PR #715, opened 2014-11-27
Add a function to prepare URL for logging

aledeg:
Please review my code to check if the fix is correct. I think that will covers our use cases. I am not very comfortable to have a modified version of SimplePie in the project. I know that it supports plugins, maybe we should look that way to make it easier for us if we need to update the library.

See #711


Alkarex:
Please see https://github.com/FreshRSS/FreshRSS/pull/815
I have moved the cleaning function to SimplePie (since it has to be used there too).

Issue #648, opened 2014-10-01
Introducing i18n.freshrss.org


marienfressinaud:
To @Alkarex and @aledeg 

I have created a new website this afternoon : http://i18n.freshrss.org
Code is available at https://github.com/marienfressinaud/i18n.freshrss.org

Objective is to facilitate translations. It's a first draft : there are surely missing features and it may doesn't work as expected (and it's not secured!).

For the moment, I have to add translation keys in `list_keys.txt` and they appear then to be translated on the website. By clicking on "Save", file is saved in `./i18n_files/`.

Later I will try to connect the application to the FreshRSS repository in order to do something like `git grep '_t'` (note I have to take care to `Minz_Translate` too) and parse the result to find automatically translation keys.

I didn't separate installation file from main one, but with this system I hope it will be possible to group translation keys easily…

What do you think about that? I see with Werner (who give a hand for the German file) if it's good for him.

ealdraed:
Hello,

I used http://i18n.freshrss.org to rework and finish the German translation of FreshRSS.

I know there is some work going on with the i18n, so maybe @marienfressinaud can easily transfer the translated strings to the _new_ multi-file i18n. That would allow to take a look at the translation _in action_.

I also notiecd some inconsistency in the English (and possibly French) languages with sentences sometimes ending with a period and sometimes not.

marienfressinaud:
Hi, 

first, thanks a lot for your contribution! Unfortunately, problem is not only transferring your translated strings into the new files (that could be easy) but there are new strings, some sentences have been improved or entirely rewritten :s. If you want, I can have a look to help you in order to transfer strings. It could be really great to have a German translation of FRSS for the version 1.0 :).

And it's my fault, I should shut-down the ability to translate on i18n.freshrss.org or, at least, add a word on the website to explain how to contribute.

sjeon87:
Is there any guide to reproduce a certain situation showing a certain message? Or should I find `_t` symbol through all files?

I am working on Korean, but a bit hard to find that which message appears when and which condition is triggered.

marienfressinaud:
@gsongsong I'm sorry but there is no such method to find locations of messages :( (unless someone has an idea…). However, I see that you finally managed to translate FRSS in Korean so thanks a lot! :) (I'll have a look to your PR later or let @Alkarex take the lead).

Btw, I close this ticket since i18n.freshrss.org is dead since a long time and there is no plan to set it up back.

PR #766, opened 2015-01-22
i18n: Add German translation

ealdraed:
Sat down again and finished translation work based on latest dev branch (see #648).


marienfressinaud:
That was quick! Thanks a lot :). It is now merged!

aledeg:
good job :+1:

PR #767, opened 2015-01-23
Fix double quotes not escaped in RSS title (share by email)

ealdraed:
If an RSS feed title contains double quotes, this prematurely terminates the subject line and prohibits the URL to be added to the email body when using Share->Email.
Replace urlencode() with rawurlencode() for all transform functions.

### DIFF
--- a/data/shares.php
+++ b/data/shares.php
@@ -19,7 +19,7 @@
 return array(
 	'shaarli' => array(
 		'url' => '~URL~?post=~LINK~&amp;title=~TITLE~&amp;source=FreshRSS',
-		'transform' => array('urlencode'),
+		'transform' => array('rawurlencode'),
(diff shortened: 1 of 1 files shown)


ealdraed:
You're right, there isn't. The main difference is that a space " " is transformed to "%20" by rawurlencode() whereas urlencode() replaces a space by "+".

The essential part here is in lines 64-67 where line 66 (transform->title) should additionally contain an encode transformation. It seems that the "+" is not decoded back to a space (at least not with Thunderbird). The same should be true for decodeURI() in JavaScript (http://stackoverflow.com/a/8591035).

Since it should do no harm, I thought about also replacing it in all other instances in data/shares.php.

Alkarex:
Ok, let's try

PR #792, opened 2015-02-12
Extract search from context

aledeg:
I figured that the code for the search could be extracted from the context to have separation of concern.
It supports multiple keywords. It supports also multiple tag keywords.

It will make filtering easier and clearer.


app/Models/Search.php
```diff
@@ -0,0 +1,187 @@
...
+	 */
+	private function parseIntitleSearch($input) {
+		if (preg_match('/intitle:(?P<delim>[\'"])(?P<search>.*)(?P=delim)/U', $input, $matches)) {
+			$this->intitle = $matches['search'];
+			$input = str_replace($matches[0], '', $input);
+		} else if (preg_match('/intitle:(?P<search>\w*)/', $input, $matches)) {
+			$this->intitle = $matches['search'];
+			$input = str_replace($matches[0], '', $input);
+		}
+		return $this->cleanSearch($input);
```
marienfressinaud: Is it really useful to call `cleanSearch()` each time? We may call it once in constructor before `$this->search = $input;`, don't you think?

aledeg: you're right. I'll change that tonight.

Issue #787, opened 2015-02-08
Add more options to use Twitter share


KillianKemps:
In the parameters, when trying to set a new sharing service. It does function for Diaspora, Blogotext, Wallabag, Shaarli.
However trying to add Google +, Facebook, Twitter, Courriel and Imprimer just adds a red cross without a text input.

In fact, it seems that they are the services positioned after Diaspora on the list.

As you see on the attached image.
![freshrsssharingbugscreenshot](https://cloud.githubusercontent.com/assets/6884630/6098589/f5203fac-afdb-11e4-8e15-367c3ce3f8f0.png)

Alwaysin:
I don't think an input is necessary for those sharing options, it works without the need of inputting something.

KillianKemps:
Oh I see, you are right.
I thought I could tweet directly clicking on the share button from the post.
In fact, it opens a new tab on Twitter to edit the message and where you need to be logged in.
My apologies.

Alwaysin:
The page could be redesigned so that it does not confuse the user, because I admit it can be confusing.

For example, why can we only choose a name for a share that uses a self-hosted service?

aledeg:
Ideas are welcome :)
I do not use twitter, facebook nor google+ so I don't know how it will make it better. So if you can share your ideas, that would be great.

Alwaysin:
I don't use them either, but from the screenshot given byKillianKemps, I think we should at least add the input which lets you rename the sharing option to every option. Maybe even add the second box, but greyed out, and labeled as "not applicable".

KillianKemps:
In my case for Twitter I would prefer to have a link that tweets directly the message with the link of the article and to have another link which lets me edit the message before publishing it. It could be better to merge the two links into one but it has to be fast to publish the default message.

As I don't like to be logged in on Twitter in the same window as FreshRSS, it could be useful to just add FreshRSS as a Twitter application, so it got the rights to publish without I have to login.

I know there is also quite lot of people who are using several Twitter accounts, so it should be easy for them to choose on which one to publish.

aledeg:
Do you have a link to the documentation that allows to register an application as a Twitter application?

KillianKemps:
To register a Twitter application, you need to log in a Twitter account and to create it at this link https://apps.twitter.com/ 

[...]

PR #793, opened 2015-02-14
Harmonize share configuration view.

aledeg:
Before, for shares that don't need options, only a button to remove it was visible. It was source of confusion for users.
I changed the look of those shares by using the same layout as others (minus the help). As there is no configuration possible for the url, the field is disabled but it is possible to change the name of the share.

See #787

### DIFF
--- a/app/i18n/en/gen.php
+++ b/app/i18n/en/gen.php
@@ -156,7 +156,7 @@ return array(
 		'damn' => 'Damn!',
 		'default_category' => 'Uncategorized',
 		'no' => 'No',
-		'not_applicable' => 'N/A',
+		'not_applicable' => 'Not available',
 		'ok' => 'Ok!',
 		'or' => 'or',
 		'yes' => 'Yes',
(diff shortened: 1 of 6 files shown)


app/i18n/en/gen.php
```diff
@@ -156,6 +156,7 @@
 		'damn' => 'Damn!',
 		'default_category' => 'Uncategorized',
 		'no' => 'No',
+		'not_applicable' => 'N/A',
```
marienfressinaud: I'm not fan of "N/A", I would prefer complete words such as "Not available" or maybe the more explicit "URL cannot be configured" (but it's a bit too long)

aledeg: I was thinking of "N/A" because people are used to it. Like when using Excel.
Anyway, it's easy to change if needed!

marienfressinaud: I always found "N/A" too formal and a bit ugly. I find "Not applicable" a lot better.

aledeg: No problem. I'll change that.

aledeg:
Please review. I changed the translation of N/A

PR #794, opened 2015-02-15
Add a cleaning process in the auto-remove feature.

aledeg:
Before, when the status change was too quick, sometimes, articles weren't removed from the view.
Now they are removed by adding a statement that grabs past articles with specific properties and remove them.

See #738


p/scripts/main.js
```diff
@@ -258,7 +253,7 @@ function toggleContent(new_active, old_active) {
 
 	if (context['sticky_post']) {
 		var prev_article = new_active.prevAll('.flux'),
-		    new_pos = new_active.position().top,
```
marienfressinaud, line 261: I use spaces instead of tabs to align variables but it's opened to discussion. We may add this to coding style rules.

Issue #610, opened 2014-09-09
[feature request] improve filter


aledeg:
If I want to filter the title with a sentence, I need to use this filter string

```
intitle:first_word intitle:second_word
```

if I use the syntax

```
intitle:first_word second_word
```

it filters the title with the first word and every thing with the second word.
if I use the syntax

```
intitle:"first_word second_word"
```

There is no match.

It could be nice to support the latest syntax to make the search more efficient.

For testing purposes, I was using that feature on Reflets.com feed with the string "un homme comme les autres"

Issue #732, opened 2014-12-17
Give possibility to manipulate database easily


marienfressinaud:
We need an abstract way to add / remove / update tables or columns. It will be helpful for repairing database during an update but also for extensions which can need to update the database schema.

It will begin to look like an ORM system but it's a very interesting feature for developers!

marienfressinaud:
In fact, I would like to refactor the DAO level before implementing such a system. I'm really not fan of the actual implementation. Since it is a big and long work, this ticket has been planned after the 1.0 release.

aledeg:
Agreed.
I think the way of filtering entries is a nightmare to maintain for instance. That's why I add few things yesterday in the context object.

marienfressinaud:
I'm no longer sure that is really needed. SQL is not that complicated, and it is more powerful than an abstraction. It would give a more integrated and nice interface to the database, but I would no longer say it's "important". At least, it is too complicated to add an ORM now.

Issue #577, opened 2014-08-12
Automated test (Travis + PHPUnit)


marienfressinaud:
I was thinking today it would be awesome to have a strong test suite. I think PHPUnit is a good framework for our tests (at least, the best known). Travis CI could be used as continuous integration platform to test automatically commits and PR. The best would be to check test coverage!

I know it's a lot of works but it can help to not lose too much time to find bugs and regressions. And it's a really good for communication :D

aledeg:
I am trying to test the class FreshRSS_Share and I wonder the reason for a private constructor and static methods. It makes testing hard and I am not sure if we can not split the class in 2.

I think the static methods are related to a Shares class which will own a Share collection. Any thoughts?

marienfressinaud:
Yes probably but I would like to avoid this kind of changes before the release of the 1.0.

aledeg:
I am now trying to test the Category class but it is highly coupled. My guess is that most of the code of FreshRSS is build the same way. If we want to be able to test properly, I think we need to refactor again the code by introducing more dependency injection and making the functions into classes.
I don't say we need to do everything in one big refactoring but I think it will be a good thing if we want to make it testable.

marienfressinaud:
I would like to try to set up travis-ci with FRSS. We'll need a .travis.yml file but I can already try to enable the Travis switch button to see what happens :).

aledeg:
I think in the future, with a nice test suite, we should consider using [humbug](https://github.com/padraic/humbug) for mutation testing.

Frenzie:
Copying this from my inadvertent duplicate:

It looks like FreshRSS isn't making the most of GitHub integration yet. Here are some hopefully helpful suggestions to get started.

* [Scrutinizer](https://scrutinizer-ci.com/tour/eliminate-bugs-before-they-hit-production) (code quality)
* [Codacy](https://en.wikipedia.org/wiki/Codacy) (code quality)
* [CircleCI](https://circleci.com/) (or Travis, for running unit tests and custom coding style/quality tests)
  See [rss-bridge](https://github.com/RSS-Bridge/rss-bridge/blob/8ed4812e00e6f6bb55a3ee6ab374a4642b135a62/.travis.yml#L12-L13) and [Elkarte](https://github.com/elkarte/Elkarte/tree/8cac408e6726a64ee81294a915b3d80962cb829c/tests)

Of course I'm willing to help think about and implement these.

Alkarex:
[...]

PR #796, opened 2015-02-21
Improve filter

aledeg:
Now you can search for intitle:'something with spaces'.
The search splitting is done in the search object instead of the DAO.

See #610

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -441,7 +441,7 @@ class FreshRSS_EntryDAO extends Minz_ModelPdo {
 			$where .= 'AND e1.id >= ' . $date_min . '000000 ';
 		}
 		$search = '';
-		if ($filter instanceof FreshRSS_Search) {
+		if ($filter !== null) {
 			if ($filter->getIntitle()) {
 				$search .= 'AND e1.title LIKE ? ';
 				$values[] = "%{$filter->getIntitle()}%";
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -441,53 +441,45 @@ class FreshRSS_EntryDAO extends Minz_ModelPdo {
 			$where .= 'AND e1.id >= ' . $date_min . '000000 ';
 		}
 		$search = '';
-		if ($filter !== '') {
-			require_once(LIB_PATH . '/lib_date.php');
-			$filter = trim($filter);
-			$filter = addcslashes($filter, '\\%_');
-			$terms = array_unique(explode(' ', $filter));
-			//sort($terms);	//Put #tags first	//TODO: Put the cheapest filters first
-			foreach ($terms as $word) {
-				$word = trim($word);
-				if (stripos($word, 'intitle:') === 0) {
-					$word = substr($word, strlen('intitle:'));
-					$search .= 'AND e1.title LIKE ? ';
-					$values[] = '%' . $word .'%';
-				} elseif (stripos($word, 'inurl:') === 0) {
-					$word = substr($word, strlen('inurl:'));
-					$search .= 'AND CONCAT(e1.link, e1.guid) LIKE ? ';
-					$values[] = '%' . $word .'%';
-				} elseif (stripos($word, 'author:') === 0) {
-					$word = substr($word, strlen('author:'));
-					$search .= 'AND e1.author LIKE ? ';
-					$values[] = '%' . $word .'%';
-				} elseif (stripos($word, 'date:') === 0) {
-					$word = substr($word, strlen('date:'));
-					list($minDate, $maxDate) = parseDateInterval($word);
-					if ($minDate) {
-						$search .= 'AND e1.id >= ' . $minDate . '000000 ';
-					}
-					if ($maxDate) {
-						$search .= 'AND e1.id <= ' . $maxDate . '000000 ';
-					}
-				} elseif (stripos($word, 'pubdate:') === 0) {
-					$word = substr($word, strlen('pubdate:'));
-					list($minDate, $maxDate) = parseDateInterval($word);
-					if ($minDate) {
-						$search .= 'AND e1.date >= ' . $minDate . ' ';
-					}
-					if ($maxDate) {
-						$search .= 'AND e1.date <= ' . $maxDate . ' ';
-					}
-				} else {
-					if ($word[0] === '#' && isset($word[1])) {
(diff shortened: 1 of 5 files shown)


app/Models/EntryDAO.php
```diff
@@ -441,54 +441,46 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
...
-					}
-				} else {
-					if ($word[0] === '#' && isset($word[1])) {
-						$search .= 'AND e1.tags LIKE ? ';
-						$values[] = '%' . $word .'%';
-					} else {
-						$search .= 'AND ' . $this->sqlconcat('e1.title', $this->isCompressed() ? 'UNCOMPRESS(content_bin)' : 'content') . ' LIKE ? ';
-						$values[] = '%' . $word .'%';
-					}
+		if ($filter instanceof FreshRSS_Search) {
```
marienfressinaud: Don't test with "instanceof". Method signature should indicate $filter is FreshRSS_Search or null. Test should be `if ($filter !== null)`

aledeg: I made the change but I am not sure if it's the proper way to do it. In my mind, it is better to test for an interface (which we don't have now but I am working on it) than testing for null. Because if for any reason we have something that is not what we are expecting, it will break.
At the moment, it is not a big deal because we are in charge of the code, but if there is extensions that inject bad things, this will break.

marienfressinaud: Errors are normal if you don't respect method signature. That's not the problem of this method but the responsibility of the caller. What we need here is a comment to define the signature.

aledeg: Agreed. That the use of the interface when using dependency injection.
Anyway, I made the change to test against null value.

aledeg:
Any chance this will be reviewed and merge before Sunday? There is things in that PR I am waiting for to work on other things.
Thank you

marienfressinaud:
Done! Sorry about that, I'm a bit busy these days ;).

aledeg:
Don't worry, we all do. There was no rush since I didn't have time to do anything.
Thank you.

Alwaysin:
Could this possibly break API? After updating, it seems I cannot sync anymore.

aledeg:
Maybe, I forgot to check the API. I never think of that since I am not using it.
I will check that.
That's why we need #732 . There should be only one place where we access database.
That's also why we need #577 to avoid regressions

aledeg:
@Alwaysin do you have any errors? logs? you can share. What reader are you using?

Alkarex:
Quick message: The API does not "access the database" (i.e. does not write SQL requests). It just uses the "model" classes, which probably have been broken by recent changes.
A way to debug is to activate the logs for the database, and check that the SQL requests are exactly identical compared to before those changes. Watch out that even subtle changes in the SQL syntax may have huge performance impacts. If the SQL requests are to be changed, please check with a very large dataset.

aledeg:
There was no changes in the queries. Only the parameter extraction is removed from the DAO to be included in a separate object.

app/Models/EntryDAO.php
```diff
@@ -441,54 +441,46 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
...
+			if ($filter->getAuthor()) {
+				$search .= 'AND e1.author LIKE ? ';
+				$values[] = "%{$filter->getAuthor()}%";
+			}
+			if ($filter->getMinDate()) {
+				$search .= 'AND e1.id >= ? ';
+				$values[] = "{$filter->getMinDate()}000000";
+			}
+			if ($filter->getMaxDate()) {
+				$search .= 'AND e1.id <= ?';
```
aledeg, line 462: Might be missing a space here

aledeg:
@Alwaysin are you using a query with a max date?

Alwaysin:
I have this in /data/users/_/log_api.txt : 

```
----------------------------------------------------------------
User => alwaysin
tagList()
----------------------------------------------------------------
User => alwaysin
subscriptionList()
----------------------------------------------------------------
User => alwaysin
streamContents(reading-list, , 1425205822, 500, n, user/-/state/com.google/read, )
```

I'm using News+. I am not sure what you mean by "are you using a query with a max date?" ; I'm just syncing with the app, it shouldn't use a max date from what I understand (no parameter in the app stating "I want no article from last week", etc).

aledeg:
Thank you.
I noticed that I forgot a space in the max date query. That's why I wonder if this was the root of the problem.
So far, I have no ideas how the API works. But while reading the code, there is no change I need to apply there. @Alkarex confirmed that already.

Alwaysin:
Manually applied the correction, but still no luck. Logs look the same.

aledeg:
I will have a look at the query in the API and how it gets its parameters. The problem is probably there.
I'll let you know.
I hope to have some time tonight!

Alwaysin:
Ok super great, thanks :) Feel no rush though, I can live with it some days!

Alkarex:
If you use MySQL with the logs enabled, there are likely some errors in
them.

aledeg:
Is there a way to use the API without using a phone?

Alwaysin:
@Alkarex I have this : 

```
150303 17:49:06    41 Connect   freshrss@localhost on freshrss
                   41 Query     SET NAMES utf8
                   41 Query     SELECT c.name FROM `freshrss_alwaysin_category` c
                   41 Quit
                   42 Connect   freshrss@localhost on freshrss
                   42 Query     SET NAMES utf8
                   42 Query     SELECT f.id, f.name, f.url, f.website, c.id as c_id, c.name as c_name FROM `freshrss_alwaysin_feed` f
                INNER JOIN `freshrss_alwaysin_category` c ON c.id = f.category
                   42 Quit
                   43 Connect   freshrss@localhost on freshrss
                   43 Query     SET NAMES utf8
                   43 Query     SELECT f.id, f.name, c.name as c_name FROM `freshrss_alwaysin_feed` f INNER JOIN `freshrss_alwaysin_category` c ON c.id = f.category
                   43 Quit
```

Nothing in error.log :(

Alkarex:
@aledeg It is not trivial to test manually, but an option is to run one of the Android apps in an emulator.

aledeg:
@Alwaysin I made a change in the API. I didn't test it since I did not take the time to set up an app emulator.
Let me know!

Alwaysin:
It works perfect! Thank you very much :D

PR #799, opened 2015-03-01
Introduce user queries objects

aledeg:
There is now an object to manipulate user queries. It allows to move logic to handle those from the view and the controller in the model.
Thus making the view and the controller easier to read.
I introduced a new interface to start using dependency injection. There is still some rough edges but we are moving in the right direction.
The new object is fully tested but it still need some improvements, for instance, it is still tied to the search object.
There might be a better way to do that.

### DIFF
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -282,6 +282,7 @@ class FreshRSS_configure_Controller extends Minz_ActionController {
 			$queries[$key] = new FreshRSS_UserQuery($query, $feed_dao, $category_dao);
 		}
 		$params = Minz_Request::params();
+		$params['url'] = Minz_Url::display(array('params' => $params));
 		$params['name'] = _t('conf.query.number', count($queries) + 1);
 		$queries[] = new FreshRSS_UserQuery($params, $feed_dao, $category_dao);
 
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -241,13 +241,16 @@ class FreshRSS_configure_Controller extends Minz_ActionController {
 	 * checking if categories and feeds are still in use.
 	 */
 	public function queriesAction() {
+		$category_dao = new FreshRSS_CategoryDAO();
+		$feed_dao = FreshRSS_Factory::createFeedDao();
 		if (Minz_Request::isPost()) {
-			$queries = Minz_Request::param('queries', array());
+			$params = Minz_Request::param('queries', array());
 
-			foreach ($queries as $key => $query) {
+			foreach ($params as $key => $query) {
 				if (!$query['name']) {
 					$query['name'] = _t('conf.query.number', $key + 1);
 				}
+				$queries[] = new FreshRSS_UserQuery($query, $feed_dao, $category_dao);
 			}
 			FreshRSS_Context::$user_conf->queries = $queries;
 			FreshRSS_Context::$user_conf->save();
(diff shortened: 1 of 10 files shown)


marienfressinaud:
Seems really nice! I don't have time to review your PR before Wednesday but it looks like a great improvement :).

aledeg:
Thanks :)

aledeg:
I fixed few bugs I missed during the first batch of devs. Sorry about that.

app/Models/UserQuery.php
```diff
@@ -0,0 +1,218 @@
...
+	 * @param FreshRSS_Searchable $category_dao
+	 */
+	public function __construct($query, FreshRSS_Searchable $feed_dao = null, FreshRSS_Searchable $category_dao = null) {
+		$this->category_dao = $category_dao;
+		$this->feed_dao = $feed_dao;
+		if (isset($query['get'])) {
+			$this->parseGet($query['get']);
+		}
+		if (isset($query['name'])) {
+			$this->name = $query['name'];
```
marienfressinaud, line 35: Name should be required, isn't it? In this case `isset` test should be facultative. But it's ok like that too ;)

app/Models/UserQuery.php
```diff
@@ -0,0 +1,218 @@
...
+		}
+		if (isset($query['order'])) {
+			$this->order = $query['order'];
+		}
+		if (!isset($query['search'])) {
+			$query['search'] = '';
+		}
+		// linked to deeply with the search object, need to use dependency injection
```
marienfressinaud, line 43: Do you plan to use dependency injection here?

app/Models/UserQuery.php
```diff
@@ -0,0 +1,218 @@
...
+		$this->get_type = 'all';
+	}
+
+	/**
+	 * Parse the query string when it is a "category" query
+	 * 
+	 * @param integer $id
+	 */
+	private function parseCategory($id) {
+		$category = $this->category_dao->searchById($id);
```
marienfressinaud, line 113: You should test `category_dao` is not null since it is optional in the constructor (i.e. default value at null)

app/Models/UserQuery.php
```diff
@@ -0,0 +1,218 @@
...
+		$this->get_type = 'category';
+	}
+
+	/**
+	 * Parse the query string when it is a "feed" query
+	 * 
+	 * @param integer $id
+	 */
+	private function parseFeed($id) {
+		$feed = $this->feed_dao->searchById($id);
```
marienfressinaud, line 132: Same here

marienfressinaud:
I really like your new implementation! I added some comments but there is still a bug on the main page: you didn't update `./app/layout/nav_menu.phtml` so the queries cannot be used. Links on the configuration page don't work neither.

app/Controllers/configureController.php
```diff
@@ -241,13 +241,16 @@ public function archivingAction() {
 	 * checking if categories and feeds are still in use.
 	 */
 	public function queriesAction() {
```
marienfressinaud, line 243: Don't you think query-related action could be moved into a dedicated controller? It's just an idea but it can help to reduce `configureController` file size.

app/Models/UserQuery.php
aledeg, line 113: You're right

app/Controllers/configureController.php
aledeg, line 243: That's a good idea. I think that we should move gradually, while introducing unit tests. This way refactoring would be easier.

app/Models/UserQuery.php
aledeg, line 43: I am not sure to understand your question. Do you mean for the search object?
If your question is related to the search object, yes and no. I want the constructor not be dependent from that object but still want to use it. I want to use a container that holds the search object as well as DAO. I think the perfect location for that is in the context object.
This way we can remove lots of dependencies by using DAO from the context instead of recreating one every time we need one.
The new constructor signature would be `__construct($query, FreshRSS_Context $context)`

aledeg:
Did you get the latest changes? Because I noticed also that links weren't working but I fixed that already in a573954

marienfressinaud:
Ok I see, problem happens after updating queries: we should keep url as an hidden input in the form.

app/Models/UserQuery.php
marienfressinaud, line 43: Ok, you answer my question :).

aledeg:
I see. I will work on that tonight

PR #801, opened 2015-03-04
Fix API to use the search object

aledeg:
Since the internal of the listWhere method was changed, the API wasn't working. It was still calling the method with the old parameters.
I didn't test it but now, it should work.

See #796

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -371,7 +371,7 @@ function streamContents($path, $include_target, $start_time, $count, $order, $ex
 	}
 
 	$entryDAO = FreshRSS_Factory::createEntryDao();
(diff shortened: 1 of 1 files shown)


Alwaysin:
I've spoken too quick... The behavior is erratic. I don't even know how to explain what's happening. 
I deleted my account on News+. When signing in, I got all my articles. When marking an article read in the app, I don't see it in the web, which is normal, but when I mark an article read in the web, it stays unread in the app...

aledeg:
@Alwaysin the new fix should solve your problem. I hope!

Alwaysin:
Fixed :) Thanks!
Hoping there won't be any more bugs ^^

aledeg:
Me too :)
It is hard to work blindly

PR #802, opened 2015-03-05
Introduce user queries objects

aledeg:
There is now an object to manipulate user queries. It allows to move logic to handle those from the view and the controller in the model.
Thus making the view and the controller easier to read.
I introduced a new interface to start using dependency injection. There is still some rough edges but we are moving in the right direction.
The new object is fully tested but it still need some improvements, for instance, it is still tied to the search object.
There might be a better way to do that.

### DIFF
--- a/app/Exceptions/DAOException.php
+++ b/app/Exceptions/DAOException.php
@@ -1,5 +1,5 @@
 <?php
 
-class FreshRSS_DAOException extends \Exception {
+class FreshRSS_DAO_Exception extends \Exception {
 
 }
--- a/app/Exceptions/DAOException.php
+++ b/app/Exceptions/DAOException.php
@@ -1,0 +1,5 @@
+<?php
+
+class FreshRSS_DAOException extends \Exception {
+
+}
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -1,6 +1,6 @@
 <?php
 
-class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable{
+class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable {
 	public function addCategory($valuesTmp) {
 		$sql = 'INSERT INTO `' . $this->prefix . 'category`(name) VALUES(?)';
 		$stm = $this->bd->prepare($sql);
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -1,6 +1,6 @@
 <?php
 
-class FreshRSS_CategoryDAO extends Minz_ModelPdo {
+class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable{
 	public function addCategory($valuesTmp) {
 		$sql = 'INSERT INTO `' . $this->prefix . 'category`(name) VALUES(?)';
 		$stm = $this->bd->prepare($sql);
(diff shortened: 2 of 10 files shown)


aledeg:
I made a new PR and I squash all my small commits into one.
It make things clearer in the git history.
I included everything we talk about in the other PR. I should be ok now.

app/Exceptions/DAOException.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+class FreshRSS_DAOException extends \Exception {
```
marienfressinaud: Could you add an underscore after `DAO`?

app/Models/CategoryDAO.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class FreshRSS_CategoryDAO extends Minz_ModelPdo {
+class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable{
```
marienfressinaud: There is a missing space before `{`

app/Models/EntryDAO.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class FreshRSS_EntryDAO extends Minz_ModelPdo {
+class FreshRSS_EntryDAO extends Minz_ModelPdo implements FreshRSS_Searchable{
```
marienfressinaud: Same here

app/Models/FeedDAO.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class FreshRSS_FeedDAO extends Minz_ModelPdo {
+class FreshRSS_FeedDAO extends Minz_ModelPdo implements FreshRSS_Searchable{
```
marienfressinaud: Same here

app/Models/UserQuery.php
```diff
@@ -0,0 +1,226 @@
...
+
+	/**
+	 * Parse the query string when it is a "category" query
+	 *
+	 * @param integer $id
+	 * @throws FreshRSS_DAOException
+	 */
+	private function parseCategory($id) {
+		if (is_null($this->category_dao)) {
+			throw new FreshRSS_DAOException('Category DAO is not loaded i UserQuery');
```
marienfressinaud: "i" UserQuery?

app/Models/UserQuery.php
```diff
@@ -0,0 +1,226 @@
...
+
+	/**
+	 * Parse the query string when it is a "feed" query
+	 *
+	 * @param integer $id
+	 * @throws FreshRSS_DAOException
+	 */
+	private function parseFeed($id) {
+		if (is_null($this->feed_dao)) {
+			throw new FreshRSS_DAOException('Feed DAO is not loaded i UserQuery');
```
marienfressinaud: Same here

marienfressinaud:
Sorry I missed a bunch of details yesterday.

I'm not fan of squashing commits: we lose the work process (i.e. how the work has been done). The main argument is usually to clean history but it seems to me that is the role of creating a dedicated branch.

aledeg:
I agree for the work process but when it comes for minor changes, I find that it is too verbose.
My two cents.

marienfressinaud:
I mess up a bit my merging but everything seems to work.

marienfressinaud:
I merged your work locally and rebased it on your latest changes (#803) and push it so it is the reason the ticket has not been merged properly.

PR #804, opened 2015-03-05
Refactor exceptions

aledeg:
I removed unnecessary constructors and unnecessary inheritance

### DIFF
--- a/app/Exceptions/BadUrlException.php
+++ b/app/Exceptions/BadUrlException.php
@@ -1,6 +1,6 @@
 <?php
 
-class FreshRSS_BadUrl_Exception extends \Exception {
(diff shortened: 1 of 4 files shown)


app/Exceptions/BadUrlException.php
```diff
@@ -1,6 +1,9 @@
 <?php
-class FreshRSS_BadUrl_Exception extends FreshRSS_Feed_Exception {
+
+class FreshRSS_BadUrl_Exception extends \Exception {
```
marienfressinaud: Mmm, did you check there is no problem by changing inheritance here? I'm pretty there are some `FreshRSS_Feed_Exception` catch which are used to catch `FreshRSS_BadUrl_Exception` too.

PR #812, opened 2015-03-14
Add an unsaved changes alert on config pages

aledeg:
Before, you could leave a configuration page without knowing if you saved your changes or not.
Now, there is an alert poping up if you have unsaved changes. It will ask you if you want to stay on the page and save your changes or leave the page and loose your changes.

See #739


Alkarex:
Looks good :-)

Issue #754, opened 2015-01-17
Facebook's RSSes aren't displayed correctly


ghost:
One picture is worth a thousand words :

![frss](https://cloud.githubusercontent.com/assets/3429231/5789235/8f046156-9e62-11e4-94f2-030cb8c7db89.png)

Here is the RSS that causes problems :

https://www.facebook.com/feeds/page.php?format=rss20&id=166081103477984

aledeg:
It looks like an encoding problem. When I open the feed with Firefox, I have the same problem. I think it is not specific to FRSS.
I checked also the feed on the W3C validator and it is not valid. It is enough valid to be parsed though.

marienfressinaud:
> It is enough valid to be parsed though.

Yes and SimplePie demo and Liferea work great: that's why I ask @AugierLe42e to open a ticket. I didn't inspect code deeper but it could be a problem from FRSS. By the way, this problem reminds me another but I cannot find it in ticket archive.

Alwaysin:
Maybe this one : #661 ? It is still opened.

Alkarex:
This RSS2 feed uses a [CDATA container](http://en.wikipedia.org/wiki/CDATA#Issues_with_encoding) together with HTML-entities in the `<title>` section which is [supposed to be text-only](http://validator.w3.org/feed/docs/warning/ContainsHTML.html) (see comment below), not HTML. As far as I can see, the result in FrehsRSS is as per the specifications (XML, RSS2).
Note however that I cannot immediately find an official reference stating explicitly that `<title>` is text-only and not HTML (except the note from the W3C validator, which is already quite a good reference). This is what I lack to be sure.
If `<title>` is indeed text-only, then I do not believe it is possible to support this feed without breaking other valid feeds and without doing an exception for Facebook.

There is also the Atom 1.0 version of the same feed: https://www.facebook.com/feeds/page.php?format=atom10&id=166081103477984
In this case it is clear, and the Facebook Atom feed is wrong: http://tools.ietf.org/html/rfc4287#section-3.1.1

> If the "type" attribute is not provided, Atom Processors MUST behave as though it were present with a value of "text"

ghost:
Here is another one causing problem :

http://www.atterres.org/rss.xml

Alkarex:
[...]

PR #813, opened 2015-03-16
SimplePie: decode special chars for MAYBE_HTML

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/754
Needs to check with many feeds to see if this does not introduce
incompatibilities with some valid feeds.

### DIFF
--- a/lib/SimplePie/SimplePie/Sanitize.php
+++ b/lib/SimplePie/SimplePie/Sanitize.php
@@ -249,6 +249,11 @@ class SimplePie_Sanitize
 		{
 			if ($type & SIMPLEPIE_CONSTRUCT_MAYBE_HTML)
 			{
+				if (preg_match('/&amp;#(x[0-9a-fA-F]+|[0-9]+);/', $data))	//FreshRSS
+				{
+					$data = htmlspecialchars_decode($data, ENT_QUOTES);	//FreshRSS
+					//syslog(LOG_DEBUG, 'SimplePie sanitize MAYBE_HTML htmlspecialchars_decode: ' . $data);	//FreshRSS
+				}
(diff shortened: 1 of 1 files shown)


Alkarex:
I have been using this patch and it seems to work fine :-)
Anybody would like more time to test? Otherwise I will merge during the week-end.

aledeg:
Let me do it since we shouldn't merge our own commits :) 

See the
new rules for more info 

Le 2015-03-21 7:27, Alexandre Alapetite a
écrit : 

> I have been using this patch and it seems to work fine :-)
> 
> Anybody would like more time to test? Otherwise I will merge during the
> week-end. 
> 
> ## 
> 
> Reply to this email directly or view it on GitHub
> [1].

## Links:

[1]
https://github.com/FreshRSS/FreshRSS/pull/813#issuecomment-84317854

Alkarex:
I have just found an example of an RSS2 feed ([Clubic](http://com.clubic.feedsportal.com/c/33464/f/581983/index.rss)) with an escaped HTML character entity reference:

``` xml
<title>NVIDIA GeForce GTX Titan X : la plus rapide à 1 200 &amp;euro; ?</title>
```

The current patch only handles escaped numeric character references (e.g. `&amp;#160;`). I am considering extending it to the above case, but this has higher risk of unwanted side effect.

Edit: new version https://github.com/Alkarex/FreshRSS/commit/0e08b5ba56b2117e0f3d9bea6ae725295e8700d5

aledeg:
So should we merge that now or should we wait? I haven't tested it and I probably won't have time to do it before weekend

Alkarex:
@aledeg I am running with this patch on my instance, and so far everything is good with my ~200 feeds. It would be nice to get some feedback from more users.

aledeg:
ok, I'll wait 'til we have more feedback

Alkarex:
Actually, I was trying to say that it was good to merge to get more feedback. I doubt anybody will test a branch not merged in /dev.

Alwaysin:
I haven't encountered any problems with my ~40 feeds.

Alkarex:
Thanks for the feedback @Alwaysin; sounds good :-)

Alkarex:
Reference to SimplePie's identical bug https://github.com/simplepie/simplepie/issues/350

Alkarex:
https://github.com/simplepie/simplepie/pull/400

Issue #798, opened 2015-02-28
Thread update


Pierrotd:
Hi,
In Tiny Tiny RSS, updated threads come back in the principal stream as non read.
Is it be possible to add this option in FreshRSS?
Thanks

Alkarex:
That could be possible. We would have to change a bit our approach to determinate the unicity of an article.
There could two alternative approaches at least:
1. Mark the old article as unread, and change its content with the new version;
2. or make a new article.

Alkarex:
To test the second alternative, could can try to change to `$item->get_id(true),`
https://github.com/FreshRSS/FreshRSS/blob/beta/app/Models/Feed.php#L302
(Not tested, so proceed with caution)

Alkarex:
Everybody is welcome to comment/vote the two possible options above regarding possible behaviours when feed entries are modified.

Alwaysin:
I am in favor of the first option; I don't want "duplicating" articles.
But we could have a threaded view of an article with the different versions able to be reviewed.

Else, the first option seems good, but we should add an indicator saying the article has been updated (icon, text added in front of the title of the article, or whatever).

Pierrotd:
I think the first option is the best (Mark the old article as unread, and change its content with the new version)

Alkarex:
Example of feed updating old entries: 01Net http://rss.feedsportal.com/c/629/f/502199/index.rss
It looks like articles such as http://www.01net.com/editorial/624966/applis-android-iphone-ipad-les-bons-plans/ and http://www.01net.com/editorial/638020/01net-evening-les-5-infos-quil-ne-fallait-pas-rater-aujourdhui/ keep their GUID and are regularly updated.

Another example at OVH http://travaux.ovh.net/rss.php

Alkarex:
I have started the work on that and should have a patch soon. I need to address other pending issues at the same time, e.g. a better way to purge old articles, because now we have old articles that may get updated, and because our old approach was not robust enough anyway.
https://github.com/FreshRSS/FreshRSS/issues/493
https://github.com/FreshRSS/FreshRSS/blob/e2da6e6e6b871dc3dbc289cdd40ba401a21d8e91/app/Controllers/feedController.php#L335

Alkarex:
Example of problematic feed for change detection, which makes irrelevant random hidden changes to the content of items: LeMonde.fr http://rss.lemonde.fr/c/205/f/3050/index.rss

``` xml
[...]

PR #815, opened 2015-03-22
SimplePie logging of HTTP requests

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/711
New option (in FreshRSS and in SimplePie) to turn on/off the logging (via syslog) of HTTP requests done by SimplePie.
Passwords are removed from URLs before logging. The regular expression has been slightly updated since https://github.com/FreshRSS/FreshRSS/pull/715 (should be faster and safer), taking advantage of RFC 3986.

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -240,7 +240,7 @@ class FreshRSS_Feed extends Minz_Model {
 					$subscribe_url = $feed->subscribe_url(true);
 				}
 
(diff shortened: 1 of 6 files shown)


Alkarex:
Any comment on this patch? I have to make some changes for https://github.com/FreshRSS/FreshRSS/issues/798, which will conflict with this patch if not committed before.

Issue #819, opened 2015-04-05
Can't add RSS feed with Cyrillic URL -> support idn urls for feeds


Trim:
Hello,

I've just read about next Thunderbird release (v38) and I've seen an interesting bug about the RSS reader: [support idn urls for feeds](https://bugzilla.mozilla.org/show_bug.cgi?id=1018589).

Apparently, URLs like http://партиявеликоеотечество.рф/feed are valid and should be possible to read by FreshRSS.I don't know Cyrillic, but I think it could be great to have support of UTF8 in FreshRSS (apprently the Mozilla's patch just add UTF8 support: https://bug1018589.bugzilla.mozilla.org/attachment.cgi?id=8439925).

I'm using FreshRSS 1.1.0 and after I've added [this feed](http://партиявеликоеотечество.рф/feed) I've read this in my logs:

> [Sun, 05 Apr 2015 07:06:51 +0200] [warning] --- `` is not a valid URL

and I had a poup saying something like this:

> The url http://партиявеликоеотечество.рф/feed is invalid

Otherwise FreshRSS is really a great RSS reader, I'm enjoyed to use it ! Many thanks for the great work :)

Note: that's not really like the bug #293 or #301  because it's UTF8 in the URL not in articles/title

Alkarex:
Thanks for the bug report. I will add support for IDNs.

Alkarex:
Ok, I have done a first version that is working with your example (which is automatically converted to Punycode `http://xn--80aealaaatcuj2bbdbr2agjd6hzh.xn--p1ai/feed` for internal compatibility - see [PHP bug 53474](https://bugs.php.net/bug.php?id=53474)). If you can, please test https://github.com/FreshRSS/FreshRSS/pull/820

The current version requires PHP 5.3+ IDN extension http://php.net/intl.idn (`php5-idn` package
on Debian/Ubuntu).

A possible future improvement could be to convert from Punnycode back to IDN when showing back URLs in the Web user interface.

Alkarex:
Thanks for the feedback, @Trim :-)

Alkarex:
Corrected in /dev branch

PR #820, opened 2015-04-05
Support for Internationalized Domain Names (IDN)

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/819
Add explicit conversion from IDN to Punycode.

Requires PHP 5.3 IDN extension http://php.net/intl.idn (php5-idn package
on Debian/Ubuntu).

For systems without PHP 5.3+ IDN extension, we may consider adding a
dependency (322 kB) to the third-party library
https://phlymail.com/en/downloads/idna-convert.html

See PHP bug 53474 FILTER_VALIDATE_URL should not fail URL's that use
IDN https://bugs.php.net/bug.php?id=53474


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/819

Issue #823, opened 2015-04-21
SQL: Search problem


Alkarex:
Somewhere during a recent refactoring, the search behaviour has been changed. It does not seem possible anymore to search for multiple independent keywords.

Now when searching for e.g. `employee stack`, the SQL is `LIKE '%employee stack%'`, meaning that the search input is treated like it had quotes `"employee stack"`.

aledeg:
I will have a look at that

Alkarex:
Thanks :-) BTW, it would be nice to be able to search for `"employee stack"` too, in the future

aledeg:
that's my plan :)

Alkarex:
FWIW, the latest version of Ubuntu (15.04) ships with MySQL 5.6, first version of MySQL to support full-text indexes on InnoDB http://dev.mysql.com/doc/refman/5.6/en/fulltext-search.html
https://github.com/FreshRSS/FreshRSS/issues/302

Alkarex:
Solved by @aledeg

PR #825, opened 2015-04-22
Split the search into values

aledeg:
Before, the search was a single value.
Now it is splited in chuncks when separated by spaces.
Except if they are enclosed by single quotes or double quotes.

For some reasons, the unit tests are working for both single and double quotes but the
search box isn't. It is working only with single quotes.
We need to investigate the reason of this behavior.

See #823

### DIFF
--- a/app/Models/Search.php
+++ b/app/Models/Search.php
@@ -34,9 +34,9 @@ class FreshRSS_Search {
 		$input = $this->parsePubdateSearch($input);
 		$input = $this->parseDateSearch($input);
 		$input = $this->parseTagsSeach($input);
-		$this->search = $this->cleanSearch($input);
+		$this->parseSearch($input);
 	}
-	
+
 	public function __toString() {
 		return $this->getRawInput();
 	}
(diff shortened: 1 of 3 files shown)


app/Models/Search.php
```diff
@@ -188,6 +188,34 @@ private function parseTagsSeach($input) {
...
+	 *
+	 * @param string $input
+	 * @return string
+	 */
+	private function parseSearch($input) {
+		$input = $this->cleanSearch($input);
+		if (strcmp($input, '') == 0) {
+			return;
+		}
+		if (preg_match_all('/(?P<delim>[\'"])(?P<search>.*)(?P=delim)/U', $input, $matches)) {
```
Alkarex, line 203: ~~I have not tested yet, but just by reading quickly, isn't this regular expression greedy? I think it should search for the shortest `"..."`, not the longest `"..." ... "..."`. A suggestion is to replace the middle mask with `(?P<search>*?)`~~
Edit: Ah, sorry, I did not see the `PCRE_UNGREEDY` flag at the end.

Alkarex:
Seems fine, thanks :-)

Issue #770, opened 2015-01-25
Cannot create an account with sqlite


dhoko:
Hello,

I cannot create an account with freshRSS and a sqlite database. There is an error when i'm clicking on Validate:

`An invalid form control with name='user' is not focusable.`

![capture du 2015-01-25 19 08 27](https://cloud.githubusercontent.com/assets/713283/5892598/b68a0a28-a4c5-11e4-9102-66964d21dc4e.png)

Edit: It works with firefox.

marienfressinaud:
Had you filled the username field before changing the database field? I can reproduce the bug by this way:
- Fill username field with incorrect characters (e.g. "èèèè" so the pattern `[0-9A-Za-z_.-]{1,16}` doesn't match).
- Change the database type to "SQLite"
- Try to submit

Chrome/Chromium shows an error in the console while Firefox shows a box to indicate the field is not valid in the top left corner of the window (which is very strange btw).

It could be easy to fix that by renaming the pattern attribute at the `mySqlShowHide()` JS function call (e.g. rename all the attributes `pattern` into `data-pattern` so it could be easy to reverse it when switching back to MySQL database).

It's not a critical issue. I'll fix that if I have time ;).

dhoko:
Yup. I did the same process with the same credentials with firefox and chrome.
username: dhoko
password: test123
sqlite

It worked on Firefox.

marienfressinaud:
When I said "username", I spoke about database username which was, I guess, the cause of the bug. I think, with Chrome, you filled the database username with wrong characters (a "é" for example) before switching on SQLite. Chrome was not able to validate the form nor showing your error because concerned field was hidden (`display: none`)

dhoko:
So you have to clean user input. :) If I choose to use sqlite, it doesn't have to check if the mySQL username is set or not, it has to ignore it.

marienfressinaud:
Yep, sure, it's what I was saying in my first post but I don't have time to do that immediately (I am writing some doc about Minz and extension system ;)). However, it seems to me it is an easy task (only some JavaScript code to write in install script)

Alkarex:
@dhoko This should be solved in /dev branch. Thanks again for the report.
https://github.com/FreshRSS/FreshRSS/pull/827

Alkarex:
@dhoko The correction is now merged in /dev.

PR #827, opened 2015-05-10
Cannot create an account with sqlite

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/770


Alkarex:
Grr, I again merged in FreshRSS/dev when it should have been my own /dev... Sorry, I will try to be more careful... Anyway, let me know if there is any comment on this patch.

PR #835, opened 2015-05-16
Typo fix

Alwaysin:


Alwaysin:
Thanks Alkarex.

Also, not sure, but "Attention!" in english is not the same that in french, shouldn't we rename it to "Careful!"?

Alkarex:
Indeed @Alwaysin, probably "Warning!"

Alwaysin:
Hmm, I've also just noted that there is a mix between British and American English (_actualize_ -> en_US, _favourite_ -> en_GB).

Should we stick with EN_US, which seems to be the default for international communication? Or even provide two translations?

But maybe we should create a ticket about it.

Alkarex:
Being in Europe, I use en_GB, but at some time we could have both, indeed.

PR #836, opened 2015-05-16
Language fix + en_GB standardisation

Alwaysin:


app/i18n/en/gen.php
```diff
@@ -14,7 +14,7 @@
 		'import' => 'Import',
 		'manage' => 'Manage',
 		'mark_read' => 'Mark as read',
-		'mark_favorite' => 'Mark as favourite',
+		'mark_favorite' => 'Mark as favorite',
```
Alkarex, line 17: If the rest of the file is en_GB, then it is _favourite_ https://en.wiktionary.org/wiki/favourite
_favorite_ is en_US

app/i18n/en/gen.php
```diff
@@ -2,7 +2,7 @@
 
 return array(
 	'action' => array(
-		'actualize' => 'Actualize',
+		'actualize' => 'Actualise',
```
Alkarex, line 5: Oxford's en_GB is _actualize_ https://en.wiktionary.org/wiki/actualise but the -ise form is fine too in en_GB

app/i18n/en/gen.php
Alwaysin, line 17: Erf, I've always thought it was "o" for British and "ou" for American. I got it all wrong all my life during... but anyway, thanks, I've learned something today :)

What shall we do? I close all PR with this error and re-make them?

Alkarex, line 17: Haha :-) Yes, that is the easiest I think.

Issue #443, opened 2014-03-02
API: Improve Google Reader API level 2 compatibility


Alkarex:
More specific follow up of https://github.com/FreshRSS/FreshRSS/issues/13 and https://github.com/FreshRSS/FreshRSS/issues/328

FreshRSS already supports quite well [News+](https://play.google.com/store/apps/details?id=com.noinnion.android.newsplus) (https://github.com/noinnion/newsplus ).

It was also tested against a [modified version](https://github.com/Alkarex/gReader-Library) of https://github.com/ericmann/gReader-Library

It is currently being tested (~~alpha~~ beta quality) against a [fork of EasyRSS](https://github.com/Alkarex/EasyRSS) (https://play.google.com/store/apps/details?id=com.pursuer.reader.easyrss).

To improve compatibility, it would be nice to check with additional clients.

See [how to get started with the API](https://github.com/FreshRSS/FreshRSS/issues/443#issuecomment-36666133)

ghost:
I try to login with News+ and the forked EasyRSS. With both apps, I receive a failed login.

I'm running the dev branch, LAMP, http, checked successfully the API, generated the API key, enabled the API, etc... There are NO logs in data/logs.

Please help!?

ghost:
FYI: I use this info:

http://my.serv.er/p/api/greader.php
jeroen
theAPIkey

Alkarex:
@jeroenpraat If you are running /dev, please have a look at the logs into `FreshRSS/data/users/_/log_api.txt`. Also, do you have any error code shown in News+ (e.g. 401).

ghost:
There is no log_api.txt there. Only log.txt and that had one or two feed errors. The error code in News+ is 'login failure' and the error code in EasyRSS was something like 'wrong username and/or password'.

Alkarex:
Please be sure to set your API password in `/p/i/?c=user&a=profile` for the username you are trying to log into from the mobile app. This is the password to use in the mobile apps.
If this is not enough,  then please test greader.php from your browser. The expected answer is "Bad Request!", which should also be written in the `FreshRSS/data/users/_/log_api.txt`. Please check that you have this underscore directory under users, that your web server is allowed to write there, and check if there is no error in your PHP/Apache log.

ghost:
Oh I'm sorry, looked indeed in the directory of the default user and not with the underscore. There is indeed an api log in there. Where to look for?

ghost:
```
User =>
[...]

PR #847, opened 2015-05-21
Google Reader API: work-around for News+ bug

Alkarex:
https://github.com/noinnion/newsplus/issues/84#issuecomment-57834632
https://github.com/FreshRSS/FreshRSS/issues/443
Copy from @jochenberger https://github.com/noinnion/newsplus/issues/84#issuecomment-57600231

> Imagine there is 1 unread item (let's call it A) in tt-rss and that is synced to News+ correctly, so News+ shows A as unread too.
> Then, I read the item A in tt-rss, so there are no unread items left. Then, I start the sync in News+ again. A is _not_ marked as read in News+.
> If, _however_, a new item B arrives in tt-rss (so there is a new unread item again) and I start the sync in News+, B is fetched and A is marked as unread, leaving only B as unread.

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -467,6 +467,9 @@ function streamContentsItemsIds($streamId, $start_time, $count, $order, $exclude
 	$entryDAO = FreshRSS_Factory::createEntryDao();
 	$ids = $entryDAO->listIdsWhere($type, $id, $state, $order === 'o' ? 'ASC' : 'DESC', $count, '', new FreshRSS_Search(''), $start_time);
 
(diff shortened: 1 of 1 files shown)


Alkarex:
@Alwaysin Here is a patch for a little bug that should have your interest if you are still using News+ for the mobile client.

Alwaysin:
Yes I'm still using it!
Awesome patch which looks really simple, but it's so good to have it! Thanks :D

Alkarex:
Please let me know the results, if you test this patch, @Alwaysin , so that it may be included in the upcoming beta 1.1.1.

Alwaysin:
Just tested, it works fine!

Issue #848, opened 2015-05-22
Number of entries in stream


subzero79:
I've looked around issues but i couldn't find my answer.

I am trying to get a rss stream with a number of entries greater than 20, i tried to add n=50 to the URL but it didn't work.
http://server.com/freshrss/p/i/?a=rss&n=50&get=f_2

Can someone enlighten me how's this achieved? in tt-rss this can be done with the &limit= appended to the URL.

Alkarex:
Hello @subzero79,
We use the `nb` parameter, as such: `http://example.com/freshrss/p/i/?a=rss&nb=50&get=f_2`.
If it is not at the end, it will return n+1 entries.

While looking at your question, I have found a small bug in the `self` attribute of the generated feed, which will soon be corrected in the /dev branch https://github.com/Alkarex/FreshRSS/commit/27d2b88a19345dfc665dc086d3c2b2e4547e1b7f

subzero79:
Thanks very much @Alkarex . It works now. 
BTW great project.

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1251

PR #831, opened 2015-05-15
PubSubHubbub

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/312
http://blog.superfeedr.com/howto-pubsubhubbub/

PR #849, opened 2015-05-23
Minz getBaseUrl correction and RSS template bug

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/848
Corrections in Minz (HTTP_HOST was not sanitized, getURI() was never used and not working anyway with absolute base_url)
$this->url was not defined in rss.phtml , which raised an error and could brake FreshRSS RSS feeds

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -85,44 +85,26 @@ class Minz_Request {
 	}
 
 	/**
-	 * Retourn le nom de domaine du site
-	 */
-	public static function getDomainName() {
-		return $_SERVER['HTTP_HOST'];
-	}
-
-	/**
 	 * Détermine la base de l'url
 	 * @return la base de l'url
 	 */
-	public static function getBaseUrl() {
+	public static function getBaseUrl($baseUrlSuffix = '') {
 		$conf = Minz_Configuration::get('system');
-		$defaultBaseUrl = $conf->base_url;
-		if (!empty($defaultBaseUrl)) {
-			return $defaultBaseUrl;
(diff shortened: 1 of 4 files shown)


lib/Minz/Request.php
```diff
@@ -85,44 +85,26 @@ public static function init() {
 	}
 
 	/**
-	 * Retourn le nom de domaine du site
-	 */
-	public static function getDomainName() {
-		return $_SERVER['HTTP_HOST'];
```
Alkarex, line 91: This should be sanitized, if we keep it. But this function is not used anymore

lib/Minz/Request.php
```diff
@@ -85,44 +85,26 @@ public static function init() {
...
-		} else {
-			return '/';
-		}
-	}
-
-	/**
-	 * Récupère l'URI de la requête
-	 * @return l'URI
-	 */
-	public static function getURI() {
```
Alkarex, line 114: getURI() was not used and not working well with absolute base_url

lib/Minz/Url.php
```diff
@@ -10,7 +10,6 @@ class Minz_Url {
 	 *                    $url['c'] = controller
 	 *                    $url['a'] = action
 	 *                    $url['params'] = tableau des paramètres supplémentaires
-	 *                    $url['protocol'] = protocole à utiliser (http par défaut)
```
Alkarex, line 13: This was not used

Alkarex:
Corrections in the handling of base_url are especially needed for the upcoming PubSubHubbub:
https://github.com/FreshRSS/FreshRSS/pull/831
https://github.com/FreshRSS/FreshRSS/issues/312

Issue #845, opened 2015-05-18
Call for testing /dev before 1.1.1 beta


Alkarex:
Hello everybody,
Without further delay, it is now time to release a new beta of FreshRSS, [version 1.1.1](https://github.com/FreshRSS/FreshRSS/milestones/1.1.1), and [following the tradition](https://github.com/FreshRSS/FreshRSS/issues/725), I would like to invite you all to test the [current /dev branch](https://github.com/FreshRSS/FreshRSS/tree/dev), which will become the next beta.

We need to test in particular:
- [x] Fresh installation with MySQL
- [x] Update from the previous beta, using MySQL
- [x] Fresh installation with SQLite
- [x] Update from the previous beta, using SQLite

So when some of you report whether _it works_ or not, please remember to state in which case you are (fresh or update, MySQL or SQLite).

The list of issues included in this beta is listed in the [milestone 1.1.1](https://github.com/FreshRSS/FreshRSS/issues?q=milestone%3A1.1.1). The [main changes](https://github.com/FreshRSS/FreshRSS/blob/a19e908b73b3fe011bc40609d1edb7f7da6030c8/CHANGELOG) include:
- [New option to detect and mark updated articles as unread](https://github.com/FreshRSS/FreshRSS/issues/798).
  - Be aware of that if you start seeing [articles that are _reappearing_](https://github.com/FreshRSS/FreshRSS/issues/493#issuecomment-102491457) (should be off be default, though).
- [Support for internationalized domain name](https://github.com/FreshRSS/FreshRSS/issues/819) ([IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name)).
- [Improved logic for automatic deletion of old articles](https://github.com/FreshRSS/FreshRSS/issues/493).
- [Attempt to better handle encoded titles](https://github.com/FreshRSS/FreshRSS/issues/754).
- A [new option to turn on/off the logging of SimplePie requests in the syslog](https://github.com/FreshRSS/FreshRSS/issues/711)
- Updated languages (French, English, German), including a new [Czech translation](https://github.com/FreshRSS/FreshRSS/issues/814) by @Tets42.
- Several other smaller improvements

[...]

PR #860, opened 2015-05-30
Minz URL encoding bug

Alkarex:
See also https://github.com/FreshRSS/FreshRSS/pull/849


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/845

Issue #857, opened 2015-05-29
Automatically change strings when changing language


Alwaysin:
As stated here : https://github.com/FreshRSS/FreshRSS/issues/845#issuecomment-104421126

> it could be nice to automatically change the strings to the language selected so that no one will be lost if they do not speak French

This was considering there was a bug which made the default language picked up being French. However if for some reason the auto-detection is not working or the user is using a browser in a different language than theirs, changing the language of the page "on-the-fly" when selected could be great.

Feel free to close if not interesting/not feasible.

aledeg:
As mentioned by @Alkarex, this will add a similar problem as the one encountered in #855. That's the reason I closed #866.

I think there is no real value of working on that. I suggest we close that issue.

Alwaysin:
As you want, it was just a suggestion ^^

Alkarex:
I have just checked: with a Web browser that is set up with English as the preferred language, or with a list of language that does not contain any language known by FreshRSS, then English is automatically used. So it seems to work as expected.

@Alwaysin You probably had French higher than English in your list of languages in the parameters of your Web browser. Feel free to re-open if you still believe there is a bug, but then please provide the HTTP headers sent by your browser, such as `Accept-Language: fr-FR,fr;q=0.8,en-GB;q=0.6,en;q=0.4,da;q=0.2`. You can for instance see them with https://addons.mozilla.org/firefox/addon/live-http-headers/ or other Web programming tools.

Issue #856, opened 2015-05-29
Language selection when installing


Alwaysin:
As stated here : https://github.com/FreshRSS/FreshRSS/issues/845#issuecomment-104945293

> My system is in English, and so is my browser. However, French is pre-selected. Is the detection based on the IP address (which indeed is French)?

During fresh installation.

aledeg:
If I look at the code, the language selection search:
- the selected language
- the server language
- English

I guess your server accept French that's why you have French selected. There is no IP detection to select the language at this point.

I don't know what would be best here.
I think we could remove the server language step and default to English since it is more widespread than other languages.

aledeg:
@Alwaysin could you confirm your server configuration?

Alkarex:
See also my comment https://github.com/FreshRSS/FreshRSS/issues/857#issuecomment-109580995

Alwaysin:
My server uses en_GB.UTF-8 but indeed as pointed by Alkwarex in the other thread, my Accept-Language header states FR and then EN.

Sorry for the false issue!

PR #866, opened 2015-06-02
Add JS to change language automatically

aledeg:
The language is now changed automatically when selected during the install process. It reloads the page automatically to take into account the selected language.
#857

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -72,7 +72,11 @@ function saveLanguage() {
 
 		$_SESSION['language'] = $_POST['language'];
 
-		header('Location: index.php?step=1');
+		if ($_POST['submit']) {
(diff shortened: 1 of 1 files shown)


Alkarex:
I do not think this is good for accessibility (e.g. keyboard navigation)
http://stackoverflow.com/questions/5095336/html-select-onchange-accessibility-concerns
http://webaim.org/techniques/javascript/eventhandlers#onchange
I believe the original issue #857 was "only" about automatically selecting a language based on language preferences in the browser.

aledeg:
I agree for accessibility but the original issue was to be able to change on-the-fly the page language. The automatic selection of the language was in a different issue (see #856)

Issue #868, opened 2015-06-03
MySQL over SSL


nikaro:
Hello,

It would be nice to add the support of MySQL database connection over SSL/TLS. Otherwise, if you use a remote SQL server, the connection is not secure.

Thanks :smiley:

Alkarex:
For reference http://php.net/ref.pdo-mysql#103501

Alkarex:
@NicolasKAROLAK Are you using client keys like in the example below?

``` php
array (
    PDO::MYSQL_ATTR_SSL_KEY     => '/path/to/client-key.pem',
    PDO::MYSQL_ATTR_SSL_CERT    => '/path/to/client-cert.pem',
    PDO::MYSQL_ATTR_SSL_CA      => '/path/to/ca-cert.pem',
)
```

aledeg:
:+1:

Alkarex:
@aledeg Is it also this mechanism with the certificate that you would like?

nikaro:
@Alkarex yes I am.

aledeg:
@Alkarex Actually I don't need it but I think it is a good idea :)

Alkarex:
See my suggestion (not tested at all): https://github.com/FreshRSS/FreshRSS/pull/869
In your `data/config.php`, you should be able to have something like:

``` php
'db' => array(
    'type' => 'mysql',

    ...,

    'pdo_options' => array(
        PDO::MYSQL_ATTR_SSL_KEY     => '/path/to/client-key.pem',
        PDO::MYSQL_ATTR_SSL_CERT    => '/path/to/client-cert.pem',
        PDO::MYSQL_ATTR_SSL_CA      => '/path/to/ca-cert.pem',
    ),
),
```

nikaro:
Just tested it, and... it works ! :smile: 
Thanks.

Alkarex:
The work about loading the existing config continues in https://github.com/FreshRSS/FreshRSS/issues/890 and if there is a need to have a UI, it can be another issue too.

PR #869, opened 2015-06-03
PDO options for e.g. SSL

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/868

Not tested. Not ready for merge yet.


aledeg:
Is there a plan to get the configuration available in the UI?

Alkarex:
@aledeg I do not plan to do it myself, but it could be fine. More urgent: we need to improve the install process to load the existing `data/config.default.php` and `data/config.php` before creating a new `data/config.php`. This could be another issue, though, since it is more general than this bug.

Alkarex:
The work about loading the existing config continues in https://github.com/FreshRSS/FreshRSS/issues/890 and if there is a need to have a UI, it can be another issue too.

Issue #880, opened 2015-06-14
autocomplete off problem


Alkarex:
`<form autocomplete="off">` no longer prevents passwords from being saved
https://developer.mozilla.org/en-US/Firefox/Releases/30/Site_Compatibility#Security

It is a problem for the feed configuration, where username/password may be pre-filled, even though they are not appropriate for a given feed.

The browser might pre-fill the feed's username/password with the credentials saved for another purpose on the domain on which FreshRSS is hosted. This might happen when adding a new feed, or when updating a feed's settings. And this may result in these credentials been sent to a third-party server when requesting a feed.

Alkarex:
Patch https://github.com/FreshRSS/FreshRSS/pull/881 merged

Alkarex:
Follow up in https://github.com/FreshRSS/FreshRSS/pull/1877

PR #881, opened 2015-06-14
Force autocomplete off

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/880
Put a space in the user field instead of empty to avoid autocomplete.
Use feed ID in the username/password field name.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -98,10 +98,10 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 
 			// HTTP information are useful if feed is protected behind a
 			// HTTP authentication
-			$user = Minz_Request::param('http_user');
-			$pass = Minz_Request::param('http_pass');
+			$user = trim(Minz_Request::param('http_user', ''));
+			$pass = Minz_Request::param('http_pass', '');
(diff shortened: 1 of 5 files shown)


Alkarex:
Follow up in https://github.com/FreshRSS/FreshRSS/pull/1877

app/Controllers/feedController.php
```diff
@@ -98,10 +98,10 @@ public function addAction() {
 
 			// HTTP information are useful if feed is protected behind a
 			// HTTP authentication
-			$user = Minz_Request::param('http_user');
-			$pass = Minz_Request::param('http_pass');
+			$user = trim(Minz_Request::param('http_user', ''));
+			$pass = Minz_Request::param('http_pass', '');
 			$http_auth = '';
-			if ($user != '' || $pass != '') {
+			if ($user != '' && $pass != '') {	//TODO: Sanitize
```
Inverle, line 104: What needs to be sanitized?

Alkarex: Hum, I can see that my comment is 10 years old, and I cannot remember for sure, but looking at the code, a username or password with `:` might give problems, for instance

Issue #894, opened 2015-07-10
hex2bin() not available with PHP < 5.4


marienfressinaud:
`hex2bin` function is used twice in `app/Models/EntryDAO.php` but is not available before PHP 5.4. Since we should be compatible at least with PHP 5.3, we have to find a solution.

See http://stackoverflow.com/a/17963343

Log example:

```
PHP Fatal error:  Call to undefined function hex2bin() in /var/www/html/app/Models/EntryDAO.php on line 80, referer: http://localhost/p/i/
```

Alkarex:
`hex2bin()` is used to support SQLite, and we support SQLite 3.7.4+, which implies PHP 5.3.6+ http://php.net/ChangeLog-5.php#5.3.6
So to keep the existing compatibility, we need to support MySQL with PHP 5.2.1+ and SQLite with PHP 5.3.6+

marienfressinaud:
Yes but PHP 5.3.6 < 5.4 and `hex2bin` is only >= 5.4 so it doesn't change anything ;).

Alkarex:
Yes, it was just for the record

Alkarex:
Patch available https://github.com/FreshRSS/FreshRSS/issues/895

Alkarex:
Just for reference, link to another commit for PHP 5.2 compatibility https://github.com/Alkarex/FreshRSS/commit/214a5cc9a4c2b821961bc21f22b4b08e34b5be68

Alkarex:
Done.

PR #895, opened 2015-07-10
Avoid hex2bin for PHP 5.3

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/894
And use native hexadecimal function when available (MySQL) to avoid
having binary data in the SQL logs.

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -6,6 +6,10 @@ class FreshRSS_EntryDAO extends Minz_ModelPdo implements FreshRSS_Searchable {
 		return parent::$sharedDbType !== 'sqlite';
 	}
 
+	public function hasNativeHex() {
+		return parent::$sharedDbType !== 'sqlite';
+	}
+
 	protected function addColumn($name) {
 		Minz_Log::debug('FreshRSS_EntryDAO::autoAddColumn: ' . $name);
 		$hasTransaction = false;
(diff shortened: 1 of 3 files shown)


app/Models/EntryDAO.php
```diff
@@ -77,7 +83,7 @@ public function addEntry($valuesTmp) {
 			substr($valuesTmp['link'], 0, 1023),
 			$valuesTmp['date'],
 			time(),
-			hex2bin($valuesTmp['hash']),	// X'09AF' hexadecimal literals do not work with SQLite/PDO
+			$this->hasNativeHex() ? $valuesTmp['hash'] : pack('H*', $valuesTmp['hash']),	// X'09AF' hexadecimal literals do not work with SQLite/PDO	//hex2bin() is PHP5.4+
```
Alkarex, line 86: I did not like using `hex2bin()` with MySQL anyway, because having binary data in the SQL logs was not nice, sometimes interfering with my SSH client.

constants.php
```diff
@@ -25,4 +25,4 @@
 	define('APP_PATH', FRESHRSS_PATH . '/app');
 	define('EXTENSIONS_PATH', FRESHRSS_PATH . '/extensions');
 
-define('TMP_PATH', sys_get_temp_dir());
+define('TMP_PATH', sys_get_temp_dir());	//Requires PHP 5.2.1+
```
marienfressinaud: I don't think we need this comment...

marienfressinaud:
I think I'm ok with that but don't we need to `unpack` the value somewhere?

constants.php
Alkarex: Removed https://github.com/Alkarex/FreshRSS/commit/d9d9801e8da07b917aa89ce830dad2304f6ab6fc

Alkarex:
@marienfressinaud ~~I am not sure to understand your comment. If you say we should use an intermediate variable, I do not think there is a great advantage in doing so: the return value is used only once, and it is only a simple kind of string formatting / serialization (just like `hex2bin()`). Furthermore, it would be a (tiny bit) more work for PHP with an extra variable.
But I do not really mind.~~

Ahh, now I understand :-) The unpack is done natively in the SQL requests with `hex(hash)`, e.g.
https://github.com/Alkarex/FreshRSS/blob/96b720b03269826ea0a6d58d8aea8ebf7ce7c1d9/app/Models/EntryDAO.php#L605

marienfressinaud:
:+1: perfect!

Issue #862, opened 2015-06-01
Sticky Feeds looks broken with latest pull


Jucgshu:
Sticky Feeds looks broken with last commits. 

I dont' use the other two extensions, so I can't tell whether the problem is the Sticky Feeds extension or the extensions themselves.

Don't hesitate to ask for logs/whatever :-)

Alkarex:
Reference https://github.com/FreshRSS/Extensions/tree/master/xExtension-StickyFeeds

marienfressinaud:
All the extensions which use additional script / style files are broken. Bug is in https://github.com/FreshRSS/FreshRSS/blob/beta/lib/Minz/Extension.php#L171 and it has been highlighted by https://github.com/FreshRSS/FreshRSS/commit/a5a24108 (but the patch was incomplete :p)

I'm writing the complete patch.

Alkarex:
@Jucgshu Could you please test again the current /dev ?

Jucgshu:
Yup, it's working.

Thanx guys !

PR #902, opened 2015-07-13
Fix broken links for extension script/style files

marienfressinaud:
Fix https://github.com/FreshRSS/FreshRSS/issues/862


Alkarex:
Ah, I forgot: consider adding a note in the changelog.

Issue #890, opened 2015-07-05
Install: load existing config.default.php and config.php


Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/869#issuecomment-109581185

Alkarex:
Also needed for
https://github.com/FreshRSS/FreshRSS/issues/680
https://github.com/FreshRSS/FreshRSS/issues/892
https://github.com/FreshRSS/FreshRSS/issues/865
https://github.com/FreshRSS/FreshRSS/issues/897

marienfressinaud:
I don't really understand why we need to load existing config: install.php is used only the first time we install FRSS, ~~so when there is no user yet~~ (sorry, we talk about general configuration :)).

**Edit**: well, I see benefit for SSL option for DB but not for the other tickets...

**Edit 2**: in fact, no, I don't see any interest. Install process is aimed to create `config.php`... and we don't need to load default values since we only want to overwrite them.

Alkarex:
I see, we need to clarify the desired behaviour:
1) Even during the first install, I think it is cleaner to load `config.default.php` instead of having the values hard-coded in `install.php` https://github.com/FreshRSS/FreshRSS/blob/4032bf5d8501768647eac67b4692662defd50e46/app/install.php#L171
2) I thought `install.php` was supposed to be called at every update, except in the case of automatic update;
3) At least in the past, there has been some update recommendations in the lines of a) keep the `data`directory, b) deploy the new code (which also contains some files in `data`. This is also what I [suggest for git update](https://github.com/FreshRSS/FreshRSS/blob/c31a6bf5ef7babe1fd5440e406e5bda372aa0f08/README.md#example-of-full-installation-on-linux-debianubuntu), with a `git reset --hard && git pull` which will also re-create the `data/install.txt`;
4) Since there is therefore ( a risk | an expectation ) that `install.php` is called after an update, if we do not want `install.php` to re-create `config.php`, it should maybe jump over some steps when `config.php` already exists;

Do we want `install.php` to be called after an update not done by an automatic update?

marienfressinaud:
1. Ok but the line you commented is not the most appropriate because we can just remove these two hard-coded variables (environment and simplepie_syslog_enabled). But it makes sense in the view, true.
2. Not since I introduced the automatic update. The goal was to simplify this script which was very hard to maintain and not the best place to handle updating process.
[...]

PR #911, opened 2015-07-20
Load existing default configuration

marienfressinaud:
See https://github.com/FreshRSS/FreshRSS/issues/890


Alkarex:
The code looks good :-)
I will test with MySQL + SQLite during the testing of /dev and before the next beta.

Issue #906, opened 2015-07-17
dirname() has an unexpected behaviour in Minz_Request::getBaseUrl()


marienfressinaud:
In `Minz_Request::getBaseUrl()`, `dirname()` function is called to remove the last part of `$_SERVER['REQUEST_URI']` (see https://github.com/FreshRSS/FreshRSS/blob/beta/lib/Minz/Request.php#L103).

But we can face two cases:
- /p/i/ (i.e. most of the time) which is transformed into /p/
- /p/i/index.php (e.g. during installation) which is transformed into /p/i/

Expected result: always have the same result (i.e. /p/i/) for both cases.

Note: if we keep /p/i/, we should change `Minz_Url::display()` which calls `Minz_Request::getBaseUrl(PUBLIC_TO_INDEX_PATH)`.

Needed for #865

Alkarex:
Ok, I will have a look.

Alkarex:
Not tested yet, but should already be more consistent with https://github.com/Alkarex/FreshRSS/commit/2d22bf300a8dabcc77237f4bb56bf6532486fca2
It does not return the trailing slash, so it would be `/p/i` in both cases, so we can keep the convention we have in `constants.php` in which path constants start with a slash.
The value parameter `PUBLIC_TO_INDEX_PATH` is only used when the ~~using~~ retrieving the base URL from the `config.php`.

marienfressinaud:
>  The value parameter PUBLIC_TO_INDEX_PATH is only used when the using retrieving the base URL from the config.php.

Do you mean `getBaseUrl()` will return:
- `http://example.com/p/i` if `base_url` is not present in `config.php`
- `http://example.com/p` if it is (and so we should call `getBaseUrl(PUBLIC_TO_INDEX_PATH)` to have the previous behaviour)

I think it's a bit confusing. It would be more consistent to always return `http://example.com/p` in both cases (I didn't think about it when writing my first message) and to always add the `$baseUrlSuffix` if given.

Alkarex:
If `base_url` is present in config.php, then this is the value that is returned, together with the suffix `PUBLIC_TO_INDEX_PATH`.
If `base_url` is empty in config.php, then the base url is calculated (and the value `PUBLIC_TO_INDEX_PATH` is not used because we directly get the `/p/i` path)

In both cases, we should have identical results when calling FreshRSS from a Web browser.

marienfressinaud:
[...]

PR #915, opened 2015-07-23
Fix getBaseUrl behaviour

marienfressinaud:
Fix https://github.com/FreshRSS/FreshRSS/issues/906
Use and merge https://github.com/FreshRSS/FreshRSS/pull/910

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -85,29 +85,37 @@ class Minz_Request {
 	}
 
 	/**
-	 * Détermine la base de l'url
-	 * @return la base de l'url
+	 * Try to guess the base URL from $_SERVER information
+	 *
+	 * @return the base url (e.g. http://example.com/)
 	 */
-	public static function getBaseUrl($baseUrlSuffix = '') {
-		$conf = Minz_Configuration::get('system');
-		$url = $conf->base_url;
-		if ($url == '' || !preg_match('%^https?://%i', $url)) {
-			$url = 'http';
-			$host = empty($_SERVER['HTTP_HOST']) ? $_SERVER['SERVER_NAME'] : $_SERVER['HTTP_HOST'];
-			$port = empty($_SERVER['SERVER_PORT']) ? 80 : $_SERVER['SERVER_PORT'];
-			if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on') {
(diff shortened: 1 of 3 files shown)


lib/Minz/Request.php
```diff
@@ -85,26 +85,37 @@ public static function init() {
...
-			$url = rtrim($url, '/\\') . $baseUrlSuffix;
+			$url .= '://' . $host . ($port == 80 ? '' : ':' . $port);
 		}
-		return filter_var($url . '/', FILTER_SANITIZE_URL);
+		if (isset($_SERVER['REQUEST_URI'])) {
+			$path = $_SERVER['REQUEST_URI'];
+			$url .= substr($path, -1) === '/' ? substr($path, 0, -1) : dirname($path);
+		}
+
+		return $url;
```
Alkarex: Is it on purpose that it is not `return filter_var($url, FILTER_SANITIZE_URL);` there?
Probably the only data input that is not already sanitized is `$_SERVER['HTTP_HOST']` and `$_SERVER['REQUEST_URI']`

lib/Minz/Request.php
```diff
@@ -85,26 +85,37 @@ public static function init() {
...
+	/**
+	 * Return the base_url from configuration and add a suffix if given.
+	 *
+	 * @param $base_url_suffix a string to add at base_url (default: empty string)
+	 * @return the base_url with a suffix.
+	 */
+	public static function getBaseUrl($base_url_suffix = '') {
+		$conf = Minz_Configuration::get('system');
+		$url = rtrim($conf->base_url, '/\\') . $base_url_suffix;
+		return filter_var($url, FILTER_SANITIZE_URL);
```
Alkarex, line 118: Since the data comes directly from the config file, maybe `filter_var` is superfluous there? But it is indeed safer to keep it.

lib/Minz/Request.php
marienfressinaud: Not on purpose, no, I'll fix it.Thanks!

lib/Minz/Request.php
marienfressinaud, line 118: There is also `$base_url_suffix` so I prefer to keep this function: I don't know how this method will be called...

Alkarex, line 118: Ok :-)

lib/Minz/Request.php
marienfressinaud: Done!

Issue #865, opened 2015-06-01
PubSubHubbub: base_url


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/312#issuecomment-102414760
- [x] Generate `base_url` during install
- [x] Check if `base_url` is public? (optional)
- [ ] Documentation

Alkarex:
Needs https://github.com/FreshRSS/FreshRSS/issues/890

Alkarex:
If `base_url` is not public, then PubSubHubbub should not try to subscribe.
Probably, if `base_url` is not public, we should just leave it empty (or add another option/flag).

marienfressinaud:
I would prefer to add a flag since a user can change the config himself

marienfressinaud:
I wrote a very basic test at https://github.com/marienfressinaud/FreshRSS/commit/339e32424fa60fc0c99a4c10890abef139444f6d but I'm not sure where I should use `enable_pubsubhubbub` (`Feed->pubSubHubbubEnabled`? `Feed->pubSubHubbubSubscribe`? `feedController`?)

Alkarex:
Two possibilities:
- https://github.com/marienfressinaud/FreshRSS/blob/339e32424fa60fc0c99a4c10890abef139444f6d/app/Controllers/feedController.php#L445
- https://github.com/marienfressinaud/FreshRSS/blob/339e32424fa60fc0c99a4c10890abef139444f6d/app/Models/Feed.php#L391

In both cases, just before the `if` or as the first statement of the `if`.
It depends if you think it is better to have this test in a controller or in a model.
Probably doing the test in the controller (first option) is cleaner, but I do not have strong feelings either way.

P.S.: Indeed, now that we have another option to turn PubSubHubbub on/off, it could be good to be clear about the role of `pubSubHubbubEnabled()`: it checks that the system is in place, has worked, and has not expired. Maybe this function could be renamed to e.g. `pubSubHubbubWorking()` or something.

Alkarex:
If done in the controller, for performance, it would be good to add a check on https://github.com/marienfressinaud/FreshRSS/blob/339e32424fa60fc0c99a4c10890abef139444f6d/app/Controllers/feedController.php#L305 : get the configuration flag before the loop, and do not call `pubSubHubbubEnabled()´ if PubSubHubbub is disabled in the configuration

marienfressinaud:
[...]

PR #917, opened 2015-07-23
Generate base_url during installation

marienfressinaud:
Fix https://github.com/FreshRSS/FreshRSS/issues/865
- `base_url` is stored without `/i` (e.g. http://localhost/FreshRSS/p)
- Add a `pubsubhubbub_enabled` configuration key
- Add a very basic `server_is_public()` function to test if the server is reachable by an external service. Note it tests only if host is a localhost address,it can be improved if you have ideas.
- `pubsubhubbub_enabled` is set to `false` if the server is not public
- PubSubHubbub process is not handled if `pubsubhubbub_enabled` is `false`


Alkarex:
Looks good :-)

Issue #923, opened 2015-07-25
Minz: Bug loading config.default for array values


Alkarex:
If arrays such as `limits` exist in `config.php`, their values always come from `config.php` even when they do not exist, which create an _undefined index_ exception.

For instance, I do not have the new `max_registrations` value (https://github.com/FreshRSS/FreshRSS/issues/905) in my `config.php` (but it is in my `config.default.php`), which generates a 
- Notice: Undefined index: max_registrations in ./FreshRSS/app/views/user/manage.phtml on line 12
- Notice: Undefined index: max_registrations in ./FreshRSS/lib/lib_rss.php on line 305

Reference https://github.com/FreshRSS/FreshRSS/blob/dev/lib/Minz/Configuration.php
https://github.com/FreshRSS/FreshRSS/issues/919

Alkarex:
I suggest loading `config.default.php` and replacing its values with `config.php` by calling http://php.net/array-replace-recursive

marienfressinaud:
Yes sure! I don't even know why I didn't do this in first place ><...

marienfressinaud:
Fixed by https://github.com/FreshRSS/FreshRSS/pull/926

PR #926, opened 2015-07-27
Load configuration by recursion

marienfressinaud:
- Remove Minz_Configuration::$data_default
- Default values are loaded first in $data
- $data values are replaced by values from config file

Fix https://github.com/FreshRSS/FreshRSS/issues/923


Alkarex:
Looks good :+1:

Issue #930, opened 2015-07-29
Czech and German translations for 1.1.2-beta


marienfressinaud:
Hi @Tets42, @thomasE1993 and @ealdraed,

we still have some strings untranslated in FreshRSS and we plan a new release tomorrow and it could be nice to have a fully-translated version. You can give the translations in comments.

Here is what we need.

**Czech**:

```
'There is %d account created yet',
'There are %d accounts created yet',
'Allow account creation',
'0 means that there is no account limit',
'Max number of accounts',
'Account deletion',
'Your account and all the related data will be deleted.',
'The maximum amount of accounts has been updated.',
'New account',
'Create an account?',
'Account creation',
'Keep previous installation',
'Reinstall FreshRSS',
'We have detected that FreshRSS is already installed!',
'You will lose your previous configuration by reinstalling FreshRSS. Are you sure you want to continue?',
```

**German**:

```
'There is %d account created yet',
'There are %d accounts created yet',
'Allow account creation',
'0 means that there is no account limit',
'Max number of accounts',
'Account deletion',
'Your account and all the related data will be deleted.',
'The maximum amount of accounts has been updated.',
'New account',
'Create an account?',
'Account creation',
```

Thanks for your help!

By the way, I plan to open this kind of ticket each time we plan to release a new version. Is it fine for you or you would prefer another process? Note we may have to change in the future if we have more languages to support.

ghost:
Hello together

German:

'There is %d account created yet',
Es wurde bis jetzt noch keinen Account erstellt.

 'There are %d accounts created yet', 
Es wurden bis jetzt x Accounts erstellt

'Allow account creation', 
Erlaube die Accounterstellung

'0 means that there is no account limit',
0 meint, dass es kein Account Limit gibt

 'Max number of accounts', 
Maximale Anzahl von Accounts

'Account deletion', 
Accountlöschung

'Your account and all the related data will be deleted.', 
Dein Account und alle damit bezogenen Daten werden gelöscht.

'The maximum amount of accounts has been updated.', 
Die maximale Anzahl von Accounts wurde aktualisiert.

'New account', 
Neuer Account

'Create an account?', 
Erstelle einen Account?

'Account creation',
Accounterstellung

Instead of Account --> "Konto" is also possible

What means the %d? Maybe the sentence has to be corrected. 

Please check for errors, I am on holiday and on my smartphone :-)

marienfressinaud:
Thank you!

[...]

PR #933, opened 2015-07-29
Add translations for German and Czech

marienfressinaud:
Fix https://github.com/FreshRSS/FreshRSS/issues/930


marienfressinaud:
I merged the PR myself since it's i18n only

PR #940, opened 2015-08-04
PubSubHubbub prevent subscribing too often in case of error

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/939
To be tested a bit more.


Alkarex:
Is anybody about to try? Otherwise I suggest merging to test more in /dev

Issue #892, opened 2015-07-05
Admin UI to control the limits


Alkarex:
Add a UI to control the maximum number of feeds and categories per user https://github.com/FreshRSS/FreshRSS/issues/680

Alkarex:
Needs https://github.com/FreshRSS/FreshRSS/issues/890

Alkarex:
Implemented by @aledeg

Issue #858, opened 2015-05-29
Remove "title" field when installing


Alwaysin:
As stated here : https://github.com/FreshRSS/FreshRSS/issues/845#issuecomment-104421126

> Not sure about "Installation · FreshRSS" field. I think it is the name which will then be used in the title but it could be more descriptive.

I think this field is confusing and has no real usefulness.

aledeg:
The name is confusing but there is a real use of it.
When you want to have a specific name for your installation (like for Framasoft), it is useful to name it the way you want.
I agree we need a better name since I had to make the full install to understand what was the use of it.

Maybe we could rename it to **Instance name** or something similar.

Alkarex:
I think we could keep the functionality, but remove it from the install process since it is not an essential feature, just like many other settings we have.

aledeg:
I think the more we can add in the installation process, the better.
We can argue why, but it is only what I feel. I hate when a software has a lot of steps after installation to finish configuration. I like the way of Symfony that has a hook on composer that allows to configure the application at the end of installation. There is default values if you don't want to configure after installation.

We could have either default values for all non mandatory fields or have 2 path of installation (regular and custom).

aledeg:
Any insight for that?
I just checked the interface and I can't find where to modify this value. Is it available somewhere I am not aware of?

Alkarex:
It is the first field of step 2. But the translation of the label `install.title` is misleading indeed:
![clipboard01](https://cloud.githubusercontent.com/assets/1008324/8023426/6444b826-0d0c-11e5-963d-8306d67ae70b.png)

aledeg:
I got that one. I was wandering after install, is there a way to change it from the interface. So far I found only the value in the configuration file.

Alkarex:
Indeed, after install, it can only be changed from the config file.

aledeg:
So I guess we should keep it during install.

marienfressinaud:
I feel like @Alkarex that we don't need this information during installation (such as the "remove articles after…"). But it still can be useful to administrators who want to have specific configuration so I suggest to ask the user at the end if he wants to continue to configure additional options.

aledeg:
@marienfressinaud Or we can add it in the new "system configuration" page.
Let me know what you think, so I can include that in #958

marienfressinaud:
Let's go! I'm not sure if it's useful but since several persons asked for it, I think we can include it.

aledeg:
[...]

PR #958, opened 2015-08-22
Add a system configuration page

aledeg:
It allows to modify system configuration from the interface. At the moment, only limits are modifiable. The user limit was removed from the user page and added here along with categories and feeds limits.

See #892 and #680


Alkarex:
I have not tested yet, but looks good :-) Do you want to put anything more in this pull request?

aledeg:
Thanks!
I cannot think of anything else. If you have to test it, maybe you can find something I should add.

marienfressinaud:
I'm not sure I like the idea of moving user limit: it makes more sense to keep it with user management for me…

aledeg:
For me, user management and system configuration is 2 different things. It feels more natural to have the two separated.
I know that the user limit is linked to the users. But I think it safe to say it's a system configuration parameter.

app/views/configure/system.phtml
```diff
@@ -0,0 +1,47 @@
...
+		</div>
+
+		<div class="form-group form-actions">
+			<div class="group-controls">
+				<button type="submit" class="btn btn-important"><?php echo _t('gen.action.submit'); ?></button>
+				<button type="reset" class="btn"><?php echo _t('gen.action.cancel'); ?></button>
+			</div>
+		</div>
+	</form>
+</div>
```
marienfressinaud: Miss a blank line here

app/i18n/cz/admin.php
```diff
@@ -146,6 +146,15 @@
 		'title' => 'Statistika',
 		'top_feed' => 'Top ten kanálů',
 	),
+	'system' => array(
+		'_' => 'System configuration',
```
marienfressinaud: Could you add a "TODO: translate" each time you don't know the translation?

marienfressinaud:
Ok, it could make sense too. Let me test (today) and I'll tell you ;). Also, I don't forget your email on the mailing

aledeg:
I was thinking it would be a good idea to include on that page the "authentication method" page inside the system configuration page.
It's also a system configuration and having it on the same page than the one I created in this PR would make sense. Later in the process, if we add a lot of system configuration parameters, it would be nice to have a section dedicated for system configuration as we already have the user configuration.

marienfressinaud:
Don't spend too much time on layout organization: we have to do it but I think it should be discussed on a dedicated ticket, with mockups and design proposals otherwise we will have to change again in 6 months ;).

aledeg:
I wasn't going to change more. That was just a thought.

aledeg:
I fixed the minor issues concerning the code formatting.
I added a configuration item for the instance name.

See #858.

Issue #980, opened 2015-09-11
[bug] removing a feed or a category deletes all user queries


aledeg:
No matter what are the user queries, they are dropped when removing a feed or a category. Even when they are not related to the feed or the category deleted.

The normal behavior should be to drop only the user queries related to that feed or category.

marienfressinaud:
It could explain why I lost all my queries, I thought it was a bad manipulation from me :(

PR #982, opened 2015-09-11
Fix feed and category side effect

aledeg:
Before, when deleting a feed or a category, the user queries were deleted as well. No matter if they were related or not.
Now, they are deleted only if they are related.

I this this fix is not the best way to handle that. I think it would be better if we could find a way to create a UserQuery object from the array.
The same applies when displaying the user queries in the interface.

See #980

### DIFF
--- a/app/Models/ConfigurationSetter.php
+++ b/app/Models/ConfigurationSetter.php
@@ -119,6 +119,8 @@ class FreshRSS_ConfigurationSetter {
 		foreach ($values as $value) {
 			if ($value instanceof FreshRSS_UserQuery) {
 				$data['queries'][] = $value->toArray();
+			} elseif (is_array($value)) {
+				$data['queries'][] = $value;
 			}
 		}
 	}


Alkarex:
Thanks!
Where was it called from with an array?
Do you know since when the bug was present in FreshRSS?

marienfressinaud:
It is called when deleting categories or feeds, when we use `remove_query_by_get` (https://github.com/FreshRSS/FreshRSS/blob/beta/lib/lib_rss.php#L471). I guess the problem had been introduced with the `ConfigurationSetter` :( My bad...

aledeg:
That's why we need tests :)
I know that's not really exiting when the code is done. But we should try to write tests first, it will become more fun. At least, that's what I try to do in my day job. Of course it is not always possible.

marienfressinaud:
:+1: I will open a new ticket quite soon to discuss about future of FreshRSS and TDD (test-driven development) is a part of my idea ;).

PR #1000, opened 2015-10-17
Movim

aledeg:
Add a share option for Movim


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/992

Issue #988, opened 2015-09-18
Logs


Liandriz:
Is is possible to auto cleanup pubsub logs, in "/data/users/_/log_pshb.txt" when cleaning profile logs?
Just see that this file was over 1G, rhaaaaa.

And for user log, any way to contain it under a sizelimit? Got blank page when visiting "i/?a=logs", had to manually delete my log file.

Alkarex:
@Liandriz Please test this patch if you can https://github.com/FreshRSS/FreshRSS/issues/1010

Liandriz:
Both log_api.txt and log_pshb.txt cleared when personal admin logs cleared.
Tried an update to beta dev from this one, still got my theme wiped :0

Alkarex:
Thanks for testing. The theme bug is another issue https://github.com/FreshRSS/FreshRSS/issues/935

PR #1009, opened 2015-10-24
Clean logs

Alkarex:
Reduced login of API and PubSubHubbub (both are quite stable now).
When clearing logs as admin, also clear API and PubSubHubbub logs.
https://github.com/FreshRSS/FreshRSS/issues/988


Alkarex:
Re-done in https://github.com/FreshRSS/FreshRSS/pull/1010

Issue #938, opened 2015-08-01
SEO (robots, meta)


Wanabo:
I think FreshRSS is build for personal use (Robots: noindex,nofollow), but I would like to use FreshRSS for the public.
To become a successful website with the use of FreshRSS SEO is a must!
Not only Robots: noindex,nofollow has to be changed (easy enough to do myself), but at least meta description and meta keywords are needed. Best dynamically generated from articles, categories, tags, titles etc.

I'm setting up a new website and love to use FreshRSS, but SEO is missing.
Will SEO ever be implemented?

Regards

Alkarex:
Hello @Wanabo 
We do not currently have a plan to add SEO, put it may be doable depending on the amount of work.
Adding a control for the `robots` directive would not be a problem:

``` html
<meta name="robots" content="xxx" />
```

and adding a meta description for the feeds would also be easy:

``` html
<meta name="description" content="Description given by the feed" />
```

Would that be useful for you?
For other, more advanced, behaviours, I would rather see that in an extension.
BTW, I do not think any major search engine uses the meta keyword for anything.

Wanabo:
[quote]BTW, I do not think any major search engine uses the meta keyword for anything.[/quote]
Personally I am not convinced that is the case. When there are no description or keywords set google is quite capable to generate their own description. Setting them yourself will certainly help to be found in google for specific articles, tags and the like.

I've changed in layout.phtml

meta name="robots" content="noindex,nofollow"

to

meta name="robots" content="index,follow"

All though in my opinion    meta name="robots" content="index,follow" can be removed all together. 

Do you mean for meta name="description" content="Description given by the feed" to have this in the xml? Now it is description>RSS feed of Your RSS feeds | FreshRSS</description
That would be nice!

I have a website with Rnews 1.01 (http://nieuwskop.nl and modded some own seo), but for my new website, http://nieuwskop.be I would like to try an other self hosted RSS software and see if that is more successful. I have setup two possible candidates. FreshRSS: http://www.nieuwskop.be/fr/p/i/ and Selfoss: http://www.nieuwskop.be/ 

With SEO would have my preference.

Regards.

Alwaysin:
> BTW, I do not think any major search engine uses the meta keyword for anything.
> 
> > Personally I am not convinced that is the case. 

Alkarex is right, Google does not use it : https://www.youtube.com/watch?v=jK7IPbnmvVU and from latest news I heard, Bing even considers only spammers use this metatag.

[...]

PR #1011, opened 2015-10-24
Config allow robots

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/938

If `allow_robots` is true, then an HTML meta description is generated (there is only some content when one single feed is shown):

``` html
<meta name="description" content="Le Monde.fr | Toute l'actualité au moment de la connexion" />
```

Otherwise, we still generate the robots exclusion HTML meta tag:

``` html
```

### DIFF
--- a/app/Models/Context.php
+++ b/app/Models/Context.php
@@ -10,6 +10,7 @@ class FreshRSS_Context {
 	public static $categories = array();
 
 	public static $name = '';
+	public static $description = '';
 
 	public static $total_unread = 0;
 	public static $total_starred = array(
@@ -94,6 +95,13 @@ class FreshRSS_Context {
 	}
 
 	/**
+	 * Return true iif the current requests target a feed and not a category or all articles.
+	 */
+	public static function isFeed() {
+		return self::$current_get['feed'] != false;
+	}
+
+	/**
 	 * Return true if $get parameter correspond to the $current_get attribute.
 	 */
 	public static function isCurrentGet($get) {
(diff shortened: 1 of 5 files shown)


app/Models/Context.php
```diff
@@ -94,6 +95,13 @@ public static function currentGet($array = false) {
 	}
 
 	/**
+	 * Return true iif the current requests target a feed and not a category or all articles.
```
marienfressinaud: "if" ;)

Alkarex: It was intentional :-)
It is the abbreviation of "if and only if". Otherwise, one must write "or false otherwise" or something similar.
https://en.wikipedia.org/wiki/Iff

marienfressinaud: So I would prefer "if and only if", I'm not a fan of abbreviations in comments ;)

app/layout/layout.phtml
```diff
@@ -36,7 +36,11 @@
 		<meta name="apple-mobile-web-app-status-bar-style" content="black" />
 		<meta name="apple-mobile-web-app-title" content="<?php echo FreshRSS_Context::$system_conf->title; ?>">
 		<meta name="msapplication-TileColor" content="#FFF" />
+<?php if (FreshRSS_Context::$system_conf->allow_robots) { ?>
+		<meta name="description" content="<?php echo htmlspecialchars(FreshRSS_Context::$name . ' | ' . FreshRSS_Context::$description, ENT_NOQUOTES, 'UTF-8'); ?>" />
```
Alkarex, line 40: Arf, it should be ENT_COMPAT there

Alkarex, line 40: https://github.com/FreshRSS/FreshRSS/pull/1022

PR #1012, opened 2015-10-24
Remove "title" field when installing

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/858
Now we have a configuration UI instead.


marienfressinaud:
:+1:

Issue #975, opened 2015-09-09
base_url is not correct when FreshRSS is behind SSL backed reverse proxy


zirkome:
Hi,

I'm actually using FreshRSS on Apache2.4 and put it behind an NGINX reverse proxy. I set my reverse proxy to use SSL and redirect traffic to my Apache. The problem is that the traffic between Nginx and Apache is HTTP (port 80) so when `config.php` is created `base_url` is totally wrong: `http` + port 80 at the end. When redirecting traffic like that there is a variable set in `$_SERVER` which take in account the front protocol: `HTTP_X_FORWARDED_PROTO`.

I think adding a check for that variable must be included in `install.php` to use the right port and right protocol: the real ones :)

Alkarex:
Hello @kokaz,
Could you please give a try to this patch? https://github.com/FreshRSS/FreshRSS/pull/1013

PR #1013, opened 2015-10-24
Use HTTP_X_FORWARDED_

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/975

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -91,9 +91,30 @@ class Minz_Request {
 	 */
 	public static function guessBaseUrl() {
 		$url = 'http';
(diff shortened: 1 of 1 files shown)


Alkarex:
Not yet tested.

Alkarex:
The test to be done is to check if `base_url` is generated correctly during install, without proxy. Then possibly with a proxy.

marienfressinaud:
Seems good to me be I would like to validate it's working first. However, I don't have the corresponding configuration to test it :(

marienfressinaud:
Merged into https://github.com/FreshRSS/FreshRSS/commit/529a28567f7166b7e4ed1ff62388b86765f71226 (I made a little fix before merging, that's the reason why GitHub think this PR is not merged)

Issue #965, opened 2015-08-30
Some kind of alert when a feed has a problem


Alwaysin:
Sometimes we can not be aware that a feed has a problem, and it can be problematic if it lasts too much. Could it be possible for example to have a little exclamation point next to the name of the category containing the feed, or any other way that can make it seen from the main view?

aledeg:
I checked in the interface and I am not sure what kind of alert you want since there is already something.

Here is what I see in the categories. Red feeds are feeds with errors.
![feed error](https://cloud.githubusercontent.com/assets/3056148/9569551/c1890bdc-4f3c-11e5-8076-a7e183696806.png)

Could you elaborate a little more?

Alwaysin:
I never unroll the categories, so I never see the red feeds. That's what I meant by being "seen from the main view".

aledeg:
We could add the same rule on the feed name in the main view.

Alwaysin:
On the category name you mean?

I was thinking of just a little exclamation point next to the category name to make it subtle enough.

aledeg:
@Alwaysin I made a PR. You can try it even if it is not merged yet. Please comment.

Alwaysin:
Thanks!

I think it is not enough visible, it should catch the eye (not too much, but a bit). I've try to add a color to your CSS rule (color: #880000;) and I find it better (and a space after the warning sign also!)
EDIT : I think we can re-use the color used for the feed name, as displayed on your screenshot. But just for the warning sign, not the whole category name of course

What do you think?

PR #1014, opened 2015-10-25
More viewable visual alert for categories with problematic feeds

Alwaysin:


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/965

Issue #1015, opened 2015-10-25
Log unsuccessful login attemps


Alwaysin:
I wanted to do a fail2ban rule modeled on this one : http://www.sublimigeek.fr/fail2ban-securiser-instance-wallabag
But unsuccessful login attempts seem not to be registered on Apache logs, I only have : 

```
freshrss.example.org IP.xxx.xxx.xxx - - [25/Oct/2015:11:16:02 +0100] "GET /i/?c=javascript&a=nonce&user=hacker HTTP/1.1" 200 688 "http://freshrss.example.org/i/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.71 Safari/537.36"
```

Is there something to change on my Apache conf, or is it something in FreshRSS?

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/1001

Alkarex:
Please test https://github.com/FreshRSS/FreshRSS/pull/1016
You will now see an HTTP status 403 in the Web server logs.

Alwaysin:
It works, thanks.

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2903

PR #1016, opened 2015-10-25
HTTP 403 for invalid login

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1015
And does not leak if user exists or not.
Must not use redirect to send status 403.

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -124,6 +124,7 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 			$conf = get_user_configuration($username);
 			if (is_null($conf)) {
 				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
+				return;
 			}
 
 			$ok = FreshRSS_FormAuth::checkCredentials(
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -123,8 +123,7 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 
 			$conf = get_user_configuration($username);
 			if (is_null($conf)) {
-				Minz_Request::bad(_t('feedback.auth.login.invalid'),
-				                  array('c' => 'auth', 'a' => 'login'));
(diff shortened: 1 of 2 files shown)


app/Controllers/authController.php
```diff
@@ -123,8 +123,7 @@ public function formLoginAction() {
 
 			$conf = get_user_configuration($username);
 			if (is_null($conf)) {
-				Minz_Request::bad(_t('feedback.auth.login.invalid'),
-				                  array('c' => 'auth', 'a' => 'login'));
+				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
```
marienfressinaud, line 126: If there is no redirection (`$redirect = false`), rest of the code in the method is executed as well. Are you sure it does not cause more troubles?

Alkarex, line 126: Ok, we need to return/exit after this line, then.

Alkarex, line 126: This one was problematic indeed. I have added a `return`.
The other two below should be fine.

app/Controllers/javascriptController.php
```diff
@@ -43,7 +43,11 @@ public function nonceAction() {
 		} else {
 			Minz_Log::notice('Nonce failure due to invalid username!');
 		}
-		$this->view->nonce = '';	//Failure
-		$this->view->salt1 = '';
+		//Failure: Return random data.
+		$this->view->salt1 = sprintf('$2a$%02d$', FreshRSS_user_Controller::BCRYPT_COST);
+		for ($i = 22; $i > 0; $i--) {
+			$this->view->salt1 .= './ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'[rand(0, 63)];
```
marienfressinaud: Just realized: is this syntax valid with PHP 5.3?

Alkarex: Well spotted. It is indeed first PHP 5.5 that added support for accessing characters within string literals using []...
http://php.net/manual/en/language.types.string.php

Issue #1025, opened 2015-11-02
Installation does not work anymore


marienfressinaud:
Due to https://github.com/FreshRSS/FreshRSS/pull/1012, we can't finish installation process since `title` is required.

Alkarex:
Hum, it was supposed to be set by https://github.com/marienfressinaud/FreshRSS/blob/fix/remove-requirement-on-title/app/install.php#L102

marienfressinaud:
This line is executed only during a reinstallation if we keep the previous configuration

Alkarex:
Ok

PR #1026, opened 2015-11-02
Remove title from requirements on title page

marienfressinaud:
Fix https://github.com/FreshRSS/FreshRSS/issues/1025

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -122,6 +122,8 @@ function saveStep1() {
 function saveStep2() {
 	$user_default_config = Minz_Configuration::get('default_user');
 	if (!empty($_POST)) {
+		$system_default_config = Minz_Configuration::get('default_system');
+		$_SESSION['title'] = $system_default_config->title;
 		$_SESSION['old_entries'] = param('old_entries', $user_default_config->old_entries);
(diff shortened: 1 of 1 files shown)


marienfressinaud:
Ups, it does not work yet, please don't merge ;)

EDIT: ok :)

app/install.php
```diff
@@ -122,6 +122,7 @@ function saveStep1() {
 function saveStep2() {
 	$user_default_config = Minz_Configuration::get('default_user');
 	if (!empty($_POST)) {
+		$_SESSION['title'] = 'FreshRSS';
```
Alkarex: It would be nicer to get the value from https://github.com/marienfressinaud/FreshRSS/blob/fix/remove-requirement-on-title/data/config.default.php#L25 with something like the other section (`$system_conf->title`)

marienfressinaud:
Done! :)

Alkarex:
:+1:

marienfressinaud:
@Alkarex > note about the "Update action" label: it should be used when we need to add a specific patch during auto-update process

Alkarex:
All right. I was using it also for install- and update-related code.

PR #1027, opened 2015-11-04
Use font with international glyphs

asmfreak:
I saw that OpenSans font used for default UI theme (`Origine`) doesn't use international font glyphs (cyrillic and greek).
I proposed to use fonts from Google.

### DIFF
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -1,10 +1,8 @@
 @charset "UTF-8";
 
 /*=== FONTS */
-@font-face {
-	font-family: "OpenSans";
-	src: url("../fonts/openSans.woff") format("woff");
-}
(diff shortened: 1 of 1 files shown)


Alwaysin:
I think your PR won't be accepted as you are making calls to Google servers. Many are using self-hosted software in order not to be dependant on any company.

marienfressinaud:
Exactly :) we prefer to include fonts directly in FreshRSS. If the PR is about OpenSans is not up to date, could you update font in `./p/themes/fonts`.

Also, please note you should work in the `dev` branch and make your PR on it (instead of `beta`).

asmfreak:
Ok. I'm fixing the PR. [See here](https://github.com/FreshRSS/FreshRSS/pull/1028)

PR #1035, opened 2015-11-23
Update greader.php

Purexo:
streamContents can know exclude target unread item


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/704#issuecomment-158750428

Issue #634, opened 2014-09-24
[feature request] add a context object


aledeg:
See #632 for the context of the discussion.

It would be nice to have a context object that handle the current state of the application and the current configuration for the user.
This way we can have a more efficient way to modify the layout and query the database.

At the moment, here is the list of the thing I can think of:
- compute the current state to display depending of the user configuration and selection
- simplify templates by using context methods instead of complex tests
- simplify queries by using context methods instead of complex tests

PR #1050, opened 2016-01-09
Update jQuery, and UI bugs (scroll, global and reader views)

Alkarex:
http://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/


Alkarex:
Hum, there seems to be a bug with jQuery 2.2 for auto-loading articles when scrolling down.

Alkarex:
Now it should work. There were several small bugs with loading on auto-scroll (which became more apparent with the new jQuery), as well as several small issues with global and reader views.
Simplified a little bit the code at the same time.
Related to these old issues :
https://github.com/FreshRSS/FreshRSS/issues/634
https://github.com/FreshRSS/FreshRSS/issues/275

Alkarex:
It would be nice if you could have a quick look at this one, @marienfressinaud

marienfressinaud:
I found a small bug on the global view: when I select an item (by clicking on the header or by pressing the next key), the previous article header is truncated. Here is a screenshot illustrating the problem:

capture d ecran de 2016-01-26 19-37-21

marienfressinaud:
And thanks for maintaining FRSS while I don't do that :)

Alkarex:
Should be fixed now https://github.com/Alkarex/FreshRSS/commit/f62ab5fe5bb15a846a653e291f42828df090426e
But I have just spotted another bug when using autoshare in the global view https://github.com/Alkarex/FreshRSS/blob/f62ab5fe5bb15a846a653e291f42828df090426e/p/scripts/main.js#L436

marienfressinaud:
Fine for me. JavaScript is too messy to don't have any bug at each modification anyway :(

Alkarex:
:-P But to be fair, not all the bugs fixed here are due to the new jQuery

Alkarex:
See dedicated issue https://github.com/FreshRSS/FreshRSS/issues/1076 for the autoshare bug in the global view

PR #1070, opened 2016-01-28
initial version of freshrss port

jon1012:


Alkarex:
Looked interesting :-) Do not hesitate to send another PR.

jon1012:
Thanks ! Yeah, it's not finished, that's why I canceled it.
I'll send a PR when the port is finished.

Issue #1075, opened 2016-02-14
inline scripts and Content-Security-Policy


oupala:
FreshRSS contains some inline javascript on the homepage (in `/i/` directory):

``` js
<script src="../scripts/bcrypt.min.js?1443537072" defer="defer" async="async"></script>
        <script>//<![CDATA[
...
```

This is a bad habit as inline scripts and css are a good entrance for XSS attacks. See https://scotthelme.co.uk/content-security-policy-an-introduction/

With a secure configuration such as `Content-Security-Policy :"default-src 'self'"`, FreshRSS do not work any more. The configuration has to be set to `Content-Security-Policy :"default-src 'self' 'unsafe-inline'"` which is working but less secure.

A good improvement would be to follow [this advice](https://scotthelme.co.uk/content-security-policy-an-introduction/#realworldlimitations):

> To be able to accommodate this, you need to externalise the content of any `<script>` tags. Instead of a block of code for your Google Analytics, you would have to create an external file and reference it like `<script src="/assets/js/ga.min.js"></script>` instead. That's not such a big problem.

Alkarex:
Hello @oupala 
And thanks for your suggestion. The reason for the inline JavaScript is that it contains dynamically generated values. Putting them in an external file would make another HTTP request that cannot be cached, so would hurt the performances.
I will try to think about other approaches (e.g. through cookie, local cache, URL parameters, ...).

Alkarex:
https://hacks.mozilla.org/2016/02/implementing-content-security-policy/

Alkarex:
@oupala See a first draft https://github.com/FreshRSS/FreshRSS/issues/1078 using a ~~one-time cookie~~ (edit: now inline JSON) instead of the inline dynamic JavaScript. There are still several places to change (such as in the install script and statistics). There are also a few `onclick` events here and there. CSS would need some work too.
The main features are working fine with
`Content-Security-Policy: default-src 'self'; img-src * data:; media-src *; style-src 'self' 'unsafe-inline'`
which I have currently hard-coded https://github.com/Alkarex/FreshRSS/blob/e4a459a6edc40b64cba7845b52f3e90666b2818a/app/FreshRSS.php#L171
Anyway, let me know what you think with this version.

oupala:
Excellent @Alkarex !

[...]

PR #1099, opened 2016-03-01
CSP compatibility with Firefox older than 45

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1075
https://github.com/FreshRSS/FreshRSS/pull/1078
https://developer.mozilla.org/en-US/Firefox/Releases/45#Security
https://bugzilla.mozilla.org/show_bug.cgi?id=1045891

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2016-03-xx FreshRSS 1.3.1-beta
 
 * Security
-	* Added CSP `Content-Security-Policy: default-src 'self'; child-src *; img-src * data:; media-src *` [#1075](https://github.com/FreshRSS/FreshRSS/pull/1075)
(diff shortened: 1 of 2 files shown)


Alkarex:
Firefox 44 only knows `frame-src` (and not `child-src`) but Firefox 45 emits a warning when `frame-src` is used... I am not sure of the best approach.

Alkarex:
Firefox 45 (which solves the problem - I am running the beta) will be released on 8 March 2016
https://wiki.mozilla.org/Releases#Based_on_Gecko_45

Alkarex:
Also needed for IE, Edge, Safari

Alkarex:
`frame-src` is back from the dead and un-deprecated https://www.w3.org/TR/CSP3/#directive-frame-src
> The frame-src directive, which was deprecated in CSP Level 2, has been undeprecated, and a worker-src directive added. Both defer to child-src if not present (which defers to default-src in turn).

Issue #1047, opened 2015-12-25
Deleting a feed in statistics page result in an error


Alwaysin:
Steps to reproduce :
- Go to `http://<freshrss>/i/?c=stats&a=idle`
- Click on "delete" on one feed

Then you are redirected to `http://<freshrss>/i/http%3A%2F%2F<freshrss>%2Fi%3Fc%3Dstats%26a%3Didle` and you get this :

```
Not Found

The requested URL /i/http://<freshrss>/i?c=stats&a=idle was not found on this server.
```

But the feed is indeed deleted.

Alwaysin:
https://github.com/FreshRSS/FreshRSS/blob/dev/app/views/stats/idle.phtml#L29, I've replaced

```
formaction="<?php echo _url('feed', 'delete', 'id', $feed['id'], 'r', $current_url); ?>"
```

with

```
formaction="<?php echo _url('feed', 'delete', 'id', $feed['id']); ?>"
```

and it works, but we are redirected to `http://<freshrss>/i/?c=subscription`

Alkarex:
@Alwaysin  Could you please provide a pull request with your modification?

marienfressinaud:
The idea was to stay on the idle statistics page, I would prefer a PR which fix the redirection instead of redirecting to the subscription page...

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/1109

PR #1109, opened 2016-03-06
Fix redirection while deleting a feed

aledeg:
Before, when deleting a feed from the statistics idle page, there was an error in the redirection process.
Now, the redirection works properly and redirects to the idle page.

See #1047

### DIFF
--- a/app/views/stats/idle.phtml
+++ b/app/views/stats/idle.phtml
@@ -6,10 +6,10 @@
 	<h1><?php echo _t('admin.stats.idle'); ?></h1>
 
 	<?php
-		$current_url = urlencode(Minz_Url::display(
+		$current_url = Minz_Url::display(
 			array('c' => 'stats', 'a' => 'idle'),
 			'php', true
-		));
+		);
 		$nothing = true;
 		foreach ($this->idleFeeds as $period => $feeds) {
 			if (!empty($feeds)) {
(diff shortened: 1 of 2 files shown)


Alkarex:
There seems to be a remaining bug. I am redirected to `.../p/i?c=stats&a=idle` with a slash missing after the `/i`

Alkarex:
Probably a bug in `Minz_Url::display()`

Alkarex:
Seems to work with this additional fix:
https://github.com/aledeg/FreshRSS/blob/5f04462e554e4604c04c2c9ca2fe97df6f6f6376/lib/Minz/Url.php#L30

``` php
            if ($url_string === PUBLIC_TO_INDEX_PATH) {
                $url_string = Minz_Request::guessBaseUrl();
            } else {
                $url_string .= '/';  //This was missing
            }
```

aledeg:
That's weird because I don't have that problem with my test instance

Alkarex:
In my case, it is only visible when looking at the HTTP requests in the browser and/or the Apache logs, because Apache ends-up doing the proper redirection

Alkarex:
Can you please try to see if it still works in your case with my additional patch, and if yes, include it in your PR?

aledeg:
done

Alkarex:
Thanks :-)

Issue #1049, opened 2016-01-05
All articles appear unread after upgrade


jjk-jacky:
So I was still running FreshRSS 1.0.0 and decided to (finally) look into updating (to 1.2.0), but got a "little" issue. After the upgrade, once a feed is actualized (whether manually from FreshRSS or via actualize_script.php) every "recent" article will appear unread, much like if I was adding a new feed.

I'm now running 1.1.0 (at commit ca58a265 to be exact), since that's the last one to work fine. After the merge in 185dba88 the problem appears. Seems some db format changes took place, I tried to have a look in the db and they were apparently created just fine, so I'm assuming that part was ok (didn't see any notes about manual operation needed to perform/help with the conversion in changelog).

I have to say I don't know anything about FreshRSS internals, so I just quickly tried to set all rows in entry with lastseen=0 to a current timestamp, in case, to no avail.

If anyone knows how to fix this, or has a clue that could help, I'd appreciate it. (Manually updating every feed one by one and having to restore which articles were read and which weren't is obviously not a good solution...)

Thanks.

PHP 5.6.16; SQLite 3.9.2

Alkarex:
Hello,
Please try to uncheck the option "Mark updated articles as unread" in
FreshRSS configuration, and see if it works with the latest FreshRSS
version.

Alkarex:
If this solves your problem, see also https://github.com/FreshRSS/FreshRSS/issues/1044#issuecomment-166002227 and https://github.com/FreshRSS/FreshRSS/issues/798#issuecomment-89655370

jjk-jacky:
Hi,

Thanks! That seems to have done the trick indeed, now feeds are
"properly" updated. Based on the dev logs, it seems entries go from an
empty hash to a new one, I'm guessing this hash is what's used to
determine if a post changed, and since obviously all the "old" ones
have no hash, they were seen as updated.

I'm also guessing that after a while, leaving time for all entries to
have a hash, I should be able to turn the option on and it would
then work as expected.

Maybe when the hash is empty, it should only be silently updated to a
new value without been marked unread, to avoid such issue (Notably, as
here, in case of an upgrade) ?

Also, this is out of topic (sorry), but in the logs I also see messages
such as this one:
    Error while PubSubHubbub unsubscribing from
    http://www.optipess.com/feed/
Even though I have added
    'pubsubhubbub_enabled' => false,
in my config.php (unless I'm mistaken it should be disabled by default,
[...]

PR #1111, opened 2016-03-06
SQLite: Articles from old version were marked as unread on upgrade

Alkarex:
Bug fixed https://github.com/FreshRSS/FreshRSS/issues/1049


jjk-jacky:
Did a quick test w/ an old db, and it does fix the post-upgrade issue reported in #1049, even with "Mark updated articles as unread" enabled everything seems to work fine/as expected now.

Alkarex:
Thanks for the feedback @jjk-jacky :+1:

PR #1132, opened 2016-04-22
MySQL compatibility only_full_group_by

Alkarex:
Fix MySQL error, which appeared in MySQL 5.7:

`ERROR 1055 (42000): Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'freshrss.c.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`

The consequences were that categories and feeds were gone in the left menu.


Alkarex:
Needed for Ubuntu 16.04 LTS

Issue #1089, opened 2016-02-23
Prepare for HTTP/2


Alkarex:
- Most browsers support HTTP/2 http://caniuse.com/#search=http2
- Apache 2.4.17 (October 2015) ships with mod_http2 by default http://www.apache.org/dist/httpd/CHANGES_2.4.17 https://httpd.apache.org/docs/trunk/mod/mod_http2.html
- Ubuntu 16.04 LTS will have Apache 2.4.18 http://distrowatch.com/table.php?distribution=Ubuntu as well as Fedora 24

https://www.smashingmagazine.com/2016/02/getting-ready-for-http2/

Alkarex:
Presentation about HTTP/2 https://docs.google.com/presentation/d/1r7QXGYOLCh4fcUq0jDdDwKJWNqWK1o4xMtYpKZCJYjM/present

https://blog.samat.org/2015/11/26/Enabling-HTTP2-on-Apache-2.4-on-Debian-Ubuntu/

PR #1133, opened 2016-04-23
HTTP2 optimization

Alkarex:
Fast flush HTTP headers, push promise CSS.
Requires PHP 5.3+ due to anonymous function.
Do not load syles, scripts, and notifications for Ajax requests.
https://github.com/FreshRSS/FreshRSS/issues/1089


Alkarex:
With this patch, when using e.g. Apache with mod_http2 http://httpd.apache.org/docs/trunk/mod/mod_http2.html one can see in e.g. Chrome debugger `HTTP2_SESSION_RECV_PUSH_PROMISE`

PR #1147, opened 2016-05-01
Fix validator link

Frenzie:


app/views/helpers/feed/update.phtml
```diff
@@ -48,7 +48,7 @@
 					<a class="btn" target="_blank" href="<?php echo $this->feed->url(); ?>"><?php echo _i('link'); ?></a>
 				</div>
 
-				<a class="btn" target="_blank" href="http://validator.w3.org/feed/check.cgi?url=<?php echo $this->feed->url(); ?>"><?php echo _t('sub.feed.validator'); ?></a>
+				<a class="btn" target="_blank" href="http://validator.w3.org/feed/check.cgi?url=<?php echo urlencode(str_replace('&amp;', '&', $this->feed->url())); ?>"><?php echo _t('sub.feed.validator'); ?></a>
```
Alkarex: Thanks!
I think instead of `str_replace()`, it would be better to use:

``` php
htmlspecialchars_decode(..., ENT_QUOTES);
```

Furthermore, regarding the handling of spaces, if think `rawurlencode()` should be used instead of `urlencode()`.

Alkarex:
Example of link to test, with plenty of special chars:
http://cordis.europa.eu/search/result_en?format=rss&srt=/result/contentUpdateDate:decreasing&q=/result/relations/categories/resultCategory/code=%27story%27%20AND%20language=%27fr%27

`http://cordis.europa.eu/search/result_en?format=rss&srt=/result/contentUpdateDate:decreasing&q=/result/relations/categories/resultCategory/code='story' AND language='fr'`

app/views/helpers/feed/update.phtml
Frenzie: Interestingly, the validator itself opts for `+` instead of `%20`. But yes, I didn't realize `urlencode` didn't abide by the URI RFC.

Frenzie:
Actually FreshRSS refuses to add that test link in the latter form (with spaces in it), at least on stable.

The error message is a cryptic:

```
`` is not a valid URL
```

In any case, it's understood properly by the validator if I add the first form to FreshRSS.

Alkarex:
Thanks!

Frenzie:
Oh, I only just realized I probably should've used the dev branch instead?

Alkarex:
Ah yes indeed. I did not pay enough attention either!

Frenzie:
To be fair to me _cough, cough_, normally the dev branch (or whatever name it goes by) is what you get by default when you check out. =P

Alkarex:
Should be repaired now :-)

PR #1085, opened 2016-02-17
1029 translation to russian

asmfreak:
This PR adds russian translation for FreshRSS.
I will bump this PR when I'll translate other files.

PR #1150, opened 2016-05-03
A share for Wallabag v2

AmauryCarrade:
The new version of Wallabag (2) was rewrote completly, and the sharer changed.

This PR keeps the old Wallabag v1 sharer, for people still using it, and adds a new one for Wallabag v2. The old sharer for wallabag v1 is renamed to `wallabag v1` (instead of `wallabag`).

Please note that as this changes a couple of translations, it may impact #1085.


Alkarex:
Thanks :+1:

PR #1172, opened 2016-07-08
Sticky sidebar

romibi:
With lots of articles on one page, all expanded (how i like it), it was a bit inconvenient to change the feed…

I solved this by making the Sidebar sticky with a jQuery Plugin ["Sticky-Kit"](http://leafo.net/sticky-kit/)

For the plugin to work I needed to disable the defer and async on jQuery script tag.

I hope the "mobile check" is ok this way…

Before:
[https://gyazo.com/2816b7b2a928a97a8bebe5277f04ea7d](https://gyazo.com/2816b7b2a928a97a8bebe5277f04ea7d)
After:
[https://gyazo.com/059b66ce0590eb9cc96fa7e92e9abda5](https://gyazo.com/059b66ce0590eb9cc96fa7e92e9abda5)

(sorry for creating the PR twice … choose the wrong target branch first)


marienfressinaud:
There is an extension for that ;) https://github.com/FreshRSS/Extensions

romibi:
ooh …  
I didn't see the Extensions … sorry … (perhaps the Extensions should be linked in the main readme)

Thou i think this should be default …  Therefore i'll keep this PR open for the moment …
Just close this if you don't think so …

Alkarex:
@romibi Thanks :+1: 
I have merged your proposition, after a few changes https://github.com/FreshRSS/FreshRSS/commit/a180485f9146f22d48470f3ba541317dbc2ee076
Please give a try to the /dev branch of FreshRSS if you can

romibi:
Nice, thanks…
Yes I'll give the dev branch a try once i find time … (I'm having exams the next few weeks)
Probably everything is ok :smile:

Alkarex:
One more little change https://github.com/FreshRSS/FreshRSS/pull/1214 for recalc after auto loading more articles

Alkarex:
I can see a little bug when collapsing the categories

romibi:
sounds familiar and i think i fixed it … can't remember exactly … but wasn't that the reason why i added those 3 'false' to the jquery append script function?
don't really have time to test it at the moment …
the last few weeks i was testing the extension marienfressinaud mentioned instead of this modification … so i can't exactly remember if it worked 100% fine …

Alkarex:
The tree `false` were to ensure that jQuery was loaded first. But I ensured another way (keeping async and defer) and this is not the reason for this bug.

PR #1184, opened 2016-07-30
PSHB: work-around for SuperFeeder

Alkarex:
Feeds using https://superfeedr.com/ for PubSubHubbub push had timeout problems during substription.
SuperFeeder bot was returning 422 "We could not verify your callback Error: ETIMEDOUT"
It seems to be due to the the fact that SuperFeeder bot uses keep-alive but expects the server to close the connection.
https://github.com/FreshRSS/FreshRSS/issues/312#issuecomment-73716936


Alkarex:
@julien51 Niceness for you :-)
I have sent more info by e-mail, but please comment here if you have any technical feedback.

julien51:
Hello! Thanks for the message and sorry for the delay! I don't think we have changed anything here in a couple weeks.
Could you provide more details?

Basically the error code you're seeing happens when we perform the verification of intent synchronously and we are unable to reach the callback (ETIMEDOUT).

Alkarex:
@julien51 I have just forwarded the additional info to your GitHub e-mail

PR #1185, opened 2016-07-30
PSHB: Articles in the right order

Alkarex:
When push by PubSubHubbub contains multiple articles, ensure they are in
the right order (do not apply the reversing used for SimplePie sources)


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/312

Issue #955, opened 2015-08-21
AnonymousReferer should anonymize all links


oupala:
Marien started to write an extension called AnonymousReferer.

This extension aims at deleting the http referrer when the FreshRSS user click on a post to see it on the original website. The http referrer is deleted by the use of an anonymous redirection service known as **anonym.to**, deleting the referrer is as simple as adding the `http://anonym.to/?` prefix to each link. The extension is available for [download](http://marienfressinaud.fr/data/files/upload/xExtension-AnonymousReferer.zip).

Unfortunately, the extension is not as complete as the original extension that came from KrissFeed. The extension only deletes the referrer for the link to the original post while ignoring all links inside the post.

The extension from KrissFeed has the following options :
- leave the configuration field empty to not modify the behaviour (leave links as is)
- fill-in the field with `http://anonym.to/?` to use the **anonym.to** service (what is _partially_ done by the actual extension)
- fill-in the url of another service which API si similar (in case **anonym.to** would be blacklisted by your ISP)
- fill-in `noreferrer` to use the html5 property

The KrissFeed extension does only change links (but **all** links) and not other media files which are then downloaded directly (images and others).

Alkarex:
See https://www.w3.org/TR/referrer-policy/
https://developer.mozilla.org/en/docs/Web/Security/CSP/CSP_policy_directives#referrer

``` http
​Content-Security-Policy: default-src 'self'; referrer origin; ...
```

http://caniuse.com/#search=referrer

JD557:
I don't see any extension for this. Does it exists?

Anyway, can't we just add `<meta name="referrer" content="no-referrer" />` to all pages (or add an option to insert this)?

https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-delivery-meta

Alkarex:
@JD557 See my comment just above yours. You should be able to use a CSP rule in your Web server configuration

Alkarex:
P.S. @JD557 No, no extension exists

Alkarex:
I will make a patch directly in FreshRSS

Alkarex:
The appropriate directive would be `same-origin` https://www.w3.org/TR/referrer-policy/#referrer-policy-same-origin but:
- In Mozilla Firefox 48:
[...]

PR #1198, opened 2016-08-07
meta referrer origin

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/955
Tested in Firefox 48, Chrome 53, Edge 25

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -41,6 +41,8 @@
 		$url_rss['a'] = 'rss';
 ?>
 		<link rel="alternate" type="application/rss+xml" title="<?php echo $this->rss_title; ?>" href="<?php echo Minz_Url::display($url_rss); ?>" />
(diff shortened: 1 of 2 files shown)


Frenzie:
I think this may have broken the ability to update stuff in extensions, unless something went wrong for me while upgrading to 1.5.

> Error 403 - Forbidden
> 
> You don’t have permission to access this page [HTTP_REFERER=]
> ← Go back to your RSS feeds

Frenzie:
Specifically it's this codepath that gets executed: https://github.com/FreshRSS/FreshRSS/blob/17c8c039df675b3b0f8d88d14f7316a240eabe76/app/FreshRSS.php#L58-L73

Alkarex:
Yes indeed. Related issue https://github.com/FreshRSS/FreshRSS/issues/1253
I believe the CSRF token is missing for the extensions.

Issue #1227, opened 2016-08-25
Update all Feed not working


Barrnet:
Greetings,
I'm experiencing a strange bug on my installation of FreshRSS that's affecting - version 1.2.0. - 1.4.0. It's impossible to update any feed via the update button in the webgui or thought php-cli command "/app/actualize_script.php", the webgui simply print "There is no feed to refresh…", sometimes it prints in the FreshRSS log  page "feed error: [insert here feed log]", and after that a red allert icon appear in the category .
The error_log doesn't show anything, and if I try to download the last RSS from curl it works, at this point I don't know what to do.

My configuration:
OS: Debian 8
Web-Server: nginx .10.1 (for the FressRSS Virtual Host I used your file configuration in the documentation)
PHP: 5.6.24-0+deb8u1 (via php-fpm)
MySQL: 5.5.50-0+deb8u1-log
And i'm using Cloudflare for DNS and caching.

That is my nginx configuration:
`server {
  listen 80; # http on port 80
  #listen 443 ssl; # https on port 443

  # https configuration 
  # ssl on;
  # ssl_certificate      /etc/nginx/server.crt;
  # ssl_certificate_key  /etc/nginx/server.key;

  # your server's url(s)
  server_name newser.videogamezone.eu;

  # the folder p of your FreshRSS installation
  root /home/newser/p/;

  index index.php index.html index.htm;

  # nginx log files
  access_log /home/newser/rss.access.log;
  error_log /home/newser/rss.error.log;

  # php files handling
  # this regex is mandatory because of the API
  location ~ ^.+?.php(/._)?$ {
    fastcgi_pass unix:/var/run/php5-fpm-newser.sock;
    fastcgi_split_path_info ^(.+.php)(/._)$;
    # By default, the variable PATH_INFO is not set under PHP-FPM
    # But FreshRSS API greader.php need it. If you have a "Bad Request" error, double check this var !
    fastcgi_param PATH_INFO $fastcgi_path_info;
    include fastcgi_params;
    fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
  }

  location / {
    try_files $uri $uri/ index.php;
  }
}
`
That is my php5-fpm configuration:
`
[newser.videogamezone.eu]
user  = newser 
group = newser
# listen = 127.0.0.1:9010

listen = /var/run/php5-fpm-newser.sock
listen.owner = www-data
listen.group = www-data
[...]

Issue #1219, opened 2016-08-18
Call for testing /dev before version 1.5


Alkarex:
The future version 1.5 of FreshRSS is [ready](https://github.com/FreshRSS/FreshRSS/milestone/18) for testing. There is a nice changelog https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md with new features and bug corrections, all in all a good improvement since the [previous version](https://github.com/FreshRSS/FreshRSS/issues/1113) :-)

Please give a try to the [/dev version](https://github.com/FreshRSS/FreshRSS/tree/dev) if you can.
Watch out, login by Mozilla Persona has been removed, and the minimum versions for PHP and MySQL have slightly increased.

@marienfressinaud @aledeg @galuel @cjrobe @ecmu @romibi @j8r @AmauryCarrade @Frenzie @cborne @Alwaysin @lapineige @NerosTie @JD557 @Wanabo @Leepic ...

lapineige:
Happy to see a new version so soon :) (with a good list of features)

I'll try to test it on Yunohost. Unfortunately, the maintainer of the app (https://github.com/YunoHost-Apps/freshrss_ynh) is in holidays, so I guess I'll not be able to do so soon.

marienfressinaud:
Mmm, I just realised the last version was 1.4, I didn't pay attention when you send me an email but since we don't have `beta` anymore, the next stable version can be 1.5 (sorry about that!)

Wanabo:
Great to see a new version. I setup a new install of 1.5.0-dev.

First step installation: Choose a language. Need to save/submit it before proceeding. Is it possible when you select a language from the dropdownbox and then you only have to click go to the next step to make the language choice stick? 
Selected English for installation.
Step 3: HTTP username and HTTP password should be DB username and DB password
Database should be DB name

Installation of 1.5.0-dev went fine.
Imported OPML file and all feeds and categories were ok.
All the descriptions of the feeds are empty. I would request an extra option to check for feed descriptions once during import or as an extra button to retrieve the description manually. If a feed has a description then no auto update of the description should take place. 

Thanks for this new version.

Wanabo:
Was playing with enabling/disabling lazy load setting because I noticed html errors. Enabled or disabled both settings produces lazyload="" postpone="" in output html. I was hoping to suppress html errors by disabling lazy load.

Alkarex:
[...]

PR #1229, opened 2016-08-28
PSHB base_url bug

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1227
Bug if base_url does not end with /


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1219

PR #1231, opened 2016-08-29
HTTP security: prevent reflection abuse

Alkarex:
https://msdn.microsoft.com/en-us/library/gg622941(v=vs.85).aspx


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1219

Issue #1245, opened 2016-09-07
target="_blank" vulnerability


Cypouz:
As explained [here](https://dev.to/ben/the-targetblank-vulnerability-by-example), it would be nice to use `rel="noopener noreferrer"` with `target="_blank"` in links to avoid any risk of phishing.

Alkarex:
Interesting :+1: 
I will have to check if this not already covered by `<meta name="referrer" content="never" />`
https://github.com/FreshRSS/FreshRSS/pull/1210

Alkarex:
I have quickly checked that `<meta name="referrer" content="never" />` is not enough, and that `rel="noreferrer"` works in Firefox 48, Edge 25, Chrome 53.

Alkarex:
Pull request https://github.com/FreshRSS/FreshRSS/issues/1246

Alkarex:
Should be solved in /dev branch. Please give it a try.

PR #1246, opened 2016-09-07
Prevent a target _blank attacks with window.opener

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1245
https://mathiasbynens.github.io/rel-noopener/
noopener is implied by noreferrer https://html.spec.whatwg.org/multipage/semantics.html#link-type-noreferrer
The API for window.open() does not seem stable yet https://bugzilla.mozilla.org/show_bug.cgi?id=1267339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 2016-XX-XX FreshRSS 1.6.0-dev
 
+* Security
+	* Prevent `<a target="_blank">` attacks with `window.opener` [#1245](https://github.com/FreshRSS/FreshRSS/issues/1245)
 * UI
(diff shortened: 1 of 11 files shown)


marienfressinaud:
There is still a risk if a link from an article contains a `<a href="..." target="_blank" />` link, isn't it? Or is `target` attribute filtered?

Alkarex:
If there is no existing `target="_blank"`, we add one anyway https://github.com/Alkarex/FreshRSS/blob/6450839b87b6a05b71af87490325acda9ffd5a44/p/scripts/main.js#L755 but that is fine with the `rel="noreferrer"` (which implies `rel="noopener"`)

marienfressinaud:
Oops, I didn't pay attention to that line!

Issue #1249, opened 2016-09-10
Number of RSS items in public URL


ryancom16:
Related issue: #1018

Currently, the number of items in the public rss matches those shown on the web page. Generally this is not a problem unless the feed has a lot of content.

Request: 
Allow a parameter in the public url that would specify the number of items retrieved in the public rss.

Ex: 
http://Domain/FreshRSS/p/i/?a=rss&get=c_6**&Items=300**&token=xxxxxxx

Reason: 
For feeds that contain a lot of content, The Freshrss web ui is slowed down a lot(Nearly freezes) when that feed is loaded into view. This is made much worse when the number of items per page is increased.

I have an external program that monitors the public FreshRss feeds (at the category level) at regular intervals. There are many feeds in this category and updates can be frequent. In order to make sure that public feed displays all updates, I have set the feed limit per page to a very high number (300). 

This var is global so the feed with a lot of content also gets loaded at 300 per page. This causes massive lag to the system. If the item number could be specified at the public url level, then the global value could be set lower and system integrity and performance would increase dramatically.

Alkarex:
Here is a proposition https://github.com/FreshRSS/FreshRSS/pull/1251
Could you please test it?

Alkarex:
Related: https://github.com/FreshRSS/FreshRSS/issues/848

Alkarex:
Done

PR #1251, opened 2016-09-11
Better control of number of entries per page or RSS feed

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1249
- Since X hours: `https://freshrss.example/i/?a=rss&hours=3`
- Explicit number: `https://freshrss.example/i/?a=rss&nb=10`
- Limited by `min_posts_per_rss` and `max_posts_per_rss` in user config


Alkarex:
Related: https://github.com/FreshRSS/FreshRSS/issues/848

ryancom16:
**Test**: Pass

**Limits Tested**:
-user/config.php
'since_hours_posts_per_rss'
'min_posts_per_rss'
'max_posts_per_rss'

-url:
&hours=#
&nb=#

**notes**:
Fantastic work, all the limits are working correctly. Limits in config.php override url parameters as expected. &hours is appended to the rss feed when it is generated. One thing of note, when both &hours and &nb are included in the url, hours overrides nb. This is not a problem because users should be using one or the other.

ryancom16:
note: Existing users will have to submit a change to their config in the web UI in order for the parameters to bin included in config.php.

Future suggestions: Have these parameters configurable in the web UI.

Alkarex:
Thanks for the feedback :-)
Merged in the /dev branch

alexis-:
Hi. First, thanks and drum rolls for this excellent project.
I was looking for a solution to aggregate and archive RSS feeds when I found FreshRSS.

It is my understanding that the _hours_ parameters of the _a=rss_ module allows one set a time window, in hours, for RSS retrieval, and based on the entries **pubDate**.

After toying with `https://demo.freshrss.org/i/?a=rss&hours=240` I found that items from 2018 were displayed in response, which is obviously way past the 240h window.

Is it intended behavior, or a glitch ?

Thanks !

Frenzie:
I would imagine it's retrieval time, unless you mean these items were actually retrieved in 2018.

Alkarex:
@alexis- : Indeed, as @Frenzie writes, `hours` filters on the discovery date, not the publication date. If you would like to filter out articles based on their declared date (which is not reliable) you can use the search parameter using a pubdate filter, such as:

`https://freshrss.example.net/i/?a=rss&hours=240&search=pubdate:PT250h`

Please open a new issue if needed.

P.S.: Do keep `hours` in the query or a `date:` filter in the search query, as filtering only on `pubdate` is slow (not indexed).

alexis-:
Excellent answer. That makes sense and I understand now.

Issue #1247, opened 2016-09-08
Subscribe URL 


jwhipp:
I'd like to get FreshRSS working with this tool: 

https://chrome.google.com/webstore/detail/rss-subscription-extensio/nlbjncdgjeocebhnmkbbbdekmmmcbfjd

One of the things required to configure it to work with FreshRSS is a URL for subscriptions as indicated in the image below:

![selection_004](https://cloud.githubusercontent.com/assets/556332/18363030/917b8d1a-75c5-11e6-99ca-3c8dd471691c.png)

What URL beyond https://myfreshrss.example/p/i/ to make this work?

Alkarex:
Without having tried, something like that:
`https://myfreshrss.example/p/i/?c=feed&a=add&url_rss=%s`

P.S.: If you can, try to have your Web server pointing to the `./freshrss/p/` directory instead of `./freshrss/` so your URL becomes `https://myfreshrss.example/i/`

jwhipp:
Ok I tried that and I get this:

The data could not be converted to UTF-8. You MUST have either the iconv or mbstring extension installed. Upgrading to PHP 5.x (which includes iconv) is highly recommended. [http://www.noobslab.com/feeds/posts/default]

I am running PHP 7 so iconv should be in it.   I also installed php-mbstring, to no avail.

If I grab the URL manually and plug it in, it works fine.

Alkarex:
I have just tested this Chrome extension, and it seems to work fine for me, also when testing with `http://www.noobslab.com/`. My setup is running on Ubuntu 16.04 with PHP 7.0.8, and the extensions listed on https://github.com/FreshRSS/FreshRSS/#example-of-full-installation-on-linux-debianubuntu
In particular, try:

``` sh
apt install php-intl php-mbstring
```

Please provide more information about your setup if you still get problems.

jwhipp:
I came across a different URL I wanted to add and it worked => http://feeds.arstechnica.com/arstechnica/index/ 

There must've been something mangling it before, but I am not sure.

PHP 7.0.8-0ubuntu0.16.04.2 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.8-0ubuntu0.16.04.2, Copyright (c) 1999-2016, by Zend Technologies

I had already added mbstring and tried it again before I submitted the issue.

[...]

PR #1292, opened 2016-10-07
Show address for adding feeds from an external software

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1247#issuecomment-248118352
I am not so glad about how it looks like. Alternative suggestions welcome.

### DIFF
--- a/app/views/subscription/index.phtml
+++ b/app/views/subscription/index.phtml
@@ -46,7 +46,6 @@
 			</div>
 			<button class="btn" type="submit"><?php echo _i('add'); ?></button>
 		</div>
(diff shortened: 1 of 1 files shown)


kevinpapst:
You already have these kind of URLs in the admin section (in the authenticate screen for example).
Just a quick idea that passed my mind: putting these URLs behind a "info" icon is not uncommon. If you combine that with a click action, which would show the URL in a flash message (for copy&paste) you could 1) beware the screen of getting cluttered and 2) would not confuse the regular users?

aledeg:
Maybe we could add an entry in the subscription tool page.
At the moment, we have *Bookmarklet* and *Firefox feed reader*. I think this will be a perfect location for that information.
What do you think ?

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/1660 instead

Issue #1258, opened 2016-09-20
Auto marking as read and hiding is not woking.


ghost:
I may being stupid, but the app isn't working as i'd expect. 

I had checked:
Mark article as read… [x] while scrolling 
and
Hide articles after reading  [x]

Then i go to the "reading" mode, and look at the articles, skipping then with j. The old ones wont get hide, even if i reload the page.

Where is the mistake?

Alkarex:
@aledeg Do you think you could have a chance to look at this bug?

Alkarex:
@techmago Could you please try this patch https://github.com/FreshRSS/FreshRSS/pull/1309 and let me know what you think?

Alkarex:
Merged in the /dev branch. Tests welcome :-)

PR #1309, opened 2016-10-13
Scroll as read

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1258
https://github.com/FreshRSS/FreshRSS/issues/707
Should works for all views. Simplified logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 	* Experimental support for PostgreSQL [#1195](https://github.com/FreshRSS/FreshRSS/pull/1195)
 	* New client supporting FreshRSS on Linux: FeedReader 2.0+ [#1252](https://github.com/FreshRSS/FreshRSS/issues/1252)
 * Features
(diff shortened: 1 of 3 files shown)


Alkarex:
Addresses some of the needs of https://github.com/FreshRSS/FreshRSS/issues/329

chemical1979:
In the german translation, "... while scrolling" is translated into ".. beim blättern", which is a term that is mainly is used for "switching or skipping pages" which does not match an endless list of elements. 

Background: I just clicked in an article that was not at the top of my feed in the normal view, and after this, I noticed that all other articles in front of that one "were gone" (I noticed this too late and I have no idea which articles slipped through now). But anyway, this is not what I wanted to say: I jumped into the configuration and try to find a setting that matches this behaviour and I looked for sth like "mark as read ... while skipping". But I did not find it. I went here and searched an eventually found this issue.

@Alkarex Long story short: The translation "... while scrolling" should be "... while scrolling/skipping" in English and "... beim Scrollen/Überspringen" in German. :-)

Alkarex:
@chemical1979 Thanks. Would you be able to make a pull request for this change?

chemical1979:
I'm not a git expert, but as this is the second time you ask me, perhaps it is the challenge I need to become one. :-)

Alkarex:
You did very well :-) For reference https://github.com/FreshRSS/FreshRSS/pull/2026

PR #1316, opened 2016-10-15
Remove superflous SQL category request

Alkarex:
A category request was systematically done, although it is not always
needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,10 @@
 * UI
 	* Download icon 💾 for podcasts [#1236](https://github.com/FreshRSS/FreshRSS/issues/1236)
 * Extensions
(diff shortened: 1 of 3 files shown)


Alkarex:
Especially needed now that the number of "mark as read" requests will increase with "scroll as read" https://github.com/FreshRSS/FreshRSS/pull/1309

Alwaysin:
What did motivate the choice of a default "mark read as scroll"? I think this is a bad idea.

Alkarex:
It is an option, so you can turn it off. The main reason is to address the UI problem of marking previous articles as read, for which there is currently no other option in FreshRSS https://github.com/FreshRSS/FreshRSS/issues/329

Alwaysin:
Yes I've turned it off, I just find it very disturbing when you first install an rss reader that it marks the articles as read directly, when you usually go around and scroll to see how it looks. You may think this is a bug.

Issue #1312, opened 2016-10-14
[Postgres] Duplicate violation


anayrat:
Hi,

On a fresh install with dev version I import an OPML. I have plenty of duplicate key violation in logs : 

```
2016-10-14 19:42:22 CEST [14249]: [7-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=23505 ERROR:  duplicate key value violates unique constraint "freshrss_admin_category_pkey"
2016-10-14 19:42:22 CEST [14249]: [8-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=23505 DETAIL:  Key (id)=(1) already exists.
2016-10-14 19:42:22 CEST [14249]: [9-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=23505 STATEMENT:  INSERT INTO "freshrss_admin_category"(name) VALUES($1)
```

```
2016-10-14 19:43:12 CEST [14249]: [10-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=23505 ERROR:  duplicate key value violates unique constraint "freshrss_admin_entry_guid_key"
2016-10-14 19:43:12 CEST [14249]: [11-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=23505 DETAIL:  Key (guid)=(https://commitfest.postgresql.org/11/814/) already exists.
2016-10-14 19:43:12 CEST [14249]: [12-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=23505 STATEMENT:  INSERT INTO "freshrss_admin_entry" (id, guid, title, author, content, link, date, "lastSeen", hash, is_read, is_favorite, id_feed, tags) VALUES($1, $2, $3, $4, $5, $6, $7, $8, decode($9, 'hex'), $10, $11, $12, $13)
2016-10-14 19:43:12 CEST [14249]: [13-1] user=freshrss,db=freshrss,client=[local],application=[unknown],state=25P02 ERROR:  current transaction is aborted, commands ignored until end of transaction block
[...]

PR #1322, opened 2016-10-15
PostgreSQL workaround for SERIAL strange behaviour

Alkarex:
http://stackoverflow.com/questions/18389537/does-postgresql-serial-work-differently
https://github.com/FreshRSS/FreshRSS/issues/1312
https://github.com/FreshRSS/FreshRSS/pull/1195
https://github.com/FreshRSS/FreshRSS/issues/416

### DIFF
--- a/app/SQL/install.sql.pgsql.php
+++ b/app/SQL/install.sql.pgsql.php
@@ -51,7 +51,7 @@ $SQL_CREATE_TABLES = array(
 'CREATE INDEX %1$sis_read_index ON "%1$sentry" ("is_read");',
 'CREATE INDEX %1$sentry_lastSeen_index ON "%1$sentry" ("lastSeen");',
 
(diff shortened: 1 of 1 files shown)


anayrat:
As I said in #1312 
It is not a strange behavior.  See this comment : http://stackoverflow.com/a/18393072
See documentation : https://www.postgresql.org/docs/current/static/datatype-numeric.html#DATATYPE-SERIAL

> To insert the next value of the sequence into the serial column, specify that the serial column should be assigned its default value. This can be done either by excluding the column from the list of columns in the INSERT statement, or through the use of the DEFAULT key word.

When you do : `INSERT INTO "names" (id, name) SELECT 1, 'default' WHERE NOT EXISTS (SELECT id FROM "names" WHERE id = 1);`

Sequence is not incremented (you don't ask Postgres to do). So when you do `INSERT INTO "names"(name) VALUES ('second');`

Sequence is incremented and take the value 1 that's why you get a duplicate key.

Issue #1317, opened 2016-10-15
Bug scroll as read in Global view


Alkarex:
Wrong offset

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1309

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/1324

PR #1324, opened 2016-10-16
Global view scroll

Alkarex:
- Fix global view scroll top
- Add a throttle for scroll event (performance)


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1317
https://github.com/FreshRSS/FreshRSS/pull/1309

Issue #1328, opened 2016-10-16
Call for testing /dev before version 1.6 (second round)


Alkarex:
The future version 1.6 of FreshRSS is [ready](https://github.com/FreshRSS/FreshRSS/milestone/19) for testing. Since the [previous version](https://github.com/FreshRSS/FreshRSS/issues/1219), the theme has been to introduce PostgreSQL as an additional database choice, but there is also a number of new features and bug corrections, as visible [in the changelog](https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md):
- API
  - Support for editing feeds and categories from client applications [#1254](https://github.com/FreshRSS/FreshRSS/issues/1254)
- Compatibility:
  - Support for PostgreSQL [#416](https://github.com/FreshRSS/FreshRSS/issues/416)
  - New client supporting FreshRSS on Linux: FeedReader 2.0+ [#1252](https://github.com/FreshRSS/FreshRSS/issues/1252)
- Features
  - Rework the “mark as read during scroll” option, enabled by default for new users [#1258](https://github.com/FreshRSS/FreshRSS/issues/1258), [#1309](https://github.com/FreshRSS/FreshRSS/pull/1309)
    - Including a _keep unread_ function [#1327](https://github.com/FreshRSS/FreshRSS/pull/1327)
  - In a multi-user context, take better advantage of other users’ refreshes [#1280](https://github.com/FreshRSS/FreshRSS/pull/1280)
  - Better control of number of entries per page or RSS feed [#1249](https://github.com/FreshRSS/FreshRSS/issues/1249)
    - Since X hours: `https://freshrss.example/i/?a=rss&hours=3`
    - Explicit number: `https://freshrss.example/i/?a=rss&nb=10`
    - Limited by `min_posts_per_rss` and `max_posts_per_rss` in user config
  - Support custom ports `localhost:3306` for database servers [#1241](https://github.com/FreshRSS/FreshRSS/issues/1241)
  - Add date to exported files [#1240](https://github.com/FreshRSS/FreshRSS/issues/1240)
  - Auto-refresh favicons once or twice a month [#1181](https://github.com/FreshRSS/FreshRSS/issues/1181), [#1298](https://github.com/FreshRSS/FreshRSS/issues/1298)
    - Cron updates will also refresh favicons every 2 weeks [#1306](https://github.com/FreshRSS/FreshRSS/pull/1306)
- Bug fixing
[...]

PR #1335, opened 2016-10-19
guid and urls should not contain low/high characters

Alkarex:
It looks like SimplePie does not always filter everything
Having a character not in latin1 would create MySQL collate errors

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -314,7 +314,9 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 			if (count($entries) > 0) {
 				$newGuids = array();
 				foreach ($entries as $entry) {
-					$newGuids[] = $entry->guid();
+					$guid = $entry->guid();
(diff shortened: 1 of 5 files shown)


Alkarex:
I got a guid 'http://synd.co/2cuJ0lM​' which ends with `\xE2\x80\x8B` (in ansi: 'http://synd.co/2cuJ0lMâ€‹') from _A List Apart_, which somehow did not get sanitized upstream.

Alkarex:
```
SQL error addEntry: HY000: 1366 Incorrect string value: '\xE2\x80\x8B' for column 'guid' at row 1 while adding entry in feed 5 with title: This week's sponsor: OPTIMAL WORKSHOP

SQL error listHashForFeedGuids: HY000: 1271 Illegal mix of collations for operation ' IN ' while querying feed 5
```

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1328

PR #1340, opened 2016-10-22
CSS line-height bug with sup in dates

Alkarex:
Not visible in Firefox, but in Chrome / Internet Explorer / Edge.
https://github.com/FreshRSS/FreshRSS/issues/534#issuecomment-255519269
https://github.com/FreshRSS/FreshRSS/issues/1328
https://www.cs.tut.fi/~jkorpela/www/linespacing.html


Alkarex:
@Alwaysin

Issue #1355, opened 2016-11-04
Web update permissions access


paul-mesnilgrente:
Hello everyone,

I have used your script in the README to install FreshRSS last week. The only difference is that I have placed the symlink in `/var/www` instead of `/var/www/html/` and I use an apache virtualhost :

```xml
<VirtualHost *:80>
    ServerName ...
    DocumentRoot /var/www/FreshRSS

    ErrorLog ${APACHE_LOG_DIR}/freshrss_error.log
    CustomLog ${APACHE_LOG_DIR}/freshrss_access.log combined
RewriteEngine on
RewriteCond %{SERVER_NAME} =...
RewriteRule ^ https://%{SERVER_NAME}%{REQUEST_URI} [END,QSA,R=permanent]
</VirtualHost>

<IfModule mod_ssl.c>
<VirtualHost *:443>
    ServerName ...
    DocumentRoot /var/www/FreshRSS

    ErrorLog ${APACHE_LOG_DIR}/freshrss_error.log
    CustomLog ${APACHE_LOG_DIR}/freshrss_access.log combined
SSLCertificateFile /etc/letsencrypt/live/...
SSLCertificateKeyFile /etc/letsencrypt/live/...
Include /etc/letsencrypt/options-ssl-apache.conf
</VirtualHost>

</IfModule>
```

The problem is that when I click on "Check the updates", an error message prints that `/usr/share/FreshRSS` is not accessible by the HTTP user.

So I have run again that line `sudo chown -R :www-data . && sudo chmod -R g+r . && sudo chmod -R g+w ./data/` but the problem is not solved. If you have an idea...

Thank you very much.

PS: I am on ubuntu, with FreshRSS 1.5

subic:
You also have to chmod g+w on the application folder (/usr/share/FreshRSS in your case) to allow the webuser group write access to the installation directory. I personally only do this while upgrading and reduce the privileges after (chmod g-w on the application folder).

paul-mesnilgrente:
This work fine, but this not only on the /usr/share/FreshRSS but recursively too.
So the script is

``` bash
sudo chmod -R g+w .
# make the update with the web interface
sudo chmod -R g-w .
sudo chmod -R g+w ./data/
```

Alkarex:
@paul-mesnilgrente can you please show the rights of `/var/www/FreshRSS` and `/usr/share/FreshRSS` ?

Alkarex:
[...]

PR #1356, opened 2016-11-04
Document access-rights for Web updates

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1355

### DIFF
--- a/README.fr.md
+++ b/README.fr.md
@@ -85,13 +85,16 @@ sudo git checkout -b dev origin/dev
 
 # Mettre les droits d’accès pour le serveur Web
 sudo chown -R :www-data . && sudo chmod -R g+r . && sudo chmod -R g+w ./data/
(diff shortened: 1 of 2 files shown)


subic:
I would move it before the "cd FreshRSS" line and include base directory just to be sure (mostly in cases with symlink):

...
sudo git clone https://github.com/FreshRSS/FreshRSS.git
+
+# If you would like to allow Web updates (not needed for updates with git)
+sudo chmod g+w FreshRSS -R
+
cd FreshRSS
...

Alkarex:
Aren't the following two approaches producing the exact same result?
`.` is the base directory

```
sudo chmod g+w FreshRSS -R
cd FreshRSS
```

```
cd FreshRSS
sudo chmod g+w . -R
```

subic:
Yes, my bad, ignore the previous comment, you are correct.

Issue #1380, opened 2016-11-23
Favicons are not loaded


thomasrenes:
I have a issue were favicons of some RSS feeds are not loaded in FreshRSS since I've updated to version 1.6.1. For these RSS feeds Chrome displays the "image not found icon" on the place where the favicon of that feed should be. When manually going to to the URL "[domain]/p/f.php?8fff8de3" I don't get any output.

Example RSS Feed:
https://developer.apple.com/news/rss/news.rss

The favicon.ico file does exist:
https://developer.apple.com/favicon.ico

The file is created on the server but is empty:
```
[favicons]# file 8fff8de3.ico
8fff8de3.ico: empty
[favicons]# cat 8fff8de3.txt
https://developer.apple.com/news
```

```
[tmp]# wget https://developer.apple.com/favicon.ico
--2016-11-23 09:29:47--  https://developer.apple.com/favicon.ico
Resolving developer.apple.com... 17.146.1.14
Connecting to developer.apple.com|17.146.1.14|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 9062 (8.8K) [image/vnd.microsoft.icon]
Saving to: `favicon.ico'

100%[======================================>] 9,062       --.-K/s   in 0s

2016-11-23 09:29:47 (626 MB/s) - `favicon.ico' saved [9062/9062]

[tmp]# file favicon.ico
favicon.ico: MS Windows icon resource - 4 icons, 32x32, 256-colors
```

The permissions on the folders seems to be correct. I've tried manually removing all .ico and .txt files from the folder but the same feeds will have the same issue after the files are recreated.

This issue seems to be for 5 of the 164 RSS feeds I'm using:
```
[favicons]# find . -size 0 | grep ".ico" | wc -l
5
[favicons]# ls -ail *.ico | wc -l
164
```

Alkarex:
A quick test gives an HTTP error 401:

```php
<?php
$favicon_url = 'https://developer.apple.com/favicon.ico';
$c = curl_init($favicon_url);
curl_setopt($c, CURLOPT_HEADER, false);
curl_setopt($c, CURLOPT_RETURNTRANSFER, true);
curl_setopt($c, CURLOPT_BINARYTRANSFER, true);
$img_raw = curl_exec($c);
$status_code = curl_getinfo($c, CURLINFO_HTTP_CODE);
curl_close($c);
echo $status_code, "\n";
```

I will try to find out why

[...]

PR #1385, opened 2016-11-28
Add curl user-agent to retrieve favicon

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1380

### DIFF
--- a/lib/favicons.php
+++ b/lib/favicons.php
@@ -9,7 +9,7 @@ $default_favicon = PUBLIC_PATH . '/themes/icons/default_favicon.ico';
 function download_favicon($website, $dest) {
 	global $favicons_dir, $default_favicon;
 
-	syslog(LOG_DEBUG, 'FreshRSS Favicon discovery GET ' . $website);
+	syslog(LOG_INFO, 'FreshRSS Favicon discovery GET ' . $website);
 	$favicon_getter = new \Favicon\Favicon();
 	$favicon_getter->setCacheDir($favicons_dir);
 	$favicon_url = $favicon_getter->get($website);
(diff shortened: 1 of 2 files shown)


lib/favicons.php
```diff
@@ -18,11 +18,12 @@ function download_favicon($website, $dest) {
 		return @copy($default_favicon, $dest);
 	}
 
-	syslog(LOG_DEBUG, 'FreshRSS Favicon GET ' . $favicon_url);
+	syslog(LOG_INFO, 'FreshRSS Favicon GET ' . $favicon_url);
 	$c = curl_init($favicon_url);
 	curl_setopt($c, CURLOPT_HEADER, false);
 	curl_setopt($c, CURLOPT_RETURNTRANSFER, true);
 	curl_setopt($c, CURLOPT_BINARYTRANSFER, true);
+	curl_setopt($c, CURLOPT_USERAGENT, 'FreshRSS/' . FRESHRSS_VERSION . ' (' . PHP_OS . '; ' . FRESHRSS_WEBSITE . ')');
```
kevinpapst, line 26: Probably doesn't hurt, but I am against exposing the operating system of the webserver. 
Could that line be in the config file to allow users to overwrite the value?

lib/favicons.php
```diff
@@ -9,7 +9,7 @@
 function download_favicon($website, $dest) {
 	global $favicons_dir, $default_favicon;
 
-	syslog(LOG_DEBUG, 'FreshRSS Favicon discovery GET ' . $website);
+	syslog(LOG_INFO, 'FreshRSS Favicon discovery GET ' . $website);
```
kevinpapst, line 12: This one could stay debug, as there is already one info log for the same procedure below

Alkarex: Yes, but not with the same address (two distinct requests) :-)

lib/favicons.php
kevinpapst: Well my comment was indeed without any background knowledge how FreshRSS handles this in all the other requests. Probably that should be streamlined (and configurable for the future).
Actually I just thought I give a little back to a great open source project by doing a code review ... 
I know that browsers spread the useragent information even further (that why I use extensions to switch them) but leaking server side information to other sites is probably not the best idea. Every admin/devop I know (me included) tries to hide every possible information about server-side OS and software versions (like type of webserver, php version and so on). 
Just my 2 cents - like in every code review - sharing thoughts is the most important part :)

lib/favicons.php
kevinpapst: I just imagined what happens in your log when you import an OPML with 200+ feeds ;-)
In most cases the favicon will be hosted on the same domain like the feed, so I would say this information is likely redundant - but you are right - these are two different urls. And this action won't happen to often^^

Alkarex: We should indeed improve the logging options. I think it is important to have logs for the outgoing queries.

PR #1420, opened 2017-02-04
Share to sites based on Known (http://withknown.com)

danc:
Hi,
I propose a new share for sites based on Known, which I often use to publish data from FreshRSS.
tested on latest version 1.6.1


Alkarex:
Thanks @danc 👍

PR #1435, opened 2017-02-23
Fixed empty db-prefix in do-install.php

dswd:
The existing code did not take an empty string as `db-prefix`.


Alkarex:
Thanks! Could you please target the /dev branch instead of /master ?

Issue #1432, opened 2017-02-22
FreshRSS on Cloudron


gramakri:
Hi,
Thanks for FreshRSS! The Cloudron community (@dswd) has contributed a FreshRSS app for Cloudron. You can try it out here - https://my-demo.cloudron.me (install FreshRSS from the App Store).

The code for the app itself is at https://git.cloudron.io/dswd/freshrss-app and also has selenium based tests at https://git.cloudron.io/dswd/freshrss-app/tree/master/test.

Let me know if you had any comments/suggestions. I will make a PR to add Cloudron as an installation method.

Alkarex:
Hello @gramakri 
And thanks for this contribution 👍 

I have a few questions:
1. What happens when there is a new version of FreshRSS?
2. What was the problem that required this patch? https://git.cloudron.io/dswd/freshrss-app/blob/master/change-ttrss-file-path.patch Maybe we can address in FreshRSS itself
3. Did you see there is a CLI (command-line interface)? https://github.com/FreshRSS/FreshRSS/tree/master/cli Could you try to see whether it can be used instead of the manual settings of https://git.cloudron.io/dswd/freshrss-app/blob/master/start.sh
    * Let me know if any other CLI option could help
4. What would you need for Cloudron authentication? https://git.cloudron.io/dswd/freshrss-app/blob/master/POSTINSTALL.md

dswd:
Hi @Alkarex, let me answer these questions as I did the packaging.

1) I am subscribed to Github releases for FreshRSS (via RSS) and I will hopefully be able to package new releases for Cloudron when they come out. I got a personal interest in packaging the new releases as I use FreshRSS on Cloudron for my daily feed reading.

2) The underlying problem of that patch was that Cloudron distinguishes between code and data in a way that makes code read-only. Since the api directory is clearly code I wanted to keep it in the read-only code directory. As the ttrss plugin wanted to write into that api folder a simple symlink approach is not enough as the folder itself needs to be writable. Therefore I went with the small patch that makes the plugin write its file to a writable location instead which in turn allows me to use a symlink in the api folder again. (I hope that was no too confusing)

3) Actually I saw it too late. I did not expect to find such a tool and I did not see it in the documentation (at least not in the english version). As far as I can see it could help me with setting up the initial config and db scheme but currently it does not seem to support overwriting the db settings on already installed instances. Such a feature would be useful to me, yes.

[...]

PR #1436, opened 2017-02-23
Add config option to disable and hide self-update

dswd:
This PR contains the following changes:
* Add a config option `disable_update` to default config (default: `false`)
* Hide the "Update" entry in header an configuration menu when this option is set
* Disable the update action when this option is set

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -162,7 +162,7 @@ class FreshRSS_update_Controller extends Minz_ActionController {
 	}
 
 	public function applyAction() {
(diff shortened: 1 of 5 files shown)


Alkarex:
Thanks!
Related discussion https://github.com/FreshRSS/FreshRSS/issues/1432

Alkarex:
Could I please ask you to add this new option `disable_update` in the CLI interface as well?
https://github.com/FreshRSS/FreshRSS/blob/dev/cli/do-install.php
You can use one of the other boolean parameters such as `allow_anonymous` as example

PR #1439, opened 2017-02-24
Added cli tool to reconfigure FreshRSS

dswd:
This PR adds a tool `cli/reconfigure.php` that takes the same parameters as `do-install.php` (including `disable_update`) and changes them in `config.php`.


Alkarex:
Thanks!

PR #1465, opened 2017-03-21
1464 Improve English Translations

### LINKED ISSUE #1464: Improve English Translations
Some English translations use poor grammar or improper words. This ticket is a placeholder so I can submit a pull request to improve the English translation.

jlefler:
This pull request has no functional changes. It merely updates the English translation with minor changes to word choice and grammar to improve the user experience for English-speaking users. The incorrect translation that prompted the change was 'Actualise' when performing a feed update. I have replaced that with "Updating" and made other similar changes. Fixes #1464.


app/i18n/en/conf.php
```diff
@@ -142,7 +142,7 @@
 		'_' => 'Shortcuts',
 		'article_action' => 'Article actions',
 		'auto_share' => 'Share',
-		'auto_share_help' => 'If there is only one sharing mode, it is used. Else modes are accessible by their number.',
+		'auto_share_help' => 'If there is only one sharing mode, it is used. Other modes are accessible by their number.',
```
Alwaysin: I think that "else" was used as in "otherwise" here - not "other", which would make more sense.

Alwaysin (changes requested):

jlefler:
I'm fine with otherwise rather than other here.

app/i18n/en/admin.php
```diff
@@ -138,7 +138,7 @@
 		'number_entries' => '%d articles',
 		'percent_of_total' => '%% of total',
 		'repartition' => 'Articles repartition',
-		'status_favorites' => 'Favourites',
+		'status_favorites' => 'Favorites',
```
aledeg: Both are correct. One is for US, the other one is for UK. I have no preference even though I am more used to "Favourites"

aledeg:
Look good to me.

@jlefler thanks

Alkarex:
Merged. Thanks again!

app/i18n/en/gen.php
```diff
@@ -2,7 +2,7 @@
 
 return array(
 	'action' => array(
-		'actualize' => 'Actualize',
+		'actualize' => 'Updating',
```
Alkarex, line 5: I think this one is wrong:
image
Should be something like *Actualize* or *Update*

Alkarex: Fixed in https://github.com/FreshRSS/FreshRSS/pull/1492

Frenzie:
I would also change the "damn!" in error messages to something slightly more proper even if it's only a stylistic issue.

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1561

PR #1528, opened 2017-05-09
Fix syslog_enabled = false not working

cez81:
`$syslog_enabled` will default to `SIMPLEPIE_SYSLOG` when initalizing a new `SimplePie_File`. To correct this I have added `$syslog_enabled` as an argument when creating a new file.


Alkarex:
Thanks! Please add yourself to https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

Issue #1531, opened 2017-05-11
config.default.php should be  along the code


oupala:
In my opinion, if `config.default.php` should be updated with FreshRSS update, it does not belong to user data, and hence **should not be present in `data` directory. It should be along the code**, somewhere else. In my case, the data directory is outside FreshRSS's directory (and symlinked) to that I can update FreshRSS source code without any risk of erasing my user data.

Alkarex:
Addressed in https://github.com/FreshRSS/FreshRSS/pull/1532

oupala:
Great !

You should also mention in the changelog or somewhere in the update instructions to remove the existing `config.default.php` file from the `USERS_PATH`.

Alkarex:
Moving the old files should be done automatically by `git pull` or the future ZIP update.

Alkarex:
There is a note in the changelog https://github.com/Alkarex/FreshRSS/commit/68d3deee1e1d5480bff1a61547188576e13ffcb4

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1519

Issue #907, opened 2015-07-19
Update process: include in FreshRSS code


Alkarex:
Update code should be shipped with new versions of FreshRSS, so that updating through FTP upload or git will work.
See https://github.com/FreshRSS/FreshRSS/issues/890#issuecomment-122309779

Alkarex:
A possibility to  avoid a dependency to the update server would be to query GitHub directly (simple `HEAD` request), which returns the latest release:

``` sh
curl -I https://github.com/FreshRSS/FreshRSS/releases/latest
HTTP/1.1 302 Found
Server: GitHub.com
Location: https://github.com/FreshRSS/FreshRSS/releases/tag/1.4.0
```

oupala:
While using the current *upgrade check* process, I'm still getting an error while I just want to know if there is a new version. I think it would be an improvement if it could be a *two-step* process.

1. first checking if there is a new version available (this step should **never** fail, except for a network problem where it display *can't check for a new release*)
2. then, try to update it (but **only** if the user requested to do so), this step could preventively check if the upgrade process is possible (due to permission issue) to determine if it should display the upgrade button or not

As a consequence, messages could be more explicit. Let me give you an example:

- first step:

> [Vérifier les mises à jour] button activated
> Une nouvelle version **1.7.1** est disponible.
> or
> Impossible de vérifier si une nouvelle version est disponible. Une erreur réseau est peut-être survenue.

- second step

> [Mettre à jour FreshRSS vers la version **1.7.1**]
> or
> Impossible de mettre à jour FreshRSS vers une nouvelle version car FreshRSS n'a pas les permissions en ériture sur son propre répertoire. Vous pouvez mettre à jour les permissions si vous voulez bénéficier d'une mise à jour automatisée.

This idea was first mentioned in #1450

Alkarex:
GitHub API: https://developer.github.com/v3/repos/releases/

```sh
curl -s 'https://api.github.com/repos/FreshRSS/FreshRSS/releases/latest' | jq .
```

```json
{
  "url": "https://api.github.com/repos/FreshRSS/FreshRSS/releases/7956249",
  "assets_url": "https://api.github.com/repos/FreshRSS/FreshRSS/releases/7956249/assets",
  "upload_url": "https://uploads.github.com/repos/FreshRSS/FreshRSS/releases/7956249/assets{?name,label}",
  "html_url": "https://github.com/FreshRSS/FreshRSS/releases/tag/1.8.0",
[...]

PR #1532, opened 2017-05-12
Move default configuration files

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1531


Alkarex:
@oupala Please give it a try if you can

Alkarex:
Some of the distributions might have to change a path if they do anything special with those default configuration files:

* Moved `./data/force-https.default.txt` to `./force-https.default.txt`,
 `./data/config.default.php` to `./config.default.php`,
 and `./data/users/_/config.default.php` to `./config-user.default.php`

Any comment from Cloudron (@dswd), DPlatform (@j8r), YunoHost (@plopoyop) ?

aledeg (approved):
LGTM

plopoyop:
Since Yunohost is only using the cli interface, it won't change anything for yunhost package.

dswd:
I just checked my Cloudron package code and it seems I am using the cli too.

j8r:
Thanks @Alkarex , a line need to be changed on DPlatform. These paths are clearer than the previous ones 👍

oupala:
@Alkarex I can test for sure. But what should I test? The install process, the usage after installation? The usage with my own data directory?

Alkarex:
@oupala If you just try the /dev branch for your own case, that will do it :-)

oupala:
I tested your fork (Alkarex/FreshRSS branch dev) and I end up with so many configuration files:

```
>find -L FreshRSS-dev/ -name "*config*"
FreshRSS-dev/config-user.default.php
FreshRSS-dev/config.default.php
FreshRSS-dev/data/users/_/config.default.php
FreshRSS-dev/data/users/<user>/config.php
FreshRSS-dev/data/config.php
FreshRSS-dev/data/config.default.php
```

Is it normal to have so many configuration files?

Even after a clean install (no user data directory), there is 4 configurations files!

```
>find -L FreshRSS-dev/ -name "*config*"
FreshRSS-dev/config-user.default.php
FreshRSS-dev/config.default.php
FreshRSS-dev/data/users/<user>/config.php
FreshRSS-dev/data/config.php
```

It becomes hard to say *where* is the configuration stored.

Alkarex:
Yes, that looks fine. There is a default set of settings for the system, and a default set of settings for a user. Those are not supposed to be edited by end-users. Then there is the settings of the concrete instance, and one configuration file per concrete user.

oupala:
Ok, but then there is 2 configurations files in `data` directory.

How do you ensure that these files stay up to date? Are they both containing just the modified settings from the default ones?

In my case, the data directory is never updated by an upgrade. Is there an automated process to upgrade them whenever ansible upgrades the FreshRSS instance? (such as the cookie setting)

Alkarex:
@oupala Sorry for the delay.
The files you are mentioning are user files. Yes, they override default settings. When a new property is introduced, it does not matter if it is not included in those files.
Very few changes are made during versions in the `./data/` directory, but a little bit, such as creating new sub-directories. Those things are done automatically by a ZIP update or git pull update. There is currently no mechanism to perform a partial update.
I have not yet seen an Ansible rule for FreshRSS, so I am not sure of the details.

oupala:
Ok @Alkarex thanks a lot.

In my opinion, the data directory is the user domain. There should not be any update in this directory made by the upgrade process, at least not without warning the user.

With my ansible role, the data directory is never updated by the upgrade process. So if any of your upgrade needs an update in the data directory, you should add a **BREAKING CHANGE** in the changelog so that the user knows that he has to change something in the data directory.

Alkarex:
`./data/` is not (only) user domain. It is the sub-directory where FreshRSS (via the Web server) must have write access, while read access is enough for other directories.
I think everything you ask should be done via the CLI. Not everything is available through the CLI yet, though, so please open issues for the missing features you need.
https://github.com/FreshRSS/FreshRSS/tree/dev/cli

oupala:
Thanks @Alkarex for pointing the difference between "user space" and "write access".

Can you help me to define where is the "user space" then?

My point is not about manually editing files. I do fully agree that all operation should be done via the CLI, which knows better that any human how to manage data.

My point is about separating the user data from the app data. When my ansible role installs a new release of FreshRSS, it installs it in a new directory (either by deleting the previous directory, or by creating a new directory). After that step, the ansible role creates a symlink to *insert* the user data so that the user can find all his settings, rss history, favorites articles, that he had on the old instance. This is only possible if user data are clearly separated from the app data, and if any breaking changed in user data are clearly documented (in the changelog).

Alkarex:
From version 1.7.0, copying the content of your old `./data/` directory into the new `./data/` directory should be enough, and will also cover the case of new sub-directories created under `./data/`.
Related to https://github.com/FreshRSS/FreshRSS/issues/907

oupala:
Good point.

Will the following work?

- delete old FreshRSS directory
- install new version of FreshRSS
- delete new `./data` directory
- symlink `/somewhere-on-a-nas/freshrssdata/` to `./data`

This way, it won't merge the two `data` directory, and new sub-directories won't exist after upgrade...

Could it be FreshRSS's responsibility to update a data directory when needed?

PR #1032, opened 2015-11-20
Add more glyphs for opensans font

asmfreak:
This pull request is a continuation of work for #1028, but in a cleaner way. I deleted my fork and reforeked the repository. This pr fixes #1028 .

PR #1541, opened 2017-05-23
simplified chinese translation

hoilc:
The PR adds simplified chinese translation for FreshRSS.

It's translated from English and some words could be inaccurate.

I will improve it later.


Alkarex:
Hello @hoilc 
Many thanks for this first translation.
Do you need to change the theme font to include Chinese characters?

hoilc:
Hi @Alkarex

Thanks for reminding me.

I added the default simplified chinese font of Windows and MacOS into templete font-family.

Alkarex:
For the font, I was also thinking of adding glyphs to our OpenSans font (see e.g. https://github.com/FreshRSS/FreshRSS/pull/1032 ). I need to look at it also (I would like to avoid linking to yet another CSS, and also ensure that the font does not become too large - maybe with an option)

Alkarex:
Please add yourself to https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md

Alkarex:
Merged in /dev. Thanks again!

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/5930

Issue #1548, opened 2017-05-26
Call for testing /dev before version 1.7


Alkarex:
The future version 1.7 of FreshRSS is [ready](https://github.com/FreshRSS/FreshRSS/milestone/20) for testing. This time, the main theme has been to introduce a more more natural order of entries when refreshing feeds, but there is also a number of new features and bug corrections, as visible [in the changelog](https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md):

* Features:
	* Deferred insertion of new articles, for better chronological order [#530](https://github.com/FreshRSS/FreshRSS/issues/530)
	* Better search:
		* Possibility to use multiple `intitle:`, `inurl:`, `author:` [#1478](https://github.com/FreshRSS/FreshRSS/pull/1478)
		* Negative searches with `!` or `-` [#1381](https://github.com/FreshRSS/FreshRSS/issues/1381)
			* Examples: `!intitle:unwanted`, `-intitle:unwanted`, `-inurl:unwanted`, `-author:unwanted`, `-#unwanted`, `-unwanted`
		* Allow double-quotes, such as `author:"some name"`, in addition to single-quotes such as `author:'some name'` [#1478](https://github.com/FreshRSS/FreshRSS/pull/1478)
	* Multi-user tokens (to access RSS outputs of any user) [#1390](https://github.com/FreshRSS/FreshRSS/issues/1390)
* Compatibility:
	* Add support for PHP 7.1 [#1471](https://github.com/FreshRSS/FreshRSS/issues/1471)
	* PostgreSQL is not experimental anymore [#1476](https://github.com/FreshRSS/FreshRSS/pull/1476)
* Bug fixing
	* Fix PubSubHubbub bugs when deleting users, and improved behaviour when removing feeds [#1495](https://github.com/FreshRSS/FreshRSS/pull/1495)
	* Fix SQL uniqueness bug with PostgreSQL [#1476](https://github.com/FreshRSS/FreshRSS/pull/1476)
		* (Require manual update for existing installations)
	* Do not require PHP extension `fileinfo` for favicons [#1461](https://github.com/FreshRSS/FreshRSS/issues/1461)
	* Fix UI lowest subscription popup hidden [#1479](https://github.com/FreshRSS/FreshRSS/issues/1479)
	* Fix update system via ZIP archive [#1498](https://github.com/FreshRSS/FreshRSS/pull/1498)
[...]

PR #1550, opened 2017-05-26
fix font-sizes of input fields for exotic browsers

mszkb:
There was no Placeholder inside input fields. Also the written text inside password-fields and adding subscription were super small (font-size: 0; inherited from .stick origine.css:115)

See my comment below for screenshots.

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -88,7 +88,7 @@ input.extend:focus {
 input, select, textarea {
 	display: inline-block;
 	max-width: 100%;
-	font-size: inital;
+	font-size: initial;
 }
 input[type="radio"],
 input[type="checkbox"] {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -88,6 +88,7 @@ input.extend:focus {
 input, select, textarea {
 	display: inline-block;
 	max-width: 100%;
+	font-size: inital;
 }
 input[type="radio"],
 input[type="checkbox"] {
@@ -345,7 +346,7 @@ a.btn {
 /*=== Tree */
 .tree {
 	margin: 0;
-	padding: 0 0 15em 0;
+	padding: 0;
 	list-style: none;
 	text-align: left;
 }
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -88,7 +88,7 @@ input.extend:focus {
 input, select, textarea {
 	display: inline-block;
 	max-width: 100%;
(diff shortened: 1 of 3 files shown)


p/themes/base-theme/template.css
```diff
@@ -88,6 +88,7 @@ input.extend:focus {
 input, select, textarea {
 	display: inline-block;
 	max-width: 100%;
+	font-size: inital;
```
Alkarex: There is a typo: it is `initial` ;-)

Alkarex:
Please add yourself to https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md

mszkb:
Oh yes, stupid me. Fixed in 20281fb

Alkarex:
Actually, the `font-size: initial;` makes input fields quite larger. It would be nice to find a fix that corrects the problem in QupZilla without changing too much the rendering in other browsers:

Before (in Firefox):

image

After:

image

mszkb:
Hmmm, to fix this issue on all browsers, it needs a reset.
If changing the font-size to 15px, the fonts on Chrome gets bigger (13.3333px default on chrome).

Firefox and Chrome ignore the font-size from **.post origine.css:635** and **.stick origine.css:115**. They take the fallback to the user agents default. QubZilla inherites font-size: 0 from **.stick**. 
Not sure if this comes from the special implementation of QubZillas rendering engine or the correct interpretation.

Anyway.. It would be your call to risk a reset of the font-size or ignore it, due to low user count of QubZilla.

p/themes/base-theme/template.css
```diff
@@ -345,7 +346,7 @@ a.btn {
 /*=== Tree */
 .tree {
 	margin: 0;
-	padding: 0 0 15em 0;
+	padding: 0;
```
Alkarex: What was the reason for this change?
https://github.com/FreshRSS/FreshRSS/pull/1489

p/themes/base-theme/template.css
```diff
@@ -88,6 +88,7 @@ input.extend:focus {
 input, select, textarea {
 	display: inline-block;
 	max-width: 100%;
+	font-size: initial;
```
Alkarex: I suggest trying with `font-size: 0.8rem;` instead.

Alkarex:
@mszkb Trying with `font-size: 0.8rem;`. Let me know how it works on your side.
Quick merge to test the /dev branch prior to 1.7 release.
https://github.com/FreshRSS/FreshRSS/issues/1548

Issue #1545, opened 2017-05-25
Feature Request: "Add to favourites" on Reading View


Aasemoon:
Hi there,

I have a request for a tiny adjustment. Could you possibly make it so that the "add to favourites" star button shows up for each item on the reading view as well? Would be very helpful I believe.

Cheers!

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/1400

Alkarex:
Thanks @Aasemoon 
Planned for 1.8.0.

Aasemoon:
@Alkarex Great! Thanks! =)

aledeg:
@Aasemoon Could you check the PR I've just made and tell me if it suits your needs?

Aasemoon:
Hi @aledeg, hmmm I'm not sure exactly what I'm missing, but when I apply the suggested changes to my main.js, I don't see anything different in the reading view. Is that the only file you have changed?

aledeg:
There is nothing different in the view. But you can use the shortcut available in the normal view. I was wondering if that change is enough for you or if we need to go a bit further. Thank you

Aasemoon:
@aledeg , aha, then I guess not. I'd like an "add to favourite" button on the reading view mainly because usually I view my feeds on various tablets. Keyboard shortcuts I've basically never used.

aledeg:
Thank you for your feedback. I'll try to add something in the view. I am not very comfortable with front-end but I'll give it a try.

Aasemoon:
@aledeg Thanks a lot! That would be great. =)

aledeg:
@Aasemoon The PR is merged. Does it covers what you had in mind?
If yes, I suggest we close that issue.

PR #1555, opened 2017-05-27
Added a toggle to quickly change from/to collapsed articles

romibi:
[https://gyazo.com/38564bc66e2256deba3b893d24ffa086](https://gyazo.com/38564bc66e2256deba3b893d24ffa086)

* I hope the Icons are ok...
* Translations need updating...
* Toggle does **not** keep its state through reload
* slight duplicated code for the lazy-loading part

### DIFF
--- a/app/Models/Themes.php
+++ b/app/Models/Themes.php
@@ -104,6 +104,8 @@ class FreshRSS_Themes extends Minz_Model {
 			'view-normal' => '☰',
 			'view-global' => '☷',
 			'view-reader' => '☕',
+			'out' => '↕',
+			'in' => '✖',
 		);
 		return isset($name) ? $alts[$name] : '';
 	}
--- a/app/i18n/cz/index.php
+++ b/app/i18n/cz/index.php
@@ -54,6 +54,7 @@ return array(
 		'stats' => 'Statistika',
 		'subscription' => 'Správa subskripcí',
 		'tags' => 'My labels',	// TODO - Translation
+		'toggle_collapse' => 'Toggle collapse',
 		'unread' => 'Zobrazovat nepřečtené',
 	),
 	'share' => 'Sdílet',
--- a/app/i18n/de/index.php
+++ b/app/i18n/de/index.php
@@ -54,6 +54,7 @@ return array(
 		'stats' => 'Statistiken',
 		'subscription' => 'Abonnementverwaltung',
 		'tags' => 'Meine Tags',
+		'toggle_collapse' => 'Artikel auf/zu-klappen',
 		'unread' => 'Nur ungelesene zeigen',
 	),
 	'share' => 'Teilen',
--- a/app/i18n/en/index.php
+++ b/app/i18n/en/index.php
@@ -54,6 +54,7 @@ return array(
 		'stats' => 'Statistics',
 		'subscription' => 'Subscriptions management',
 		'tags' => 'My labels',
+		'toggle_collapse' => 'Toggle collapse',
 		'unread' => 'Show unread',
 	),
 	'share' => 'Share',
--- a/app/i18n/fr/index.php
+++ b/app/i18n/fr/index.php
@@ -54,6 +54,7 @@ return array(
 		'stats' => 'Statistiques',
 		'subscription' => 'Gestion des abonnements',
 		'tags' => 'Mes étiquettes',
+		'toggle_collapse' => 'Toggle collapse',
 		'unread' => 'Afficher les non-lus',
 	),
 	'share' => 'Partager',
--- a/app/i18n/it/index.php
+++ b/app/i18n/it/index.php
@@ -54,6 +54,7 @@ return array(
 		'stats' => 'Statistiche',
 		'subscription' => 'Gestione sottoscrizioni',
 		'tags' => 'My labels',	// TODO - Translation
+		'toggle_collapse' => 'Toggle collapse',
 		'unread' => 'Mostra solo non letti',
 	),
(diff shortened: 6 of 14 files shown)


Alkarex:
Thanks @romibi for this pull request :-) I will have a look after the release of version 1.7.

Alkarex:
Maybe we should try to merge this feature with an improved "read view". Any comments?
Indeed, the read view should also get icons to mark as unread, favourite, etc. and it will get more similar to this view when articles are expanded.

For reference:
* https://github.com/FreshRSS/FreshRSS/issues/603
* https://github.com/FreshRSS/FreshRSS/issues/1400
* https://github.com/FreshRSS/FreshRSS/issues/1545

romibi:
Hm… my french isn't too good …
But I don't use the read view because i like having the feed lists on the left …

But yes … I ignored the other views completely while making this change …
This button should only be visible in the normal view.
(And perhaps in read view **if** it works there with a new version of that view.)

marienfressinaud:
Hi! What's the status of this PR? I'd like to close some and this one is pretty old…

I think this button meets a too specific need to be included in the core of FreshRSS. I rather would see it in an extension. Looking at the code, I would say there's only one missing hook to add a new button to the menu.

romibi:
It's no problem for me if you take your time to decide. If you don't merge I'll just rebase the branch everytime I update my FreshRSS installation. (yes I think I went from 0.12 directly to 0.15)

For me this is as basic a functionality as the reorder button. Which is why I put it next to it.

### My Usecase (and why I think this is basic functionality)
I have "Show articles unfolded by default" enabled (which I think is disabled by default).
Because I don't like it when the page content changes (collapse/uncollapse) while switching beetween articles. (And normally I skim over every article)

But if I have too many unread articles in a category I'll click this button and only skim over the titles and then click "Mark everything as read".

Basically this is just a button to quickly and only temporarily change the "Show articles unfolded by default" option …

### My thoughts why this might be unwanted
Hm, maybe users who don't have "Show articles unfolded by default" enabled are not interested in this button. We could hide it if that option is not enabled. (Although I don't see why it would matter.)
But if that option is enabled I really think this is a really basic and useful functionality.

### Why I think this was not merged yet.
As far as I understood it, this isn't merged yet because it doesn't work in the "Global"- and the "Read"-View even though the button is shown … but the reorder button is show in "Global"-View and does nothing there, too…
Later it might have been just overlooked and/or was ignored because it had merge conflicts for some time…

To fix the "Button does not work"-Issue we could quite easily add the following css:
`.reader #toggle-collapse, .global #toggle-collapse {
	display: none;
}`
(Although `.stick .btn:first-child` and `.stick .btn:last-child` does not seem to work with nav-menu groups containing hidden buttons and/or only one button)

marienfressinaud:
> If you don't merge I'll just rebase the branch everytime I update my FreshRSS installation.

Yes but it's not really fun for you :(

> My Usecase (and why I think this is basic functionality)
> […]

I understand your use case, but most of it is the reason why the "Show articles unfolded by default" is disabled by default. Though I also understand that you don't like that the content changes while you're navigating between articles.

The main reason for me to not merge this code is that the menu bar has already too many buttons. While I did user tests, I observed some cognitive issues to discover and understand the purpose of the buttons. Unless necessary, adding a new button here would not help. It's also more code for developers to maintain "officially" while not being used by them.

Yet, I think it's not a big deal to move your code in an extension (I can do it myself if you prefer, maybe next week) and let's reconsider having it in the core of FreshRSS if it becomes popular, or if one of the maintainer starts to use it and aims to maintain the code :)

romibi:
> Yet, I think it's not a big deal to move your code in an extension (I can do it myself if you prefer, maybe next week) and let's reconsider having it in the core of FreshRSS if it becomes popular, or if one of the maintainer starts to use it and aims to maintain the code :)

Yes, I would be glad if you could do that. You are most likely way faster than I would be to achieve this.

Although I don't share the opinion about "too many buttons".
There are only:
* the Filters
* the "Mark read"-option
* the View-Selector
* sort
* (collapse) 
* refresh

Similar to comparable menubars in tools like Thunderbird which have even more additional menubars.
There's like only a third of the space available in the menubar used on most of my screens. (And I'm not even using HiDPi screens)

But if you want it as an extension, so be it.

Frenzie:
@romibi 
> There's like only a third of the space available in the menubar used on most of my screens. (And I'm not even using HiDPi screens)

The window width is besides the point @marienfressinaud is making. It's about difficulty understanding the purpose of buttons. One might counter that obvious at first glance doesn't necessarily make for good UI/UX depending on context. As you said, FreshRSS is probably more like Thunderbird and less like some random website that should be as clear as possible at first glance.

Incidentally, I didn't customize Thunderbird to display the labels. ;-)

Screenshot_2020-01-18_12-16-15

marienfressinaud:
Here's the code of the extension: https://github.com/FreshRSS/Extensions/pull/71

aledeg:
Should we close this as @marienfressinaud created an extension that addresses the original problem?

romibi:
I'm still in the optinion that this would be nice directly in FreshRSS … therefore i'll not be the one to close it.
But I used the extension now for some time and that works fine too.
So its ok for me if you close it …

math-GH:
> Hm… my french isn't too good …
> But I don't use the read view because i like having the feed lists on the left …

The feed lists on the left are available in the "reader view" (by default the list is not displayed, via button you can show the list on the left side)

For me it would be fine to close this PR, because there is now a extension and the feed list is available in the reader view.

Issue #1556, opened 2017-05-27
CSS font-size mark-as-read


Alkarex:
@Frenzie https://github.com/FreshRSS/FreshRSS/issues/1548#issuecomment-304445616

> Is "Mark as read" supposed to look extra large compared to the other options?

![screenshot_2017-05-27_13-02-35](https://cloud.githubusercontent.com/assets/202757/26521007/46aeeb06-42de-11e7-88e0-93d6ccf8892e.png)

Alkarex:
Indeed, there is a problem in some themes. Pafat seems to work fine already.

Alkarex:
Candidate fix in https://github.com/FreshRSS/FreshRSS/pull/1557

PR #1557, opened 2017-05-27
More uniform styling of A and BUTTON in drop-downs

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1556
https://github.com/FreshRSS/FreshRSS/issues/1548#issuecomment-304445616

### DIFF
--- a/p/themes/BlueLagoon/BlueLagoon.css
+++ b/p/themes/BlueLagoon/BlueLagoon.css
@@ -368,16 +368,13 @@ a.btn {
 .dropdown-header {
 	display:none;
 }
-.dropdown-menu > .item > a {
-	padding: 0 25px;
-	line-height: 2.5em;
-	color: #ccc;
-}
+.dropdown-menu > .item > a,
 .dropdown-menu > .item > span,
 .dropdown-menu > .item > .as-link {
 	padding: 0 22px;
-	line-height: 2em;
+	line-height: 2.5em;
 	color: #ccc;
+	font-size: 0.8rem;
 }
 .dropdown-menu > .item:hover {
 	background: linear-gradient(180deg,  #0090FF 0%, #0062BE 100%) #E4992C;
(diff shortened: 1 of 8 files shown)


Alkarex:
@Frenzie Please give it a try if you can

Frenzie:
Not tonight; probably not till Tuesday.

Frenzie:
The font-size seems okay, but the alignment is off a fair bit in Fx. To get it to render the same as in other browsers it'd need something like this:

```css
button::-moz-focus-inner {
    padding: 0;
    border: 0
}
```

I imagine this is part of the reason why so many fake link buttons exist.

Alkarex:
@Frenzie I cannot immediately see the problem on my side. Could you please send a screenshot?

Frenzie:
screenshot_2017-05-28_11-42-37

I can't quite get at the bottom of it because the Fx web inspector seems to be broken for me and it's not showing me CSS rules properly (no, I'm not filtering anything).

Also see https://stackoverflow.com/questions/5517744/remove-extra-button-spacing-padding-in-firefox

Alkarex:
With Firefox 53 on Windows 10:
image

With Firefox 52 on Ubuntu 17.04:
image

With Firefox 53 on Ubuntu 17.04:
image

Alkarex:
I was able to see the problem with Firefox 52 on Ubuntu, but with the current version (53) of Firefox, I cannot reproduce the problem on Windows or Ubuntu.
Related Firefox issues:
* https://bugzilla.mozilla.org/show_bug.cgi?id=140562
* https://bugzilla.mozilla.org/show_bug.cgi?id=435697
Seems to be fixed from Firefox 53.

Frenzie:
The current Firefox ESR is 45 but since it's a minor visual glitch let's just ignore it then.

Edit: incidentally, on my Debian Stretch Fx is still at 52. I mistakenly thought it was the latest.

PR #1559, opened 2017-05-28
(chore) Some random Dutch i18n improvements

Frenzie:


Alkarex:
Thanks. Maybe you could translate this new string as well?
https://github.com/Frenzie/FreshRSS/blob/f4bd7431dfbb87f906f95141fe3201eb9677f3a9/app/i18n/nl/conf.php#L95

Frenzie:
Added the new string and a bunch of other random stuff.

PR #1561, opened 2017-05-28
(minor) Eng i18n: "Damn" to "blast"

Frenzie:
I think French *arf*/*zut* would be better translated as something like *blast* or *rats* than *damn*. The word *arf* doesn't have a lemma in my *Petit Robert*, so since French is my fourth and worst language I can't be completely sure. But even if I'm wrong about the French, the word *damn* simply seems slightly too expletive for an English localization. By contrast, in Dutch I'm not sure if anything even exceeds the level of offensiveness reached by a mild expletive like *damn* in English. English speakers are just much more sensitive to words in and of themselves regardless of context.


marienfressinaud (approved):
You're English is probably better than mine, go for it! 👍

PR #1388, opened 2016-12-03
Added Minimal theme

kevinpapst:
Hi guys,
I am not sure if you want to add more themes... but I'll give it a try ;-)
When I encountered FreshRSS I really enjoyed the software package - fantastic work!
The default theme - in my eyes - is beautiful but at the same time is wasting a lot space in the viewport with paddings, margins, line-height and to large font-sizes... so I created a fork from Origine and adjusted some sizes to fit more information onto the screen.
In the first place I opened this PR to get some feedback and input on what could be further improved and also contribute a bit back to the community.
Any thoughts?

PR #1580, opened 2017-06-10
Show full title of article when opened

sjeon87:
I am subscribing some conference and journal feeds. Oftentimes, titles are a bit long. FreshRSS cuts the remained part away. Especially, it's hard to use FreshRSS on mobile phones. Added an option for it.

It looks like

- on PC

- on Mobile

screenshot_20170610-220718

Cons: Looks thick.


sjeon87:
TODO

1. i18n for this feature is not complete. All i18n are written in English
2. I've made 2 PR to the current `dev` branch. If both are merged, `i18n/kr` for this feature are needed.

Alwaysin:
I've always wanted this 👍 
Especially useful with the minimal theme !

app/i18n/cz/conf.php
```diff
@@ -91,6 +91,7 @@
 		'confirm_enabled' => 'Vyžadovat potvrzení pro akci “označit vše jako přečtené”',
 		'display_articles_unfolded' => 'Ve výchozím stavu zobrazovat články otevřené',
 		'display_categories_unfolded' => 'Ve výchozím stavu zobrazovat kategorie zavřené',
+		'full_title' => 'Show the full title of the article when opened <small>(Title bar might look thicker)</small>',
```
Alkarex: Could you please add a ` //TODO` at the end of this line and all others that requite a translation?

Alkarex:
Ah, I should have noticed before: this is only relevant for the "origin-compact" theme, isn't it? Other themes already show the full title in the body of the article. E.g.
image
I would then suggest to change the origin-compact theme with your patch
https://github.com/FreshRSS/FreshRSS/pull/1388

sjeon87:
Then, it's somewhat complex. My PR still works with other themes, but showing full title in other themes looks redundant.

So it's a theme-dependent feature. I'll consider other ways to this end. (I think opt-in/out are needed because some people may like compact theme without full title)

Alkarex:
Without having checked the details, I would suggest you merge your CSS changes https://github.com/FreshRSS/FreshRSS/pull/1580/files#diff-ba1139aea9808c596c3bb85ce1b1233a with the `origin-compact` theme.
P.S.: Meaning: always show the full title in this theme, without an option

Alkarex:
Maybe @kevinpapst has some comments?

sjeon87:
If no opt-in/out are needed (but I think it should be careful, because it's compact theme!), I'll work on it. But if not, It rather not be merged.

kevinpapst:
When I created the compact theme I removed the title on purpose from the body as it consumes rather much space and is (at least for me) in more than 90% of all feeds either not important or completely visible in the list. I would rather not re-add it. 

But we could probably have it in the HTML and use display:none as default. 
Interested users could then still use custom CSS rules to display it.

Or if it should be more user friendly: add a opt-in setting which works in all themes and toggles the title?

Alkarex:
@kevinpapst In this case, the title in the body would still be hidden, but the full title would be shown in the title bar even in the case of titles longer than one line

Frenzie:
The compact theme never displays the full title at all? Seems like a slightly odd decision to me. :-) I'm actually much more interested in having this prior to opening in any theme. (Yeah, yeah, I'm sure I can easily figure out the CSS but I'm just throwing it out there since someone's already on it.) Just compare EasyRSS and FreshRSS on my phone. The full title is only somewhat redundant *after* you've already read it.

It's the only reason I use EasyRSS on my phone really, except for when I want to take along some articles offline (which only seems to work out half the time anyway).

kevinpapst:
Last time I read this thread I might have misunderstood the PR.

I don't think that we should keep the discussion focused on the compact theme:
- this PR affects all themes by bringing a new feature
- if you want to see the full title before opening the article, then its not a big help to have the full title in the body => thats why I removed it from the body (redundancy)
- the mobile view is probably not the best UX in any theme and this might be a small step for improvement => we should have a different discussion how to improve the UX in mobile view in general

Now what?
1. Keep this PR as it is, add it as a new and theme independent feature
2. Having this PR for compact only (@Alkarex how? this PR changes core features), but not without a toggle to turn it on and off

My preference:
I am not convinced the solution from this PR is a good fit for the current UX/UI concept but I totally understand that people want to see the full title - so give them the option to do so ... for any theme.

Frenzie:
On desktop I'd also prefer to see the full title but it's even less important because

a. Most of the title is generally showing regardless.
b. It's easy enough to hover, which is not the case on mobile.

But I don't necessarily mean the full title, just more of it. Exceeding five lines or so would become a bit much. Like how on good old Opera/Presto you had this experimental property:
 
```css
text-overflow: -o-ellipsis-lastline;
```

Well, perhaps my interest is sufficiently piqued to experiment a little a few days from now. We don't have the CustomCSS extension for nothing. ;-)

Alkarex:
@Frenzie Please let us know the result of your experiments :-)

Frenzie:
None yet. I might've been too optimistic wrt my time.

Frenzie:
@Alkarex The README mentions "A recent browser like Firefox, Internet Explorer 11 / Edge, Chrome, Opera, Safari." but does that mean any concrete minimum? For example, flexbox could be interesting here besides regular CSS tabular stuff as already used.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes

Alkarex:
@Frenzie Good idea to use flexbox. I suggest for anybody interested in using more recent CSS features to just start a new theme. This way we can keep some legacy themes for some time, while having better layouts in new themes. We can write in the theme's description any higher requirement.

Alkarex:
https://github.com/FreshRSS/Extensions/issues/46

marienfressinaud:
I'm closing this PR because there's a bunch of conflicts and the approach needs to be improved. Also there's alternative thanks to the CustomCSS extension (see https://github.com/FreshRSS/FreshRSS/issues/2344 thread)

Issue #1479, opened 2017-04-03
UI: lowest subscription's popup gets hidden


psukys:
Steps to reproduce:

1. Have a big (filling whole screen vertically) category of subscribed feeds;
2. Click on wrench (settings) of the lowest subscription;
3. Verify that the popup shows ups, part of it hidden below
4. Verify that attempts to scroll down (specifically using mouse scrolling wheel, keyboard arrows and _Page down_) does not scroll down.

Checked on

-  Chrome: Version 57.0.2987.133 (64-bit)

![screenshot from 2017-04-03 21-11-25](https://cloud.githubusercontent.com/assets/1302529/24626719/806a5dfe-18b2-11e7-92ee-f179a4e2c919.png)

Alkarex:
Yes indeed. I was waiting for someone to complain :-P
A quick fix would probably to give a bit more margin at the bottom of the column of categories.
In the meantime, scrolling down *after* you have opened the menu should work :-)

psukys:
@Alkarex I've downloaded and checked on Firefox and was able to scroll down _after_ opening menu, but in Chrome it does not :/

Alkarex:
Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1489

Alkarex:
Please test https://github.com/FreshRSS/FreshRSS/pull/1489 or even better, the [/dev version of FreshRSS](https://github.com/FreshRSS/FreshRSS/tree/dev).

PR #1589, opened 2017-06-21
Make feedlist padding a bit dynamic

romibi:
With the fix #1489 (Bottom padding menu) for issue #1479 (UI: lowest subscription's popup gets hidden) now I had always a huge unnecessary space below the feed list while scrolling up… (about 1/3rd of the screen)

With this change I reduced the padding a lot (but not completely) and made sure that the padding is big enough again when a dropdown-menu is open to enable scrolling it into the field of view…

I think this is a good compromise…


Alkarex:
Could you please make a screenshot of the original problem?

romibi:
How it looks without this PR for me:
[https://gyazo.com/b7f0e7b5fec9afb9273a7f3e65a435af](https://gyazo.com/b7f0e7b5fec9afb9273a7f3e65a435af)

Alkarex:
Ah, I see. Thanks 👍

Issue #1590, opened 2017-07-07
Closed (folded) articles are drawn open (unfolded).


sjeon87:
An example feed is http://export.arxiv.org/rss/cs. 

The problem is that body parts of multiple closed (unfolded) articles still appear and I cannot close such articles. Strangely, this feed passes RSS validation: https://validator.w3.org/feed/check.cgi?url=http%3A%2F%2Fexport.arxiv.org%2Frss%2Fcs

You can check the problem via https://songsong.postech.ac.kr/FreshRSS/i/?get=f_33 without login (or via your own FreshRSS.)

aledeg:
There is probably something in the body that prevent such action. I've experienced such things with other feeds. It was a little bit different though since after a type of article all the other disappear. We should probably add something to render the article content in a safe way.

Good catch

sjeon87:
Yes, I suspect missing tag escape or something (maybe not.) If time allows, I'll look into the issue.

An attachment is the XML feed backup causing the issue.
[cs.zip](https://github.com/FreshRSS/FreshRSS/files/1130143/cs.zip)

sjeon87:
`cs.xml` in the above, I found:

    <dc:creator> &lt;a href=&quot;http://arxiv.org/find/cs/1/au:+Gorbachev_V/0/1/0/all/0/1&quot;&gt;V.N. Gorbachev&lt;/a&gt;, &lt;a href=&quot;http://arxiv.org/find/cs/1/au:+Denisov_L/0/1/0/all/0/1&quot;&gt;L.A. Denisov&lt;/a&gt;, &lt;a href=&quot;http://arxiv.org/find/cs/1/au:+Kaynarova_E/0/1/0/all/0/1&quot;&gt;E.M. Kaynarova&lt;/a&gt;, &lt;a href=&quot;http://arxiv.org/find/cs/1/au:+Metelev_I/0/1/0/all/0/1&quot;&gt;I.K. Metelev&lt;/a&gt;</dc:creator>

When I look into the corresponding article with `RSS feed mode` of FreshRSS:

    <dc:creator>
[...]

PR #1592, opened 2017-07-11
Validate creator field

sjeon87:
Realted Issue: https://github.com/FreshRSS/FreshRSS/issues/1590#issuecomment-313914174

## Summary

Some feeds (e.g., arXiv) include hyperlinks in author fields (`<dc:creator>`). FreshRSS has a limited author field length of 255 bytes. This may make author information to have an invalid HTML format.

## Changes

Before displaying an entry, validate author information, if not, replace it with some warning e.g., *Cannot parse author information properly*

### Before

Other entries are hidden due to malformed author information of 1st entry

When 1st entry is opened, the others appear, but seem broken

## Consideration

Changes are made in `Models/Entry.php` and `i18n/*/gen.php`. But I am not sure other users and developers agree with my code style, error handling policy, or something.

## Future work

Validate author information before pushing it to database.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -80,7 +80,7 @@ class FreshRSS_Entry extends Minz_Model {
 		$tags_close = $this->_classify_tag($string, '/(?<tag>\<\/(?<keyword>\w)+.*?\>)/', 'close');
 		$tags_self = $this->_classify_tag($string, '/(?<tag>\<(?<keyword>\w)+.*?\/\>)/', 'selfclosing');
 		$tags = array_merge($tags_open, $tags_close, $tags_self);
-		usort($tags, '_sort_by_pos');
+		usort($tags, array('FreshRss_Entry', '_sort_by_pos'));
 		$stack = array();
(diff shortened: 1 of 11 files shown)


Alkarex:
Hello @gsongsong 
Many thanks for the bug report and the pull request.
I do not think we should try to save some HTML in a field as short as 255 characters. For the time being, I would rather like to parse the full HTML of the author information, and then keep only the text, and then truncate that text if it is too long. This should also make the code much shorter and simpler.

If it is OK with you, I would appreciate if you could update this pull request to do that. Please let me know :-)
(Using something like DOMDocument and node->nodeValue)

I understand that we will loose the author hyperlinks but from the sample of feeds I have, it does not seem to be a common practice. If we want to keep the full author HTML, we could either change the database, or add the information to the body of the article.

sjeon87:
Ok, but I can't make sure when it will be done. Maybe I can start working on it after August.

Alkarex:
Simple fix in https://github.com/FreshRSS/FreshRSS/pull/1615
Feedback appreciated

PR #1594, opened 2017-07-12
Fix Illegal String Offset warning (PHP 7.1) in greader.php

arokettu:
greader.php throws Illegal String Offset warning because $ALL_HEADERS should be array not string


Alkarex:
Hello @sandfoxme 
Many thanks for spotting and correcting this error.
Could you please use the classic syntax `array()` ? We are still supporting PHP 5.3.3.
I have changed the branch in the pull request (please use /dev and not /master)

arokettu:
@Alkarex done

Issue #1560, opened 2017-05-28
Language nitpick: uppercase May in statistics


Frenzie:
![screenshot_2017-05-28_12-31-53](https://cloud.githubusercontent.com/assets/202757/26528048/df2cc61e-43a1-11e7-9eca-69c4ab735210.png)

It's debatable which of these is actually wrong in English (as the default is a capitalized first letter), but in any case it's inconsistent.

A quick peek suggests that months all exist in two versions, both abbreviated in spite of what the index suggests:

```php
		'mar' => 'mar',
		'march' => 'Mar',
```

However, because May is already three letters it only exists in an uppercase capacity:
https://github.com/FreshRSS/FreshRSS/blob/be0bcfef7e38f27284ec7b377b342ba389515964/app/i18n/en/gen.php#L82

Alkarex:
I would go for an uppercase everywhere

aledeg:
If we check in the [code](https://github.com/FreshRSS/FreshRSS/blob/master/app/i18n/en/gen.php#L45-L56), we already use the *ucfirst* string but I don't know why.
If it's useless, it's a matter of reusing the strings.

Alkarex:
Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1617

PR #1617, opened 2017-08-20
May abbreviation

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1560


Alkarex:
@Frenzie

app/i18n/en/gen.php
```diff
@@ -79,6 +79,7 @@
 		'last_year' => 'Last year',
 		'mar' => 'mar',
 		'march' => 'Mar',
+		'may_' => 'May',
```
Frenzie: Note that for the English localization this May should be lowercase for consistency with the rest, although I think it'd be better to change the others to `'mar' => 'Mar',` etc.

app/i18n/nl/gen.php
```diff
@@ -79,6 +79,7 @@
 		'last_year' => 'Vorig jaar',
 		'mar' => 'mar',
```
Frenzie, line 80: I can fix this myself in a follow-up PR, but the Dutch abbreviation for `maart` is `mrt.` or `mrt`, not `mar`. Feel free to update it yourself. :-)

app/i18n/ru/gen.php
```diff
@@ -79,6 +79,7 @@
 		'last_year' => 'Last year',
 		'mar' => 'mar',
 		'march' => 'Mar',
+		'may_' => 'May',
```
Frenzie, line 82: Ah yes, I can tell this is Russian. xD

app/i18n/en/gen.php
Alkarex: Thanks, should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1617/commits/b80c2ee014827b2fcaa72676263359115a55230a

app/i18n/nl/gen.php
Alkarex: 👍

app/i18n/ru/gen.php
Alkarex: Yes, the Russian translation is not complete :-)

Issue #1604, opened 2017-07-26
iframe sanitizing bug with PHP 5.3.3


giorgiobeggiora:
Is there a way to always show controls?
Sometime they are not visible, i think when there are just few news.
Thanks

Alkarex:
Hello @giorgiobeggiora 
Could you please upload a screenshot illustrating the problem?

giorgiobeggiora:
Here they are. Note: on the left of the 4th image the sidebar is repeated 2 times: that's an error of my screenshot software. However, the presence or absence of the controls reflect the reality in all of the screenshots.  Thanks.

![6017 daily freshrss](https://user-images.githubusercontent.com/9007227/28649516-c0f22a92-7275-11e7-9260-b2becfade9d6.jpg)
![39964 daily geek freshrss](https://user-images.githubusercontent.com/9007227/28649517-c11a1278-7275-11e7-84a7-79b044d0a018.jpg)
![78379 your rss feeds freshrss](https://user-images.githubusercontent.com/9007227/28649518-c148ed00-7275-11e7-8cab-c2242f7fddd0.jpg)
![78 378 your rss feeds freshrss](https://user-images.githubusercontent.com/9007227/28649520-c1834cd4-7275-11e7-832c-172d9f4d8fa2.jpg)
![526 dev game freshrss](https://user-images.githubusercontent.com/9007227/28649521-c188e928-7275-11e7-80e3-c28f92461087.jpg)
![1258 gdr freshrss](https://user-images.githubusercontent.com/9007227/28649522-c199aad8-7275-11e7-8478-a237e1b88c4c.jpg)

giorgiobeggiora:
the cause may be this javascript error (seems that it blocks the loading of many posts too)

<img width="831" alt="schermata 2017-07-27 alle 16 48 41" src="https://user-images.githubusercontent.com/9007227/28676695-987127aa-72eb-11e7-85f1-c28f70b250ea.png">

Alkarex:
Strange error you have there. What browser? Could you try with browser extensions disabled? Could you please try with a couple other browsers?

giorgiobeggiora:
With addons disabled, i tried with  last version of Chrome, Internet Explorer and Firefox: the error remains.

In firefox, the error message is:

TypeError: notification.find(...).html(...) is undefined

aledeg:
I have the same error sometimes. I've dig in the code and it seems that there is some entry contents that are interfering with the page tags.
[...]

Issue #789, opened 2015-02-09
FreshRSS community processes


marienfressinaud:
Hi all!

I have worked a bit on some important processes to be more « community-friendly ». Here are the points I wanted to discuss. It is a bit long, but it is important for the future of FreshRSS ! :)
### General community process
- A mailing list could be great to communicate between developers and users
- I have to finish the CONTRIBUTING file
- I thought to move all the *.freshrss.org domains on a dedicated VPS to be able to give access to some trusted contributors. It also means it will cost more money to me so it will probably wait until I have a new job ;)
- The new doc (https://github.com/FreshRSS/documentation) has to be installed. daux.io does not support image hosting as I would like so I'll have a look at [couscous.io](http://couscous.io/) (used by wallabag): we can keep our current doc files, only internal links will have to be updated if we choose to use this doc software.
- Do we need a forum? It seems to me that if we have a mailing list, we don't need a forum.
- What do you think of using Loomio for taking important decisions? Or the mailing list is enough?
- I wanted to create a sort of schema or map to explain to users who want to contribute, where they should begin.
- And a simple question but hard answers I guess: who own the code? I mean, when I was alone to work on FreshRSS, it was easy, I owned the code. But since people can contribute, it's harder to say. Do we need a sort of « collaboration agreement »?
### Internationalization process
- We have to update i18n.freshrss.org, at least to explain to not use it!
- CONTRIBUTING file has to explain in few steps how to participate to the i18n
- I think a new string must have, at least, an English version. Other versions (French, German and future supported languages) can use the English version in a first time if contributor don't know the correct translation.
- We'll need later a developer script (e.g. `i18n.php`) to manipulate translations (e.g. to add a sentence, to rename a key). For now it's ok since we have to take care of only 3 languages.
### Quality process
- Use a view template system to have a better MVC architecture separation and facilitate theme contributions
- Considerations about a new, supported and documented PHP framework:
  - Clearly, it's not realistic to entirely rewrite FRSS with a new framework BUT…
  - … we have to improve some controller and model source code by reducing the longest methods and think a bit more to the framework logic (ping @dhoko ;))
  - Comments and code documentation are necessary too
  - I would like to release a new proper version of Minz
- Tests + Travis integration are welcomed (@aledeg has already begun, thanks!)
- Improve documentation
- Reduce size of the README: it's quite a complete documentation page actually
[...]

PR #1618, opened 2017-08-21
Increase minimal support to PHP 5.3.8+

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1604

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 2017-XX-XX FreshRSS 1.7.1-dev
 
+* Compatibility:
+	* Minimal PHP version increased to PHP 5.3.8+ to fix sanitize bug [#1604](https://github.com/FreshRSS/FreshRSS/issues/1604)
 * UI
 	* New page for subscription tools [#1534](https://github.com/FreshRSS/FreshRSS/issues/1354)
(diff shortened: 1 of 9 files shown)


Alkarex:
PHP 5.3.6+ is needed to fix the sanitize bug of https://github.com/FreshRSS/FreshRSS/issues/1604
https://github.com/simplepie/simplepie/blob/45c50e895f0657480c57069b25722dfebec1c8e0/library/SimplePie/Sanitize.php#L368-L375

Patch to increase the minimal PHP version from PHP 5.3.3 to 5.3.8+, based on the fact that 5.3.6 and 5.3.7 do not seem to be used much, and we need 5.3.7+ for some crypto functions.
https://w3techs.com/technologies/details/pl-php/5.3/all (loosing <4.8% of PHP installed base)

* Minimal Debian version increases from 6 (squeeze) to 7 (wheezy, 2013/05) https://distrowatch.com/table.php?distribution=debian
* Minimal CentOS version increases from 6 to 7 (2014/07) https://distrowatch.com/table.php?distribution=centos

Alkarex:
Reference to FreshRSS PHP needs: https://github.com/FreshRSS/FreshRSS/issues/789#issuecomment-73878076

Alkarex:
@marienfressinaud Do you still use CentOS? If yes, which version?

marienfressinaud:
I'm using CentOS 7 but I plan to move quite soon to another distribution

Issue #1570, opened 2017-06-05
Do we have an account on Travis?


aledeg:
If not, could we create one with an email similar to "travis <at> freshrss <dot> org" ?

marienfressinaud:
I can configure Travis through my personal account (and I guess you also can), I'm not sure to understand the need to create a generic account. Otherwise, such address would be a mailing list since it has been configured that way for freshrss.org.

aledeg:
The idea was to decouple a user account from the travis user. This way we can protect ourselves from the [bus factor](https://en.wikipedia.org/wiki/Bus_factor).

marienfressinaud:
Didn't forget the ticket :). Travis is not our weakest point against bus factor, I worry a lot more about the server access! Anyway, all people belonging to FreshRSS organization on GitHub should have the exact same access on Travis, so I don't think we need to create a generic account, do you?

However, we may open the problematic to the different accounts we may have (server, Twitter, etc.)

aledeg:
Are you talking about where each new version are hosted?
If so, maybe we can change that process and use github tags directly in github.

marienfressinaud:
I'm talking about freshrss.org and all its subdomains: demo., rss., update., doc. (but this one will be hosted on GH Pages soon). We don't host any version of FRSS anywere, update.freshrss.org provides a script to download an archive from GH, that's all.

Alkarex:
A bit related to https://github.com/FreshRSS/FreshRSS/issues/907 (which could remove the need for the update server)

Frenzie:
Yeah, Travis doesn't have a bus factor provided there are multiple people with sufficient access in the organization.

Alkarex:
https://travis-ci.org/FreshRSS/FreshRSS

PR #1619, opened 2017-08-21
[ci] Add Travis

Frenzie:
References #577.

Some caveats:

1. This won't work until someone enables Travis integration.
2. I hope I've captured the [coding style](http://doc.freshrss.org/en/developers/01_First_steps/#coding-style) correctly. It'll be a while before it passes. If nothing else there are many spaces in the code base where there should be tabs. See [here](https://travis-ci.org/Frenzie/FreshRSS/jobs/266845131) for a detailed log. The fixes can be applied almost entirely automatically.
3. This is just for basic linting and coding standards compliance. Adding [PHPUnit](https://phpunit.de/) is for another day. ;-)


Alkarex:
@Frenzie Super. I think you should have the rights in the project settings, if you want to have a look.

Frenzie:
As a "member" I can now push to all the repos, but only "owners" can change project settings like continuous integration.

A couple of preliminary issues I've spotted.

### Style guide vs. reality

The style guide quite clearly says:
```php
if ($a == 10) {
    // do something
}

if ((int)$a == 10) {
    // do something
}
```

but I noticed that there's a great quantity of this kind of thing:

```
@@ -718,20 +716,15 @@ class SimplePie
-               if ((version_compare(PHP_VERSION, '5.3', '<') || !gc_enabled()) && !ini_get('zend.ze1_compatibility_mode'))
-               {
-                       if (!empty($this->data['items']))
-                       {
-                               foreach ($this->data['items'] as $item)
-                               {
+               if ((version_compare(PHP_VERSION, '5.3', '<') || !gc_enabled()) && !ini_get('zend.ze1_compatibility_mode')) {
+                       if (!empty($this->data['items'])) {
+                               foreach ($this->data['items'] as $item) {
                                        $item->__destruct();
                                }
                                unset($item, $this->data['items']);
                        }
-                       if (!empty($this->data['ordered_items']))
-                       {
-                               foreach ($this->data['ordered_items'] as $item)
-                               {
+                       if (!empty($this->data['ordered_items'])) {
+                               foreach ($this->data['ordered_items'] as $item) {
```

I tend to prefer what the style guide says, but the code itself seems to be 50/50 on it (not an actual stat).

### The /libs issue

On a related note, should everything in /libs be ignored or only the stuff that doesn't start with `lib_`? It looks like the lib_ files might be FreshRSS-specific.

Alkarex:
Quick partial response:
* favicons, lib_rss, lib_install, lib_date, lib_opml : FreshRSS code to be included
* Minz library is from @marienfressinaud and could probably be included
* SimplePie (this is the code of your example): excluded
* other libs: probably excluded

Ah, I did not know for the rights. Maybe @marienfressinaud or @aledeg could be interested in enabling the Travis config?

Frenzie:
I hadn't looked at it in too much detail yesterday, but sensibly excluding a few libraries makes a lot of difference.

https://travis-ci.org/Frenzie/FreshRSS/jobs/267081829

Travis with all the libs:
> Time: 57.69 secs; Memory: 90.75Mb

Travis without (most of it):
> Time: 12.63 secs; Memory: 22.25Mb

There are a few occurrences of this kind of thing that don't error on my local phpcs but do on Travis. On my Debian I'm currently using PHP_CodeSniffer version 2.7.1, whereas the Travis box installs a brand new PHP_CodeSniffer-3.0.2.tgz. I hope https://github.com/squizlabs/PHP_CodeSniffer/pull/1404 didn't fix this the wrong way.
https://github.com/FreshRSS/FreshRSS/blob/fd43ee546e419fc9fbb9a3ad974d2234d94cffaa/lib/lib_install.php#L70-L72

Frenzie:
I've added a few commits to start addressing some of the errors. I'm keeping them all separate so they're easy to drop or amend if desired.

https://travis-ci.org/Frenzie/FreshRSS/jobs/267229385

marienfressinaud:
I just enabled repo on Travis, you can add `[Build Status](https://travis-ci.org/FreshRSS/FreshRSS)` in README if you want :)

Alkarex:
Thanks for all this cleaning effort @Frenzie 👍

Frenzie:
It's actually surprisingly relaxing combined with listening to [this album I just bought in hi-res](http://www.qobuz.com/be-fr/album/haydn-4-piano-sonatas-hiroaki-takenouchi/3775000078508).

I haven't yet found out if there's a setting to make it shut up about these things `66 | ERROR | [x] Expected 1 space after ":"; newline found`

https://github.com/FreshRSS/FreshRSS/blob/fd43ee546e419fc9fbb9a3ad974d2234d94cffaa/lib/lib_date.php#L66-L67

Frenzie:
I dropped a comment here regarding what I consider a regression compared to the version 2.7.1 I'm still happily using on my desktop: https://github.com/squizlabs/PHP_CodeSniffer/pull/1404/files#r135108075

Personally I would just toss in an extra tab without bothering with pixel art alignment, but it looks beautiful the way it is and obviously @marienfressinaud went through all the trouble of doing it in the first place. I think I might force an older version of phpcs in the meantime.

Alkarex:
@Frenzie How much work do you think there is left in this PR? Should we wait for it before releasing FreshRSS 1.7.1? https://github.com/FreshRSS/FreshRSS/milestone/24

Frenzie:
I'll have another look to see if I can silence the remaining errors this weekend. (Stuff like `Class name is not valid; consider FreshRSS_Update_Controller instead` is a configuration issue.) But we shouldn't delay a bug fix release for just a bit of a cleanup if it's ready. :-)

Frenzie:
Alright, I got it down to 0 errors in the test. Please take a look at the changes and let me know if there are any changes you dislike.

Except for PHP 5.3 (`pear/PHP_CodeSniffer requires PHP (version >= 5.4.0)`) I'm not sure why the test is only passing on PHP 7.1 on Travis, but that aside. My whole point in setting this up is to run the unit tests on all supported versions of PHP and perhaps a few allowed to fail other versions to boot, but for the linting I suppose it doesn't matter.

Alkarex:
Good work! Quickly looking at the changes, I do not see anything unwanted :-)

Frenzie:
I just added everything except PHP 7.1 to allowed failures, but now 5.4 and 5.5 are also passing. In any case, I think this should be safe to squash merge without every PR getting a fail if no one has any objections.

Alkarex:
Please go ahead @Frenzie , so we can give it a try :-)

Alkarex:
Ok, let's give it a try :-)

app/Models/Feed.php
```diff
@@ -481,10 +485,12 @@ function pubSubHubbubSubscribe($state) {
 				CURLOPT_FOLLOWLOCATION => true,
 				CURLOPT_RETURNTRANSFER => true,
 				CURLOPT_USERAGENT => 'FreshRSS/' . FRESHRSS_VERSION . ' (' . PHP_OS . '; ' . FRESHRSS_WEBSITE . ')',
-				CURLOPT_POSTFIELDS => 'hub.verify=sync'
-					. '&hub.mode=' . ($state ? 'subscribe' : 'unsubscribe')
-					. '&hub.topic=' . urlencode($url)
-					. '&hub.callback=' . urlencode($callbackUrl)
+				CURLOPT_POSTFIELDS => http_build_query(
```
Alkarex, line 488: Strange this syntax error was not caught by the linter.
Fixed in https://github.com/FreshRSS/FreshRSS/pull/1633

app/Models/Feed.php
```diff
@@ -481,10 +485,12 @@ function pubSubHubbubSubscribe($state) {
...
+					'hub.verify' => 'sync',
+					'hub.mode' => ($state ? 'subscribe' : 'unsubscribe'),
+					'hub.topic' => urlencode($url),
```
Alkarex, line 491: This broke PSHB. Fix on the way

Frenzie: Should've been an array around it I suppose. Whoops. :)

Alkarex: The array was the first issue (a bit higher up). I wonder why it was not caught by the linter when running Travis. Any idea?
This is another issue, with a double-encoding of hub.topic value :-)

Alkarex: Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1643

Frenzie: I'm not sure to what extent the linter picks up on things the regular parser will. Its main purpose is that which the parser won't pick up. (This particular error is all on me. The parser said it was a big bunch of string concatenation and I realized that PHP is a bit more efficient now than it once was... unfortunately introduced a slight type error in the process.)

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1570

@Frenzie , have you found out why PHP versions 5.3, 5.6, 7.0 are failing?
It looks to be due to some (mis-)configuration, and not due to FreshRSS code:
image

Frenzie:
5.3 won't work with a recent version of CodeSniffer but I don't consider that of any importance (unlike PHPUnit). All the PHP versions are just there as prep for PHPUnit really since just one PHP would do for the linter.

In the case of 5.6 and 7.0 it looks like Travis hasn't set up the include dirs correctly but I don't understand why that should be the case. What is clear is that 5.6 and 7.0 are installed by default while the others are installed only on request.

A workaround mentioned in https://github.com/travis-ci/travis-ci/issues/7154 is to be more specific, such as `5.4.44` (except in this case that'd be something like `5.6.5` or `5.6.6`, also see https://github.com/travis-ci/travis-ci/issues/5410#issuecomment-182756734). Or maybe just a simple symlink would do.

Anyway, I'll look into the specifics if it's relevant for PHPUnit. :)

app/Models/Entry.php
```diff
@@ -189,6 +189,7 @@ public function loadCompleteContent($pathEntries) {
 					);
 				} catch (Exception $e) {
 					// rien à faire, on garde l'ancien contenu(requête a échoué)
+					Minz_Log::warning($e->getMessage());
```
Inverle, line 192: https://github.com/FreshRSS/FreshRSS/issues/9222

Issue #1591, opened 2017-07-10
Logged Out when using Global View


redinsight:
When using the global view to view articles and choosing Mark All As Read on the bottom of the page I am immediately logged out and the articles are not marked as read. I do not have this problem in normal view or reading view, in those views everything works as expected.

Alkarex:
Strange. I will try to have a look in a few days.

Alkarex:
Sorry for the delay. Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1640

PR #1640, opened 2017-09-23
Fix global view CSRF

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1591


Alkarex:
@redinsight Please give this little patch a test if you can

Issue #1506, opened 2017-04-24
Problem showing share option


GeoZzNL:
Hi there,

As you can see, i have some trouble with the share options. Its it out of the screen.

This is not in every width. Its only on phones

![image](https://cloud.githubusercontent.com/assets/12541266/25336070/1ffb9c04-28f6-11e7-8e5b-1ffc6fb911ee.png)

![image](https://cloud.githubusercontent.com/assets/12541266/25336128/625ab742-28f6-11e7-81e0-eef0719892f5.png)

I think it has to do something with a fixed style or something, but i cant find the right code to edit.
I also disabled all extensions, but this does not work.

Do you already know the wrong, and the solution? I maybe know the problem, but cant find the right file to edit

Thanks

Alkarex:
You are almost at the right place. I think it is the style of `.dropdown.menu` that needs to be fixed for mobile views.

![image](https://cloud.githubusercontent.com/assets/1008324/25336551/54931c42-28f8-11e7-9cde-492535994342.png)

GeoZzNL:
I actually already tried that. But when i change the position to fixed, the box is not showing anymore. When i change it to static, this happen: 

![image](https://cloud.githubusercontent.com/assets/12541266/25336764/4ffbfa7c-28f9-11e7-80d8-c7f7a2614dcf.png)

All other positions are not working

Alkarex:
I think one needs to add a rule for small screens (e.g. `@media all and (min-width:XXX) {}`) and redo the styling of the drop-down menu in that case.

Alkarex:
@GeoZzNL Could you please give a try to the /dev version of FreshRSS, or try this patch https://github.com/FreshRSS/FreshRSS/pull/1645 and confirms whether it fixes your problem?

PR #1645, opened 2017-09-27
Change layout for the share options on mobile

aledeg:
Before, it was impossible to click on links to share on small screens.
Now, all dropdowns are opened at the bottom of the screen and stays there.

Linked to #1506

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -835,6 +835,15 @@ input:checked + .slide-container .properties {
 	.no-mobile {
 		display: none;
 	}
+	.dropdown-menu {
+		border-radius: 0 !important;
+		bottom: 0;
+		position: fixed;
+		width: 100%;
+	}
+	.dropdown-menu::after {
+		display: none;
+	}
 	.aside .toggle_aside,
 	.nav-login {
 		display: block;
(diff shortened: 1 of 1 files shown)


Frenzie:
Where is this dropdown menu? Whenever I see `!important` my spidey senses start tingling. :-P

aledeg:
@Frenzie I know :)
The dropdown menu are the one that you can see when you click on the share button and on the user queries. I used ```!important``` because I want to make sure that there is no radius on that dropdown since I display it at the bottom of the screen and it uses the whole space. The round corner was quite disgraceful.
Feel free to come with a better solution.

Alkarex:
Would it be possible to find a more qualified rule, hence avoiding the need of `!important` ?

Frenzie:
I don't really want to figure out where this sharing thing is right now[1] but for the user query button you could increase specificity by using something like `.dropdown .dropdown-menu`.

[1] It looks like my settings say sharing in the bottom line but I suppose I'd have to set up some sharing for it to display.

aledeg:
@Frenzie @Alkarex Done. I removed the ```!important```

Frenzie (approved):

PR #1647, opened 2017-09-29
Remove limitation on toggleContent

### LINKED ISSUE #1177: Go back to previous view when collapsing article
J'espère que je vais être clair : quand j'ouvre un article, la vue est calée sur l'article de dessus, et se décale à chaque fois au fur et à mesure du défilement. Cependant, quand les articles sont longs et qu'il faut descendre pour les lire, au moment du repliement (avec le raccourci ou en cliquand sur les côtés), la vue est complétement décalée ; j'aimerais pouvoir revenir au calquage avec l'article du dessus.

aledeg:
Before, articles were repositioned only when the new and the old article were different. When they were the same, it was just closed.
It was frustrating with long articles, especially during reading.

Now, when an article is collapsed, it is repositionned no matter the collapsing method.

This should fix #1177

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -266,11 +266,9 @@ function toggleContent(new_active, old_active) {
 				new_pos += old_scroll;
 			}
 
-			if (old_active[0] !== new_active[0]) {
-				new_active.children(".flux_content").first().each(function () {
-					box_to_move.scrollTop(new_pos).scrollTop();
-				});
-			}
+			new_active.children(".flux_content").first().each(function () {
+				box_to_move.scrollTop(new_pos).scrollTop();
+			});
 		} else {
 			if (relative_move) {
 				new_pos += old_scroll;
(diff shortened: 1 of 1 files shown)


Frenzie:
There's a setting I've disabled because I don't like it at all. (I sometimes like to click and highlight while reading.) But nice improvement to it. 👍

Alwaysin:
It seems I cannot update:

```The update process has encountered an error: Git error: Updating 3ce2b4d..e53c10f```

Any idea to why?

@Frenzie The problem is (hopefully was thanks to aledeg) here even with the setting disabled, when you use 'c' to collapse the article.

aledeg:
I almost committed without fixing the collapsing mecanism :D

Alkarex:
@Alwaysin Is it an error message from the Web interface? If yes, it is likely due to some local permission problems.
Please try https://github.com/FreshRSS/FreshRSS/tree/master/cli#using-git

Alwaysin:
@Alkarex thank you, it works perfectly after this set of commands!
@aledeg thank you, it works perfectly after this patch!

:D :D

PR #1653, opened 2017-10-01
Add a translation validation tool.

aledeg:
It's triggered by Travis to check what is missing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,16 +14,27 @@ install:
 
 script:
   - phpenv rehash
-  - phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
+  - |
+    if [[ $VALIDATE_STANDARD == yes ]]; then
+      phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
+    fi
+  - |
+    if [[ $CHECK_TRANSLATION == yes ]]; then
+      php tools/check.translation.php -r
+    fi
 
-env: # important! otherwise no job will be allowed to fail
+env:
+  - CHECK_TRANSLATION=no VALIDATE_STANDARD=yes
+
 matrix:
-  # PHP 5.3 only runs on Ubuntu 12.04 (precise), not 14.04 (trusty)
+  fast_finish: true
   include:
     - php: "5.3"
       dist: precise
-  fast_finish: true
+    - php: "7.1"
+      env: CHECK_TRANSLATION=yes VALIDATE_STANDARD=no
   allow_failures:
+    # PHP 5.3 only runs on Ubuntu 12.04 (precise), not 14.04 (trusty)
     - php: "5.3"
       dist: precise
     - php: "5.4"
@@ -32,3 +43,4 @@ matrix:
     - php: "7.0"
     - php: hhvm
     - php: nightly
+    - env: CHECK_TRANSLATION=yes VALIDATE_STANDARD=no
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -14,16 +14,27 @@ install:
 
 script:
   - phpenv rehash
-  - phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
+  - |
+    if [[ $CHECK_TRANSLATION == no ]]; then
+      phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
+    fi
+  - |
+    if [[ $CHECK_TRANSLATION == yes ]]; then
```
Frenzie: Awesome work on the PHP side, but I have a little nitpick on the Travis/shellscript side. I think it makes more sense to just split it up in a bunch of true vars. :-)

```
env:
  matrix:
    - CHECK_TRANSLATION=1 # or true/yes/whatever
    - CODE_SNIFFER=1
    - PHP_UNIT=1
    - CHECK_TRANSLATION=1 PHP_UNIT=1 # look, a combo!
```

Then the script is simply something like:
```
script:
  - if [ "$CHECK_TRANSLATION" = 1 ]; then php tools/check.translation.php -r; fi
  - if [ "$CODE_SNIFFER" = 1 ]; then phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p; fi
```

If more complicated scripts are desired in the future you can adapt it like:

```
  - if [ "$SOMETHING" = something ]; then source .ci/something_whatever.sh; fi # or without source; but don't forget some kind of source common.sh in each script in that case
```

Implicitly, I also object to Bashisms like `[[ $bla == lala ]]` unless they bring something unique to the table. :-P

tools/translation.ignore.php
```diff
@@ -0,0 +1,59 @@
...
+$ignore['fr']['index.php'][] = 'about.version';
+$ignore['fr']['index.php'][] = 'log._';
+$ignore['fr']['index.php'][] = 'log.title';
+$ignore['fr']['install.php'][] = 'title';
+$ignore['fr']['install.php'][] = 'this_is_the_end';
+$ignore['fr']['sub.php'][] = 'bookmarklet.title';
+$ignore['fr']['sub.php'][] = 'feed.description';
+$ignore['fr']['sub.php'][] = 'feed.number_entries';
+
+
```
Frenzie: Stray newline. (I really am nitpicking today, aren't I. 👼 )

.travis.yml
aledeg: I've done the ```source``` thing in the past but it can become a mess. I prefer having everything in the same file.

aledeg:
I've updated the code

.travis.yml
Frenzie: I disagree as I think .travis.yml is horribly unsuited for evaluating the correctness of scripts. Even for simple one liners it's awkward, but that's neither here nor there. My example was poorly chosen because it suggests multiple files even though for our purposes a single script.sh would be much better for the foreseeable future.

What I'm saying is that I think explicit enabling of checks is more elegant than explicit everything.

PS
A `.ci/script.sh` would look like this:

```bash
#!/bin/bash
if [ "$CHECK_TRANSLATION" = 1 ] then
  php tools/check.translation.php -r
fi
if [ "$CODE_SNIFFER" = 1 ] then
  phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
fi
```

Frenzie:
Alright, cool!

Issue #1627, opened 2017-08-30
Impossible to subscribe to some RSS feed because of lack of HTTP headers


Massedil:
Some feed required a browser's User-Agent HTTP header, an "Accept-Language" HTTP header or an "Accept" HTTP header, for example.

It would be great if we could add or override the HTTP headers for each RSS feed.

aledeg:
It's a good idea

aledeg:
@Massedil Do you have any feeds falling in that use case?

Massedil:
Yes. 
"https://www.mondialisation.ca/feed" needs a "User-Agent" HTTP header.
"http://www.nord.gouv.fr/syndication/flux/rss_feed_69" doesn't work with FreshRSS, but I don't know why.

aledeg:
Thank you. I'll have a look at them to figure out why they are not working and how to make them work in FreshRSS.

Alkarex:
`nord.gouv.fr` is another example blocking SimplePie:

```sh
curl -v -A "SimplePie/1.3" 'http://www.nord.gouv.fr/syndication/flux/rss_feed_69'
*   Trying 90.85.16.36...
* Connected to www.nord.gouv.fr (90.85.16.36) port 80 (#0)
> GET /syndication/flux/rss_feed_69 HTTP/1.1
> Host: www.nord.gouv.fr
> User-Agent: SimplePie/1.3
> Accept: */*
>
< HTTP/1.1 403 Forbidden
< Server: nginx
< Date: Sat, 07 Oct 2017 11:16:27 GMT
< Content-Type: text/html; charset=iso-8859-1
< Content-Length: 230
< Connection: keep-alive
< Vary: Accept-Encoding
<
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>403 Forbidden</title>
</head><body>
<h1>Forbidden</h1>
<p>You don't have permission to access /syndication/flux/rss_feed_69
on this server.</p>
</body></html>
```

Alkarex:
Can now be implemented more easily with https://github.com/FreshRSS/FreshRSS/pull/1838

Alkarex:
There is a new simple extension mechanism to do that.
See https://github.com/FreshRSS/FreshRSS/pull/1924 and included example for Tumblr:

[...]

Issue #1622, opened 2017-08-23
Cannot add this feed - validator say is ok


planetahuevo:
https://ayudawp.com/feed/

http://validator.w3.org/feed/check.cgi?url=https://ayudawp.com/feed/

Not sure why...

Any ideas?

Thanks!

planetahuevo:
Another one: same issue.
http://validator.w3.org/feed/check.cgi?url=https://tellado.es/feed/

Frenzie:
Hah, it sure has a very helpful error in the log.
```
2017-08-23 15:12:08Unknown error for feed [https://ayudawp.com/feed/]
```

You could try adding `#force_feed` to the end of the URL (see #1410). However, the mimetype on both of them seems to be `application/rss+xml` so the root cause is different.

planetahuevo:
Tried and not working but at least I have a nicer error log. :) 
I have the same problem with this other feed from the same person (server issue?)
```
2017-08-23 14:22:29

cURL error 22: The requested URL returned error: 400 Bad Request [https://ayudawp.com/feed/]

2017-08-23 12:55:30

cURL error 22: The requested URL returned error: 400 Bad Request [https://tellado.es/feed/]```

Frenzie:
It seems fine when I try something like `curl -v -A "SimplePie/1.7" https://ayudawp.com/feed/` so I'm not really sure what could be happening.

It works in rss-bridge on my same server, so if it is SimplePie-related then it's only in the FreshRSS version. `/rss-bridge/?action=display&bridge=Filter&url=https%3A%2F%2Fayudawp.com%2Ffeed%2F&filter=&filter_type=permit&format=Atom`, you can try it on a public instance [here](https://bridge.suumitsu.eu/?action=display&bridge=Filter&url=https%3A%2F%2Fayudawp.com%2Ffeed%2F&filter=&filter_type=permit&format=Atom).

planetahuevo:
No idea, this goes far beyond my knowledge on code. 
I have found other issues adding sites but those didn´t pass the validator, so I didn´t think the problem is related.
Is there anything I can test or help?

Frenzie:
You could play around with the curl parameters but right now I don't have the time to investigate which ones are relevant or even where they're located. And maybe use `var_dump()` or `print_r()` on the output, as well. I like to use `error_log(print_r($relevant_var, true))` but if all of that is meaningless to you then the fastest option is probably to wait for @Alkarex to notice this. :-P

[...]

Issue #1607, opened 2017-07-30
Can't add a RSS


IPv777:
Hi, I can't add this RSS : http://homputersecurity.com/feed/
An another person could try please?

Raymondcal:
Doesn't work here, too.
I don't know why, there is no log in FreshRSS either.

Alkarex:
The error message is: `cURL error 22: The requested URL returned error: 400 Bad Request [http://homputersecurity.com/feed/]`

Trying manually, it looks like there is a server-side rule on the user-agent *SimplePie*. Maybe someone has been refreshing that feed too much...

```sh
curl -v -A 'SimplePie/1.5' 'http://homputersecurity.com/feed/'
```

```
HTTP/1.1 400 Bad Request
...
The page you are trying to access is restricted due to a security rule.
```

elevatorsimulator:
I can add `http://homecomputersecurity.com/feed` but not `https://borodust.org/atom.xml`; I get `Unknown error for feed [https://borodust.org/atom.xml]` in the log file.

azlux:
I have contact the creator of homecomputersecurity , that why it's working now. @elevatorsimulator you can send an email to the creator of borodust to remove the filter on the rss.

elevatorsimulator:
Sorry, what filter? I have no trouble borodust's feed to other "rss"
readers (e.g. RSSOwl), so I'm not sure the problem lies with the feed
and not with FreshRSS.

Frenzie:
I have no trouble adding borodust anymore. But the filter would presumably have been for the "SimplePie" user agent, like @Alkarex said.

elevatorsimulator:
You are able to add his feed? What version of FreshRSS are you using? I'm using the latest repo version, and it doesn't work.

I traced back the error message (`Unknown error for feed`). It gets thrown in line 258 of ./app/Models/Feed.php, and it gets thrown because `$mtime = $feed->init()` evaluates to false as a boolean. The object $feed is an instance of SimplePie, and for some reason it fails to initialize.

Update: It now works all of a sudden.

Frenzie:
Actually I never checked before, so "anymore" might be inaccurate. Anyway, I'm using 1.8.

Alkarex:
Confirmed to work in FreshRSS 1.9.0 with the new default (and customisable) user-agent.

PR #1656, opened 2017-10-07
Remove SimplePie name from HTTP User-Agent string

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1622#issuecomment-334928486
https://github.com/FreshRSS/FreshRSS/issues/1627
https://github.com/FreshRSS/FreshRSS/issues/1607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,15 @@
 
 ## 2017-1X-XX FreshRSS 1.8.1-dev
 
+* SimplePie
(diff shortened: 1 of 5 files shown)


Frenzie (approved):
Looks good to me.

Alkarex:
It looks like that SimplePie, which is used by many different software solutions, is regularly banned by RSS servers when indicated in the HTTP User-Agent header.
This patch removes the "SimplePie" name from FreshRSS HTTP request headers.
And we need to ensure good HTTP / netizen practices in FreshRSS.

Issue #1655, opened 2017-10-03
Favicons not loading from feeds?


donlencho:
Hello,
I just did a fresh install of version 1.8 and imported my feeds form a previous 1.6 installation.
Everything went fine but I got no favicons for my feeds, all articles have this icons:

![fc97eec4](https://user-images.githubusercontent.com/8362161/31146424-f9987df0-a886-11e7-89a7-28f18b961146.jpg)

Is there a way to force it to load icons, or will it come at some point, or is there a problem somewhere?

Thanks!

Alkarex:
Hello @donlencho 
favicons should appear by themselves, and are regularly refreshed. If you still have the problem, let us know.

donlencho:
Well it's been a few days now, the feeds refresh regularly, but still no favicons.
I tried changing the permissions on the data/favicons folder and the files inside, but it's not working better.
Thanks for the help in advance (although I admit I can live without favicons!).

Alkarex:
Could you please show the content of `./FreshRSS/data/favicons/` ?

```sh
ls -alh FreshRSS/data/favicons/
```

If you see some *.ico files there, try to manually open some of them and check whether they are the default icon or a proper one. If they are only default icons, try to delete them:

```sh
rm FreshRSS/data/favicons/*.ico
```

And then force-refresh your FreshRSS home page in your Web browser.

donlencho:
I had actually done that before, there is one icon per feed, but they are all the default one, like mentioned in the first post. Deleting them just makes it recreate them identically on next refresh.

Alkarex:
Please put the following script in `./FreshRSS/p/test-fav.php` and access it from your browser:

```php
<?php
error_reporting(E_ALL);

$url = 'https://alexandre.alapetite.fr/';

require('../constants.php');
require(LIB_PATH . '/lib_rss.php');
require(LIB_PATH . '/favicons.php');

header('Content-Type: text/plain; charset=UTF-8');
header('Content-Disposition: inline; filename="test-fav.txt"');

function get_favicon($url) {
	$url = trim($url);
	echo 'searchFavicon ', $url, "\n";
	$favicon = searchFavicon($url);
	if ($favicon == '') {
		$rootUrl = preg_replace('%^(https?://[^/]+).*$%i', '$1/', $url);
		if ($rootUrl != $url) {
			$url = $rootUrl;
			echo 'searchFavicon ', $url, "\n";
			$favicon = searchFavicon($url);
		}
[...]

PR #1657, opened 2017-10-08
CURLOPT_FOLLOWLOCATION open_basedir bug

Alkarex:
CURLOPT_FOLLOWLOCATION cannot be activated when an open_basedir is set
https://github.com/FreshRSS/FreshRSS/issues/1655#issuecomment-334999448

https://stackoverflow.com/questions/6918623/curlopt-followlocation-cannot-be-activated

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -481,18 +481,21 @@ class FreshRSS_Feed extends Minz_Model {
 			}
 			$ch = curl_init();
 			curl_setopt_array($ch, array(
-				CURLOPT_URL => $hubJson['hub'],
-				CURLOPT_FOLLOWLOCATION => true,
-				CURLOPT_RETURNTRANSFER => true,
-				CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
-				CURLOPT_POSTFIELDS => http_build_query(array(
-					'hub.verify' => 'sync',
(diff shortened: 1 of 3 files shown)


app/Models/Feed.php
```diff
@@ -481,18 +481,22 @@ function pubSubHubbubSubscribe($state) {
...
+					CURLOPT_RETURNTRANSFER => true,
+					CURLOPT_POSTFIELDS => http_build_query(array(
+						'hub.verify' => 'sync',
+						'hub.mode' => $state ? 'subscribe' : 'unsubscribe',
+						'hub.topic' => $url,
+						'hub.callback' => $callbackUrl,
+						)),
+					CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
+					CURLOPT_MAXREDIRS => 10,
+					CURLOPT_REDIR_PROTOCOLS => CURLPROTO_HTTP | CURLPROTO_HTTPS,
```
Frenzie: Is there any need to limit this? It's my understanding that possibly problematic stuff like `FILE` is already disabled by default.

Alkarex: You are right. If the CURLOPT_REDIR_PROTOCOLS options exists, then it already limits the problematic ones. I will remove it.

Frenzie (approved):
Looks good to me.

PR #1660, opened 2017-10-09
Add an entry in the subscription tool page

aledeg:
I reworked @Alkarex idea proposed in #1292. I though it was a good idea to merge everything in the same location.

### DIFF
--- a/app/i18n/en/sub.php
+++ b/app/i18n/en/sub.php
@@ -1,6 +1,10 @@
 <?php
 
 return array(
+	'api' => array(
+		'documentation' => 'Copy the following URL to use it within an external tool.',
+		'title' => 'API',
+	),
 	'bookmarklet' => array(
 		'documentation' => 'Drag this button to your bookmarks toolbar or right-click it and choose "Bookmark This Link". Then click "Subscribe" button in any page you want to subscribe to.',
 		'label' => 'Subscribe',
--- a/app/i18n/fr/sub.php
+++ b/app/i18n/fr/sub.php
@@ -1,6 +1,10 @@
 <?php
 
 return array(
+	'api' => array(
+		'documentation' => 'Copier l’URL suivante dans l’outil qui utilisera l’API.',
+		'title' => 'API',
+	),
 	'bookmarklet' => array(
 		'documentation' => 'Glisser ce bouton dans la barre des favoris ou cliquer droit dessus et choisir "Enregistrer ce lien". Ensuite, cliquer sur le bouton "S’abonner" sur les pages auxquelles vous voulez vous abonner.',
 		'label' => 'S’abonner',
--- a/app/i18n/nl/sub.php
+++ b/app/i18n/nl/sub.php
@@ -1,6 +1,10 @@
 <?php
 /* Dutch translation by Wanabo. http://www.nieuwskop.be */
 return array(
+	'api' => array(
+		'documentation' => 'Copy the following URL to use it within an external tool.',// TODO
+		'title' => 'API',// TODO
+	),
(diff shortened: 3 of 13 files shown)


app/i18n/en/sub.php
```diff
@@ -1,6 +1,10 @@
 <?php
 
 return array(
+	'api' => array(
+		'documentation' => 'Copy the following URL to consume it with an external tool.',
```
Frenzie: Consume? Seems like a slightly odd word choice. Maybe just `use`?

* `for use in an external tool`
* `to use it with an external tool`

app/i18n/fr/sub.php
```diff
@@ -1,6 +1,10 @@
 <?php
 
 return array(
+	'api' => array(
+		'documentation' => 'Copier l\'URL suivante pour l\'utiliser dans l\'outil qui consommera l\'API.',
```
Frenzie: See, even in French you're using use, although apparently the tool consumes the API. :-P

aledeg: In French, the consume translation seems odd. That's why I used "use".

Frenzie: Had I read it in French I would have thought "oh well, I guess that's French", or perhaps more Obélix-style "ils sont fous, ces français" but I can confidently say that I don't much like it in English. ;-)

Alkarex: May I suggest: `'Copier l’URL suivante dans l’outil qui utilisera l’API.'` ?
And I would rather go with *Unicode curly apostrophe* rather than *backslash straight single quote*.

Alkarex:
For the languages not yet translated, could you please add the English version with a `//TODO` at the end of the line? E.g. like https://github.com/FreshRSS/FreshRSS/blob/4e174ed9dd84ce4f8de410baf6a6e0fde9160055/app/i18n/tr/conf.php#L96

aledeg:
@Frenzie I've removed the use of *consume*.
@Alkarex I've added missing translations in all languages.

app/i18n/nl/sub.php
```diff
@@ -1,6 +1,10 @@
 <?php
 /* Dutch translation by Wanabo. http://www.nieuwskop.be */
 return array(
+	'api' => array(
+		'documentation' => 'Copy the following URL to use it within an external tool.',// TODO
```
Frenzie, line 5: Kopieer de volgende URL om hem in een externe toepassing te gebruiken.

Frenzie (commented):
I should probably just stick this in a follow-up PR.

app/i18n/nl/sub.php
```diff
@@ -1,6 +1,10 @@
...
+		'title' => 'API',// TODO
+	),
 	'bookmarklet' => array(
 		'documentation' => 'Drag this button to your bookmarks toolbar or right-click it and choose "Bookmark This Link". Then click "Subscribe" button in any page you want to subscribe to.',// TODO
```
Frenzie, line 9: Sleep deze knop naar je bladwijzerwerkbalk of klik erop met de rechtermuisknop en kies "Deze link aan bladwijzers toevoegen."

Frenzie (approved):

Alkarex:
@Frenzie Will you do a new PR for your translations?

Frenzie:
Translations added in #1661.

Alkarex:
Hum, a bit late to come with my comment, sorry, but for a future improvement:
image

* I think the two sections (Firefox and API) are a bit redundant.
* The formulation of the API section could be misinterpreted for the link to use in e.g. mobile clients.

Issue #1583, opened 2017-06-13
Command-line interface for the database maintenance


lapineige:
Following the discussions here #1582, a CLI tool to do the database maintenance can be useful in case of broken configuration.

aledeg:
What are the feature that we should include in such tool?

lapineige:
In that case it was needed to do the database optimisation from the commandline (the web version always timed out).
That's all I need, but well if you have other ideas that might help in case of broken configuration...

Don't hesitate to ask me if you need any test.

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/918

Alkarex:
Tests welcome https://github.com/FreshRSS/FreshRSS/pull/1663

Issue #918, opened 2015-07-23
Perform VACUUM on SQLite db


marienfressinaud:
We have an "Optimize DB" button but there is nothing behind when using SQLite (https://github.com/FreshRSS/FreshRSS/blob/beta/app/Models/EntryDAOSQLite.php#L183). An idea would be to implement a VACUUM action BUT WAIT, we have to discuss about it first!

I don't have time to write a proper explanation about benefits and drawbacks of VACUUM so I let two interesting links:
- An article from Jim Nelson about SQLite in Geary software: https://blogs.gnome.org/jnelson/2015/01/06/sqlite-vacuum-and-auto_vacuum/
- And an article about Liferea's startup performance: http://jeff.ecchi.ca/blog/2011/12/24/investigating-lifereas-startup-performance/

Both are interesting and should be read before trying to implement a solution.

Alkarex:
See https://github.com/FreshRSS/FreshRSS/pull/1663

PR #1663, opened 2017-10-11
CLI optimize database

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1583
And VACUUM in SQLite https://github.com/FreshRSS/FreshRSS/issues/918


cli/db-optimize.php
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/php
+<?php
+require('_cli.php');
+
+$options = getopt('', array(
+		'user:',
+	));
+
+if (empty($options['user'])) {
+	fail('Usage: ' . basename(__FILE__) . " --user username");
```
aledeg: Can't you just use ```__DIR__``` instead of ```basename(__FILE__)```?

aledeg (approved):
I've tested for MySql. It's working fine. Good job. Should I move the i18n tools in the cli folder?

Alkarex:
Ah yes, this sounds like a good idea. I have not looked much at your i18n tools yet, but I assume they are all invoked from command line, aren't they?

Alkarex:
Ah, PostgreSQL also has `VACUUM` https://www.postgresql.org/docs/9.2/static/sql-vacuum.html

Alkarex:
Would some our of dear PostgreSQL users be willing to give a try to our VACUUM implementation? :-)
@anayrat @Trim @tcitworld ?
It is available in the dev branch of FreshRSS, accessible by CLI or with a Web button in FreshRSS settings.

anayrat:
Hello,
I don't do tests for the moment. But normally, manual vacuum is not necessary. PostgreSQL perform autovacuum (enabled by default) depending on the activity (row inserted, updated...) : https://www.postgresql.org/docs/current/static/routine-vacuuming.html#autovacuum

Eventually, you can perform `VACUUM FULL` which rewrite all the tables and rebuild indexes. But it acquire an exclusive lock. It must be an exceptional maintenance.

PR #1671, opened 2017-10-14
GitHub update

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/907
https://github.com/FreshRSS/update.freshrss.org
Work in progress

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,15 @@
 
 ## 2017-1X-XX FreshRSS 1.8.1-dev
 
+* SimplePie
+	* Remove "SimplePie" name from HTTP User-Agent string [#1656](https://github.com/FreshRSS/FreshRSS/pull/1656)
 * Misc.
 	* Travis translation validation tool [#1653](https://github.com/FreshRSS/FreshRSS/pull/1653)
 
 
 ## 2017-10-01 FreshRSS 1.8.0
 
-* Compatibility:
+* Compatibility
 	* Minimal PHP version increased to PHP 5.3.8+ to fix sanitize bug [#1604](https://github.com/FreshRSS/FreshRSS/issues/1604)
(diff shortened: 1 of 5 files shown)


app/update/update_util.php
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	$index = $dir . '/index.html';
+	$data = <<<EOF
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en-GB" lang="en-GB">
+<head>
+<meta charset="UTF-8" />
+<meta http-equiv="Refresh" content="0; url=/" />
```
Frenzie: Is this root `/` quite what is wanted here?

Alkarex: This is some code from Marien, directly from the current update server.
This part is generating an index.html. I have a habit of putting an index.html file in directories that should not be browsed (even when there are other rules such as .htaccess, etc.). The redirect address is not very important, as it is not supposed to be used (except during manual / accidental browsing).

kevinpapst:
This might not be directly related to this PR, but its related to the update mechanism: 
I would kindly ask for a more thoroughly permission check (updateController ~line 80). When I crashed my FreshRSS installation, that was due to a stupid permission problem. Some files were writable, some not. The current is_writable() check only tests the main directory and does not act friendly if some files are not writable. Adding a simple recursive file and directory check would likely solve the issue. Yes, it should not be a FreshRSS task to handle file permissions properly, but it should warn the user if the update could potentially fail.

Alkarex:
I can see I am getting in late for this PR. Any help welcome :-)

Alkarex:
Damn, I have broken my PR :-(

Alkarex:
Replaced by https://github.com/FreshRSS/FreshRSS/pull/1760
Sorry for the wrong manipulation.

PR #1673, opened 2017-10-15
Move translation tools into the cli folder

aledeg:
Translation tools must be used on cli. It is better to have them in the cli folder.


Alkarex:
@aledeg Could you please have a look? It seems like the translation tools are failing in Travis with PHP 7.1:

https://travis-ci.org/FreshRSS/FreshRSS/jobs/297333559
image

By the way, very nice with the translation statistics 👍

aledeg:
@Alkarex it's on purpose to see that there is untranslated strings. But the failing tests are not blocking the final status.
When everything is translated, the result would be green.

Alkarex:
Ok, but then why is it there a difference between PHP 7.0 and PHP 7.1? https://travis-ci.org/FreshRSS/FreshRSS

Frenzie:
@Alkarex They're not running the same command. ;-)

The main 7.1 is at the top with `CHECK_TRANSLATION=no VALIDATE_STANDARD=yes`, which is the same as the passing 7.0. The allowed_failure 7.1 uses `CHECK_TRANSLATION=yes VALIDATE_STANDARD=yes`.

aledeg:
translation is checked only in 7.1. The failing 7.0 has nothing to do with the translation but with an unsatisfied dependency on phpcs

Alkarex:
Thanks for the info, @aledeg and @Frenzie 👍

Issue #1521, opened 2017-05-06
[Feature Request] Share to Mastodon


prolibre:
tout est dans le titre :-) 

et un grand merci pour votre travail... FreshRSS est une perle !

Alkarex:
@prolibre Merci. Auriez-vous sous la main la syntaxe du partage avec Mastodon ?

prolibre:
désolé je reviens tardivement et les mains vides ... non pour l'heure je n'ai rien trouvé pour vous aider :-(

Epy:
Je crois que tout est décrit dans cette page de doc: https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md

aledeg:
I guess it's called a Status on Mastodon. See [here](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#posting-a-new-status).
The main difference between that share and the one we already have is the HTTP action. ATM we only use the GET action but this one requires a POST action.
It's not as easy as it sounds at the first glance.

Alkarex:
I have not read the spec' yet. Do any of you have an example of cURL request(s) that would do the job?

aledeg:
[Here](https://stackoverflow.com/a/33880971/4424102) is something that might help!

aledeg:
So, I was working on that. I'm almost done but something came in the way. Firefox is blocking the post due to Content Security Policy. Apparently, there is no easy way to overcome that.
If you have any ideas, I would love to hear them.

Alkarex:
@aledeg Could you give some more details about the problem?

aledeg:
@Alkarex When I send the post, the message is blocked by the browser because the destination URL is not on the same server. Thus triggering the rules of Content Security Policy.
I don't know if you need more than that.

Frenzie:
Is the destination URL something controlled by a third-party?

Edit: silly question, of course it is or there would be an easy way around. :-P

Alkarex:
Maybe the request can be sent server-side, using OAuth https://github.com/tootsuite/documentation/blob/master/Using-the-API/Testing-with-cURL.md

marienfressinaud:
Asking Mastodon developers to add a GET action for sharing might be easier, don't you think? :p

aledeg:
@marienfressinaud it could be done for Mastodon alone, but if we want to ask all providers, it would be a mess. For reference, #1513 .

Alkarex:
[...]

PR #1674, opened 2017-10-17
Add a Mastodon share

aledeg:
See #1521

This is a work in progress. It needs some testing on a real Mastodon instance.
I am not really satisfied with the code eventhough it's working fine.
Advices are welcome.


Frenzie:
Any part in particular?

aledeg:
The header part and the js. It feels oddish.

Frenzie:
The header thing looks a bit odd I suppose but it's my understanding that this is a necessity?

The JS seems fine to me, aside from me not particularly liking jQuery/jQuery-style. :-P

aledeg:
The header is mandatory to be able to post on other servers.

p/scripts/main.js
```diff
@@ -1172,6 +1172,14 @@ function init_print_action() {
 	});
 }
 
+function init_post_action() {
+	$('.item.share > a[href="POST"]').click(function(event){
```
Alkarex: I can better like with spaces before the anonymous function and before the curly bracket like:
`.click(function (event) {` but this is just a matter of taste :-)

app/FreshRSS.php
```diff
@@ -111,7 +111,12 @@ private static function loadNotifications() {
 	public static function preLayout() {
 		switch (Minz_Request::controllerName()) {
 			case 'index':
-				header("Content-Security-Policy: default-src 'self'; child-src *; frame-src *; img-src * data:; media-src *");
+				$urlToAuthorize = array_filter(array_map(function($a) {
+					if ('POST' === $a['method']) {
+						return $a['url'];
+					}
+				}, FreshRSS_Context::$user_conf->sharing));
+				header(sprintf("Content-Security-Policy: default-src 'self' %s; child-src *; frame-src *; img-src * data:; media-src *", implode(' ', $urlToAuthorize)));
```
Alkarex: It is not a big deal, but it would be better to keep `default-src 'self'` and update only what you need, which is likely `connect-src`. Otherwise, `default-src` also applies to scripts, css, etc.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src

Alkarex:
This looks quite good to me :-) See a couple of comments inline

p/scripts/main.js
Frenzie: I prefer that as well, but it's indeed a matter of taste. ;-)

aledeg: I'll change that later. I have no preference on that matter.

app/FreshRSS.php
aledeg: I wasn't aware of that. I'll modify accordingly.

aledeg:
I've added the changes we talked about earlier.

Alkarex (approved):
Not tested, but looks good to me 👍

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1904

PR #1691, opened 2017-11-15
Also check pdo_pgsql driver after successful installation

blankoworld:
This should fix #1690

Issue #1690, opened 2017-11-15
Check install don't see pdo_pgsql after a successful installation on postgreSQL


blankoworld:
### Problem

After a successful installation on postgreSQL, system don't detect any PDO driver.

### How to reproduce

1. Use php with pdo_pgsql and pgsql extensions. No other one.
2. Install FreshRSS
3. Go to Administration > Vérification de l'installation
4. System say « Vous ne disposez pas de PDO ou d’un des drivers supportés (pdo_mysql, pdo_sqlite, pdo_pgsql). »

Alkarex:
Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1692

PR #1692, opened 2017-11-16
[FIX] #1690 - Also check pdo_pgsql extension in check_install()

blankoworld:
After this pull request: https://github.com/FreshRSS/FreshRSS/pull/1691

Create a new one with dev branch and without previous PR proposals.

Hope all is OK now!

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -397,12 +397,13 @@ function is_referer_from_same_domain() {
  */
 function check_install_php() {
 	$pdo_mysql = extension_loaded('pdo_mysql');
+	$pdo_pgsql = extension_loaded('pdo_pgsql');
 	$pdo_sqlite = extension_loaded('pdo_sqlite');
 	return array(
 		'php' => version_compare(PHP_VERSION, '5.3.8') >= 0,
 		'minz' => file_exists(LIB_PATH . '/Minz'),
 		'curl' => extension_loaded('curl'),
-		'pdo' => $pdo_mysql || $pdo_sqlite,
+		'pdo' => $pdo_mysql || $pdo_sqlite || $pdo_pgsql,
 		'pcre' => extension_loaded('pcre'),
 		'ctype' => extension_loaded('ctype'),
 		'fileinfo' => extension_loaded('fileinfo'),
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -38,6 +38,7 @@ People are sorted by name so please keep this order.
 * [MSZ](https://github.com/mszkb): [contributions](https://github.com/FreshRSS/FreshRSS/commits?author=mszkb)
 * [Nicolas Elie](https://github.com/nicolaselie): [contributions](https://github.com/FreshRSS/FreshRSS/commits?author=nicolaselie)
 * [Nicolas Lœuillet](https://github.com/nicosomb): [contributions](https://github.com/FreshRSS/documentation/commits?author=nicosomb), [Web](http://www.loeuillet.org/)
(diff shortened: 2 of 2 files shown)


lib/lib_rss.php
```diff
@@ -398,11 +398,12 @@ function is_referer_from_same_domain() {
 function check_install_php() {
 	$pdo_mysql = extension_loaded('pdo_mysql');
 	$pdo_sqlite = extension_loaded('pdo_sqlite');
+	$pdo_pgsql = extension_loaded('pdo_pgsql');
```
Frenzie: Is this order (also below) the most sensible? With only three I suppose it doesn't matter but I'd go for alphabetic. :-)

Frenzie:
You can force push onto a branch btw. Something like this.

```bash
git checkout messed-up-branch
git checkout -b backup-branch # just in case
git checkout messed-up-branch
git reset HEAD~5 # or a few more, whatever
git pull upstream dev # you'll need to set up a remote called upstream for that to work of course
git cherry-pick e5ba61e # apply the one commit on top
git push -f
```

Alternatively you can use `git rebase -i HEAD~5`; depends a bit on what you want.

blankoworld:
Wow, I see. A little bit complex for me, but hey, I found a way to give you the change you expect. I think. Probably :).

Alkarex:
Many thanks, @blankoworld 👍 
(If you want), please add your name to the contributor list: https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1690

aledeg (approved):
Looks good to me. Thank you for your effort.

lib/lib_rss.php
```diff
@@ -397,12 +397,13 @@ function is_referer_from_same_domain() {
...
 function check_install_php() {
 	$pdo_mysql = extension_loaded('pdo_mysql');
+	$pdo_pgsql = extension_loaded('pdo_pgsql');
 	$pdo_sqlite = extension_loaded('pdo_sqlite');
 	return array(
 		'php' => version_compare(PHP_VERSION, '5.3.8') >= 0,
 		'minz' => file_exists(LIB_PATH . '/Minz'),
 		'curl' => extension_loaded('curl'),
-		'pdo' => $pdo_mysql || $pdo_sqlite,
+		'pdo' => $pdo_mysql || $pdo_sqlite || $pdo_pgsql,
```
Frenzie, line 406: This one's still at the end. :-P

Frenzie (approved):
Looks good. Don't mind me going a bit overboard. xD

lib/lib_rss.php
Alkarex: In the rest of the code, there are many places with that order, for historical reasons (MySQL was used first, and then SQLite compatibilité, and last PostgreSQL). So I do not mind either way :-)

Frenzie: It's one of those things where on the one hand it seems like an enormous nitpick. On the other hand I've learned through bitter experience how innocent code where different people just randomly add some stuff at the end or bottom can become an unparseable blob before you good and well realize what happened.

You could quite validly point out that the set will always remain minuscule, but it still serves as an example that might be followed.

Anyway, that's just a braindump. I already explicitly marked it approved after all. ;-)

blankoworld:
> (If you want), please add your name to the contributor list: https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

How should I add it? A new commit above this pull request? Or into the same commit with ```git commit --amend```? What's the best?

Frenzie:
Either way is fine. I prefer more smaller commits but that's just me. ;-)

blankoworld:
Seems that all is fine for me. Hope everything is OK for you now ;)

Frenzie:
Thanks!

Issue #861, opened 2015-05-30
How to upgrade?


ghost:
Hello together

How can i upgrade Freshrss? I fount nothing in the documentation. Just overriding the files do not work. :-)

regards
Thomas

Alkarex:
Hello Thomas,
It depends a bit on how you installed FreshRSS, and also from which to which version you are upgrading. (Please note that I am right now in the process of releasing a new beta version, so today/tomorrow is a bit special).

1) There is a built-in upgrade system in the Web interface of FreshRSS. (But I was not the one developing this mechanism and I have never used it, so I am not sure of the details)

2) If you are familiar with git, it is also convenient. But you may have to keep a eye on the release notes if there are some things to do manually. If your current version is >= 1.0, then it should be fine.

Install:

``` bash
git clone https://github.com/FreshRSS/FreshRSS.git
```

Update:

``` bash
cd ./FreshRSS/
git reset --hard
git pull
```

3) Otherwise, override your files with the files of the new version, and do not delete your existing `./data/` directory. So "just overriding the files" should work in most cases. What is your problem, more precisely ?

Alkarex:
To be revised as part of https://github.com/FreshRSS/FreshRSS/issues/861

Alkarex:
Some more examples of managing FreshRSS updates with git are included in https://github.com/FreshRSS/FreshRSS/tree/dev/cli#using-git

kevinpapst:
Your linked article should go somewhere else, e.g. the official documentation. I updated my installation with the web installer and that crashed my FreshRSS installation. I guess because some file permissions where wrong and the installer did only half of the work (a check upfront  would be a good idea to be included in the updater).
I tried about an hour until I found this URL and the info which files I had to restore and that I had to delete the "force installation" file. 
Would you consider to move that docu to the GitHub pages documentation? I would sit down and write it/copy & paste it if you are willing to include it via PR. What do you think @Alkarex ?

PR #1696, opened 2017-11-29
Docu updates

kevinpapst:
Hi @Alkarex 

here is a first draft to talk about some documentation updates.
What do you think, shall we move all information into one place? 
Only then I looked through the open issues and found your comment https://github.com/FreshRSS/FreshRSS/issues/861#issuecomment-304220690 
It was not obvious to me to look into the cli documentation - I only searched through the docu for infos on how to update.


docs/en/users/02_First_steps.md
```diff
@@ -24,3 +24,6 @@ Now that you know the basic usages, it is time to configure FreshRSS to improve
 * [Filter articles](03_Main_view.md#filtering-articles) for a fast access to a selection
 * [Search an article](03_Main_view.md#searching-articles) published some time ago
 * [Access your feeds on a mobile device](06_Mobile_access.md)
+* [Add some extensions](https://github.com/FreshRSS/Extensions)
+* [Frequently asked questions](07_Frequently_Asked_Questions.md)
+* [Update your installation](08_Updating.md)
```
Frenzie: Nitpick: a newline character terminates a line, so a line is incomplete without one. ;-)

docs/en/users/08_Updating.md
```diff
@@ -0,0 +1,30 @@
+The update process depends on your installation type. 
+If you installes FreshRSS using git, then please refer to our [cli documentation](https://github.com/FreshRSS/FreshRSS/tree/dev/cli).
```
Frenzie: Installed, Git.

The then is less nice stylistically.

docs/en/users/08_Updating.md
```diff
@@ -0,0 +1,30 @@
...
+# Perform all commands below in your FreshRSS directory:
+cd /usr/share/FreshRSS
+
+# Download the latest version
+wget https://github.com/FreshRSS/FreshRSS/archive/master.zip
+
+# And unzip it
+unzip master.zip
+
+# Now overwrite all your existig files with the new ones
```
Frenzie: Existing

Frenzie (commented):
Couple of quickies from my phone.

kevinpapst:
Thanks for the first feedback. I did not check the spelling whatsoever, just wanted to get an idea on  how you want to structure the documentation. In my impression, the cli page is simply not obvious but thats mainly what I wanted to check with you. I'll fix the spellings once you let me know if the suggested way is okay with you.

docs/en/users/08_Updating.md
```diff
@@ -0,0 +1,30 @@
+The update process depends on your installation type. 
+If you installes FreshRSS using git, then please refer to our [cli documentation](https://github.com/FreshRSS/FreshRSS/tree/dev/cli).
+
+# Using the web admin panel
+
+*TODO*
+
+# Using git
+
+See [cli documentation](https://github.com/FreshRSS/FreshRSS/tree/dev/cli#install-and-updates).
```
Frenzie: Shouldn't this point to master?

Frenzie:
You mean whether update should go before or after FAQ? The only objection I have to the update document itself is that Git is essentially also shell, so I'm not too sure about the particular name.

docs/en/users/08_Updating.md
kevinpapst: So, true - it was a bit to late yesterday ;-)

kevinpapst:
Hey Frenzie, 
let me try to re-phrase my proposal:

I wanted to improve the docu for all users and gather most of the information in one place - the documentation. 

Why do I want to do that:
Even though I already contributed to FreshRSS and wrote an extension, I struggled to get my installation back up after a failed update ... just because I couldn't locate the docu. 
Currently its simply unclear where the users should look for more infos. 
There are at least 3 places currently:
- the main README
- the README in the cli directory
- the documentation
Some of the documentation is not even really linked somewhere (like the FAQ or the cli/README.md with the update infos.

My questions:
As I currently can't see a strategy on where you want the documentation to be located - what do you think where users look for more infos first. Shall we move all the infos into one place and just place some strategic links, so users can help themself? I think by doing that, you can even save time when answering issues. Just drop the proper link and thats it.

Normally the project devs don't need the docu (or only to link to it from issues) and thats why they don't see the problems for end-users. Lets make it that way: I will re-arrange the docu in another PR and you have a look if that fits your ideas as well.

Alkarex:
Hello @kevinpapst and thanks for helping 👍 
The current strategy is to have a summary in the readme files, and a full, detailed documentation on https://freshrss.github.io/FreshRSS/ (which is generated from https://github.com/FreshRSS/FreshRSS/tree/master/docs ).

I think we should keep the readme files (maybe slightly reduce some sections), and add some more links from the readme files to the main documentation. The main documentation should indeed be very much improved, and help is welcome :-)

To be fair, there are 4 links from the main readme to the CLI readme ;-)

I have already added some more links from the readme to the documentation in the /dev version https://github.com/FreshRSS/FreshRSS/blob/dev/README.md but more would be needed when the corresponding sections are actually developed in the main documentation.

image

There is also a bit more substance in the issues themselves, often with the "Documentation" tag, which would need to be copied into the main documentation. https://github.com/FreshRSS/FreshRSS/issues?q=label%3ADocumentation

Frenzie:
@kevinpapst Other than needing to look up some stuff in the French docs in the past, just going to the docs seemed mostly reasonably self-evident to me. :-)

kevinpapst:
Right on, thats what I wanted to read! 
@Alkarex maybe it was really to late, as I couldn't find the links in the main README ;-)
@Frenzie to me as well, thats why it took so long for me to find the right infos.

Ok. Then I will create a new PR based on the dev branch and look through the open issues as well. 
I'll leave this one open until then for discussion.

PR #1698, opened 2017-12-01
I18n - DE

kevinpapst:
Added some missing german translations

### DIFF
--- a/app/i18n/de/admin.php
+++ b/app/i18n/de/admin.php
@@ -146,11 +146,11 @@ return array(
 		'top_feed' => 'Top 10-Feeds',
 	),
 	'system' => array(
-		'_' => 'System configuration', // @todo translate
-		'auto-update-url' => 'Auto-update server URL', // @todo translate
-		'instance-name' => 'Instance name', // @todo translate
-		'max-categories' => 'Categories per user limit', // @todo translate
-		'max-feeds' => 'Feeds per user limit', // @todo translate
+		'_' => 'System Einstellungen',
+		'auto-update-url' => 'Auto-update URL',
+		'instance-name' => 'Dein Reader Name',
+		'max-categories' => 'Maximale Anzahl von Kategorien pro Benutzer',
+		'max-feeds' => 'Maximale Anzahl von Feeds pro Benutzer',
 		'registration' => array(
 			'help' => '0 meint, dass es kein Account Limit gibt',
 			'number' => 'Maximale Anzahl von Accounts',
--- a/app/i18n/de/admin.php
+++ b/app/i18n/de/admin.php
@@ -146,11 +146,11 @@ return array(
 		'top_feed' => 'Top 10-Feeds',
 	),
 	'system' => array(
-		'_' => 'System Einstellungen',
+		'_' => 'Systemeinstellungen',
 		'auto-update-url' => 'Auto-update URL',
 		'instance-name' => 'Dein Reader Name',
-		'max-categories' => 'Maximale Anzahl von Kategorien pro Benutzer',
-		'max-feeds' => 'Maximale Anzahl von Feeds pro Benutzer',
+		'max-categories' => 'Anzahl erlaubter Kategorien pro Benutzer',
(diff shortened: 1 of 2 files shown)


app/i18n/de/admin.php
```diff
@@ -146,11 +146,11 @@
 		'top_feed' => 'Top 10-Feeds',
 	),
 	'system' => array(
-		'_' => 'System configuration', // @todo translate
-		'auto-update-url' => 'Auto-update server URL', // @todo translate
-		'instance-name' => 'Instance name', // @todo translate
-		'max-categories' => 'Categories per user limit', // @todo translate
-		'max-feeds' => 'Feeds per user limit', // @todo translate
+		'_' => 'System Einstellungen',
```
Frenzie: Systemeinstellungen?

app/i18n/de/gen.php
```diff
@@ -143,7 +143,7 @@
 		'sharing' => 'Teilen',
 		'shortcuts' => 'Tastaturkürzel',
 		'stats' => 'Statistiken',
-		'system' => 'System configuration',// @todo translate
+		'system' => 'System Einstellungen',
```
Frenzie: Same as above.

app/i18n/de/admin.php
```diff
@@ -146,11 +146,11 @@
...
+		'auto-update-url' => 'Auto-update URL',
+		'instance-name' => 'Dein Reader Name',
+		'max-categories' => 'Maximale Anzahl von Kategorien pro Benutzer',
```
Frenzie: `von Kategorien` or `an Kategorien`?

app/i18n/de/admin.php
kevinpapst: Seems both are legit but Systemeinstellungen is more often used according to Google.

aledeg (approved):
My German is decaying so I can't judge. Anyway, thank you for your effort.

Frenzie (commented):
Thanks!

app/i18n/de/admin.php
Frenzie: On what basis do you judge both to be possible? (Not a rhetorical question.) Splitting up compounds (Zusammensetzungen) as if it were English seems contradictory to German orthography.

On double-checking in my Duden I can also only find regular (non-split) compounds like Systemanalyse, Systemkritik, Systemlehre, Systemtheorie, etc.

Frenzie: PS My copy of *Sprachdummheiten* has a discussion of Binde-s, whether to insert an -n, etc. but there's no discussion of ever writing compounds as anything other than one word. I claim no authority in the matter (whereas I'll stand on my stripes in English and Dutch) but it looks like English pollution to me. :-)

kevinpapst: I am in no way trustable with grammar and I would never ever look into a Duden for backend translations ;-) I just used Google and found both variants...

Frenzie: I think we can safely discard Google as useless unless it leads to a properly argued/referenced answer on StackExchange or some such. :-P

kevinpapst: I tend to use the "Sprachgebrauch" rule. So when anything its used often enough it goes into officially accepted language rules. But as I said - as long as there is no compiler, don't trust me with grammar rules.

Frenzie: There's no grammar involved. It's just spelling. :-P

kevinpapst: Hahaha, see thats what I am talking about

Frenzie: It's not your fault. ;-) For some reason "grammar" has caught on as the term for — well, everything, when it'd make much more sense to say something like "writing."

Grammar is what form of a word you use. Like how I pointed out somewhere earlier that the past tense of split is split, not splitted.

Most things people refer to as "grammar" are not grammar. It's primarily style and register, with a good sprinkling of punctuation and a touch of spelling.

I suppose you could also try the *Sprachgebrauch* argument on the term grammar itself, and probably with more validity to boot. :-P (Because plain Google works alright to prove that something *exists* but is otherwise useless as a corpus.)

PR #1701, opened 2017-12-02
Update panel shows latest version message as success

### LINKED ISSUE #1586: upgrade check should not warn if there is no new version available
In FreshRSS, when someone check if a new version is available, the answer is displayed as a warning (in red) even if there is no new version available.

screenshot

If there is no new version available, the version used is the last one. Hence it should be a green answer, not a red one.

kevinpapst:
This PR changes the display of the "you are running the latest version" message.

Currently it shows that message as "alert-error" which was not only irritating to me, see also #1586.
My PR changes that message to "alert-success" to indicate a positive state. 

FIXES #1586

Further enhancement idea:
Change the wording from "Blast! No update to apply" to "Congratulations, you are running the latest version" - IMO its always better to give a positive message, but thats not part of this PR.

### DIFF
--- a/app/views/update/index.phtml
+++ b/app/views/update/index.phtml
@@ -14,7 +14,21 @@
 	</p>
 
 	<?php if (!empty($this->message)) { ?>
-	<p class="alert <?php echo $this->message['status'] === 'bad' ? 'alert-error' : 'alert-warn'; ?>">
+    <?php
+        $class = 'alert-warn';
+        switch ($this->message['status']) {
+            case 'bad':
+                $class = 'alert-error';
+                break;
+            case 'latest':
+                $class = 'alert-success';
+                break;
+            default:
+                $class = 'alert-warn';
+                break;
+        }
+    ?>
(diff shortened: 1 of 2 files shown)


app/views/update/index.phtml
```diff
@@ -14,7 +14,21 @@
 	</p>
 
 	<?php if (!empty($this->message)) { ?>
-	<p class="alert <?php echo $this->message['status'] === 'bad' ? 'alert-error' : 'alert-warn'; ?>">
+    <?php
```
Frenzie: Shouldn't these be tabs?

Frenzie:
Exaltations and celebrations your liege, you are running the latest version!

/me was reminded of Civ

app/views/update/index.phtml
kevinpapst: Seems so, looked fine over here. Guess I have to change my IDE settings

kevinpapst:
Pushed an tab'ed update

Frenzie (approved):
lgtm

PR #1708, opened 2017-12-03
Show existing extensions in admin panel

kevinpapst:
Hi guys,

I thought about promoting the extensions feature more, so we get the community not only better informed about existing extension but also in the hope that new people sit down and add more.

First, how does it look like:
bildschirmfoto 2017-12-03 um 16 11 24

What it does:
Currently it fetches a json file from the extensions repo, so all extensions need to be added there manually. It will list all extensions in the admin-extension flyout and compare them with the installed ones, also showing if there is an update.

There is a lot of future improvement possible, for example:
- caching the packages file, so it doesn't get fetched on every flyout load
- we could add one-click installation
- probably fetching the data dynamically instead of hard-coding it in the packages file (but thats not necessary unless we have dozens of extensions)

This pull request is only a first proposal to start a discussion, I know that there are improvements necessary:
- whitespace vs tabs issues
- currently only english translation
- error handling missing for the file load

### DIFF
--- a/app/i18n/en/admin.php
+++ b/app/i18n/en/admin.php
@@ -112,13 +112,13 @@ return array(
 		),
 		'title' => 'Extensions',
 		'user' => 'User extensions',
-        'community' => 'Available community extensions for download',
-        'name' => 'Name',
-        'version' => 'Version',
-        'description' => 'Description',
-        'author' => 'Author',
-        'latest' => 'Installed',
-        'update' => 'Update available'
+		'community' => 'Available community extensions for download',
+		'name' => 'Name',
+		'version' => 'Version',
+		'description' => 'Description',
+		'author' => 'Author',
+		'latest' => 'Installed',
+		'update' => 'Update available'
 	),
 	'stats' => array(
 		'_' => 'Statistics',
--- a/app/i18n/en/admin.php
+++ b/app/i18n/en/admin.php
@@ -112,7 +112,7 @@ return array(
 		),
 		'title' => 'Extensions',
 		'user' => 'User extensions',
-		'community' => 'Available community extensions for download',
+		'community' => 'Available community extensions',
 		'name' => 'Name',
 		'version' => 'Version',
 		'description' => 'Description',
--- a/app/i18n/en/admin.php
+++ b/app/i18n/en/admin.php
@@ -112,6 +112,11 @@ return array(
 		),
 		'title' => 'Extensions',
 		'user' => 'User extensions',
+        'community' => 'Available community extensions for download',
+        'name' => 'Name',
+        'version' => 'Version',
+        'description' => 'Description',
+        'author' => 'Author',
 	),
 	'stats' => array(
 		'_' => 'Statistics',
--- a/app/i18n/en/admin.php
+++ b/app/i18n/en/admin.php
@@ -117,6 +117,8 @@ return array(
         'version' => 'Version',
         'description' => 'Description',
         'author' => 'Author',
+        'latest' => 'Installed',
+        'update' => 'Update available'
 	),
 	'stats' => array(
 		'_' => 'Statistics',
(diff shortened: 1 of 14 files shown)


Alkarex:
Looks good so far 👍

kevinpapst:
Ok, sounds as if you are fine with the general idea. Then I will go ahead and address the open issues.

kevinpapst:
Your whitespace check is a nightmare ;-)

From a feature perspective I say its good for 1.9, so we have still room for improvements for 2.0.
Any comments @Frenzie and @Alkarex ?

app/i18n/en/admin.php
```diff
@@ -112,6 +112,13 @@
 		),
 		'title' => 'Extensions',
 		'user' => 'User extensions',
+		'community' => 'Available community extensions for download',
```
Frenzie: Either just `Available community extensions` or `Community extensions available for download`.

app/i18n/nl/admin.php
```diff
@@ -112,6 +112,13 @@
 		),
 		'title' => 'Uitbreidingen',
 		'user' => 'Gebruikersuitbreidingen',
+		'community' => 'Available community extensions for download', // @todo translate
+		'name' => 'Name', // @todo translate
```
Frenzie: Naam
Versie
Beschrijving
Auteur
Geïnstalleerd
Update beschikbaar

app/i18n/nl/admin.php
```diff
@@ -112,6 +112,13 @@
 		),
 		'title' => 'Uitbreidingen',
 		'user' => 'Gebruikersuitbreidingen',
+		'community' => 'Available community extensions for download', // @todo translate
```
Frenzie: Gebruikersuitbreidingen beschikbaar voor download

PS You don't need to add the Dutch translations but I figured since you're on it. :-P

Frenzie:
No objections here besides the one somewhat ungrammatical string (word order is grammar :-P).

I should probably add that I prefer the shorter version `Available community extensions`.

kevinpapst:
@Frenzie thanks for the checks and suggestions

kevinpapst:
Oh and you have to merge https://github.com/FreshRSS/Extensions/pull/22 first, then I need to update the JSON URL

kevinpapst:
@Alkarex pushed the changed JSON URL ... and tweaked the version compare, so it doesn't trigger if someone has locally a higher version than the one offered by our json file

app/i18n/es/admin.php
```diff
@@ -112,6 +112,13 @@
 		),
 		'title' => 'Extensiones',
 		'user' => 'Extensiones de usuario',
+		'community' => 'Available community extensions for download', // @todo translate
```
Frenzie: Could you do a mass replace of these? :-)

app/i18n/nl/admin.php
```diff
@@ -112,6 +112,13 @@
 		),
 		'title' => 'Uitbreidingen',
 		'user' => 'Gebruikersuitbreidingen',
+		'community' => 'Gebruikersuitbreidingen beschikbaar voor download',
```
Frenzie: With the new shorter version this is now `Gebruikersuitbreidingen beschikbaar` or `Beschikbare gebruikersuitbreidingen `:-)

app/i18n/fr/admin.php
```diff
@@ -112,6 +112,13 @@
 		),
 		'title' => 'Extensions',
 		'user' => 'Extensions utilisateur',
+		'community' => 'Available community extensions for download', // @todo translate
```
Frenzie: Something like `Extensions utilisateur disponibles` @Alkarex ?

kevinpapst:
@Frenzie please have a look again
@Alkarex could you provide the french translations?

app/i18n/fr/admin.php
Alkarex: Sounds good 👍

Alkarex:
Done :-)

Frenzie (approved):

kevinpapst:
Thanks for the support guys! :-)

PR #1711, opened 2017-12-08
Check requirement in CLI script

### LINKED ISSUE #1710: Undefined class constant 'MYSQL_ATTR_INIT_COMMAND'  on Synology DSM6
I'm trying to create a cronjob on my synology system using the info provided in [this error report](https://github.com/FreshRSS/FreshRSS/issues/1126)

So, I run this cron job with user root:
php70 /volume1/web/freshrss/app/actualize_script.php > /volume1/web/freshrss/cron.txt 2>&1

And this is the outpot of the logfile:

Starting feed actualization at 2017-12-08T10:08:53+01:00
Actualize YannickV...
PHP Fatal error:  Uncaught Error: Undefined class constant 'MYSQL_ATTR_INIT_COMMAND' in /volume1/web/freshrss/lib/Minz/ModelPdo.php:65
Stack trace:
#0 /volume1/web/freshrss/app/actualize_script.php(62): Minz_ModelPdo->__construct('YannickV')
#1 {main}
  thrown in /volume1/web/freshrss/lib/Minz/ModelPdo.php on line 65
Results: 
YannickV 

The pdo_mysql extension is activated in the options. Anyone with idea's?

Thx!
Yannick

kevinpapst:
Just a small followup which fixes #1710 

As that problem occurred at least for two users, its a good idea to support users with a better error message.

Probably the choosen way to load the lib_install.php is not the best idea, as it does more than only defining some functions.


Alkarex:
Maybe we could check for more than only PDO.

kevinpapst:
I was using a foreach first checking all requirements, but I am not sure which ones are required for that script. Do you know?

Alkarex:
The question can probably be reformulated as: which ones are *not* needed for the refresh cron.
We do not need `http_referer`, but it returns true when empty, so there should be no need of doing anything special.

We can use: https://github.com/FreshRSS/FreshRSS/blob/4e174ed9dd84ce4f8de410baf6a6e0fde9160055/cli/do-install.php#L69-L81

This code should probably be moved into https://github.com/FreshRSS/FreshRSS/blob/af8960b8b3be3c78573d0319a8a537083fae2d4d/cli/_cli.php

In general, I would like to refactor `actualize_script.php` to use the same approaches than the other CLI commands. For the time being, it could be a simple `require('../cli/_cli.php');`.

kevinpapst:
I hope I understood you correctly @Alkarex

app/actualize_script.php
```diff
@@ -1,6 +1,6 @@
 <?php
-require(dirname(__FILE__) . '/../constants.php');
-require(LIB_PATH . '/lib_rss.php');	//Includes class autoloader
+require(__DIR__ . '/../constants.php');
+require(__DIR__ . '/../cli/_cli.php');
```
Alkarex: ~~`__DIR__` should not be required at all. Something like `require('../cli/_cli.php');` should work in all cases.~~
Same for the other lines.

kevinpapst: Nope, I would disagree. There is no chdir() in FreshRSS and therefor the current working path is unspecified and depends on how you call the script. That goes especially for cron jobs.

Alkarex: ~~`require(relative_path)` is relative to the calling script, and does not depend on current execution dir~~ http://php.net/manual/function.include.php

kevinpapst:
Now we have to include the installer lib for every _cli call.

kevinpapst:
Is there a reason why you use require instead of require_once in all the cli scripts?
require_once would be more error safe.

Alkarex:
I use `*_once` functions when there are multiple paths or when dependencies become a bit more complicated. So far in CLI, it is pretty straightforward and I would rather see if there is a dependency logic issue.

Alkarex:
I will revise the require/include more generally in another PR. So this one should be good enough as it is.

PR #1714, opened 2017-12-10
Added extension hook for reading modes

kevinpapst:
I am currently working on a full-fledged extension, which uses its own controllers and views.

Problem is: I cannot route the user to the page (controller/action) that I want him to see as there is currently no way of displaying additional links in the main UI.

To achieve my goal, I refactored the reading modes to be a simple array structure which gets injected into the hook. Extensions can then work on that navigation structure like that:

```
class MyExtension extends Minz_Extension
{
    public function init() {
	$this->registerHook('nav_reading_modes', array($this, 'addReadingModeToMenu'));
    }

    public function addReadingModeToMenu($menu) {
    	$request = Minz_Request::currentRequest();
	$menu[] = new FreshRSS_ReadingMode(
		'example',
		'example title',
		array_merge($request, array('c' => 'my', 'a' => 'any')),
		($request['c'] == 'my' && $request['a'] == 'any')
	);
	return $menu;
    }
}
```

Now extensions are capable of adding their own reading modes into the menu.

What do you think of this approach?

### DIFF
--- a/app/Models/ReadingMode.php
+++ b/app/Models/ReadingMode.php
@@ -107,7 +107,7 @@ class FreshRSS_ReadingMode {
 	public static function getReadingModes() {
 		$actualView = Minz_Request::actionName();
 		$defaultCtrl = Minz_Request::defaultControllerName();
-		$isDefaultCtrl = Minz_Request::controllerName() == $defaultCtrl;
+		$isDefaultCtrl = Minz_Request::controllerName() === $defaultCtrl;
 		$urlOutput = Minz_Request::currentRequest();
 
 		$readingModes = array(
--- a/app/Models/ReadingMode.php
+++ b/app/Models/ReadingMode.php
@@ -1,0 +1,136 @@
+<?php
+
+/**
+ * Manage the reading modes in FreshRSS.
+ */
+class FreshRSS_ReadingMode {
+
+	/**
+	 * @var string
+	 */
+	protected $name;
+	/**
+	 * @var string
+	 */
+	protected $title;
+	/**
+	 * @var string[]
+	 */
+	protected $urlParams;
+	/**
+	 * @var bool
+	 */
+	protected $isActive = false;
+
+	/**
+	 * ReadingMode constructor.
+	 * @param string $name
+	 * @param string $title
+	 * @param string[] $urlParams
+	 * @param bool $active
+	 */
+	public function __construct($name, $title, $urlParams, $active) {
+		$this->name = $name;
+		$this->title = $title;
+		$this->urlParams = $urlParams;
+		$this->isActive = $active;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getName() {
+		return $this->name;
+	}
+
+	/**
+	 * @param string $name
+	 * @return FreshRSS_ReadingMode
+	 */
+	public function setName($name) {
+		$this->name = $name;
+		return $this;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getTitle() {
+		return $this->title;
+	}
+
+	/**
+	 * @param string $title
+	 * @return FreshRSS_ReadingMode
+	 */
+	public function setTitle($title) {
+		$this->title = $title;
+		return $this;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getUrlParams() {
+		return $this->urlParams;
+	}
+
+	/**
+	 * @param string $urlParams
+	 * @return FreshRSS_ReadingMode
+	 */
+	public function setUrlParams($urlParams) {
+		$this->urlParams = $urlParams;
+		return $this;
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function isActive() {
+		return $this->isActive;
+	}
+
+	/**
+	 * @param bool $isActive
+	 * @return FreshRSS_ReadingMode
+	 */
+	public function setIsActive($isActive) {
+		$this->isActive = $isActive;
+		return $this;
+	}
+
+	/**
+	 * Returns the built-in reading modes.
+	 * return ReadingMode[]
+	 */
+	public static function getReadingModes() {
+		$actualView = Minz_Request::actionName();
+		$defaultCtrl = Minz_Request::defaultControllerName();
+		$isDefaultCtrl = Minz_Request::controllerName() == $defaultCtrl;
+		$urlOutput = Minz_Request::currentRequest();
+
+		$readingModes = array(
+			new FreshRSS_ReadingMode(
+				_i("view-normal"),
+				_t('index.menu.normal_view'),
+				array_merge($urlOutput, array('c' => $defaultCtrl, 'a' => 'normal')),
+				($isDefaultCtrl && $actualView === 'normal')
+			),
+			new FreshRSS_ReadingMode(
+				_i("view-global"),
+				_t('index.menu.global_view'),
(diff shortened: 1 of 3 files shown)


Frenzie:
Should all of that `$readingModes` definition stuff really go in the template? That looks like template pollution to me. But in principle I think it's clearer this way even without extensions.

kevinpapst:
True, its not the best place to keep that logic, but I was unsure where to put it otherwise. Any idea? We could create a base `FreshRSS_Controller` which extends `Minz_Controller` and then let the `FreshRSS_index_Controller` extend that new `FreshRSS_Controller`. 
Extensions could also extend this controller if they want to display the `nav_menu.phtml` partial.

Alkarex:
I have not looked in details yet, but could this code be in `FreshRSS_index_Controller` and you could override `FreshRSS_index_Controller` when needed?

app/layout/nav_menu.phtml
```diff
@@ -131,20 +131,38 @@
...
-		<a class="view_normal btn <?php echo $actual_view == 'normal'? 'active' : ''; ?>" title="<?php echo _t('index.menu.normal_view'); ?>" href="<?php echo Minz_Url::display($url_output); ?>">
-			<?php echo _i("view-normal"); ?>
-		</a>
+		<?php
+		$readingModes = array(
+			'normal' => array(
+				'name' => _i("view-normal"),
+				'title' => _t('index.menu.normal_view'),
+				'url' => array_merge($url_output, array('c' => Minz_Request::defaultControllerName(), 'a' => 'normal')),
+				'active' => (Minz_Request::controllerName() == Minz_Request::defaultControllerName() && $actual_view == 'normal'),
```
Alkarex: Please put the "fast" string comparisons first, and use triple-equal, like:

```php
'active' => $actual_view === 'normal' && Minz_Request::controllerName() === Minz_Request::defaultControllerName(),
```

and maybe the multiple `Minz_Request::controllerName() === Minz_Request::defaultControllerName()` could be factorised somehow

kevinpapst:
Updated the PR: refactored ReadingMode to a Model class. 
Keeping it in IndexController would force Extensions to extend that IndexController, which I would prefer not to do. But I could still move it there, if you disagree with the model.

Alkarex:
Sorry for the delay.
Happy New Year 2018 :-)

app/Models/ReadingMode.php
```diff
@@ -0,0 +1,136 @@
...
+	 */
+	public static function getReadingModes() {
+		$actualView = Minz_Request::actionName();
+		$defaultCtrl = Minz_Request::defaultControllerName();
+		$isDefaultCtrl = Minz_Request::controllerName() === $defaultCtrl;
+		$urlOutput = Minz_Request::currentRequest();
+
+		$readingModes = array(
+			new FreshRSS_ReadingMode(
+				_i("view-normal"),
```
Frenzie, line 115: What if we just make the argument here `"view-normal"` and then above something like

```php
$this->id = $name;
$this->name = _i($name);
```

This is in reference to #1758.

app/layout/nav_menu.phtml
```diff
@@ -131,20 +131,19 @@
...
-			<?php echo _i("view-reader"); ?>
-		</a>
+		<?php
+		$readingModes = FreshRSS_ReadingMode::getReadingModes();
+		$readingModes = Minz_ExtensionManager::callHook('nav_reading_modes', $readingModes);
+
+		/** @var FreshRSS_ReadingMode $mode */
+		foreach ($readingModes as $mode) {
+			?>
+			<a class="view_normal btn <?php if ($mode->isActive()) { echo 'active'; } ?>" title="<?php echo $mode->getTitle(); ?>" href="<?php echo Minz_Url::display($mode->getUrlParams()); ?>">
```
Frenzie, line 141: And then the ID would be part of the className here. Sorry for kind of polluting your PR but I figure it's easier to discuss this way than in #1758.

kevinpapst:
I did not follow the changes in #1755. From a quick view I would say its slightly incompatible with my changes?" 
Now extension can add views, which then would not have a shortcut. And the list of views could even be completely rewritten by the event, so the hard coded shortcuts in #1758 may result in unexpected behaviour.
I can add your changes @Frenzie then we have view IDs which can be hardwired to the shortcuts. In the unrealistic case that a view link is not existing in the list/DOM the JS would not trigger.
Is that what you suggested @Frenzie ? @aledeg any objections?

Frenzie:
@kevinpapst I'm not entirely sure what you're suggesting with hardwiring but I suppose you mean what I mean, lol. I was suggesting to dynamically restore the HTML/DOM to the way it was or at least something close to it. That way a selector like  `#nav_menu_views .view-rss` would work, although I mean that mostly for styling purposes.

I'm not the biggest fan of the JS using the DOM like in #1755 since my DOM might be different, but what I thought reasonable enough with classes could yield odd and unexpected results while using the *n*th child node. (Including not working, but that's pretty much the best case scenario.) Ideally the info from `Minz_Request::currentRequest()` would probably also be easily available to the JS. But that aside.

PS Incidentally, [here](https://stackoverflow.com/a/11654596/2470572) is the first search result for a premade function that modifies a single parameter in the query string, although just the value of `a=blabla` would likely do. Anyway, that's just a quick brainstorm.

aledeg:
The problem is the class used. Before, each view had a different CSS class (```view_normal```, ```view_global```, ```view_reading```)(See https://github.com/FreshRSS/FreshRSS/pull/1755/files#diff-99959fbf8401f92306d451127be5e01a). Now, they all have the ```view_normal``` class.
I am not saying that my fix was perfect. But it fixes the issue at the moment.

What should be done is restoring a way to have a different CSS class for each of the buttons. This way we can revert the fix to its initial setting.

kevinpapst:
Ok, now I am slightly confused^^ Let me adjust the code and submit a PR hopefully tomorrow.

I am concerned about the changes in #1758 as they can fail in rare circumstances. The hardcoded index used in the get() call can be wrong if an extension manipulates the available reading modes in any way. So we are concerned about the same point but probably for different reasons ;-)
Therefor we need to re-add the class or an index to the links, which were previously available and I accidentaly removed them  (I guess because they were just not used before @aledeg introduced his changes).

app/layout/nav_menu.phtml
aledeg: That was the idea behind the class ```view_normal```, ```view_global```, and ```view_reading```. We already had that. I think we just need a small adjustment on how you generate the reading modes.
@kevinpapst

aledeg:
No hard feelings, we all make mistakes. I've recently rewrote a PR after Alkarex mentioned something I didn't see :)

kevinpapst:
Hahaha, I blame everything on you @aledeg ... nobody would have noticed without your changes LOL
I ping you with my fixes tomorrow for review.

Frenzie:
> What should be done is restoring a way to have a different CSS class for each of the buttons. This way we can revert the fix to its initial setting.

@kevinpapst This is what I meant to convey in my comments, btw. I figured the translation ID could also serve well doing double duty as a className.

aledeg:
@kevinpapst @Frenzie I think we were all taking about the same thing :)

aledeg:
@kevinpapst any news regarding the class fix?

kevinpapst:
Ups, totally forgot about it, thanks for the reminder. I'am on it right now

kevinpapst:
@aledeg see #1773
@Frenzie thanks, just did it as you proposed

aledeg:
@kevinpapst great!

PR #1715, opened 2017-12-10
Use relative require / include everywhere

Alkarex:
http://php.net/manual/function.include.php
https://github.com/FreshRSS/FreshRSS/pull/1711#discussion_r155922574
https://github.com/FreshRSS/FreshRSS/pull/1711#issuecomment-350581350


Alkarex:
Damn, it appears that did not understand the documentation correctly, regarding `include('./something.php');` :-(

Alkarex:
The more precise part is in http://php.net/manual/ini.core.php#ini.include-path

PR #1717, opened 2017-12-11
Use __DIR__ for relative include and require

Alkarex:
For uniformity, and to avoid having PHP searching in include_path.
http://php.net/manual/function.include.php
https://github.com/FreshRSS/FreshRSS/pull/1715
https://github.com/FreshRSS/FreshRSS/pull/1711#issuecomment-350581350

### DIFF
--- a/cli/_cli.php
+++ b/cli/_cli.php
@@ -3,7 +3,7 @@ if (php_sapi_name() !== 'cli') {
 	die('FreshRSS error: This PHP script may only be invoked from command line!');
 }
 
-require(dirname(__FILE__) . '/../constants.php');
+require(__DIR__ . '/../constants.php');
 require(LIB_PATH . '/lib_rss.php');
 require(LIB_PATH . '/lib_install.php');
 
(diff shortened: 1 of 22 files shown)


kevinpapst:
The only two picky comments I have: 
- require and include are language constructs and  brackets should IMO be avoided
- there are clear use cases for both require and require_once ... and from what I see the use case here is always require_once

But apart from that ;-) Looks good to me

aledeg (approved):
We can also use ```DIRECTORY_SEPARATOR``` to replace ```/``` in paths. But it's not mandatory. I guess PHP works fine on Windows with ```/```.

Alkarex:
Thanks for the feedbacks :-) Quick notes:
* I do not mind parentheses or not, but if we change, we should probably do it consistently in the rest of the code base as well (another PR).
* For the `*_once`, it is also a habit of not using them if I do not need them (and rather see an error and fix it). Years ago, they were known to be quite slower (random googling https://www.roojs.com/index.php/View/76/require_once_one_optimization_too_many.html ) but I do not whether that still has any relevance with PHP 5.3.8+ (the current minimal version we are supporting).
* Indeed, I have been using `/` on Windows without any problem.

PR #1720, opened 2017-12-12
FreshRSS 1.9.0

Alkarex:
Dear all,
Let's finalise a few issues https://github.com/FreshRSS/FreshRSS/milestones/1.9.0 and then release 1.9.0 in a few days.
A main feature (rework update mechanism) that was originally planned for 1.9 is not ready yet (pushed to 1.10), but there is already enough substance to justify a bump to 1.9.0:
https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md
So please give the /dev branch a spin :-)


Frenzie:
I suppose I should hold off on an approval until I've actually tried the latest commits but w00t! :-)

kevinpapst:
Impressive list ... next time please squash my commits ;-)
I'd say #1494 and #1562 can be closed.

Frenzie:
I don't really have time to investigate in detail right now but it looks like this broke my auto-refresh cronjob. (I ran a manual refresh and that took care of the lack of new feeds since December 12.)

Edit: also I suppose we changed something about the unique ID? My feed http://feeds.feedburner.com/TCF-TV5MONDE-CIEP thought everything was new.

Frenzie:
Update: found some time to look at the log:

```
Starting feed actualization at 2017-12-14T12:15:01+01:00
PHP Fatal error:  Uncaught Error: Call to undefined function performRequirementCheck() in /var/www/html/feeds/app/actualize_script.php:27
Stack trace:
#0 {main}
  thrown in /var/www/html/feeds/app/actualize_script.php on line 27
Results:
```

kevinpapst:
Are you sure you updated fully to dev branch? The actualize_script.php includes cli/_cli.php which defines the function. Check:
- https://github.com/FreshRSS/FreshRSS/blob/dev/app/actualize_script.php#L2
- https://github.com/FreshRSS/FreshRSS/blob/dev/app/actualize_script.php#L27
- https://github.com/FreshRSS/FreshRSS/blob/dev/cli/_cli.php#L52

Its working flawless on my installation.

constants.php
```diff
@@ -1,11 +1,16 @@
 <?php
-define('FRESHRSS_VERSION', '1.8.0');
+define('FRESHRSS_VERSION', '1.8.1-dev');
```
Frenzie: Should be 1.9.0, right?

Alkarex: Yes yes, I do that at the last minute before release :-)

Frenzie:
> Are you sure you updated fully to dev branch?

Of course I'm not sure; the entire point is not to do anything silly like that but to use the regular built-in update mechanisms. Otherwise I might as well not test. Imo anyway. :-P

In any event, it looks like `/app` updated but `/cli` didn't. This is probably my fault (it's owned by a different user than www-data) but I would consider the fact that I wasn't warned about it in any way a bug.

Alkarex:
@Frenzie https://github.com/FreshRSS/FreshRSS/pull/1671#issuecomment-349464343

marienfressinaud:
I have same problem as @Frenzie but don't have time to investigate. Nevertheless, I don't think current system deploy `cli` folder (see https://github.com/FreshRSS/update.freshrss.org/blob/master/scripts/update_util.php#L154), it should be related ;)

Frenzie:
Yes, I fixed the permissions (assuming they were in fact broken; I didn't double-check if the permissions were quite right but the ownership felt off regardless) and found that it didn't resolve the problem. The files simply aren't updated like @marienfressinaud says.

kevinpapst:
@Frenzie Silly hm? What exactly? Using the update function which is known to have problems with wrong file permissions or my comment which was meant to be helpful for investigating the problem? Its for comments like yours, that people leave oss projects.

Alkarex:
Let's come back to a friendly tone :-)
Indeed, the update system is one of the next priorities to be revised. Especially for /dev, it is not a reliable update method. git or manual ZIP override are safer.

Frenzie:
@kevinpapst I apologize if I offended you, but I just meant that it would be silly *for me* to treat my real test the same as my regular dev instance. On the latter I would switch to the relevant git branch and execute something like `git pull` or `git pull -r` and I'd therefore be sure that all files were exactly as they should be. On my real test making sure that all files are how I expect them to be is not step one (save for catastrophic failure) but step three at the earliest.

Back on topic, does your known issue include a successful "Update completed!" message?

Frenzie:
@Alkarex I only use it to update from stable to RC. I ignore dev on my real instance.

Alkarex:
Sounds good! And to close this little parenthesis, let's try to avoid sentences, which risk being perceived as aggressive :-)

Frenzie:
Like @marienfressinaud mentioned I can confirm that the issue is that `/cli` simply isn't deployed regardless of permissions, which apparently failed to manifest into any concrete problems over the past year.

To answer my own question about the update, no error message is given even if the system clearly failed to overwrite certain files due to insufficient permissions. To test I made sure the www-data group couldn't write to the app/Controllers subdir and checked based on file modified time. Ideally I suppose it should avoid trying at all, but I'm ultimately more concerned about the lack of an indication of failure. But that's a separate issue.

If I understand correctly the lack of `/cli` deployment problem can be fixed on the update server?

marienfressinaud:
Yes, and it is my own responsability, ahah. I'll take time this weekend to do it ;)

Frenzie (approved):
All good except for the cli thing.

kevinpapst:
@Frenzie thanks for your feedback! I might have overreacted on a clumsy(?) wording ... if that sounds aggressive I blame it on google translate ;-)

No, my update that failed ended in a 500. So its two different problems (as you already found out). I'll try to submit a patch for the permission check, but I wouldn't incorporate that in 1.9. Lets wait for the upgrade improvements in 1.10.

kevinpapst:
Please check and retest #1731 - either my install is broken or a nasty 500 bug slipped into the subscription management.

marienfressinaud:
`./cli` deployment should work now, can you give it a try @Frenzie ? Just update FreshRSS and check `cli` directory exists

Alkarex:
Thanks all for your good work! We have a [new version 1.9.0](https://github.com/FreshRSS/FreshRSS/releases/tag/1.9.0) :-)

Alkarex:
@Frenzie , @kevinpapst , @aledeg I do not know whether you use Mastodon / Diaspora* / Friendica or any similar decentralised social network, but FYI https://f.a80.fr/display/bda90b95135a36ddef0c11c202219992 , https://f.a80.fr/search?tag=freshrss

Frenzie:
No, but I do have a centralized Twitter account if that isn't unwelcome news. :-P

Edit: wait, I do have a Diaspora account. Not sure if I'd really get anybody reading it there though.

Alkarex:
@Frenzie Then you can spread the news there as well :-) Let me know if you ever open an account with a more standard protocol somewhere.

marienfressinaud:
I have tweeted with FreshRSS account: https://twitter.com/FreshRSS/status/942670499905196032

Frenzie:
@marienfressinaud Everything about `/cli` deployment seems to be working fine now. :-)

Issue #1705, opened 2017-12-03
Proposal: Move freshrss lock file into web dir


sleeksorrow:
Hello there,

my proposal is to not write the .freshrss.lock file into the system temp dir, but find a cosy place inside the web directory tree for it.

Reason:
A friend of mine had the issue where the lock file got stuck somehow and feed updating was just giving the message "Feed already being actualized". Since this installation is on shared hosting where the system temp dir is not reachable for the customer, it was not possible to delete it. As soon as I changed the lock file path in app/Models/Feeds.php:367 from TMP_PATH to CACHE_PATH, FreshRSS was working all fine, and when it should get stuck again in the future, it's removable via FTP.

It would be awesome if you could implement some change similar to this so my friend doesn't need to merge this change after every update. Thanks!

Alkarex:
Hello,
It is easier to change the location of your temp path in https://github.com/FreshRSS/FreshRSS/blob/master/constants.php#L27

lock files are supposed to be outdated after one hour. If the problem persisted for more than one hour, then there might be another issue (such as access rights changed).

For the record, we should probably split this file in two: one for the parts that are changing more often (e.g. version number), and one for the rest.

sleeksorrow:
The issue was more than a day, so yes, there is some problem behind that is causing this, and access rights can easily be the culprit. But since the system temp dir is out of user control on this installation I could not do any debugging.

But I don't want to dig into this, as this is not a fault of FreshRSS, it was just the occation. My proposal is way more general. Basically it is:

Better keep every file needed for FreshRSS in the FreshRSS directory itself, because on shared hosting any other location could be inaccessible and people cannot debug and fix anything there.

So yes, this reasoning in mind, moving the whole temp directory using the constant would be a simple yet powerful approach. It would be perfect if you could implement this and e.g. create a temp directory within FreshRSS tree. This way every shared hosting user would benefit.

Alkarex:
Yes, I understand your reasoning @sleeksorrow . On the other hand, I would also like to avoid writing to physical disk when possible (use cases when e.g. a SD card is used like on a Raspberry Pi), and `/tmp` is often an in-memory location (tmpfs). (By the way there is still a bit of work to do there to make it easier, like the location of the logs.)
I will try to make it easier to switch between `/tmp/` and `./FreshRSS/data/tmp/`

sleeksorrow:
[...]

PR #1725, opened 2017-12-16
Customisable constants.local.php

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1562
https://github.com/FreshRSS/FreshRSS/issues/1607
https://github.com/FreshRSS/FreshRSS/pull/1656
https://github.com/FreshRSS/FreshRSS/issues/1705
https://github.com/FreshRSS/FreshRSS/pull/1712

### DIFF
--- a/constants.php
+++ b/constants.php
@@ -1,35 +1,64 @@
 <?php
+//NB: Do not edit; use ./constants.local.php instead.
+
+//<Not customisable>
 define('FRESHRSS_VERSION', '1.8.1-dev');
 define('FRESHRSS_WEBSITE', 'https://freshrss.org');
 define('FRESHRSS_WIKI', 'https://freshrss.github.io/FreshRSS/');
 
-define('FRESHRSS_USERAGENT', 'FreshRSS/' . FRESHRSS_VERSION . ' (' . PHP_OS . '; ' . FRESHRSS_WEBSITE . ')');
+define('FRESHRSS_PATH', __DIR__);
+define('PUBLIC_PATH', FRESHRSS_PATH . '/p');
+define('PUBLIC_TO_INDEX_PATH', '/i');
+define('INDEX_PATH', PUBLIC_PATH . PUBLIC_TO_INDEX_PATH);
+define('PUBLIC_RELATIVE', '..');
+define('LIB_PATH', FRESHRSS_PATH . '/lib');
+define('APP_PATH', FRESHRSS_PATH . '/app');
+define('EXTENSIONS_PATH', FRESHRSS_PATH . '/extensions');
+//</Not customisable>
 
-// PHP text output compression http://php.net/ob_gzhandler (better to do it at Web server level)
-define('PHP_COMPRESSION', false);
+if (file_exists(__DIR__ . '/constants.local.php')) {
+	//Include custom / local settings:
+	include(__DIR__ . '/constants.local.php');
+}
 
-// Maximum log file size in Bytes, before it will be divided by two
-define('MAX_LOG_SIZE', 1048576);
-
-// Constantes de chemins
-define('FRESHRSS_PATH', dirname(__FILE__));
-
-	define('PUBLIC_PATH', FRESHRSS_PATH . '/p');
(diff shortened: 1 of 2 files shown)


constants.php
```diff
@@ -1,35 +1,64 @@
...
+//</Not customisable>
 
-// Constantes de chemins
-define('FRESHRSS_PATH', dirname(__FILE__));
-
-	define('PUBLIC_PATH', FRESHRSS_PATH . '/p');
-		define('PUBLIC_TO_INDEX_PATH', '/i');
-		define('INDEX_PATH', PUBLIC_PATH . PUBLIC_TO_INDEX_PATH);
-		define('PUBLIC_RELATIVE', '..');
+if (file_exists(__DIR__ . '/constants.local.php')) {
```
Frenzie: Would it be clearer to call it something like `constants.persistent.php` instead?

constants.php
```diff
@@ -1,35 +1,64 @@
...
+	//Include custom / local settings:
+	include(__DIR__ . '/constants.local.php');
+}
 
+if (!defined('FRESHRSS_USERAGENT')) {
```
Frenzie: Do you think this whole block would be more elegant with some kind of array/for loop construction?

constants.php
Alkarex: I hesitated a bit on the naming. I wanted to make it clear that it was something for local administrators and not from a default FreshRSS configuration. "local" is a term used in Linux FHS as well as Windows for related needs.

constants.php
Alkarex: Could you please explain a bit more what you have in mind?

Alkarex: I could propose something like:

```php
function safe_define($name, $value) {
	if (!defined($name) {
		return define($name, $value);
	}
}

safe_define('MAX_LOG_SIZE', 1048576);
safe_define('DATA_PATH', FRESHRSS_PATH . '/data');
...
```

constants.php
Frenzie: But my FreshRSS itself is already local, conceptually speaking. :-P

How does Windows use it?

constants.php
Frenzie: Yes, that's fine too, probably better even. I'm not saying I object per se, although I'm not overly fond of how the code looks right now, just that I'd like it to be a conscious choice. :-)

I meant simply something like

```php
$defines = [
    'MAX_LOG_SIZE' => 1234,
    'DATA_PATH' => '/data',
];

foreach ($defines as $key => $value) {
  safe_define($key, $value)
}
```

Edit: added that I prefer your proposal.

Alkarex: Let's go for it, then :-) https://github.com/FreshRSS/FreshRSS/pull/1725/commits/5dd2906eb2e9a19a1f3277880bc37a259d3cb7ec

constants.php
Alkarex: I was referring to https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard , where *local* means *specific to this host*, which is a bit what I wanted to achieve there.
In Windows, there is *%LOCALAPPDATA%* for a similar usage.

Frenzie: I meant conceptually it's like a script in `/usr/local` even if I personally happen to store it in `/var/someplace`.

I suppose the distinction is somewhat similar — I assume that in Windows you have settings provided over the network or from a system default, and in %LOCALAPPDATA% a Windows program might store things like recently opened files and preferences.

But I'm not sure if it clarifies or obscures. If you already know that your user profile stores stuff in `~/.local` or `%USERPROFILE%\Local Settings` it's fine, but if you don't it's somewhat mystifying. That's why I suggested an imo more self-explanatory word like `persistent`.

Frenzie (approved):
My comment about the filename is not an objection to merging.

constants.php
Alkarex: While I see what you mean with *persistent*, I am unsure it will be well understood as something users can edit, and it does not seem to be so commonly used. In any case, it would be nice to add some documentation somewhere :-)

PR #1726, opened 2017-12-16
Add message after log rotation

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1712
https://github.com/FreshRSS/FreshRSS/issues/1562

### DIFF
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -98,6 +98,7 @@ class Minz_Log {
 				rewind($fp);
 				ftruncate($fp, 0);
 				fwrite($fp, $content ? $content : '');
+				fwrite($fp, sprintf("[%s] [notice] --- Log rotate.\n", date('r')));
 				fflush($fp);
 				flock($fp, LOCK_UN);
 			} else {


lib/Minz/Log.php
```diff
@@ -98,6 +98,7 @@ protected static function ensureMaxLogSize($file_name) {
 				rewind($fp);
 				ftruncate($fp, 0);
 				fwrite($fp, $content ? $content : '');
+				fwrite($fp, '[' . date('r') . "] [notice] --- Log rotate.\n";);
```
aledeg: I am not a big fan of string concatenation. I'd like ```sprintf``` better. My 2¢

Frenzie: Is PHP one of the languages where it makes a difference in speed? In this case it'd make no difference in legibility but if you end up with `%s%s%s` then concat can still be clearer.

(Ideally I suppose you'd have something more like Python's string.format instead.)

Alkarex: Updated to `sprintf()` for @aledeg :-)

aledeg: @Frenzie @Alkarex it makes a difference on how the memory is allocated. String concatenation use more memory than sprintf.

Frenzie: @aledeg In that case I too prefer sprintf, although without seeing benchmarks it might be premature optimization. ;-D

aledeg: @Frenzie I know that from one of the PHP core dev. He made a speak about optimization and that was one of it.

PR #1704, opened 2017-12-03
[doc] Extensions: translate various sections from French

Frenzie:
See https://github.com/FreshRSS/FreshRSS/pull/1697#discussion_r154290857

The file is slightly larger than I anticipated, so I won't be able to finish it in one "breakfast translation". Here are this morning's results. I can keep adding to this PR or we can split it into multiple PRs; whichever you prefer.

The quality is first draft, which probably suffices but it means that I might've unintentionally copied French sentence structures or ways of expression.

PR #1729, opened 2017-12-16
Minz Dispatcher Controllers path

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1704

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * I18n
 	* Improved German [#1698](https://github.com/FreshRSS/FreshRSS/pull/1698)
 * Extensions
+	* Breaking change: uppercase `./Controllers/` directory [#1729](https://github.com/FreshRSS/FreshRSS/pull/1729)
 	* Show existing extensions in admin panel [#1708](https://github.com/FreshRSS/FreshRSS/pull/1708)
 	* New function `$entry->_hash($hex)` for extensions that change the content of entries [#1707](https://github.com/FreshRSS/FreshRSS/pull/1707)
 * Misc.
--- a/lib/Minz/Dispatcher.php
+++ b/lib/Minz/Dispatcher.php
@@ -10,7 +10,6 @@
  * C'est un singleton
  */
 class Minz_Dispatcher {
-	const CONTROLLERS_PATH_NAME = '/Controllers';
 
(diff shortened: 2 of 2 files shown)


Alkarex:
* The constant was apparently never used
* I have not found any path using this code (maybe some extensions, but I have checked the listed ones, and they do not use it)

https://github.com/FreshRSS/FreshRSS/blob/4e174ed9dd84ce4f8de410baf6a6e0fde9160055/lib/Minz/Dispatcher.php#L152

* FreshRSS uses an uppercase `./Controllers/` directory
* Other places in the code are uppercased 

https://github.com/FreshRSS/FreshRSS/blob/d91a92434f516a9e25bf0dca608a0221094d0489/lib/Minz/Error.php#L24

Alkarex:
@kevinpapst Did you have an example using the lowercase `./controllers/` ?

kevinpapst:
I was trying to write a full fledged extension and the controllers didn't load, exactly due to the code part that you changed now. It did not work until I used lowercase "controllers" like here: https://github.com/kevinpapst/freshrss-wall/tree/master/xExtension-Wall

The code that you changed is very likely ONLY used by extensions. I have not seen an extension using controllers by now, but surely there are (probably in Mariens private playground only) otherwise the code wouldn't be there and work ;-)

Its only executed if you manually register a controller like that:
https://github.com/kevinpapst/freshrss-wall/blob/master/xExtension-Wall/extension.php#L18

You probably will only break my demo extension, which is no big deal, but you better write it in bold letters into the release notes that there is a backward compatibility break coming in.

Edit: Its a BC break and I personally like lowercase more, but I would still vote for uppercase for consistency within FreshRSS

Alkarex:
Thanks for the feedback. Yes, maybe @marienfressinaud , otherwise I believe you are likely the first to use this feature :-)

marienfressinaud (approved):
It seems ok to me but I don't understand why it was working until now

Alkarex:
It was working because FreshRSS and extensions are using two different if/else paths:
https://github.com/FreshRSS/FreshRSS/blob/4e174ed9dd84ce4f8de410baf6a6e0fde9160055/lib/Minz/Dispatcher.php#L77-L82
and the lowercase was in `self::loadController(...);`

PR #1731, opened 2017-12-17
fixed bug when adding a category and feed at the same time

kevinpapst:
Fixed "Call to a member function id() on a non-object in /freshrss/app/Controllers/feedController.php on line 57"

That happened when you want to add a feed to a on-the-fly created category.

Because `$catDAO->addCategory()` returns an int and not an instance of `FreshRSS_Category`.

See:
https://github.com/FreshRSS/FreshRSS/blob/dev/app/Controllers/feedController.php#L40
https://github.com/FreshRSS/FreshRSS/blob/dev/app/Models/CategoryDAO.php#L16

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -42,7 +42,14 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		if ($cat == null) {
 			$catDAO->checkDefault();
 		}
-		$cat_id = $cat == null ? FreshRSS_CategoryDAO::DEFAULTCATEGORYID : $cat->id();
+
+		if ($cat === null) {
+			$cat_id = FreshRSS_CategoryDAO::DEFAULTCATEGORYID;
+		} else if ($cat instanceof FreshRSS_Category) {
+			$cat_id = $cat->id();
+		} else {
+			$cat_id = $cat;
+		}
 
 		$feed = new FreshRSS_Feed($url);	//Throws FreshRSS_BadUrl_Exception
 		$feed->_httpAuth($http_auth);
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -149,7 +149,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 				// User want to create a new category, new_category parameter
 				// must exist
 				$new_cat = Minz_Request::param('new_category');
-				$new_cat_name = isset($new_cat['name']) && !empty(trim($new_cat['name'])) ? $new_cat['name'] : '';
+				$new_cat_name = isset($new_cat['name']) && !empty(trim($new_cat['name'])) ? trim($new_cat['name']) : '';
 			}
 
 			// HTTP information are useful if feed is protected behind a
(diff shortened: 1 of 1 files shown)


Alkarex:
Good catch!

May I suggest a slightly different fix?

```patch
diff --git a/app/Controllers/feedController.php b/app/Controllers/feedController.php
index 883f7af0..2cb39acd 100755
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -33,12 +33,13 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
                $catDAO = new FreshRSS_CategoryDAO();

                $cat = null;
+               if ($new_cat_name != '') {
+                       $new_cat_id = $catDAO->addCategory(array('name' => $new_cat_name));
+                       $cat_id = $new_cat_id > 0 ? $new_cat_id : $cat_id;
+               }
                if ($cat_id > 0) {
                        $cat = $catDAO->searchById($cat_id);
                }
-               if ($cat == null && $new_cat_name != '') {
-                       $cat = $catDAO->addCategory(array('name' => $new_cat_name));
-               }
                if ($cat == null) {
                        $catDAO->checkDefault();
                }
```

kevinpapst:
Edit: looks good, I'll try

kevinpapst:
Next try, please squash!

kevinpapst:
I slipped in another little fix, as you could add a feed that only had spaces as name, like "   ". 
That worked mostly but led to some really strange frontend issues.

app/Controllers/feedController.php
```diff
@@ -136,7 +149,7 @@ public function addAction() {
 				// User want to create a new category, new_category parameter
 				// must exist
 				$new_cat = Minz_Request::param('new_category');
-				$new_cat_name = isset($new_cat['name']) ? $new_cat['name'] : '';
+				$new_cat_name = isset($new_cat['name']) && !empty(trim($new_cat['name'])) ? $new_cat['name'] : '';
```
Alkarex: Good idea. But it would probably be good to apply trim() in all cases, to remove whitespaces at start and end.

kevinpapst: Hahaha, I was thinking about it, but wasn't unsure, alright give me 3 minutes

kevinpapst:
next try

app/Controllers/feedController.php
```diff
@@ -136,7 +149,7 @@ public function addAction() {
 				// User want to create a new category, new_category parameter
 				// must exist
 				$new_cat = Minz_Request::param('new_category');
-				$new_cat_name = isset($new_cat['name']) ? $new_cat['name'] : '';
+				$new_cat_name = isset($new_cat['name']) && !empty(trim($new_cat['name'])) ? trim($new_cat['name']) : '';
```
Alkarex: `empty(function())` is only accepted since PHP 5.5 http://php.net/manual/function.empty.php

I think the following should be sufficient:

```php
$new_cat_name = isset($new_cat['name']) ? trim($new_cat['name']) : '';
```

kevinpapst:
damned you are right, would have been easier if I just reported the bug and you applied the fix ;-)

PR #1733, opened 2017-12-17
Fix favicon for open_basedir

kevinpapst:
PHP error log says:
`PHP message: PHP Warning:  curl_setopt(): CURLOPT_FOLLOWLOCATION cannot be activated when an open_basedir is set in /freshrss/lib/favicons.php on line 39`

That is already known, as stated in favicons.php but not handled.
I applied a possible fix, but I do not know how to trigger that code to test my fix, please advice.

I also fixed a unknown variable problem, inside an if block that is probably not executed very often...

### DIFF
--- a/lib/favicons.php
+++ b/lib/favicons.php
@@ -36,7 +36,7 @@ function downloadHttp(&$url, $curlOptions = array()) {
 			CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 			CURLOPT_MAXREDIRS => 10,
 		));
-	if (!empty(ini_get('open_basedir'))) {
+	if (ini_get('open_basedir') == '') { // see PHP bug 65646
 		curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
 	}
 	if (defined('CURLOPT_ENCODING')) {
--- a/lib/favicons.php
+++ b/lib/favicons.php
@@ -37,8 +37,8 @@ function downloadHttp(&$url, $curlOptions = array()) {
 			CURLOPT_MAXREDIRS => 10,
 		));
 	if (!empty(ini_get('open_basedir'))) {
-        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);    //Keep option separated for open_basedir bug
-    }
+		curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+	}
 	if (defined('CURLOPT_ENCODING')) {
 		curl_setopt($ch, CURLOPT_ENCODING, '');	//Enable all encodings
 	}
--- a/lib/favicons.php
+++ b/lib/favicons.php
@@ -24,7 +24,7 @@ function isImgMime($content) {
 function downloadHttp(&$url, $curlOptions = array()) {
 	syslog(LOG_INFO, 'FreshRSS Favicon GET ' . $url);
 	if (substr($url, 0, 2) === '//') {
-		$url = 'https:' . $favicon;
+		$url = 'https:' . $url;
 	}
 	if ($url == '' || filter_var($url, FILTER_VALIDATE_URL) === false) {
 		return '';
@@ -36,7 +36,9 @@ function downloadHttp(&$url, $curlOptions = array()) {
 			CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 			CURLOPT_MAXREDIRS => 10,
 		));
-	curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);	//Keep option separated for open_basedir bug
+	if (!empty(ini_get('open_basedir'))) {
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);    //Keep option separated for open_basedir bug
+    }
 	if (defined('CURLOPT_ENCODING')) {
 		curl_setopt($ch, CURLOPT_ENCODING, '');	//Enable all encodings
 	}


Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/pull/1657 , https://github.com/FreshRSS/FreshRSS/issues/1655

lib/favicons.php
```diff
@@ -36,7 +36,9 @@ function downloadHttp(&$url, $curlOptions = array()) {
 			CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 			CURLOPT_MAXREDIRS => 10,
 		));
-	curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);	//Keep option separated for open_basedir bug
+	if (!empty(ini_get('open_basedir'))) {
```
Alkarex: I am not sure whether I actually want to fix this warning. It is only a warning; it still works when no redirection is needed. It is supposed to be fixed in PHP 5.6.0+ https://bugs.php.net/bug.php?id=65646

The thing is that with this bug, the favicons might not be fetched correctly because HTTP redirections will not be followed, and this cannot easily be addressed from FreshRSS (otherwise we need to implement our own HTTP redirection following mechanism).

If we only want to get rid of the warning, a simple `@` could do the job.

Otherwise, please change to `ini_get('open_basedir') == ''` because `empty(function())` is only allowed since PHP 5.5 (and is slower anyway) http://php.net/manual/function.empty.php

For reference, this is what SimplePie is doing to support even older PHP versions:
https://github.com/simplepie/simplepie/blob/0edf39018709c0d66cfa6d9b972907bb92461efb/library/SimplePie/File.php#L112-L116

lib/favicons.php
```diff
@@ -24,7 +24,7 @@ function isImgMime($content) {
 function downloadHttp(&$url, $curlOptions = array()) {
 	syslog(LOG_INFO, 'FreshRSS Favicon GET ' . $url);
 	if (substr($url, 0, 2) === '//') {
-		$url = 'https:' . $favicon;
+		$url = 'https:' . $url;
```
Alkarex, line 27: Nice catch 👍

Alkarex:
On what version of PHP and OS did you experience this issue?

lib/favicons.php
kevinpapst: What about only using the shutup operator @ instead?
We could write a freshrss log entry as well, which could potentially be a benefit for some admins.
But at least I would make sure that this message does not show up in the global error log.

Alkarex: Let's go for the `ini_get('open_basedir') == ''` version, which is more explicit.
Please add a comment that this is for PHP 5.6.0- (PHP bug 65646)

kevinpapst:
Both LTS versions
```
Linux Ubuntu-1310-saucy-64-minimal 3.13.0-57-generic #95-Ubuntu SMP Fri Jun 19 09:28:15 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```
and
```
PHP 5.5.9-1ubuntu4.22 (cli) (built: Aug  4 2017 19:40:28)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
```
with curl
```
curl 7.35.0 (x86_64-pc-linux-gnu) libcurl/7.35.0 OpenSSL/1.0.1f zlib/1.2.8 libidn/1.28 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp smtp smtps telnet tftp
Features: AsynchDNS GSS-Negotiate IDN IPv6 Largefile NTLM NTLM_WB SSL libz TLS-SRP
```

lib/favicons.php
Alkarex: It is when `open_basedir` is empty, that the curl option can be set, not the other way around

Alkarex:
Huh, Ubuntu 13.10 Saucy is not LTS :P
But Ubuntu 14.04 LTS Trusty has PHP < 5.6 indeed https://distrowatch.com/table.php?distribution=ubuntu

kevinpapst:
Ah, sorry, wording issue. My definition of LTS: it still gets regular updates. Its almost 2018, thats more than 4 years and I would consider that a long time ;-)

Nevertheless, I have public demos running on that server, thats why I use open_basedir in the first place (I know that open_basedir is considered a broken feature).

lib/favicons.php
kevinpapst: brr, thats not my day, totally correct

Alkarex:
You may want to do the same for https://github.com/Alkarex/FreshRSS/blob/9c9411e911f8a21f5b4c9a87ff12c30084fa040f/app/Models/Feed.php#L495

PubSubHubbub can be tested with e.g. http://push-pub.appspot.com/

For the record, Ubuntu 13.10 Saucy is EOL since July 2014 ;-) http://fridge.ubuntu.com/2014/06/20/ubuntu-13-10-saucy-salamander-reaches-end-of-life-on-july-17-2014/
While Ubuntu 14.04 LTS Trusty continues to be supported until April 2019.

kevinpapst:
Just found out that it is actually 14.04 / trusty. Seems uname -a cannot be trusted.
I have no clue what PubSubHubbub is, well yes I read what it does, but I never used it actively or from a developer pov. What does it have to do with the favicon?
Edit: understand, you mean the other linked resource.

Alkarex:
PubSubHubbub is a push technology that several feeds are using (so that FreshRSS receives news instantly). On your server side, it only requires that your FreshRSS endpoint `./api/pshb.php` is publicly routable (it must be able to receive an HTTP POST).
This piece of code uses the same construction and CURLOPT_FOLLOWLOCATION than favicon.

Alkarex:
P.S. One can check whether PubSubHubbub  is enabled for a feed in the feed's options:
image

Alkarex:
Done in https://github.com/FreshRSS/FreshRSS/pull/1734

PR #1734, opened 2017-12-17
Remove warning for CURLOPT_FOLLOWLOCATION with open_basedir

Alkarex:
For PHP 5.6.0- http://www.php.net/ChangeLog-5.php#5.6.0
https://bugs.php.net/bug.php?id=65646
https://github.com/FreshRSS/FreshRSS/pull/1733
https://github.com/FreshRSS/FreshRSS/pull/1657
https://github.com/FreshRSS/FreshRSS/issues/1655

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -491,7 +491,9 @@ class FreshRSS_Feed extends Minz_Model {
 					CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 					CURLOPT_MAXREDIRS => 10,
 				));
-			curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);	//Keep option separated for open_basedir bug
+			if (version_compare(PHP_VERSION, '5.6.0') >= 0 || ini_get('open_basedir') == '') {
+				curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);	//Keep option separated for open_basedir PHP bug 65646
+			}
 			if (defined('CURLOPT_ENCODING')) {
 				curl_setopt($ch, CURLOPT_ENCODING, '');	//Enable all encodings
 			}
--- a/lib/favicons.php
+++ b/lib/favicons.php
@@ -36,8 +36,8 @@ function downloadHttp(&$url, $curlOptions = array()) {
 			CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 			CURLOPT_MAXREDIRS => 10,
 		));
-	if (ini_get('open_basedir') == '') { // see PHP bug 65646
-		curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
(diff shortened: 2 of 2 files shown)


Alkarex:
I have just changed a bit my mind. Please check again, now with a `version_compare(PHP_VERSION, '5.6.0') >= 0` in order for newer versions of PHP to actually use `open_basedir` if they want without breaking cURL redirections.

kevinpapst:
What about adding a lib_php-compatibility.php file to collect all the nasty little pieces of code? 
Currently its only that one, but I am pretty sure there are more hidden gems in other places.
But besides that the logic looks good. If PHP_VERSION is 5.6.0 or higher it will always get executed.
How do you know that it is fixed in 5.6.0? I only see that it was reported for 5.5.3.

Alkarex:
> How do you know that it is fixed in 5.6.0?

From the changelog http://www.php.net/ChangeLog-5.php#5.6.0

kevinpapst:
Hahaha, that was too easy.

Frenzie:
Btw, with 5.3 and 5.4 severely EOL will we still be actively taking <5.5 into account going forward?

Alkarex:
@Frenzie We are slowly dropping old PHP versions when there is a clear need. See e.g. the move to PHP 5.3.8+ in FreshRSS 1.8.0 https://github.com/FreshRSS/FreshRSS/pull/1618#issuecomment-323708633

I am checking the usage statistics https://w3techs.com/technologies/details/pl-php/all/all as well as the status for a few (conservative but popular) Linux distributions:
* https://distrowatch.com/table.php?distribution=ubuntu (LTS branches)
* https://distrowatch.com/table.php?distribution=debian
* https://distrowatch.com/table.php?distribution=centos

While those old PHP versions are EOL upstream, they are still actively maintained with security patches by the various distributions.

Finally, I also look at the status at some large Web hosting services such as http://90plan.ovh.net/infos/

Reference to an older discussion https://github.com/FreshRSS/FreshRSS/issues/789#issuecomment-73878076

For the record, I run currently run my FreshRSS on PHP 7.1.

Frenzie:
I recently switched to 7.0 (from 5.6 which I'd been using for a few years) alongside upgrading from Jessie to Stretch. 

> While those old PHP versions are EOL upstream, they are still actively maintained with security patches by the various distributions.

* Debian Squeeze (6) LTS support ended in February 2016. That had 5.3.
* Debian Wheezy (7) is on 5.4, support for which will end in a few months (March 2018).
* Debian Jessie (8, now oldstable) is on 5.6.
* Debian Stretch (9, stable) is on 7.0.

Ubuntu's should be similar. That's actually part of why I brought it up. :-)

Now CentOS — yes, that one is generally quite remarkable in how long support lasts. Do they have an easy online packages browser like Debian?

PR #1739, opened 2017-12-18
Fix a warning when retrieving the list of entries

aledeg:
When retrieving the list of entries when the context was 'all' or 'starred', there was the following warning:

> Warning: count(): Parameter must be an array or an object that implements Countable in /home/alexis/FreshRSS/app/Controllers/indexController.php on line 206

I fixed that by changing how the array is tested.


Alkarex:
I wonder why I have not seen this warning yet. Is it a PHP 7.2 thing maybe?

Alkarex (approved):

aledeg:
I think it's PHP 7.2. I only have that on my local environment. My server runs PHP 7.0 and I cannot see that warning.

Frenzie:
Same warning showed up in Elkarte on 7.2. ;-)

Issue #1738, opened 2017-12-18
[Feature request] Pause feeds


Jucgshu:
Hi guys,

I tried to look for this feature on the issue log, and I'm a little surprised that no one asked for that yet?

I know we can prevent a feed from being displayed on the main stream, but I don't think we can pause a feed (meaning don't display it on the main stream, nor refresh it on feeds sync, nor take it into account it in the categories). Or am I completely wrong?

As always, kudos for your work guys.

-j

aledeg:
In the feed configuration, you can configure if you want to retrieve new articles from the feed. Check the ```Do not automatically refresh more often than``` option in the ```Archiving``` section. Then select ```∞```

For the 2 other topics, you're right.

Jucgshu:
Oww right, will try that.

Thanks @aledeg

Alkarex:
Implemented by @aledeg in https://github.com/FreshRSS/FreshRSS/pull/1750 :-)

PR #1750, opened 2017-12-22
Add mute strategy configuration [WIP]

aledeg:
There is a new configuration setting in the feed configuration which allow to mute it.
There is 4 supported modes:
  - none → nothing is muted, it's the default and previous mode.
  - refresh → feed is not refreshed anymore but is still visible in the interface.
  - display → feed is not displayed anymore but is still refreshed.
  - all → feed is not refreshed nor displayed anymore.

For more info, see #1738

----

This is far from perfect. There is probably room for improvement and/or fine tuning, but the general idea is here.
Here is a list of things I think are missing:
 - [x] ~~an icon to show the mute strategy on feeds in the configuration page. If you can provide me with an icon, that would be nice.~~ I've changed the color of the icon instead
 - [x] ~~a way to add the new field to the database when upgrading FRSS. I couldn't find where I should put that.~~ I reuse existing fields to add the feature, see https://github.com/FreshRSS/FreshRSS/pull/1750#issuecomment-353962134
    - [x] ~~mysql.~~
    - [x] ~~sqlite.~~
    - [x] ~~postgres.~~
 - [x] check if the api is filtered
 - [x] check if the RSS feed is filtered
 - [x] some thorough testing.


aledeg:
@Jucgshu if you could have a look at that. I think this will solve you're problem.

Jucgshu:
Hi @aledeg 

Just switched to your branch to test it. Looks like I have an error when trying to move the feeds to 'all' muted: ```SQL error updateFeed: syntax error```

Tried with https://nextcloud.com/blogfeed/ and https://www.myopenrouter.com/rss

aledeg:
@Jucgshu You need to add a field in the feed table to hold the configuration. Sorry I forgot to mention that. Do you need the syntax ?

Jucgshu:
Yes please @aledeg  ;-)

aledeg:
alter table feed add column muteStrategy int(11) not null default 0;

Don't forget to add your prefix before feed.

Alkarex:
@aledeg We need to auto-add this new field. Here is an example for the entry table: https://github.com/FreshRSS/FreshRSS/blob/f632a346269100d6a93bef318ffa66c97f16f6fa/app/Models/EntryDAO.php#L21

aledeg:
@Alkarex I've added the auto-update of the column

app/Models/Category.php
```diff
@@ -45,7 +45,7 @@ public function nbNotRead() {
 		return $this->nbNotRead;
 	}
 	public function feeds() {
-		if ($this->feeds === null) {
+		if (null === $this->feeds) {
```
Alkarex: Any advantage of writing the comparison in reverse?

aledeg: On this one not really, but when you're working with variables, you're sure you are not assigning a variable by mistake. For instance:

    // This is always true
    if ($var = 'hello')...

    // This fails
    if ('hello' = $var)...

I found that it's a good habit to have tests in reverse. My 2¢

app/Models/FeedDAO.php
```diff
@@ -421,4 +423,48 @@ public static function daoToFeed($listDAO, $catID = null) {
...
+		} catch (Exception $e) {
+			Minz_Log::error('FreshRSS_FeedDAO::addColumn error: ' . $e->getMessage());
+			if ($hasTransaction) {
+				$this->bd->rollBack();
+			}
+		}
+		return false;
+	}
+
+	public function autoUpdateDb($errorInfo) {
```
Alkarex: We need to add the code for SQLite and PostgreSQL too, in FeedDAOSQLite.php and FeedDAOPGSQL.php.
Examples for the entry table:
* https://github.com/FreshRSS/FreshRSS/blob/f632a346269100d6a93bef318ffa66c97f16f6fa/app/Models/EntryDAOSQLite.php#L9-L26
* https://github.com/FreshRSS/FreshRSS/blob/f632a346269100d6a93bef318ffa66c97f16f6fa/app/Models/EntryDAOPGSQL.php

Since the code is similar for all tables, we could consider factorising it somehow (can wait to another PR later).

As it is the first new field after we have introduced PostgreSQL, I do not yet have the full example for it.

app/i18n/cz/sub.php
```diff
@@ -2,13 +2,13 @@
 
 return array(
 	'api' => array(
-		'documentation' => 'Copy the following URL to use it within an external tool.',// TODO
-		'title' => 'API',// TODO
+		'documentation' => 'Copy the following URL to use it within an external tool.', // TODO
```
Alkarex: If possible, please avoid changes in parts of the code not related to the PR. (It was probably automatic, but you might have a way of avoiding it, don't you?)

aledeg: Sure

aledeg: Done

app/Models/FeedDAO.php
aledeg: Will do

aledeg:
@Jucgshu @Alkarex If you could test more thoroughly what I've done, that would be great :)

aledeg:
@Alkarex I've changed the greader file. I've added my filter and it works fine. I think the modified query doesn't work with SQLite because of the prefix. Why don't we use directly the DAO there on the first place ?

Alkarex:
@aledeg As far as I can see, it does not look like you have changed anything that should influence the prefix, have you? The API has a few requests that are specific to it. Yes, they could probably be moved into one of the DAO. I cannot remember whether there were any other reasons.

Alkarex:
Don't you think it could be enough to have a display/do-not-display attribute? Indeed, the "refresh" part seems redundant with the existing refresh option, which can be set to infinite (same effect than the new mute refresh). As far as I can see, the only advantage is to be able to remember the old refresh setting, and if this is a desirable feature, we could use some of the available negative range of feed.ttl for that. It would also simplify the queries and the UI.

Alkarex:
I am of course a bit late to come with suggestions after one implementation is mostly done, but what would you think of using a specific category for muted feeds? It would allow drag&drop to that category and provide a clear visualisation of the feeds in that state. Here also, I believe it would simplify the UI (and the queries).

aledeg:
For the prefix thingy, I was talking about the original query. I think we are lucky that nobody used SQLite and the API at the same time :)

I'll give you some insight on what I thought was a good idea when I read the ticket opened by @Jucgshu . The main idea is to be able to quickly turn the feed on and off while keeping the settings unchanged. The main advantage of that is that you have a more fine-grained solution on how you want your feeds to react.

I have that idea to keep the refresh unchanged while hiding the feed to be able to get new entries and manipulate them before reaching the database. For instance, having some extensions rerouting the original entry to some dummy feeds. Let say you have few feeds with a lot of different topics and you're interested in only one topic, you could just silent all feeds for display and get the related entries in that dummy feed.

I don't know if I am clear enough. Let me know if you need more information.

I am not a big fan of the specific category for muted feeds for the same reason I mentioned earlier. It won't keep track of the previous settings. My idea is that if you want to mute a feed, it's temporary and you don't want to set it back to the previous settings manually. A little bit like when you mute the television.

Alkarex:
I cannot see a problem with SQLite on the current /dev, and I have just tested (quickly) with success :-)

I like the general idea. I would though like to investigate whether there could be ways to implement it in a lighter way, both in terms of concepts (for the user) and of DB (for the performances in particular).

## Architecture

We already have a `feed.priority` column, which is indexed (because it is used in some of the most performance-critical requests). And we have a `feed.ttl` column, which is not indexed (not used during expensive queries). They both have room to be used for more features, and I would like to avoid overlapping / redundant parameters.

### Feed priority
As far as I can see, the `feed.priority` column can be used for the purpose of display/do-not-display, at least. It is also already conveyed to the UI with a `data-priority` HTML attribute. We could agree on a scale with more values, such as:

* -10 (new): mute display (not shown in list of feeds);
* 0 (existing): not shown in main feed;
* 10 (existing): normal.

The UI could upgrade the current checkbox to e.g. a drop-down or radio-boxes:
image

If needed, the priority attribute could allow more values, e.g. to decide what to do with existing favourites.

### Feed TTL
The `feed.ttl` is already controlling when feeds should be refreshed (if they should at all).
We can have a shortcut in the UI to easily toggle "mute refresh", without needing a new field in DB.
We currently use `-1` for disabling refresh (same as "mute refresh" concept), `-2` to use the global settings, and values higher than 900 to control how often to refresh. We can exploit more negative values to remember the previous settings when using the "toggle mute refresh" feature, such as:

* ...
* -1200 (new): muted / before was 1200
* -900 (new): muted / before was 900
* -20 (new): muted / before was -2
* -2 (existing): Use global settings (I think this should be moved to use `0` instead, if we can)
* -1 (existing): Manually set to never refresh (in that case, the toggle "mute refresh" should be unavailable)
* 0: undefined
* 900 (existing): refresh rate every 15min
* 1200 (existing):  refresh rate every 20min
* ...

## UI
* [ ] UI to see in which state the feeds are (check in various browsers)
* [ ] Get feedback on whether "muted" is the right wording / is understood. Maybe "archived" would be a more generic concept.
* [ ] Consider a UI to toggle view (show/hide muted feeds), following the same logic than we already have for e.g. favourites (could be postponed to another PR)
image

## Tests
* [x] ~~Test MySQL column auto-create (if we need to create a new column at all)~~
* [x] Test MySQL impacted requests
* [x] Measure MySQL SELECT performance impact on large bases on modest systems / EXPLAIN
* [x] ~~Test SQLite column auto-create (if we need to create a new column at all)~~
* [ ] Test SQLite impacted requests
* [ ] Quick SQLite performance test
* [x] ~~Test PostgreSQL column auto-create (if we need to create a new column at all)~~
* [ ] Test PostgreSQL impacted requests
* [ ] Quick PostgreSQL performance test
* [x] ~~Check `FreshRSS_DatabaseDAO` (if we need to create a new column at all)~~

Looong comment, but I look forward to read what you think.

aledeg:
@Alkarex Thank you for your comment. You've seen what I didn't. You're right about reusing what we have already. I completly rewrote the PR to follow what you suggested.
Suggestions welcomed

Alkarex:
@aledeg That was impressively fast. And thanks for your flexibility :-)
I have only spotted a couple of minor points so far. I will have a look in more details during the day.

app/Controllers/feedController.php
```diff
@@ -24,6 +24,7 @@ public function firstAction() {
 				Minz_Error::error(403);
 			}
 		}
+		$this->updateTTL();
```
Alkarex, line 27: We need to call that from a couple of additional places. I suggest:
* In `install.php` for people going through it after an upgrade (a bit redundant, but the earlier the better);
* In `FreshRSS_FeedDAO->listFeedsOrderUpdate()` to cover the CLI and the JavaScript manual refresh;

app/Models/EntryDAO.php
```diff
@@ -727,15 +727,13 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
 		switch ($type) {
 		case 'a':
 			$where .= 'f.priority > 0 ';
```
Alkarex: You could use one of your new constants, such as `>= PRIORITY_MAIN_STREAM `

app/Models/EntryDAO.php
```diff
@@ -752,8 +750,9 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
 
 		return array(array_merge($values, $searchValues),
 			'SELECT e.id FROM `' . $this->prefix . 'entry` e '
-			. ($joinFeed ? 'INNER JOIN `' . $this->prefix . 'feed` f ON e.id_feed=f.id ' : '')
-			. 'WHERE ' . $where
+			. 'INNER JOIN `' . $this->prefix . 'feed` f ON e.id_feed = f.id '
+			. 'WHERE f.priority >= ' . FreshRSS_Feed::PRIORITY_NORMAL . ' '
```
Alkarex: * We already have a priority filter for the `'a'` case. I suggest to add an explicit priority for the other cases as well, instead of having it there.
* Maybe we should already make the priority a parameter of this `sqlListWhere()` function. I expect we will soon have cases when we want to search in all feeds, also muted/archived.

app/Models/EntryDAO.php
```diff
@@ -873,12 +872,28 @@ public function countNotRead($minPriority = null) {
...
+          WHERE e1.is_favorite = 1
+            AND f1.priority >= :priority_normal
+         UNION
+         SELECT COUNT(e2.id) AS c
+              , 2 AS o
+           FROM `{$this->prefix}entry` AS e2
+           JOIN `{$this->prefix}feed` AS f2 ON e2.id_feed = f2.id
+          WHERE e2.is_favorite = 1
+            AND e2.is_read = 0
+            AND f2.priority >= :priority_normal
```
Alkarex, line 892: I am a bit unsure of what we should do regarding favourites in muted/archived feeds. What do you think?
The same goes for export.

aledeg: I think that if you don't want to see a feed, you probably don't want to see the favorites either. That's why I've added that filter here.

app/Controllers/subscriptionController.php
```diff
@@ -87,17 +87,23 @@ public function feedAction() {
...
 				'website' => checkUrl(Minz_Request::param('website', '')),
 				'url' => checkUrl(Minz_Request::param('url', '')),
 				'category' => $cat,
 				'pathEntries' => Minz_Request::param('path_entries', ''),
-				'priority' => intval(Minz_Request::param('priority', 0)),
+				'priority' => intval(Minz_Request::param('priority', FreshRSS_Feed::PRIORITY_MAIN_STREAM)),
 				'httpAuth' => $httpAuth,
 				'keep_history' => intval(Minz_Request::param('keep_history', -2)),
-				'ttl' => intval(Minz_Request::param('ttl', -2)),
+				'ttl' => $ttl * ($mute ? -1 : 1),
```
Alkarex, line 108: Does that work for default TTL values (0)?

app/Models/FeedDAO.php
```diff
@@ -421,4 +423,22 @@ public static function daoToFeed($listDAO, $catID = null) {
...
+UPDATE `{$this->prefix}feed`
+   SET ttl = :ttl_default
+ WHERE ttl = -2
+SQL;
+		$stm = $this->bd->prepare($sql);
+		$stm->execute(array(':ttl_default' => FreshRSS_Feed::TTL_DEFAULT));
+
+		$sql = <<<SQL
+UPDATE `{$this->prefix}feed`
+   SET ttl = -900
```
Alkarex: When we "unmute", will that feed then be set to `+900`? (That seems too fast)
Maybe we could keep this `-1`? This relates to another comment higher up: isn't there a need to have a value for muted when it was zero. (I have not completely though about the logic yet)

app/i18n/zh-cn/sub.php
```diff
@@ -32,19 +32,25 @@
...
 		'empty' => '此源为空。请确认它是否正常更新。',
 		'error' => '此源遇到一些问题。请在确认是否能正常访问后重试。',
-		'in_main_stream' => '在首页中显示',
 		'informations' => '信息',
 		'keep_history' => '至少保存的文章数',
 		'moved_category_deleted' => '删除分类时，其中的 RSS 源会自动归类到 <em>%s</em>',
+		'mute' => 'mute', // TODO
 		'no_selected' => '未选择 RSS 源。',
 		'number_entries' => '%d 篇文章',
+		'priority' => array(
```
Alkarex, line 41: I do not know whether it is better, just a suggestion: we could have a `'_' => 'Visibility'` (or `'Priority'`) instead of the standalone key a bit lower (just to keep things grouped)

app/views/subscription/index.phtml
```diff
@@ -130,7 +130,7 @@
 			    data-feed-id="<?php echo $feed->id(); ?>"
 			    dropzone="move">
 				<a class="configure open-slider" href="<?php echo _url('subscription', 'feed', 'id', $feed->id()); ?>"><?php echo _i('configure'); ?></a>
-				<img class="favicon" src="<?php echo $feed->favicon(); ?>" alt="✇" /> <?php echo $feed->name(); ?>
+				<img class="favicon <?php if ($feed->mute()) {echo 'mute';}?>" src="<?php echo $feed->favicon(); ?>" alt="✇" /> <?php echo $feed->name(); ?>
```
Alkarex: I think we should output the same parameters here, and also in:
* https://github.com/FreshRSS/FreshRSS/blob/63a8f5e167e4bf17f76eb6e8baa9b75814c4512d/app/layout/aside_feed.phtml#L56
* https://github.com/FreshRSS/FreshRSS/blob/63a8f5e167e4bf17f76eb6e8baa9b75814c4512d/app/views/index/global.phtml#L40

See the `data-priority` attribute

app/Controllers/subscriptionController.php
aledeg: yes it does. The code force the use of the default TTL when used with default. The only drawback is that you need to set it back to default when you unmute

Alkarex:
This looks quite good 👍 I am done with my comments (before testing). I will try to test a bit later tonight.

app/Models/FeedDAO.php
aledeg: It behave the way you understood. It might be too fast, but I can change the value for something more suitable. I don't like having the ```-1``` option because it means introducing a different behavior for that value. We already have a different behavior for the default value.

app/i18n/zh-cn/sub.php
aledeg: Sure, why not

app/Models/FeedDAO.php
Alkarex: ~~Ok, then let's change it to `TTL_DEFAULT` then.~~ Sorry, wrong option, with the current values.

app/i18n/zh-cn/sub.php
aledeg: Done

app/views/subscription/index.phtml
aledeg: I am not sure what you mean here. I'll need some more explanations

app/Models/EntryDAO.php
aledeg: I've done the first item. I think we could do the second one later when the question arise.

app/Models/FeedDAO.php
aledeg: I've changed the value to ```3600```. 20 minutes is not that fast when refreshing. Especially when you're going from muted to unmuted.

app/Controllers/feedController.php
aledeg: I've done it for the CLI and JavaScript, but after looking into it, I think it's a lot of work to do it during installation. I am not sure it's that important.

app/views/subscription/index.phtml
Alkarex: We have 3 places with a similar construct, which I suggest to keep uniform.
A bit higher up, there is a line 128 which currently contains `<li class="item feed<?php echo $error, $empty; ?>"`.
I suggest to replace it with something along the lines of:
```php
<li class="item feed<?php echo $error, $empty, $feed->mute() ? ' mute' : ''; ?>" data-priority="<?php echo $feed->priority(); ?>"`
```
So that we keep all the feed properties at the same element level.
(And adjust your CSS rule below accordingly)

And do the same (i.e. add the mute info) to the two other referenced lines.

app/Models/FeedDAO.php
```diff
@@ -251,16 +252,18 @@ public function arrayFeedCategoryNames() {	//For API
...
-		if (!($stm && $stm->execute())) {
-			$sql2 = 'ALTER TABLE `' . $this->prefix . 'feed` ADD COLUMN ttl INT NOT NULL DEFAULT -2';	//v0.7.3
+		if (!($stm && $stm->execute(array(':priority_normal' => FreshRSS_Feed::PRIORITY_NORMAL)))) {
+			$sql2 = 'ALTER TABLE `' . $this->prefix . 'feed` ADD COLUMN ttl INT NOT NULL DEFAULT ' . FreshRSS_Feed::TTL_DEFAULT;	//v0.7.3
 			$stm = $this->bd->prepare($sql2);
 			$stm->execute();
 			$stm = $this->bd->prepare($sql);
 			$stm->execute();
 		}
+		$this->updateTTL();
```
Alkarex: I think it needs to be placed a bit higher up, before the `SELECT` otherwise we risk returning old values

app/Models/EntryDAO.php
```diff
@@ -726,22 +726,24 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
...
-			$joinFeed = true;
 			break;
 		case 'f':
+			$where .= 'f.priority >= ' . FreshRSS_Feed::PRIORITY_NORMAL . ' ';
 			$where .= 'e.id_feed=? ';
 			$values[] = intval($id);
 			break;
 		case 'A':
+			$where .= 'f.priority >= ' . FreshRSS_Feed::PRIORITY_NORMAL . ' ';
 			$where .= '1=1 ';
```
Alkarex: The `1=1 ` can now be removed. It was only used to simplify the logic and allow an ` AND` in all cases

aledeg: done

Alkarex: Oh, I cannot see the line removed in the git diff. Have you committed already?

app/Models/FeedDAO.php
aledeg: done

app/Models/EntryDAO.php
aledeg: a minute ago :D

app/views/subscription/index.phtml
aledeg: Ok, I see what you mean.

app/views/helpers/feed/update.phtml
```diff
@@ -125,6 +126,10 @@
 						echo '<option value="' . intval($this->feed->ttl()) . '" selected="selected">' . intval($this->feed->ttl()) . 's</option>';
```
Alkarex, line 126: Isn't there a little bug there?
When the feed is muted, and we edit it, then no TTL option will be selected (the browser will pick the first one in the list). When we save, it looks like we will override the TTL with a wrong value.

One option could be to show the old value.

app/Models/Feed.php
```diff
@@ -104,9 +111,12 @@ public function keepHistory() {
 	public function ttl() {
 		return $this->ttl;
 	}
+	public function mute() {
+		return $this->mute;
```
Alkarex, line 115: Do we need a variable, or is it always `return $this->ttl < self::TTL_DEFAULT;` ?
I have become confused along the way: I do not have a strong feeling either way, but I thought the `muted` state would be a kind of meta property to quickly change TTL and priority at once.

app/views/helpers/feed/update.phtml
aledeg: That how it's done. The TTL is an absolute value of the field in the DB. It is splitted in the ttl and mute fields. So if you have ```-900``` in db, you'll have ```900``` in ttl and ```true``` in mute.

app/Models/Feed.php
aledeg: It's always ```return $this->ttl < self::TTL_DEFAULT;```.
Muting the refresh won't change the display. I can force that behavior the same way I did for the ttl. So the mute is not a meta property at the moment.

app/views/subscription/index.phtml
aledeg: done

aledeg:
@Jucgshu There is a brand new version of this feature. Please try it and give us some feedback. You can drop the ```muteStrategy``` column as it's not needed anymore.

Jucgshu:
@aledeg Just give it a try, and this looks really cool ! Tried it with SQLite successfully. 

So the feeds disappear correctly when selecting the 'archived' option, and the category count is updated accordingly. Same goes with the 'display in category' option, where the feeds count doesn't change and the feeds are hidden in the feeds list.

So far, everything looks good. This is really impressive what you did in such a short period of time. Thank you so much !

-j

aledeg:
@Jucgshu My pleasure. It was fun to write!

app/Models/Feed.php
```diff
@@ -219,8 +228,8 @@ public function _keepHistory($value) {
 	public function _ttl($value) {
 		$value = intval($value);
 		$value = min($value, 100000000);
-		$value = max($value, -2);
-		$this->ttl = $value;
+		$this->ttl = abs($value);
```
Alkarex, line 231: I am not sure to understand what the logic is when selecting ∞ / -1 for the TTL. Could you please explain?

Alkarex:
Sorry for the delay.
@aledeg Do you have a suggestion of what to do with archived/muted feeds, when it comes to purging old articles? Without being refreshed, the number of articles will automatically decrease until reaching the minimum number of articles to keep.
P.S.: Happy New Year 2018! :-)

aledeg:
I think that this is the expected behavior. If you mute a feed, in the end you'll end up with only the favorites.
Happy New Year too!

app/Models/Feed.php
aledeg: There is none. I've just forgot to remove that option from the drop-down :)

app/Models/EntryDAO.php
```diff
@@ -726,23 +726,24 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
...
 		case 'c':
-			$where .= 'f.category=? ';
+			$where .= 'f.priority >= ' . FreshRSS_Feed::PRIORITY_NORMAL . ' ';
+			$where .= 'AND f.category=? ';
 			$values[] = intval($id);
-			$joinFeed = true;
 			break;
 		case 'f':
-			$where .= 'e.id_feed=? ';
+			$where .= 'f.priority >= ' . FreshRSS_Feed::PRIORITY_NORMAL . ' ';
```
Alkarex: When we query a feed by ID, we should not filter by feed priority, don't you think?

aledeg: That's true. I removed it.

Alkarex:
I have just tested (execution time and EXPLAIN) the main modified SQL requests on my little Atom server using MySQL 5.7.20-0ubuntu, with 275k articles, and it looks fine.

app/Controllers/feedController.php
Alkarex: I think it would be good to add an `updateTTL()` in `FreshRSS_subscription_Controller::firstAction()`.
In my case, I am seeing:
image

aledeg: Done, but not tested.

Alkarex:
I think we can merge soon, to get some more testing in the /dev branch.

My final comment is about the visualisation of the muted feeds (currently a greyed favicon): it does not do anything on favicons already in back&white, and makes some favicons unrecognisable (the ones which happen to use multiple colours giving the same grey). There are multiple approaches, but I am thinking of adding a Unicode character like *pause* ⏸ or *stop* ⏹
image

p/api/greader.php
```diff
@@ -261,8 +261,8 @@ function subscriptionList() {
 
 	$pdo = new MyPDO();
 	$stm = $pdo->prepare('SELECT f.id, f.name, f.url, f.website, c.id as c_id, c.name as c_name FROM `%_feed` f
-		INNER JOIN `%_category` c ON c.id = f.category');
-	$stm->execute();
+		INNER JOIN `%_category` c ON c.id = f.category AND f.priority >= :priority_normal');
```
Alkarex, line 264: Ups. Typo here. It is a `WHERE`, not an `AND`.

aledeg: It's on purpose. The inner join is done on the category and the priority

aledeg:
@Alkarex I've changed the rendition of the feed when muted

Alkarex:
My API is not working anymore (empty). I have not checked why yet. I will come back later.

aledeg:
@Alkarex I've found why the API wasn't working anymore. I've fixed it.

p/themes/base-theme/template.css
```diff
@@ -71,6 +71,9 @@ img.favicon {
 	width: 16px;
 	vertical-align: middle;
 }
+.feed.mute::after {
+	content: '🔇';
```
Alkarex: Good idea with the mute sign. The CSS rule does not seem to work so well:

image

Maybe something like:

```css
.feed.mute > .item-title::after {
	content: ' 🔇';
}
```

aledeg: Good idea, but it doesn't work in the subscription list :(

aledeg: It doesn't work either if there is an ellipsis

aledeg: I've put it before instead of after. It works everywhere.

PR #1751, opened 2017-12-22
Update travis config

aledeg:
I've enabled tests on PHP 7.2 since it's the latest stable version.
I've also installed phpcs from composer since PEAR is not available for all versions of PHP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,18 +5,20 @@ php:
   - '5.6'
   - '7.0'
   - '7.1'
+  - '7.2'
   - hhvm
   - nightly
 
 install:
   # newest version without https://github.com/squizlabs/PHP_CodeSniffer/pull/1404
-  - pear install PHP_CodeSniffer-3.0.0RC4
+  - composer global require squizlabs/php_codesniffer "<=3.0.0RC4"
 
 script:
   - phpenv rehash
   - |
     if [[ $VALIDATE_STANDARD == yes ]]; then
-      phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
+      COMPOSER_BIN=$(composer global config --absolute bin-dir)
+      $COMPOSER_BIN/phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
     fi
   - |
     if [[ $CHECK_TRANSLATION == yes ]]; then
@@ -29,18 +31,11 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: "5.3"
-      dist: precise
-    - php: "7.1"
-      env: CHECK_TRANSLATION=yes VALIDATE_STANDARD=no
-  allow_failures:
     # PHP 5.3 only runs on Ubuntu 12.04 (precise), not 14.04 (trusty)
     - php: "5.3"
       dist: precise
-    - php: "5.4"
-    - php: "5.5"
-    - php: "5.6"
-    - php: "7.0"
-    - php: hhvm
-    - php: nightly
+    - php: "7.2"
+      env: CHECK_TRANSLATION=yes VALIDATE_STANDARD=no
+  allow_failures:
     - env: CHECK_TRANSLATION=yes VALIDATE_STANDARD=no
+    - dist: precise
--- a/app/Controllers/extensionController.php
+++ b/app/Controllers/extensionController.php
@@ -240,7 +240,7 @@ class FreshRSS_extension_Controller extends Minz_ActionController {
 			$res = recursive_unlink($ext->getPath());
 			if ($res) {
 				Minz_Request::good(_t('feedback.extensions.removed', $ext_name),
-				                  $url_redirect);
+				                   $url_redirect);
 			} else {
 				Minz_Request::bad(_t('feedback.extensions.cannot_delete', $ext_name),
 				                  $url_redirect);


aledeg:
Tests are even better now since they are running on every version of PHP \o/

p/api/pshb.php
```diff
@@ -128,8 +128,8 @@
 	try {
 		Minz_Session::_param('currentUser', $username);
 		Minz_Configuration::register('user',
-		                             join_path(USERS_PATH, $username, 'config.php'),
-		                             join_path(FRESHRSS_PATH, 'config-user.default.php'));
+									 join_path(USERS_PATH, $username, 'config.php'),
+									 join_path(FRESHRSS_PATH, 'config-user.default.php'));
```
Alkarex: Is the extra whitespace on purpose?

aledeg: Nope, phpcbf made the correction automatically. It's probably because we have some kind of weird rules for checking the syntax.
Maybe we should move to defacto standards like PSR

Frenzie:
Since this seems to solve all phpcs-related failures we can remove all of the allowed failures except for the translation check.

Frenzie:
Btw, I was just looking into the 91.2% complete translation rating for Dutch, but except for one or two terms (like "Project website" and "Subscription tools") it's all stuff that just so happens to be the same as in English.

aledeg:
@Frenzie  So you need to add them in ```cli/i18n/ignore``` folder. You can look what I have done for French and English.

Frenzie:
@aledeg Thanks, I hadn't noticed. In my defense I didn't really look into it yet. :-P Done in #1752.

aledeg:
@Frenzie I've updated the ```.travis.yml``` file.

Frenzie:
Nice. :-)

app/Controllers/authController.php
```diff
@@ -54,10 +54,10 @@ public function indexAction() {
 
 			if ($ok) {
 				Minz_Request::good(_t('feedback.conf.updated'),
-				                   array('c' => 'auth', 'a' => 'index'));
+								   array('c' => 'auth', 'a' => 'index'));
```
Alkarex: I do not normally align my own code like that - this comes from @marienfressinaud - but I believe the intention is to get a stable alignment between all the lines. If I understand correctly, this change will only work for a precise number of spaces per tab, and not for other settings.
Maybe @marienfressinaud has a comment.
For the record, I normally use a single tab indentation in that case, but I do not really mind (as long as we keep tabs for general indentation 😃).

aledeg: I'll have a look. I personally prefer spaces :)

marienfressinaud: I used to prefer tab for indentation and spaces for alignment (it was the case here), but today I use spaces only to avoid mixing both of them.

I think best option is to keep tab for indentation indeed (or we would have to change all the code!) but if we want to avoid mix with spaces, I would prefer something like that:

```php
Minz_Request::good(
	_t('feedback.conf.updated'),
	array('c' => 'auth', 'a' => 'index')
);
```

Frenzie: That's the way I prefer it (all tabs, alignment optional and perhaps discouraged). Spaces for indentation just messes up ASCII art alignment in editors if you try to set a custom width and is therefore much worse than proper "mixing" (there's no mixing) of tabs and spaces. Everything looking the same everywhere is a bug, not a feature.

Anyway, that's my two cents. It's all fine though. ;-)

kevinpapst: Thats one way to look at it, another one is http://www.php-fig.org/psr/psr-2/

aledeg: I'd love to implement that one

aledeg:
I fixed the conflicts and updated the formatting we talked about a bit earlier.

app/Controllers/authController.php
aledeg: @kevinpapst just for fun, I've run phpcs with PSR2 and it's terrible. Every single file needs to be modified. Most of the errors are about spacing and can be fixed automatically, but some are really bad. Those are the one related to namespacing, class names for instance.
We have a long way to get there. Eventually, we will need to do that in order to get the latest php features by dropping support for older version. I can't wait to be there.

kevinpapst: Hm, not sure if I wanna go down that road. Don't get me wrong, I would totally prefer PSR-2 ... but has changing the complete codebase any benefits besides our personal feelings (especially when the other devs prefer tabs)? 
At least my IDE doesn't fu*k up all the whitespaces since I added the .editorconfig file. Its was a pain, as FreshRSS is the only project I work on that uses tabs.
Dropping support for older PHP version indeed has a lot of benefits, but a lot of downsides for users as well. So both topics should be carefully discussed, but likely in another place!?

Frenzie: In the event of spaces, let the record show that I vote for 13 spaces per indentation. :angel: 

(That'll teach you how four spaces feel.)

Ftr, I've been called "worse than luacheck" for enforcing four spaces. :-P The project standard is the project standard; there's typically little reason to change that kind of thing.

Visually I typically prefer 2 spaces. In most editors there's the option to display tab as *n* spaces, but not to display, e.g., 4 spaces as 1 through 3 spaces. And in the rare case that there is, *unlike* proper indentation + spaces it messes up ASCII art, while proper tab indentation + spaces like in FreshRSS works fine with ASCII art. (Although I generally find ASCII art more distracting than helpful, regardless of indentation.) Using spaces is like using `\r\n`. :smiling_imp: 

Anyway, even if I preferred spaces I'd prefer to keep tabs because that's what currently happens to be there. Otherwise you'd just get a diff messing with git blame on every single line in every single file.

> At least my IDE doesn't fu*k up all the whitespaces since I added the .editorconfig file. Its was a pain, as FreshRSS is the only project I work on that uses tabs.

@kevinpapst Was auto-detect malfunctioning for some reason? I come across both plenty and it tends to work out fine.

Edit: btw, any editors [like this](https://softwareengineering.stackexchange.com/a/103278) for PHP? :-P

kevinpapst: @Frenzie I literally haven't worked on a "tab" project for the last ten years, so I might have missed the correct IDE setting. Auto-detection wasn't working unless I added the .editorconfig. Problem could be in front of the screen, without doubt. 
I am voting against almost all code formatting rules for a longer time now, as everyone can easily reformat the code to his own liking. But that problem can't be fixed unless we see changes in the IDEs and probably the VCS as well (whitespaces changes are a nightmare not only for the history). The linked article is a great write down! Until then I would always vote for sticking to the project standards and ... ehm ... strongly against 13 spaces :-D

app/Controllers/authController.php
```diff
@@ -141,12 +145,12 @@ public function formLoginAction() {
 
 				// All is good, go back to the index.
 				Minz_Request::good(_t('feedback.auth.login.success'),
-				                   array('c' => 'index', 'a' => 'index'));
+								   array('c' => 'index', 'a' => 'index'));
```
Alkarex: There is still a strange mix of tabs and spaces, isn't it?

aledeg: done

Alkarex: This line was only an example :-)
Please have a look from the GitHub UI https://github.com/FreshRSS/FreshRSS/pull/1751/files

aledeg: how do you spot those strange mix. I don't know if we are talking about the same thing. I've looked in the UI you linked and I am not sure what to look for.

Alkarex: Here is a screenshot, just to be sure we look at the same thing :-)

image

I see a lot of replaced whitespace, which have kept the same apparent indenting, but produced a git change anyway, which means that there has been a tab <-> space conversion.

Just putting the mouse on one of those lines shows that it is a mixture of tabs and space.

My local text editor also shows a difference (showing white spaces is quite common in various text editors).

It is even more visible when the tab size is different than 4, such as the default value of 8 used in gedit:

image

To be a bit more systematic, a regex search with something like `\t[ ][^*]` should highlight them.

aledeg: Ok, I see what you mean. Thank you

aledeg: @Alkarex So I've tried to restore the old spacing, but it doesn't work with phpcs. Here is the error I've got if I try to use spaces instead of tabs.

    -------------------------------------------------------------------------------------------------------
    FOUND 3 ERRORS AFFECTING 3 LINES
    -------------------------------------------------------------------------------------------------------
     153 | ERROR | [x] Tabs must be used to indent lines; spaces are not allowed
     154 | ERROR | [x] Tabs must be used to indent lines; spaces are not allowed
     155 | ERROR | [x] Tabs must be used to indent lines; spaces are not allowed
    -------------------------------------------------------------------------------------------------------
    PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
    -------------------------------------------------------------------------------------------------------

I don't know what to do now. On one hand I want to keep phpcs running smoothly, on the other end, it is not configured to allow what we defined for the coding standards.
If you have any idea, that would be appreciated.

aledeg:
@Frenzie Thank you, I'll include the specific version in the config.

aledeg:
@Frenzie @Alkarex I think it's all good now.
It's working everywhere except on PHP 5.3 but it's no big deal I think.

Frenzie (approved):

PR #1755, opened 2018-01-01
Add shortcuts to switch views

aledeg:


p/scripts/main.js
```diff
@@ -704,6 +704,29 @@ function init_shortcuts() {
 		'disable_in_input': true
 	});
 
+	shortcut.add(shortcuts.normal_view, function () {
+		$('#nav_menu_views .view_normal').get(0).click();
```
Frenzie, line 708: Hm, is this the best way? I'm inclined to think that the `?a=normal` part of the URL is in principle more stable than the specific DOM tree.

aledeg: The thing is that when you filter the view, you might want to keep the same filter while switching views. The DOM tree hold that value. That's why I went that way.

Frenzie: Ah, of course!

Frenzie (approved):
Lgtm.

PR #1756, opened 2018-01-01
Fix login bug when HTTP REMOTE_USER changes

Alkarex:
https://github.com/YunoHost-Apps/freshrss_ynh/issues/33

### DIFF
--- a/app/Models/Auth.php
+++ b/app/Models/Auth.php
@@ -13,6 +13,11 @@ class FreshRSS_Auth {
 	 * This method initializes authentication system.
 	 */
 	public static function init() {
+		if (Minz_Session::param('REMOTE_USER', '') !== httpAuthUser()) {
+			//HTTP REMOTE_USER has changed
+			self::removeAccess();
+		}
+
 		self::$login_ok = Minz_Session::param('loginOk', false);
 		$current_user = Minz_Session::param('currentUser', '');
 		if ($current_user === '') {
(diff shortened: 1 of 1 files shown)


Alkarex:
@plopoyop Could you please give a try to this patch, if you can?

Using the username in the session name as you proposed might be problematic because session names have their own limitations in terms of lengths and allowed characters. So I went for a session invalidation instead;

plopoyop:
It works for me.
I think invalidate the session is the best way to fix the problem.
Well done.

Alkarex:
Thanks for the quick feedback. Merged in /dev :-)

littabird:
Hello Alkarex

> @plopoyop Could you please give a try to this patch, if you can?
  
While I am still waiting for the PATCH https://github.com/FreshRSS/FreshRSS/issues/3617
 
I managed to identify the cause > **Rocket Loader™**

It's appear my log in problem is caused by Cloudfare's Rocket Loader™ https://support.cloudflare.com/hc/en-us/articles/200168056-What-does-Rocket-Loader-do-

Now I could log in after switching off the Rocket Loader™
> Meaning there is conflict between FRESHRSS and cloudfare.
> Good thing is there is solution >

If you have a Content Security Policy, you will need to update it to allow Rocket Loader to run: What is Content Security Policy (CSP), and how can I use it with Cloudflare?
https://support.cloudflare.com/hc/en-us/articles/216537517-What-is-Content-Security-Policy-CSP-and-how-can-I-use-it-with-Cloudflare-

Can you take a look if I can switch back on the Rocket Loader™?

Alkarex:
See https://github.com/FreshRSS/FreshRSS/issues/3617#issuecomment-841795350

PR #1759, opened 2018-01-02
Replace "keep history" magic value by a constant

aledeg:
I think the use of a magic value repeated many times in the code is prone to have some errors made by people not knowing its meaning. Using a constant is a bit more safe. Judging by some comments in the code, I am not the only one.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -317,10 +317,8 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 
 			$feed_history = $feed->keepHistory();
 			if ($isNewFeed) {
-				$feed_history = -1; //∞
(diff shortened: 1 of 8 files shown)


app/Controllers/feedController.php
```diff
@@ -318,9 +318,7 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 			$feed_history = $feed->keepHistory();
 			if ($isNewFeed) {
 				$feed_history = -1; //∞
```
Alkarex: Maybe it could be a good opportunity to do the same with -1 / ∞

aledeg: Ok, let's do that :)

aledeg: Done

PR #1760, opened 2018-01-02
Github update

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/907
https://github.com/FreshRSS/update.freshrss.org


Alkarex:
I can see I am getting in late for this PR. Any help welcome :-)

Inverle:
So why is GitHub update needed if there's already git update? Is it because git isn't available on every environment, and GH is supposed to be used as fallback?

Frenzie:
> because git isn't available on every environment

Yes, it's just a zip update. The "GitHub" part is more or less incidental.

PR #1761, opened 2018-01-03
Allow favorite in reading mode

aledeg:
I've got the inspiration from the mark as read action. So I've used the id of the feed to trigger the action and not the url in the link. This way, it works all the time. It works in reading mode, it works in normal mode when icon are not displayed.

See #1545

### DIFF
--- a/app/views/index/reader.phtml
+++ b/app/views/index/reader.phtml
@@ -18,11 +18,25 @@ if (!empty($this->entries)) {
 				<?php
 					$feed = FreshRSS_CategoryDAO::findFeed($this->categories, $item->feed());	//We most likely already have the feed object in cache
 					if (empty($feed)) $feed = $item->feed(true);
+					$favoriteUrl = array('c' => 'entry', 'a' => 'bookmark', 'params' => array('id' => $item->id()));
+					if ($item->isFavorite()) {
+						$favoriteUrl['params']['is_favorite'] = 0;
+					}
+					$readUrl = array('c' => 'entry', 'a' => 'read', 'params' => array('id' => $item->id()));
+					if ($item->isRead()) {
+						$readUrl['params']['is_read'] = 0;
+					}
 				?>
-				<a target="_blank" rel="noreferrer" class="go_website" href="<?php echo $item->link(); ?>">
+				<a class="read" href="<?php echo Minz_Url::display($readUrl); ?>">
+					<?php echo _i($item->isRead() ? 'read' : 'unread'); ?>
+				</a>
+				<a class="bookmark" href="<?php echo Minz_Url::display($favoriteUrl); ?>">
+					<?php echo _i($item->isFavorite() ? 'starred' : 'non-starred'); ?>
+				</a>
+				<a href="<?php echo _url('index', 'reader', 'get', 'f_' . $feed->id()); ?>">
 					<img class="favicon" src="<?php echo $feed->favicon(); ?>" alt="✇" /> <span><?php echo $feed->name(); ?></span>
 				</a>
-				<h1 class="title"><?php echo $item->title(); ?></h1>
+				<h1 class="title"><a target="_blank" rel="noreferrer" class="go_website" href="<?php echo $item->link(); ?>"><?php echo $item->title(); ?></a></h1>
 
 				<div class="author"><?php
 					$author = $item->author();


aledeg:
As @Frenzie mentioned in another PR, I feel that we should less rely on the DOM to get information for the JS code and more about what we know (ie URL).
I think at some point, we will need to render the whole entry in AJAX and send it back to replace the current one. This way, we only have one way of doing things and we are more flexible if we have different DOM architecture (ie new modes).

Frenzie (approved):
This makes me a lot happier! ;-D

aledeg:
@Aasemoon I've updated that PR, I am not really happy with the outcome but let me know what you think.
@Frenzie @Alkarex I am not really good with UI, what do you think ?

Aasemoon:
@aledeg YESSS!! This is perfect! It works great and I think the buttons are exactly where they should be. 
Thanks a lot! =) This makes my life so much easier..

Frenzie:
@aledeg I should probably be able to take a look tomorrow but I'll be able to drop a quick comment with a screenshot. :-)

p/scripts/main.js
```diff
@@ -169,16 +169,14 @@ function mark_favorite(active) {
 		return false;
 	}
 
-	var url = active.find("a.bookmark").attr("href");
```
Alkarex: It looks like the previous version should work fine. No need to change anything in `main.js`, is there?

Alkarex:
Sorry for the delay. Thanks @aledeg , this was a long awaited feature :-)
This does a good part of https://github.com/FreshRSS/FreshRSS/issues/603
Maybe it could be a good opportunity to make the title a link to the original article.
We would then only lack a share button and a tag button.

Alkarex:
~~Ah, and it looks like there is a tiny bug: in "auto mark as read", when scrolling down, the "read" icon is not updated.~~

aledeg:
If you mean ```Mark article as read… while scrolling```, I made some tests and it's working properly. If you mean something else, could you give me some more information, I am not sure what you're talking about.

Alkarex:
The ```Mark article as read… while scrolling``` works fine after all, sorry.

aledeg:
@Alkarex you've said
> Maybe it could be a good opportunity to make the title a link to the original article.

I've just checked and the feed name is a link to the original article (shortcut is working properly on that). Do you mean changing the article title to a link to the original article? I think it's redundant but it's true that it's surprising that the link is on the feed name.

Alkarex:
@aledeg Yes indeed. I think it is a bit confusing to have such differences with the normal view, and I would suggest to try to get more uniformity by (like in the normal view) making the title link to the original article, and the feed name link to a filter showing only articles of that feed.

aledeg:
I'll do that. Not tonight though.

aledeg:
@Alkarex I've changed the links while keeping the shortcut working

p/scripts/main.js
aledeg: That's what @Frenzie explained somewhere. It's better to rely on what we know (URL) and not the DOM content. If we do so, this part is not needed anymore.

Frenzie: I'd say it was more of an opinion than an explanation but the gist is that I think the DOM is much more likely to break expectations for some reason than the URL. That applies not only to personal mods but also to the program itself.

aledeg: We can also include ISP modifying page content :cry:

Alkarex: We were not relying on a random DOM structure (which might indeed change), but on a quite semantic class.

In this case, we end up having the URL in two places (in the HTML, and in the JavaScript), make assumptions about the URL format, and rely on the DOM even more (having an ID and a class on the active object) :-P

Frenzie: Yes, a.bookmark should be fine. It's not child n of firstchild yada yada.

aledeg: So do I revert it?

Frenzie: No objections either way in this case on my end. :-)

aledeg:
Oups, I've deleted the branch by mistake 😊

p/scripts/main.js
Alkarex: Yes please, I would better like the previous version

aledeg:
@Alkarex I've reverted my earlier change

PR #1762, opened 2018-01-04
count() parameter must be an array

Crupuk:
Since Php 7.2 i have this warning :

`Warning: count(): Parameter must be an array or an object that implements Countable in /app/Controllers/indexController.php on line 206`

currentGet(true) return "a" on the index page.


Frenzie:
Unfortunately (or luckily, depending on your perspective) #1739 already fixed this. Development takes place in the `dev` branch, you see. :-)

PR #1764, opened 2018-01-11
Option to hide nav_entries

mdemoss:
This adds a display option to hide the 3 nav buttons.

Needs text for other languages.

### DIFF
--- a/app/Models/ConfigurationSetter.php
+++ b/app/Models/ConfigurationSetter.php
@@ -183,6 +183,10 @@ class FreshRSS_ConfigurationSetter {
 	private function _mark_updated_article_unread(&$data, $value) {
 		$data['mark_updated_article_unread'] = $this->handleBool($value);
 	}
+	
+	private function _hide_nav_entries(&$data, $value) {
+		$data['hide_nav_entries'] = $this->handleBool($value);
+	}
 
 	private function _display_categories(&$data, $value) {
 		$data['display_categories'] = $this->handleBool($value);
--- a/app/Models/ConfigurationSetter.php
+++ b/app/Models/ConfigurationSetter.php
@@ -184,8 +184,8 @@ class FreshRSS_ConfigurationSetter {
 		$data['mark_updated_article_unread'] = $this->handleBool($value);
 	}
 
-	private function _hide_nav_entries(&$data, $value) {
-		$data['hide_nav_entries'] = $this->handleBool($value);
+	private function _show_nav_buttons(&$data, $value) {
+		$data['show_nav_buttons'] = $this->handleBool($value);
 	}
 
 	private function _display_categories(&$data, $value) {
--- a/app/Models/ConfigurationSetter.php
+++ b/app/Models/ConfigurationSetter.php
@@ -183,7 +183,7 @@ class FreshRSS_ConfigurationSetter {
 	private function _mark_updated_article_unread(&$data, $value) {
 		$data['mark_updated_article_unread'] = $this->handleBool($value);
 	}
-	
+
 	private function _hide_nav_entries(&$data, $value) {
 		$data['hide_nav_entries'] = $this->handleBool($value);
 	}
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -37,7 +37,7 @@ return array(
 			'no_limit' => 'No limit',
 			'thin' => 'Thin',
 		),
-		'hide_nav_entries' => 'Hide the lower-left navigation buttons',
+		'show_nav_buttons' => 'Show the navigation buttons',
 	),
 	'query' => array(
 		'_' => 'User queries',
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -37,6 +37,7 @@ return array(
 			'no_limit' => 'No limit',
 			'thin' => 'Thin',
 		),
+		'hide_nav_entries' => 'Hide the lower-left navigation buttons'
 	),
 	'query' => array(
 		'_' => 'User queries',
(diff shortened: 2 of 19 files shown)


app/layout/nav_entries.phtml
```diff
@@ -1,5 +1,7 @@
-<ul id="nav_entries">
-	<li class="item"><a class="previous_entry" href="#"><?php echo _i('prev'); ?></a></li>
-	<li class="item"><a class="up" href="#"><?php echo _i('up'); ?></a></li>
-	<li class="item"><a class="next_entry" href="#"><?php echo _i('next'); ?></a></li>
-</ul>
\ No newline at end of file
+<div <?php if (FreshRSS_Context::$user_conf->hide_nav_entries) echo 'hidden'; ?> > <!-- adding hidden to the ul doesn't have the desired effect -->
```
aledeg: It would be better to not render that template at all instead of rendering it and hidding it.
You could move the test in ```app/views/index/normal.phtml```.

app/i18n/en/conf.php
```diff
@@ -37,6 +37,7 @@
 			'no_limit' => 'No limit',
 			'thin' => 'Thin',
 		),
+		'hide_nav_entries' => 'Hide the lower-left navigation buttons'
```
aledeg: There is a missing comma at the end of the line, plus all the entries in other languages. You could use the CLI tool to add them automatically.

aledeg (changes requested):
You've made your PR on the master branch. You must do it on the dev branch.

app/layout/nav_entries.phtml
mdemoss: init_nav_entries() in p/scripts/main.js looks for it by id, might need to test to make sure that doesn't cause problems if it's not rendered at all.

aledeg: If there is a problem in js when the tag is not rendered, you should fix the js instead of just hiding the content.

mdemoss: I tested it, and init_nav_entries() doesn't have a problem, but it looks like the [StickyFeeds Extension](https://github.com/FreshRSS/Extensions/blob/b4aeeefc3fa98ccad0837f329ecd99287fc7a4ae/xExtension-StickyFeeds/static/script.js) may have an issue. 

`$(window).height() - $('#nav_entries').height()` evaluates to `NaN` when `#nav_entries` is not rendered because the height will be undefined. When `#nav_entries` is hidden, `$('#nav_entries').height()` evaluates to `0`.

I've pushed the change here to conditionally render `#nav_entries`, but it may be simpler and less disruptive to revert to hiding it.

aledeg: Good. In that case, we need to fix StickyFeeds :)

app/Models/ConfigurationSetter.php
```diff
@@ -183,6 +183,10 @@ private function _auto_remove_article(&$data, $value) {
 	private function _mark_updated_article_unread(&$data, $value) {
 		$data['mark_updated_article_unread'] = $this->handleBool($value);
 	}
+
```
Alkarex: There is a bit of forgotten whitespace here, which is flagged by Travis:

```
FILE: /home/travis/build/FreshRSS/FreshRSS/app/Models/ConfigurationSetter.php
-----------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------
186 | ERROR | [x] Whitespace found at end of line
```

aledeg:
I was wondering why not creating a CSS rule with the [Custom Css Extension](https://github.com/FreshRSS/Extensions/tree/1b5eb9c141c65003a6dad132190903a71298f5e8/xExtension-CustomCSS) instead of having extra configuration?
What is the use-case?

mdemoss:
I found that I never used the nav buttons myself, and I wanted to make use of the space they occupied, particularly on mobile.

Custom CSS would certainly work if the user knows CSS, installs the extension, and is willing and able to use their browser's dev tools to find the ID of the nav_entries, but the check box is usable by more people.

aledeg:
> Custom CSS would certainly work if the user knows CSS, installs the extension, and is willing and able to use their browser's dev tools to find the ID of the nav_entries, but the check box is usable by more people.

That's true, but I wonder if we should include it in the core since nobody complains over the years. I am not saying that it shouldn't be. I am just questioning the reasoning behind that PR to understand the underlying motives.

On the first place, I thought you wanted that option because you don't use FRSS on mobile so you didn't want the buttons. That's what I do and even if they never bothered me, but I figured why not. But now, I am not sure how you are using FRSS on mobile. I am curious to understand your process.

mdemoss:
>But now, I am not sure how you are using FRSS on mobile. I am curious to understand your process.

I found that I preferred scrolling and tapping item titles on mobile. Part of it is that there are a number of items interspersed in the feed that I don't want to open on mobile -- they'd require sound, too much bandwidth, or too much time. That is to say, when I'm on mobile I might want to read blog posts and skip YouTube videos. Using the buttons on mobile I could too-easily open and mark read an item without being able to first see its source and title when I'd rather get to it later on the desktop and it'd be a bit tedious to mark it as unread via touch.

Frenzie:
I think I mostly just scroll and tap. Sometimes I use the buttons to skip to the next article but now I'm wondering if I might not prefer just having more screen estate available. Also marking things as unread can be harder on mobile since it's a smaller target and perhaps your connection might fail at the wrong moment.

aledeg (approved):

Alkarex:
I have made the option positive (*show* instead of *hide*) https://github.com/FreshRSS/FreshRSS/pull/1764/commits/8568b2c2e77dccbbda81b64fe7e25f174ca9a52e
Please test if you can

aledeg (approved):
I haven't tested this version, but it looks good to me.

Issue #960, opened 2015-08-24
[feature] Add reset user password for admin


aledeg:
It would be nice to have a way to reset the user password as an admin. Maybe also the email address.
Because at the moment action on user management are pretty limited.

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/901

blankoworld:
What's about non-admin users?

I got a user that don't have any way to reset the password. And admin interface don't permit to change its password.

What should I do for this user?

Frenzie:
I'm pretty sure the FR is for an admin to change any user's password. :-)

Alkarex:
@blankoworld For the time being, the easiest is if you are able to use the command-line, then you can use:
```sh
./cli/update-user.php --user username --password 'password' --api_password 'api_password'
```
https://github.com/FreshRSS/FreshRSS/blob/master/cli/README.md#commands

PR #1765, opened 2018-01-12
Allow admin user to reset passwords

aledeg:
See #960

Comments are welcomed.


app/i18n/en/feedback.php
```diff
@@ -101,6 +101,10 @@
 			'_' => 'User %s has been deleted',
 			'error' => 'User %s cannot be deleted',
 		),
+		'updated' => array(
+			'_' => 'The user %s has been updated',
```
Frenzie: For consistency with the other strings (see a couple of lines above) that should be "User %s…"

aledeg: Done

app/views/user/manage.phtml
```diff
@@ -46,12 +46,48 @@
...
 
+	<form method="post" action="<?php echo _url('user', 'update'); ?>">
+		<input type="hidden" name="_csrf" value="<?php echo FreshRSS_Auth::csrfToken(); ?>" />
+		<legend><?php echo _t('admin.user.update_users'); ?></legend>
+
+		<div class="form-group">
+			<label class="group-name" for="current_user"><?php echo _t('admin.user.selected'); ?></label>
+			<div class="group-controls">
+				<select id="current_user" class="select-change" name="username">
+					<option selected="selected">&nbsp;</option>
```
Alkarex: Good idea, and could you please do the same thing (i.e. empty first option) for the next list of users a bit further down?
Detail: I would prefer a simple space, or Unicode non-breaking space, as I am not so found of entities.

Alkarex:
At a first glance, it looks very good to me 👍 
I will try to test very soon.
I am wondering whether it is necessary to have two times the list of users, but there may be a point in having distinct areas for updating vs. deleting.

app/views/user/manage.phtml
aledeg: No pbs for the space.
On the other hand, I've tried to replicate the same thing for the other list but there is JS involved in that one. It's not that easy and I didn't take the time to dig into it.

aledeg:
My initial idea was to split things between updating and deleting to avoid confusion.

app/views/user/manage.phtml
aledeg: I've found a way :)

app/Controllers/userController.php
```diff
@@ -110,12 +135,7 @@ public function manageAction() {
 
 		Minz_View::prependTitle(_t('admin.user.title') . ' · ');
 
-		// Get the correct current user.
-		$username = Minz_Request::param('u', Minz_Session::param('currentUser'));
-		if (!FreshRSS_UserDAO::exist($username)) {
-			$username = Minz_Session::param('currentUser');
-		}
-		$this->view->current_user = $username;
+		$this->view->current_user = Minz_Request::param('u');
```
Alkarex, line 138: When simply browsing to `?c=user&a=manage`, `$this->view->current_user` seems to be null, and I get the following error:
```
<br />
<b>Fatal error</b>:  Uncaught Error: Call to a member function execute() on boolean in ./FreshRSS/app/Models/EntryDAO.php:879
Stack trace:
#0 ./FreshRSS/app/Controllers/userController.php(142): FreshRSS_EntryDAO-&gt;count()
#1 ./FreshRSS/lib/Minz/Dispatcher.php(118): FreshRSS_user_Controller-&gt;manageAction()
#2 ./FreshRSS/lib/Minz/Dispatcher.php(47): Minz_Dispatcher-&gt;launchAction('manageAction')
#3 ./FreshRSS/lib/Minz/FrontController.php(84): Minz_Dispatcher-&gt;run()
#4 ./FreshRSS/p/i/index.php(46): Minz_FrontController-&gt;run()
#5 {main}
  thrown in <b>./FreshRSS/app/Models/EntryDAO.php</b> on line <b>879</b><br />
```

Alkarex: Things seem to work fine with the previous code, for this little section

aledeg: @Alkarex That's weird. I cannot reproduce.
My error_reporting is set to E_ALL. My user doesn't have authentication. I cannot think of something else that could break that.
Any ideas ?

Alkarex: When landing on the page the first time, `Minz_Request::param('u')` is not defined, is it?

aledeg: It's not defined but set automatically to ```false```.
By the way, I am using mysql. Maybe you're using something different.

Alkarex: Tested with SQLite, not tested yet with MySQL - but it would be an error if it does not fail :-P
Anyway, `$entryDAO->count()` (as well as `$databaseDAO->size()` probably) should not be called with an undefined (or false) username.
So I suggest to either keep the previous code, which loaded the default user (seems like the easiest), or move the count elsewhere.

aledeg: @Alkarex I've kept what I was doing previously but I've added a validation on the user. This way, we don't have an error.
I've tried on Sqlite but I couldn't make frss work :'(

Alkarex:
Damn, a bit too fast:
```
Notice: Undefined property: Minz_View::$nb_articles in ./FreshRSS/app/views/user/manage.phtml on line 100

Notice: Undefined property: Minz_View::$size_user in ./FreshRSS/app/views/user/manage.phtml on line 101
```

aledeg:
Oups, should I make a new PR or update the previous one?

Alkarex:
New one please :-)

aledeg:
👍

PR #1773, opened 2018-01-24
Fix css classes for reading modes #1714 #1758

kevinpapst:
See #1714 and #1758


Frenzie (approved):
Didn't test but lgtm

aledeg (approved):
Same here

PR #1774, opened 2018-01-25
API /reader/api/0/stream/items/contents

Alkarex:
For FeedMe https://play.google.com/store/apps/details?id=com.seazon.feedme

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -270,6 +270,8 @@ function subscriptionList() {
 	$res = $stm->fetchAll(PDO::FETCH_ASSOC);
 
 	$salt = FreshRSS_Context::$system_conf->salt;
+	$faviconsUrl = Minz_Url::display('/f.php?', '', true);
+	$faviconsUrl = str_replace('/api/greader.php/reader/api/0/subscription', '', $faviconsUrl);	//Security if base_url is not set properly
 	$subscriptions = array();
 
 	foreach ($res as $line) {
@@ -286,7 +288,7 @@ function subscriptionList() {
 			//'firstitemmsec' => 0,
 			'url' => $line['url'],
 			'htmlUrl' => $line['website'],
-			'iconUrl' => Minz_Url::display('/f.php?' . hash('crc32b', $salt . $line['url']), '', true),
+			'iconUrl' => $faviconsUrl . hash('crc32b', $salt . $line['url']),
 		);
 	}
 
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -435,6 +435,51 @@ function unreadCount() {	//http://blog.martindoms.com/2009/10/16/using-the-googl
 	exit();
 }
 
+function entriesToArray($entries) {
+	$items = array();
+	foreach ($entries as $entry) {
+		$f_id = $entry->feed();
+		if (isset($arrayFeedCategoryNames[$f_id])) {
+			$c_name = $arrayFeedCategoryNames[$f_id]['c_name'];
+			$f_name = $arrayFeedCategoryNames[$f_id]['name'];
+		} else {
+			$c_name = '_';
+			$f_name = '_';
+		}
+		$item = array(
+			'id' => /*'tag:google.com,2005:reader/item/' .*/ dec2hex($entry->id()),	//64-bit hexa http://code.google.com/p/google-reader-api/wiki/ItemId
+			'crawlTimeMsec' => substr($entry->id(), 0, -3),
+			'timestampUsec' => '' . $entry->id(),	//EasyRSS
+			'published' => $entry->date(true),
+			'title' => $entry->title(),
+			'summary' => array('content' => $entry->content()),
+			'alternate' => array(
+				array('href' => htmlspecialchars_decode($entry->link(), ENT_QUOTES)),
+			),
+			'categories' => array(
+				'user/-/state/com.google/reading-list',
+				'user/-/label/' . $c_name,
+			),
+			'origin' => array(
+				'streamId' => 'feed/' . $f_id,
+				'title' => $f_name,	//EasyRSS
+				//'htmlUrl' => $line['f_website'],
+			),
+		);
+		if ($entry->author() != '') {
+			$item['author'] = $entry->author();
+		}
+		if ($entry->isRead()) {
+			$item['categories'][] = 'user/-/state/com.google/read';
+		}
+		if ($entry->isFavorite()) {
+			$item['categories'][] = 'user/-/state/com.google/starred';
+		}
+		$items[] = $item;
+	}
+	return $items;
+}
+
 function streamContents($path, $include_target, $start_time, $count, $order, $exclude_target, $continuation) {
 //http://code.google.com/p/pyrfeed/wiki/GoogleReaderAPI
 //http://blog.martindoms.com/2009/10/16/using-the-google-reader-api-part-2/#feed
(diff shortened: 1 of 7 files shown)


Frenzie:
So they use a slightly incompatible Google Reader API?

Alkarex:
@Frenzie More some features that I did not implement because I had not seen anybody using them :-)

p/api/greader.php
```diff
@@ -216,9 +216,13 @@ function token($conf) {
 function checkToken($conf, $token) {
 //http://code.google.com/p/google-reader-api/wiki/ActionToken
 	$user = Minz_Session::param('currentUser', '_');
+	if ($user !== '_' && $token == '') {
```
Frenzie, line 219: == or ===?

p/api/greader.php
```diff
@@ -476,73 +528,37 @@ function streamContents($path, $include_target, $start_time, $count, $order, $ex
 			break;
 	}
 
-	if (!empty($continuation)) {
+	if ($continuation != '') {
```
Frenzie, line 533: So we're fine if it's FALSE or 0? Also, is $continuation trimmed anywhere or could you end up with something empty-adjacent like a space?

Also, like above, != or !==?

Frenzie (approved):
Is there such a thing as API docs?

p/api/greader.php
Alkarex: Strict for the first comparison, loose for the second (match empty string and 0 in particular)

p/api/greader.php
Alkarex: Yes, loose comparison there. The test should reject empty string, false, 0, null, etc. and only keep substantial strings.
But good point with sanitising those parameters. Will do so.

Frenzie: Ah, so `!= something falsy` (loose) basically means exactly the same thing as `empty()`? I know that PHP plays a bit fast and loose with things like 1 and 0 vs true and false, but I didn't realize empty strings were included in being falsey. I thought any string was truthy. Apparently the string `'0'` is also falsey.

Confusingly, on that last note, `0.0` (float) is falsey and `'0.0'` (string) is truthy.

Alkarex: Yes indeed. `empty()` is mostly relevant when working with potentially non-existing entries such as in a array, and it is similar to doing `!isset($t[1]) || $t[1] == ''`. In the case where we have a variable and not an index, there is little point of using `empty()` and it used to be slower (I am not sure whether there is any difference nowadays).

Frenzie: I find `empty()` slightly clearer in the sense that I don't necessarily think of an empty string as falsey while I do think of it as empty, but that's just my problem I suppose. :-P

Python is much more consistent in counting empty arrays, empty strings, etc. as falsey than the likes of JavaScript or PHP. But in Ruby, for example, empty strings are truthy, and in a sense they are in Lua too. In POSIX and related shells you also have to be well aware to use checks like `-z` (unset or empty) and `-n` (set and non-empty).

Incidentally, can we get this kind of fun out of PHP?
```javascript
bla = '0'; if (bla && bla == false) console.log('/sigh')
/sigh
```

It's truthy (non-empty string) but at the same time '0' == false. Perhaps even more confusingly, you can get the same behavior out of `new Boolean(false)` (the existence of any object is inherently truthy in JS, but its value == false).

Alkarex: I do not think there is anything in PHP that can evaluate positively `$bla && $bla == false`

Frenzie: I didn't mean that specific scenario, just if there's something similarly crazy lurking in PHP that I don't know about. ;-)

Alkarex: https://www.destroyallsoftware.com/talks/wat :-)

Frenzie: Fun, but no PHP. Well, at least I got the last one right after the hint. I didn't really dare think not a number at first because that just seemed too logical. xD

Alkarex: All right, I have one for you. Guess the output :-P

```php
<?php
echo ctype_digit(1234) ? 'true' : 'false', "\n";
echo ctype_digit(123) ? 'true' : 'false', "\n";
echo ctype_digit('123') ? 'true' : 'false', "\n";
```

Frenzie: My intuitive guess would be false false true but I'm guessing there's a catch. Now to check… ;-)

Edit: hm, true false true. Well then…

> If an integer between -128 and 255 inclusive is provided, it is interpreted as the ASCII value of a single character (negative values have 256 added in order to allow characters in the Extended ASCII range). Any other integer is interpreted as a string containing the decimal digits of the integer.

Alkarex: Close enough: true false true. :-)
> If an integer between -128 and 255 inclusive is provided, it is interpreted as the ASCII value of a single character (negative values have 256 added in order to allow characters in the Extended ASCII range). Any other integer is interpreted as a string containing the decimal digits of the integer.

Alkarex:
A new version of FeedMe will be released when we release FreshRSS 1.10.0 with this patch.

Alkarex:
Preview:
screenshot_20180208-140434

Alkarex:
@MaluNoPeleke and others: The /dev branch of FreshRSS (soon 1.10.0) and FeedMe 3.5.3+ (currently available as beta in the Play Store) should work fine together. Tests welcome.
* https://play.google.com/store/apps/details?id=com.seazon.feedme
* https://github.com/FreshRSS/FreshRSS/tree/dev

bramboomen:
Hi, I just set-up a fresh install of FreshRSS dev and FeedMe 3.5.4 and encountered an issue.
The list of articles does not synchronize. As far as I can tell the connection is working because starred items do synchronize and FeedMe has the complete list of my subscriptions but no unstarred, read or unread articles.
I did have some issues with setting up the API, in particular with the PATH_INFO variable in nginx, but currently both api checks pass and the EasyRSS application is working without problem.
I would try some manual API calls to further debug the problem, but I have no idea where to begin in that endeavour...

Alkarex:
@blankoworld Thanks for the test.
Please make a detailed log by enabling the two following lines
https://github.com/FreshRSS/FreshRSS/blob/c79b8fb46e25cd513799224b83cb93282c3a8e26/p/api/greader.php#L740-L743
and either set `'environment' => 'development',` in `./data/config.php` or change the lines to `Minz_Log::warning()`.
It will be in `./data/users/_/log_api.txt`

bramboomen:
@Alkarex 
Happy to, let me know if I can do anything else!
[log_api.txt](https://github.com/FreshRSS/FreshRSS/files/1735111/log_api.txt)

Alkarex:
@bramboomen Could you please ensure that you have a number of unread articles before trying a first synchronisation?
If that still does not work, would it be possible for you to make me a test account on your FreshRSS instance?

bramboomen:
@Alkarex 
There were both unread and read articles but they only show up if I star them.
I will send you an email with the information for the test-account.

Alkarex:
Thanks for the test account. It seems to work fine for me :-)
Can you give a try to the test account you have just made for me?

Alkarex:
Ah, in FeedMe, under Settings / Synchronisation Mode, pick "All". (I am not sure of the exact wording of the English version)

bramboomen:
@Alkarex 
Synchronizing 'all' instead of 'per feed' did the trick, thank you!

Alwaysin:
Had the same problem!
I guess a lot of people might conclude "this is not working" and go somewhere else.

Anyway, this reader is a nice addition, I'll use it a couple days to see if it fits my needs :)

Alkarex:
The next version of FeedMe will have the proper synchronisation setting by default.

fcoiffie:
Regarding the FeedMe configuration, it's not very easy to guess that the Host is: `http://adresse_serveur/p/api/greader.php`

Alkarex:
@fcoiffie Indeed, therefore:
* It is shown in FreshRSS, in the API test page (same page than where one set the API password)
* It is in the documentation https://freshrss.github.io/FreshRSS/en/users/06_Mobile_access.html#compatible-clients
* I have already suggested to the developer of FeedMe to auto-add `api/greader.php` if not already included in the address provided by the user.
* I will try to see if we can add a redirect without affecting performances.

Please note that is is better to expose only the `./p/` directory of FreshRSS (and not its parent), to have a public URL like `https://freshrss.example/` instead of `https://freshrss.example/p/`.

Issue #1757, opened 2018-01-02
Views shortcuts are broken


aledeg:
The shortcuts introduced in #1755 are broken since #1714. It relies on css class but those have been replaced by the same class.
I think we should rethink #1714 for the nav bar generation. Or find a better way to match the buttons.

aledeg:
I've got that idea for matching the buttons. There might be a better way to handle that though.

kevinpapst:
Can you please have a look at the comments in #1714 - I guess the last fix merge could lead to strange behaviour ... in the case an extension changes the list of reading modes which is possible since #1714. 
Even though there is only one demo extension existing by now that makes use of the dynamic list of reading modes / views, I guess we could improve the fix to match the links more safely.

Alkarex:
Can this be closed?

PR #1775, opened 2018-01-26
Fix shortcuts for view switching

aledeg:
I've introduced shortcuts to switch between view in #1755. They have been broken by #1714.
Then I've made an ugly fix in #1758.

This change revert all changes to have something better.

See #1757


Alkarex (approved):

Frenzie (approved):
Nice :-)

PR #1776, opened 2018-01-30
Translated to Korean

sjeon87:
I wonder how FreshRSS behaves if `mute` of a feed is enabled, specifically in `sub.php`.


aledeg:
@gsongsong I'll check that tonight and let you know

app/i18n/kr/sub.php
```diff
@@ -2,8 +2,8 @@
 
 return array(
 	'api' => array(
-		'documentation' => 'Copy the following URL to use it within an external tool.',// TODO
-		'title' => 'API',// TODO
+		'documentation' => '외부 도구에서 API를 사용하기 위해서 아래 URL을 사용하세요.',
+		'title' => 'API',
```
aledeg, line 6: @gsongsong  As the translation is the same as in english. You should add exceptions for the validation tool.
Check in ```cli/i18n/ignore``` for examples. All keys added in those files are discarded from the completion validation tool.

aledeg (changes requested):

app/i18n/kr/sub.php
gsongsong: Thank you. I fixed it. But validation will also fail due to `mute`. I need to know the exact behavior of it before translation.

aledeg:
@gsongsong Thank you for your contribution. It's appreciated that you're maintaining the Korean translations.

To answer your original question, when you mute a feed, it won't refresh anymore. The behavior is the same as before when selecting ```∞``` in '다음 시간이 지나기 전에 새로고침 금지'. It's linked to 
'표시'. You can now mute a feed and still see it. Or archive it and still updating it. Or any combination of those actions.
I hope it helps you to understand. If not, please ask more questions.

If you're planning on improving the Korean translation, there is probably some keys you can add to the ignore files on top of the one I've mentioned earlier. If you check the [Travis build](https://travis-ci.org/FreshRSS/FreshRSS/jobs/335202855), you can see that the Korean translation is 91.8% complete. If you need help to find the missing translations, you can use the cli tool to do that.

aledeg (approved):
LGTM

Frenzie:
Thanks!

PR #1777, opened 2018-02-02
Trim spaces from user given url

keltroth:
If a user gives an url with spaces an error is thrown... I suggest to avoid those by triming the given url.


Frenzie:
Note that while trim does strip tabs (which you might accidentally get from tables) it doesn't strip non-breaking spaces (which you might also get from websites). I'm not sure if that matters. :-)

Frenzie (approved):
Looks good to me.

keltroth:
good point about non breaking spaces... anyway it's still better to trim "normal" spaces :)

keltroth:
Oops, those commits where not intended to come here... Sorry...

Frenzie:
Oh, that needn't be a problem. You can either do something like:

```bash
git reset --hard HEAD~4
```

(because there are four commits you want to get rid of; test on a copied branch first)

Or somewhat safer:

```bash
git rebase -i HEAD~4
```

Then `git push -f`.

keltroth:
Ok thanks for the tip !

PR #1778, opened 2018-02-02
Trim spaces from user given url

keltroth:
Forget the other PR, I made this one to avoid new commit on my own dev branch to come to you :)

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -39,13 +39,13 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 	 * @throws FreshRSS_Feed_Exception
 	 * @throws Minz_FileNotExistException
 	 */
-	public static function addFeed($param_url, $title = '', $cat_id = 0, $new_cat_name = '', $http_auth = '') {
+	public static function addFeed($url, $title = '', $cat_id = 0, $new_cat_name = '', $http_auth = '') {
 		FreshRSS_UserDAO::touch();
 		@set_time_limit(300);
 
 		$catDAO = new FreshRSS_CategoryDAO();
 
-		$url = trim($param_url);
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
As I said, lgtm.

Frenzie:
See #1777 for some (minimal) discussion. As I said in https://github.com/FreshRSS/FreshRSS/pull/1777#issuecomment-362551550

> Note that while trim does strip tabs (which you might accidentally get from tables) it doesn't strip non-breaking spaces (which you might also get from websites). I'm not sure if that matters. :-)

app/Controllers/feedController.php
```diff
@@ -39,12 +39,14 @@ public function firstAction() {
 	 * @throws FreshRSS_Feed_Exception
 	 * @throws Minz_FileNotExistException
 	 */
-	public static function addFeed($url, $title = '', $cat_id = 0, $new_cat_name = '', $http_auth = '') {
+	public static function addFeed($param_url, $title = '', $cat_id = 0, $new_cat_name = '', $http_auth = '') {
```
Alkarex: Please either keep `$url` (and later below use `$url = trim($url);`) or update the `@param` documentation a few lines higher up. I would better like option 1.

Alkarex:
Thanks @keltroth 👍 
Please add your name to https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md

app/Controllers/feedController.php
Frenzie: I thought of writing something like that and decided against it without realizing that GitHub was hiding the documentation a few lines up. :-P

I'd also go with option 1.

keltroth: Did option 1 👍

PR #1794, opened 2018-02-25
Fix MySQL minimal requirement

laedit:


Alkarex:
Thanks! Could you please do the same fix in https://github.com/FreshRSS/FreshRSS/blob/32e9d3a5a17edd33ed19158bf567c857fb4fea78/docs/fr/users/01_Installation.md ?

laedit:
Yes, of course, thanks for the ultra-quick response!

Frenzie:
For context:
https://github.com/FreshRSS/FreshRSS/commit/fe18d12551f626c56257f825caf8f97a4a5461ce
https://github.com/FreshRSS/FreshRSS/issues/789#issuecomment-73878076

That being said, it's still not entirely clear if there is a specific point to the .3+. :-P Did 5.5 coincidentally fix something important in the same point release number or was that just sort of carried over when replacing 5.0 with 5.5?

Alkarex:
@Frenzie The reason for this specific MySQL version is that it is the first to support utf8mb4 https://dev.mysql.com/doc/relnotes/mysql/5.5/en/news-5-5-3.html

Frenzie:
Funny that it turned out to be the same point release. :-P

Issue #901, opened 2015-07-13
Give indications to change the password


marienfressinaud:
Sometimes, people send emails to ask how to change the password because they forgot the initial one. We should write a small how-to.

marienfressinaud:
See some comments in https://github.com/FreshRSS/FreshRSS/issues/676

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/960

aledeg:
#1765

PR #1798, opened 2018-02-26
Add a section in the FAQ about changing a password

aledeg:
See #901

### DIFF
--- a/docs/en/users/07_Frequently_Asked_Questions.md
+++ b/docs/en/users/07_Frequently_Asked_Questions.md
@@ -33,3 +33,14 @@ Here is a list of feeds which don't work:
 * http://foulab.org/fr/rss/Foulab_News: is not a W3C valid feed (November 2014)
 * http://eu.battle.net/hearthstone/fr/feed/news: is not a W3C valid feed (Novembre 2014)
 * http://webseriesmag.blogs.liberation.fr/we/atom.xml: is not working for the user but succeed on all the described validations (November 2014)
+
+## How to change a forgotten password?
+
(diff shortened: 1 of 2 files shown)


Alkarex (approved):

docs/en/users/07_Frequently_Asked_Questions.md
```diff
@@ -33,3 +33,14 @@ Here is a list of feeds which don't work:
...
 * http://eu.battle.net/hearthstone/fr/feed/news: is not a W3C valid feed (Novembre 2014)
 * http://webseriesmag.blogs.liberation.fr/we/atom.xml: is not working for the user but succeed on all the described validations (November 2014)
+
+## How to change a forgotten password?
+
+Since [1.10.0](https://github.com/FreshRSS/FreshRSS/releases/tag/1.10.0) release, admins are able to change user passwords directly from the interface. This interface is available under  ```Administration → Manage users```.
+Select a user, enter a password, and validate.
+
+Since [1.8.0](https://github.com/FreshRSS/FreshRSS/releases/tag/1.8.0) release, admins are able to change user passwords using a terminal. It worth mentioning that it must have access to PHP CLI. Open a terminal, and type the following command:
+```bash
```
Alkarex: Very minor detail: Even if this is only for code highlighting, `sh` should be enough there, as this command is not using any Bash-specific syntax :-)

aledeg: here, ```bash``` is a keyword to have colors when displaying the command in github.

Frenzie: You get the same highlighting with sh though. ;-) (They're probably just synonyms for GH Markdown.)

````
```sh
cd bla #sdfd
```
````

```sh
cd bla #sdfd
```

Issue #1799, opened 2018-02-27
1.10 displays _ for author in EasyRSS


Frenzie:
I suspect it was triggered by #1774

# 1.10
![screenshot_20180227-090116](https://user-images.githubusercontent.com/202757/36750090-f1aaff5c-1bfc-11e8-904e-5586ace189bf.png)

# Some older stuff from 1.9 with author
![screenshot_20180227-090139](https://user-images.githubusercontent.com/202757/36750091-f1ca5334-1bfc-11e8-80a5-84276dbb06fb.png)

Alkarex:
Arg. It is not the author information but the feed name, which is missing apparently.

Alkarex:
Patch coming in FreshRSS

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/1800

PR #1800, opened 2018-02-27
Fix API bug with EasyRSS

Alkarex:
Regression in FreshRSS 1.10.0
https://github.com/FreshRSS/FreshRSS/issues/1799


Frenzie:
That was even simpler than I thought it might be. :-)

PR #1801, opened 2018-02-28
Feed parsing: use author email when there's no author name

perrinjerome:
This is especially useful because when author is given as
`<author>Author Name</author>` ( as in this example
https://cyber.harvard.edu/rss/rss.html#ltauthorgtSubelementOfLtitemgt ),
SimplePie will expose *Author Name* as `email`.

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -355,7 +355,7 @@ class FreshRSS_Feed extends Minz_Model {
 				$this->id(),
 				$item->get_id(false, false),
 				$title === null ? '' : $title,
-				$author === null ? '' : html_only_entity_decode(strip_tags($author->name)),
+				$author === null ? '' : html_only_entity_decode(strip_tags($author->name == null ? $author->email : $author->name)),
 				$content === null ? '' : $content,
 				$link === null ? '' : $link,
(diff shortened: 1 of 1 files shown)


perrinjerome:
A random feed where authors were not parsed properly:  https://devnet.kentico.com/rss/articles 

I could not find which test could be extended to cover this, so I did not write any new test.

app/Models/Feed.php
```diff
@@ -355,7 +355,7 @@ public function loadEntries($feed) {
 				$this->id(),
 				$item->get_id(false, false),
 				$title === null ? '' : $title,
-				$author === null ? '' : html_only_entity_decode(strip_tags($author->name)),
+				$author === null ? '' : html_only_entity_decode(strip_tags($author->name === null ? $author->email : $author->name)),
```
Alkarex: I would suggest `== ''` instead of `=== null` to catch other nullish values such as empty strings

Alkarex:
Thanks! 👍

app/Models/Feed.php
perrinjerome: Good point, I pushed-force with a fix for that

Issue #1723, opened 2017-12-13
SSL certificate problem


qwertygc:
Hi,
I can't add a feed : 
cURL error 60: SSL certificate problem: unable to get local issuer certificate [https://www.lagedefaire-lejournal.fr/feed/]

Thx.

kevinpapst:
Seems to be a GeoTrust certificate - wasn't that one of the problematic subcompanies of Symantec? I guess thats a problem only the site owner can fix:
```
wget https://www.lagedefaire-lejournal.fr/feed/
--2017-12-13 18:48:48--  https://www.lagedefaire-lejournal.fr/feed/
Resolving www.lagedefaire-lejournal.fr (www.lagedefaire-lejournal.fr)... 137.74.93.217
Connecting to www.lagedefaire-lejournal.fr (www.lagedefaire-lejournal.fr)|137.74.93.217|:443... connected.
ERROR: cannot verify www.lagedefaire-lejournal.fr's certificate, issued by '/C=US/O=GeoTrust Inc./CN=RapidSSL SHA256 CA':
  Unable to locally verify the issuer's authority.
```
But lets see if someone has a better idea

Alkarex:
Hello,
One can disable SSL verification globally in FreshRSS, but this is not
recommended:
https://github.com/FreshRSS/FreshRSS/issues/1144#issuecomment-216024457

See also:
https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fwww.lagedefaire-lejournal.fr%2Ffeed%2F

https://curl.haxx.se/docs/sslcerts.html

Alkarex:
P.S.: You can also manually add the missing root certificate to your server

qwertygc:
> P.S.: You can also manually add the missing root certificate to your server

How ? It's a shared server.

Alkarex:
> How ? It's a shared server.

Then you cannot :-)

They have at least a couple of SSL problems: https://www.ssllabs.com/ssltest/analyze.html?d=www.lagedefaire-lejournal.fr

I have not looked in details but it looks like their certificate chain is incomplete, or using a certificate that is not currently standard in e.g. Ubuntu LTS.

```
curl --insecure -v -I 'https://www.lagedefaire-lejournal.fr/feed/'
*   Trying 137.74.93.217...
* TCP_NODELAY set
* Connected to www.lagedefaire-lejournal.fr (137.74.93.217) port 443 (#0)
* ALPN, offering http/1.1
[...]

Issue #1654, opened 2017-10-02
DB add JSON field


Alkarex:
I suggest adding a JSON field to the feed table, to add custom properties (e.g. user-agent, mark as unread upon update, strip data...) without having to add many new columns.
E.g. useful for:
* https://github.com/FreshRSS/FreshRSS/issues/1627
* https://github.com/FreshRSS/FreshRSS/issues/891

Frenzie:
Also potentially useful for extensions.

Alkarex:
It looks like letting PHP parse and stringify JSON would be the most portable.
Notes, if we wanted to use databases native JSON abilities:
* MySQL 5.7.8+ https://dev.mysql.com/doc/refman/5.7/en/json.html (we so far support MySQL 5.5.3+)
  * In the Ubuntu world, this means 16.04 LTS instead of 12.04 LTS
* PostgreSQL 9.2+ https://www.postgresql.org/docs/9.2/static/functions-json.html
* SQLite requires an extension :-( https://sqlite.org/json1.html

Frenzie:
MariaDB starting with 10.0.16 https://mariadb.com/kb/en/library/connect-json-table-type/ (that means Debian Jessie and Ubuntu 16.04LTS)

I don't know how adding JSON to these databases works, but with any luck someone already created a wrapper function?

Alkarex:
Well, for the needs I foresee, the native JSON functions would have been nice but not really needed. A `varchar` type with JSON parsing / stringifying in PHP should be enough.

Frenzie:
Sure. It's just a performance thing. I don't know if it's a performance thing that *matters* but it just sounds so self-evident that I think a solution must already exist.

perrinjerome:
👍 making it a simple `varchar` column seems enough. I just want to suggest using a column type where there's no limit on the length, such as `text` or `blob`.

Alkarex:
Continues in https://github.com/FreshRSS/FreshRSS/pull/1838

PR #1802, opened 2018-02-28
WIP: allow ignoring SSL certificate for some feed

perrinjerome:
Implementing suggestion from https://github.com/FreshRSS/FreshRSS/issues/1723#issuecomment-369171458

This is very early work, I still have to polish several things, but I'm already interesting on feedback on:
 * wether this is a good idea at all
 * this introduce a new column. Do you guys already have anything to manage schema migrations ? I did b092527 because I saw the addition of `feed.ttl` have been managed like this, but I know this does not support all cases.

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -23,6 +23,7 @@ class FreshRSS_Feed extends Minz_Model {
 	private $priority = self::PRIORITY_MAIN_STREAM;
 	private $pathEntries = '';
 	private $httpAuth = '';
+        private $tolerateInvalidSSLCertificate = false;
 	private $error = false;
 	private $keep_history = self::KEEP_HISTORY_DEFAULT;
 	private $ttl = self::TTL_DEFAULT;
@@ -105,6 +106,9 @@ class FreshRSS_Feed extends Minz_Model {
 			);
 		}
 	}
+        public function tolerateInvalidSSLCertificate() {
+                return $this->tolerateInvalidSSLCertificate;
+        }
 	public function inError() {
 		return $this->error;
 	}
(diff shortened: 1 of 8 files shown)


Alkarex:
Thanks for this draft!
For such secondary feed properties, I was planning to use a new field containing some JSON https://github.com/FreshRSS/FreshRSS/issues/1654 . Indeed, there is a bunch of various properties we could want to save for feeds (including some created by extensions).
P.S. Yes, this is a good idea 👍

app/Models/Feed.php
```diff
@@ -23,6 +23,7 @@ class FreshRSS_Feed extends Minz_Model {
 	private $priority = self::PRIORITY_MAIN_STREAM;
 	private $pathEntries = '';
 	private $httpAuth = '';
+        private $tolerateInvalidSSLCertificate = false;
```
Frenzie, line 26: Btw, there are some indentation errors like these (spaces instead of tabs). See the CI report for specifics but they also jump out here on GH. :-)

Frenzie:
I was going to say I like that JSON idea but I already did a few months ago. :rofl:

perrinjerome:
Thanks @Alkarex , a JSON fields for extra-properties seems very appropriate for this purpose. I'll update this MR once the JSON field is available.

app/Models/Feed.php
perrinjerome: Thanks. I should have mentioned it because I noticed this when commiting my changes. I wanted to submit my current state "as is" to make sure the direction was correct.
Good to see the test suite is watching this kind of mistakes 😉

perrinjerome:
I opened a feature request issue #1811 for this (just to keep things organised )

samwei12:
> Implemented in #1838 . Tests and feedback welcome

2021-06-04 02:03:08
cURL error 7: Failed to connect to raw.githubusercontent.com port 443: Connection refused [https://github.com/hellodword/wechat-feeds/raw/feeds/MzU4Mjg3MDAyMQ==.xml]

hi, I had met this problem when I use FreshRSS in a docker , and I can't find any option about SSL, How can I resole this problem ?
thanks

samwei12:
image

I closed this option, and tried add `#force_feed`
they both don't work for me

PR #1805, opened 2018-03-02
Add tooltips on share configuration buttons

aledeg:
Add a tooltip on the add and remove buttons in the share configuration page.

### DIFF
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -126,6 +126,7 @@ return array(
 	),
 	'sharing' => array(
 		'_' => 'Sharing',
+		'add' => 'Add a sharing method',
 		'blogotext' => 'Blogotext',
 		'diaspora' => 'Diaspora*',
 		'email' => 'Email',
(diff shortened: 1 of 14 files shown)


app/i18n/en/conf.php
```diff
@@ -126,13 +126,15 @@
 	),
 	'sharing' => array(
 		'_' => 'Sharing',
+		'add' => 'Add a sharing method',
```
Frenzie, line 129: I think just `Add sharing method` would be better? Shall I change it while adding the Dutch translation?

Alkarex: I do not mind either way. I think both forms are ok.

Frenzie: It's also inconsistent with `Remove sharing method`. ;-)

aledeg: I went with ```Add a sharing method``` because it's undefined since you haven't choose it now, and with ```Remove sharing method``` because it's defined since you know which one you're deleting.
I don't mind if you want to change it.

Frenzie: I'll take a look around at similar interfaces elsewhere.

PR #1808, opened 2018-03-03
Improve translating tools

aledeg:
To ease the use of the translating tools, I've added a formatting method to
keep the output files identicals. This will reduce the amount of time needed
to handle translations.


Frenzie:
@Alkarex was two seconds ahead of me and I got a "Can only review open pull requests." error.

(To approve. :-P)

PR #1813, opened 2018-03-03
New Docker

Alkarex:
Based on Alpine Linux. Size ~78MB.
https://github.com/FreshRSS/docker-freshrss/issues/4
https://github.com/FreshRSS/FreshRSS/issues/520
https://github.com/FreshRSS/docker-freshrss
https://github.com/FreshRSS/docker-freshrss-production


Alkarex:
I would like to redirect both https://github.com/FreshRSS/docker-freshrss and https://github.com/FreshRSS/docker-freshrss-production to this new version. Any comments?

Frenzie:
I'm not entirely convinced it belongs in the main repo but that's not a real objection, just a very slightly raised eyebrow. But yeah, two mini repos is a bit much.

Alkarex:
Having it in the same repository makes the scenarios of using git update + Docker (e.g. for development or testing) easier.
And a bit like with the documentation, it is also easier to maintain the versions in sync (a Dockerfile might change with a new FreshRSS version).
P.S.: Also more convenient for automated builds

Alkarex:
For the record, I initially planned to use [`php:7.2-apache`](https://hub.docker.com/_/php/) as a base image, but the large size and manual handling of library configurations and dependencies made me move to [Alpine](https://hub.docker.com/_/alpine/) instead.

.dockerignore
```diff
@@ -0,0 +1,10 @@
+*/.git
+*/docs
+*/data/config.php
```
Alkarex: Hum, maybe we should exclude all `*/data/`. Will do that tomorrow.

Alkarex:
Merging to try auto-build on https://hub.docker.com/r/freshrss/freshrss/
Let's open a new PR for any remaining change.

Alkarex:
Use cloud.docker.com instead of hub.docker.com, to get better auto-build features:
https://cloud.docker.com/app/freshrss/repository/docker/freshrss/freshrss/general

Alkarex:
@aledeg and @Frenzie Let me know if you have an account on cloud.docker.com so I can add you in the team (already done for @marienfressinaud )

Docker/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## Update
+
+```sh
+# Rebuild an image (see build section above) or get a new online version:
+sudo docker pull freshrss/freshrss
+# And then 
+sudo docker stop freshrss
+docker rename freshrss freshrss_old
```
perrinjerome, line 49: Isn't `sudo` missing here ? ( and two lines below ) or maybe this section should not have `sudo`, assuming user will use `sudo` if it's needed in his environment

Alkarex: Well spotted :-)

Alkarex:
An automatic build is already working on the dev branch. To test:
```sh
mkdir -p ./data/
sudo docker run -dit --restart unless-stopped --log-opt max-size=10m -v $(pwd)/data:/var/www/FreshRSS/data -p 8080:80 --name freshrss freshrss/freshrss:dev
```

Docker/README.md
Alkarex: Fixed in https://github.com/FreshRSS/FreshRSS/pull/1817

Alkarex:
Follow-up in https://github.com/FreshRSS/FreshRSS/pull/1817

Frenzie:
@Alkarex 
> Let me know if you have an account on cloud.docker.com so I can add you in the team

The sign-up is a half-broken nightmare (complains about things without telling you what's wrong, expiring validations, etc.) but at least it didn't complain that my password was too long or contained difficult characters.

Anyway, my username is `frenzie`.

Alkarex:
@Frenzie Done 👍

PR #1829, opened 2018-03-11
Improve i18n tools

aledeg:


Alkarex:
I am not yet so familiar with the i18n tools, so I trust this is fine :-)
Just merge when you think it is ready.

Alkarex (approved):

aledeg:
Actually, that's a good thing since I keep changing it to improve it. :)
I think I will keep it this way for a while now. I am not 100% satisfied, but I think that's good enough for now.

PR #1833, opened 2018-03-12
SVG version of FreshRSS logo + text

Alkarex:
In two versions, one using a font, another one an SVG path.
Tested in Web browsers (Firefox, Chrome, Edge, IE) + Inkscape


Frenzie:
Original lost? Photoshop, not vector? :-)

Alkarex:
I think @marienfressinaud lost the original, yes :-)

marienfressinaud:
I would like to avoid having two versions that look different. I wanted to suggest to keep the SVG-path one, but it looks slightly less well than the font one (my opinion :)). What do you think? Is it not possible to use a SVG version of OpenSans without having to require the font on the computer? (I know barely nothing about SVG :D).

@Frenzie Well, I think to remember that I did it with Gimp ^^'

Alkarex:
It is possible to include glyphs in SVG (kind of including a subset of the font in the SVG file).
But the text rendering will never be pixel perfect between two systems.

Alkarex:
I will prepare (not before tomorrow) a third version which uses text + embedded font, so we can compare.

marienfressinaud:
Ok :+1: if you don't have time to do it before sending to Guénaël, I suggest to use the SVG-path version anyway, it will work for more people… or maybe you can send both and tell him the pros and cons :)

Alkarex:
@marienfressinaud I had already hit the "Send" button (copy to you), including both versions as well as PNG versions.
Edit: I am sure they are used to work with images and will understand :-)

marienfressinaud:
Ah ah, yes I just saw it :D perfect!

Frenzie:
@Alkarex I've noticed that zoom level in Inkscape can affect stuff like object to path. Just to be sure you might want to try zooming in as much as possible.

@marienfressinaud There's also the fact that fonts contain optimizations for linking up with pixels at smaller sizes.

Alkarex:
I am not able to immediately find the status of the support for embedded fonts / glyphs in Inkscape, but my initial search is not so positive :-(
I think Inkscape should be the minimum requirement (or at least another open source software - maybe Krita?), followed by Web browser, and if possible the GitHub renderer (I have just noticed that my current versions are truncated in the GitHub renderer for some reason - maybe a viewport bug).
It of course works in e.g. Illustrator, but that misses the point...

Alkarex:
https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_fonts
I suggest to keep it as it is, due to the low compatibility of embedded fonts

Issue #328, opened 2013-12-23
Liste des lecteurs RSS mobiles


Alkarex:
> Pour utiliser l'API, lire https://github.com/FreshRSS/FreshRSS/issues/443#issuecomment-36666133

Pour aider l'écriture de l'API FreshRSS https://github.com/FreshRSS/FreshRSS/issues/13, faire une liste des lecteurs RSS mobiles (par exemple sous Android https://github.com/FreshRSS/FreshRSS/issues/200, iOS), et en particulier les applications permettant l'utilisation d'une source auto-hébergée :
- News+ https://github.com/noinnion/newsplus
  - https://play.google.com/store/apps/details?id=com.noinnion.android.newsplus
  - gReader https://play.google.com/store/apps/details?id=com.noinnion.android.greader.reader
- PaperBoy https://play.google.com/store/apps/details?id=com.dottedcircle.paperboy
- FeedMe https://play.google.com/store/apps/details?id=com.seazon.feedme
- NewsFeed https://play.google.com/store/apps/details?id=com.smedilepaolo.newsfeed
- Sismics Reader https://play.google.com/store/apps/details?id=com.sismics.reader

yafp:
Hi,
my french is very limited - what clients would you recommend for Android?

Best regards

Alkarex:
@macfidelity I personally use News+ (https://play.google.com/store/apps/details?id=com.noinnion.android.newsplus.extension.google_reader with https://play.google.com/store/apps/details?id=com.noinnion.android.newsplus )

wilfried18:
@macfidelity J'utilise depuis le début easyrss , et je le trouve trés bien (opensource) alors que news+ @Alkarex me contredira si ce n'est pas vrai , mais il semblerais que se soit une application privatrice

Alkarex:
@wilfried18 Yes indeed, News+ is not open source, while my EasyRSS fork is fully open source and without any tracking / advertisement / unneeded permission. EasyRSS is though not (yet) as good, for instance with a less efficient synchronisation mechanism.

yafp:
Thanks for the responses.

Regarding EasyRSS - somehow i dont understand how to set that android application up. It asks just for a "google Reader Account" and a "password" - while i would expect something like
- server
- username
- password

marienfressinaud:
@macfidelity > for EasyRSS you have to use the fork by @Alkarex. It is available at https://github.com/Alkarex/EasyRSS/tree/master/Release

[...]

PR #1836, opened 2018-03-17
Add Fever API and user documentation

### LINKED ISSUE #1406: Make API fever compatible
Lot's of clients (f.e. Reeder) have Fever support. Fever is an old RSS aggregator which is also self-hosted. It would be great if you guys made the API `fever-compatible`, so we can point or Reeder apps, running on our mobile phones, to FreshRSS.

kevinpapst:
Fixed #1406 

This is the import of https://github.com/kevinpapst/freshrss-fever-api with adjusted documentation.
This can for sure be simplified, especially when I can move the classes to the lib folder.
But I need your input first on how to do that and on how to get rid of the custom SQL stuff.

Please start with your input about the basic structure and lets save stuff like grammar fixes ;-) for later.


Frenzie:
At a quick skim I see a a bunch of what looks like custom SQL? I suspect any major feedback would focus primarily on that.

README.md
```diff
@@ -161,6 +161,39 @@ FreshRSS supports further customizations by adding extensions on top of its core
...
+	* [FeedReader 2.0+](https://jangernert.github.io/FeedReader/) (Open source)
+
+## Fever API
+
+You can find more information about this API at our [Fever API documentation](https://freshrss.github.io/FreshRSS/en/users/06_Fever_API.html) page.
+
+Supported clients are:
+
+- iOS
+  - Fiery Feeds
```
Alkarex: Could we have links? I believe the first one is https://itunes.apple.com/app/fiery-feeds-rss-reader/id1158763303 for instance

Alkarex: I can see that Fiery Feeds already supports the Google Reader API. We could try to ask the developer to just add the possibility of typing a server address (like I did for FeedMe recently, which was in the same situation).

README.md
```diff
@@ -161,6 +161,39 @@ FreshRSS supports further customizations by adding extensions on top of its core
...
+  - Unread
```
Alkarex: Seems to be https://itunes.apple.com/app/unread-rss-reader/id1252376153
Same as above, the service seems to already support Google Reader API.

README.md
```diff
@@ -161,6 +161,39 @@ FreshRSS supports further customizations by adding extensions on top of its core
...
+
+- MacOS
+  - Readkit
```
Alkarex: Seems to be https://readkitapp.com.
Same as above, the service seems to already support Google Reader API.

README.md
```diff
@@ -161,6 +161,39 @@ FreshRSS supports further customizations by adding extensions on top of its core
...
+     
+- Android 
+  - Press
```
Alkarex: I cannot immediately find a link. Any success from someone else?

docs/en/users/06_Fever_API.md
```diff
@@ -0,0 +1,114 @@
...
+
+Please point your application to the URL of `fever.php` (e.g. `http://your-freshrss-url/api/fever.php`).
+
+Special client implementation:
+- The Press Android client needs (tested with 1.5.4) needs the additional file `fever-press.php` (use that file as endpoint in the Fever account setting)
+
+There is a drawback when using this plugin: the username and password combination that you have to use.
+You have to set the API password as **lowercased** md5 sum from the string 'username:password' (this limitation exists, because the Fever API was designed in a way which is incompatible with the authentication system used by FreshRSS).
+
+So if you use **kevin** as username and **freshrss** as password you have to set your FreshRSS API password to **4a6911fb47a87a77f4de285f4fac856d** (it MUST be lowercased, some tools seem to create it in uppercase which will not work!).
```
Alkarex: What do the users have to type on the mobile app side? The plaintext password, right? If so, we can do some magic server-side to hide this complexity to the users.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+
+/**
+ * Class FeverAPI_FeedDAO for more feed functions than FreshRSS offers.
+ */
+class FeverAPI_FeedDAO extends FreshRSS_FeedDAO
+{
+	/**
+	 * @return FreshRSS_Feed
+	 */
+	public function getLastUpdatedFeed() {
```
Alkarex: Seems to be very similar to `FreshRSS_FeedDAO::listFeedsOrderUpdate()`, to which we could add a `limit` parameter if needed.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+	 * @return []
+	 */
+	public function countFever()
+	{
+		$values = [
+			'total' => 0,
+			'min' => 0,
+			'max' => 0,
+		];
+		$sql = 'SELECT COUNT(id) as `total`, MIN(id) as `min`, MAX(id) as `max` FROM `' . $this->prefix . 'entry`';
```
Alkarex: It would be good to check that the clients, which we would like to support, accept 64-bit integers for ID.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+			$values = $result[0];
+		}
+
+		return $values;
+	}
+
+	/**
+	 * @return []
+	 */
+	public function getUnread()
```
Alkarex: Seems to be done already by `FreshRSS_EntryDAO::listWhere()`.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+		$stm->execute();
+		$result = $stm->fetchAll(PDO::FETCH_COLUMN);
+
+		return $result;
+	}
+
+	/**
+	 * @return []
+	 */
+	public function getStarred()
```
Alkarex: Seems to be done already by `FreshRSS_EntryDAO::listWhere()`.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+	/**
+	 * FeverAPI constructor executes authentication and initialization.
+	 */
+	public function __construct()
+	{
+		$this->authenticate();
+
+		// are we xml or json?
+		if (isset($_REQUEST["api"]) && strtolower($_REQUEST['api']) === 'xml') {
+			$this->xml = true;
```
Alkarex: All the clients I have tested for the Google Reader API only support JSON. It would be nice to check what the clients, which we would like to support, are actually using. If all of them are using e.g. JSON, we could drop the other serialization.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+			if (method_exists($this, $method_name)) {
+				$id = intval($_REQUEST["id"]);
+				switch (strtolower($_REQUEST["mark"])) {
+					case 'item':
+						$this->{$method_name}($id);
+						break;
+					case 'feed':
+					case 'group':
+						$before = (isset($_REQUEST["before"])) ? $_REQUEST["before"] : null;
+						if ($before > pow(10, 10)) {
```
Alkarex: That is a constant, right? If yes, let's avoid calling an expensive function.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+		$dao = $this->getDaoForEntries();
+		$dao->markFavorite($id, false);
+	}
+
+	/**
+	 * TODO check this method for validity - is this required?
+	 * @param $html
+	 * @return string
+	 */
+	protected function rewriteUrls($html)
```
Alkarex: FreshRSS already does some sanitizing, which might be enough. What is this function trying to do?

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+		$node = $doc->getElementsByTagName('body')->item(0);
+
+		if ($node) {
+			return $doc->saveXML($node);
+		}
+		return $html;
+	}
+
+	/**
+	 * TODO check this method for validity - is this required?
```
Alkarex: Same as above. Let's see if it is actually needed.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+	 * @param int $beforeTimestamp
+	 * @return int
+	 */
+	protected function convertBeforeToId($beforeTimestamp)
+	{
+		// if before is zero, set it to now so feeds all items are read from before this point in time
+		if ($beforeTimestamp == 0) {
+			$before = time();
+		}
+		$before = PHP_INT_MAX;
```
Alkarex, line 586: This can be different between 32-bit and 64-bit systems. FreshRSS IDs are 64 bits.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+		$dao->markReadCat($id, $before);
+	}
+}
+
+// ================================================================================================
+// Start the Fever API handling
+$handler = createFeverApiInstance();
+
+if ($handler->isXmlRequested()) {
+	header("Content-Type: text/xml");
```
Alkarex: The following would probably be better: `Content-Type: application/xml; charset=UTF-8`

Alkarex:
Thanks for this effort @kevinpapst 👍 
Would you mind uploading a log of the calls, if you have any ready? (login + first synchronisation + mark as read + mark as favourite).

kevinpapst:
It might be good if you test it yourself. In the original feature request thread there is a link to an APK for press, its not available through Playstore any longer. Just add that one line from the README to the fever.php to see a log for the calls:
```
file_put_contents(__DIR__ . '/fever.log', $_SERVER['HTTP_USER_AGENT'] . ': ' . json_encode($_REQUEST) . PHP_EOL, FILE_APPEND);
```
An example output looks like that:
```
Dalvik/2.1.0 (Linux; U; Android 5.0.1; GT-I9515 Build/LRX22C): {"api":"","api_key":"my-api-key"}
Dalvik/2.1.0 (Linux; U; Android 5.0.1; GT-I9515 Build/LRX22C): {"api":"","groups":"","api_key":"my-api-key"}
Dalvik/2.1.0 (Linux; U; Android 5.0.1; GT-I9515 Build/LRX22C): {"api":"","feeds":"","api_key":"my-api-key"}
Dalvik/2.1.0 (Linux; U; Android 5.0.1; GT-I9515 Build/LRX22C): {"api":"","unread_item_ids":"","api_key":"my-api-key"}
Dalvik/2.1.0 (Linux; U; Android 5.0.1; GT-I9515 Build/LRX22C): {"api":"","items":"","since_id":"0","api_key":"my-api-key"}
Dalvik/2.1.0 (Linux; U; Android 5.0.1; GT-I9515 Build/LRX22C): {"api":"","items":"","since_id":"50","api_key":"my-api-key"}
```
The API was not very well designed IMO, so for example it only send the md5 hash of the username:password. No plaintext username unfortunately.
If we add fever to the core, we can save the API password twice, once in its current state and one time with the current username and the submitted password as md5 hash. But we would still have to iterate above all registered users. I don't see how that could be solved otherwise.

p/api/fever.php
kevinpapst: This is currently completely wrong, as the request should send a timestamp. We have to convert that to an ID with a select statement. But no client that I tested even used that call...

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+	 */
+	public function process()
+	{
+		$response_arr = array();
+
+		if (!$this->isAuthenticatedApiUser()) {
+			throw new \Exception('No user given or user is not allowed to access API');
+		}
+
+		if (isset($_REQUEST["groups"])) {
```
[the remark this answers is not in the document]
kevinpapst: It does. II re-used $_REQUEST from the original plugin as I assumed that the dev tested a lot of clients and figured that some misbehave. I didn't give it more thoughts.

p/api/fever.php
```diff
@@ -0,0 +1,894 @@
...
+	{
+		if (!isset($_POST["api_key"]) || empty($_POST["api_key"])) {
+			FreshRSS_Context::$user_conf = null;
+			return;
+		}
+
+		$apiKey = $_POST["api_key"];
+		$usersDir = join_path(DATA_PATH, 'users');
+
+		foreach (glob($usersDir . '/*', GLOB_ONLYDIR) as $username) {
```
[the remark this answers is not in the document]
kevinpapst: Nope, the login is passed with every call (called `api_key`).
All clients seem to test it by sending this call:
```
{"api":"","api_key":"my-api-key"}
```
Thats actually a POST to `https://yourdomain/api/fever-press.php?api` with the body send as `form-data` and a key value pair of `api-key` and `my-api-key`

p/api/fever.php
kevinpapst: None that I tested used XML. But XML support is already baked in and works: so why remove it?

p/api/fever.php
kevinpapst: I checked that method but it fetches too much data, executes a join and does not allow to add a limit. But we could add a simple `fetchIdsWhere` method to the core!?

p/api/fever.php
kevinpapst: Same as above

p/api/fever.php
Alkarex: Not very important indeed, but if nobody uses a code, removing it reduces the amount of code to maintain as well as the risks of bugs and other vulnerabilities. If we do not have any client supporting XML, it is not even easy to test that it actually works as intended.

p/api/fever.php
kevinpapst: I checked that method before, but I didn't support limit. If you add that I can switch. Can you push to this PR?

docs/en/users/06_Fever_API.md
kevinpapst: As explained below, we could save two API password but only showing one input in the user profile. Once like it is and once md5 hashed with username:password. Do you want to add that to the PR?

README.md
kevinpapst: No "official" APK in the Playstore existing, but the APK can still be downloaded on other websites. But I wouldn't link there from the FreshRSS website...

README.md
kevinpapst: How can you see that? The APP has nothing in "add account" that sounds like Google Reader. Does one of the supported services use GReader API in the background?
It states "Fiery Feeds syncs with Feedly, Feedbin, Feed Wrangler, Fever, Feedja, NewsBlur, Tiny Tiny RSS, Inoreader, Bazqux, and The Old Reader. "

README.md
kevinpapst: How can you see that? The APP has nothing in "add account" that sounds like Google Reader. Does one of the supported services use GReader API in the background?
It says "Feed Wrangler, Feedbin, Feedly, Fever, Inoreader, NewsBlur"

README.md
kevinpapst: How can you see that? The APP has nothing in "add account" that sounds like Google Reader. Does one of the supported services use GReader API in the background? If so, we should document that FreshRSS is compatible to service XYZ somewhere. 

Its says "ReadKit supports services from Instapaper, Pocket, Readability, Pinboard, Feedly, Fever, NewsBlur, Feed Wrangler and Feedbin and it has its own built-in RSS engine."

README.md
kevinpapst: If you are fine with linking to paid apps, I will add the links

README.md
Alkarex: Feedly, The Old Reader, InnoReader, etc. are all using the Google Reader API, only with minor variations.

README.md
Alkarex: Feedly, The Old Reader, etc. are using the Google Reader API. This is how the other clients supporting FreshRSS are working.

Alkarex: I do not have anything Apple around, so could you please check for which services Fiery Feeds gives the user the option to type a server address?

README.md
Alkarex: I do not have anything Apple around, so could you please check for which services Unread RSS gives the user the option to type a server address?

README.md
Alkarex: It supports e.g. Feedly, which is using the Google Reader API. I do not have anything Apple around, so could you please check for which services Readkit gives the user the option to type a server address?

p/api/fever.php
kevinpapst: Hm, how? All I can say is that the linked clients work.

Alkarex: Then it is good enough :-)

p/api/fever.php
kevinpapst: I don't mind removing it, it makes things easier.

Alkarex: If we cannot find any client using it so yes, I would indeed rather like removing the code

README.md
kevinpapst: FeedWrangler: Email/Password
Feedbin: Email/Password
Feedly: OAuth
Fever: Domain/Email/Password
Inoreader: OAuth
Newsblur: OAuth

README.md
kevinpapst: Instapaper: Username/Password
Pocket: OAuth
Pinboard: Username/Password
Feedly: OAuth
Fever: Domain/Email/Password
NewsBlur: Username/Password
Feedwrangler: Email/Password
Feedbin: Email/Password

p/api/fever.php
kevinpapst: I will remove this, as we have to handle the before timestamp completely different

README.md
Alkarex: Thanks for the info. I am writing to the developers of those 3 apps, to hear whether they could enable a generic service using Google Reader API.

docs/en/users/06_Fever_API.md
Alkarex: I would like to find an approach that scales a bit better. For instance, we could have the username somewhere in the Fever API URL, such as `https://rss.example.net/api/fever.php/alice/`. Another solution could be to have a directory with text files `./data/fever/[md5].txt` and containing the name of respective accounts (similarly to what we do for PubSubHubbub).

p/api/fever.php
Alkarex: Not tested, but https://github.com/FreshRSS/FreshRSS/pull/1836/commits/2c271f0b725dd978b28bf86f964ee7a396e6278b should do the job :-)

p/api/fever.php
Alkarex: Could the existing `FreshRSS_EntryDAO::listIdsWhere()` do the job then?

p/api/fever.php
kevinpapst: Works, thanks! I don't really get the calculation happening there as I had to use `listFeedsOrderUpdate(-1, 1)` - but I guess thats not a problem?

docs/en/users/06_Fever_API.md
kevinpapst: If you come up with a solution that you prefer, please let me know. I can add it, but I don't want to change the core without some guidance. Privacy and security is a very important issue and I don't want to decide that by myself.

Alkarex: Yes, I will do so. Let's check that the rest works fine first. I am slowed down by family activities for a couple of weeks

kevinpapst: Me too, sorry for not being more active, but free time is rare atm

p/api/fever.php
Alkarex: Could you please give a try without these two sanitising/rewrite functions, and see if it still works as intended with your client?

dylan-k:
Will this include the Hot Links feature that Fever and its API used to provide?

kevinpapst:
I don't know what this feature is/was. I guess there is no equivalent in FreshRSS, so likely no.
But you seem to have used it in the past, can you explain what "Hot Links" is?

dylan-k:
The feature's the reason it's called Fever and is arguably its salient feature.

As for what it is/does though... The Fever homepage says:

> Fever reads your feeds and picks out the most frequently talked about links from a customizable time period. Unlike traditional aggregators, Fever works better the more feeds you follow.

Essentially the feature looks through the content of all the feeds and notes how many times a link has been mentioned recently. The more often the blogs are mentioning a link recently, the more "hot" that link is. One early review of Fever described it, "Besides offering a full-featured feed reader, the application attempts to create a personalized Techmeme by scanning a user’s feed list for popular (or hot) links. Fever then groups these links into stories and assigns each a “temperature.” This allows a user to quickly keep a pulse on what’s going on in his or her “slice of the web.”

The Fever API documentation has an endpoint for it, documented under the "hot links" heading on the page at https://feedafever.com/api

**EDIT**: I'm not sure I'm skilled enough to do a viable PR, but I do have some more detail about [how the original "hot links" feature works](https://github.com/passiomatic/coldsweat/issues/68#issuecomment-272963268), in case that's useful.

kevinpapst:
Thanks for the explanation @dylan-k . In this case I stick with my "no". 
Let me quickly explain why: this PR was never meant to implement Fever logic but only making FreshRSSS compatible with the most important features of the Fever API. I wanted to finally use FreshRSS with iOS apps that did not support Greader compatible endpoints by now. 
If you want to Hot links feature, please open a PR for FreshRSS. Then it could be integrated into the API as well. But the Fever API itself will not offer any feature that is beyond the scope of the FreshRSS core.

p/api/fever.php
kevinpapst: Still works (as expected) without. These function were just small helper to convert a text url to a clickable link and some function to add a target="_blank" to links. I don't think it necessary.

Alkarex:
Indeed, it could be an interesting feature to add into FreshRSS itself (new issue), which could then be supported by the APIs

kevinpapst:
Guys, is anyone still interested in helping me with the authentication issue? You wanted me to add this to the core, so lets rock together ;-)
I do understand that an Android/Linux has probably no big interest if there are enough clients to choose from, but for all users living with Mac/iOS this new API is a major change.

Frenzie:
I don't really know what @Alkarex has in mind. :-)

Alkarex:
Sorry for the delay. I will try to look at the login part in the coming
days. For the rest, as I do not have any Apple thing to test (and there is
apparently no Android app alive with this API), I would be pleased to get
some logs and other info, in particular regarding performances &
scalability.

Alkarex:
@kevinpapst I have made a draft of the rework of the login part https://github.com/FreshRSS/FreshRSS/pull/1836/commits/76f892f22ab098c49c467b46569179655995d306
Not tested yet. Feedback welcome already.

TODO:
* Delete the token when a user is deleted
* French translations

data/fever/index.html
```diff
@@ -0,0 +1,13 @@
+﻿<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en-GB" lang="en-GB">
```
kevinpapst, line 2: html5 doctype vs xhtml

Alkarex: Any problem? This is correct as far as I can see.

Frenzie: Not really. This is exactly how you're supposed to do XHTML5. Is there a particular reason for using it though?

kevinpapst: I was just reading through your changes (thanks for then btw!) and stumbled upon this. Was just curious why you would use an xhtml declaration nowadays.

Alkarex: Please note that HTML vs. XHTML is not the same question than to use HTML5 vs. HTML4 or XHTML1.

XHTML can be parsed as standard XML easily in almost any language/platform, which is not the case for HTML, even less if it is supposed to be in a consistent manner. This makes XHTML easier to consume. The fact that it can optionally be served strictly makes it easier to spot errors, at least during development. In deployment, one can serve it as HTML if one does not desire its strictness.

Alkarex: * https://html.spec.whatwg.org/multipage/xhtml.html
* https://www.w3.org/TR/html5/xhtml.html
* https://www.w3.org/TR/html-polyglot/#doctype

Alkarex:
I have tried with an old copy of Press on Android, and could not get the sync to work.
The login seems to work, but then there is several hundred requests like:

```
POST /.../api/fever-press.php/?api&items&since_id=29050
```

where ID does not even seem to be ordered.

p/api/fever-press.php
```diff
@@ -0,0 +1,54 @@
...
+					return ' LIMIT 50';
+				} else if (!empty($max_id)) {
+					return ' LIMIT 50';
+				}
+
+				if ($since_id < 0) {
+					$since_id = 0;
+				}
+
+				return ' LIMIT '.intval($since_id).', 50';
```
Alkarex: This syntax is MySQL-specific. I am not sure we will keep this file though, so no hurry in adding SQLite + PostgreSQL.

Alkarex:
I would be pleased to include this PR in FreshRSS 1.11, which I would like to release in the coming days.
I need some help in particular:
* To get some info (logs of a full sync) from iOS
* To know whether Press on Android can work or not

data/fever/index.html
kevinpapst: Besides the X/HTML topic, couldn't we use a simple one liner index.php instead?
<?php header('Location: /'):

Alkarex: This file is not even supposed to be used. It is mainly there to avoid directory indexing. We have the same template in several other directories. A one-line PHP 301 or 302 redirect is not valid, as per HTTP/1.1 which requires sending a proper body, and `index.html` is more commonly used in the list of default files than `index.php`.

Edit: HTTP/1.1 uses a *SHOULD* for the body: https://tools.ietf.org/html/rfc2616#section-10.3.2

> Unless the request method was HEAD, the entity of the
>   response SHOULD contain a short hypertext note with a hyperlink to
>   the new URI(s).

kevinpapst:
Press somewhat works, but only when using that extra php file as endpoint. It has a crappy API implementation, see this discussion: https://github.com/kevinpapst/freshrss-fever-api/issues/1

I am going to check all the changes today and catch up later

p/api/fever-press.php
kevinpapst: I only added it, as many requests by Android users where pointing to Press. I don't need it and for Android there seem to be much more/better clients with greader support, so I don't mind if we remove it.

kevinpapst:
@Alkarex the credential saving should to be changed. Each time I save a new API password, a new credential file is saved to the data/fever/ directory. So old API passwords do not expire. 
Any ideas on that?

Alkarex:
@kevinpapst Yes, that is very important indeed. I will address that at the same time as a few remaining TODOs for the login, such as removing the file when the user is deleted.

Alkarex:
@kevinpapst Done https://github.com/FreshRSS/FreshRSS/pull/1836/commits/af4cac7871799dc89d3dae91f2b896eef2f5972e
I have changed the filenames of the keys, so please set your API password once again.

kevinpapst:
Works, thanks @Alkarex! 

And here is the full log of me adding my instance as new "Fever account" in the iOS app Unread:
```
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":""}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","api_key":"XXXXXXXXXXXX"}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","feeds":"","api_key":"XXXXXXXXXXXX"}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","groups":"","api_key":"XXXXXXXXXXXX"}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","saved_item_ids":"","api_key":"XXXXXXXXXXXX"}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","unread_item_ids":"","api_key":"XXXXXXXXXXXX"}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","items":"","with_ids":"1526824296671940,1526824296671939,1526824296671938,1526824296671937,1526723498489876,1526665886926073,1526540594321730,1525710669342897,1524898620380741,1524898620380739","api_key":"XXXXXXXXXXXX"}
Unread/109043 CFNetwork/897.15 Darwin/17.5.0: {"api":"","items":"","with_ids":"1526003494262265,1520445116936397,1515351114757028,1513983113542001,1509540716471827,1502988666500755,1502650338543063,1500728700880224,1500303076866879,1498834266516508,1498186281346581,1494715880653072,1492138280637078,1492008668716590,1491785493526772,1489114280923150,1489063865764453,1488739908671010,1486522282136870,1483642327503590,1483534265485953,1482843065383606,1480708834916255,1480624974438655,1480624877408917,1415229480578865","api_key":"XXXXXXXXXXXX"}
```
I created the log by adding this to the top of the fever.php:
```
file_put_contents('/tmp/fever.log', $_SERVER['HTTP_USER_AGENT'] . ': ' . json_encode($_REQUEST) . PHP_EOL, FILE_APPEND);
```

What is still missing for you? I think this can still be improved, but I'd like to get some feedback from real users ...

Alkarex:
@kevinpapst Thanks. That logs looks much better than with Press. By the way, did you manage to get Press to work? If yes, does it still works with the latest changes? Mainly by curiosity and for testing, as I indeed believe it is better not to propose it officially. I suggest we keep the Press-specific code for now, but remove mentions from the documentation.

Yes, let's try to get a few test-users. We can soon merge in /dev to ease testing.

p/api/fever.php
```diff
@@ -0,0 +1,649 @@
...
+			$response_arr["unread_item_ids"] = $this->getUnreadItemIds();
+		}
+
+		if (isset($_REQUEST["saved_item_ids"])) {
+			$response_arr["saved_item_ids"] = $this->getSavedItemIds();
+		}
+
+		if (isset($_REQUEST["mark"], $_REQUEST["as"], $_REQUEST["id"]) && is_numeric($_REQUEST["id"])) {
+			$method_name = "set" . ucfirst($_REQUEST["mark"]) . "As" . ucfirst($_REQUEST["as"]);
+			if (method_exists($this, $method_name)) {
```
Alkarex: I am a bit worried about calling methods based on user input. I would rather like to test (e.g. with a switch) against a white-list of acceptable methods, instead of solely `method_exists()`.

p/api/fever.php
```diff
@@ -0,0 +1,649 @@
...
+		return implode(',', array_values($ids));
+	}
+
+	/**
+	 * @return string
+	 */
+	protected function getUnreadItemIds()
+	{
+		$dao = $this->getDaoForEntries();
+		$entries = $dao->listIdsWhere('a', '', FreshRSS_Entry::STATE_NOT_READ, 'ASC', PHP_INT_MAX);
```
Alkarex: The last parameter `limit` can be set to zero to disable it

kevinpapst:
I haven't tried Press lately. It took quite a long time until everything was indexed. As fas as I remember, it tried to query every single article ID for indexing all articles that FreshRSS knows.

Even though Press was mentioned by some of our users, its development stopped a while ago and its only available as APK download from some more-or-less trustable sources. At least thats what I found while searching for it.

So I would rather go the opposite way and:
- remove the press specific stuff and simplify the codebase (some "nasty workarounds" were only added to get it to work)
- remove every mention in the docu
- probably even mention that it does not work for its wrong usage of the API

If thats okay for you @Alkarex I could probably finalize the PR today

In addition: what do you think about extracting some of the methods from greader.php into a lib_api.php file? Methods like `debugInfo()`, `serviceUnavailable()` ... could be useful for both APIs.

Alkarex:
@kevinpapst That sounds like a good plan indeed 👍 
Let me have a look about mutualised methods a bit later today

p/api/fever.php
```diff
@@ -0,0 +1,638 @@
...
+	 * @param string $prefix
+	 * @param array $values
+	 * @param array $bindArray
+	 * @return string
+	 */
+	protected function bindParamArray($prefix, $values, &$bindArray)
+	{
+		$str = "";
+		foreach ($values as $index => $value) {
+			$str .= ":".$prefix.$index.",";
```
Alkarex: Could you please cast `$index` to int (or make a normal `for` loop)? I am worried there might be ways of arriving here with an associative array in `$values` and a risk of SQL injection.

Alkarex:
@kevinpapst I guess you developed for MySQL, didn't you? Did you have a chance to test with SQLite and/or PostgreSQL too?

Alkarex:
@kevinpapst I have made a few changes. Could you please have a quick look? Then I suggest we merge and ask for additional testing in the /dev branch.

kevinpapst:
Sorry, I have not much time during the week, so I can currently only work on this at the weekend.
Lets be "agile" and fix upcoming stuff in /dev - the main use cases are definitely working ... even though I had no time to test the different DBs

Your changes look good and thanks for your ongoing support!

Alkarex:
@kevinpapst I have simplified the reverse key check, so for testers of this PR, one needs to re-set the API password once again.
Proceeding to merging in /dev and call for more testing.

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1861
https://github.com/FreshRSS/FreshRSS/issues/328

Issue #1811, opened 2018-03-03
[Feature Request] Allow to ignore SSL certificate errors on a feed


perrinjerome:
Some feeds are hosted on servers with invalid SSL certificates, it would be nice to be able to configure in the feed parameters that problems with SSL certificate while updating the feed can be ignored.

Alkarex:
Usecase for https://github.com/FreshRSS/FreshRSS/pull/1838

papaxris:
This SSL certificate ignore feature seems to be working.  I consistently had issues with a particular feed because of their invalid certificate.  This has allowed the feed to populate within the FreshRSS feeds.

Logging the SSL issue is still maintained and is not simply ignored (thank you).

Issue #1702, opened 2017-12-02
Mark as read when updating feeds


Liandriz:
Hi, just a question to know if a feature is actually feasible, with or without plugin : mark as read all the posts from a feed, when updating them. 

This is for some feeds i archive to search later on them, not reading them when they are published.
So is it possible to do that please?

Alkarex:
Hello @Liandriz 
Yes; see the last option:

![image](https://user-images.githubusercontent.com/1008324/33515516-127b6b4a-d765-11e7-84ef-13c746215a03.png)

Alkarex:
Ah sorry, maybe you meant for only one specific feed?

Liandriz:
Yes, didn't see it cause i didn't check in global options, but this is it, but per feed, is it hard to add by plugin?  👍

Alkarex:
@Liandriz Making an extension is not so difficult.
See the latest one as an example https://github.com/kevinpapst/freshrss-dilbert/blob/master/xExtension-Dilbert/extension.php from @kevinpapst
Different "read" options per feed could become a FreshRSS feature in the future, but it is not something we have right now (easy to do, the main part is to translate the title of the option in the different languages).

Liandriz:
You really don't need any traduction : Just add a section on feed options page, with title "Reading", same that is in global section, and same for options, something like this (maybe not the 4 options) : 
![feed options](https://user-images.githubusercontent.com/563539/33525193-36f63920-d82b-11e7-9244-82ab50c72ace.png)

Alkarex:
Usecase for https://github.com/FreshRSS/FreshRSS/pull/1838

Issue #891, opened 2015-07-05
Control of article updates per feed


Alkarex:
Add a DB column and a UI to control for each feed how to react to article updates
https://github.com/FreshRSS/FreshRSS/issues/798#issuecomment-107078678

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/1838

PR #1838, opened 2018-03-18
JSON column for feeds

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1654


Alkarex:
Lacks auto-create column for SQLite and PostgreSQL.

aledeg (commented):
I am not very keen of the column name but apart from that, good job!

Alkarex:
@aledeg Any other name suggestion?

aledeg:
I prefer to have the full name like ```attributes```. We're not in the 80s when you needed to shorten everything to 8 chars :)

Alkarex:
@aledeg What about `json` as a name, then? Short, and with additional information

aledeg:
I think ```json``` is good, but it lacks content description. My 2¢.

Frenzie:
I agree with @aledeg. `json` could be "anything." I have no real objection to something like `attr`, although I don't see why two characters more should count against `attributes`. :-)

Alkarex:
This JSON column will actually be with anything inside :-)
So what do you vote for?
1. attribs
2. attributes
3. json
4. Yet another

I think I can now better like `json`, as it provides a bit more information, which we lack from the type (we cannot use the proper JSON types unfortunately due to poor compatibility)

Frenzie:
json_attributes, jattr, j_attr, jsonAttributes
json_settings
json_properties

aledeg:
```jsonAttributes```?

**Comment by Alkarex** [comment]
Ok, go for `attributes` before it gets even longer then :P

@aledeg Didn't you make the code to auto-add a column in PostgreSQL when you did the first version of the mute feed thing? I cannot find it anymore due to the squash inside the pull request. If you can find it again, I am interested :-)

**Comment by aledeg** [comment]
@Alkarex I am not sure I've done such thing. And if I did, I am afraid I've deleted it. Sorry

**Comment by Alkarex** [comment]
Tested with success in MySQL, SQLite, PostgreSQL

File: app/Controllers/feedController.php [hunk 1]
```diff
@@ -365,15 +367,18 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 						// This entry should not be added considering configuration and date.
 						$oldGuids[] = $entry->guid();
 					} else {
+						$read_upon_reception = $feed->$attributes['read_upon_reception'] !== null
```
Alkarex: @Frenzie Do you know how those lines are supposed to be written to pass Travis / PHPCS?
I have already tried two variations without success...

Alkarex:
* Feed cURL timeout
* Mark updated articles as read #891
* Mark as read upon reception #1702
* Ignore SSL (unsafe) #1811

app/Controllers/feedController.php
Frenzie: Unfortunately it wants it all to be on one line like this:

```diff
diff --git a/app/Controllers/feedController.php b/app/Controllers/feedController.php
index eb82a797..aa322a8b 100755
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -353,9 +353,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
                                                } else {        //This entry already exists but has been updated
                                                        //Minz_Log::debug('Entry with GUID `' . $entry->guid() . '` updated in feed ' . $feed->id() .
                                                                //', old hash ' . $existingHash . ', new hash ' . $entry->hash());
-                                                       $mark_updated_article_unread = $feed->$attributes['mark_updated_article_unread'] !== null
-                                                               ? $feed->$attributes['mark_updated_article_unread']
-                                                               : FreshRSS_Context::$user_conf->mark_updated_article_unread;
+                                                       $mark_updated_article_unread = $feed->$attributes['mark_updated_article_unread'] !== null ? $feed->$attributes['mark_updated_article_unread'] : FreshRSS_Context::$user_conf->mark_updated_article_unread;
                                                        $needFeedCacheRefresh = $mark_updated_article_unread;
                                                        $entry->_isRead(FreshRSS_Context::$user_conf->mark_updated_article_unread ? false : null);      //Change is_read according to policy.
                                                        if (!$entryDAO->inTransaction()) {
@@ -367,9 +365,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
                                                // This entry should not be added considering configuration and date.
                                                $oldGuids[] = $entry->guid();
                                        } else {
-                                               $read_upon_reception = $feed->$attributes['read_upon_reception'] !== null
-                                                       ? $feed->$attributes['read_upon_reception']
-                                                       : FreshRSS_Context::$user_conf->mark_when['reception'];
+                                               $read_upon_reception = $feed->$attributes['read_upon_reception'] !== null ? $feed->$attributes['read_upon_reception'] : FreshRSS_Context::$user_conf->mark_when['reception'];
                                                if ($isNewFeed) {
                                                        $id = min(time(), $entry_date) . uSecString();
                                                        $entry->_isRead($read_upon_reception);
```

So for legibility it might be better to use an if/else statement. I could also try to find out if there's a way to coax PHPCS into accepting this if you like but I don't know if that's an option.

Frenzie:
One workaround would be to simply remove the `<rule ref="Squiz.WhiteSpace.OperatorSpacing"/>` rule. Doing so will fail to detect incorrect spacing where we would want it to but it's not the biggest issue.

There are a few more targeted approaches:

```diff
diff --git a/phpcs.xml b/phpcs.xml
index d79e2a4d..1d10d277 100644
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -92,7 +92,9 @@
        <!-- Do not add spaces when casting -->
        <rule ref="Squiz.WhiteSpace.CastSpacing"/>
        <!-- Operators must have a space around them -->
-       <rule ref="Squiz.WhiteSpace.OperatorSpacing"/>
+       <rule ref="Squiz.WhiteSpace.OperatorSpacing">
+               <exclude-pattern>./app/Controllers/feedController.php</exclude-pattern>
+       </rule>
        <!-- Do not add a whitespace before a semicolon -->
        <rule ref="Squiz.WhiteSpace.SemicolonSpacing"/>
        <!-- Do not add whitespace at start or end of a file or end of a line -->
```

If we updated to a newer PHPCS we could also selectively ignore rules for a few lines instead of for whole files ([docs](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file)):

```php
// phpcs:disable Generic.Commenting.Todo, PSR1.Files.SideEffects -- Because reasons
breaking_code();
// phpcs:enable Generic.Commenting.Todo, PSR1.Files.SideEffects
```

However, updating it would require some effort as well, also due to indentation issues.

That being said, in our current version we can still do this (i.e., selectively disable all checks):

```diff
diff --git a/app/Controllers/feedController.php b/app/Controllers/feedController.php
index eb82a797..4f478994 100755
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -353,9 +353,11 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
                                                } else {        //This entry already exists but has been updated
                                                        //Minz_Log::debug('Entry with GUID `' . $entry->guid() . '` updated in feed ' . $feed->id() .
                                                                //', old hash ' . $existingHash . ', new hash ' . $entry->hash());
+                                                       // @codingStandardsIgnoreStart
                                                        $mark_updated_article_unread = $feed->$attributes['mark_updated_article_unread'] !== null
                                                                ? $feed->$attributes['mark_updated_article_unread']
                                                                : FreshRSS_Context::$user_conf->mark_updated_article_unread;
+                                                       // @codingStandardsIgnoreEnd
                                                        $needFeedCacheRefresh = $mark_updated_article_unread;
                                                        $entry->_isRead(FreshRSS_Context::$user_conf->mark_updated_article_unread ? false : null);      //Change is_read according to policy.
                                                        if (!$entryDAO->inTransaction()) {
```

Alkarex:
@Frenzie Thanks for the feedback and for trying to find a better rule. I think the current workaround is good enough for now, while waiting for a potential better approach later.
Let me know if you have a chance to test the new features of this PR :-)

Frenzie:
@Alkarex Test in what sense exactly? Just some basic use it for a few days or actually trying to find where it might break? (The first is easier to squeeze in somewhere, naturally. :-P)

Alkarex:
@Frenzie I only meant a quick look / run before merging, as further testing by more people can also be done once landed in /dev. I am already running this patch on my main instance.

Frenzie:
@Alkarex Nothing obviously broken so far. :-P

PR #1847, opened 2018-03-28
Add \ for global namespace

ColonelMoutarde:


Alkarex:
Hello @ColonelMoutarde , and thanks for your contributions 👍 
Could you please try to make a single pull request, instead of several related ones?
* https://github.com/FreshRSS/FreshRSS/pull/1851
* https://github.com/FreshRSS/FreshRSS/pull/1852
Hint: 1) fork, 2) make a new branch for one specific change, 3) make all the needed changes in that branch, 4) make a pull request from that branch into FreshRSS/dev. Just ask if you need more guidance.

Also, please try to add a little message explaining why the change is good / needed. (Vous pouvez écrire en français si vous n'êtes pas à l'aise en anglais).

For instance, in the current case of the backslash for the global namespace, I am not sure this improves anything, does it? Indeed, the context is already a global namespace.

ColonelMoutarde:
Bonjour,
A partir de PHP7, le moteur de php vas d'abord vérifier si la class Excection existe dans le namespace courant. Sinon il vérifiera la présence dans le namespace global : le " \ ". On gagne donc du temps machine en lui indiquant dès le départ dans quel namespace chercher.

Pour l'ajout de la visibilité, ce n'est que pour rendre cohérentes les classe dans les quelles le l'ai fait, en effet une partie des méthodes écrites plus haut affichaient la visibilité. C'est aussi considéré comme une bonne pratique de développement.

Frenzie:
Could you share a link to a reference about relevant upgrading tips related to PHP7? (No problem if it's in French.) I glanced over at some of the nice new features in PHP 7+ but I haven't looked into potential downsides to older patterns.

Edit: btw, you can just distinguish separate commits within the same PR if they're all about the same thing.

ColonelMoutarde:
backslah is compatible with php5.6 but no performance gain

ColonelMoutarde:
http://php.net/manual/fr/language.namespaces.faq.php#language.namespaces.faq.shortname2

Frenzie:
That page doesn't say anything about performance though, does it? It looks like it just explains how it works should you choose to use it.

ColonelMoutarde:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/
the conclution:
"As expected, the fully qualified global function call is the fastest one. This is because PHP does not need to go through the INIT_NS_FCALL_BY_NAME opcode. When calling the global function in a non-fully qualified way, it is slower. Running functions inside a namespace are always slower then running global functions."

Alkarex:
The article says:

> When calling functions in a namespaced context, additional actions are triggered in PHP which result in slower execution

We currently do not use namespaces in FreshRSS (with a few minor exceptions, which could be removed), so it seems to be a good reason for continuing like that. Isn't it? In which case, this patch is probably not needed.

ColonelMoutarde:
You are in namespace since PHP 5.3.0 By default, the global namespace -> \

Alkarex:
> You are in namespace since PHP 5.3. By default the global namespace -> \

I understand that, and from what I can see, as long as we stick to that global namespace (i..e not using custom namespaces), we should be good, and without needing the `\` prefix everywhere (which would probably mean in many places). Isn't it?

ColonelMoutarde:
en français... ...j'ai pas tout compris svp

kevinpapst:
My vote goes also for a consistent codebase. If that should be changed in the future, adapt all places at once.

ColonelMoutarde:
OK, i need help :)

ColonelMoutarde:
For your information:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048

https://speakerdeck.com/nicolasgrekas/tirer-le-maximum-du-moteur-php-7-lexemple-de-symfony

Alkarex:
@ColonelMoutarde Merci pour les liens, qui semblent conforter ma conclusion un peu plus haut.

1) https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048#issue-256982855

> php-cs-fixer should have a rule to list those functions and enforce adding this \ **when used inside a namespaced file/class**.

2) https://speakerdeck.com/nicolasgrekas/tirer-le-maximum-du-moteur-php-7-lexemple-de-symfony

The example is

```php
namespace App;
\count([123]);
```

See `namespace` (which we - almost - do not use).

Refers also to https://github.com/symfony/symfony/pull/25854/files#diff-75877649dec82c1428deacf1ef0978baR12

```php
namespace Symfony\Component\DependencyInjection\Compiler;
...
    if (\is_array($value)) {
```

Dans les deux cas, il est bien précisé que le problème de performance survient lorsqu'on utilise des noms de domaines explicites, ce qui n'est pas le cas de FreshRSS :-)

Alkarex:
If there is no more comment, I will close this issue, keeping it as a reference as to why we do not use namespaces in FreshRSS for the time being.

S'il n'y a plus d'autres commentaires, je ferme ce ticket, qui pourra être utilisé comme référence sur pourquoi nous n'utilisons pas les espaces de nom dans FreshRSS pour le moment.

PR #1848, opened 2018-03-28
Add visibility "public"

ColonelMoutarde:


Alkarex:
Hello @ColonelMoutarde and thanks again 👍 
Please see my comment on https://github.com/FreshRSS/FreshRSS/pull/1847#issuecomment-377693139 as it also applies there.
* https://github.com/FreshRSS/FreshRSS/pull/1849

The default visibility of PHP class methods is already *public*, so I am unsure of the added value, and a little explanation would help.

Alkarex:
Similar reasoning than https://github.com/FreshRSS/FreshRSS/pull/1847#issuecomment-384594048 : if we were to use explicit `public`, we should do it everywhere and not only in one file.
The added value is unclear, and it does make the code heavier.
I am closing if there is no more comment.

Raison similaire à https://github.com/FreshRSS/FreshRSS/pull/1847#issuecomment-384594048 : si nous nous métions à utiliser `public` explicitement, il faudrait le faire partout de manière cohérente, et pas seulement dans un fichier.
De plus, la valeur ajoutée n'est pas claire, et cela par contre rend le code plus lourd.
Je ferme s'il n'y pas d'autres commentaires.

PR #1857, opened 2018-04-01
fix update user for empty username

kevinpapst:
Minor fix for the page `/i/?c=user&a=manage` where you could previously hit the Update button and get a 500 error:
```
PHP message: PHP Fatal error:  Uncaught Error: Call to a member function save() on null in /xxx/app/Controllers/userController.php:67
```

I added a minor validation to prevent that.

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -76,10 +76,14 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 
 			$apiPasswordPlain = Minz_Request::param('apiPasswordPlain', '', true);
 
+			$ok = false;
+
 			$username = Minz_Request::param('username');
-			$ok = self::updateUser($username, $passwordPlain, $apiPasswordPlain, array(
+			if (!empty($username)) {
+				$ok = self::updateUser($username, $passwordPlain, $apiPasswordPlain, array(
 					'token' => Minz_Request::param('token', null),
 				));
+			}
 
 			if ($ok) {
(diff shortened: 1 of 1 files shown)


app/Controllers/userController.php
```diff
@@ -76,10 +76,14 @@ public function updateAction() {
 
 			$apiPasswordPlain = Minz_Request::param('apiPasswordPlain', '', true);
 
+			$ok = false;
+
 			$username = Minz_Request::param('username');
-			$ok = self::updateUser($username, $passwordPlain, $apiPasswordPlain, array(
+			if (!empty($username)) {
```
Frenzie: Would the check make more sense at the top of updateUser() itself? Perhaps similar issues could occur elsewhere.

```php
if (empty($user) {
  return false;
}
```

kevinpapst: But then the user won't get the visible feedback (see below for the `Minz_Request::bad()` call).
Depends on the wanted behavior in a case where the user submits an empty form.

Frenzie: I said `updateUser()`, ergo line 48. ;-)

(Not saying it's the way to go, but I definitely wasn't suggesting to avoid feedback.)

kevinpapst: Right, misunderstood your suggestion. Thats better, as the function is used from multiple places.

Alkarex:
I have made a slight change, checking the user config object instead: https://github.com/FreshRSS/FreshRSS/pull/1857/commits/6fda58435101cfcef911669779ad9da73f8b7285

Issue #1841, opened 2018-03-20
Error 403 when changing user password


dylanbob:
Hi
I just installed FreshRSS Version 1.10.2 on : Debian 9 using up to date MySQL, php, Apache as a webserver in http, nginx as a reverse proxy in https. I use the auth form to login.
Problem :
Changing my user password (currently connected and admin) on the profile configuration page /i/?c=user&a=profile return a 403 error and disconnect yourself from the service. The password gets changed though.

It is a bit misleading. If you change your password while having deactivated the login, you do not get the error and get a success message like "Your password has been succesfully updated"

I was personnaly trying to set an API password, but firefox automatically filled the first password field (to change the user password) while I wrote my API password on the second field
![Trying to change the API password but the browser automatically filled the first password field](https://user-images.githubusercontent.com/25272703/37683255-ea71b482-2c83-11e8-993f-9481c769a974.png)
![get the error message](https://user-images.githubusercontent.com/25272703/37683271-fbc8dce2-2c83-11e8-97d6-3c643611f679.png)
Nginx and apache logs do not get any useful information. FreshRSS logs do not show anything.

Alkarex:
Hum, we could consider not killing the session after a password change, but if so, it needs to be done carefully. In particular, we still want to kill other potentially open sessions with the same account

pled:
Hum, I had a similar issue with Firefox (59 on Ubuntu) : while trying to set an API password (for EasyRSS use), the first password field (for form) was already filled. Setting the API pwd results in being disconnected (error 403 - Forbidden), and API pwd was not set. Even clearing the first pwd before failed as well. Only way to make it works was using Chrome. Strange !

Alkarex:
@pled I will implement the `autocomplete="new-password"` approach suggested by https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

nikoss:
Alive and kicking in...
   Android 7.0
   Freshrss 1.11.2
   Firefox 62.0.3
   Lightning 4.5.1 (browser) 

Of course, I tested it using the desktop web interface of FreshRSS.

It killed the sessions in both browsers. That is, when I tried it in Lightning, I got logged out. Then I switched to Firefox to find out that I was logged out in there, as well! 

Maybe the message could be more clear. Something like...
   "
    Session killed
    Error 403- blah blah
    You don't have blah blah...
   "
[...]

PR #1877, opened 2018-05-01
Use HTML autocomplete new-password

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1841#issuecomment-376551901
autocomplete="new-password" for user management pages

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -537,7 +537,7 @@ function printStep2() {
 		<div class="form-group">
 			<label class="group-name" for="default_user"><?php echo _t('install.default_user'); ?></label>
 			<div class="group-controls">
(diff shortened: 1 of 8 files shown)


Frenzie (approved):

kevinpapst:
> though support for this has not been implemented in all browsers yet

I did not check the documentation of all alternative browsers, but:
when considering the browser market-share I would vote for sticking with an attribute that works in ALL browsers.

Alkarex:
Thanks to the way unknown values for this parameter are processed, this
change is supposed to be fine in all browsers. And for sure, the previous
"off" value is very problematic in Firefox, and problematic in Chrome.
In any case, tests welcome 🤓

>

kevinpapst:
:D so the fallback for invalid/unknown values is always "off"?

Alkarex:
@kevinpapst Apparently yes, and an even stronger "off" than the real "off".

Alkarex:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9847360/

Alkarex:
Let's continue the testing in /dev to have a broader audience

Alkarex:
Follow up of https://github.com/FreshRSS/FreshRSS/issues/880
https://github.com/FreshRSS/FreshRSS/pull/881

PR #1879, opened 2018-05-04
[i18n] Add hint about adding #force_feed to URL

### LINKED ISSUE #1878: Pippa.io podcast feeds cannot be added
Hello,

I've tested with two shows and FreshRSS 1.10.1 on PHP 7.1.17:
https://shows.pippa.io/btln
https://shows.pippa.io/brise-glace

Two errors are raised in the logs:

```
Unknown error for feed [https://feed.pippa.io/public/shows/brise-glace]
```
```
A feed could not be found at `https://feed.pippa.io/public/shows/btln`; 
the status code is `200` and content-type is `application/xhtml+xml; charset=utf-8` [https://feed.pippa.io/public/shows/btln]
```

Thanks in advance

Frenzie:
Fixes #1878

### DIFF
--- a/app/i18n/fr/feedback.php
+++ b/app/i18n/fr/feedback.php
@@ -72,7 +72,7 @@ return array(
 			'already_subscribed' => 'Vous êtes déjà abonné à <em>%s</em>',
 			'deleted' => 'Le flux a été supprimé.',
 			'error' => 'Une erreur est survenue',
-			'internal_problem' => 'Le flux ne peut pas être ajouté. <a href="%s">Consulter les logs de FreshRSS</a> pour plus de détails.',
+			'internal_problem' => 'Le flux ne peut pas être ajouté. <a href="%s">Consulter les logs de FreshRSS</a> pour plus de détails. Vous pouvez essayer de forcer l’ajout par addition de <code>#force_feed</code> à l’URL.',
 			'invalid_url' => 'L’url <em>%s</em> est invalide.',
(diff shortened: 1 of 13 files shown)


aledeg (approved):

app/i18n/fr/feedback.php
```diff
@@ -72,7 +72,7 @@
 			'already_subscribed' => 'Vous êtes déjà abonné à <em>%s</em>',
 			'deleted' => 'Le flux a été supprimé.',
 			'error' => 'Une erreur est survenue',
-			'internal_problem' => 'Le flux ne peut pas être ajouté. <a href="%s">Consulter les logs de FreshRSS</a> pour plus de détails.',
+			'internal_problem' => 'Le flux ne peut pas être ajouté. <a href="%s">Consulter les logs de FreshRSS</a> pour plus de détails. Vous pouvez essayer de forcer l’ajout par addition de <code>#force_feed</code> à l’URL.',
```
Frenzie, line 75: @aledeg No weird turn of phrase in the French?

Alkarex:
This is only useful in the case of bad MIME type, such as text/html.
Maybe we should generate an additional message when the MIME type is wrong and include this hint.
Otherwise, maybe we could find a formulation that is not too long, and which explains when this `#force_feed` is useful.

Alkarex:
Let's keep it like that for now. I would like to ship version 1.11 soon.

PR #1882, opened 2018-05-08
Add a docker compose example.

upils:
Using postgresql and with traefik specific labels.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -70,12 +70,28 @@ sudo docker run -d --restart unless-stopped --log-opt max-size=10m \
   --name freshrss freshrss/freshrss
 ```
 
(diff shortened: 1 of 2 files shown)


Alkarex:
Thanks @upils , that looks good.
Could you please consider adding a little sub-section in https://github.com/FreshRSS/FreshRSS/blob/dev/Docker/README.md , under *Deployment in production*, briefly explaining what parts of the  docker-compose file to edit, and how to launch it afterwards?

upils:
Yeah, sure !

Alkarex:
Great @upils 👍 
Could you please move the new section at the end, so that is becomes a sub-section under *Deployment in production* ?

Alkarex:
Merged. Thanks again :-)

PR #1887, opened 2018-05-19
Detection of non-unique GUIDs

Alkarex:
Some feeds are using GUIDs, but fail to make them unique.
Example: https://www.kbh-sprogcenter.dk/en/category/danish-break/feed/
This patch detects non-unique GUIDs, and disable GUIDs in that specific
case.

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -335,6 +335,8 @@ class FreshRSS_Feed extends Minz_Model {
 
 	public function loadEntries($feed) {
 		$entries = array();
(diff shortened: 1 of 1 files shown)


Alkarex:
@Frenzie I have added a log when the state (good GUIDs or not) changes https://github.com/FreshRSS/FreshRSS/pull/1887/commits/d2932d6e50da324d11be6957b0c66db43be38642

Frenzie (approved):
lgtm :-)

Issue #1888, opened 2018-05-20
Postgresql - unix socket not working


bartlomiejcieszkowski:
Hi,
I've passed the unix socket path in installation - in the field host:
/var/run/postgresql
and it worked fine, from the db logs I can see that it used the socket, did whole db creation but afterwards in the logs i get that it can't connect to a database:
[Sun, 20 May 2018 16:17:52 +0200] [error] --- Access to database is denied for `freshrss` (`pgsql:host=;dbname=freshrss`)

it seems that the issue is with parse_url, as it returns empty dbServer['host'] instead of keeping it to aforementioned host path.

I did a local workaround in ModelPdo.php for postgresql that uses the original host value if it begins with / and it works fine, the unix socket is used without a problem

Alkarex:
Addressed in https://github.com/FreshRSS/FreshRSS/issues/1889

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/1364 , https://github.com/FreshRSS/FreshRSS/pull/1366

Issue #1364, opened 2016-11-07
Allow socket as database path


hrehfeld:
My shared host opens postgres connections only as a socket to /home/$USER/tmp, but the field in the installation doesn't allow me to put in a path; it complains that it's not the right format.

(Hoster doc says that all libs linking to the original postgres lib should be able to deal with filepaths/sockets fine)

hrehfeld:
Work-Around: add slash in regexp here: https://github.com/FreshRSS/FreshRSS/blob/master/app/install.php#L637

Alkarex:
Thanks. Fixed in patch https://github.com/FreshRSS/FreshRSS/pull/1366

Alkarex:
Tests welcome. And I will then merge in /dev

hrehfeld:
thanks!

Alkarex:
Merged in /dev

PR #1889, opened 2018-05-20
1888 postgresql unix socket workaround

bartlomiejcieszkowski:
#1888 this enables using of postgresql unix socket outside of installer

### DIFF
--- a/lib/Minz/ModelPdo.php
+++ b/lib/Minz/ModelPdo.php
@@ -58,7 +58,7 @@ class Minz_ModelPdo {
 		try {
 			switch ($db['type']) {
 				case 'mysql':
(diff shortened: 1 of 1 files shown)


Alkarex:
Thanks @bartlomiejcieszkowski 
You have made the change for MySQL. You might want to do do the same for PostgreSQL.
I suggest you introduce a boolean variable `$isSocket` line 57 instead of the inline test.

Alkarex:
@bartlomiejcieszkowski I have changed the code to support PostgreSQL in addition to MyQSL.
Could you please check that it works for you?

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/1364 , https://github.com/FreshRSS/FreshRSS/pull/1366

PR #1891, opened 2018-05-20
Strip credentials from HTTP Referer

Alkarex:
https://github.com/simplepie/simplepie/blob/0edf39018709c0d66cfa6d9b972907bb92461efb/library/SimplePie/File.php#L109
Candidate for pull-request upstream to SimplePie


Alkarex:
Thanks to @oumph from https://linuxfr.org
https://framateam.org/freshrss/pl/q8mqryc4abbaixjxxditzrxgsh

Issue #1892, opened 2018-05-21
xkcd "What If?" feed links not working


Frenzie:
I haven't investigated it, but my guess is some kind of namespace issue. For some reason this feed uses `ns:href="https://what-if.xkcd.com/157/"` instead of just a `href` attribute.

https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fwhat-if.xkcd.com%2Ffeed.atom

```xml
<link ns:href="https://what-if.xkcd.com/157/" ns:rel="alternate" xmlns:ns="http://www.w3.org/2005/Atom" ns:type="text/html"/><published>2018-01-24T00:00:00Z</published></entry><entry><id>https://what-if.xkcd.com/156/</id><title xmlns:ns="http://www.w3.org/2005/Atom" ns:type="text">Electrofishing for Whales</title><updated>2017-03-09T00:00:00Z</updated><author><name>xkcd</name><email>whatif@xkcd.com</email></author><content xmlns:ns="http://www.w3.org/2005/Atom" ns:type="html">&lt;article class=&#34;entry&#34;&gt;
```

Anyway, just dropping it here as a thing that is making my life slightly less nice. I'm not convinced we should work around it on our end. It's not like it works in other feed readers afaik.

Alkarex:
The current SimplePie fails on it too http://simplepie.org/demo/?feed=https%3A%2F%2Fwhat-if.xkcd.com%2Ffeed.atom
But it might be easy to fix. Will look into it.

Alkarex:
Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1893

PR #1893, opened 2018-05-21
SimplePie fix for Atom feeds using namespace for type

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1892
https://what-if.xkcd.com/feed.atom

```xml
<content xmlns:ns="http://www.w3.org/2005/Atom" ns:type="html">&lt;...</content>
```


Alkarex:
@Frenzie

Frenzie (approved):
Looks straightforward. :)

PR #1900, opened 2018-05-28
Fix CLI no_default_feeds

Alkarex:
The option was not working anymore


cli/create-user.php
```diff
@@ -20,7 +20,7 @@
 	empty($options['password']) ? '' : $options['password'],
 	empty($options['api_password']) ? '' : $options['api_password'],
 	$values,
-	!isset($options['no-default-feeds']));
```
Frenzie, line 23: Sneaky typo. :-)

PR #1905, opened 2018-05-29
Feed attributes only for admin

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1838


Alkarex:
cURL feed options can be problematic, due to partial feed caching shared between users.
Enabling only for admin users for now.

Frenzie:
What about the SSL?

Alkarex:
@Frenzie timeout and SSL are both some cURL options. Both options risk to affect other users.

Frenzie:
I see. At a glance only timeout made sense to me as something that could be problematic on the server side.

Alkarex:
Yes, for instance with/without SSL could very well retrieve two different version, or retrieve a malicious content (instead of nothing) if someone disables SSL.

PR #1911, opened 2018-05-31
Use realpath of EXTENSIONS_PATH

gramakri:
This handles the case where the extensions directory might be a symlink.

On the FreshRSS Cloudron app (https://git.cloudron.io/cloudron/freshrss-app/issues/2), the extensions directory is a symlink to a writable location (so that users can install freshrss extensions, rest of the code is read-only).


Alkarex:
Thanks @gramakri :-)

Issue #1870, opened 2018-04-18
Css Selector HTML tag section


ghost:
Hello, 

I am trying to figure out how retrieve the full article by using the CSS selector for this feed
https://anotherwhiskyformisterbukowski.com/feed/

it appears that the content is enclosed there
{section class="single-article__content"}

but I've tried with 
.single-article__content
section.single-article__content

and it does not works. Is there any tools that I can use or cli command to test ?
Thanks.

ghost:
thanks I will try with .single-article, but I don't understand why because the classname is .single-article__content

Frenzie:
The other two I mentioned are not equivalent but the parent. If you want `.single-article__content` you should use it. ;-)

ghost:
It does not work (.single-article__content or single-article), I'm still getting the truncated article.

Alkarex:
Merged in the /dev branch. Tests very much welcome.

PR #1913, opened 2018-06-02
Use cURL for GET full content

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/1870

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -193,7 +193,8 @@ class FreshRSS_Entry extends Minz_Model {
 				try {
 					// l'article n'est pas en BDD, on va le chercher sur le site
 					$this->content = get_content_by_parsing(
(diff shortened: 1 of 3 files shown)


app/Models/Entry.php
```diff
@@ -193,7 +193,8 @@ public function loadCompleteContent($pathEntries) {
 				try {
 					// l'article n'est pas en BDD, on va le chercher sur le site
 					$this->content = get_content_by_parsing(
-						htmlspecialchars_decode($this->link(), ENT_QUOTES), $pathEntries
+						htmlspecialchars_decode($this->link(), ENT_QUOTES), $pathEntries,
+						$this->feed->attributes()
```
Alkarex, line 197: Type error fixed in https://github.com/FreshRSS/FreshRSS/pull/1918

Issue #1904, opened 2018-05-29
Update share to Mastodon


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1521#issuecomment-354178330

Alkarex:
https://framagit.org/framasoft/ttrss/ttrss_mastodon

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/1916

PR #1916, opened 2018-06-04
use GET method for mastodon

hoilc:
Update #1904

According to tootsuite/mastodon#6278 , mastodon has supported sharing with GET method since version 2.2.0.

It seems that the POST method doesn't work properly and most of the top 20 instance are running 2.2.0+ version, therefore I replaced the old configuration.


Alkarex:
Thanks! 👍

Issue #1930, opened 2018-06-10
Fever API - Mark all as read not working


cutzenfriend:
Hi,
I have the newest version 1.11.0 and tried the new Fever API integration with two iOS Apps: Fiery Feeds and Unread. In both Apps i can‘t mark all Feeds as read, they always get resynced as unread.

If I go into a single feed and mark as read it works for this feed.

Alkarex:
Thanks for the report. Could you please tell what works or not?
* Mark an individual article as read
* Mark a feed as read
* Mark a category as read
* Mark all as read

Alkarex:
Thanks. What database are you using? (MySQL, PostgreSQL, SQLite)

Alkarex:
@cutzenfriend Could you please give a try to https://github.com/FreshRSS/FreshRSS/pull/1931 ? If you did not install by git, the easiest is to just replace `./FreshRSS/p/api/fever.php` with the new version
https://github.com/Alkarex/FreshRSS/blob/4327903c9e03ff6bddd0a83da650f866793d5d26/p/api/fever.php

I do not have any iOS device myself, so not tested.

cutzenfriend:
Hi @Alkarex 
I‘m using this docker image: https://hub.docker.com/r/linuxserver/freshrss/

Tried your fix. Works perfectly 😏 👍🏻

Alkarex:
Thanks for the feedback @cutzenfriend 
The fix is available in the /dev branch. Additional testing much welcome.

PR #1931, opened 2018-06-10
Fever fixes

Alkarex:
Was hardcoded for MySQL. Bug in "before" parameter. Bug in mark all as read.
https://github.com/FreshRSS/FreshRSS/issues/1930

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -112,11 +112,6 @@ class FeverAPI
 	private $entryDAO = null;
 	private $feedDAO = null;
 
-	public function __construct() {
-		$this->entryDAO = FreshRSS_Factory::createEntryDao();
-		$this->feedDAO = FreshRSS_Factory::createFeedDao();
-	}
-
 	/**
 	 * Authenticate the user
 	 *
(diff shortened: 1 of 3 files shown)


Alkarex:
The lack of "before" parameter from the clients is quite problematic, as newer articles than the last synchronisation with the client will be marked as read erroneously.

There is also a few TODOs in the code to address

Alkarex:
I hear from a user that Fiery crashed when marking a category as read

p/api/fever.php
```diff
@@ -511,17 +448,17 @@ protected function getItems()
...
-				$feed_ids = explode(",", $_REQUEST["feed_ids"]);
+		if (isset($_REQUEST['feed_ids']) || isset($_REQUEST['group_ids'])) {
+			if (isset($_REQUEST['feed_ids'])) {
+				$feed_ids = explode(',', $_REQUEST['feed_ids']);
 			}
 
-			$dao = $this->getDaoForCategories();
-			if (isset($_REQUEST["group_ids"])) {
-				$group_ids = explode(",", $_REQUEST["group_ids"]);
+			if (isset($_REQUEST['group_ids'])) {
```
Alkarex, line 456: I cannot find any mention of a `group_ids` parameter in the documentation https://feedafever.com/api .
There is however a `group_id` one.

Alkarex:
Merging in /dev to make it easier for testers

Issue #1942, opened 2018-06-17
Some files aren't tracked by git


marienfressinaud:
After a fresh install of FRSS, I discovered some files have been automatically created and aren't tracked by FRSS. It seems fine to me to track them, what do you think? Is there any reason to not track them?

```console
$ git status
Sur la branche dev
Votre branche est à jour avec 'origin/dev'.

Fichiers non suivis:
  (utilisez "git add <fichier>..." pour inclure dans ce qui sera validé)

	data/PubSubHubbub/feeds/index.html
	data/PubSubHubbub/index.html
	data/PubSubHubbub/keys/index.html
	data/users/_/index.html
```

Frenzie:
If they're automatically created doesn't that imply they should be ignored?

marienfressinaud:
Not necessarily, they are required for security reasons so it's better to have them by default. By the way, these files are empty while some others `index.html` redirects to the home page: https://github.com/FreshRSS/FreshRSS/blob/3306a1679c2570c30d4b662c887b4a71ce147398/data/users/index.html#L1-L13

Alkarex:
Yes, I would suggest adding them, but the full version with the redirect.
I think the empty files are auto-created by https://github.com/FreshRSS/FreshRSS/blob/b5300e2ed511426bc72e092b7af4d2951df413b9/cli/prepare.php if they do not exist.

PR #1943, opened 2018-06-17
[#1942] Track missing index files

marienfressinaud:
These files must exist for security reasons: they hide contents of their
directories if the webserver isn't well configured and redirect to the
home page.

They were automatically created by `./cli/prepare.php` script so it was
annoying to have them in the working tree. Also, the files created by
the script were empty.

Closes https://github.com/FreshRSS/FreshRSS/issues/1942


Frenzie (approved):
This is all of them?

marienfressinaud:
I did `git add .` so I guess yes ^^

PR #1950, opened 2018-06-28
before_yesterday

ParoTheParrot:
"älter" better than "vor vorgestern"


kevinpapst:
Hm, in my perception "Ältere Beiträge" (older posts) would fit better, or at least "Ältere" => so starting with an uppercase Umlaut . 
Aside from that, I think changing this would be good, it sounds a bit strange currently.

Alkarex:
@ParoTheParrot Could you please include @kevinpapst 's suggestion? Thanks :-)

PR #1960, opened 2018-07-16
Share with LinkedIn

Alkarex:
https://developer.linkedin.com/docs/share-on-linkedin
https://github.com/FreshRSS/freshrss.org/issues/40


Alkarex:
@paduparchy Please test this patch if you can

paduparchy:
Hello,
Thanks for your fast answer and thanks for opening the case at the right place, i'm not used to with Girhub :(
I will try to test your patch soon.
Regards

Issue #1962, opened 2018-07-20
Entry ID's and 32 bit environment


Vincentv92:
Recently I ran into an issue when I migrated my FreshRSS installation from a 64 bit environment to a 32 bit environment. Specifically the use case when using the Fever API and the Reeder client for macOS and iOS. 

It appears, that some entry id's for the articles have a very high number, as in for example `1531867867902117` as unique identifier.
For storing these entries on a 32 bit environment when using MariaDB or MySQL works perfectly fine since the `id` column is stored as a `bigint`.
But there are issue's within PHP handling these integers if the environment is 32 bit.
I only ran into issues with the Fever API since that's the only component I currently use extensively.

### The issue
After migration my Reeder client's successfully downloaded the articles but failed to fetch new articles through the Fever API. So I opened up my Nginx access logs to see what happend. I saw I had around hundreds of requests per minute which looked something along the lines of this:
```POST /api/fever.php/?api&items&since_id=1532025245261275 HTTP/1.1" 200 72900 "-" "Reeder/3010.29.01 CFNetwork/902.1 Darwin/17.7.0 (x86_64)"
POST /api/fever.php/?api&items&since_id=1532025245261275 HTTP/1.1" 200 72900 "-" "Reeder/3010.29.01 CFNetwork/902.1 Darwin/17.7.0 (x86_64)"
POST /api/fever.php/?api&items&since_id=1532025245261275 HTTP/1.1" 200 72900 "-" "Reeder/3010.29.01 CFNetwork/902.1 Darwin/17.7.0 (x86_64)"
POST /api/fever.php/?api&items&since_id=1532025245261275 HTTP/1.1" 200 72900 "-" "Reeder/3010.29.01 CFNetwork/902.1 Darwin/17.7.0 (x86_64)"
POST /api/fever.php/?api&items&since_id=1532025245261275 HTTP/1.1" 200 72900 "-" "Reeder/3010.29.01 CFNetwork/902.1 Darwin/17.7.0 (x86_64)"
POST /api/fever.php/?api&items&since_id=1532025245261275 HTTP/1.1" 200 72900 "-" "Reeder/3010.29.01 CFNetwork/902.1 Darwin/17.7.0 (x86_64)"
```

So I went ahead and manually verified the returned data from the Fever API and noticed the `id` `1532025245261275` was the last element from the the Fever API response. At this point I wasn't sure what was happening, I compared this to my previous installation and it worked perfectly fine there. 
[...]

PR #1964, opened 2018-07-21
FeverAPI 32-bit fixes

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1962

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -69,14 +69,16 @@ class FeverDAO extends Minz_ModelPdo
 		if (!empty($entry_ids)) {
 			$bindEntryIds = $this->bindParamArray('id', $entry_ids, $values);
 			$sql .= " id IN($bindEntryIds)";
-		} elseif ($max_id !== null) {
+		} elseif ($max_id != null) {
 			$sql .= ' id < :id';
 			$values[':id'] = $max_id;
 			$order = ' ORDER BY id DESC';
-		} elseif ($since_id !== null) {
+		} elseif ($since_id != null) {
 			$sql .= ' id > :id';
 			$values[':id'] = $since_id;
 			$order = ' ORDER BY id ASC';
+		} else {
+			$sql .= ' 1=1';
 		}
 
 		if (!empty($feed_ids)) {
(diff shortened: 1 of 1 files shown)


Alkarex:
@kevinpapst

Frenzie (approved):
Looks straightforward

p/api/fever.php
```diff
@@ -69,11 +69,11 @@ public function findEntries(array $feed_ids, array $entry_ids, $max_id, $since_i
 		if (!empty($entry_ids)) {
 			$bindEntryIds = $this->bindParamArray('id', $entry_ids, $values);
 			$sql .= " id IN($bindEntryIds)";
-		} else if (!empty($max_id)) {
+		} elseif ($max_id !== null) {
 			$sql .= ' id < :id';
 			$values[':id'] = $max_id;
 			$order = ' ORDER BY id DESC';
-		} else {
+		} elseif ($since_id !== null) {
```
kevinpapst: This was wrong before, so just the question if we want to make it fail safe:
Looks like the calling part makes sure that either since_id or max_id is not null. But if we consider that someone is trying to use the API programmatically, then we should move both the WHERE and the ORDER (or at least the WHERE) clause into the if-else. Otherwise the SQL would be broken if someone calls this function with max_id=null and since_id=null. 
Or probably use an array to collect all WHERE parts and append it with `implode(' AND ', $where)` before $order is added to $sql.

Alkarex: Good point. Here is a suggestion https://github.com/FreshRSS/FreshRSS/pull/1964/commits/44d3e5172655d2d9be4ae5ac383371b251eeae62

p/api/fever.php
```diff
@@ -471,17 +471,18 @@ protected function getItems()
...
+			if (!ctype_digit($max_id) || $max_id < 0) {
+				$max_id = null;
 			}
 		} else if (isset($_REQUEST['with_ids'])) {
 			$entry_ids = explode(',', $_REQUEST['with_ids']);
 		} else {
 			// use the since_id argument to request the next $item_limit items
-			$since_id = isset($_REQUEST['since_id']) && is_numeric($_REQUEST['since_id']) ? intval($_REQUEST['since_id']) : 0;
+			$since_id = '' . $_REQUEST['since_id'];
+			if (!ctype_digit($since_id) || $since_id < 0) {
```
kevinpapst: If I read the docu correct, a negative sign would already not be accepted by `ctype_digit` as it is a non numeric value. So I guess the <0 check is not required.
Also from the documentation: the string 000000001 would be accepted, but I don't know what the database makes out of it in that case `id > 000000001`. Thats why I originally used intval().
But to be honest, I didn't test these edge cases.

kevinpapst:
So you pushed while I was still reviewing and typing. LOL, seems you catched my remarks before I even submitted them ;-) 
Looks good to me.

p/api/fever.php
kevinpapst: The old 1=1 trick is also fine :-D

PR #1975, opened 2018-08-09
Boolean return values

ColonelMoutarde:
since php7 you can typeHinted return value:
http://php.net/manual/en/functions.returning-values.php

### DIFF
--- a/app/Models/Context.php
+++ b/app/Models/Context.php
@@ -288,7 +288,7 @@ class FreshRSS_Context {
 	 *
 	 * @return boolean
 	 */
-	public static function isAutoRemoveAvailable() {
+	public static function isAutoRemoveAvailable(): bool {
 		if (!self::$user_conf->auto_remove_article) {
 			return false;
 		}
@@ -309,7 +309,7 @@ class FreshRSS_Context {
 	 *
 	 * @return boolean
 	 */
-	public static function isStickyPostEnabled() {
+	public static function isStickyPostEnabled(): bool {
 		if (self::$user_conf->sticky_post) {
 			return true;
 		}


Frenzie:
But is the syntax backwards compatible?

ColonelMoutarde:
only since php7

on your website:

Software | Recommended | Works also with
-- | -- | --
Web server | Apache2 | Nginx, Lighttpd
PHP | PHP 7 | older PHP versions*: 5.6+, 5.5+, 5.4+ or 5.3.3+
PHP modules | Required: libxml, cURL, DOM PDO_MySQL 					Recommended: JSON, GMP, Zlib, IDN, mbstring, iconv, Zip | Database: PDO_SQLite or PDO_PGSQL
Database | MySQL 5.7+ | older MySQL versions*: 5.6+, 5.5.3+other engines: SQLite 3.7.4+, PostgreSQL
Web browser | Firefox | Chrome, Opera, Safari, Edge or IE 11+

dont use old version of php... php5.3, php5.4 realy ?

Alkarex:
@ColonelMoutarde Thanks, but I believe we should continue supporting PHP5 a bit longer.
PHP 5.6 is still supported at PHP level http://php.net/supported-versions.php

Yes, I have myself been using PHP7+ for more than 2 years, but the reality is that a majority (~80% according to w3techs) of users, and some big distributions are still on PHP5: https://w3techs.com/technologies/details/pl-php/all/all

Example with RedHat or CentOS, a family of distributions popular for servers (which I do not use myself) https://distrowatch.com/table.php?distribution=centos , until 2024 (we might not wait that long for FreshRSS though...). Even Debian and Ubuntu still have some supported versions that include PHP5 (with security updates) until 2019-2020.

We do regularly increase the minimal version of PHP when there are some good reasons such as important bug fixes or needed features (and when this happens, we have users noticing / complaining). Example https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md#2017-10-01-freshrss-180

So patience, such patches will one day be very welcome, but not quite yet :-)

For reference : https://github.com/FreshRSS/FreshRSS/issues/789#issuecomment-73878076

PR #1980, opened 2018-08-12
 bottom padding to enable scroll to mark all as read 

primaeval:
100vh bottom padding to enable scrolling to mark as read and a 100% fallback for ancient browsers. 

It works on the desktop (Windows Firefox and Chrome) and mobile (Android Firefox and Chrome). Mobile needed the 100vh version, desktop was ok with just the 100% version. 

You can stop scrolling when you see the button if you want the original behaviour.

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -630,7 +630,7 @@ br + br + br {
 #bigMarkAsRead {
 	display: block;
 	width: 100%;
-	padding: 1em 0;
+	padding: 1em 0 100% 0;
 	text-align: center;
 	font-size: 1.4em;
 }
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -631,6 +631,7 @@ br + br + br {
 	display: block;
 	width: 100%;
 	padding: 1em 0 100% 0;
+	padding: 1em 0 100vh 0;
 	text-align: center;
 	font-size: 1.4em;
 }


Frenzie:
You should merge into dev, not master. But what does scroll to mark as read mean? It sounds undesirable to me.

primaeval:
You're the one who suggested this fix. ;) https://github.com/FreshRSS/FreshRSS/pull/1974 I've just rearranged my repo to get rid of the other ideas for a fix.

It means if you turn on "Settings \ Reading \ Mark article as read… \ while scrolling" you can scroll past the last article to mark All articles read. Without the fix the last few articles would not get marked and you have to manually press the "Mark All As Read" button. It is how Feedly works and how most people expect apps to work these days in my experience. Once you scroll past something you want to mark that you have seen it so you don't waste your time reading it again.

Frenzie:
I only commented on the odd complexities involved in adding some padding.

So your title is deceptive and it's just a bug fix for a feature I don't like. ;-)

primaeval:
The important word in the title is "all". If FreshRSS has a feature to "mark article as read" "while scrolling" it should really finish the job and mark "all" the articles read, in my opinion.

I can't see what is wrong with the travis build. Is it something I've done or not done?

Alkarex:
Thanks @primaeval 👍 
Could you please add you to https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md ?

Issue #1966, opened 2018-07-31
[Feature Request] Mark all items (younger than...) as unread


hrehfeld:
I would like to be able to _mark everything younger than a certain time as unread_. Basically, the same as with "mark read", just inverse. 

I accidently marked 50k posts as read, because even on mobile it's just two misclicks away and notifications get into the way :cry: This should be easy to implement and help slightly.
Thanks a bunch in advance! ;-)

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/1995

Alkarex:
@hrehfeld Could you please try the following patch? https://github.com/FreshRSS/FreshRSS/pull/1995
You can select recent articles with a search like `date:P30d` and then mark them as unread from the drop-down menu.

Alkarex:
@hrehfeld Example to mark as unread articles younger than e.g. 7 days:

![image](https://user-images.githubusercontent.com/1008324/44998991-53d44c00-afba-11e8-862c-fbbe4dced152.png)

PR #1995, opened 2018-09-01
Add mark-as-unread

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1966


app/i18n/en/feedback.php
```diff
@@ -52,6 +52,10 @@
 	),
 	'sub' => array(
 		'actualize' => 'Updating',
+		'articles' => array(
+			'marked_read' => 'The articles have been marked as read.',
```
Frenzie: The selected articles have been marked as read?

app/i18n/nl/index.php
```diff
@@ -40,6 +40,7 @@
 		'mark_all_read' => 'Markeer alles als gelezen',
 		'mark_cat_read' => 'Markeer categorie als gelezen',
 		'mark_feed_read' => 'Markeer feed als gelezen',
+		'mark_selection_unread' => 'Mark selection as unread',	//TODO
```
Frenzie, line 43: Markeer selectie als ongelezen

Frenzie (approved):
I'll toss in a follow-up PR for the Dutch translation if you prefer. :-)

app/layout/nav_menu.phtml
```diff
@@ -77,8 +78,12 @@
 				'idMax' => FreshRSS_Context::$id_max,
 				'search' => htmlspecialchars_decode(FreshRSS_Context::$search, ENT_QUOTES),
 				'state' => FreshRSS_Context::$state,
-			)
+			),
```
Frenzie, line 81: I always get annoyed by languages that don't allow this. ;-)

app/i18n/nl/index.php
Frenzie: Alright, I'll wait until several of these are merged (or at the very least until tomorrow) before I do that. :-)

app/layout/nav_menu.phtml
Alkarex: Like JSON!

Alkarex:
I have disabled *mark-as-unread* when read articles are not shown.
This is to prevent users from marking-as-unread the whole reader/category/feed by mistake. It might be possible to find a better logic (the ones I have in mind has the drawback of always requiring an additional expensive SQL request at each view to find the boundaries, or require some not-so-nice JavaScript).

Alkarex:
image

Alkarex:
@Frenzie Ready for translation PR ;-)

app/Controllers/entryController.php
```diff
@@ -63,39 +64,38 @@ public function readAction() {
 
 			if (!$get) {
 				// No get? Mark all entries as read (from $id_max)
-				$entryDAO->markReadEntries($id_max);
+				$entryDAO->markReadEntries($id_max, $is_read);
```
Alkarex, line 67: Fixed in https://github.com/FreshRSS/FreshRSS/pull/4456

Issue #1968, opened 2018-08-03
Only one dc:creator is shown as author


hrehfeld:
Example: https://haukerehfeld.de/feed/index.xml 

(feed validates: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fhaukerehfeld.de%2Ffeed%2Findex.xml, albeit with warnings about other stuff.)

Only the first dc:creator is shown in the feed view.

    <?xml version="1.0" encoding="utf-8"?>
    <rss version="2.0" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:atom="http://www.w3.org/2005/Atom">
      <channel>
        <title>Hauke Rehfeld</title>
        <link>https://haukerehfeld.de/</link>
        <description>Latest posts on Hauke Rehfeld</description>
        <pubDate>Mon, 30 Jul 2018 14:19:30 +0200</pubDate>
        <atom:link href="https://haukerehfeld.de/feed/index.xml" rel="self" type="application/rss+xml"/>
        <item>
          <title>Eindruck einer friedlichen und geruhsamen Ansiedlung</title>
          <link>https://haukerehfeld.de/art/2018-houses</link>
          <guid isPermaLink="true">https://haukerehfeld.de/art/2018-houses</guid>
          <dc:creator>First Author</dc:creator>
          <dc:creator>Second Author</dc:creator>
          <category>art</category>
          <pubDate>Mon, 16 Jul 2018 22:00:00 +0000</pubDate>
          <description><![CDATA[...]]></description>
        </item>
    ...

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/1997 . Tests welcome

Alkarex:
@hrehfeld Only the new ones, or by removing/re-adding a feed

PR #1997, opened 2018-09-01
Allow multiple authors

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/1968

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -341,7 +341,7 @@ class FreshRSS_Feed extends Minz_Model {
 
 		foreach ($feed->get_items() as $item) {
 			$title = html_only_entity_decode(strip_tags($item->get_title()));
-			$author = $item->get_author();
+			$authors = $item->get_authors();
 			$link = $item->get_permalink();
 			$date = @strtotime($item->get_date());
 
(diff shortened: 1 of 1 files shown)


app/Models/Feed.php
```diff
@@ -409,12 +409,19 @@ public function loadEntries($feed) {
 			$guid = $item->get_id(false, false);
 			$hasUniqueGuids &= empty($guids['_' . $guid]);
 			$guids['_' . $guid] = true;
+			$author_names = '';
+			if (is_array($authors)) {
+				foreach ($authors as $author) {
+					$author_names .= html_only_entity_decode(strip_tags($author->name == '' ? $author->email : $author->name)) . ', ';
```
Frenzie, line 415: Will that work right in RTL scenarios?

Alkarex: I have to admit that I have near-zero experience with RTL.
The current Hebrew translation seems to do something similar:

https://github.com/FreshRSS/FreshRSS/blob/32d9c3b7905f4e43ffdf4bf2bf37723cfd18390c/app/i18n/he/gen.php#L184

I guess it would also depend on the alphabet of authors' names, so I do not know how to make it better.

Frenzie: Let's not worry about it prematurely then. :-)

Frenzie (approved):

PR #2010, opened 2018-09-08
Better test if server has public address

Alkarex:


lib/lib_rss.php
```diff
@@ -109,24 +109,29 @@ function safe_ascii($text) {
...
 		'[::1]',
+		'ip6-localhost',
 		'localhost6',
 		'localhost6.localdomain6',
 	));
 
-	return $is_public;
-}
+	if ($is_public) {
+		$ip = gethostbyname($host);
```
Alkarex, line 128: This does not work as expected when FreshRSS is in a container (e.g. Træfik + Docker). Will fix.

Alkarex: Fixed with https://github.com/FreshRSS/FreshRSS/pull/2084

PR #2020, opened 2018-09-12
Fix Mark as read button spacing in Chrome

sirideain:
The button was shifted up in Chrome, but not Firefox. This fixes that and doesn't have an impact on Firefox.

image


Alkarex:
Thanks @sirideain !
Please add yourself in https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md as well :-)

Alkarex:
If you can, could you please have a look at the other themes? Some of the others might have the same problem.

Issue #1999, opened 2018-09-03
Reeder 3/iOS does not work with the fever API implementation


chemical1979:
I just installed FreshRSS master on my system to switch away from Feedly Cloud Solution to a self-hosted one. I'm pleasantly surprised how well the setup is described. Everything works out well.

I started with Unread on my iPad as a mobile reader which worked great, but isn't easy for me to adopt. So I started using [Reeder 3 (iOS)](https://itunes.apple.com/us/app/reeder-3/id697846300?mt=8) which is a better fit for me, but somehow does not work (Login failed). 

As Unread worked, I'm pretty sure my setup is correct. I also followed [this page](https://freshrss.github.io/FreshRSS/en/users/06_Fever_API.html) to check my setup. At the last part it hints about a change in fever.api (I guess it should read "/p/api/fever.php" instead) to log the requests that go in via API.

The only things that go into the log if I try to login with Reeder is this:

fever.log
`Reeder/3.1.21 CFNetwork/902.2 Darwin/17.7.0: {"api":""}`

There is not even content in the JSON requests, so I'm not sure whether the problem is within Reeder or within FreshRSS Fever API implementation. But as Reeder is free, maybe someone else more capable is able to help me here.

If I remove my login from Unread and re-add the source, this gets dumped into the log (comparision):

`Unread/109055 CFNetwork/902.2 Darwin/17.7.0: {"api":""}
Unread/109055 CFNetwork/902.2 Darwin/17.7.0: {"api":"","api_key":"409283a4322b2a61xxxxxxxxxxxxxxxx"}
Unread/109055 CFNetwork/902.2 Darwin/17.7.0: {"api":"","feeds":"","api_key":"409283a4322b2a61xxxxxxxxxxxxxxxx"}
Unread/109055 CFNetwork/902.2 Darwin/17.7.0: {"api":"","groups":"","api_key":"409283a4322b2a61xxxxxxxxxxxxxxxx"}
Unread/109055 CFNetwork/902.2 Darwin/17.7.0: {"api":"","unread_item_ids":"","api_key":"409283a4322b2a61xxxxxxxxxxxxxxxx"}
Unread/109055 CFNetwork/902.2 Darwin/17.7.0: {"api":"","saved_item_ids":"","api_key":"409283a4322b2a61xxxxxxxxxxxxxxxx"}`

Did someone positive-test Reeder 3 with FreshRSS? Maybe the problem is on my side.

Alkarex:
[...]

PR #2026, opened 2018-09-16
Small docs update (see #1999) and i18n fix (see #1309)

chemical1979:
Adding Reeder-3/iOS as an supported client via Fever API to the docs and a small translation fix for an option in the config (german)


Alkarex:
Thanks @chemical1979 👍 
Could you please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md ? :-)

Frenzie (approved):

Issue #1367, opened 2016-11-09
[FR] Tag category


anayrat:
Hi,

It will be usefull if we can use tag as category with advanced filter : 

- Main stream
  - Tags
    - Tag 1
    - Tag 2
  - My category 1
    - Feed 1
  - My category 2
    - Feed 2

anayrat:
For information this functionnality is in tinyrss. It is missing for my usage :(

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/928

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/2027
Tests welcome

Issue #928, opened 2015-07-28
Add custom tags to articles, and show custom tags as virtual categories


Alkarex:
See https://github.com/FreshRSS/FreshRSS/issues/922#issuecomment-124880593

Alkarex:
I believe the main needs will be:
1. add new tag to existing article (entry);
2. retrieve the list of articles matching a tag (virtual category);
   - Requires an index on the tag
3. show all tags for an article;
   - Requires an index on the article or article reference

There are several approaches for the SQL model, with different advantages/drawbacks.

Here is the variant 1, in which tags are referenced by an integer. It offers the easy possibility of changing the name of the tag, and using a long tag name should not affect performances. Articles (entries) can have several tags.

``` SQL
CREATE TABLE IF NOT EXISTS `tag` (
    `id` SMALLINT NOT NULL AUTO_INCREMENT,
    `name` VARCHAR(255) NOT NULL,
    PRIMARY KEY (`id`),
    UNIQUE KEY (`name`)
) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci
ENGINE = INNODB;

CREATE TABLE IF NOT EXISTS `tags` (
    `id_tag` SMALLINT NOT NULL,
    `id_entry` BIGINT NOT NULL,
    FOREIGN KEY (`id_tag`) REFERENCES `tag`(`id`) ON DELETE CASCADE ON UPDATE CASCADE,
    FOREIGN KEY (`id_entry`) REFERENCES `entry`(`id`) ON DELETE CASCADE ON UPDATE CASCADE,
    INDEX (`id_entry`),
    UNIQUE KEY (`id_tag`,`id_entry`)
) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci
ENGINE = INNODB;
```

Alkarex:
We also need to decide what to do with the tags provided by the RSS feed:
1. Keep them in the `entry` table as they are now;
2. Merge them with the new tagging system, with a flag to distinguish custom tags;

rolero:
[...]

PR #2027, opened 2018-09-16
Custom labels

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/928
https://github.com/FreshRSS/FreshRSS/issues/1367


Alkarex:
First draft

Alkarex:
image

Alkarex:
image

app/Models/EntryDAO.php
```diff
@@ -129,13 +129,18 @@ protected function autoUpdateDb($errorInfo) {
 						return $this->addColumn($column);
 					}
 				}
-			} elseif ($errorInfo[0] === '42S02' && stripos($errorInfo[2], 'entrytmp') !== false) {	//ER_BAD_TABLE_ERROR
-				return $this->createEntryTempTable();	//v1.7
+			} elseif ($errorInfo[0] === '42S02') {	//ER_BAD_TABLE_ERROR
```
aledeg: I think that the magic value 42S02 should be in class constant.

app/Models/EntryDAOPGSQL.php
```diff
@@ -12,8 +12,13 @@ public function sqlHexEncode($x) {
 
 	protected function autoUpdateDb($errorInfo) {
 		if (isset($errorInfo[0])) {
-			if ($errorInfo[0] === '42P01' && stripos($errorInfo[2], 'entrytmp') !== false) {	//undefined_table
-				return $this->createEntryTempTable();
+			if ($errorInfo[0] === '42P01') {	//undefined_table
```
aledeg: Same here

app/Models/EntryDAO.php
Alkarex: Done in https://github.com/FreshRSS/FreshRSS/pull/2027/commits/036be54ac70b707c1113788a6f9e4146f9981592

app/Models/EntryDAOPGSQL.php
Alkarex: Done in https://github.com/FreshRSS/FreshRSS/pull/2027/commits/036be54ac70b707c1113788a6f9e4146f9981592

Alkarex:
Now with the possibility to add new tags:

image

Alkarex:
If someone tried this branch, the last commit introduces a new field. Either drop the tag tables (which will be re-created automatically), or add the field with:

```sql
ALTER TABLE freshrss_alice_tag ADD COLUMN attributes TEXT;
```

app/Models/CategoryDAO.php
```diff
@@ -9,7 +9,7 @@ public function addCategory($valuesTmp) {
 		$stm = $this->bd->prepare($sql);
 
 		$values = array(
-			mb_strcut($valuesTmp['name'], 0, 255, 'UTF-8'),
+			mb_strcut($valuesTmp['name'], 0, 191, 'UTF-8'),
```
aledeg: Not a big fan of magic values. It would be better to have constants. Same comment goes for all magic values.

aledeg (changes requested):

app/Models/CategoryDAO.php
Alkarex: Done in https://github.com/FreshRSS/FreshRSS/pull/2027/commits/d8a3c8cfb90c97b72f549c7ba13870ebf03249b5 (but not a big fan of the additional complexity of a constant in this case :-) )

Alkarex:
TODO:
* [x] Drop-down menu to e.g. remove tags
* [x] Mark as read
* [x] Read/unread views
* [x] Live-change unread tags when marking as read/unread
* [x] Keep tagged articles during purge?
* [x] Investigate tag options in API / mobile clients

aledeg (approved):
:+1: :tada:

aledeg:
Nice work @Alkarex

Alkarex:
Thanks @aledeg , I hope you will like it :-)

Alkarex:
Google Reader API implemented. Fully functional with News+:

screenshot_20180924-232132_news

screenshot_20180924-232212_news

PR #2032, opened 2018-09-26
Escape cron step values with backslash

jaltek:
As I had this error myself I propose the following change:
Escape cron step values with backslash in $CRON_MIN.

Otherwise you will get the error ```sed: bad option in substitution expression```.


Alkarex:
Thanks for the bug report 👍

Alkarex:
@jaltek I have changed the `sed` delimiter to avoid having to escape slashes.
Could you please double-check that it works fine for you? https://github.com/FreshRSS/FreshRSS/pull/2032/commits/7546ec43adfc5ea95e0707522fa8e385dfe0775f

jaltek:
@Alkarex Yes, works fine for me :+1:

Alkarex:
Thanks for the feedback :-)

PR #2035, opened 2018-09-29
Message when PHP is not installed

Alkarex:
image


Frenzie (approved):
Maybe something like "not installed or configured correctly"?

PR #2040, opened 2018-10-07
Remove some preload / prefetch

Alkarex:
This approach was only efficient in the specific case when no change was made while reading the first page (no mark-as-read, favourites, tags), and useless in the other situations.
Removed to reduce server load.

### DIFF
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -90,7 +90,7 @@ class FreshRSS extends Minz_FrontController {
 				}
 				$filetime = @filemtime(PUBLIC_PATH . '/themes/' . $theme_id . '/' . $filename);
 				$url = '/themes/' . $theme_id . '/' . $filename . '?' . $filetime;
-				header('Link: <' . Minz_Url::display($url, '', 'root') . '>;rel=preload', false);	//HTTP2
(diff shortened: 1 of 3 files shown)


Frenzie (approved):

app/FreshRSS.php
```diff
@@ -90,7 +90,7 @@ public static function loadStylesAndScripts() {
 				}
 				$filetime = @filemtime(PUBLIC_PATH . '/themes/' . $theme_id . '/' . $filename);
 				$url = '/themes/' . $theme_id . '/' . $filename . '?' . $filetime;
-				header('Link: <' . Minz_Url::display($url, '', 'root') . '>;rel=preload', false);	//HTTP2
+				//header('Link: <' . Minz_Url::display($url, '', 'root') . '>;rel=preload', false);	//HTTP2
```
aledeg: If it's useless, it's better to remove it completely.

Alkarex: Removed in https://github.com/FreshRSS/FreshRSS/pull/2040/commits/a666cd8307290995711c2f597d3e7354ac30e022
It might be useful associated with a smarter way of guessing whether the client needs it or not (e.g. first request)

PR #2045, opened 2018-10-13
Ensure fast flush of HTTP headers and HTML head

Alkarex:
The fast flush optimisation done in https://github.com/FreshRSS/FreshRSS/pull/1133 does not seem to work anymore (need to check if it is related to a PHP version).
Work-around when PHP flush() is not working.
Can be tested by adding a `sleep(5);` after:

https://github.com/FreshRSS/FreshRSS/blob/ee902ee7c4370421802768c3105ba269a4f97b16/app/layout/layout.phtml#L27
Follow-up of the performance checks of https://github.com/FreshRSS/FreshRSS/pull/2040

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -14,7 +14,6 @@ WORKDIR ${FRESHRSS_ROOT}
 
 COPY . ${FRESHRSS_ROOT}
 COPY ./Docker/*.Apache.conf /etc/apache2/conf.d/
-COPY ./p/.user.ini /etc/php7/conf.d/FreshRSS.php.ini
 
 RUN sed -r -i "/^[ ]*(CustomLog|ErrorLog|Listen) /s/^/#/" /etc/apache2/httpd.conf && \
 	echo "17,37 * * * * php ${FRESHRSS_ROOT}/app/actualize_script.php 2>&1 | tee /tmp/FreshRSS.log" >> \
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -14,6 +14,7 @@ WORKDIR ${FRESHRSS_ROOT}
 
 COPY . ${FRESHRSS_ROOT}
 COPY ./Docker/*.Apache.conf /etc/apache2/conf.d/
+COPY ./Docker/*.php.ini /etc/php7/conf.d/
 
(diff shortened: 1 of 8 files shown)


Alkarex:
This is quite strange. `flush()` used to work fine - as far as I can recall - but I cannot make it to work now. Tested on:
* PHP 7.2.10 on Ubuntu 18.04 with Apache 2.4.29
* PHP 7.0.32 on Ubuntu 16.04 with Apache 2.4.18
* PHP 7.2.8 on Alpine 3.8 with Apache 2.4.34
* PHP 5.6.36 on Alpine 3.4 with Apache 2.4.33

Turning on/off Apache DEFLATE makes a difference when PHP runs as CGI, not when running as module.

Documentation:
* http://php.net/manual/function.flush.php
* http://php.net/manual/outcontrol.configuration.php#ini.output-buffering
* https://httpd.apache.org/docs/current/mod/mod_deflate.html

In my tests, it looks like `flush()` only works when the value of `output_buffering` is smaller than what `flush()` releases.

To test in FreshRSS, add a delay after `flush()`, e.g. add `sleep(10);` after the line
https://github.com/FreshRSS/FreshRSS/blob/ee902ee7c4370421802768c3105ba269a4f97b16/app/layout/layout.phtml#L27

To test outside FreshRSS:

```php
<html>
<body>
<pre>
Before...
<?php flush(); sleep(5); ?>
Between...
<?php flush(); sleep(5); ?>
End.
</pre>
</body>
</html>
```

```sh
curl -v -N 'http://example.net/flush.php'
# And
curl -v -N --compressed 'http://example.net/flush.php'
```

You are supposed to see an output with HTTP headers and the first part of the HTML, then a pause, then some more. If you only see one long pause and get all the output at once, it is wrong.

Additional help / testing welcome.

Alkarex:
https://www.jeffgeerling.com/blog/2016/streaming-php-disabling-output-buffering-php-apache-nginx-and-varnish

Alkarex:
The fast flush speed optimisation works fine together with compression when PHP runs as Apache module (like we do in the Docker image).
However, I have not found a way to have at the same time compression and fast flush when PHP runs as CGI / FastCGI / FPM. Tips welcome.

PR #2048, opened 2018-10-14
init confirm also for configuration pages

Alkarex:
There are some actions that require confirmation also in the
configuration pages, such as delete user.


Alkarex:
Bug introduced in https://github.com/FreshRSS/FreshRSS/pull/1374

Issue #2056, opened 2018-10-21
403 Error when trying to update passwords


Twilek-de:
Hi all,

my FreshRSS installation seems to work but when I try to update a password I get a 403 You don´t have permission to view this page error (I am logged in as an admin at the time). I have set my environment to development and the log reads "gen.share.Known is not a valid key". I have no idea wether that error is connected to the permission problem or how to fix it.

Regards
Henning

Twilek-de:
Hmm... I have looked at this further. The users config file gets rewritten with new password hashes and the new passwords work. At the same time I get logged out. Is it supposed to work that way? If so, that would be really inelegant.

Alkarex:
Hello,
What browser are you using?
Could you please try again with another browser without password manager / form autocomplete?

Twilek-de:
I am using the latest Firefox. I have just tried Edge and Chrome (both latest versions) with the same result. Always 403 when changing the password. This happens in the "Users" subection as well as in the "Profile" subsection of the settings.

Twilek-de:
The "gen.share.Known is not a valid key" seems to pop up whenever FreshRSS is called and not be linked to the 403 problem.

Twilek-de:
I think I solved the problem with the gen.share.Known" thing. I copied the en.php in cli/i18n/ignore to de.php in the same directory (my user has "de" as a locale) and turned of developement. Now it doesn´t show up any more. This of course does not change the 403 page.

Alkarex:
Sorry, I thought it was when you were changing another user's password from the admin account.
Indeed, when you change your own password, the current behaviour is that you are logged out and need to log in again. The 403 error page is indeed unfortunate, and I will try to improve that.

Twilek-de:
Ah ok, then it is not that much of a deal, it just feels like the wrong behaviour so you start to search for the error. Couldn´t you simple force a logout and return the user to the starting page (whatever a guest user sees)?!

Alkarex:
@Twilek-de Could you please try this patch? https://github.com/FreshRSS/FreshRSS/pull/2057

Alkarex:
@Twilek-de Another patch for the *Known* bug in German: https://github.com/FreshRSS/FreshRSS/pull/2058

Alkarex:
P.S. Could you please try to edit the following line for a better German translation?

[...]

PR #2058, opened 2018-10-21
i18n de: Known

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2056


Alkarex:
Proper translation in https://github.com/FreshRSS/FreshRSS/pull/2059

PR #2062, opened 2018-10-21
Allow dot in username

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2061


app/Controllers/userController.php
```diff
@@ -38,7 +38,7 @@ public static function hashPassword($passwordPlain) {
 	 * The username is also used as folder name, file name, and part of SQL table name.
 	 * '_' is a reserved internal username.
 	 */
-	const USERNAME_PATTERN = '[0-9a-zA-Z_]{2,38}|[0-9a-zA-Z]';
+	const USERNAME_PATTERN = '[0-9a-zA-Z_][0-9a-zA-Z_.]{1,38}|[0-9a-zA-Z]';
```
Frenzie, line 41: In MySQL I note that this means all identifiers should be quoted since only these are permitted in unquoted:

https://dev.mysql.com/doc/refman/8.0/en/identifiers.html

I'm not sure if hat could cause any issues at this point in time?
> ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)

Alkarex: We do quote identifiers for MySQL / SQLite / PostgreSQL so that should be fine (require quick testing)

PR #2052, opened 2018-10-16
Option to force clear cache

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1020#issuecomment-428515868
image

PR #2068, opened 2018-10-22
German translation according to #2052

spackmat:


aledeg (approved):

Alkarex:
GitHub is still recovering from a technical incident https://status.github.com/messages so let's wait a bit before merging, etc.

PR #2070, opened 2018-10-23
mark_selection_unread

Twilek-de:
Translated that

### DIFF
--- a/app/i18n/de/index.php
+++ b/app/i18n/de/index.php
@@ -40,7 +40,7 @@ return array(
 		'mark_all_read' => 'Alle als gelesen markieren',
 		'mark_cat_read' => 'Kategorie als gelesen markieren',
 		'mark_feed_read' => 'Feed als gelesen markieren',
-		'mark_selection_unread' => 'Auswahl als gelesen markieren',	
+		'mark_selection_unread' => 'Auswahl als gelesen markieren',
 		'newer_first' => 'Neuere zuerst',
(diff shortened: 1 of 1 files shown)


Alkarex:
Thanks for all the translations 👍 
Another time, please make the edits in the /dev branch instead of /master (I was able to retarget the correct branch without problem, but it does not always work in case of version conflicts)

app/i18n/de/index.php
```diff
@@ -40,7 +40,7 @@
 		'mark_all_read' => 'Alle als gelesen markieren',
 		'mark_cat_read' => 'Kategorie als gelesen markieren',
 		'mark_feed_read' => 'Feed als gelesen markieren',
-		'mark_selection_unread' => 'Mark selection as unread',	//TODO
+		'mark_selection_unread' => 'Auswahl als gelesen markieren',
```
Frenzie: I think you missed an `un(gelesen)`. ;-)

Twilek-de: Ah chute I missed that
```suggestion
		'mark_selection_unread' => 'Auswahl als ungelesen markieren',
```

Frenzie: You'll have to commit the change yourself. I don't have write access to your repo. :-)

Twilek-de: A thank you, I am new to the whole committing thing :-) I hope that did it...

Frenzie:
Thanks!

Issue #2054, opened 2018-10-20
Nginx config in documentation not compatible with php7


Maltimore:
Hi,
I know extremely little about php, or sysadmin in general, and only randomly made this work, so I don't know exactly what the problem is.
But on my debian 9 system, I had to modify the suggested nginx configuration here:

https://freshrss.github.io/FreshRSS/en/admins/02_Installation.html

to set this lock file to /var/run/php/php7.0-fpm.sock instead of /var/run/php5-fpm.sock

Alkarex:
Thanks. Could you please try to edit the following two documents accordingly?

* https://github.com/FreshRSS/FreshRSS/blob/dev/docs/en/admins/02_Installation.md
* https://github.com/FreshRSS/FreshRSS/blob/dev/docs/fr/users/01_Installation.md

Maltimore:
Will do!

PR #2076, opened 2018-10-24
doc: update nginx config to reflect debian php7.0 changes

Maltimore:
Solves #2054


Alkarex:
Thanks @Maltimore 👍

Frenzie:
It's a pity that there doesn't seem to be a more version-agnostic way.

Issue #2077, opened 2018-10-24
<guid> matching of entries is done case-insensitively, should be case-sensitive


Dan-Q:
I have an RSS feed which summarises URLs produced by a URL-shortening service. The <guid> of each entry in the feed is the shortened URL, each of which is unique. However, FreshRSS discards those where the guid matches that of an existing guid, even where they differ in case.

E.g. the following feed:
```
<?xml version="1.0" encoding="UTF-8" ?>
<rss version='2.0' xmlns:atom='http://www.w3.org/2005/Atom'>
  <channel>
    <item>
      <title>An Oral History of ‘Leisure Suit Larry’ – MEL Magazine</title>
      <description></description>
      <link>https:&#x2F;&#x2F;melmagazine.com&#x2F;an-oral-history-of-leisure-suit-larry-ef41bc374802</link>
      <guid>https://url-shortening-service/Mw</guid>
      <pubDate>Sat, 04 Aug 2018 18:52:06 -0000</pubDate>
    </item>
    <item>
      <title>The Web That Never Was - Dylan Beattie - YouTube</title>
      <description></description>
      <link>https:&#x2F;&#x2F;www.youtube.com&#x2F;watch?v=j51Fmn4JVwU</link>
      <guid>https://url-shortening-service/MW</guid>
      <pubDate>Sat, 04 Aug 2018 13:14:21 -0000</pubDate>
    </item>
  </channel>
</trr>
```

When imported into FreshRSS produces only ONE entry in the database: it appears that the guids are being compared in a case-insensitive manner.

This is incorrect behaviour because (with the exception of the protocol and domain name parts) URLs, which are often used as guids, are NOT (necessarily) case-sensitive.

Frenzie:
Does the behavior differ if you add `isPermaLink="true"`?

At a glance I only see a whole bunch of characters being stripped out (cf. #1335) but perhaps it's down to SimplePie.

Alkarex:
@Dan-Q Thanks for the bug report.

In FreshRSS, GUID is currently case-insensitive only when using MySQL (`COLLATE utf8mb4_unicode_ci`), and should already be case-sensitive when using SQLite (`COLLATE BINARY`) or PostgreSQL.
[...]

PR #2078, opened 2018-10-24
MySQL GUID case sensitive

Alkarex:
latin1_bin
https://github.com/FreshRSS/FreshRSS/issues/2077


Alkarex:
~~The current approach is good for cron updates, but not good for Ajax updates (which perform several requests in parallel). I will find another approach.~~
Fixed

Alkarex:
@Dan-Q Merged in /dev . Tests welcome

PR #2088, opened 2018-10-29
Fix Swage theme issues

pattems:
I fixed the issue with the active tab not being highlighted in the config view.

In order to make the configure-feeds buttons be able to adapt to fill the width of the aside, I had to add ID tags to them, as without this change, I couldn't target the buttons individually, and the only solution would be to make each button 50% the width of the aside, a less than ideal solution that would cut text in pretty much any language (except possibly languages which use logograms instead of alphabets).

Ideally, I'd like to add ID's to all of the UI components, which would greatly increase the themeability of FreshRSS without breaking any existing themes, but that's not work I'm going to take on without any sort of discussion first.

### DIFF
--- a/app/layout/aside_feed.phtml
+++ b/app/layout/aside_feed.phtml
@@ -12,8 +12,8 @@
 
 	<?php if (FreshRSS_Auth::hasAccess()) { ?>
 	<div class="stick configure-feeds no-mobile">
-		<a class="btn btn-important" href="<?php echo _url('subscription', 'index'); ?>"><?php echo _t('index.menu.subscription'); ?></a>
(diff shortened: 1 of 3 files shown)


Alkarex:
Thanks @pattems 😃 
A couple of versions ago, there was this style fix https://github.com/FreshRSS/FreshRSS/pull/1980 to allow scrolling enough to the bottom to mark all articles as read. Could you please try to get the same in Swage?

Alkarex:
Yes please, another PR is fine. Too many PRs are not a problem (yet) :-)

pattems:
I'll get it to you a bit later on, then!

PR #2089, opened 2018-10-29
Fix Dropdown being obscured by other elements

pattems:
I noticed that in every theme but Swage (and I can't quite figure out what my own theme's doing differently that makes this a non-issue), the properties tile of the theme select slider floats over the menu on a small screen, like so:

image

massively increasing the dropdown-menu's z-index will ensure that it should stay on top of everything; if this solution's non-ideal for some reason, .nav label's z-index could be reduced from 11 to 9, and properties' z-index could be reduced to 8.

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -256,7 +256,7 @@ a.btn {
 }
 .dropdown-target:target ~ .dropdown-menu {
 	display: block;
-	z-index: 10;
+	z-index: 1000;
 }
 .dropdown-close {
 	display: inline;


Frenzie:
Don't forget to check overflow before going all in on z-index.

(Generic remark; haven't looked yet.)

pattems:
@Frenzie are you talking about checking if the overlapping tile was overflowing, causing it to draw on top of the dropdown? That's not the case, if that's what you're talking about. Otherwise, I don't think I understand what you mean.

Frenzie:
@pattems Apologies, I slightly misinterpreted the issue on my phone.

Looking at it I don't really understand why .properties has a z-index other than the default at all?

Anyway, note that you can also use things like `inherit 1` (1 higher than parent) instead of absolute values in z-index stacks.

pattems:
I don't really understand why .properties has a z-index specified either. Everything in the slider seems to work fine without setting any z-indexes at all, but since I didn't understand why they were there, I opted for not touching them, in case there's some really important reason that's just not immediately evident.

In the template.css file, there's nothing up the chain for .dropdown-menu to inherit a z-index from. the only z-index in the dropdown menu path in template.css is https://github.com/FreshRSS/FreshRSS/blob/e04804d0f67dd43fd3f072b9a127768ee7b7b56c/p/themes/base-theme/template.css#L257-L260

Also, as way of check, I copied the base-theme to a theme so I could check it raw, making sure it wasn't just a persistent issue with all the themes:

image

In any case, my rationale for a z-index of 1000 is to be high enough up the z-index that it should float above anything else in any theme, as one would expect for a pop-up menu; I can't think of a better way of ensuring that it stays on top of everything else.

Issue #2085, opened 2018-10-28
Cannot close an article that is opened and marked as unread


Alwaysin:
Steps to reproduce : 
- open an article,
- mark it as unread
- try to close it by clicking on the title

-> not working

Alwaysin:
Cannot say for sure but I think I would have noticed if it was already there!

pattems:
I'm not really well-versed in javascript, but I'm poking at this a bit to see if I can work something out; the issue is bigger than reported, however. The addition of the .keep_unread class makes a flux element completely non-interactive as long as the keep_unread class exists. (Manually marking an item previously marked as keep_unread only clears not_read, and doesn't clear keep_unread until a page reload; this has to be unrelated since it also worked that way in 11.2, but possibly behaviour that should be changed?)

Alkarex:
Fixed by @pattems in https://github.com/FreshRSS/FreshRSS/pull/2092

PR #2025, opened 2018-09-15
Improve authors

Alkarex:
Multiple authors display with links, search compatible Unicode.
Improve https://github.com/FreshRSS/FreshRSS/pull/1997
Related to https://github.com/FreshRSS/FreshRSS/pull/2023

PR #2092, opened 2018-10-30
Fix for #2085

pattems:
Issue #2085 seems to have been caused by commit #2025, where .keep_unread was added to the onclick that would cause a simple return, intended for clicking on things that shouldn't cause interaction.


Frenzie (approved):

Alkarex:
Thanks! I cannot find why I added `.keep_unread`... It was for the authors / search / tags, but everything seems to work find without it...

Issue #2091, opened 2018-10-29
Vienna Support?


meminens:
I am sorry if Github is not appropiate to ask but I wonder what I can do to make FreshRSS work with Vienna? Choosing Other from the Sync menu and entering username/password did not work.

`http://www.vienna-rss.com/`

Alkarex:
Related: https://github.com/ViennaRSS/vienna-rss/issues/1197

> I can see the Vienna-RSS supports the [Open Reader API](http://rss-sync.github.io/Open-Reader-API/rssconsensus/) . As far as I can see, this API has never been written / completed (or I cannot find it - links welcome), and it is the first client I stumble upon that seems to support it.
I am happy to collaborate from FreshRSS side to make the compatibility happen.

@bisherbas I cannot test myself because I do now own anything Apple, but we could start by making a log of the first calls ViennaRSS is doing:

1. In your FreshRSS, please uncomment the following two lines:

https://github.com/FreshRSS/FreshRSS/blob/ec56aa9e815b58f66e1c4b5bbf05592ee34b8063/p/api/greader.php#L867-L868

2. Erase (if it exists) `./FreshRSS/data/users/_/log_api.txt`

3. Try to connect from Vienna RSS
    * Be sure to set your API password and to use the full URL to the Google Reader API endpoint, which you can find by clicking in the Profile menu of FreshRSS:
![image](https://user-images.githubusercontent.com/1008324/47675926-831ec800-dbbb-11e8-8068-ddb9fc2cd9cf.png)

4. Paste here the content of `./FreshRSS/data/users/_/log_api.txt` (you are welcome to anonymise your server address, username, etc.)

bisherbas:
Hi @Alkarex 

I added the full API url into Vienna and the Vieann app itself crashed on my Mac. So much so that I cannot even reopen the app. Crashes immediately. Had to fully uninstall the Vienna app.

I did uncomment the debug lines but no file is generated after attempting a sync within the Vienna app.

For reference, on my Android device, EasyRSS works perfectly fine using the Google Reader API.

pattems:
Vienna's OpenReader API file, linked directly to a problematic line,at the end of this post.

Part of (possibly the only? I'm in over my head here) the problem is that Vienna seems to expect the API at a specific URL, "https://\<your-server\>/reader/api/0/"

[...]

PR #2093, opened 2018-10-30
Tweaks for Vienna RSS

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2091
https://github.com/ViennaRSS/vienna-rss/issues/1197

### DIFF
--- a/app/Models/BooleanSearch.php
+++ b/app/Models/BooleanSearch.php
@@ -45,6 +45,14 @@ class FreshRSS_BooleanSearch {
 		return $this->searches;
 	}
 
+	public function add($search) {
+		if ($search instanceof FreshRSS_Search) {
+			$this->searches[] = $search;
+			return $search;
+		}
+		return null;
+	}
+
 	public function __toString() {
 		return $this->getRawInput();
 	}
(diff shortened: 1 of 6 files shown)


Frenzie (commented):
These encoding changes wouldn't cause any trouble elsewhere?

Frenzie:
Pardon, I just saw the comment stating that it's technically a bug. GH had cleverly hidden the full commit message from my view.

Alkarex:
@Frenzie Those changes should not break existing clients. But additional tests very much welcome. We need to try:
* with `< & > " '` in all text fields such as category name, label name, feed name, feed URL, entry title, entry URL, entry author
* all actions such as retrieve, change read/favourite status, label / unlabel, add, rename, delete
* from the various clients we have access to

Note: the version in FreshRSS's database is always the HTML-escaped version.

Alkarex:
It is a mess for News+, which decodes some HTML entities, but not all of them...

https://github.com/noinnion/newsplus/blob/b06d52a101f3913f5b395b1f512f3cc9c1b86c99/extensions/GoogleReaderCloneExtension/src/com/noinnion/android/newsplus/extension/google_reader/GoogleReaderClient.java#L138-L142

But is it not worse than the situation before my changes.

PR #2094, opened 2018-10-30
Changes in Swage theme for Mark As Read on Scroll

pattems:
Remove theme sizing to make scroll as read work per #1980, as requested in https://github.com/FreshRSS/FreshRSS/pull/2088#issuecomment-434309422

I just want to say that I really don't like this solution; it doesn't seem intuitive to me if scroll to mark as read is enabled, and looks downright broken if you don't, which is why I'd overridden the height in the first place. I'll be adding looking into another way to do this to my personal to-do list.

### DIFF
--- a/p/themes/Swage/swage.css
+++ b/p/themes/Swage/swage.css
@@ -778,7 +778,6 @@ display: none;
 text-align: center;
 text-decoration: none;
 background: #e3e3e3;
-padding: 20px !IMPORTANT;
 }
 #bigMarkAsRead:hover {
 background: #22303d;
--- a/p/themes/Swage/swage.scss
+++ b/p/themes/Swage/swage.scss
@@ -873,7 +873,6 @@ form {
 	text-align: center;
 	text-decoration: none;
 	background: darken( $color_light, 10%);
-	padding: 20px !IMPORTANT;
 	&:hover {
 		background: $color_aside;
 		color: $color_light;


Alkarex:
Thanks. Alternative suggestions welcome.
By the way, for me, when it comes to scrolling, the most urgent is to find a more elegant way to decouple the list of feeds (on the left) and the list of articles (on the right) to allow scrolling down to the bottom of a long list of feeds without scrolling the articles, and thus without auto-marking them as read

pattems:
A simple decoupling could be done by just removing sticky-kit, and would leave the feed list at the top, like the behavior of feeds in versions prior to... 1.5, it looks like (#1214)?  

Without sticky-kit, individual scrolling of the sidebar and main feeds could be accomplished purely through css. I've tossed together a quick proof-of-concept here: https://github.com/pattems/FreshRSS/tree/Sticky_aside

This isn't a terribly robust solution, however. There's problems with the last feed or two being obscured by nav buttons if they're visible, but with a bit more work it's a viable, if slightly hacky, pure-css solution.

I'm not terribly good with Javascript, but I think I can probably throw together a JS replacement for sticky-kit that will let us have those decoupled feeds, it might just need a little extra polishing by more capable hands after I'm done with it. I'm not making any promises, but I hope to have something together by My Friday evening/Your Saturday morning.

Issue #2079, opened 2018-10-25
[Feature Request] TT-Rss Favorites import option


typhoe:
Hi,

would it be possible to add an option to import favorites from tt-rss?

Its favorites can be exported as an xml (gzipped) file.

Thank you.

Issue #2018, opened 2018-09-10
Reconsidering read/unread distinction


marienfressinaud:
Over the years, I realized that people were using the distinction between read and unread articles to keep a list of articles they wanted to read later. I honestly never thought of it this way and I personally use the "favorite" articles as a way to list articles that I want to read later (but I realize that it's also a diversion of functionality).

The problem is that it's very easy to lose all its unread articles (there are two buttons, and the warning is regularly skipped :p). The "favorite" feature is less problematic but might not fit with all the use cases.

It would help a lot if people who don't mark all their articles as read regularly can explain what they try to achieve!

lapineige:
(special mention to this number 2018 ;)

lapineige:
> It would help a lot if people who don't mark all their articles as read regularly can explain what they try to achieve!

For my part, I receive too much articles (it's kind of intended) to read them each day - as I'm not always able to read them on a daily (sometimes weekly) basis. But I often have little time in a day to sort them (without reading too much). Then I have a selection that is ready to be read later.

That's a short/mid-term solution. On the long term, I keep the unread articles (that might be interesting - as explained I sort them before) in case I'm searching something one day, to know if I've read the article (I might remember reading this information somewhere in my RSS feed) or not (that's something new).

Alkarex:
By default, the option *mark as read during scroll* is enabled, in particular to reduce this behaviour of not marking as read.

lapineige:
I think the current behaviour is pretty good. I'm just using it in a different way, it's more a diversion of functionality, and maybe an niche use case.
(I'm also using Wallabag as a read-it-later service, but it's an extra software)

GLLM:
Hi !
I'm also using the unread status to keep all articles I want to read but do it later ...
I mark as read all articles I already have read and all those I find not interesting.
And I keep the longer ones unread to see them later.
Once read, if an article was really worth it and I want to keep it, I'll mark it as Favorite !

I'm always very super mega careful when I mark a selection of items as read since I'm always afraid to mark my selection of to-read-later items as unread unwillingly !!

 
**Edit**  : actually, what could be great could be that the articles can be marked as to-read-later ... which would keep them unread until individually marked as read.
This means mark all as read would not affect them ... 

### Workflow
Article arrives in feed:  it is unread.
Then: 
1. I find it useless, I mark it as read
2. I like it and read it : I mark it as read
[...]

PR #2099, opened 2018-10-31
TT-RSS import

Alkarex:
Import of Tiny Tiny RSS favourites from a file like `TinyTinyRSS_exported_2018-10-24.xml`
https://github.com/FreshRSS/FreshRSS/issues/2018#issuecomment-432710462
https://github.com/FreshRSS/FreshRSS/issues/2079

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -398,7 +398,7 @@ class FreshRSS_importExport_Controller extends Minz_ActionController {
 			$item['id'] = isset($item['guid']) ? $item['guid'] : (isset($item['feed_url']) ? $item['feed_url'] : $item['published']);
 			$table['items'][$i] = $item;
 		}
-		return json_encode($table, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
+		return json_encode($table);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


Alkarex:
@typhoe for you :-)
Tests welcome (I have only quickly tested myself).
I have not implemented the import of custom labels yet. It does not look like it is something you use, do you?

Alkarex:
P.S. If you have an example of export with custom labels, I am interested.

Alkarex:
@typhoe I have added the ability to import custom labels as well from TT-RSS. Tests welcome if you can - please let me know :-)
We also need to re-test FreshRSS own export / import, as it was modified during the process.

typhoe:
Hi,

I have tested the custom labels import function from TTRSS favorites with labels and it seems to work fine.

I tried the export/import from FreshRSS and didn't notice any problem.

I also tried to import in TTRSS the exported opml from FreshRSS (just had to unzip it first).
It works without problem. I get back the full feeds with theirs categories.

If I have a few suggestions: 
Display the user name currently connected in FreshRSS on the main page or in the dropdown menu (something like "disconnect USER" in place of simply "disconnect"). The user currently connected is only displayed in the profil page for now (isn't it?).
And add it to the name of the FreshRSS export (freshrss_2018-11-05_export.zip -> freshrss_2018-11-05_export_USER.zip)

Thank you.

Alkarex:
Thanks for the test @typhoe 👍 
I will address the other points in another PR.

Alkarex:
@typhoe I have implemented your suggestions in https://github.com/FreshRSS/FreshRSS/pull/2133
Tests welcome

PR #2108, opened 2018-11-02
Improve long dropdown menu lists

romibi:
### Issue:
Dropdown menus with lots of entries were a bit difficult to use …
I was only able to see either the bottom of the labels-list or the article and not both …
And I was unable to close the labels menu on my Phone after adding more than 12 labels.

### Changes:
I set the max height to ~~60%~~ **75%** of Vieport-height and enabled vertical scrolling … ~~in the theme I use …~~
**for a new dropdown class now used for dropdowns of the "My labels" feature**

### To Further Improve:
- ~~Maybe similar changes to other themes are needed.~~
- On my (windows) phone i still can't add all labels because i cant scroll the menu … but at least i can close it again without reload …
Works ok on my Android …

**Edit:** Changed PR based on responses below

### DIFF
--- a/app/views/helpers/index/normal/entry_bottom.phtml
+++ b/app/views/helpers/index/normal/entry_bottom.phtml
@@ -42,7 +42,7 @@
 			<a class="dropdown-toggle" href="#dropdown-labels-<?php echo $this->entry->id();?>"><?php
 				echo _t('index.menu.tags');
 			?></a>
-			<ul class="dropdown-menu">
+			<ul class="dropdown-menu dropdown-menu-scrollable">
 				<li class="dropdown-close"><a href="#close">❌</a></li>
 				<!-- Ajax -->
 			</ul>
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -230,6 +230,11 @@ a.btn {
 	background: #fff;
 	border: 1px solid #aaa;
 }
+.dropdown-menu-scrollable {
+	max-height: 75vh;
+	overflow-x: hidden;
+	overflow-y: auto;
+}
 .dropdown-header {
 	display: block;
 }


p/themes/Origine/origine.css
```diff
@@ -330,6 +330,8 @@ a.btn {
 	box-shadow: 3px 3px 3px #ddd;
 	font-size: 0.8rem;
 	text-align: left;
+	max-height: 60vh;
+	overflow: hidden auto;
```
Frenzie: That's not valid, is it?

Frenzie:
> And I was unable to close the labels menu on my Phone after adding more than 12 labels.

Could you please explain that a bit more in-depth in order to understand the problem?

Regardless, some kind of media queries should definitely be involved here. Otherwise you get something like this, where it would fit on the screen if it weren't for max-height thing. (That's with what seemed like a more reasonable value, like 75 or 80.)

screenshot_2018-11-02_15-03-43

romibi:
This is the issue:
[Image from Gyazo](https://gyazo.com/bb213593bb0d4f868be03ef35753ba17)

True … feel free to change 60% to another value … it seemed about right for the labels dropdown … there 60% menu and 40% article seemed right

p/themes/Origine/origine.css
romibi: vh is relative to 1% screen height: https://www.w3schools.com/cssref/css_units.asp
true … i'm not sure if overflow-x and overflow-y need to be defined separately for other browsers …

Frenzie: I see, this overflow shorthand is very recent so I wasn't familiar with it. It's only been implemented properly in Fx 63, which came out last week.

https://www.fxsitecompat.com/en-CA/docs/2018/overflow-shorthand-syntax-has-been-updated-to-swap-2-values/

The official line is IE11, but personally I care more about Opera 12. ;-)

romibi: uh they swapped it from firefox 61/62 to 63 … true … better 2 explicit lines?

romibi:
i'd be fine with removing the position: fixed;
but then there is still the issue that if you scroll down the long list of labels you have to scroll up again to get back to the article …

oh and for whatever reason … adding overflow to the css removes the white triangle from the top in the desktop version of the dropdown-menu

and why is there a hidden, unused close button in the dropdown menu?

Frenzie:
> but then there is still the issue that if you scroll down the long list of labels you have to scroll up again to get back to the article …

You could add an `overflow-y: scroll` of some sort I suppose, but it should surely be inside that media query. :-)

> oh and for whatever reason … adding overflow to the css removes the white triangle from the top in the desktop version of the dropdown-menu

To me that's the least of the horrors of applying that style outside of the confines of the problem you were trying to solve. ;-D

> and why is there a hidden, unused close button in the dropdown menu?

Interesting, and it's functional too. I was actually going to suggest to try adding something like that, but that makes it a whole lot easier to experiment…

romibi:
> > but then there is still the issue that if you scroll down the long list of labels you have to scroll up again to get back to the article …
> 
> You could add an `overflow-y: scroll` of some sort I suppose, but it should surely be inside that media query. :-)

no that same issue exists on desktop aka outside that media query …

I was waiting for something like labels for a long time … to categorize my 300+ Favorites i've added 36 labels now … thats a long list … often longer than 2 articles … on some twitter rss feeds even as long as 4 entries

p/themes/Origine/origine.css
```diff
@@ -330,6 +330,8 @@ a.btn {
 	box-shadow: 3px 3px 3px #ddd;
 	font-size: 0.8rem;
 	text-align: left;
+	max-height: 60vh;
```
Frenzie:
```suggestion
	max-height: 75vh;
```
60 seems really tiny to me?

p/themes/Origine/origine.css
Frenzie:
```suggestion
	overflow-x: hidden;
	overflow-y: auto;
```

p/themes/Origine/origine.css
romibi: yea … ok … i still can apply a personal custom stylesheet if i personally want it smaller

Frenzie:
The principle is fine. Outlook online isn't hugely inferior to Outlook desktop (in this particular way; of course it is). At the same time, I always enjoy how easily I can go through FreshRSS's equivalent list by merely pressing page up and down instead of having to scroll on some particular section of the screen.

This is alright, note how it doesn't randomly cut out a big slice of space like Gmail does:
screenshot_2018-11-03_07-41-08

Okay:
screenshot_2018-11-03_07-43-45

But Outlook's online category management, for instance, utterly unusable if you add more, very disappointing and very Gmail-esque:
screenshot_2018-11-03_07-50-06
screenshot_2018-11-03_07-51-34

> i'd even say … yes, a searchbar: let's filter the list below if you start typing a new label … maybe the label you want to create already exists in similar form

Yes, it's a good addition. The problem is that realistically speaking it's the only way of interacting with it. It's just too small otherwise. Like various bizarrely unresizable dialogs in Windows. Most recently I installed new AMD drivers for my GPU and the end user license agreement could be resized neither in window nor font size, nor could it be copied. Making a window resizable is *one* tiny little flag, but ooooh no, perish the though that someone reads it.

Anyway, back to menus and such, on real desktop applications you get functionality from keys like Home and End. Outlook 365 replicates this. Gmail doesn't.

> compare the content to potentially fitting labels:

Interesting use case! I hadn't thought of that.

Setting aside the discussion on menus in general, I really do think that at its best this change wouldn't mess up the other menus too badly while they're better left without. The labels menu seems to be a special case. The application of the scrolling principle can be made generalized, like `class="dropdown-menu dropdown-menu-scrollable"`.

It could also be shorter on thinner windows (so as to cover up less of the content) than on wider ones.

You made a convincing argument that it might be okay for the labels menu not to take up as many pixels as possible, but my biggest objection is to the indiscriminate application of this principle.

screenshot_2018-11-03_08-23-16

Let's say we want all menus to be scrollable. Most menus should take up as much space as they possibly can. 100vh-some padding and stuff. Then there may be menus where that doesn't apply, although like I said before, I think the problem is just that it doesn't open with a negative offset. :-)

Alkarex:
I have not followed everything. What is the status of this PR, @Frenzie ?

Frenzie:
Not quite ready yet I'd say. @romibi could you perhaps refactor it slightly with the `dropdown-menu-scrollable` (or something akin to that) to limit it to the filter menu? :-)

romibi:
Wasn't sure if you wanted to do it …
… is this ok?

Frenzie:
Lgtm, thanks!

Alkarex:
Thanks @romibi 👍 
Merged in /dev

Issue #1593, opened 2017-07-12
Feed menu jumps down to bottom


thomasrenes:
When using the the keyboard shortcuts to browse through the articles, sometimes the feed menu on the left randomly jumps down to the bottom of the screen.

After jumping to the previous feed (up arrow on keyboard) the menu jumps back up (where it's suppose to be) and stays there, even when going to the next article again.

I'm not able to replicate this issue, it just sometimes (once or twice a week) happen.

![feed_menu](https://user-images.githubusercontent.com/4504804/28104467-793018f4-66db-11e7-9cfc-00b8811706be.png)

FreshRSS Version: 1.7.0
FreshRSS Theme: Flat design
Browser: Chrome 59.0.3071.115 (64-bit) on OS X
Screen Resolution: 1440x900

Alkarex:
I have also noticed the issue occasionally. Let me know whether you find out when it happens a bit more precisely

Alkarex:
Might be the same as https://github.com/FreshRSS/FreshRSS/issues/1588

Alkarex:
@thomasrenes Could you please give a try to this tiny patch? https://github.com/FreshRSS/FreshRSS/pull/1616

thomasrenes:
@Alkarex Have applied the patch and will let you know in a few days if it works.

Alkarex:
Any improvement, @thomasrenes ? :-)

thomasrenes:
Sorry for not responding quicker.

This "jumping" issue has not occurred with the fix, but this seem to introduce some issues with the menu when opening FreshRSS in Chrome on an iPad. When scrolling down the menu disappears and seems to be only available when the browser position is on the top or bottom of the page.

thomasrenes:
One more thing: In the desktop version the top item of the menu (the one below "Favorites") also disappears when scrolling down the news items list, while there is still enough space left to display the full menu. When scrolling up again the menu is fully displayed.

![scrolling_up](https://user-images.githubusercontent.com/4504804/30760638-3bf25e32-9fdb-11e7-974b-eb63bb7793b1.png)

![scrolling_down](https://user-images.githubusercontent.com/4504804/30760639-3bfb1e6e-9fdb-11e7-950a-60acb659f7bd.png)

Alkarex:
The UI of the list of categories and feeds has been fully reworked in the /dev branch (FreshRSS 1.13.0 will be released in a few days). Tests and feedback much welcome, especially before the stable release.

Issue #1588, opened 2017-06-20
Scrolling Jumps Upon Feed Retrieval (Paging)


vindicatorr:
FreshRSS: 1.7.0
Chrome: 59.0.3071.86

I'm finding that when I'm scrolling through the feeds, when the next bunch of feeds are retrieved in the browser (paging), the feed position jumps much farther down and I have to scroll back up to find where I was.

Alkarex:
The UI of the list of categories and feeds has been fully reworked in the /dev branch (FreshRSS 1.13.0 will be released in a few days). Tests and feedback much welcome, especially before the stable release.

PR #2117, opened 2018-11-04
Decouple scrolling of feeds and articles

pattems:
Per https://github.com/FreshRSS/FreshRSS/pull/2094#issuecomment-434488436.

There's a couple of minor issues I'm not skilled enough with javascript to resolve:

- Resizing  the window doesn't trigger a recalculation of the tree height; that's not done until a scroll event. All that's needed is an onresize event to trigger sticky_recalc(), but I wouldn't begin to know where to put it.
- I stuck the line to style the scrollbar at the end of simple-scrollbar.js; I couldn't work out how to call it from main.js after injecting it.

I also have a branch that doesn't include the addition of simple-scrollbar.js, and just uses the browser's built-in scrollbar, if you all would rather stick with a default scrollbar for the sidebar, but that doesn't seem to be how most other examples I've seen with independent scrolling have gone, and (in my opinion, anyways) it's rather ugly.


Frenzie:
Do you have a screenshot for curious people on cellphones? :-P

Alkarex:
@pattems I have made a few changes https://github.com/FreshRSS/FreshRSS/pull/2117/commits/286217d6577d89dcb6befc7f20d96d6f90fefb85
Please have a look. When testing the loading, it is interesting to test with various network throttling and cache settings in the browser developer tools.
There seems to be a condition when the tree is all white before the first scroll.

@Frenzie 
image

Alkarex:
@Frenzie Hum, could be a good idea with JSHint indeed, if you could please try to add it to Travis.
P.S.: Only for our own JS, not for third-party libraries.

Alkarex:
@pattems Could you please post a couple screenshots highlighting the pros and cons of simple-scrollbar.js vs. native?

Alkarex:
The mobile view does not seem to be optimal yet, with some kind of double vertical scroll-bar in my test

Alkarex:
@pattems Please say also if I have made any regression compared to your first version

pattems:
@Frenzie I was having trouble figuring out how to adequately show things with just screenshots, so I recorded a quick video: https://www.youtube.com/watch?v=eKpLNKH8y5I

@Alkarex Right after I recorded that, I managed to lock myself out of my test instance, so it'll be a bit before I've got more screenshots, or will be able to comment on the commit you made.

Alkarex:
If you have lost your FreshRSS admin password, edit data/config.php and set auth_type to none, after which you can set back your password and login method from within the UI

pattems:
Thanks for telling me how to get back in! Much faster than blowing away the instance and recreating it. 

There is a regression; if the page is scrolled down and then back to the top, the sticky class isn't being removed. I just pushed a commit fixing that, but there might be a more elegant way to do that. I'll have screenshots in a couple minutes.

pattems:
Screenshots, as promised.

First, a baseline reference; this is the same with or without simple-scrollbar. 
is_desktop_noscroll


Here's with the native scrollbar:
is_desktop_scroll


And with simple-scrollbar, when the aside is hovered over (otherwise the scrollbar disapears): 
ss_desktop_scroll_mouseover


Other than matching what I'd expect from a UI pattern for an independent scrollable area in a modern web-application, simple-scrollbar is an overlay, and doesn't push content over by the width of a scrollbar.

pattems:
@Alkarex When you were telling me about testing the loading, was that general advice, or a reference to the fact that with caching disabled, it could take a few seconds after initial pageload to display the overlay scrollbar? Because that seems to have been resolved by your rewriting of my script, and I'm not seeing anything that particularly alarms me; more experienced eyes may disagree, though. I'm working on improving my skills, but mostly I'm just a tinkerer.

The condition when the tree is all white before the first scroll is something I haven't encountered. Is it something you can reproduce, or something that just happens sometimes?

Alkarex:
@pattems Testing with various caching / throttling was more a general advise, but especially here due to the fact that we want to load this feature without blocking the rest. The all-white tree happened a few times, but I have not yet looked into it (maybe a wrong offset in some race conditions). I will test some more tonight.

Alkarex:
@pattems I could reproduce the bug as well as identify the problematic rule:

image

```
$sidebar.parent().width() = 1203
$(window).height() = 969
$sidebar[0].getBoundingClientRect().top = 1830.86669921875
h1 = -918.86669921875
```

Alkarex:
Incidently, I can see that the native CSS3 `position:sticky` has gained some good support (Firefox, Chrome, Edge, Safari, ...) https://caniuse.com/#feat=css-sticky
and there is a polyfill for older browsers https://github.com/wilddeer/stickyfill
Demo: http://wd.dizaina.net/en/scripts/stickyfill/

Alkarex:
Just for testing, I have temporarily introduced an option to toggle between the JavaScript scrollbar and the native one.

Firefox with JavaScript scrollbar
image

Firefox with Windows native scrollbar
image

Firefox with Ubuntu native scrollbar
image

I tend to agree that the JavaScript scrollbar looks nicer, but:
1) A significant amount of JS + CSS is needed + listening on expensive events
2) It might be less accessible, at least on some platforms
3) I can still produce some small bugs in various situations
4) It seems to increase the frequency at which I observe layout flashes during page load

So I am a bit unsure. I would be pleased to see some tests with the newer native approach `position:sticky` (which will not help for the style, but should reduce the code and event listeners even more)

pattems:
Sorry for the slow reply; I've been swamped with my day job.

I just added a bunch of commits from a branch I was working on to test the native css sticky.

I also took steps to mitigate the frequency of recalculation on scroll, which should lessen some of the expense of the independent scrolling sidebar.

Since you added the ability to easily toggle simple-scrollbar, if you find that the layout flashing and bugs you're encountering are fixed by turning it off, I'm certainly okay with yanking that part out of the code for now, and exploring alternate approaches on my own as future improvement.

Alkarex:
Thanks @pattems and no worries; I was also waiting for the week-end myself to have a bit more time :-)
I will check! What is your impression with native sticky CSS so far?

p/scripts/main.js
```diff
@@ -1259,7 +1234,54 @@ function init_crypto_form() {
...
 
+var scrollTimeout;
+function init_sticky_sidebar(){
+	$sidebar = $('#sidebar');
+	if (useNiceScrollbar) {
+		init_simple_scrollbar();
+	}
+	$(window).scroll(function () {
+				if (scrollTimeout) {
+					clearTimeout(scrollTimeout);
```
Alkarex: Just at a quick glance, it looks like if the user scrolls continuously, the recalculation will never be called (until the scrolling ends), will it? :-)

Alkarex:
I need to test a bit more, but this looks quite good 👍 

* Regarding native vs. JavaScript-enhanced scrollbar, I am still very much undecided (screenshots / argumentations for one or the other very much welcome), and might end up making it an option for the user to pick

* Regarding native vs. JavaScript sticky, it would be a pity to always load `stickyfill.min.js` even for browsers not needing it, so I believe we should implement one of the following approaches (I think I can better like approach number 2):
  1. Revert to previous full-JavaScript approach (based on your comment that there was no clear gain)
  2. Full native CSS3 sticky with no sticky functionality for older browsers, but ensuring that the menu still works without sticky (e.g. using a fixed, or absolute position)
  3. Native sticky with dynamic loading of the `stickyfill.min.js` when needed, based on feature detection

Alkarex:
By the way, in my slightly delayed reading list, this article, slightly related: https://css-tricks.com/practical-css-scroll-snapping/

pattems:
- On the scrollbar: another option to consider would be using the -webkit-scrollbar css property, which would target 86-ish% of global users, and on non-webkit based browsers it'd just fallback to the native scrollbar. https://caniuse.com/#search=webkit-scrollbar That removes one JS dependency entirely, and simplifies the css. I've got a branch read to merge into this one with those changes made, if you like that option. It's the css_scrollbar branch in my repo, if you want to give it a test-run.

- As for stickyfill - I agree with that. There's a slight gotcha to using a fixed/absolute position as a fallback with option 2: The sticky property requires a top; in order to have it stick to the top of the screen, we need to specify `top: 0;`.

 We could specify `position: relative;`, which would allow scrolling down the list of feeds separately from the list of articles. The downside to this would be that, on browsers not supporting sticky, the list of feeds would be stuck at the top of the page, like things worked before the addition of sticky-kit. This wouldn't effect mobile browsers, and the only desktop browser we should see this behavior on, if I'm reading canIUse right, would be Internet Explorer.

Alkarex:
Thanks @pattems , it looks like we are converging to something quite good, with your work.
From memory, Internet Explorer is also supporting some custom scrollbar CSS settings (in FreshRSS, we try to support Internet Explorer 11 and Edge).
I will test your branch tonight.

pattems:
Internet Explorer supports changing the color of the scrollbar, but doesn't seem to support transparency, and the vendor prefix was dropped entirely in Edge, giving us no scrollbar customization options. There also aren't any in Firefox, so on those browsers we'd fallback to default scrollbars.

Also, in testing, I've discovered that sticky also doesn't work on Edge either. I personally don't feel that the behavior is broken on the Microsoft Browsers, just less functional than on other browsers, and I'd be perfectly happy with leaving it as-is. If you want a more consistent UX, we could look at your suggested approach 3, and apply the polyfill to IE and Edge, and use the JS scrollbar on IE, Edge, and Firefox.

pattems:
As an aside, looking through open issues, this decoupling should take care of #1588 & #1593 when merged.

Alkarex:
@pattems I am testing your branch `css_scrollbar`, which looks very promising 👍 I think we should go this way.
One bug spotted: expand a few categories, so that the category tree is higher than the window, then scroll down to the last feed, and try to open its configuration menu. First, the menu is not accessible, second, the feed gets a bit hidden, and third, the scrolling does not work (well) anymore. Tested in Chrome and Firefox.

image

P.S.: Similar problem in Edge and IE.

Alkarex:
@pattems For the sticky native fallback, maybe a combination of `top` and `margin` could help (not tried).

pattems:
It's not so much that scrolling stops working, as it is that the dropdown becomes the scrollable element. This is decidedly weird and unacceptable behavior. It's the absolute positioning of .dropdown-menu that causes this, and I think I've got the start of a fix in my head, I'll start working on this evening.

on the css-scrollbar styling- What would you prefer to do about non-webkit browsers? leave those scrollbars unstyled, or attempt polyfilling simple-scrollbar.js on non webkit browsers?

Alkarex:
For standards / Firefox, there is some hope: https://drafts.csswg.org/css-scrollbars-1/ 😃 
https://bugzilla.mozilla.org/show_bug.cgi?id=1460109 (I have just voted for this issue)

In the meantime, I think we could conditionally inject `simple-scrollbar.js`, if possible based on feature detection.

Alkarex:
@pattems To implement the fallback for browsers without sticky, I think the easiest might be the `@support` CSS rule https://developer.mozilla.org/docs/Web/CSS/@supports
Likewise, for conditionally injecting the JavaScript `simple-scrollbar.js` , we can use the `CSS.supports()` function https://developer.mozilla.org/docs/Web/API/CSS/supports

Let me know if there is any difficulty.

Alkarex:
@pattems I have manually merged your `css_scrollbar` branch into this branch, to use the more native features of `css_scrollbar` while keeping e.g. JS scrollbar for non-webkit browsers.
I have added feature detection of `::-webkit-scrollbar-thumb` and `position: sticky`.
Please have a look :-) https://github.com/FreshRSS/FreshRSS/pull/2117/commits/866ef64c8aeee8c5f9eac7be38d6ebc2f72b9616

Alkarex:
Note: I have not been able to immediately find from when the non-prefixed version of `-webkit-sticky` has been available, nor whether Firefox has ever used `-moz-sticky`
https://developers.google.com/web/updates/2012/08/Stick-your-landings-position-sticky-lands-in-WebKit

Frenzie:
According to https://caniuse.com/#feat=css-sticky never, just a config switch.

pattems:
Sorry I've been away, I've been having some computer troubles. But echoing what Frenzie said, it looks like Firefox never used a prefixed version, and has worked with the non-prefixed version since 2014, and every mainstream desktop browser except for IE11 has supported it either natively or prefixed since the beginning of 2017. 

@Alkarex I haven't had a chance to test your code yet, but it looks good, and I should have a chance to test tomorrow.

Todo, as it stands now:

- [x] Workaround for lack of sticky on IE11
- [x] Fix the Configuration menu bug

I think the cleanest suggestion for tackling IE11 is checking for IE11 with a couple of lines of JS and injecting stickyfill if needed. I can't foresee this causing a significant performance impact, but I'll do a little testing, and try to come up with something else if it turns out to be problematic.

I'm pretty sure the fix I've got in mind for issue 2 will work, I've just got to scrape together a few hours to code it, and will hopefully have that done by the end of the weekend.

Alkarex:
@pattems I am glad to read from you again, as I was getting worried :-)
Let's not do too much for IE11: I think it is sufficient to have a non-sticky menu for IE11 - it is still usable. As long as it only requires a few fallback CSS rules or lightweight JS helper, then fine, but shipping a JS library for it too much.

Frenzie:
> I think it is sufficient to have a non-sticky menu for IE11 - it is still usable.

My point in #2159 to a tee. ;-)

pattems:
I tracked down the first half of the issue with not being able to scroll the nav when the menu was expanded, referenced here https://github.com/FreshRSS/FreshRSS/pull/2117#issuecomment-439187010. The absolute positioning of the dropdown, coupled with the way that they're currently closed (invisible href over the entire canvas pointing to #close) was causing issues.... though I couldn't replicate the bug on Firefox. I've got a promising alternative solution using javascript click events that I'm testing out in another branch, found here: https://github.com/pattems/FreshRSS/tree/Scrollbar_menuFix

So far, it seems to work fine across the browsers that are tested for, but I want to do a little more thorough testing when I haven't just wrapped up writing the code, to make sure I haven't missed anything obvious, before I merge it into the pr branch.

As far as the menus getting cut off goes, my plan is to move the dropdown target https://github.com/FreshRSS/FreshRSS/blob/7cbfdb4e0932230742060b1d04e86f221a7cb10d/app/layout/aside_feed.phtml#L82
after the title (line 86, same file), removing the absolute positioning from sidebar dropdown menus. This works fine and fixes the second half of the problem when I make the edits in dev tools, except for the fact that, because the dropdown-target comes before the feed name, it obscures the name of the feed that it'd supposed to be dropping down from by pushing it right, so I'll need to move the dropdown-target after the feed name.

Where I've hit a wall in that I can't figure out how to get a moved .dropdown-target assigned an id. I'm pretty sure that the assignment's being done here https://github.com/FreshRSS/FreshRSS/blob/e04804d0f67dd43fd3f072b9a127768ee7b7b56c/app/views/helpers/index/normal/entry_bottom.phtml#L56 but I'm not really sure.

That id assignment is really the last stumbling block, and once that's cracked, I should be able to wrap up the rest of the code in an evening... so if anybody can give me pointers about how to get that done, it'd be greatly appreciated!

Alkarex:
@pattems Before going into more advanced solutions, wouldn't a bigger margin at the bottom of the tree of categories / feeds solve the problem?

The dropdown is using a template, and is instantiated in:

https://github.com/FreshRSS/FreshRSS/blob/7cbfdb4e0932230742060b1d04e86f221a7cb10d/p/scripts/main.js#L570-L585

P.S.: And declared in:

https://github.com/FreshRSS/FreshRSS/blob/7cbfdb4e0932230742060b1d04e86f221a7cb10d/app/layout/aside_feed.phtml#L110-L131

pattems:
@Alkarex I got so caught up in my idea that I didn't stop to consider that there might be a simpler solution. Adding a margin didn't work, but fixing the overflow turned out to be as simple as removing the treepadding that I should have removed when I started the decoupling process, so I'm really glad you suggested that.

Eventually, I still think it would be cleaner to expand the menus inline the same way the fluxes are so that they don't obscure other menu items, but that's easy enough to push for future improvement.

I've just merged in the current dev branch and, unless anybody spots anything else broken, we should be good to go on this one.

p/scripts/main.js
```diff
@@ -1511,6 +1540,23 @@ function parseJsonVars() {
 	window.icons = json.icons;
 }
 
+function init_dropdown_handler() {
```
pattems: One thing that still needs to be done for sure is stick this in a more sensible place... I just kind of arbitrarily added it in before the actual initializations because I haven't really groked the organizational structure of main.js yet.

Alkarex: I am not sure there is a great logic behind the current organisation 😄 
Things are supposed to be more or less grouped, trying also to put functions before their calls (minimising our dependency on [JavaScript hoisting](https://developer.mozilla.org/docs/Glossary/Hoisting)

Alkarex:
Thanks @pattems 👍 I will try to test tonight.
Are the changes made in https://github.com/FreshRSS/FreshRSS/pull/2117/commits/0e5188f327540190d8599c773bb816d6152f7edc still necessary?
The menu used to work without JavaScript (but I rarely test that, and quickly checking now, it does not seem to be the case anymore, even before this commit).

p/scripts/main.js
pattems: Ah, well that would certainly explain why I was having a hard time getting a feel for it, then!

pattems:
Moving the dropdown target can probably be reverted. I can't actually remember why I did that, so I'm testing reverting that bit now. As for the rest of the changes in https://github.com/FreshRSS/FreshRSS/commit/0e5188f327540190d8599c773bb816d6152f7edc, swapping out the overlay href for a JavaScript close, those changes are still necessary. If the sidenav dropdowns were displayed inline instead of absolutely positioned, I think we could get away with using the overlay href still, but I'd abandoned that approach for the more straightforward css solution with the padding issue. 

The dropdown's in the menu definitely depend on JavaScript, because (checked in 1.12.0) the .dropdown-menu objects don't exist in the page on pageload, but each menu is inserted the first time it's corresponding toggle is clicked. But you probably wouldn't even notice that, because without JavaScript, you can't even expand a category to get to a toggle in the first place.

pattems:
Explaining that last commit: Firefox (at least my testing version, Firefox 63.0.3 on Windows 10) was ignoring the scrollbar color for dark themes when it was grouped with the webkit style. I left them grouped, but also broke a single .ss-scrollbar rule out on it's own as a duplicate, on the thinking that this is probably a Gecko bug that might get fixed before (if?) Firefox implements native scrollbar styling, rendering the JS scrollbar unnecessary.

p/themes/BlueLagoon/BlueLagoon.css
```diff
@@ -566,6 +567,11 @@ a.btn {
 	color: #0090FF
 }
 
+/*=== Scrollbar */
+.ss-scroll, #sidebar:hover::-webkit-scrollbar-thumb {
```
Alkarex: Forgotten split here, like for the other themes?

p/themes/Dark/dark.css
```diff
@@ -502,6 +502,15 @@ a.btn {
 	color: #888;
 }
 
+/*=== Scrollbar */
+.ss-scroll {
+	background: rgba(255, 255, 255, 0.3);
+}
+
+#sidebar:hover::-webkit-scrollbar-thumb, .ss-scroll {
```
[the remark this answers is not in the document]
pattems: Not necessary, I just thought that It would make it a little more obvious that the split was intentional. On second thought, a comment probably would make more sense.

p/themes/BlueLagoon/BlueLagoon.css
pattems: Yep, certainly did!

p/themes/Dark/dark.css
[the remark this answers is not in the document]
Alkarex: Sorry, I missed your explanation higher up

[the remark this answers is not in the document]
Alkarex: Yes indeed, I also think a very short comment would be better, also to avoid a routine optimisation.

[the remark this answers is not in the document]
Alkarex: And it does not help to swap the two rules?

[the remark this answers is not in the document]
pattems: No worries! The comment's probably the better course of action, anyways.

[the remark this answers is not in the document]
pattems: Nope, tried it both ways. Firefox just stubbornly refuses to pick up on the combined line in any combination.

Alkarex:
I have just tested on a few different browsers, and it is very nice 😃 I have only some minor points left:

* [x] If possible, I would like to add a margin (slightly higher than a drop-down menu) at the bottom of the tree of categories / feeds, to be able to get the bottom-most feed or category a bit higher than the very bottom of the window. This would allow opening the drop-down menu of the bottom-most feed without having to scroll 2-3 times.

* [x] It has become a bit more difficult to close a drop-down menu without clicking on one of its links. In particular, re-clicking on the drop-down menu icon does not close the menu anymore.

* [x] (Only buggy with simple-scrollbar.js - works fine with native scrollbars) Starting from a window that is higher than the tree of categories / feeds, there is no scrollbar. Then expand a few categories / feeds to make the tree higher than the window. The scrollbar does not appear until the next scroll event. Likewise, the height of the scrollbar is not recalculated upon expanding / collapsing of items until the next scroll event. It is not a bit deal though.

* [x] Opening a down-down menu and pressing escape to close it generates a `#null`. Probably not a problem. Just a need to double-check.

And we should give a try at https://bugzilla.mozilla.org/show_bug.cgi?id=1460109 in Firefox 64 (probably in another PR)

pattems:
I'll hopefully have a chance to look into those this evening!

Alkarex:
Ha, the CSS parsing bug is fixed in Firefox 65... but my try/catch feature detection does not work anymore!

```javascript
document.body.matches('::-webkit-scrollbar-thumb');
```

~~Will fix.~~ Edit: Fixed in https://github.com/FreshRSS/FreshRSS/pull/2117/commits/0ea661c55cc09b88bbb9e044a2ba766a5cebadc1

p/scripts/main.js
```diff
@@ -1286,7 +1277,54 @@ function init_crypto_form() {
 }
 //</crypto form (Web login)>
 
+var $sidebar = null;
+var useJsScrollbar = !(parseInt(getComputedStyle(document.getElementById('sidebar'), '::-webkit-scrollbar').width) < 99);
```
Alkarex: A few notes:
* This code will be safely executed after CSS loading thanks to the script `defer` attribute, and may be executed before jQuery is ready as there is no use of jQuery there.
* `parseInt()` will return `NaN` when the property is empty (e.g. in Firefox), and `NaN < 99 == false`;
* Edge / IE would return the width of the whole sidebar, which is about ~300.

p/scripts/main.js
```diff
@@ -1286,7 +1277,57 @@ function init_crypto_form() {
 }
 //</crypto form (Web login)>
 
+var $sidebar = null;
+var useJsScrollbar = true;
+try {
+	useJsScrollbar = !(parseInt(getComputedStyle(document.getElementById('sidebar'), '::-webkit-scrollbar').width) < 99);
```
Alkarex: A few notes:
* This code will be safely executed after CSS loading thanks to the script `defer` attribute, and may be executed before jQuery is ready as there is no use of jQuery there.
* `parseInt()` will return `NaN` when the property is empty (e.g. in Firefox), and `NaN < 99 == false`;
* Edge / IE would return the width of the whole sidebar, which is about ~300.

pattems: Forgive me if there was a way to merge in this change; I added it manually since I didn't see how to.

Alkarex: I believe a simple `git pull` should have worked :-)

pattems: That was what I thought, and what I'd tried, but it didn't seem to pull it! I was probably just doing something wrong, still getting used to git.

p/scripts/main.js
```diff
@@ -1280,14 +1279,20 @@ function init_crypto_form() {
 var $sidebar = null;
 var useJsScrollbar = true;
 try {
-	useJsScrollbar = !(parseInt(getComputedStyle(document.getElementById('sidebar'), '::-webkit-scrollbar').width) < 99);
```
Alkarex: Ah, it is because my changes where manually overridden in this commit. Just a guess: what might have happened it that you pulled, but had a file opened in a local text editor, which did not get refreshed, so an older version was then saved.

p/scripts/main.js
```diff
@@ -1313,9 +1318,6 @@ function init_simple_scrollbar() {
 var scrollTimeout;
 function init_sticky_sidebar(){
 	$sidebar = $('#sidebar');
-	if ($sidebar.length < 1) {
```
Alkarex: This change was lost as well. I will check in the newest version to re-apply some of my changes.

pattems: Sorry again for the headaches I seem to have caused!

Alkarex: No worries :-)

Alkarex: Should be fixed with https://github.com/FreshRSS/FreshRSS/pull/2117/commits/ef0c7ed367664282bb055f57bbba63855ca9b93a

p/scripts/main.js
Alkarex: I am making a new detection to support Firefox 64.

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/2117/commits/68e30eb98bcaa4c167ae9f187bf6d1b829642128

Alkarex:
Updated CSS scrollbar feature detection https://github.com/FreshRSS/FreshRSS/pull/2117/commits/68e30eb98bcaa4c167ae9f187bf6d1b829642128 to support standard syntax (e.g. from Firefox 64)

Alkarex:
Minor thing, but the non-dynamic menus (e.g. configuration button at the top-right corner) cannot be closed easily by re-clicking on them, like is possible again on dynamic ones.

p/themes/BlueLagoon/BlueLagoon.css
```diff
@@ -569,6 +569,14 @@ a.btn {
 
 /*=== Scrollbar */
 
+#sidebar {
+	scrollbar-color: rgba(0, 0, 0, 0.0) rgba(0, 0, 0, 0.0);
```
Alkarex: The Alpha value at 0 make them not appear. Seems to work when fixing it.

p/themes/base-theme/template.css
```diff
@@ -425,6 +425,16 @@ a.btn {
 	user-select: none;
 }
 
+#sidebar {
+	scrollbar-width: thin;
+	scrollbar-color: rgba(0, 0, 0, 0.0) rgba(0, 0, 0, 0.0);
```
Alkarex: The Alpha value should be increased to be visible.

Alkarex: Ok. But for some reason, the current code does not work in my tests with Firefox 65: the scrollbar never gets visible, even when hovering.

Alkarex: I tested with https://www.mozilla.org/firefox/developer/

pattems: I've been able to see the issue myself. Additionally, despite the fact that it was working last night in v64, it's not working today. I even rolled back to my first implementation in ab29de9 to ensure that it wasn't a regression caused by subsequent changes, and that's not the case.

Anyways, the issue seems to be that Firefox isn't picking up on the #sidebar:hover on page load. If you make any changes in the block in inspector view, it picks it up and works fine until the next page load.

Are you content with just making the native scrollbar not hide when the tree's being not being hovered over for now?

pattems:
Latest commit should address https://github.com/FreshRSS/FreshRSS/pull/2117#issuecomment-444791226

p/scripts/main.js
```diff
@@ -565,20 +565,25 @@ function init_column_categories() {
...
 }
 
+var dropdownHref = null;
+
 function init_dropdown_handler() {
 	$('#global').not($('.dropdown')).click(function () {
 		window.location.hash = "close";
 	});
+	
+	$('.dropdown-toggle').click(function() {
```
Alkarex: This adds many listeners, which I try to avoid, even though the extra resource consumption might not be so significant. I will try to check if it is possible to move it back to a single listener.

p/themes/base-theme/template.css
```diff
@@ -373,16 +369,91 @@ a.btn {
...
 	text-align: left;
-}
-
-.treepadding {
-	padding: 0 0 15em 0;
+	position: absolute;
+	overflow-y: auto;
+	overflow-x: hidden;
+	width: 300px;
+	height: 90vh;
```
Alkarex: The Firefox bug seems to be due to this property

p/themes/base-theme/template.css
Alkarex: Please see this quick attempt https://github.com/FreshRSS/FreshRSS/pull/2117/commits/ab9817e79377c0d28bac614870a794f119c5b068
But removing the `height: 90vh;` might have broken something else. I did not have time to check yet. Will have time again on Sunday.

p/scripts/main.js
pattems: I'll be most curious to see how you do it, if you can move it back to a single listener. I _think_ that my JavaScript has improved some since I've had occasion to do more with the project, but that's still just an improvement from blundering idiot to slightly-less-blundering idiot.

p/themes/base-theme/template.css
pattems: Looking at the commit history, the height property here was when I was thinking that I could break the scrolling out without javascript. It's removal certainly doesn't seem to break anything.

p/scripts/main.js
Frenzie: I'd argue it's hard if not impossible to improve your JavaScript and DOM scripting by playing with jQuery, but that aside.

At a quick glance it looks like you're already adding effectively the same event listener a few lines higher (557 original, 543 post-diff), which might be what @Alkarex is getting at.

pattems: @Frenzie, noted on jQuery. Just wanted to be clear that, poor as it may be, I am making an effort to improve. There's just a lot I don't know that I don't know, and I'm trying to chip away at at least that. As always, I appreciate your guidance.

Pointing out the effectively duplicate listener made it obvious to me; it's the same thing, just for a different scope. I _think_ that this is on the right track to taking care of that, but I didn't want to commit it in the event I was wrong.

https://github.com/pattems/FreshRSS/blob/80caae05a595a25471f54239d20b8aa69986b80e/p/scripts/main.js#L538-L589

Alkarex:
Firefox 63 had a permanent scrollbar, fixed with conditional CSS https://github.com/FreshRSS/FreshRSS/pull/2117/commits/0e4b115485cbd8de17d92130e84818bda277a0b1
image

There was also a bug with JavaScript scrollbar, which was only refreshed at the first scroll. Fixed as well.

I also think I can better like having a subtle scrollbar than no scrollbar at all when the mouse is outside the tree area (similar to native scrollbar behaviour). Testing that at the same time (only with native scrollbars for now and with the themes not redefining the scrollbar - I think each theme should do what the author wants).

@pattems I have not understood when the `width: inherit;` is necessary. Please re-add if needed. You might have to adapt the *Swage* theme to use the same fixes as above (conditional CSS) if you want a custom scrollbar in your theme.

P.S.: It does not seem to be possible to close the top-right configuration menu without JavaScript anymore, but as long as it is possible to open it, I think this is good enough.

p/scripts/main.js
Alkarex: Not tested yet, but this looks good @pattems 👍 
Sorry for my short message. In jQuery parlance, what I wanted is called *delegated events* and there is an explanation of the advantages on https://api.jquery.com/on/

Alkarex:
One last (I hope) bug:
* [ ] After opening and closing a dropdown menu, it is not apparently possible to open the same menu again.

Alkarex:
P.S. @pattems Just in case you would like to apply e.g. scroll only in some conditions, more advanced CSS conditions can be used, such as:

```css
@supports (scrollbar-width: thin) and (not (overflow-y: overlay)) {
	#sidebar {
			overflow-y: scroll;
		}
	}
}
```

p/scripts/main.js
```diff
@@ -570,24 +572,33 @@ function init_column_categories() {
 		} else {
 			var id = $(this).closest('.item').attr('id').substr(2);
 		}
+        
+        if ($(this).next('.dropdown-menu').css('display') === 'none'){
```
Frenzie: space

p/scripts/main.js
```diff
@@ -570,24 +572,33 @@ function init_column_categories() {
...
+            if ($(this).attr('href') != '#close') { sidebarHref = $(this).attr('href'); }
+                $(this).attr('href', sidebarHref);
+            } else{
```
Frenzie: space (and syntax error? deceptive indentation for sure)

Frenzie (commented):
Also the indentation here is very weird. ;-)

p/scripts/main.js
Frenzie: Hmm, looks like I missed an update (or rather GH didn't notify me). Sorry about that.

pattems: yep. Just pushed to take care of that.

README.md
```diff
@@ -199,13 +199,13 @@ Supported clients are:
 * [php-http-304](https://alexandre.alapetite.fr/doc-alex/php-http-304/)
 * [jQuery](https://jquery.com/)
 * [lib_opml](https://github.com/marienfressinaud/lib_opml)
-* [jQuery Plugin Sticky-Kit](https://leafo.net/sticky-kit/)
 * [keyboard_shortcuts](http://www.openjs.com/scripts/events/keyboard_shortcuts/)
 * [flotr2](http://www.humblesoftware.com/flotr2)
 
-## Only for some options
+## Only for some options or configurations
```
Frenzie, line 205: Only for optional features?

p/scripts/main.js
```diff
@@ -564,10 +552,14 @@ function init_column_categories() {
...
-		$(this).parent().next(".tree-folder-items").slideToggle(300 , function() { $(document.body).trigger("sticky_kit:recalc"); });
+		$(this).parent().next(".tree-folder-items").slideToggle(300, function () {
+			if (useJsScrollbar && sidebar) { //Refresh JS scrollbar
+				sidebar.querySelector('.ss-content').dispatchEvent(new Event('scroll'));
+			}
+		});
 		return false;
 	});
-	$('#aside_feed').on('click', '.tree-folder-items .feed .dropdown-toggle', function () {
+	$('#aside_feed').on('click', '.tree-folder-items .feed .dropdown-toggle', function () {
```
Frenzie: Trailing whitespace

pattems:
Well, I've clearly been making a mess of things all morning. I've got a new branch not on this PR so I can try things without pushing frustratingly bad commits.

p/scripts/main.js
```diff
@@ -577,25 +569,42 @@ function init_column_categories() {
...
 				.append(template).find('button.confirm').removeAttr('disabled');
-			$('.tree-folder-items .dropdown-close a').click(function(){
-				$('.tree').removeClass('treepadding');
-				$(document.body).trigger("sticky_kit:recalc");
-			});
+		} else {
+			var id = $(this).closest('.item').attr('id').substr(2);
+		}
+        
+		if ($(this).next('.dropdown-menu').css('display') === 'none'){
```
```suggestion
		if ($(this).next('.dropdown-menu').css('display') === 'none') {
```

p/scripts/main.js
```diff
@@ -577,25 +569,42 @@ function init_column_categories() {
...
+			if ($(this).attr('href') != '#close') { sidebarHref = $(this).attr('href'); }
```
Frenzie: @Alkarex Should that be !==?

p/scripts/main.js
```diff
@@ -1286,7 +1296,57 @@ function init_crypto_form() {
 }
 //</crypto form (Web login)>
 
+var sidebar = document.getElementById('sidebar');
+var useJsScrollbar = true;
+try {
+	useJsScrollbar = sidebar && !CSS.supports('scrollbar-color: auto') &&
+		!(parseInt(getComputedStyle(sidebar, '::-webkit-scrollbar').width) < 99);
```
Frenzie: What's this magic number? I generally prefer using a self-explanatory named variable.

p/scripts/main.js
```diff
@@ -1286,7 +1296,57 @@ function init_crypto_form() {
...
+var scrollTimeout;
+function init_sticky_sidebar(){
+	if (!sidebar) {
+		return;
+	}
+	if (useJsScrollbar) {
+		init_simple_scrollbar();
+	}
+	$(window).scroll(function () {
+				if (scrollTimeout) {
```
Frenzie: A few tabs too many?

p/themes/BlueLagoon/BlueLagoon.css
```diff
@@ -564,6 +565,25 @@ a.btn {
...
+
+#sidebar {
+	scrollbar-color: rgba(0, 0, 0, 0.0) rgba(0, 0, 0, 0.0);
+}
+
+#sidebar:hover {
+	scrollbar-color: rgba(255, 255, 255, 0.3) rgba(0, 0, 0, 0.0);
+}
+
+/*Note: These have to be split, because Gecko ignores the line when combined, as of 4 December 2018*/
```
Frenzie: "These"? You mean Blink would do something with `rgba(255, , 0.3)` or something?

README.md
Alkarex: It is meant to cover e.g.:
* the case of the JavaScript scrollbar, which is only for browsers without native support
* the case of bcrypt.js, only for login using Web form (login is not an optional feature, but there are a few different options, only one of which requires bcrypt.js

p/themes/BlueLagoon/BlueLagoon.css
Alkarex: There is another selector `#sidebar:hover::-webkit-scrollbar-thumb {}` with the same rule `#sidebar:hover::-webkit-scrollbar-thumb` just below.
It would be natural to combine them but that does not work. The formulation of the comment could maybe be improved. Feel free to edit :-)

pattems:
Well crud, meant to merge that the other way. Sorry about the crappy commit messages that accidental merge cause.

README.md
Frenzie: It tickles my stylistic sensitivities, but it'll do. ;-)

p/scripts/main.js
pattems: @Alkarex explained that in a resolved conversation here: https://github.com/FreshRSS/FreshRSS/pull/2117#discussion_r238876288

Alkarex: It needs to be a number greater than 1 and smaller than the tree width. I am not sure a constant would make sense. I suggest to add a comment if needs be

Alkarex: Alternatively, could be changed by querying the tree width. I think I will do that

Alkarex: Addressed in https://github.com/FreshRSS/FreshRSS/pull/2117/commits/a3074b46873369081e37c629d096f7eb82fb935f

p/scripts/main.js
```diff
@@ -1286,7 +1313,57 @@ function init_crypto_form() {
 }
 //</crypto form (Web login)>
 
+var sidebar = document.getElementById('sidebar');
+var useJsScrollbar = true;
+try {
+	useJsScrollbar = sidebar && !CSS.supports('scrollbar-color: auto') &&
+		!(parseInt(getComputedStyle(sidebar, '::-webkit-scrollbar').width) < sidebar.offsetWidth);
```
Frenzie, line 1282: Excellent. Now that'll be fully comprehensible in five years without extensive use of git blame. :-)

Alkarex:
No worries @pattems We will anyway use a *Squash and merge* when all this is done, so it will be collapsed to one single commit in the master branch.

pattems:
Just to keep everybody in the loop, with the last commits, I wrote super verbose/duplicated code with a lot of different listeners, to be condensed and streamlined after everything's working. As it stands right now, the dropdowns in the nav and config work fine, but the dropdowns in the feed menu are a little iffy.

Alkarex:
@pattems Ok 👍 As there has been quite some activity, just to be sure those two things were not lost: Did you have a chance to try my patches from earlier today? I hope they work for you :-)
* https://github.com/FreshRSS/FreshRSS/pull/2117#issuecomment-445532385
* https://github.com/FreshRSS/FreshRSS/pull/2117#issuecomment-445538352

pattems:
@Alkarex just tested the old dropdown close method, and nope, that's still not going to work.

Alkarex:
Thanks for the continued effort, @pattems 👍 
I have only had time to test briefly today, and the following drop-down is not working (category when adding a new feed):
image
Would you be able to explain in a nutshell the main issue with the old dropdown? (again, it is not a problem to change, I am just trying to find the easiest way)

p/themes/Swage/swage.scss
```diff
@@ -604,14 +604,15 @@ form {
 }
 
 #sidebar {
```
Alkarex: Did you have a chance to consider the approach I introduced in `templace.css` with feature detection?

```css
@supports (scrollbar-width: thin) { }
```

It might not be needed when overriding only the `scrollbar-color` property, though. I cannot immediately think of potential problems.

pattems: I think that that's the way to go, I've just been focused on the JavaScript today.

Alkarex: Ok :-) If you are still making changes to the JavaScript, please paste your file in https://jshint.com (there are a couple of warnings) (there are also editor / CLI integrations if wanted).
P.S. One warning is due to my code - I will add an exception.
Edit: Done https://github.com/FreshRSS/FreshRSS/pull/2117/commits/07ef6bffdb3333f17aa8d84a00b30dc4d647ebe6

pattems:
@Alkarex the old dropdown close method works by overlaying a full-screen, invisible link that closes the dropdowns, meaning that you can't scroll the sidebar, except from the dropdown itself, which is absolutely positioned and has a higher z-index than the href.

I actually just had a thought about that, though; I'll either be back with a commit or letting you know my idea was a bust in a few.

pattems:
So my idea, which I've merged, is to add back in the overlay href, but have it not overlay the tree area. Add just enough JS to allow a second click on the toggle button in the tree to open/close a dropdown.

I've got to fully test the idea to make sure it's consistently working as expected, but the approach has enough merit I went ahead and merged it.

Todo:
- [ ] Fully vet this approach
- [ ] set .`dropdown-close a`'s left position to be equal to the width of the sidebar.
- [ ] set `dropdown-close a`'s left position to 0 if in a screen without a sidebar.

Alkarex:
I am trying without https://github.com/FreshRSS/FreshRSS/pull/2117/commits/94b904a2b2f70e3a44334323bc08c869cc2e40e8 and it seems to work fine. I am probably missing something :-)

pattems:
357dfbd throws an error that dropdown_handler isn't a function, because I removed it and forgot to remove the init call to it.

Alkarex:
Sorry, wrong commit. I meant that I am trying without https://github.com/FreshRSS/FreshRSS/pull/2117/commits/69478d974a3dcea40a9d8ada35c836ad4b3df6d4 , i.e. without *make overlay href scale to sidebar width*

pattems:
I set the `left` position in template to 300px; if you try it on a theme with a different width sidebar (Blue Lagoon, Screwdriver, or Swage I think are the 3 with non-standard width sidebars) there's dead space outside of the sidebar that you can't click to close a dropdown.

pattems:
I mean, the width could easily be fixed in css on the themes with non-300px width sidebars, but doing it in JS makes it a little more flexible.

app/views/helpers/index/normal/entry_bottom.phtml
```diff
@@ -42,7 +42,7 @@
 			<a class="dropdown-toggle" href="#dropdown-labels-<?php echo $this->entry->id();?>"><?php
 				echo _t('index.menu.tags');
 			?></a>
-			<ul class="dropdown-menu dropdown-menu-scrollable">
+			<ul class="dropdown-menu">
```
pattems: I noticed that I've removed dropdown-menu-scrollable here. I'm not finding this in the master branch; is it something that got added in dev that shouldn't be removed?

Alkarex:
I need more help :P Could you describe the problem that you are trying to solve with the `left` parameter? I am trying to reproduce the issue without success :P

app/views/helpers/index/normal/entry_bottom.phtml
Alkarex: To find such things, I use the blame function:

https://github.com/FreshRSS/FreshRSS/blame/dev/app/views/helpers/index/normal/entry_bottom.phtml#L45

It points to https://github.com/FreshRSS/FreshRSS/pull/2108

pattems:
I added a background color to `.dropdown-close a` to illustrate the point better than trying to explain it with words.

image

clicking outside of the sidebar, but not in the red-tinted region, won't close a dropdown menu.

pattems:
this could be pretty simply fixed by adding 
```
.dropdown-close a {
    left: 235;
}
````
for Blue Lagoon and Screwdriver, and `left:231` in Swage, but it increases the complexity of future themes.

**Comment by Alkarex** [comment]
I am trying with a global `left: 0` and it looks like we only lose the ability to scroll the tree alone while the dropdown is expanded, right? It seems quite acceptable to me (especially after I have added the margin-bottom). So if solving the problem fully requires some significant added complexity, it might not be worth it

**Comment by Alkarex** [comment]
P.S. Actually, the scrollbar seems to work almost fine with `left: 0` as far as I can see

**Comment by pattems** [comment]
I don't know why my editor has been so iffy on tabs/spaces recently, but I think I've gotten that all straightened out. I almost hesitate to say this again... but I think that, unless anything presents itself in testing as a problem, this might be ready.

File: p/scripts/main.js [hunk 25]
```diff
@@ -564,38 +552,39 @@ function init_column_categories() {
...
 	});
+
 	$('#aside_feed').on('click', '.tree-folder-items .feed .dropdown-toggle', function () {
+		var itemId = $(this).closest('.item').attr('id'),
+			templateId = itemId.substring(0, 2) === 't_' ? 'tag_config_template' : 'feed_config_template',
+			id = itemId.substr(2),
+			feed_web = $(this).data('fweb'),
+			template = $('#' + templateId)
+				.html().replace(/------/g, id).replace('http://example.net/', feed_web);
 		if ($(this).nextAll('.dropdown-menu').length === 0) {
```
Alkarex: `.next()` should probably be enough here

p/scripts/main.js
```diff
@@ -564,38 +552,39 @@ function init_column_categories() {
...
-			$('.tree-folder-items .dropdown-close a').click(function(){
-				$('.tree').removeClass('treepadding');
-				$(document.body).trigger("sticky_kit:recalc");
-			});
+		} else {
+			if ($(this).next('.dropdown-menu').css('display') === 'none') {
+				id = $(this).closest('.item').attr('id').substr(2);
+				$(this).attr('href', '#dropdown-' + id);
+			} else {
+				id = $(this).attr('href', "#close");
```
Alkarex: The `id = ` seems to be an error, isn't it?

p/scripts/main.js
```diff
@@ -536,23 +535,12 @@ function inject_script(name) {
...
-		$('#aside_feed .tree').stick_in_parent({parent:'#aside_feed'});
-	}
-}
-
 function init_column_categories() {
 	if (context.current_view !== 'normal') {
 		return;
 	}
+    
+	$('.dropdown-close a').css('left', ((sidebar.offsetWidth) + "px"));
```
Alkarex: If `left` stays at 0, could that simply the JS logic for the dropdown, or is it not advantageous?

At least some parentheses seem to be superfluous :-)
```suggestion
	$('.dropdown-close a').css('left', sidebar.offsetWidth + "px");
```

p/scripts/main.js
```diff
@@ -1286,7 +1276,59 @@ function init_crypto_form() {
...
+	useJsScrollbar = sidebar && !CSS.supports('scrollbar-color: auto') &&
+		!(parseInt(getComputedStyle(sidebar, '::-webkit-scrollbar').width) < sidebar.offsetWidth);
+	/*jshint +W018 */
+} catch (ex) {
+}
+if (useJsScrollbar) {
+	inject_script('simple-scrollbar.min.js');
+}
+
+function sticky_recalc() {
```
Alkarex, line 1290: Maybe not for this time, but I am wondering whether this function could be skipped in the case of native scrollbars.
The tree container is supposed to grow by itself when expanding categories, isn't it?

p/scripts/main.js
pattems: bah, I meant to pull this line out completely when it was decided to put left back to 0.

p/themes/Swage/swage.scss
```diff
@@ -603,6 +603,28 @@ form {
...
 
+@supports (scrollbar-width: thin) {
+	#sidebar {
+		scrollbar-color: rgba(255, 255, 255, 0.05) rgba(0, 0, 0, 0.0);
+	}
+	#sidebar:hover {
+		scrollbar-color: rgba(255, 255, 255, 0.3) rgba(0, 0, 0, 0.0);
+	}
+}
+/* Note: This selector has to be split from the -webkit- prefixed equivalent below,
```
Alkarex: The comments might not be needed anymore, since there is now a natural split due to the `@support {}` construction

p/scripts/main.js
pattems: If the `#nav_entries` element was reworked some, it could be. sticky_recalc takes into account the height of that element, which isn't in side .tree, when determining the height; if we left sticky_recalc out on native and left them where they were, the bottom of the scrollbar would extend behind that element. We could do something like the search bar, where one would be inside .tree and displayed on desktop, and another set would be displayed on mobile.

p/scripts/main.js
pattems: That appears to be the case.

p/themes/Swage/swage.scss
pattems: I think you're right!

pattems:
@Alkarex let me know when you're through looking through things, and I'll make one more commit to address the things you've caught

p/scripts/main.js
Alkarex: Ok. Let's come back to it at a later point

Alkarex:
@pattems I think I am done 😃

p/themes/base-theme/template.css
```diff
@@ -412,6 +494,9 @@ a.btn {
 	white-space: nowrap;
 	text-overflow: ellipsis;
 }
+.tree-folder:last-of-type {
+	margin-bottom: 15em;
```
Alkarex, line 499: That does not work when the last category is invisible 😢 such as when all articles are read.

Alkarex: Fixing

Alkarex: Second attempt in https://github.com/FreshRSS/FreshRSS/pull/2117/commits/10cb067e11cebf79a185f05dfbc4a59caa3f7204

Alkarex:
I have just fixed two small problems https://github.com/FreshRSS/FreshRSS/pull/2117/commits/10cb067e11cebf79a185f05dfbc4a59caa3f7204 , https://github.com/FreshRSS/FreshRSS/pull/2117/commits/6f2861fed4659e7ae6f0b5f70352abd961aa5919 and will give a final test round tomorrow morning :-)

pattems:
In testing, I realized that the simple-scrollbar colors hadn't been changed on dark themes to match the native scrollbars, so that's fixed. Everything else looks good to me!

Alkarex:
Thanks for the sustained work @pattems !
This was quite some effort, but I hope our users will enjoy the result. Let's continue the testing in the /dev branch.

Alkarex:
This morning, I can see that the scrollbar is not always visible in Firefox 64-65. Probably a concurrency problem. I am trying to look into it.

p/scripts/main.js
Alkarex: Attempt in https://github.com/FreshRSS/FreshRSS/pull/2181

Alkarex:
Seems be solved by https://github.com/FreshRSS/FreshRSS/pull/2181

PR #2069, opened 2018-10-22
Add Swage Theme

pattems:
Add a new Theme to FreshRSS.

This theme is a bit of a UI departure from other themes, especially on mobile:

original

capture

Issue #1620, opened 2017-08-22
Item alignement on login page


eliemichel:
On login page, the margin on the left on FreshRSS logo is much bigger than the one on the right of the "Login" link:

![freshrss2](https://user-images.githubusercontent.com/5802849/29555961-957920f0-8724-11e7-8dbb-112c935c9c0c.png)

I think the best would be not to use centering to place those element on this view, but just fixed margins. But I don't how much it would affect other things.

Alkarex:
Hello @eliemichel 
If you are able, you are welcome to send a suggestion of CSS patch

Alkarex:
Kindly fixed by @pattems with https://github.com/FreshRSS/FreshRSS/pull/2122

PR #2122, opened 2018-11-12
Login tweaks

pattems:
This is a fix for https://github.com/FreshRSS/FreshRSS/pull/2069#issuecomment-435093099, as well as a fix for #1620. I figured that it made sense to roll both of these fixes into one PR, but I can submit them separately if it's going to cause any headaches.


Alkarex:
Looks all good :-)

Issue #2106, opened 2018-11-02
figcaption tag issue


alfureu:
I noticed that the 
```
<figcaption> 
   ... 
</figcaption>
```

tags render strangely in FreshRSS. The desired behaviour would be to render centered under the Figure.

![figcaption](https://user-images.githubusercontent.com/10846169/47910735-09752b80-de94-11e8-94d4-818a0c68f9f7.png)

Frenzie:
Could you share the feed where this happens so the problem won't need to be recreated?

alfureu:
Sure, and thanks in advance! [http://444.hu/feed/](http://444.hu/feed/)

Frenzie:
Ah, it's not the figcaption but the icon that's inside the figcaption.

![screenshot_2018-11-02_11-48-16](https://user-images.githubusercontent.com/202757/47911254-42fa6680-de95-11e8-8f10-44b523f8545e.png)

alfureu:
You are right. Is it possible to remove, or scale the SVG properly?

Frenzie:
You could always add some custom CSS (or JS), but I'm not sure if there's much else you can do aside from prefiltering. Pinging @Alkarex

alfureu:
I think the easiest way would be to remove any icons/pictures from the figcaption tag. However, the most elegant would be to scale it properly to the feed rendering, as I can imagine there might be pictures which can be listed as Creative Commons (with the appropriate logo) and other cases.

Alkarex:
At a quick glance, we need to make the sanitizer more SVG aware. In particular, the URL is wrong (the `xlinl:href` should have been rewritten to an absolute URL like for the other assets) and currently generates an error 404.

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/2135 by removing SVG images for now. Proper SVG support would require custom sanitizing and URL rewriting of xlink:href, and is left for future work.
In this case, the SVG image was solely used for tracking anyway.
Tests welcome.

alfureu:
Thanks @Alkarex, I think you made the right decision. I appreciate the quick resolution of this, I believe we can close this ticket.

Frenzie:
You can close it yourself as well I believe? :-) Anyway, closing.

alfureu:
Oh, was not aware of this. Thanks for pointing this out @Frenzie. Next time will certainly do!

PR #2135, opened 2018-11-17
Strip embedded SVG images for now

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2106
Proper SVG support would require custom sanitizing and URL rewriting of
xlink:href, and is left for future work

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -226,6 +226,7 @@ function customSimplePie($attributes = array()) {
 		'font', 'form', 'frame', 'frameset', 'html',
 		'link', 'input', 'marquee', 'meta', 'noscript',
 		'object', 'param', 'plaintext', 'script', 'style',
+		'svg',	//TODO: Support SVG after sanitizing and URL rewriting of xlink:href
 	));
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
A pity though.

Frenzie:
I'm not really sure how else I'd use SVG?

https://fransdejonge.com/category/media/drawing

That being said, normally I don't put the SVG in the HTML like I seem to have done in this one:
https://fransdejonge.com/2015/09/just-a-star/

Also slightly different:
https://openclipart.org/rss/new.xml

Generally speaking the more sensible thing to do is a `picture` element with an `img` png/jpg and a `source` SVG.

But surely I can't uniquely be the only person who's used plain SVG in a "legit" manner?

Alkarex:
Thanks for the example :-) 
https://fransdejonge.com/category/media/drawing/feed/
Could you please open another ticket to remember to re-add SVG?

Frenzie:
Although note that you'll need to use the "Articles CSS path on original website" feature.

PR #2140, opened 2018-11-18
[fix] Finishing touches for next/previous without focus

Frenzie:
Cf. #1767.

### DIFF
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -162,6 +162,7 @@ return array(
 		'mark_read' => 'Mark as read',
 		'navigation' => 'Navigation',
 		'navigation_help' => 'With the "Shift" modifier, navigation shortcuts apply on feeds.<br/>With the "Alt" modifier, navigation shortcuts apply on categories.',
+		'navigation_no_mod_help' => 'The following navigation shortcuts don\'t support modifiers.',
 		'next_article' => 'Open the next article',
 		'other_action' => 'Other actions',
 		'previous_article' => 'Open the previous article',
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -162,6 +162,7 @@ return array(
 		'mark_read' => 'Mark as read',
 		'navigation' => 'Navigation',
 		'navigation_help' => 'With the "Shift" modifier, navigation shortcuts apply on feeds.<br/>With the "Alt" modifier, navigation shortcuts apply on categories.',
+		'navigation_no_mod_help' => 'The following navigation shortcuts do not support modifiers.',	//TODO - Translation
 		'navigation_no_mod_help' => 'The following navigation shortcuts do not support modifiers.',
 		'next_article' => 'Open the next article',
 		'other_action' => 'Other actions',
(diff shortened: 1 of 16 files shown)


p/scripts/main.js
```diff
@@ -252,7 +252,7 @@ function toggleContent(new_active, old_active, skipping = false) {
 	}
 
 	if (old_active[0] !== new_active[0]) {
-		if (isCollapsed && !skipping) { // BUG?: isCollapsed can only ever be true
```
Alkarex, line 255: The function declaration a bit higher up contains a default parameter, which is ES6.
```javascript
function toggleContent(new_active, old_active, skipping = false) { }
```
This might not be compatible with all the browsers we support so far (e.g. IE11?)
I suggest removing the default parameter, and being explicit in all the calls

p/scripts/main.js
```diff
@@ -252,7 +252,7 @@ function toggleContent(new_active, old_active, skipping = false) {
 	}
 
 	if (old_active[0] !== new_active[0]) {
-		if (isCollapsed && !skipping) { // BUG?: isCollapsed can only ever be true 
+		if (isCollapsed && !skipping) { // BUG?: isCollapsed can only ever be true
```
Alkarex, line 255: The function declaration a bit higher up contains a default parameter, which is ES6.
```javascript
function toggleContent(new_active, old_active, skipping = false) { }
```
This might not be compatible with all the browsers we support so far (e.g. IE11?)
I suggest removing the default parameter, and being explicit in all the calls

app/i18n/en/conf.php
```diff
@@ -162,6 +162,7 @@
 		'mark_read' => 'Mark as read',
 		'navigation' => 'Navigation',
 		'navigation_help' => 'With the "Shift" modifier, navigation shortcuts apply on feeds.<br/>With the "Alt" modifier, navigation shortcuts apply on categories.',
+		'navigation_no_mod_help' => 'The following navigation shortcuts do not support modifiers.',
```
Alkarex, line 165: Please add the TODO's in the other languages as well

Frenzie: Is there an automated way to do that?

Alkarex: `sed` ? :-)

Frenzie: lol, fair enough.

Frenzie: Yeah, that worked well enough. Good thing you can reset quickly with Git to get the escaping just right though. ;-)

```
sed -i "/navigation_help/a\\\t\t'navigation_no_mod_help' => 'The following navigation shortcuts do not support modifiers.',\t//TODO - Translation" */conf.php
```

Alkarex: Perfect :-)

Alkarex:
@Frenzie Now that you have become `sed` master of i18n files, could you please add the TODOs for the new strings? 😛  `conf.shortcut.skip_next_article` and `conf.shortcut.skip_previous_article` are visibly missing

Frenzie:
Sure, but you might have to remind me in a few days. I'm heading off to bed soon.

Issue #613, opened 2014-09-11
Implement a web-cron


marienfressinaud:
In order to schedule events (e.g. actualize feeds, check for updates), it would be useful to have a web-cron instead only a standard Cron. This must be optional and determined by administrator.

More information on http://www.smashingmagazine.com/2013/10/16/schedule-events-using-wordpress-cron/

See #585

qwertygc:
Bonjour,
Oú en est ce point ?

Frenzie:
@qwertygc The most important part of this already exists: `https://votre.serveur.net/FreshRSS/p/i/?c=feed&a=actualize`

You'll need to call that from someplace else of course, but that shouldn't be much of a problem even when real cron isn't available (e.g., on shared hosts).

planetahuevo:
Trying to use that URL returns an error 403 on my install.
I am not able to make the CRON work for me using PHP either so this was the best solution, but it is not working.
Any ideas on implementing an "update link" for using wget on my crontab? Thanks

Frenzie:
You might need something like a token or HTTP authentication.

https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html#global-view

planetahuevo:
I cannot see that option on my install 1.7.0.
And this should be easier to setup. Without a proper cronjob, the whole system does not work...

Frenzie:
It's in the settings (top right). You might want to enable anonymous refresh for simplicity. It shouldn't really matter that any random person can refresh your feeds because the refresh rate should be limited from within FreshRSS regardless. (Unless that's different for this refresh method, @Alkarex?)

![screenshot_2017-09-29_21-23-58](https://user-images.githubusercontent.com/202757/31032527-9de737b2-a55c-11e7-85ae-cda665c4fa13.png)

planetahuevo:
I couldn´t change that on my options (not sure why) but it worked editing the config file manually.
Also it is important to check the default time to refresh, because you can think the cron is not working, and be working, like mine... 👍 
Thanks!

Alkarex:
@Frenzie That is correct :-)

Alkarex:
We should add to the documentation how to use the existing Web cron possibility, also to target other users than the default one:

https://freshrss.example.net/i/?c=feed&a=actualize&user=someone&token=ABCDEFabdcef01234

Any volunteer? :-)

@opitux This is relevant for your case

Frenzie:
But it's already in the docs?

[...]

PR #2142, opened 2018-11-18
[docs] Add note about specific user for webcron

Frenzie:
Cf. https://github.com/FreshRSS/FreshRSS/issues/613#issuecomment-353585547


docs/fr/users/03_Main_view.md
```diff
@@ -35,21 +35,20 @@ Ci-dessous vous trouverez un exemple permettant la mise à jour des articles tou
...
 ##### Aucune authentification
 
 C’est le cas le plus simple, puisque votre instance est publique, vous n’avez rien de particulier à préciser :
 
 ```cron
-0 * * * * curl 'https://votre.serveur.net/FreshRSS/p/i/?c=feed&a=actualize'
+0 * * * * curl 'https://freshrss.example.net/i/?c=feed&a=actualize'
 ```
 
-You can also target a different user by adding their username to the query string, with `&user=insert-username`:
```
Frenzie: Oops. :-)

Issue #2130, opened 2018-11-16
Sharing by print gives a blank page with only the text "undefined" on it


sbrl:
Hey! I've recently set up _FreshRSS_ as an alternative to _Miniflux 1_ (v2 uses PostgreSQL exclusively, and I don't want a PostgreSQL database server!), and been generally happy with with actually.

I've encountered a bug though in the sharing system. I 'print' things often to a virtual PDF printer I've got setup. If I go to share -> print in FreshRSS, I get the text "undefined" on a white page. I'm including an image and a copy of the resulting PDF at the bottom of this report. 

![selection_092](https://user-images.githubusercontent.com/9929737/48652868-ab028e00-e9f9-11e8-94b9-721705b50d47.png)

Here's the PDF when I print:
[mozilla.pdf](https://github.com/FreshRSS/FreshRSS/files/2591264/mozilla.pdf)

Alkarex:
Hello @sbrl and welcome :-)
I cannot reproduce the problem.
Tested with Firefox on Ubuntu 18.04, Firefox + Chrome + Edge on Windows.
Do you have any information in your browser console?
Could you please try with another browser?

sbrl:
Ah, yeah I do get something in the console:

```
Content Security Policy: Directive ‘child-src’ has been deprecated. Please use directive ‘worker-src’ to control workers, or directive ‘frame-src’ to control frames respectively.
FreshRSS waiting for Sticky-kit… main.js:521:4
FreshRSS init done. main.js:1550:3
Content Security Policy: The page's settings blocked the loading of a resource at inline ("default-src"). i:1:1
This site appears to use a scroll-linked positioning effect. This may not work well with asynchronous panning; see https://developer.mozilla.org/docs/Mozilla/Performance/ScrollLinkedEffects for further details and to join the discussion on related tools and features! i
Content Security Policy: The page's settings blocked the loading of a resource at inline ("default-src").
i:1:1
Content Security Policy: The page's settings blocked the loading of a resource at inline ("default-src").
```

sbrl:
Tried Chrome, and can still reproduce - though nothing of interest in the console. I'm using the dark theme, if that helps.

Alkarex:
Would you mind making a test account for me in your FreshRSS instance? So I can see if it is something due to your instance, or due to your client. If not, I will make you a test instance.

Alkarex:
I am guessing from your screenshot that you are using Ubuntu, aren't you? Which version?

[...]

PR #2148, opened 2018-11-24
Fix print for unfolded articles

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2130
`.current` might not exist in views for which all articles are already
expanded

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1301,13 +1301,13 @@ function init_confirm_action() {
 }
 
 function init_print_action() {
-	$('.item.share > a[href="#"]').click(function () {
+	$('.item.share > a[href="#"]').click(function (e) {
 		var content = "<html><head><style>" +
 			"body { font-family: Serif; text-align: justify; }" +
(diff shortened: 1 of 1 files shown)


sbrl (approved):
I can confirm that this fixes my issue. Thanks so much, @Alkarex :smiley_cat: 

Is it possible to apply a stylesheet when printing at all?

Frenzie:
@sbrl 

> Is it possible to apply a stylesheet when printing at all?

You mean like `@media print {}`?

sbrl:
Something like that! Any CSS will do. It looks like when going to share -> print all the CSS is removed in the new tab that is opened.

I ask because I have a strong dislike for _Times New Roman_ - the default font :P

sbrl:
I guess! But that would be global.

Alkarex:
@sbrl Thanks for the feedback, and please open another issue for CSS print styling

sbrl:
Sure! I've opened #2149 about that.

Issue #2121, opened 2018-11-12
Advisory - FreshRSS - Reflected and Stored Cross-site scripting vulnerabilities 


DanielBishtawi:
Hello,

While testing the Netsparker web application security scanner we identified Reflected and Stored Cross-site scripting vulnerabilities in FreshRSS. Can you please advise whom shall we contact to disclose the vulnerability details so it can be fixed?

Please email me at daniel@netsparker.com for the technical details.

Looking forward to hearing from you.

Regards,

Daniel Bishtawi
Marketing Administrator | Netsparker Web Application Security Scanner

Frenzie:
Pinging @Alkarex

Alkarex:
Thanks @DanielBishtawi 
I have sent an e-mail to you.

PR #2157, opened 2018-11-28
Minz: Fix HTML injections

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2121
Reported by https://www.netsparker.com


Frenzie (approved):
Surprised the first existed. :-)

PR #2161, opened 2018-12-04
[BlueLagoon] Fix header's height

nicofrand:
Hi,

I'll try to sum up the current issue: the `#global` height is calculated with `calc(100% - 60px)`. 60px is supposed to be the header's height. However, even though there is a `height: 55px` set on `.header` it would not restrict the height because of `display: table`. And the height actually equals the logo's height (60px) + (0.5em * 2) of margin on the `<h1>`, which makes it around 80.5px on my screen.

This MR:

* Removes the useless `height: 55px` on `.header`
* Reduces the logo size to 40px
* Fixes the margins on `<h1>` to 10px (x 2 : top and bottom)
* Sets the `#global` height to `calc(100% - 61px)` because there also is a 1px border bottom on the header

screenshot_2018-12-04 vos flux rss freshrss

### DIFF
--- a/p/themes/BlueLagoon/icons/icon.svg
+++ b/p/themes/BlueLagoon/icons/icon.svg
@@ -1,5 +1,5 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 59.4 59.7">
 <defs>
 <linearGradient id="linearGradient3796">
 <stop stop-color="#0090ff" offset="0"/>
--- a/p/themes/BlueLagoon/BlueLagoon.css
+++ b/p/themes/BlueLagoon/BlueLagoon.css
@@ -137,8 +137,6 @@ button.as-link[disabled] {
 .stick .btn:first-child,.stick input:first-child {
 	border-radius: 6px 0 0 6px;
 }
-.stick .btn-important:first-child {
-}
 .stick .btn:last-child, .stick input:last-child {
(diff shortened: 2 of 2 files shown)


Alkarex:
Thanks for this nice contribution 👍 I will try to test tonight.
Regarding the logo, I would better like to change its dimensions in CSS instead of changing the SVG file, if possible.

nicofrand:
For the logo the dimensions were set in the SVG. I tried a few things but did not manage to do it by CSS only.

p/themes/BlueLagoon/icons/icon.svg
```diff
@@ -1,34 +1 @@
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
pattems, line 2: The reason that resizing isn't working is because you don't have a viewBox set. Instead of replacing the entire SVG and setting the exact size, you can add a viewBox to this line and then it'll scale properly with CSS. The bit you need to add is ` viewBox="0 0 59.4 59.7"`, so your resulting line will look like: 
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 59.4 59.7">>>
```

nicofrand: Thanks ! I just force-pushed on my branch sorry, I lost your commit :s

nicofrand: Nevermind, I thought you added a commit to this MR but you only referenced it -_-.
Thanks again for the tip

Alkarex:
@nicofrand Could you please add yourself in https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md ?
We start having quite a few *Nicolas* 😜

PR #2166, opened 2018-12-06
Delete unnecessary File in BlueLagoon

pattems:
I just realized (despite the fact that I've been editing it for the last month keeping it updated) that BlueLagoon has a copy of template.css in it's theme directory that doesn't do anything, and should probably be removed.


Alkarex:
Good catch 👍

Issue #2169, opened 2018-12-07
Suggestion for improvement of Reader Mode


pattems:
Perhaps Reader mode should include the sidebar, collapsed by default like in the mobile view, in order to switch feeds without leaving reader mode.

This, with a couple trivial modifications, would also enable favicon & title updating in #2168.

Alkarex:
This sounds like a great idea 👍

Alkarex:
Implemented by @pattems in https://github.com/FreshRSS/FreshRSS/pull/2180

Issue #1407, opened 2017-01-14
Not notifying in reading view


welbert:
If I use 'normal view', I received the notification about news feeds but with the 'reading view' I do not get notification of new feeds.

Browser: Mozilla Firefox 50.1.0

pattems:
@Alkarex I'm trying to track this one down. I've got a dummy test feed that issues a new article every minute, but that 20 minute cooldown between being able to fetch feeds makes it slow-going. Is there any way around this limit for debugging purposes?

Alkarex:
The easiest to debug notifications is to have two windows of FreshRSS, and mark some articles as unread in one of them :-)
And you can even call `refreshUnreads();` manually from the JavaScript console.

pattems:
okay, great! That's getting me somewhere. Thinking out loud: the new-article div was missing from reader view, but adding it isn't enough to trigger it. It triggers fine if I manually toggle the display property in css, so I'm thinking that it not triggering might be related to #545

Alkarex:
Fixed by @pattems in https://github.com/FreshRSS/FreshRSS/pull/2188 and merged in /dev branch. Tests welcome before release of FreshRSS 1.13.0 in a couple of days.

PR #2170, opened 2018-12-07
Bugfix #1407

pattems:
Take two! Notification of new articles works in reader mode for a specific feed, or on all. per #2169, I have no way to get categories in reader mode, so there's no notification there.

Feed detection in reader mode is, by necessity without the benefit of the sidebar, a little kludgey, but it works and (in my testing) doesn't seem to impact performance.


app/views/index/normal.phtml
```diff
@@ -19,8 +19,8 @@ if (!empty($this->entries)) {
 	$today = @strtotime('today');
 ?>
 
-<div id="stream" class="normal<?php echo $hidePosts ? ' hide_posts' : ''; ?>"><?php
-	?><div id="new-article">
+<div id="stream" class="normal<?php echo $hidePosts ? ' hide_posts' : ''; ?>">
+	<div id="new-article">
```
Alkarex, line 23: This was actually on purpose, to avoid sending some blanks to the client.
It is mostly irrelevant here (probably better without, indeed), but can be advantageous in particular in some loops.

p/themes/Swage/swage.scss
```diff
@@ -746,6 +746,14 @@ form {
...
 }
 
+.reader #new-article {
+	position: fixed;
+    top: 35px;
+    width: 100%;
+    padding: 10px 0;
+	bottom: unset;
+}
+
```
Alkarex, line 782: Mix of tabs and spaces ;-)

Alkarex: Was that the only theme requiring some adjustments?

pattems: It is, because the other themes don't muck with the positioning of that div the way I do with Swage.

pattems: Also, re:mixing tabs and spaces: woops. This is what happens when I switch editors mid-stream.

app/views/index/normal.phtml
pattems: Huh. I've never heard of anything like that before. I want to run by what I'm understanding from your explanation to make sure that I've got it right: When in a php loop, an empty tag like that would let there be a newline in the php code that wouldn't exist in the html that's sent to to client?

Frenzie: Except for that restrictive "in a loop" clause, that's correct.

p/themes/Swage/swage.scss
Alkarex: By the way, we have a https://github.com/FreshRSS/FreshRSS/blob/dev/.editorconfig declaration
See https://editorconfig.org

pattems:
A lot of this can be handled more elegantly if #2180 is merged first, since we'll have the sidebar to do less hacky checks for things.

pattems:
I'm going to go ahead and close this PR and open a new one based on current dev, since a lot of the changes I made to detect reader aren't needed now that we've got a sidebar in reader mode, and that way I don't have to worry about cleanly undoing the changes that aren't needed.

I'm actually questioning the necessity of a separate reader.phtml file at this point. I want to say that normal and reader have feature parity, and swapping out the body class to apply the correct css might be enough to take care of things.

Alkarex:
Indeed @pattems , let's try to simplify the code as you suggest, but first for version 1.14

Issue #2115, opened 2018-11-04
Can't create new category while adding feed


Alwaysin:
When adding a feed, you can quickly create a category by using the dropdown menu before the "+" and typing the name of the category in the box "New category". When doing so, the category is not created and the feed is added to the default one.

Alkarex:
Hum, the UI could be better, but you need to select *New category* in the list.
We should make a little piece of JavaScript, which only shows the new field name when *New category* is selected.

Alwaysin:
Ah, sure. Misleading, but minor indeed!

PR #2176, opened 2018-12-11
Fix for #2115

pattems:
`aria-hidden="true"` can be overridden by more specific css display rules. In the case of #2115, it was `.dropdown-menu .input { display: block }`.

Moving this css rule down the document doesn't help because it's not being overriden by rules of the same type, but rules with finer specificity. 

As much as I don't like using `!important`, it's the best course of action here, because otherwise we'd have to write new css for any element that might ever be in an `aria-hidden` element, to unset it's display property.

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -139,6 +139,13 @@ td.numeric {
 	text-align: center;
 }
 
+/*=== COMPONENTS */
+/*===============*/
+
+[aria-hidden="true"] {
+	display: none !important;
+}
+
 /*=== Forms */
 .form-group::after {
 	content: "";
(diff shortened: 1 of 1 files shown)


Frenzie:
Could you maybe explain the specific issue (#2115) in a bit more detail?

pattems:
In stable, while in subscriptions management, the dropdown for adding a new feed shows the option to create a category while an existing category is selected, like so:
image

This new category dropdown should only appear if you've selected new category from the category dropdown, and does nothing unless that option is selected. 
image

You can see the desired behavior once you switch categories in that dropdown. If you're not on new category, the menu should look like this: 
image


More generally (though I'm not sure if this happens anywhere else), this PR fixes anything that should be hidden by `aria-hidden="true"` not being hidden if an element contained inside an aria-hidden element has a display property of its own set.

Frenzie:
> This new category dropdown should only appear if you've selected new category from the category dropdown, and does nothing unless that option is selected.

For the value of "should" of what the code seems to intend, yes. I'm not quite convinced that's how it "should" be. :-)

Alkarex (approved):

PR #2177, opened 2018-12-11
Bugfix #1781

### LINKED ISSUE #1781: Movim sharing broken
Seems like with recent versions of Movim the sharing feature isn't working.

All I get is a giant sharing symbol and some garbled URL on the redirected page in Movim (or a login screen if not logged in before clicking the share link).

I guess something changed on the Movim side of things?

pattems:
Fixes #1781. 

Movim expects an encoded url in it's external share feature, not a raw encoded url.  I'm not sure when this change was made on Movim's end, but in any case, I've tested this as working.

### DIFF
--- a/app/shares.php
+++ b/app/shares.php
@@ -64,7 +64,7 @@ return array(
 	),
 	'movim' => array(
 		'url' => '~URL~/?share/~LINK~',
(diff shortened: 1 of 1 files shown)


Frenzie:
Cross-ref to #1000.

PR #2181, opened 2018-12-15
JavaScript-free scrollbar

Alkarex:
Make scrollbar work also without JavaScript
https://github.com/FreshRSS/FreshRSS/pull/2117#discussion_r240804225
Also implies a better support of resize scenarios.
Seems to fix at the same time the occasional glitches
https://github.com/FreshRSS/FreshRSS/pull/2117#issuecomment-447556038

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -553,7 +553,7 @@ function init_column_categories() {
 		if (sidebar) {
 			$(this).parent().next(".tree-folder-items").slideToggle(300, function () {
 				if (!useJsScrollbar && sidebar.scrollHeight > sidebar.clientHeight) {
-					//Workaround for Gecko bug in Firefox 64-66. TODO: Try newer versions
+					//Workaround for Gecko bug in Firefox 64-66(+?)
 					sidebar.style['overflow-y'] = 'scroll';
 					setTimeout(function () { sidebar.style['overflow-y'] = ''; }, 0);
 				}
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1285,7 +1285,7 @@ var useJsScrollbar = true;
 try {
 	/*jshint -W018 */
 	useJsScrollbar = sidebar && !CSS.supports('scrollbar-color: auto') &&
-		!(parseInt(getComputedStyle(sidebar, '::-webkit-scrollbar').width) < sidebar.offsetWidth);
+		!(parseInt(getComputedStyle(sidebar, '::-webkit-scrollbar').width) < sidebar.scrollWidth);
 	/*jshint +W018 */
 } catch (ex) {
 }
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1312,12 +1312,10 @@ function init_simple_scrollbar() {
 
 var scrollTimeout;
 function init_sticky_sidebar(){
-	if (!sidebar) {
+	if (!useJsScrollbar) {
 		return;
 	}
-	if (useJsScrollbar) {
-		init_simple_scrollbar();
-	}
+	init_simple_scrollbar();
 	$(window).scroll(function () {
 		if (scrollTimeout) {
 			clearTimeout(scrollTimeout);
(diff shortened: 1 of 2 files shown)


Alkarex:
@pattems @Frenzie Additional testing Welcome

Alkarex:
Illustration of the Gecko bug: https://alexandre.alapetite.fr/divers/vrac/20181215-Gecko-overflow-scrollbar-thin.html

Alkarex:
Bug submitted https://bugzilla.mozilla.org/show_bug.cgi?id=1514498

Alkarex:
@pattems Independently of this pull request, sticky does not seem to work in Edge anymore. Ideas welcome.

Alkarex:
@pattems I cannot remember for sure: has Edge sticky been working at some point of https://github.com/FreshRSS/FreshRSS/pull/2117 ? If yes, can you find again what the approach was? :-)

pattems:
Sticky was only briefly working in #2117 with the used of [stickyfill](https://github.com/wilddeer/stickyfill). It was never fully integrated, though.

Alkarex:
Ah, that is why I could not get it to work again 😛 
It should be possible to make it work natively (or with minor help) in Edge though. I have not fully understood the problem, but it looks like the parent block is moving up or something like that, maybe due to the table layout.
Let's come back to it later, if someone finds a good fix (not involving a big library).

Please let me know if you get a chance to test this PR.

I am now wondering whether it makes sense to keep the simple-scrollbar fallback, since the current Firefox (64+) and Chrome have a native support. Previous versions of Firefox, as well as IE11 and Edge have a functional native scrollbar (only not styled, but quite nice by default with e.g. Firefox in Ubuntu). What do you think?

Alkarex:
Nowadays, the design would indeed much likely better be done with more
modern layouts such as a combination of Grid + Flexbox (which might fix the
Edge Stiky bug at the same time).
I think we could keep the current themes, and introduce a new template for
future themes.

p/scripts/main.js
```diff
@@ -1312,12 +1318,10 @@ function init_simple_scrollbar() {
 
 var scrollTimeout;
 function init_sticky_sidebar(){
```
pattems, line 1320: I'm away from my dev environment until Monday, so I'm not in a position to test this right now, but I think we still need to be doing the scroll function with native scrollbars, to ensure that the bottom of the scroll doesn't extend under the nav_entries, if they're visible. It's a minor thing, since with the padding at the end of the tree, no content should be obscured by the buttons; the bottom of the scrollbar might just slide under them.

pattems:
Also, on removing the js scrollbar. the ESR release of Firefox 60 is currently slated to hit EOL on [October 23rd of 2019](https://www.mozilla.org/en-US/firefox/organizations/). It had been my intent to propose a PR to remove Simple-scrollbar at that point, but as you pointed out, the native scrollbar is perfectly functional, and in some environments even is quite nice. I'd have no problems whatsoever going ahead and pulling it; if I'd known when I started #2117 that native styling on Firefox was just around the corner, I probably wouldn't have included it in the first place.

p/scripts/main.js
```diff
@@ -1312,12 +1318,10 @@ function init_simple_scrollbar() {
...
-	if (!sidebar) {
+	if (!useJsScrollbar) {
```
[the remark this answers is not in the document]
Alkarex: That should be good: see line 1288, `useJsScrollbar` cannot be true if `sidebar` does not exists

p/scripts/main.js
Alkarex: Indeed, the scrollbar slides under the buttons, and that seems to work well. The length of the scrollbar is slightly wrong in this case, but that is a minor issue which can likely be fixed purely in CSS if wanted (with a `calc()`) and later even better if/when moving to e.g. Grid + Flexbox.

Alkarex:
Yes, I have also been surprised about the pace at which Firefox has introduced this draft of standard. Given than FreshRSS users also take some time before updating to a new FreshRSS versions, I think the remaining needs are too few and too minor (if I have not overlooked any) to justify a big fallback.

Alkarex:
I merge that in /dev to make another PR tentatively removing the JS fallback, but no problem if some reverts are needed. I would like to start testing /dev to make a new release on this side of Christmas.

PR #2184, opened 2018-12-16
Update naming to WebSub

Alkarex:
Instead of PubSubHubbub / PuSH


pattems:
Is it your intent to rename the variables in code, or just renaming the public-facing stuff? random example of what I'm talking about: https://github.com/FreshRSS/FreshRSS/blob/e04804d0f67dd43fd3f072b9a127768ee7b7b56c/app/Controllers/feedController.php#L270

Alkarex:
@pattems Just the public-facing part for now. I would like to avoid opening new potential breaking changes before the 1.13.0 release. We can do that in 1.14 if needed

PR #2188, opened 2018-12-17
Add New Article banner to reader mode

### LINKED ISSUE #1407: Not notifying in reading view

pattems:
Fixes #1407

Display the new available articles banner while in reader mode. With the sidebar now present on reader mode, all that was needed to fix this is adding the div in reader mode. Sorry I didn't have this done before 1.13.0 went to testing!

### DIFF
--- a/app/views/index/reader.phtml
+++ b/app/views/index/reader.phtml
@@ -12,8 +12,11 @@ if (!empty($this->entries)) {
 	$content_width = FreshRSS_Context::$user_conf->content_width;
 ?>
 
-<div id="stream" class="reader"><?php
-	foreach ($this->entries as $item) {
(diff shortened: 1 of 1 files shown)


Alkarex:
Thanks @pattems , and no worries. Looks straightforward and fixing a bug, so let's try to include it in 1.13.0. There will be a couple more days of testing.

For reference, same code in `normal.phtml`:

https://github.com/FreshRSS/FreshRSS/blob/0e69b279722342b4284017789f639665012dd888/app/views/index/normal.phtml#L22-L26

PR #2199, opened 2018-12-25
Batch scroll as read + reducing jQuery

Alkarex:
The overall goal is to improve performance, in particular on mobile and server-side.

In particular, this patch improves scroll performance when using the auto mark-as-read feature:
Mark-as-read requests are queued and sent max once per second. https://github.com/FreshRSS/FreshRSS/pull/2199/commits/cd9a9a93099b80fb3f8c2ba09266aa31955cbdd6

The scroll event listener is also rewritten to be much lighter https://github.com/FreshRSS/FreshRSS/pull/2199/commits/caa893e14b46bc9b52f251ca93eada0c57634302

As the same time, in distinct commits, I have started some refactoring https://github.com/FreshRSS/FreshRSS/pull/2199/commits/eefeb2341915d677e446cd733fc58b45efb7f5d9 :
* to start using ES6 syntax (limited to an IE11-subset)
* to ensure that jQuery variable names start with a `$` (to distinguish them),
* and to avoid jQuery in easy cases (we might be able to soon avoid jQuery fully).

### DIFF
--- a/app/Controllers/entryController.php
+++ b/app/Controllers/entryController.php
@@ -97,14 +97,15 @@ class FreshRSS_entry_Controller extends Minz_ActionController {
 				}
 			}
 		} else {
-			$entryDAO->markRead($id, $is_read);
-
+			$ids = is_array($id) ? $id : array($id);
+			$entryDAO->markRead($ids, $is_read);
 			$tagDAO = FreshRSS_Factory::createTagDao();
-			foreach ($tagDAO->getTagsForEntry($id) as $tag) {
-				if (!empty($tag['checked'])) {
-					$this->view->tags[] = $tag['id'];
-				}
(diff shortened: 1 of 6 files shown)


Alkarex:
Resources when searching for standard JavaScript alternatives to jQuery:
* https://github.com/Haeresis/vanilla-js-dom
* http://youmightnotneedjquery.com

Alkarex:
A bit of motivation :-) https://timkadlec.com/remembers/2019-01-09-the-ethics-of-performance/

Frenzie:
> Less often considered is the sheer waste that poor performance results in.

Funny, I never thought about performance as exclusion, but I've always thought of bad performance as a waste of my battery (and thus lifespan).

PR #2208, opened 2019-01-05
Use openlog before syslog

Alkarex:
In order to have a copy on stderr when syslog is not available.


Frenzie (approved):
I didn't even realize that was an issue.

Issue #2196, opened 2018-12-22
Improve export of labels


Alkarex:
Follow up of https://github.com/FreshRSS/FreshRSS/issues/2079

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/2217

PR #2217, opened 2019-01-15
Export labels

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2196


app/Models/EntryDAO.php
```diff
@@ -855,6 +858,9 @@ private function sqlListWhere($type = 'a', $id = '', $state = FreshRSS_Entry::ST
 		case 'T':	//Any tag
 			$where .= '1=1 ';
 			break;
+		case 'ST':	//Starred or taggued
```
Frenzie: super nitpick, but typo in comment ;-)

Frenzie (approved):

app/Models/EntryDAO.php
Alkarex: Well spotted 🤓

Issue #2153, opened 2018-11-28
Problem about item update date and time


maobaoqi:
Hi, I find a problem about freshrss item date and time, which is different from real time.
I update some items serverl times and the issue also exists.
![snipaste_20181128_170201](https://user-images.githubusercontent.com/3353791/49140878-51229200-f330-11e8-8612-e0939bcb97b9.png)
For example: now real time is 2018.11.28, but freshrss item show 2018.11.27.

Alkarex:
Hello,
The date/time on the column on the right-hand side is the date declared by the feed, not the time at which the article was received by FreshRSS. I hope this answers your question?
I am actually considering making an option to switch between the declared date and the received date.

maobaoqi:
So that's it! loolking forward to the further function, thanks!

Alkarex:
@maobaoqi Would you personally better like to have the received date in the column on the right-hand side, instead of the publication date?

maobaoqi:
I would like to know the page publication date.

xobs:
I'm experiencing similar issues.

For example, this entry:

```xml
 <item> <title>Top stories from The Straits Times on Wednesday, Jan 16</title>
 <link>http://www.straitstimes.com/singapore/top-stories-from-the-straits-times-on-wednesday-jan-16</link>
 <description>January 16, 2019 7:15 AM&lt;br /&gt;&lt;br /&gt;Good morning! Here are our top stories to kick-start your Wednesday, Jan 16.

 </description>
 <guid isPermaLink="false">http://www.straitstimes.com/singapore/top-stories-from-the-straits-times-on-wednesday-jan-16</guid>
 <pubDate>Wed, 16 Jan 2019 07:15:00 +0800</pubDate>
 <source url="https://www.straitstimes.com/news/singapore/rss.xml/20190116">The Straits Times</source>
</item>
```

However, the date is showing up incorrectly on the right-hand side:

![image](https://user-images.githubusercontent.com/238325/51221069-cf83d380-1972-11e9-97fa-a3b2f8b194fc.png)

The server's timezone is set correctly:

```sh
/var/www/FreshRSS # date
Wed Jan 16 09:41:02 +08 2019
/var/www/FreshRSS #
```

And this matches my system's timezone as well.

Alkarex:
Hello @xobs ,
Could you please try to run the following command?

```sh
php -r 'echo date_default_timezone_get(), "\n", date("c"), "\n";'
```

[...]

Issue #853, opened 2015-05-29
Date/Time Formatting Enhancements


jgeboski:
Two simple suggestions for the date/time formatting:
- Add support for 12-hour time
- Add support for local time

Alkarex:
@jgeboski When you say local time, do you mean an option to specify the user's timezone that is different than the timezone on the server hosting this user? Currently (if there is no mistake), the time is shown according to the timezone settings on the server hosting FreshRSS (php.ini, date.timezone).

jgeboski:
Correct, I am referring to the time zone present within the browsing session.

aledeg:
I leave [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset) here for information.

Frenzie:
Note that where it says
```html
<li class="item date">September 1<sup>st</sup> 2017 at 14:56&nbsp;</li>
```
It should be changed to something like
```html
<li class="item date"><time datetime="2017-09-01 14:56">September 1<sup>st</sup> 2017 at 14:56</time>&nbsp;</li>

Then you can easily do whatever you want with it.

Alkarex:
Good idea @Frenzie

aledeg:
@jgeboski I've just added an extension to customize how dates are displayed. It's available [here](https://github.com/aledeg/FreshRSS-extensions/tree/master/xExtension-DateFormat). Let me know what you think.

aledeg:
The issue has been addressed, there is no feedback from @jgeboski. I think it's time we close this.

Alkarex:
@aledeg You extension does not offer different time zone per user, does it? Maybe we could keep an issue for this aspect

aledeg:
@Alkarex You're right, it does not support different timezone per user. But this is not what was requested at the first place. I think you get confused with issue #2754.
This one is only about formatting the date. The other is for the timezone support.

Alkarex:
Ok, let's continue in https://github.com/FreshRSS/FreshRSS/issues/2754 for the timezone part

PR #2218, opened 2019-01-16
Docker Alpine PHP timezone

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2153


xobs:
Thank you!  That works for me now, and all of my news items in the feed have the correct time as well.

Alkarex:
Thanks for the feedback 👍

Alkarex:
Cf. https://github.com/FreshRSS/FreshRSS/issues/853

PR #2234, opened 2019-01-31
Less jQuery

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/2199


p/scripts/main.js
```diff
@@ -256,45 +273,51 @@ function mark_favorite(div) {
...
+			const json = xmlHttpRequestJson(this);
+			let inc = 0;
+			if (div.classList.contains('favorite')) {
+				div.classList.remove('favorite');
+				inc--;
+			} else {
+				div.classList.add('favorite');
+				inc++;
+			}
+			div.querySelectorAll('a.bookmark').forEach(function (a) { a.setAttribute('href', json.url); });
```
Frenzie: What's support for this like precisely?

p/scripts/main.js
```diff
@@ -406,7 +429,6 @@ function last_feed() {
 
 function prev_category() {
 	const $active_cat = $('#aside_feed .tree-folder.active');
-
```
Frenzie, line 409: I find a little whitespace easier to parse, personally. Whichever of course. :-)

Alkarex: This section is not done yet. Still in jQuery 🤓

p/scripts/main.js
Alkarex: I do not understand the question 😕

Frenzie: Sorry, it was a mobile post.

I thought not all modern(ish) browsers supported forEach yet but I guess I was thinking of IE8, so never mind. ;-)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

(I'd swear some older WebKit supported querySelector but not forEach, but since I'm clearly thinking of a decade ago figuring out if I recall correctly is not important.)

Alkarex: Ah, see the Polyfills section at the top. 
I am trying to support IE11 (still easy enough so far) but not older.

Alkarex:
This started as an experiment to see how much of our jQuery code could be transformed into native JavaScript. It turned out all of it could, easily :-)
So now jQuery is gone from the main view (I am not done yet with the subscription page and the global view) and will only remain for the statistics page.
In addition, our own code is now smaller (and easier, or at least more standard).

I have just done the same for the shortcuts library. Removed, and our own code is smaller, easier, faster, and with control of the order/context of the shortcuts.

All in all, it is a major rewrite, which will need more testing than usual. If you are motivated enough to test this branch, please go ahead. Otherwise it should soon land in /dev when the global view and subscription page are done. @aledeg @Frenzie @marienfressinaud 

I hope you will like it!

Frenzie:
@Alkarex Oh, I love  plain JS instead of jQuery. 

Btw, I didn't even realize that constants and such were supported by IE11. That's nice too. :-)

Alkarex:
IE11 is indeed quite capable. Among the things I would like to use that are not supported by IE11, the main one is arrow functions. Other ones: no support for JSON response type in XMLHttpRequest, no Fetch API, no async functions, but those are all easy to replace.

marienfressinaud:
Nice! I'll try as soon as it lands in `dev`! :)

aledeg:
@Alkarex I haven't tried it yet but good job :+1:

Alkarex:
Merged in /dev :-)
I expect that a few things will have to be fixed, so tests welcome !

Alkarex:
@aledeg If you have a chance to test the /dev branch, I would be pleased to get some feedback on features such as shortcuts, user queries, sharing, etc. to make sure they have not been broken in the process :-)

aledeg:
@Alkarex I'll try to find some time!

aledeg:
@Alkarex I did a quick test and changing users in the configuration page returns a 404. I need to make more thorough tests.

aledeg:
The shortcuts `Shift + K`, `Shift + J`, `Space` are not working because they collide with Firefox shortcuts.
The shortcut `n` loops through article while the shortcut `p` doesn't.

Apart from that, it seems OK

Alkarex:
@aledeg Thanks! https://github.com/FreshRSS/FreshRSS/pull/2255 fixes the 404 and space problem, and changes the navigation loop behaviour (continuing to next feed / category). I was not able to reproduce the `Shift + K` or `Shift + J` bugs (tested with Firefox on Linux + Windows).

aledeg:
@Alkarex I was probably sleepy this morning. I couldn't reproduce either :)

Kaan88:
I think this broke keep folder state extension, I know that isn't an official extension but thought I should mention it.

Frenzie:
It might be best if you mention it over at https://github.com/oyox/FreshRSS-extensions/tree/master/xExtension-KeepFolderState as well. :-)

Kaan88:
True but that repo is dead, I have a year old issue there. I think these extensions should be removed from the suggestions. I'll still create an issue though.

Alkarex:
If there are popular extensions out there not working, try to contact the original author, and if the person is not interested in maintaining the extension, let's try to find them new maintainers. We can move some in https://github.com/FreshRSS/Extensions

Kaan88:
I couldn't find a way to contact him, his website doesn't work either. How can we determine which extensions are popular? I vote for "keep folder state". :D

Alkarex:
Please make a PR in https://github.com/FreshRSS/Extensions with the current code.
What is that extension doing exactly? If enough interesting, we might move it to FreshRSS core.

Kaan88:
I have tried but couldn't make a pull request here. I ended up making one on oyox's repo.

Alkarex:
@Kaan88 Would you like to try again?
1. Fork https://github.com/FreshRSS/Extensions
2. Make new branch *KeepFolderState*
3. Download https://github.com/oyox/FreshRSS-extensions/archive/master.zip
4. Copy paste that folder xExtension-KeepFolderState from the ZIP to the Extensions
5. Send PR :-)

Frenzie:
Btw, I just came across this: https://github.com/nefe/You-Dont-Need-jQuery

(Not something any of us need, but it might be a useful reference for a passer-by.)

PR #2260, opened 2019-03-04
Use getenv instead of $_SERVER for environment variables

Alkarex:
getenv() is the proper way, and $_SERVER was not working on some
configurations.
Improvement of https://github.com/FreshRSS/FreshRSS/pull/2213


Frenzie:
I'm surprised $_SERVER would work for such a thing at all.

Alkarex:
It seems to work for CGI-like setups

Issue #2262, opened 2019-03-05
Update/Refresh behaviour


Joniator:
Im wondering how exactly does refreshing feeds behave?
It feels like refreshing doesn't really do something most of the time.
It just immediately hits you with "No feeds to refresh", until at somepoint it does not and updates every feed with many articles.
But this seems related to the "Do not automatically refresh more often than", which might be better named "Do not refresh more often than" then, since it doesn't seem to matter if automatic or not.

I think that one of the following options should be more clear to the user:

- Either rename the setting (Name seems consistently misleading at least in english and german) and make the possible values more based on reality. A few cheap requests like this don't hurt if you fire them up every 1-5 Minutes, and 20 minutes minimum seems overly defensive to me.
- Make sure this only affects automatic refreshs, not the one issued by the user via the interface.

If you can rely on the "Do not automatically refresh more often than" setting, it would be easily possible to run the cron script every minute, and check if the time is reached. This would need updated install instructions, where the cron script is run 'hourly', which seems like an eternity for me.

FreshRSS-Version: 1.13.1
PHP-Version: 7.2.14

Frenzie:
> If you can rely on the "Do not automatically refresh more often than" setting, it would be easily possible to run the cron script every minute, and check if the time is reached.

That's how it works, yes.

> This would need updated install instructions, where the cron script is run 'hourly', which seems like an eternity for me.

That's an example, not an instruction. I use 15 minutes in cron, combined with a 30 minute default in FreshRSS, as well as a few custom per-feed settings. That's mainly higher, .e.g, I've got a comic set to once a week so it shows up in the weekend. (I would support a PR to change the example to 15 minutes.)

Many feeds have WebSub, btw.

Alkarex:
Yes indeed, we should update the examples and terminology.
The examples we give for the Docker deployment are faster:
https://github.com/FreshRSS/FreshRSS/tree/master/Docker#cron-job-to-automatically-refresh-feeds

Joniator:
> That's how it works, yes.

Okay, at I'm not doing anything wrong then.
But the point still stands, for me at least 20 Minutes until update seems limiting for no real reason.
I like to see updates  as soon as possible, and maybe someone watches a gitlab teamstream and wants to see events rather soon than later, in which case 20 Minutes feel like an eternity

Another point still stands: It doesn't update if you press the update button if the time isn't reached.
[...]

PR #2263, opened 2019-03-06
Equalize archiving/feed specific refresh settings, add shorter refresh intervals

Joniator:
As discussed in #2262, I decided it might at least be a small comfort change to equalize the archiving/updating refresh times, since both had different time ranges, now they both span from 10min to 1mo.
I also added a 10 min option since it seems to be the lowest decently safe value.

For lower values there might be an option to either add a warning/info hinting that too frequent refreshs may result in exceeding your quota.
Either via JS, if a lower value is selected, or just always visible.


Alkarex:
Hello @Joniator 
The difference between default values (>= 20min) and feed-specific values (>= 15min) was on purpose. Faster refresh rates on feed-specific options are to give a way for the user to insist a bit more. Values < 15min would not do anything anyway as we also have a local disk cache as another safety barrier to prevent excessive refresh rates.

marienfressinaud:
I think we'll not merge this PR since the current values were chosen on purpose, I'm closing it.

Issue #2216, opened 2019-01-12
[Feature Requests]: Auto-apply filter on page load / Don't download based on filters / Filter Organization & Editing


TechArtisanIO:
After using FreshRSS for a bit, I think it would be great if you could auto-apply (as in when you login to your FreshRSS account it defaults to a filter you specificy) a filter to a particular feed, category or even on a "app-wide" basis. Even better would better to prevent FreshRSS from downloading materials based on filters. 

Additionally, it be nice to have the capability to reorder filters on the User Queries page and be able to directly edit those filters on that page as well.

Thanks for a great app!

Alkarex:
First implementation available in https://github.com/FreshRSS/FreshRSS/pull/2275

TechArtisanIO:
Cool. Thanks! I'll give this a shot.

aledeg:
@ahelton Could you split your request to be able to keep track of those? It's hard to follow through when there are more than one request in an issue. Thank you!

aledeg:
See #2015

math-GH:
@ahelton  How is the status? The user queries are already improved. What is left from the rest?

Issue #1368, opened 2016-11-09
[FR] Filter processing


anayrat:
Add the possibility to perform action according to rules: 

Filter "Postgres" : 
  - Postgres in title or content

Perform : 
 - Mark as read
 - Assign tag "postgres"

anayrat:
For information this functionnality is in tinyrss. It is missing for my usage :(

Alkarex:
Currently, it is already possible to (manually) search for a string, and mark the results as read.

SebCorbin:
Hello there, just moved from Feedly to FreshRSS after balancing out between TinyTinyRSS and FreshRSS. My motivation was that FreshRSS looks very nice and is mobile friendly.
Anyway: good work!

But filters are a feature that I miss too, I did not expect it to be only "filter articles list," but instead more than an "automatically forget articles containing X", maybe it should be named "blacklist" or something like that?

Alkarex:
@SebCorbin See the related issue for hiding spoilers https://github.com/FreshRSS/FreshRSS/issues/1161
Currently, this is not in the FreshRSS core (can be done manually, or by an extension, or by a cron job with curl). However, I have just increased its priority to try to include it in the next version 1.8 of FreshRSS.

Jucgshu:
@alkarex Sorry to ask: has it been included to FreshRSS? Thanks for your work as always :wink:

Alkarex:
@Jucgshu No, not yet, but for the time being, extensions are relatively easy to do for making custom rules.

Alkarex:
First implementation available in https://github.com/FreshRSS/FreshRSS/pull/2275

albertopasqualetto:
Should close as already implemented

Frenzie:
Alright.

PR #2275, opened 2019-03-16
Filter actions

Alkarex:
Feed actions based on search filters. ~~Work in progress~~


aledeg:
I've started to play around with something similar the other day but as an extension :)
I can't wait to see that in action :+1:

Alkarex:
Ah yes, an extension would work too. I think this might be a feature popular enough to make it to the core. If you have any code or concept that can be used here, do tell :-)

aledeg:
I was thinking about highlighting the matches, or highlighting the sentence of the matches, or highlighting the paragraph of the matches.

Alkarex:
For 1.14.0, I think I will only have the *auto mark as read* feature, but the structure is ready for e.g. auto-favourite, auto-tag, etc.

image

image

Alkarex:
Let's continue testing in /dev

Alkarex:
Contributes to
https://github.com/FreshRSS/FreshRSS/issues/2216
https://github.com/FreshRSS/FreshRSS/issues/1368

Jucgshu:
Hi @Alkarex 

Being too excited about this features is probably the reason I can't find it in the 1.14.0 version?

I can create filters just fine, but can't seem to find how to setup the filter actions. Is there something I miss?

I use the FreshRSS's Docker version and upgraded from the official image this PM.

Thanks.

-j

aledeg:
@Jucgshu you can find it on every feed configuration!

Jucgshu:
Oww ok, I was kinda hoping to find a global option, but that's probably for later then.

That's already a big step forward and a very cool feature to have.

Thanks team (and thanks @aledeg for the swift answer ;p)

-j

Alkarex:
@Jucgshu Is it mainly a global filter for automatically marking as read, which you want?

Jucgshu:
Having filter actions per feeds is *very* cool, but I have to say: having the ability to extend the existing filter feature would allow to do what I'd love to do with FreshRSS: achieve the no-spoiler feature by marking as read articles based on search, regardless of the setup of the feeds.

That would be a killer, yes, having this option in the 'filter' option panel and be able to define actions per filters.

SecT0uch:
Thanks @Alkarex for your work.
How does it work for a multi-word filter ? 
I added these filters last day :
```
intitle:'Revue de presse de l'April pour la semaine'
intitle:'Agenda du Libre pour la semaine'
```
But it doesn't seem to work.
Is a newline interpreted as a *OR* ?
Maybe the right way to do is to put both filters on the same line, with a *OR* between them ?

Alkarex:
Yes, with `OR` on the same line
Check the documentation https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html#with-the-search-field

Issue #2086, opened 2018-10-28
Change label for read shortcut


Alwaysin:
https://github.com/FreshRSS/FreshRSS/blob/e04804d0f67dd43fd3f072b9a127768ee7b7b56c/app/i18n/en/conf.php#L161

'r' shortcut marks as read but also marks as unread depending on the current status of the article. The label on the configuration page should reflect this, otherwise one could think there is no shortcut to mark as unread. Or we can create a specific shrtocut 'u' for this purpose.

Frenzie:
Toggle mark as read?

Alkarex:
Anyone for a PR on this? :-)

aledeg:
@Alwaysin available on dev branch

PR #2277, opened 2019-03-17
Update some translation strings

aledeg:
See #2086


Frenzie (commented):
You should lose the "marker" for simply `toggle read` and `toggle favorite`. It's more elegant and clearer at the same time. :-)

NB This doesn't (necessarily) affect localizations like `l’indicateur de favori`.

aledeg:
@Frenzie done!

Issue #2046, opened 2018-10-14
CLI should check parameter name, and exit in error when encountering an unknown parameter


oupala:
When I enter an api password in the web interface, I get an error because of an error format. But there is no tip about a specific format.

When I set the api password with the cli, I get no error, only a success. With the same password. But in the end, the password does not work with EasyRSS.

What is the minimal size of the api password? If the password is shorter and is set via the cli, should it work or should it fail?

oupala:
In the Web UI, it is not very clear to see that the api password is *also* constrained to 7 chars (as the main password.

oupala:
As far as I have tested, creating a user via CLI does not set the api password.

Here is my command line:

```bash
{{ _freshrss.install_path }}/cli/create-user.php --user {{ _freshrss.end_user.login }} --password '{{ _freshrss.end_user.password }}' --api-password '{{ _freshrss.end_user.password }}' --language {{ _freshrss.locale }} --email {{ _freshrss.end_user.email }} --token {{ _freshrss.token }}
```

As you can see, I'm trying to set the api password: `--api-password '{{ _freshrss.end_user.password }}'` (all vars are infered by ansible).

But in the end, the config file shows that no password is set:

```config
  'passwordHash' => 'filled-with-a-hash',
  'apiPasswordHash' => '',
```

I can however set the api password by the following command line:

```bash
{{ _freshrss.install_path }}/cli/update-user.php --user myusername --api_password 'myapipassword'
```

So, as far as I understand, `cli/update-user.php` works while `cli/create-user.php` does not.

Alkarex:
Please note that it is `--api_password` with an underscore, in both cases

oupala:
It works way better with the right parameter name. Thanks!

The strange thing is that hashes are different event if passwords are the same:

```bash
--password '{{ _freshrss.end_user.password }}' --api-password '{{ _freshrss.end_user.password }}'
```

By the way, this could be an opportunity to rename this issue and to improve the CLI robustness by checking the parameter's name and by exiting in error if one parameter is not recognized. This would be clearly better that silently ignoring a parameter that is wrongly spelled... Don't you think?

Alkarex:
Yes, hashes are unique, because of a random salt.
Mja, we could put some warnings

aledeg:
@oupala Please review the related PR!

oupala:
I'm not sure to understand what pull request you are talking about. Could you please be more precise?

[...]

PR #2278, opened 2019-03-17
Add an option validation on cli commands

aledeg:
If an option used on cli is not recognized, the command
aborts and displays an error message.
If the typed option is similar to one of the recognized
options, a hint is displayed.

At the moment, there is a limitation on long options.
Short options are not validated at the moment.

See #2046

### DIFF
--- a/cli/_cli.php
+++ b/cli/_cli.php
@@ -3,6 +3,9 @@ if (php_sapi_name() !== 'cli') {
 	die('FreshRSS error: This PHP script may only be invoked from command line!');
 }
 
+const REGEX_INPUT_OPTIONS = '/^--/';
+const REGEX_PARAM_OPTIONS = '/:*$/';
+
 require(__DIR__ . '/../constants.php');
 require(LIB_PATH . '/lib_rss.php');	//Includes class autoloader
 require(LIB_PATH . '/lib_install.php');
@@ -63,4 +66,26 @@ function performRequirementCheck($databaseType) {
 		}
 		fail($message);
 	}
+}
+
+function getLongOptions($options, $regex) {
+	$longOptions = array_filter($options, function($a) use ($regex) {
+		return preg_match($regex, $a);
+	});
+	return array_map(function($a) use ($regex) {
+		return preg_replace($regex, '', $a);
+	}, $longOptions);
+}
+
+function validateOptions($input, $params) {
+	$sanitizeInput = getLongOptions($input, REGEX_INPUT_OPTIONS);
+	$sanitizeParams = getLongOptions($params, REGEX_PARAM_OPTIONS);
+	$unknownOptions = array_diff($sanitizeInput, $sanitizeParams);
+
+	if (0 === count($unknownOptions)) {
+		return true;
+	}
+
+	fwrite(STDERR, sprintf("FreshRSS error: unknown options: %s\n", implode (', ', $unknownOptions)));
+	return false;
 }
--- a/cli/_update-or-create-user.php
+++ b/cli/_update-or-create-user.php
@@ -22,7 +22,7 @@ if (!$isUpdate) {
 
 $options = getopt('', $params);
 
-if (empty($options['user'])) {
+if (!validateOptions($argv, $params) || empty($options['user'])) {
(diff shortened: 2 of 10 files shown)


cli/_cli.php
```diff
@@ -64,3 +67,30 @@ function performRequirementCheck($databaseType) {
...
+
+	if (0 !== count($unknownOptions)) {
+	    $unknownOption = array_shift($unknownOptions);
+	    $possibleOption = array_filter($sanitizeParams, function($a) use ($unknownOption) {
+	        return levenshtein($a, $unknownOption) < 2;
+	    });
+	    if (0 === count($possibleOption)) {
+            fail(sprintf('FreshRSS error: unknown option: %s'), $unknownOption);
+        } else {
+            fail(sprintf('FreshRSS error: unknown option: %s. Do you mean %s?', $unknownOption, array_shift($possibleOption)));
```
Frenzie: I'd say "Did you mean" or perhaps "The most similar command/option/flag is:".

Side note, what if there are multiple results with a low Levenshtein distance?

Frenzie (commented):
Interesting idea, although given the fairly limited number of options I'm not entirely convinced that it's actually more helpful than the usage notice. :-)

cli/_cli.php
aledeg: Noted.
In the case of multiple results, the first one is displayed

aledeg:
I can remove the the suggestion part. It was just an experiment.
But the validation part is interesting because if one of the options is inexistent, the following one are discarded. In that case, you could end up doing things that you don't want to do.
See related issue for example.

Frenzie:
You could also consider printing both, so you'd get something like this:

```
Did you mean --bla

Usage notice: --bla --yada
```

aledeg:
Considering how the cli is build, it won't be possible to have both

Frenzie:
I think usage info is more useful personally. Any thoughts @Alkarex ?

aledeg:
I can do that but that means a lot of changes

Frenzie:
I don't think suggestion is very useful though. :-)

(At least not when there are fewer than 10 options.)

aledeg:
@Frenzie I've changed the behavior to display wrong parameters AND the usage. Let me know what you think!

cli/_cli.php
```diff
@@ -64,3 +67,25 @@ function performRequirementCheck($databaseType) {
...
+function validateOptions($input, $params) {
+	$sanitizeInput = getLongOptions($input, REGEX_INPUT_OPTIONS);
+	$sanitizeParams = getLongOptions($params, REGEX_PARAM_OPTIONS);
+	$unknownOptions = array_diff($sanitizeInput, $sanitizeParams);
+
+	if (0 === count($unknownOptions)) {
+		return true;
+	}
+
+	fwrite(STDERR, sprintf("FreshRSS error: unknown options: %s\n", implode (', ', $unknownOptions)));
```
Frenzie, line 89: I suppose you could consider giving `fail()` an argument regarding whether to die or not (true by default), or create a non-`die()`ing duplicate like a `warning()` function. But that might be premature abstraction.

Frenzie (approved):
lgtm!

cli/_cli.php
aledeg: That's a good point. Maybe we should do that if we intend to add more warnings.

Alkarex:
@aledeg Could you please fix the Travis white-space error? :-)

aledeg:
@Alkarex. Will do!

aledeg:
@Alkarex it's all good!

oupala:
As far as I have tested this merge request, it fixes the bug mentioned in #2046.

Thanks @aledeg for fixing this.

I still have 2 remarks:

- it is very strange to get syntax inconsistency in parameter name: `--api_password` seems not natural as we usually use dash *or* underscore, but not both of them, we should convert all parameter to dash syntax: `--api-password`
- it is also very strange to see that installing FreshRSS on an already installed FreshRSS returns an error, following idempotency we should more return a `0` return code:

```bash
>php cli/do-install.php --default_user myuser --language fr --api-enabled --db-type sqlite
FreshRSS looks to be already installed! Please use `./cli/reconfigure.php` instead.
>echo $?
1
```

PR #2279, opened 2019-03-17
Maximum number of articles to keep by feed

aledeg:


aledeg:
I am exploring a different path. I am closing that for now. I will provide a different one later.

Issue #2294, opened 2019-03-23
add optional headers to cli scripts


oupala:
The script output would be more easily readable if there was headers outputted. It could be an option to output the headers or not (`--headers true` or `--headers false`).

I was thinking about something like that:

```bash
>php cli/user-info.php -h --user myuser --headers false
*	myuser	2019-03-22T22:10:11+01:00	5.77 MiB	4 categories	20 feeds	909 reads	1 unreads	2 favourites

>php cli/user-info.php -h --user myuser --headers true
admin username last update               total space used # of categories # of feeds # of reads posts # of unreads posts # of favorite posts
-----|--------|-------------------------|----------------|---------------|----------|----------------|------------------|-------------------
*     myuser   2019-03-22T22:10:11+01:00 5.77 MiB         4 categories    20 feeds   909 reads        1 unreads          2 favourites
```

aledeg:
@oupala Please check #2296

oupala:
You guys are so fast in fixing issues!

It works well @aledeg.

Just one question: should we user *favorite* or *favourite*?

> Favorite and favourite are different spellings of the same word. Favorite is the preferred spelling in the U.S., while favourite is preferred in all the other main varieties of English ([source](https://grammarist.com/spelling/favorite-favourite/)).

And by the way, could this be extended to some other scripts where it makes sense? (if it does)

Frenzie:
@oupala I'd go for US English myself, but the English in the program consistently goes with favourites. Unless you're proposing to switch or to contribute a second variety of English, I'd say it's not really a question. ;-)

https://github.com/FreshRSS/FreshRSS/blob/8dcdde6251ae4dfc690b1a014488df125c5e5cdc/app/i18n/en/index.php#L24

https://github.com/FreshRSS/FreshRSS/blob/8dcdde6251ae4dfc690b1a014488df125c5e5cdc/app/i18n/en/index.php#L37

https://github.com/FreshRSS/FreshRSS/blob/8dcdde6251ae4dfc690b1a014488df125c5e5cdc/app/i18n/en/index.php#L45

https://github.com/FreshRSS/FreshRSS/blob/8dcdde6251ae4dfc690b1a014488df125c5e5cdc/app/i18n/en/index.php#L53

oupala:
I had no specific opinion between english english, and american english. I was only asking to be sure, and to get consistency.

But what your showing us demonstrate that **there is** consistency.

[...]

PR #2296, opened 2019-03-23
Add header to cli

aledeg:
Now there is a switch to display the header on user info.
While doing that, I've changed how the command is working to display
all users by default and to accept more than one user at once.
I also changed the display to make it more pleasing.

As this command displays all users by default. I wonder if we still
need the list user command.

See #2294

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -170,7 +170,7 @@ function format_bytes($bytes, $precision = 2, $system = 'IEC') {
 	$pow = $bytes === 0 ? 0 : floor(log($bytes) / log($base));
 	$pow = min($pow, count($units) - 1);
 	$bytes /= pow($base, $pow);
-	return format_number($bytes, $precision) . ' ' . $units[$pow];
+	return format_number($bytes, $precision) . ' ' . $units[$pow];
 }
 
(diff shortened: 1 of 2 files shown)


lib/lib_rss.php
```diff
@@ -170,7 +170,7 @@ function format_bytes($bytes, $precision = 2, $system = 'IEC') {
 	$pow = $bytes === 0 ? 0 : floor(log($bytes) / log($base));
 	$pow = min($pow, count($units) - 1);
 	$bytes /= pow($base, $pow);
-	return format_number($bytes, $precision) . ' ' . $units[$pow];
+	return format_number($bytes, $precision) . ' ' . $units[$pow];
```
aledeg, line 173: I've changed this space because it wasn't a regular space. I don't know if it was on purpose or not. The space was coded on two bytes and broke the cli display layout.

Alkarex: It was on purpose, but if it broke anything, it is not a problem to remove it

aledeg: What was the use case which needed such space?

Alkarex: thin non-breaking space, if I remember correctly, which is the typographic standard between a number and its unit.

PR #2317, opened 2019-04-01
[CI] Add PHP 7.3 to Travis configuration

Frenzie:
This also requires an update of phpcs, since the old version won't run on PHP 7.3.

By setting setting the tab-width to 40 it works around the behavior introduced in https://github.com/squizlabs/PHP_CodeSniffer/pull/1404 which erroneously interprets positioning spaces as indentation. ("If the line started with tabs, but had spaces after that, no error was thrown at all.")

That makes any line length checks ineffective, but I think line length checks aren't very useful anyway. They're basically just a (very!) rough indication that you might want to consider some refactoring.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -2,7 +2,7 @@
 <ruleset name="FreshRSS Ruleset">
 	<description>Created with the PHP Coding Standard Generator. https://edorian.github.com/php-coding-standard-generator/</description>
 	<!-- to circumvent https://github.com/squizlabs/PHP_CodeSniffer/pull/1404 -->
-	<!--<arg name="tab-width" value="10"/>-->
+	<arg name="tab-width" value="40"/>
 	<exclude-pattern>./static</exclude-pattern>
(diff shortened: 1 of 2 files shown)


phpcs.xml
```diff
@@ -2,7 +2,7 @@
 <ruleset name="FreshRSS Ruleset">
 	<description>Created with the PHP Coding Standard Generator. https://edorian.github.com/php-coding-standard-generator/</description>
 	<!-- to circumvent https://github.com/squizlabs/PHP_CodeSniffer/pull/1404 -->
-	<!--<arg name="tab-width" value="10"/>-->
+	<arg name="tab-width" value="40"/>
```
Alkarex, line 5: @Frenzie Since this workaround, have you seen anything better?
I am giving a try to https://marketplace.visualstudio.com/items?itemName=ikappas.phpcs in Visual Studio Code, and it would be nice to `<arg name="tab-width" value="40"/>`

Frenzie: I haven't checked.

PR #2331, opened 2019-04-03
Update git command

aledeg:
I removed an unnecessary option and rewrote a short option
to use the long option. I find it's easier to understand
long options when you don't type them.

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -32,7 +32,7 @@ class FreshRSS_update_Controller extends Minz_ActionController {
 		$output = array();
 		$return = 1;
 		try {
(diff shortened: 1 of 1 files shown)


Alkarex:
Ups, I merged too fast without reading carefully enough: one of the `-f` has disappeared. It was meant to be with a double `-f`
https://git-scm.com/docs/git-clean
Could you please fix @aledeg ?

aledeg:
@Alkarex Oups, I didn't read git documentation thoroughly. My bad

Frenzie:
As an aside, in principle I agree but -f is pretty straightforward. :-)

aledeg:
I've reverted it anyway :)

PR #2334, opened 2019-04-03
Fix auto_remove_article

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2323


Alkarex:
Arg, does not work with IE11 :-(

Alkarex:
Second attempt :-) Tests welcome

Issue #2328, opened 2019-04-02
Shortcuts problem when showing articles unfolded


Pitel:
j/k now (1.14.0) seems to jump by whole page, not article. It was ok in 1.13.

Chrome 75.0.3745.4, Linux

Frenzie:
What do you mean by "whole page" precisely?

ggdagg:
I confirm that I have the same symptom since Freshrss 1.14.0, however it's with Firefox 66.0.1. Before when I used j/k shortcuts it was opening the next/previous article, and the title of the article was at the top of the screen. Now it randomly open the next/previous article in the middle of the article itself, or even jump an article and ends up in the middle of the next one. I have the same issue with the navigation arrows. My options are “normal view”, “show only unread”, “show articles unfolded by default”. For testing, I also tried different themes, but the problem is still here.

Tested on 2 different Firefox 66.0.1 on Linux.

Alkarex:
@Frenzie Were you able to reproduce the issue? I could not (tested on Ubuntu 18.04 LTS with Firefox 66 and Chromium 73)
@Pitel and @ggdagg Could you please try our /dev branch (upcoming 1.14.1), where I have just fixed a number of things regarding shortcuts and mouse clicks?

Frenzie:
No, I wasn't.

ggdagg:
I just tried with 1.14.1-dev, and unfortunately I still have the same issue. Here is a GIF of the test I did, and might help you understand better what I meant: https://imgur.com/a/RPFGPD1
I also tried again with the keyboard shortcuts, and it's also the same. Even worst I would say since the "open next article" key just goes at the bottom of the page and marks all articles as read in my case. This is not shown on the GIF.

It has been test with a brand-new database, same options as before (“normal view”, “show only unread”, “show articles unfolded by default”), and tried on Windows and Linux with Firefox 60.0.2 and 60.6.0esr, respectively.

Frenzie:
Just a shot in the dark here, but on what kind of OS and filesystem are you running the server? Is it possible that something's going wrong with the `filemtime()` so that you're getting the cached version of the older script or something along those lines?

It'll be in the first few lines of the HTML source:
```html
<script src="../scripts/main.js?1554061833" defer="defer" async="async"></script>
```
[...]

PR #2336, opened 2019-04-03
Fix shortcut problem when showing articles unfolded

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2328


Alkarex:
Should be fixed with https://github.com/FreshRSS/FreshRSS/pull/2336
Tests welcome

Issue #2341, opened 2019-04-04
Does not remain marked as read


Tealk:
When I open a post with the middle mouse button it is marked as read.
If I then get the message "there are new posts" and click on it to reload the page. If this previous post is unread again.

Alkarex:
Hum, I cannot reproduce. Could you please provide a bit more info? (platform, browser, settings...)
And you are using the latest version of the /dev branch, right? I.e. including the latest commits as visible on https://github.com/FreshRSS/FreshRSS/commits/dev

Alkarex:
If is is only limited to specific feeds, it could be feeds that are updating their articles (in which case, please see the option *Mark updated articles as unread* )

Frenzie:
@Alkarex Although it occurs to me (and this doesn't match the behavior described) that you could probably unintentionally do something like this in the ~1 s interval.

Tealk:
Yes I did not update again until today at 8 o'clock.
Debian stretch
FF 67.0b7
Settings:
![image](https://user-images.githubusercontent.com/12276250/55541369-2fae2a00-56c5-11e9-81fd-cbeb9c4be994.png)

> updating their articles

I'm sure that's not the case, there's never been an update on this feed before.

Tealk:
I don't succeed in reproducing it either but I know for sure that this will happen because I still had the tab with the contribution open. I saw that it was read and it had to be read.

Alkarex:
@Frenzie Let's add a `onbeforeunload` event for that possibility.
Edit: And use `delayedClick()` for the reload link

Alkarex:
Please check https://github.com/FreshRSS/FreshRSS/pull/2346

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2346 merged in /dev

PR #2346, opened 2019-04-04
Better handling of bad request and fast unload

Alkarex:
Warnings for bad requests, confirmation before leaving a page with
pending mark-as-read requests (not the others for now)


Alkarex:
Should fix https://github.com/FreshRSS/FreshRSS/issues/2341

Issue #2345, opened 2019-04-04
"Hide articles after reading" does not work any more


oupala:
Since version 1.14.0, the option "Hide articles after reading" does not seem to work anymore. I tried to uncheck it and then re-check, identical behavior.

When I click on the *read* envelope, it passes from red to grey after a few seconds, and nothing else happens.

Here what I've found in the web console:

```
TypeError: json is null
main.js:213:1
send_mark_read_queue/req.onload
https://example.com/FreshRSS/p/scripts/main.js:213:1
```

Current version is 1.14.1-dev.

Alkarex:
Are you using the latest /dev branch? What browser? Any FreshRSS extension?
I am preparing a patch to show an error message, but that will not in itself your problem, which seems to be due to a failed request.

oupala:
Last commit is commit 71f93e62193a32da7e2da6acfaea6d5d00192c84

> Merge branch 'master' into dev

My browser is Firefox 60.6.1esr 64 bits for linux debian.

No FreshRSS extension.

When I click on the "read" icon, I can see an http request fired:

https://example.com/FreshRSS/p/i/?c=entry&a=read which return code is 200 OK

And the response content is: "### Application problem ###
Permission is denied for `/data/FreshRSS/data/users/myuser/log.txt`"

I've changed the permissions on the data directory, and I now don't have any error in the content, nor in the web console, but I still get the same bad behavior.

Alkarex:
Patch for showing a warning if there is a bad request https://github.com/FreshRSS/FreshRSS/pull/2346

I believe you might have a read-only problem. Please re-apply the correct right-access on the whole `./data/` directory, e.g.

```sh
cd ./FreshRSS/
sudo chown -R :www-data . && sudo chmod -R g+r . && sudo chmod -R g+w ./data/
```

oupala:
As I told you in my previous message, I changed the permission on the data directory.

Now, I don't have any error in the console, nor I have an error in the http response. I now have: `{"tags":[]}`.

Alkarex:
What database are you using?

oupala:
sqlite3

Alkarex:
Please double-check the access rights of `./FreshRSS/data/users/**you**/db.sqlite`

oupala:
```bash
>ls -al data/users/myuser/db.sqlite 
-rwxrwxrwx 1 ops www-data 6050816 avril  4 22:17 data/users/myuser/db.sqlite
```

[...]

Issue #2332, opened 2019-04-03
"mark as read" has a ~1sec delay


sleeksorrow:
After updating to 1.14.0 marking articles as read has a delay of about 1 second until the "eye" icon and the headline turns grey. The other way around, marking as unread, has immediate visual feedback.

Maybe marking as read is not handled asynchronously anymore? Maybe intended?

Alkarex:
Yes, intended, and even more asynchronous than before

https://github.com/FreshRSS/FreshRSS/issues/2310#issuecomment-478465498

> Regarding the delay, it is because marking as read is now sent in batch, max once per second. You can mark many as read or even scroll down without waiting.

We will follow-up in https://github.com/FreshRSS/FreshRSS/issues/2304 *instant visual feedback to users*

oupala:
I'm invitating myself into this issue, sorry if it's not right solution.

I've just updated to version 1.14.0 and it appears that an article does no longer disappear when marked as read. I have to refresh the page (with F5) to see the that the article has disappeared from the unread article list.

Is this the expected behavior?

Frenzie:
@oupala This is with the `Hide article after reading` setting checked? That's in Settings → Reading.

Alkarex:
@oupala Yes, there is a bug to fix https://github.com/FreshRSS/FreshRSS/issues/2323

sleeksorrow:
Thank you for the explanation. I can live with this behavior until #2304 is in place. It's just the impression of unresponsiveness, but after all it's all usable as before.

oupala:
@Frenzie Yes, `Hide article after reading` is checked. I'll check issue #2323.

Sorry for squatting this issue.

Alkarex:
Could you please try https://github.com/FreshRSS/FreshRSS/pull/2349 ?

sleeksorrow:
Applied to dev branch and works like a charm for me. Thank you!

Alkarex:
Thanks for the quick feedback :-)

PR #2338, opened 2019-04-03
FreshRSS 1.14.1

Alkarex:
The big changes / refactoring in the [version 1.14.0](https://github.com/FreshRSS/FreshRSS/pull/2298) released on Sunday introduced a number of bugs, requiring a [quick minor release 1.14.1](https://github.com/FreshRSS/FreshRSS/milestones/1.14.1).

Please help checking the /dev branch thoroughly so can we release it on Saturday :-)

Full [changelog](https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md):

* Bug fixing (regressions introduced in 1.14.0)
	* Fix *load more articles* when using ascending order [#2314](https://github.com/FreshRSS/FreshRSS/issues/2314)
	* Fix cron in the Ubuntu flavour of the Docker image [#2319](https://github.com/FreshRSS/FreshRSS/issues/2319)
	* Fix the use of arrow keyboard keys for shortcuts [#2316](https://github.com/FreshRSS/FreshRSS/issues/2316)
	* Fix control+click or middle-click for opening articles in a background tab [#2310](https://github.com/FreshRSS/FreshRSS/issues/2310)
	* Fix the naming of the option to unfold categories [#2307](https://github.com/FreshRSS/FreshRSS/issues/2307)
	* Fix shortcut problem when using unfolded articles [#2328](https://github.com/FreshRSS/FreshRSS/issues/2328)
	* Fix auto-hiding articles [#2323](https://github.com/FreshRSS/FreshRSS/issues/2323)
	* Fix scroll functions with Edge [#2337](https://github.com/FreshRSS/FreshRSS/pull/2337)
	* Fix drop-down menu warning [#2353](https://github.com/FreshRSS/FreshRSS/pull/2353)
	* Fix delay for individual mark-as-read actions [#2332](https://github.com/FreshRSS/FreshRSS/issues/2332)
	* Fix scroll functions in Edge [#2337](https://github.com/FreshRSS/FreshRSS/pull/2337)
* Bug fixing (misc.)
	* Fix extensions in Windows [#994](https://github.com/FreshRSS/FreshRSS/issues/994)
	* Fix import of empty articles [#2351](https://github.com/FreshRSS/FreshRSS/pull/2351)
	* Fix quote escaping on CLI i18n tools [#2355](https://github.com/FreshRSS/FreshRSS/pull/2355)
* UI
[...]

PR #2349, opened 2019-04-06
Fix hide articles + batch option

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2332
https://github.com/FreshRSS/FreshRSS/issues/2345
Re-introduces the instant-remove article.
Batch mark-as-read only used for fast actions like scroll and keyboard
shortcut for next/previous articles.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -162,6 +162,26 @@ function incUnreadsTag(tag_id, nb) {
 	}
 }
 
+function removeArticle(div) {
+	let scrollTop = box_to_follow.scrollTop;
+	let dirty = false;
+	const p = div.previousElementSibling,
+		n = div.nextElementSibling;
+	if (p && p.classList.contains('day') && n && n.classList.contains('day')) {
+		scrollTop -= p.offsetHeight;
+		dirty = true;
+		p.remove();
+	}
+	if (div.offsetHeight > 0 && div.offsetParent.offsetTop + div.offsetTop + div.offsetHeight < scrollTop) {
+		scrollTop -= div.offsetHeight;
+		dirty = true;
+	}
+	div.remove();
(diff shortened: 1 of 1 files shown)


Alkarex:
Arg, now the articles disappear during expansion. Fix coming

aledeg:
You'll get there eventually!

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/2352

squromiv:
Installed #2352. Option "Hide articles after reading" works normally, as earlier.

Alkarex:
Thanks for the test! A few more patches have just landed in the /dev branch before the imminent release of [FreshRSS 1.14.1](https://github.com/FreshRSS/FreshRSS/pull/2338). Additional tests welcome on the /dev branch.

squromiv:
@Alkarex I have checked again and must note, that currently viewing article is not hidden, when it is marked as read. I think, it should be corrected.

Alkarex:
Hum, then there would be an option conflict with marking-as-read automatically, which would make it impossible to read an article. This is actually what I fixed in https://github.com/FreshRSS/FreshRSS/pull/2352
How do you mark an article as read in your case?

squromiv:
v.1.13.1

v.1.14.1-dev
1 14 1-dev

Alkarex:
Ok @squromiv , could you please test https://github.com/FreshRSS/FreshRSS/pull/2354 ?

Alkarex:
Merged in /dev . Additional tests welcome

squromiv:
Good job!
And what do you think about such an option? I would prefer to use something like this.
screen_2019 04 07 00111
Is there anybody, who interested in it?

Issue #2359, opened 2019-04-07
Unable to cli/prepare.php - "Arrays are not allowed as constants" via lib_rss.php


bhyde:
Shortly after git clone:
```
#$ whoami
root
#$ cd /usr/share
#$ rm -rf FreshRSS
#$ git clone https://github.com/FreshRSS/FreshRSS.git
git clone https://github.com/FreshRSS/FreshRSS.git
Cloning into 'FreshRSS'...
remote: Enumerating objects: 7, done.        
remote: Counting objects: 100% (7/7), done.        
remote: Compressing objects: 100% (7/7), done.        
remote: Total 28694 (delta 1), reused 0 (delta 0), pack-reused 28687        
Receiving objects: 100% (28694/28694), 11.19 MiB | 0 bytes/s, done.
Resolving deltas: 100% (19944/19944), done.
#$ cd FreshRSS
#$ cli/prepare.php 
PHP Fatal error:  Arrays are not allowed as constants in /usr/share/FreshRSS/lib/lib_rss.php on line 559
#$ 
```

aledeg:
The reason is your PHP version. I guess you are running PHP prior to 5.6.

bhyde:
Right you are, no idea how that happened.   Where's my stick?!

Frenzie:
The README does say:

> PHP 5.3.8+ (PHP 5.4+ recommended, and PHP 5.5+ for performance, or PHP 7 for even higher performance)

bhyde:
This is on centos 7.  What I observe is:
```
$ php --version
PHP 5.4.16 (cli) (built: Oct 30 2018 19:30:51) 
```
So that should have been a sufficiently fresh version of php; if the doc Frenzie quotes is accurate.

This is all a pretty fresh install and I got futher yesterday.  Today I was switching to mariadb.

Anyhow - forcing php 7 made things get all better.  Which i did by adding these tasks to my play book.
```yaml
  - yum: {name: epel-release}
  - rpm_key:
      state: present
      key: http://rpms.remirepo.net/RPM-GPG-KEY-remi
  - yum_repository:
      name: remi
      description: Remi YUM repo
      baseurl: http://rpms.remirepo.net/enterprise/$releasever/remi/$basearch/
      gpgcheck: yes
      enabled: yes
  - yum_repository:
      name: remi-php72
      description: Remi YUM repo
      baseurl: http://rpms.remirepo.net/enterprise/$releasever/php72/$basearch/
[...]

PR #2362, opened 2019-04-07
Fix API minor code syntax warning

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2359#issuecomment-480620713
Caught by PHP lint from PHP 7.3+


Frenzie:
Do you suppose the output should be grepped for `PHP Warning` or some such? It's the only one, after all.

Edit: that is, I didn't see a command line option to make `php -l` unsuccessful in case of warnings.

Alkarex:
I have just looked for the same :-)
It might be good indeed to alert a bit more for PHP lint warnings. Maybe an allowed fail if doable, or just a fail indeed if you can find a way to do that.

PR #2363, opened 2019-04-07
[CI] Force php -l failure

Frenzie:
By redirecting stderr to a file and checking if the filesize is greater than 0 bytes, we can also force failure for warnings.

See discussion in <https://github.com/FreshRSS/FreshRSS/pull/2362#issuecomment-480623161>.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
 
 script:
   - phpenv rehash
-  - git ls-files -z "*.php" | xargs -0 -n1 -P4 php -l 1>/dev/null 2>php-l-results
+  - find . -not -path "./lib/JSON.php" -name \*.php -print0 | xargs -0 -n1 -P4 php -l 1>/dev/null 2>php-l-results
   - if [ -s php-l-results ]; then cat php-l-results; exit 1; fi
   - |
     if [[ $VALIDATE_STANDARD == yes ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ install:
 
 script:
   - phpenv rehash
-  - git ls-files -z "*.php" | xargs -0 -n1 -P4 php -l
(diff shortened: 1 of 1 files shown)


Frenzie:
I hadn't noticed this before. On the one hand I could exclude libraries (perhaps better on a case by case basis just so we are aware), on the other hand it's not completely clear to me what the library is for given the native `json_*` functions?
```
PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; Services_JSON has a deprecated constructor in lib/JSON.php on line 119
PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; Services_JSON_Error has a deprecated constructor in lib/JSON.php on line 910
PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; Services_JSON_Error has a deprecated constructor in lib/JSON.php on line 924
```

Alkarex:
The JSON library was for the infamous period when php_json was not included in the default PHP distribution on Debian due to a license problem :-/
At some point we should be able to remove it.

Alkarex:
And yes, I think third-party libraries should be excluded (and manually checked during inclusion / updates)

Frenzie:
Since they're all fine (except for that one) and it is good to be aware of it, I'd argue against a priori exclusion of third-party libraries.

Frenzie:
Btw, my local PHP 7.3.3 instance doesn't show this warning. Must be some php.ini setting?

Alkarex:
Maybe PHP 7.3.4 (e.g from https://hub.docker.com/_/php ) (I am on mobile atm).
It is indeed good to be aware, so maybe a warning if possible

Frenzie:
It's <del>5.5 or 5.6</del> <ins>7.0</ins> and up on Travis (also mobile here), so definitely not a version thing. :-)

The way I wrote it now only libs/JSON is excluded, to which other files could be added if circumstances require it. The important thing is that it's a conscious decision imo.

PR #2373, opened 2019-04-10
Refacto request class

aledeg:
I split the code into more manageable chunks to make it more readable.
It's now easier to get a grasp on it. In the process, I tried to use
the yoda style notation to avoid unintentional variable assignement.
I also tried to exit early to avoid many "if" levels.

I have also added a better way to handle headers and a way to extract
languages from the request. It will be easier to choose which language
to use for i18n.

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -11,6 +11,7 @@ class Minz_Request {
 	private static $controller_name = '';
 	private static $action_name = '';
 	private static $params = array();
+	private static $headers = array();
 
 	private static $default_controller_name = 'index';
 	private static $default_action_name = 'index';
@@ -100,6 +101,7 @@ class Minz_Request {
 	 * Initialise la Request
 	 */
 	public static function init() {
+		static::$headers = $_SERVER;
 		self::initJSON();
 	}
 
@@ -112,13 +114,15 @@ class Minz_Request {
 
 	/**
 	 * Return true if the request is over HTTPS, false otherwise (HTTP)
+	 *
+	 * @return boolean
 	 */
 	public static function isHttps() {
-		if (isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
-			return strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) === 'https';
-		} else {
-			return isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on';
+		$header = static::getHeader('HTTP_X_FORWARDED_PROTO');
+		if (null !== $header) {
+			return 'https' === strtolower($header);
 		}
+		return 'on' === static::getHeader('HTTPS');
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/Minz/Request.php
```diff
@@ -294,8 +348,41 @@ private static function magicQuotesOff() {
 		}
 	}
 
+	/**
+	 * @return array
+	 */
+	private static function getHeaders() {
```
Alkarex: What is the purpose of this function? Is seems to just make a manual copy of `$_SERVER`

aledeg: I've extracted that in case we need to pre-process some of the keys or values. At the moment, it has no extra values than storing locally the variable ```$_SERVER```

marienfressinaud: I would rename this method: it's too close of `getHeader` which does something completely different. What do you think of `initHeaders`? Or `setHeaders`, taking an array as an argument (in this case, `$_SERVER`)?

lib/Minz/Request.php
```diff
@@ -95,18 +96,20 @@ public static function _param($key, $value = false) {
...
 	 * Return true if the request is over HTTPS, false otherwise (HTTP)
+	 *
+	 * @return boolean
 	 */
 	public static function isHttps() {
-		if (isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
-			return strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) === 'https';
-		} else {
-			return isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on';
+		if (null !== $header = static::getHeader('HTTP_X_FORWARDED_PROTO')) {
```
marienfressinaud: Nitpick: I prefer to assign the variable on its own line, I think it makes the code more readable for newcomers. Just my opinion :) Looking at the rest of the code, I understand you preferred this solution anyway!

lib/Minz/Request.php
```diff
@@ -229,39 +276,46 @@ public static function bad($msg, $url = array()) {
...
 			}
 		}
 	}
 
+	/**
+	 * @return string
+	 */
+	private static function extractContentType() {
+		if (null === $contentType = static::getHeader('CONTENT_TYPE')) {
+			$contentType = static::getHeader('HTTP_CONTENT_TYPE'); // PHP < 5.3.16
```
marienfressinaud: I'm not sure we still need this since we dropped the support of PHP 5.3 :)

lib/Minz/Request.php
```diff
@@ -294,8 +348,41 @@ private static function magicQuotesOff() {
...
 	public static function isPost() {
-		return isset($_SERVER['REQUEST_METHOD']) &&
-			$_SERVER['REQUEST_METHOD'] === 'POST';
+		return 'POST' === static::getHeader('REQUEST_METHOD');
+	}
+
+	/**
+	 * @return array
+	 */
+	public static function getPreferredLanguage() {
```
marienfressinaud, line 367: I'm not sure why you added this method (it's not a problem)

marienfressinaud (commented):
This PR was here for a while, but it's nice improvements of the code :+1:

lib/Minz/Request.php
Frenzie: Not just for newcomers, it's just clearer at a glance while skimming.

lib/Minz/Request.php
aledeg: I've done that before we drop 5.3 support :)

lib/Minz/Request.php
aledeg: This method extract the preferred language from the header. You could have something like `fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5` in the header and you'll end up with an array containing `fr-CH`, `fr`, `en`, `de` and `*`. When those values are retrieved, we can match the first language that we have a translation for. In this case, `fr`.

lib/Minz/Request.php
aledeg: Here there is only one variable so it does not make any difference. I change it back. But later in the code (`extractHost` function) it will be unreadable to have multiple assignments. I think that in that case, the variable has such a limited scope that it makes no sense to assign it separately.

lib/Minz/Request.php
marienfressinaud: I understand the code, what I mean is that you don't use it anywhere so I don't know why you added it :wink:

aledeg: I don't use now because my goal was to introduce some helper methods. This one is to allow FRSS to switch language automatically to the one preferred by the user. It will become useful on the public connection interface. When the user is not connected, FRSS does not know the user config. Thus relying on its preferred language is the sole way to change the interface.

aledeg: The code to do such thing must be added though :)

lib/Minz/Request.php
```diff
@@ -307,8 +359,41 @@ private static function magicQuotesOff() {
 		}
 	}
 
+	/**
+	 * @return array
+	 */
+	private static function extractHeaders() {
```
Alkarex: I still cannot see the reason for doing such a copy, and doing so with a manual copy

lib/Minz/Request.php
```diff
@@ -307,8 +359,41 @@ private static function magicQuotesOff() {
...
+			$headers[$key] = $value;
+		}
+		return $headers;
+	}
+
+	/**
+	 * @return mixed
+	 */
+	public static function getHeader($header, $default = null) {
+		if (array_key_exists($header, static::$headers)) {
```
Alkarex: Any reason for using `array_key_exists` instead of the faster `isset`, which also has the advantage of dealing with null values (which I do not think we want to deal with manually in our case)?

lib/Minz/Request.php
```diff
@@ -113,13 +115,15 @@ public static function is($controller_name, $action_name) {
...
+	 *
+	 * @return boolean
 	 */
 	public static function isHttps() {
-		if (isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
-			return strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) === 'https';
-		} else {
-			return isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on';
+		$header = static::getHeader('HTTP_X_FORWARDED_PROTO');
+		if (null !== $header) {
```
Alkarex: I do no think this line is needed. The two following lines can be `&&`'ed

Alkarex: Sorry, I read too fast. False alert :-)

Alkarex:
You all seem to like it, so I will not oppose to this PR (which I generally do not like) but I would still like to mention a few points :-)

* The new code is slower, and doing so in our critical path, i.e. the code path that is always called for each request, including the main view. When this is theoretically the case, it would be nice to provide a quick test to quantify the damages (in %, e.g. in number of requests per second, or more locally), so that we can take an informed decision.
* The new code is using more memory (more functions, function calls, copies)
* The new code is longer
* The new code is less standard (due to custom abstraction methods)
* The new code extracts methods that are not used elsewhere (premature refactoring)
* The new code does not seem to address any current need / issue (premature refactoring)
* As soon as we move to PHP 7.0+, we will be able to use https://php.net/migration70.new-features#migration70.new-features.null-coalesce-op which will be nicer/shorter/faster/more standard than some of the parts of this PR
* And (but I understand that is is a personal preference), I find the new code harder to read, in particular being less linear, so I have to jump back and forth between the definition of the custom abstracted methods and their use.

As a side note, I would like to add some automated tests at high level (e.g. calling all the CLI methods that we have to catch things such as https://github.com/FreshRSS/FreshRSS/pull/2646 ), including some performance tests to be alerted on performance regressions. Suggestions welcome (that should be another issue)

marienfressinaud:
While I agree on most of your points (especially the two first), I would re-balance a bit the interest of this PR:

- the new code is more understandable at a higher level (just few lines, and we can easily guess what value return the `extract*` methods just by their names) -> being less linear doesn't necessarily mean less comprehensible (but indeed, if you need to understand the details of the implementation, it's a bit more difficult)
- it is subjective, but I find the new code more elegant. Should it be elegant? I think so, it's more pleasant (at least for me) to work on such a codebase, and I think it is true for people discovering the code of FRSS. It's especially true about the `initJSON` method which had 5 nested blocks (4 `if` and 1 `foreach`)

aledeg:
@Alkarex 

>     * The new code is slower, and doing so in our critical path, i.e. the code path that is always called for each request, including the main view. When this is theoretically the case, it would be nice to provide a quick test to quantify the damages (in %, e.g. in number of requests per second, or more locally), so that we can take an informed decision.
Guilty. I did not take into account that while changing the request class. I agree that it's on the critical path and that we should be careful. Testing performance is a good idea.

>     * The new code is using more memory (more functions, function calls, copies)
Same as above. But starting with PHP7.0, memory usage has been drastically improved. I am not saying that we can be sloppy but sometimes, what was true for PHP5 is not true anymore.

>     * The new code is longer
True but as it is organized in small chunks, I found it more straightforward to dig into.

>     * The new code is less standard (due to custom abstraction methods)
I don't understand. Less standard than what exactly?
 
>     * The new code extracts methods that are not used elsewhere (premature refactoring)
True, but I wrote that because I am planning to use it. I didn't want to refactor some core code and at the same time add some features.
 
>     * The new code does not seem to address any current need / issue (premature refactoring)
See above

>     * As soon as we move to PHP 7.0+, we will be able to use [php.net/migration70.new-features#migration70.new-features.null-coalesce-op](https://php.net/migration70.new-features#migration70.new-features.null-coalesce-op) which will be nicer/shorter/faster/more standard than some of the parts of this PR
Agreed. I am waiting for that. There are so many cool features in PHP7 that are both more elegant and flexible.

>     * And (but I understand that is is a personal preference), I find the new code harder to read, in particular being less linear, so I have to jump back and forth between the definition of the custom abstracted methods and their use.
That's why I've tried to have meaningful method names so you don't need to dive in if you just want to understand the basic concept.

> As a side note, I would like to add some automated tests at high level (e.g. calling all the CLI methods that we have to catch things such as #2646 ), including some performance tests to be alerted on performance regressions. 

I think that a good idea!

Alkarex:
@aledeg By less standard, I mean using custom methods instead of the normal PHP code, in places where the reduction of code is not significant, such as in the example below:

```php
public static function isHttps() {
	if (isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
		return strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) === 'https';
	}
	return isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on';
}
```

vs.

```php
public static function isHttps() {
	$header = static::getHeader('HTTP_X_FORWARDED_PROTO');
	if (null !== $header) {
		return 'https' === strtolower($header);
	}
	return 'on' === static::getHeader('HTTPS');
}
```

where the first one does not require FreshRSS-specific knowledge, while the second leaves the reader with questions such as *How does `static::getHeader()` deals with empty strings or nulls?* and comes with a code fragmentation risk because `static::getHeader()` might not be remembered when working elsewhere in the code.

Alkarex:
@aledeg If you could remove - at least for now - the headers copy, address my other comment regarding `array_key_exists`, and keep the rest of the PR as it is, I think it would be a good compromise

aledeg:
@Alkarex done

Issue #2149, opened 2018-11-26
Suggestion: CSS Print Styling


sbrl:
When I use the share -> print option, it appears as if all the CSS is removed. It's be great f I could customise this via CSS, as then I could make it look awesome :D

By using the _CSS Paged Media_ standard, we can change the background colour of the page, add heads and footers, and more! I've got a [basic stylesheet](https://starbeamrainbowlabs.com/theme/basic.css) I've thrown together to experiment with this.

aledeg:
@sbrl I've added a way to transfer css files to the print page. That means you can use the CustomCss extension to do whatever you'd like to change the display of your printed articles.
You can give it a try!

PR #2374, opened 2019-04-16
Add theme to print page

aledeg:
Before, the printed page didn't have any usable CSS.
Now, it uses the css files available in the application. It means that
custom CSS can be add to target printed page.

See #2149

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -859,15 +859,11 @@ function init_stream(stream) {
 
 		el = ev.target.closest('.item.share > a[href="#"]');
 		if (el) {	//Print
-			const content = '<html><head><style>' +
-				'body { font-family: Serif; text-align: justify; }' +
-				'a { color: #000; text-decoration: none; }' +
-				'a:after { content: " [" attr(href) "]"}' +
-				'</style></head><body>' +
-				el.closest('.flux_content').querySelector('.content').innerHTML +
-				'</body></html>';
 			const tmp_window = window.open();
-			tmp_window.document.writeln(content);
+			for (var i = 0; i < document.styleSheets.length; i++) {
+				tmp_window.document.writeln('<link href="' + document.styleSheets[i].href + '" rel="stylesheet" type="text/css" />');
+			}
+			tmp_window.document.writeln(el.closest('.flux_content').querySelector('.content').innerHTML);
 			tmp_window.document.close();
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -859,13 +859,17 @@ function init_stream(stream) {
...
-				'a:after { content: " [" attr(href) "]"}' +
-				'</style></head><body>' +
-				el.closest('.flux_content').querySelector('.content').innerHTML +
-				'</body></html>';
+			let cssUrl = [];
+			for (var i = 0; i < document.styleSheets.length; i++){
+				cssUrl.push(document.styleSheets[i].href);
+			}
+			const cssLinkReducer = (carry, item) => `${carry}<link href="${item}" rel="stylesheet" type="text/css" />`;
+			const content = `
```
Alkarex: Arrow functions and ES6 template literals are not supported by IE11
https://caniuse.com/#feat=arrow-functions
https://caniuse.com/#feat=template-literals

aledeg: Damn it :'(

p/scripts/main.js
```diff
@@ -859,13 +859,17 @@ function init_stream(stream) {
 
 		el = ev.target.closest('.item.share > a[href="#"]');
 		if (el) {	//Print
-			const content = '<html><head><style>' +
-				'body { font-family: Serif; text-align: justify; }' +
-				'a { color: #000; text-decoration: none; }' +
-				'a:after { content: " [" attr(href) "]"}' +
```
Frenzie, line 865: You're removing this here, but I'd definitely want the basic principle of it over to an `@media print` block in the common base stylesheet then.

That's serif instead of sans-serif, black links instead of blue (but I'm not sure I agree with a lack of underline), and the full URL in  `a::after`, although I'd probably just use regular parentheses instead of brackets.

aledeg: There is already a ```@media print``` block in the common base! At the moment, it's useless because we don't use it.

Frenzie: I see, indeed. :-) It's used through the regular channels, just not through this share → print thing.

aledeg:
@Alkarex I removed what could not be used in IE11.

I think the next step will be to print without opening a new page. This will be a nice addition to the process. This will make the process less awkward.

PR #2388, opened 2019-04-29
Configurable amount of feeds that gets refreshed.

lx-s:
Enhanced `FreshRSS_feed_Controller::actualizeFeed` with a `maxFeeds`-parameter, so that a user can control how many feeds he wants to refresh instead of being forced to choose between "10 or all". `maxFeeds` is read from the `$_GET['maxFeeds']` variable and is set to a minimum of `10`.


Fake4d:
Added documentation for this pull in my pull request  #2389

Alkarex:
Thanks @lx-s !

PR #2390, opened 2019-04-29
Update 06_Fever_API.md

Fake4d:
Reeder-4 is new - Reeder-3 is the old version

BUT: 3 is free ;-)
And you have to pay for the new version.. So I dont know whether this should be linked here..

- Tested with new FreshRSS and Reeder-4


Alkarex:
It looks like Reeder-3 is not visible in the store anymore, is it?

PR #2399, opened 2019-05-30
Fix filter action display

aledeg:
Before, there was extra white lines between search terms when displaying
the data. It was not an issue for the storage since they were discarded
during the save process. But it was confusing.
Now, extra white lines are stripped.


Alkarex:
P.S. Sorry for the lack of activity: I have spent most of my free time on some work in my house, but it is almost finished :-)

aledeg:
No worries. I think we all agree that FRSS is not the only thing that matter. At some point, I was a bit worried though.

PR #2401, opened 2019-05-30
docker UID/GID environment settings

alfureu:
Would it be please possible to introduce a UID and GID environment settins, so one can set up the correct file and folder priviledges towards the host? Currently it uses 0 for root and 33 as group IDs. Setting these in the ENV would be ideal.


Alkarex:
Ups, this pull request is not doing what was intended...
You can either directly edit https://github.com/FreshRSS/FreshRSS/blob/dev/Docker/Dockerfile through the Web interface and follow the instructions, or use a normal workflow (see e.g. http://www.dasblinkenlichten.com/how-to-create-a-github-pull-request-pr/ )

Alkarex:
@DOFfactory Would you like to try sending this PR again?

Issue #2391, opened 2019-04-30
Allow @ sign in username


brknkfr:
As I am using an IMAP-Backend for authentication I want to be able to use mail addresses as usernames.

Using the cli interface it's no problem to add usernames with an @ sign, but unfortunately it's not allowed on the web interface.

Please change 

`const USERNAME_PATTERN = '[0-9a-zA-Z_][0-9a-zA-Z_.]{1,38}|[0-9a-zA-Z]';`

in

`app/Controllers/userController.php`

to something more useful like maybe

`const USERNAME_PATTERN = '[0-9a-zA-Z_][0-9a-zA-Z_.@]{1,38}|[0-9a-zA-Z]';`

(Added only an @ in pattern.)

Frenzie:
Also see #2062 and #1423 for previous relaxations. I can only imagine the username characters are so restrictive for compatibility reasons of some sort.

Looking at the history it used to be [ctype_alnum()](https://www.php.net/manual/en/function.ctype-alnum.php), so perhaps it was just because it's a simple, foolproof check. See https://github.com/FreshRSS/FreshRSS/commit/1ac09e7fe4a5408290d06116c6fb8152e018fe26.

> Using the cli interface it's no problem to add usernames with an @ sign, but unfortunately it's not allowed on the web interface.

Sounds like a bug. ;-) The checkUsername() method (or whatever its precise name is) should be used there too.

aledeg:
The problem is the pattern. When use in ```checkUsername``` it behave like if we are testing ```^0-9a-zA-Z_][0-9a-zA-Z_.]{1,38}``` or ```[0-9a-zA-Z]$```. I assume that what we want is ```^0-9a-zA-Z_][0-9a-zA-Z_.]{1,38}$``` or ```^[0-9a-zA-Z]$```.

See #1511 and #1512. @Alkarex what was the basic idea behind that fix? Did I missed something?

Alkarex:
@aledeg You are correct, that is a bug. I am testing a new pattern

Edit: https://github.com/FreshRSS/FreshRSS/pull/2407#issuecomment-502468456

Alkarex:
Merged in the /dev branch. Tests welcome

Issue #1511, opened 2017-04-27
Cannot login via Edge


GeoZzNL:
![image](https://cloud.githubusercontent.com/assets/12541266/25482738/50e67640-2b53-11e7-8f2a-7f51f5095a59.png)
![image](https://cloud.githubusercontent.com/assets/12541266/25482751/5a2dc41a-2b53-11e7-9e80-0be4e29a4b22.png)
![image](https://cloud.githubusercontent.com/assets/12541266/25482758/61e61996-2b53-11e7-8f77-b245455af4b8.png)

Logging is does not work in Microsoft Edge. Evey time i press a second letter, the bar turns red

Alwaysin:
Same here.

Error message:

```
You must use this format:
```

(yes, it is empty!)

I also have these errors in the console, which might be linked:

```
CSP14321: Resource violated directive 'default-src 'self'' in Content-Security-Policy: inline style, in http://<freshrss>/i/ at line 3 column 7. The resource will be blocked.
CSP14321: Resource violated directive 'default-src 'self'' in Content-Security-Policy: inline style, in http://<freshrss>/i/ at line 4 column 7. The resource will be blocked.
CSP14321: Resource violated directive 'default-src 'self'' in Content-Security-Policy: inline style, in http://<freshrss>/i/ at line 4 column 7. The resource will be blocked.
```

Alkarex:
I think the bug was introduced in version 1.6.3 https://github.com/FreshRSS/FreshRSS/pull/1423
Patch coming

Alkarex:
* http://stackoverflow.com/questions/22360235/do-browsers-support-different-html5-pattern-regexp-features
* https://connect.microsoft.com/ie/feedback/details/836117/regex-bug-in-pattern-validator

Alkarex:
Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1512
Please pull newest /dev branch and give it a try.
For the record, it is a IE / Edge bug.

GeoZzNL:
Thanks for the help

PR #2407, opened 2019-06-12
Allow email as username

aledeg:
Before, it was possible to register email as username on cli but not in the
interface. This was caused by a bug in the pattern which was not working as
expected. If your input was "user@example.com", the PHP verification was
catching only "user" and was acting like the whole thing was catched. But
on the interface, the catching was unsuccesful.
Now, the catching should be working properly.

I needed to add "$|^" in the pattern because without, I was catching either
the beginning of a string either the last char. This was introduced as a
workaround for IE/Edge pattern matching on April 27, 2017. See #1511 for
more information.

I tested it only on FF. Tests on other browsers wanted.

See #2391

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -38,7 +38,7 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 	 * The username is also used as folder name, file name, and part of SQL table name.
 	 * '_' is a reserved internal username.
 	 */
-	const USERNAME_PATTERN = '([0-9a-zA-Z_][0-9a-zA-Z_.@+-]{1,38}|[0-9a-zA-Z])';
+	const USERNAME_PATTERN = '([0-9a-zA-Z_][0-9a-zA-Z_.@-]{1,38}|[0-9a-zA-Z])';
 
 	public static function checkUsername($username) {
 		return preg_match('/^' . self::USERNAME_PATTERN . '$/', $username) === 1;
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -38,7 +38,7 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 	 * The username is also used as folder name, file name, and part of SQL table name.
 	 * '_' is a reserved internal username.
 	 */
-	const USERNAME_PATTERN = '[0-9a-zA-Z_][0-9a-zA-Z_.@]{1,38}$|^[0-9a-zA-Z]';
+	const USERNAME_PATTERN = '([0-9a-zA-Z_][0-9a-zA-Z_.@+-]{1,38}|[0-9a-zA-Z])';
 
 	public static function checkUsername($username) {
 		return preg_match('/^' . self::USERNAME_PATTERN . '$/', $username) === 1;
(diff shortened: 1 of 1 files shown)


Alkarex:
We also need to double-check with the 3 databases: SQLite, MySQL, PostgreSQL

aledeg:
I was thinking about it and we should even write the pattern completely. Meaning we should provide ```^[0-9a-zA-Z_][0-9a-zA-Z_.@]{1,38}$|^[0-9a-zA-Z]$```.

Frenzie:
Is there a specific reason to disallow a username like `@usern@m3@` if the `@` character is not problematic?

On the subject of e-mails itself, both of the suggested patterns disallow common e-mails with a `+` or `-` in the pattern, typically in the form of `aaa+bbb@ccc.ddd`. The whole of 
```
!#$%&'*+-/=?^_`{|}~
```
is valid.

This might seem like a minor semantic point (allow some e-mails vs. allow e-mails) but `+` and `-` are definitely quite popular in Western addresses, while UTF-8 is popular in the East, so I don't think it's nearly as minor as it might seem at first glance.

aledeg:
Keep in mind that the user has a folder on disk. There might be a limitation regarding what we can write on the file system.

Frenzie:
Only `/` is problematic for that purpose.

But anyway, I'm not necessarily arguing it should all be allowed, although I would like to see + and -. Just that it shouldn't be presented to the user as e-mails can be used as usernames.

aledeg:
Agreed. I'll modify the code to allow ```+``` and ```-``` signs. I think it's a good compromise.

Alkarex:
I have just suggested a new pattern, allowing `+` and `-`:

`([0-9a-zA-Z_][0-9a-zA-Z_.@+-]{1,38}|[0-9a-zA-Z])`

The `^` and `$` are automatically added by browsers, and we also do it in PHP.
It seems to work fine in Firefox, Chrome, Edge, IE11, and PHP.

We should double-check with the 3 database types, and check whether the username appears anywhere else potentially problematic.

In any case, we should not allow `/ \ : ? " * < > |` anywhere, not allow that the full username is `_` (used internally as default), not allow non-printable / control chars, etc.
Actually, we are not strict enough on Windows :-( https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

Alkarex:
Arg, the `+` gives a problem in the login (special character in URLs)

Alkarex:
I am (gladly) surprised to see that `@` in username even works with HTTP Basic Auth in IE11

Alkarex:
Seems to work fine in SQLite, MySQL, PostgreSQL.

Alkarex:
Let's add only `@-` for now, and potentially come back to more characters later if needed.
For Unicode usernames, we could consider using Punycode (which we already use in FreshRSS for international DNS). Otherwise we will have to transition to using some hashing.

Offerel:
I tried the dev branch and created a user with e-mail as login. However i cant login. I get 

`[Fri, 21 Jun 2019 10:55:37 +0200] [warning] --- Password mismatch for user=user@mydomain.de, nonce=, c=$2a$04$tTX99qYMYRJOYHew.zm4POdfg54647756745466A0y/Hy/Fz16`

The user directory is created as "usermydomainde". I can't check the database now, but i think, its the same there. Please take also in account that there can be several ASCII signs in the password, like `?$_!'"%.`... However, for this test, is uses only numbers and letters, so that the password should be no problem.

Update: The above is the result of creating the first time user during installation. If i create a user with e-mail after that, it seems to work.

Alkarex:
@Offerel Could you please open a new issue for that? It seems to be install-specific

Issue #2369, opened 2019-04-09
[Feature request] Add configuration on category level


aledeg:
We have configuration on the application level, we have configuration on the feed level. But nothing in between. It could be nice to have configuration on category level to limit the configuration on feed level when all feeds in a category have the same configuration.

At the moment, I can think of the archiving policy or the refresh rate.

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/2410

PR #2410, opened 2019-06-16
Change category configuration

aledeg:
Before, we had a drop-down list to interract on categories. It was not
working the same way as feeds.
Now, categories and feeds behave in a similar manner. At the moment,
there is no change in features but it will allow to expand them.

See #2369


aledeg:
Help on theme created with scss welcome!

app/i18n/en/sub.php
```diff
@@ -14,7 +14,9 @@
 		'_' => 'Category',
 		'add' => 'Add a category',
 		'empty' => 'Empty category',
+		'informations' => 'Information',
```
Frenzie: I believe it's all supposed to be English? ;-)

aledeg: I know! I just copied it from the feed section.

aledeg: I've changed everything to information!

Alkarex:
@aledeg While waiting for an SCSS fan, I suggest you just write the same than your other CSS versions for now, both in the SCSS and in the CSS.

aledeg:
@Alkarex After some headache, I figured out how to make sass work. :)

Alkarex:
Very nice :-)
I have just made some minor whitespace changes.

aledeg:
Thank you :blush:

PR #2409, opened 2019-06-16
Add Dockerfile links for PHP modules list

Sp3r4z:

PR #2413, opened 2019-06-16
Fix line break in table

Sp3r4z:
Better with HTML native `<br>` than `\\`


Alkarex:
Thanks, it looks like it is working 👍

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2409#issuecomment-502469989

Issue #2096, opened 2018-10-31
Multiple admins for a larger group of users


jfmessierdotca:
I implemented an instance of FreshRSS after a meeting of our local Linux user groups. Many participants were interested to use this service to aggregate news feeds. However, when creating the site (was not the first one, but the first for a larger user group than myself), I created myself by default, and became *the* administrator, instead of a more generic userID. Since this will be a shared userid/password with my friend who also manages the site, I had to create someone else with a generic "admin" username and a password, and manually update config.php. 

I would like to see something that could let more than one user be, become, or cease to be, an administrator, with the actual privileges granted by the system. I guess this would mean a different system of rights and flags, instead of a single identified username from the config.php file, as well as slightly different page for the admin when managing users, in that now a user can be granted (or denied) admin rights.

Thank You :-)

aledeg:
@jfmessierdotca Please check aforementioned PR to check if it suits your needs.

aledeg:
@jfmessierdotca the PR is available in the master branch. It will be released in the upcoming version.

PR #2417, opened 2019-06-21
Change user management page

aledeg:
I started working on a new user management page to be able to add more feature than what we currently have. Instead of having users in a dropdown list, I displayed user card with action directly available on them.

Keep in mind this is a work in progress. There is a lot more things I want to introduce in here.
I am just testing the waters to see if this could be a good idea or if I just need to drop it.
Test it and let me know what you think.


Alkarex:
Could you please post a screenshot to get an idea of the status so far?

aledeg:
@Alkarex I will later this weekend

aledeg:
Screen Shot 2019-06-21 at 23 39 46
@Alkarex here is a screenshot. It's far from being finished. But it's a first draft.
I want to style it better and add a pager to start with. Let me know if I should continue or not.

aledeg:
the blue layout is firefox selector :)

Alkarex:
@aledeg Ok, I see. While nice for a handful of users, I am a bit worried about how that can scale. We have some installations with many users. Such a page with 100+ users would probably be quite challenging to use, don't you think?

aledeg:
@Alkarex that's why I mentionned a pager. This way we can limit the number of displayed users. And still be able to browse the list.
My intention is to keep it fast while being able to view more than one user at the time.

marienfressinaud:
I'm just adding a note here because I'll need a redesign of the users list but I'm not sure this solution is the one I would go with. It is better than the current version for sure, but it takes a lot of useless space and a pager will generate a lot of pages.

I suggest to list users in a table, by showing only some useful information (such as username, the size of the DB and the email address) and at the end of each row, a link to access a page to see details/update the user. The form to change the password doesn't look important here, it's really not something an admin will do often and it would save us a form and 2 buttons by moving this feature on another page. I'm not even sure we need to keep a "delete" button on this page.

The table-style will allow to select several users at the same time and to perform actions by batch (such as delete them or deactivate their account… in a near future ;))

Let me know what you think about my suggestion and if you're still motivated to work on it. As I said, I'll need a redesign of this page so if you want to stop working on this, I can take over (but I don't know yet when I'll be able to work on it).

aledeg:
Thank you for your feedback. I can continue working on it but it won't happen overnight since I do have other things to do.

aledeg:
Here are the modifications I've made. It's far from finished but if you think that's a good starting point, I'll continue that way.

Screen Shot 2019-08-03 at 10 37 47
Screen Shot 2019-08-03 at 10 38 37

Let me know what you think.

Frenzie:
Looks pretty good to me. You can easily fit some 30 users in the space of 6 or maybe 8 in the other style, while simultaneously probably being clearer or at least easier to scan quickly. ;-)

One small thing I'd suggest visually is a little `tr:nth-child(odd)` background color.

aledeg:
I take note of that.
Keep in mind it's a WIP. I dign't even add the translation at this point.

marienfressinaud:
That's what I had in mind! :)

aledeg:
@Alkarex @marienfressinaud @Frenzie 
Please check this PR. If you think something is missing, let me know.
I didn't add a pager on the list for now. It's pretty fast for now. If it's needed, I'll add it later.

marienfressinaud:
I just had a quick look, it looks pretty good. Two comments:

- I don't see any use of the checkboxes in the user list, what about removing them?
- what about to change the order of the user list and the new user form? When the list will grow, the form will disappear from the view

aledeg:
Done

aledeg:
I've added a new column in the list and the missing translation entries.

I was wondering if it could be a good idea to add a user table to keep track of data. I was thinking about "last connection" for instance.
We could also store if the user is an admin or not (See #2096). We could also store the user configuration as-well.

marienfressinaud:
Nice!

About adding a user table, it could be an idea (it's already discussed in https://github.com/FreshRSS/FreshRSS/issues/2504). I still have no strong opinion about it. However, I think you should not work on it in this specific PR: there's already enough to review :)

I'll be happy to merge it quick enough, I think it'll become soon a very important feature to me ;)

aledeg:
I won't add more things in that PR. I think it's a good starting point for the user management page. On top of my mind, we need to add the last connection, the credential (admin or not), a pager.
We'll see where it goes from here.

aledeg:
I've just noticed that I forgot to make a real commit message :)
Don't merge that PR before I change it.

marienfressinaud:
Don't worry, I didn't review your PR yet specifically because your commit was "wip" ^^

Also, I think we should merge this PR after releasing the version 1.15.1 (hopefully today)

aledeg:
Good! I'll change the commit message today.

aledeg:
@marienfressinaud Now we have a real commit message :)

app/i18n/en/gen.php
```diff
@@ -16,6 +16,7 @@
 		'manage' => 'Manage',
 		'mark_favorite' => 'Mark as favourite',
 		'mark_read' => 'Mark as read',
+		'purge' => 'Purge',	// TODO - Translation
```
marienfressinaud:
```suggestion
		'purge' => 'Purge',
```

app/views/user/details.phtml
```diff
@@ -0,0 +1,48 @@
...
+                    <a class="btn toggle-password" data-toggle="password-<?= $this->username; ?>"><?= _i('key') ?></a>
+                </div>
+                <?= _i('help'); ?> <?= _t('admin.user.password_format') ?>
+            </div>
+        </div>
+        <div class="form-group form-actions">
+            <noscript><b><?= _t('gen.js.should_be_activated'); ?></b></noscript>
+            <button type="submit" class="btn btn-important" name="action[update]"><?= _t('gen.action.update') ?></button>
+            <button type="submit" class="btn btn-attention confirm" name="action[purge]"><?= _t('gen.action.purge') ?></button>
+            <button type="submit" class="btn btn-attention confirm" name="action[delete]"><?= _t('gen.action.remove') ?></button>
```
marienfressinaud: What about `name="action" value="delete"`?

app/views/user/details.phtml
```diff
@@ -0,0 +1,48 @@
...
+            <label class="group-name" for="newPasswordPlain"><?= _t('admin.user.password_form') ?></label>
+            <div class="group-controls">
+                <div class="stick">
+                    <input type="password" id="newPasswordPlain" name="newPasswordPlain" autocomplete="new-password" pattern=".{7,}" <?= cryptAvailable() ? '' : 'disabled="disabled" ' ?>/>
+                    <a class="btn toggle-password" data-toggle="password-<?= $this->username; ?>"><?= _i('key') ?></a>
+                </div>
+                <?= _i('help'); ?> <?= _t('admin.user.password_format') ?>
+            </div>
+        </div>
+        <div class="form-group form-actions">
```
marienfressinaud, line 42: A blank line before this one? :)

app/Controllers/userController.php
```diff
@@ -534,4 +549,26 @@ public function deleteAction() {
 
 		Minz_Request::forward($redirect_url, true);
 	}
+
+	public function detailsAction() {
+		if (!FreshRSS_Auth::hasAccess('admin')) {
+			Minz_Error::error(403);
+		}
+
+		$username = Minz_Request::param('username');
```
marienfressinaud, line 571: We should check that the username exists, or return a 404

app/views/user/details.phtml
marienfressinaud: From a UI point of view, I would separate all these actions in different forms with explanations of what will happen by clicking on buttons. But it's a thinking we should have for the overall interface so there's no need to change it here.

app/Controllers/userController.php
```diff
@@ -186,6 +185,18 @@ public function profileAction() {
 		}
 	}
 
+	public function purgeAction() {
+		if (!FreshRSS_Auth::hasAccess('admin')) {
+			Minz_Error::error(403);
+		}
+
+		if (Minz_Request::isPost()) {
+			$username = Minz_Request::param('username');
```
marienfressinaud, line 202: same here, we should check that the user exists

marienfressinaud (changes requested):
Pretty neat PR!

aledeg:
@marienfressinaud Done! Ready to be merged

app/views/user/details.phtml
```diff
@@ -0,0 +1,49 @@
...
+                </div>
+                <?= _i('help'); ?> <?= _t('admin.user.password_format') ?>
+            </div>
+        </div>
+
+        <div class="form-group form-actions">
+            <noscript><b><?= _t('gen.js.should_be_activated'); ?></b></noscript>
+            <button type="submit" class="btn btn-important" name="action" value="update"><?= _t('gen.action.update') ?></button>
+            <button type="submit" class="btn btn-attention confirm" name="action" value="purge"><?= _t('gen.action.purge') ?></button>
+            <button type="submit" class="btn btn-attention confirm" name="action" value="delete"><?= _t('gen.action.remove') ?></button>
```
marienfressinaud: Oups, I missed it yesterday: can you put these buttons inside a `<div class="group-controls">` so there are aligned with the input?

aledeg: done!

marienfressinaud (approved):
It's good for me :+1:

Alkarex:
Sorry for the delay - I finally took the time to test. Looks great 👍 
Maybe an e-mail field could be added somewhere

PR #2424, opened 2019-06-21
jQuery 3.4.1

Alkarex:
Only used for statistics graphs


aledeg:
I don't know if those stats are making any sense. Should we keep them?

Frenzie:
They're pretty cool though. :-)

Alkarex:
Although they are pretty slow to generate, and could be better, I do look at them once in a while personally, and back when the feature was added, there were positive responses

aledeg:
I am just asking :) I never use them even though I wrote the whole thing.

PR #2430, opened 2019-06-28
Add content in share process

aledeg:
I've recently discovered that we can transmit the description directly
to Shaarli. So I modified the sharing process to include the content
in the sharing link.

### DIFF
--- a/app/views/helpers/index/normal/entry_bottom.phtml
+++ b/app/views/helpers/index/normal/entry_bottom.phtml
@@ -87,6 +87,7 @@
 						}
 						$share_options['link'] = $link;
 						$share_options['title'] = $title;
+						$share_options['content'] = $this->entry->content();
 						$share->update($share_options);
 				?><li class="item share">
 					<?php if ('GET' === $share->method()) {?>
--- a/app/Models/Share.php
+++ b/app/Models/Share.php
@@ -65,9 +65,6 @@ class FreshRSS_Share {
 		return self::$list_sharing[$type];
 	}
 
-	/**
-	 *
-	 */
 	private $type = '';
 	private $name = '';
 	private $url_transform = '';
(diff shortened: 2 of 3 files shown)


Frenzie:
Can there be a danger that it'll be too long unless cut off at some point?

aledeg:
Possibly

Frenzie:
Then I propose a maxlength substring or equivalent. ;-)

aledeg:
I'll modify this PR to include that kind of feature.

app/views/helpers/index/normal/entry_bottom.phtml
```diff
@@ -87,6 +87,7 @@
 						}
 						$share_options['link'] = $link;
 						$share_options['title'] = $title;
+						$share_options['content'] = $this->entry->content();
```
Alkarex, line 90: Hum, we have a risk of duplicating the article content many times here, right?
I think this should be replaced by a dynamic menu, either server side or client side

aledeg: I am not sure I follow your reasoning. Could you explain it?

Alkarex: Yes: I am worried about the output size. If I am not mistaken, the HTML output will contain up to several times the content of each article, considerably increasing the size of the page

aledeg: Ok, I understand all the concerns. I think it's better if we keep it the way it is. I am closing this PR.

Alkarex: Ok. Do you see any problem with a dynamic menu (either client side or server side)?

aledeg: I think that if one want the content, he will copy and paste it as I am doing at the moment. I don't think that spending time on that is necessary.

Alkarex: Ok

PR #2436, opened 2019-07-06
Docker build for ARM (Raspberry Pi)

Alkarex:
Using QEMU for building the image on Docker Hub, then native execution.
Inspired by https://github.com/stas-demydiuk/domoticz-docker

PR #2440, opened 2019-07-08
Fix ARM ca-certs

Alkarex:
Workaround strange bug: When building on Docker Hub with QEMU, the
symlinks for the CA certificates are not created properly, although it
is the case when building for ARM locally.
Manually calling `update-ca-certificates -f` fixes the problem.
Note for later: Maybe we should call this update in our entrypoint.sh.


Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/2436

Issue #2381, opened 2019-04-21
Error 403 when a new user registers


Dryusdan:
Hello :)

New user can't registers on my FreshRSS instance.
After test, I encounter this error : 
```
Error 403 - Forbidden

You don’t have permission to access this page [CSRF]
```

But registration module is native with freshrss right ? So I don't forgot update it when I update frashrss ? :sweat_smile: 

So I have FreshRSS version 1.14.2 with Swage theme.

Can you help me ? :)

Regards

wangz1yu:
I also encountered the same situation. @Alkarex

aledeg:
I've made some tests. I can reproduce. I search the code to find where this is triggered. At the moment, I don't know where. It's located deep in Minz because the process does not even enter the called method.
So far I am out of luck!

Alkarex:
@aledeg Could you please tell the sequence to reproduce?

aledeg:
@Alkarex it's very easy. You increase the number of allowed user to be greater than the current number of user. Then you logout and try to create a new user from the login page. And then 403!

Alkarex:
Ah, right. @Dryusdan and @wangz1yu : do you believe it is the same issue you are facing?
In which case, we should probably just change the UI to e.g. remove the registration field when there are too many users already.

Dryusdan:
I'm not remember if I increase number of allowed user.
This value is set to 150 (at the start of my instance there is 100 user allowed). But I have only 41 users registed so... The registeration form does appear.

![Capture d’écran du 2019-06-21 09-17-15](https://user-images.githubusercontent.com/4410431/59905207-ca392200-9405-11e9-8fae-23ef1fba7bb8.png)

I try with 0 in allow user (unlimited so) and I have the same error.

aledeg:
@Alkarex I don't think it's a UI error. You cannot register new users even though there are some spots left.

Alkarex:
Ah, indeed, I can reproduce too

Alkarex:
Here is a candidate fix https://github.com/FreshRSS/FreshRSS/pull/2442
Tests welcome

Alkarex:
Merged in the /dev branch. Tests welcome

Dryusdan:
It's work, thank you ! :D

PR #2290, opened 2019-03-22
Rework CSRF interaction with sessions

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2288
Improve security in some edge cases
Maybe relevant for
https://github.com/FreshRSS/FreshRSS/issues/2125#issuecomment-474992671

PR #2442, opened 2019-07-14
Fix user self registration

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2381
Related to https://github.com/FreshRSS/FreshRSS/pull/2290


Alkarex:
Hum, we probably need to have CSRF for the admin, but not for the self-registration case. Will fix later

PR #2443, opened 2019-07-14
Default to SQLite

Alkarex:
Move SQLite as first option in install instead of MySQL. Lower barrier to entry.


Frenzie (approved):
The impact on barrier to entry in and of itself seems fairly low to me (obviously not needing user/pass and all that makes it easier), but the ease of portability could also be considered as part of that.

Issue #2446, opened 2019-07-18
PHP / FastCGI Installation Problem


rnc:
I installed the application through cpanel which gave me version 1.14.2.

I have been following https://freshrss.github.io/FreshRSS/en/users/06_Mobile_access.html to setup the system. However on running the API test I got:
```
Google Reader API configuration test:
    ❌ FAIL get HTTP Authorization header! Wrong Web server configuration.
```
From reading https://docs.holodyn.com/KnowledgeBase/Hosting/Apache/PHP_FastCGI_Authentication_Headers and https://www.drupal.org/project/restws/issues/2101361  I tried experimenting with adding to a `.htaccess` file. 
Upon adding
```
RewriteCond %{HTTP:Authorization} ^(.+)
RewriteRule ^(.*)$ $1 [E=HTTP_AUTHORIZATION:%1,PT]
```
(from the above links) I got a pass. Unfortunately I don't know which directory needed the headers so have added it in the top level. 
* Where is best to add it?
* Could this be supported internally if this is detected?

Alkarex:
Hello,
There is already a .htaccess included, which is supposed to address your case:
https://github.com/FreshRSS/FreshRSS/blob/master/p/api/.htaccess
Please check that it is there

rnc:
It is indeed there! I removed the above code from the top level and appended it to that file - and I still get a pass. So I guess the tests aren't working in my case.

@Alkarex  I am not that familiar with htaccess conditionals but would be happy to help debug.

rnc:
@Alkarex  Having looked into this a bit more I've found the following references:

* https://httpd.apache.org/docs/2.4/en/mod/core.html#cgipassauth
* https://stackoverflow.com/questions/17018586/apache-2-4-php-fpm-and-authorization-headers
* https://stackoverflow.com/questions/26549250/apache-strips-down-authorization-header

Experimenting more with the `.htaccess` I found that 
```
	SetEnvIfNoCase "Authorization" "(.*)" HTTP_AUTHORIZATION=$1
```
works while 
```
	SetEnvIfNoCase "^Authorization$" "(.*)" HTTP_AUTHORIZATION=$1
```
does not for me. Further simply adding 
```
CGIPassAuth on
``` 
also works fine. If either of those two solutions are acceptable I'd be happy to send a PR

Frenzie:
[...]

PR #2449, opened 2019-07-20
Issue #2446 : Fix passing authentication headers

rnc:
…version is high enough

### DIFF
--- a/p/api/.htaccess
+++ b/p/api/.htaccess
@@ -1,9 +1,9 @@
 <IfModule mod_setenvif.c>
-   SetEnvIfNoCase "Authorization" "(.*)" HTTP_AUTHORIZATION=$1
+	SetEnvIfNoCase "^Authorization$" "(.*)" HTTP_AUTHORIZATION=$1
 </IfModule>
 <IfModule !mod_setenvif.c>
-   <IfModule mod_rewrite.c>
-           RewriteEngine on
-           RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
-   </IfModule>
+	<IfModule mod_rewrite.c>
+		RewriteEngine on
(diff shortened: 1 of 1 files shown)


p/api/.htaccess
```diff
@@ -1,9 +1,14 @@
...
-	SetEnvIfNoCase "^Authorization$" "(.*)" HTTP_AUTHORIZATION=$1
-</IfModule>
-<IfModule !mod_setenvif.c>
-	<IfModule mod_rewrite.c>
-		RewriteEngine on
-		RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
-	</IfModule>
-</IfModule>
+<IfVersion >= 2.4.13>
+    CGIPassAuth on
```
Alkarex: This might generate an error 500 if users do not have the `AllowOverride` rights needed (`AuthConfig`, which is less commonly granted than `FileInfo`)

Alkarex:
I do not know a way to test that `CGIPassAuth on` can be safely used (testing the version is not enough).
I am a bit hesitant, but I think that only simplifying the rule `SetEnvIfNoCase "Authorization"` would be better (although I would still be interested in knowing why the current expression does not work in your case).

rnc:
@Alkarex  Thats fine - to be honest I've never really used htaccess before so this has been a learning experiment :-) I'll update the PR to remove it.

Alkarex:
Merged in the /dev branch. Would you be able to test?

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2446

rnc:
@Alkarex I am using that change at the moment on my server. In terms of testing it, ideally I think a container image (i.e. Docker) but using litespeed instead of Apache might be suitable. However, while I have had a quick play I haven't got litespeed working yet.

PR #2454, opened 2019-07-22
[CI] Run shellcheck and shfmt

Frenzie:
Cf. https://github.com/FreshRSS/FreshRSS/pull/2436#discussion_r305640019

<hr>

I wrote this script years ago, long before these tools became a standard part of Travis. ShellCheck is a good idea on any script; shfmt has fewer obvious benefits given how few we have and how small they are but I'd say why not?

### DIFF
--- a/.ci/helper_shellchecks.sh
+++ b/.ci/helper_shellchecks.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# Based on https://github.com/koreader/koreader/blob/master/.ci/helper_shellchecks.sh
+
+# shellcheck disable=2016
+mapfile -t shellscript_locations < <({ git grep -lE '^#!(/usr)?/bin/(env )?(bash|sh)' && git submodule --quiet foreach '[ "$path" = "base" -o "$path" = "platform/android/luajit-launcher" ] || git grep -lE "^#!(/usr)?/bin/(env )?(bash|sh)" | sed "s|^|$path/|"' && git ls-files ./*.sh; } | sort | uniq)
+
+SHELLSCRIPT_ERROR=0
+
+for shellscript in "${shellscript_locations[@]}"; do
+	echo -e "${ANSI_GREEN}Running shellcheck on ${shellscript}"
+	shellcheck "${shellscript}" || SHELLSCRIPT_ERROR=1
+	echo -e "${ANSI_GREEN}Running shfmt on ${shellscript}"
+	if ! shfmt -i 4 "${shellscript}" >/dev/null 2>&1; then
+		echo -e "${ANSI_RED}Warning: ${shellscript} contains the following problem:"
+		shfmt -i 4 "${shellscript}" || SHELLSCRIPT_ERROR=1
+		continue
+	fi
+	if [ "$(cat "${shellscript}")" != "$(shfmt "${shellscript}")" ]; then
+		echo -e "${ANSI_RED}Warning: ${shellscript} does not abide by coding style, diff for expected style:"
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -16,6 +16,7 @@ script:
   - phpenv rehash
   - find . -not -path "./lib/JSON.php" -name \*.php -print0 | xargs -0 -n1 -P4 php -l 1>/dev/null 2>php-l-results
   - if [ -s php-l-results ]; then cat php-l-results; exit 1; fi
+  - bash .ci/helper_shellchecks.sh
```
Alkarex: What about moving this script to `./tests/` instead of `./.ci/` ?

.ci/helper_shellchecks.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# Based on https://github.com/koreader/koreader/blob/master/.ci/helper_shellchecks.sh
+
+# shellcheck disable=2016
+mapfile -t shellscript_locations < <({ git grep -lE '^#!(/usr)?/bin/(env )?(bash|sh)' && git submodule --quiet foreach '[ "$path" = "base" -o "$path" = "platform/android/luajit-launcher" ] || git grep -lE "^#!(/usr)?/bin/(env )?(bash|sh)" | sed "s|^|$path/|"' && git ls-files ./*.sh; } | sort | uniq)
```
Alkarex: Consider splitting this long line

Frenzie: Better yet, I didn't read through the script in detail since it was a bit of a quick breakfast commit and the part about submodules is irrelevant in this repo.

tests/shellchecks.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+# Based on https://github.com/koreader/koreader/blob/master/.ci/helper_shellchecks.sh
+
+mapfile -t shellscript_locations < <({ git grep -lE '^#!(/usr)?/bin/(env )?(bash|sh)' && git ls-files ./*.sh; } | sort | uniq)
+
+SHELLSCRIPT_ERROR=0
+
+for shellscript in "${shellscript_locations[@]}"; do
+	echo -e "${ANSI_GREEN}Running shellcheck on ${shellscript}"
```
Alkarex, line 13: Is it needed to restore the original colour at the end of the script and/or in case of error / interruption?

Frenzie: You mean for local use?

Alkarex: Yes (indeed, if the shell exits, there is no need)

tests/shellchecks.sh
```diff
@@ -0,0 +1,29 @@
...
+ANSI_GREEN="\\033[32;1m"
+ANSI_RESET="\\033[0m"
+
+mapfile -t shellscript_locations < <({ git grep -lE '^#!(/usr)?/bin/(env )?(bash|sh)' && git ls-files ./*.sh; } | sort | uniq)
+
+SHELLSCRIPT_ERROR=0
+
+for shellscript in "${shellscript_locations[@]}"; do
+	echo -e "${ANSI_GREEN}Running shellcheck on ${shellscript}"
+	shellcheck "${shellscript}" || SHELLSCRIPT_ERROR=1
```
Alkarex, line 14: It could be good to add a note in about this test in https://github.com/FreshRSS/FreshRSS/blob/dev/tests/README.md , and also that it requires installing `shellcheck` and `shfmt`

PR #2451, opened 2019-07-21
FreshRSS 1.14.3

Alkarex:
Hello,
Before some more substantial changes scheduled for 1.15.0, let's [release a version 1.14.3](https://github.com/FreshRSS/FreshRSS/milestones/1.14.3) before the summer vacations 🏖️ with a number of small improvements and bug fixes. In particular, there is now a Docker image automatically built for ARM (e.g. Raspberry Pi) testable with [`freshrss/freshrss:dev-arm`](https://hub.docker.com/r/freshrss/freshrss/tags).

Thanks to the 12 contributors, especially newcomers @rosiel , @lx-s , @Sp3r4z , @Fake4d , @SuperSandro2000, and @rnc :-)

Full [changelog](https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md):

* Bug fixing
	* Fix wrong mark-as-read limit [#2429](https://github.com/FreshRSS/FreshRSS/issues/2429)
	* Fix API call for removing a category [#2411](https://github.com/FreshRSS/FreshRSS/issues/2411)
	* Fix user self-registration [#2381](https://github.com/FreshRSS/FreshRSS/issues/2381)
	* Make CGI Authorization configuration for API more compatible [#2446](https://github.com/FreshRSS/FreshRSS/issues/2446)
	* Fix refresh icon in Swage theme [#2375](https://github.com/FreshRSS/FreshRSS/issues/2375)
	* Fix message banner in Swage theme [#2379](https://github.com/FreshRSS/FreshRSS/issues/2379)
	* Docker: Add `php-gmp` for API support in Ubuntu 32-bit [#2450](https://github.com/FreshRSS/FreshRSS/pull/2450)
* UI
	* New configuration page for each category [#2369](https://github.com/FreshRSS/FreshRSS/issues/2369)
	* Update shortcut configuration page [#2405](https://github.com/FreshRSS/FreshRSS/issues/2405)
	* CSS style for printing [#2149](https://github.com/FreshRSS/FreshRSS/issues/2149)
	* Do not hide multiple `<br />` tags [#2437](https://github.com/FreshRSS/FreshRSS/issues/2437)
	* Updated to jQuery 3.4.1 (only for statistics page) [#2424](https://github.com/FreshRSS/FreshRSS/pull/2424)
* Deployment
	* Docker: Add automatic health check [#2438](https://github.com/FreshRSS/FreshRSS/pull/2438)
[...]

PR #2438, opened 2019-07-07
Docker healthcheck+ labels

Alkarex:
And try to fix the platform metadata for ARM, and add health check.
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/2436

PR #2455, opened 2019-07-22
Replace curl healthcheck with pure php, apply haodlint suggestions

### LINKED ISSUE #2453: [Docker] replace curl healtcheck with php/wget
Curl pulls in quite some dependencies we don't really need. wget or pure php is way lighter on that.
Pure php could work like this:
````
php -r 'if(!file_get_contents(\"http://localhost/api/fever.php\")){exit(1);}'
````

SuperSandro2000:
Closes #2453

Adding ``-o pipefail`` to the shell ensures that errors will fail pipechains and forward errors correctly.
Apt is an end user tool and scripts are recommended to use ``apt-get``. See https://github.com/hadolint/hadolint/wiki/DL3027

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -1,11 +1,12 @@
 FROM ubuntu:19.04
 
 ENV TZ UTC
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apt update && \
-	apt install --no-install-recommends -y \
-	ca-certificates cron curl \
+RUN apt-get update && \
+	apt-get install --no-install-recommends -y \
+	ca-certificates cron \
 	apache2 libapache2-mod-php \
 	php-curl php-gmp php-intl php-mbstring php-xml php-zip \
 	php-sqlite3 php-mysql php-pgsql && \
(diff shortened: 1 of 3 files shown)


Alkarex:
Thanks 👍

Alkarex:
It could be nice to add https://github.com/hadolint/hadolint as part of our automatic testing, e.g. in Travis or Docker builds. What do you think @Frenzie ?
P.S.: https://github.com/hadolint/hadolint/blob/master/docs/INTEGRATION.md

Frenzie:
@Alkarex 
> It could be nice to add https://github.com/hadolint/hadolint as part of our automatic testing, e.g. in Travis or Docker builds.

Sure, I wasn't aware of it. Seems generally useful.

However, that `curl=1.1.0` will break the script long before a change in the output from `apt install` would. :-)

Alkarex:
I also noticed it :-) But it was just an example they used, I guess

Alkarex:
@SuperSandro2000 or @Frenzie , if one of you is motivated enough to add haodlint to our CI, that would be welcome :-)

Frenzie:
@SuperSandro2000 @Alkarex Maybe later today or tomorrow, not right now.

Alkarex:
@SuperSandro2000 wrote https://github.com/FreshRSS/FreshRSS/pull/2451#issuecomment-513811647
> Btw I saw you wanted to add hadolint.
> I would suggest this config https://github.com/SuperSandro2000/docker-images/blob/master/.hadolint.yaml of me and this install script of me https://github.com/SuperSandro2000/docker-images/blob/master/.travis.yml#L58 https://github.com/SuperSandro2000/docker-images/blob/master/.travis.yml#L61 .

That looks good at a first glance 👍 Would you mind sending a PR with that?

Frenzie:
Yeah, that looks basically perfect except we'll want something closer to `git ls-files --exclude='*Dockerfile*' --ignored` or you'll exclude `Docker/Dockerfile-Alpine` and `Docker/Dockerfile-QEMU-ARM`.

SuperSandro2000:
Done in #2456

Alkarex:
Cross-reference https://github.com/FreshRSS/FreshRSS/pull/2438

PR #2456, opened 2019-07-22
[CI] Run hadolint

SuperSandro2000:
As requested [here](https://github.com/FreshRSS/FreshRSS/pull/2455#issuecomment-513809772)

I don't like that it needs ``jq`` but I am to lazy to write that json select with bash. It is also probably more robust and easier to understand this way.

This would also require to add a GItHub Token under ``$GITHUB_TOKEN`` in Travis. If we leave this out the CI periodically fails cause GitHub API gets rate limited quite easily.
The URL would then change to:
``https://api.github.com/repos/hadolint/hadolint/releases/latest?access_token="$GITHUB_TOKEN"``

I am a bit unsure where to put it or if it should be blocking for the other CI jobs. Let me know if I should move it to another section.

### DIFF
--- a/.hadolint.yaml
+++ b/.hadolint.yaml
@@ -1,0 +1,7 @@
+ignored:
+  # ignore apt version pinning
+  - DL3008
+  # ignore pip version pinning
+  - DL3013
+  # ignore apk version pinning
+  - DL3018
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ script:
   - phpenv rehash
   - find . -not -path "./lib/JSON.php" -name \*.php -print0 | xargs -0 -n1 -P4 php -l 1>/dev/null 2>php-l-results
   - if [ -s php-l-results ]; then cat php-l-results; exit 1; fi
-  - bash tests/shellchecks.sh
   - |
     if [[ $VALIDATE_STANDARD == yes ]]; then
       COMPOSER_BIN=$(composer global config --absolute bin-dir)
@@ -50,6 +49,7 @@ matrix:
         - curl -sLo "$HADOLINT" $(curl -s https://api.github.com/repos/hadolint/hadolint/releases/latest?access_token="$GITHUB_TOKEN" | jq -r '.assets | .[] | select(.name=="hadolint-Linux-x86_64") | .browser_download_url') && chmod 700 ${HADOLINT}
       script:
         - node_modules/jshint/bin/jshint .
+        - bash tests/shellchecks.sh
         - git ls-files --exclude='*Dockerfile*' --ignored | xargs --max-lines=1 "$HADOLINT"
   allow_failures:
     - env: CHECK_TRANSLATION=yes VALIDATE_STANDARD=no
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,14 @@ matrix:
         - "node"
       php:
         # none
+      env:
+        - HADOLINT="$HOME/hadolint"
       install:
         - npm install jshint
+        - curl -sLo "$HADOLINT" $(curl -s https://api.github.com/repos/hadolint/hadolint/releases/latest?access_token="$GITHUB_TOKEN" | jq -r '.assets | .[] | select(.name=="hadolint-Linux-x86_64") | .browser_download_url') && chmod 700 ${HADOLINT}
       script:
         - node_modules/jshint/bin/jshint .
+        - git ls-files --exclude='*Dockerfile*' --ignored | xargs --max-lines=1 "$HADOLINT"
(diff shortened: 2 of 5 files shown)


Frenzie:
Could you please also fix the issues it points out?
```
Docker/Dockerfile:55 DL3025 Use arguments JSON notation for CMD and ENTRYPOINT arguments
Docker/Dockerfile-Alpine:51 DL3025 Use arguments JSON notation for CMD and ENTRYPOINT arguments
Docker/Dockerfile-QEMU-ARM:67 DL3025 Use arguments JSON notation for CMD and ENTRYPOINT arguments
```

Regarding placement, if you test for something like `[ -z ${HADOLINT+x} ];` you can do something similar to the VALIDATE_STANDARD/CHECK_TRANSLATION stuff and combine it with the shellchecks script. Or it's probably better to also add something like `LINT=yes` for clarity.

SuperSandro2000:
I changed it to only run in one job cause everything else is kinda pointless. The Dockerfiles don't change with the php version.

About fixing that issue: It would require us to move the CMD to it's own file cause I don't know a way to use ``(`` with the json notation.

Or we just ignore it for now.

How do we wan't to handle the GitHub Token? I don't really want to hardcode one.

.travis.yml
```diff
@@ -33,6 +33,13 @@ env:
 matrix:
   fast_finish: true
   include:
+    - env:
+      - HADOLINT="$HOME/hadolint"
+      install:
+      - sudo apt-get install jq
+      - curl -sLo "$HADOLINT" $(curl -s https://api.github.com/repos/hadolint/hadolint/releases/latest | jq -r '.assets | .[] | select(.name=="hadolint-Linux-x86_64") | .browser_download_url') && chmod 700 ${HADOLINT}
+      script:
+      - git ls-files --exclude='*Dockerfile*' --ignored | xargs --max-lines=1 "$HADOLINT"
```
Frenzie: > I changed it to only run in one job cause everything else is kinda pointless. The Dockerfiles don't change with the php version.

I wasn't suggesting to run it on every job. On second look I'd move this and the shellchecks into the jshint node_js container down below to make it the dedicated non-PHP lint job.

.travis.yml
```diff
@@ -33,6 +33,13 @@ env:
 matrix:
   fast_finish: true
   include:
+    - env:
+      - HADOLINT="$HOME/hadolint"
+      install:
+      - sudo apt-get install jq
```
Frenzie: Use

```yml
apt:
  packages:
    - jq
```

Frenzie:
> Or we just ignore it for now.

Yes, in that case it's probably clearest to stick a `# hadolint ignore=DL3025` above the relevant lines.

Frenzie:
@SuperSandro2000 I added a GITHUB_TOKEN variable to Travis with no scopes. Is that correct?

.travis.yml
Frenzie: Actually scratch that. jq is part of the standard Travis image.

.travis.yml
SuperSandro2000: Good idea and already done.

SuperSandro2000:
> @SuperSandro2000 I added a GITHUB_TOKEN variable to Travis with no scopes. Is that correct?

Yes, as we only use it for the API we don't need any special permission. Thanks.

.travis.yml
SuperSandro2000: I quickly looked at this job https://travis-ci.org/FreshRSS/FreshRSS/jobs/562212239 and for me it looks like that jq isn't installed by default.

Frenzie: Line 55 of the log. ;-)

Screenshot_2019-07-23_07-47-34

SuperSandro2000: 🤔 seems legit but why did I install it then?

SuperSandro2000:
This should be done.

Frenzie:
Awesome, thanks!

.hadolint.yaml
```diff
@@ -0,0 +1,7 @@
+ignored:
+  # ignore apt version pinning
+  - DL3008
+  # ignore pip version pinning
+  - DL3013
```
Alkarex, line 5: Next time we update, we should remove this line. We do not use `pip`

SuperSandro2000: Oh, you are totally right.

SuperSandro2000: But you already did it in https://github.com/FreshRSS/FreshRSS/commit/61af33a6f58e15d88a08765046bbaf08be7dd49e
Thanks!

Alkarex: I have just fixed it :-)

Issue #2460, opened 2019-07-25
Run on Apache 2.4+ without mod_access_compat


koocotte:
http://httpd.apache.org/docs/2.4/en/upgrading.html#access

> In Apache 2.2, access control based on client hostname, IP address, and other characteristics of client requests was done using the directives Order, Allow, Deny, and Satisfy.
> 
> In Apache 2.4, such access control is done in the same way as other authorization checks, using the new module mod_authz_host. The old access control idioms should be replaced by the new authentication mechanisms, although for compatibility with old configurations, the new module mod_access_compat is provided.

Currently, to deny access to some directories, we have Apache 2.2 directives. With Apache 2.4, we need to load mod_access_compat. The old directives are error-prone and using the module can hide errors elsewhere; so a good practice for admin with Apache 2.4 is to update directives and unload this compatibility module.

I suggest to update corresponding .htaccess files to support both Apache 2.2 (with Order...) and Apache 24 (with Require all denied)

I expect to submit a PR shortly.

Alkarex:
That sounds good 👍 
It would be good to keep a back compatibility for Apache 2.2 users (e.g. Red Hat / CentOS 6) if possible

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2461

PR #2461, opened 2019-07-25
Patch for #2460: Run on Apache 2.4+ without mod_access_compat

koocotte:
Update corresponding .htaccess files:

Test for mod_authz_core and use directives according.

### DIFF
--- a/app/.htaccess
+++ b/app/.htaccess
@@ -1,3 +1,11 @@
-Order	Allow,Deny
-Deny	from all
-Satisfy	all
+# Apache 2.2
+<IfModule !mod_authz_core.c>
+	Order	Allow,Deny
+	Deny	from all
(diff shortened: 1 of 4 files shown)


Alkarex:
Thanks 👍 Could you also please update https://github.com/FreshRSS/FreshRSS/blob/dev/Docker/FreshRSS.Apache.conf (in that one, we know it will only be Apache 2.4.38+, so no need of back-compatibility).

koocotte:
> Thanks 👍 Could you also please update https://github.com/FreshRSS/FreshRSS/blob/dev/Docker/FreshRSS.Apache.conf (in that one, we know it will only be Apache 2.4.38+, so no need of back-compatibility).

This file is not using compatibility. No change need.

PR #2464, opened 2019-07-26
Remove Google+ support

ghost:
Google+ has been dead for a while now.


Frenzie:
This looks like only the translation strings and not the code? Or do you mean this is a leftover from a previous cleanup?

aledeg:
It's only the translation strings. There are other places were it's in use. In ```app/shares.php``` for instance.

ghost:
> It's only the translation strings. There are other places were it's in use. In `app/shares.php` for instance.

I have included app/shares.php, but I didn't find anything other than that and translation strings.

https://github.com/FreshRSS/FreshRSS/pull/2464/files#diff-29a558e3108b5864daf75d27247eba30

Frenzie (approved):
lgtm

aledeg:
I think you need to rebase your branch since there is an extra commit in your PR!

ghost:
Sorry, I don't know how it crawled in. I have removed that.

PR #2467, opened 2019-07-28
Allow to change the view layout

marienfressinaud:
This PR allows to change the view layout dynamically. It could be used to separate layout code of login/registration, feeds, configuration, etc. My main need is to develop an extension to provide a custom home page: I want to customize the header for this specific page but I can't without this change.

### DIFF
--- a/lib/Minz/View.php
+++ b/lib/Minz/View.php
@@ -9,11 +9,11 @@
  */
 class Minz_View {
 	const VIEWS_PATH_NAME = '/views';
-	const LAYOUT_PATH_NAME = '/layout';
-	const LAYOUT_FILENAME = '/layout.phtml';
+	const LAYOUT_PATH_NAME = '/layout/';
+	const LAYOUT_DEFAULT = 'layout';
 
 	private $view_filename = '';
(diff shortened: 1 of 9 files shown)


lib/Minz/View.php
```diff
@@ -137,11 +140,15 @@ public function helperToString($helper) {
 	}
 
 	/**
-	 * Permet de choisir si on souhaite utiliser le layout
-	 * @param $use true si on souhaite utiliser le layout, false sinon
+	 * Choose the current view layout.
+	 * @param $layout the layout name to use, false to use no layouts.
 	 */
-	public function _useLayout ($use) {
```
Alkarex: I believe someone had built an extension doing something similar (custom view) but I cannot immediately find a reference, and I cannot remember which calls were used. Probably this one, in which case this change would break it. Maybe you could keep an additional legacy `_useLayout()` calling `_layout()`

marienfressinaud: good catch! Just updated the PR to bring back a deprecated `_useLayout` function. I also added a `Minz_Log::warning` to warn about the deprecation.

PR #2476, opened 2019-08-02
tec: Provide a Minz_Mailer class

marienfressinaud:
This class can be inherited to create Mailers dedicated to some specific tasks. You can have an example of how to use it with what I've done to send validation emails ([here](https://github.com/FreshRSS/FreshRSS/pull/2481/files#diff-deb887810d825219aa851b2d6725473dR365-R366) and [here](https://github.com/FreshRSS/FreshRSS/pull/2481/files#diff-4724fb696fbf025630cb1584f63db582R1-R31)). It is inspired by how Rails works with emails.

My main concern with this code is that the calls to the SMTP server need to be finished before redirecting the user. We have no systems of parallel queues. But it may be not critical yet and we can go with it for the moment… ?

I also needed to change the `Minz_View` API a bit by replacing the `change_view` method by a `_path` one. It is now `Minz_Request`-agnostic and can support files different than the `phtml` ones.

To merge in `--no-ff`

### DIFF
--- a/lib/Minz/Mailer.php
+++ b/lib/Minz/Mailer.php
@@ -1,0 +1,117 @@
+<?php
+
+use PHPMailer\PHPMailer\PHPMailer;
+use PHPMailer\PHPMailer\Exception;
+
+require LIB_PATH . '/PHPMailer/PHPMailer.php';
+require LIB_PATH . '/PHPMailer/Exception.php';
+require LIB_PATH . '/PHPMailer/SMTP.php';
+
+/**
+ * Allow to send emails.
+ *
+ * The Minz_Mailer class must be inherited by classes under app/Mailers.
+ * They work similarly to the ActionControllers in the way they have a view to
+ * which you can pass params (eg. $this->view->foo = 'bar').
+ *
+ * The view file is not determined automatically, so you have to select one
+ * with, for instance:
+ *
+ * ```
+ * $this->view->_path('user_mailer/email_need_validation.txt')
+ * ```
+ *
+ * Minz_Mailer uses the PHPMailer library under the hood. The latter requires
+ * PHP >= 5.5 to work. If you instantiate a Minz_Mailer with PHP < 5.5, a
+ * warning will be logged.
+ *
+ * The email is sent by calling the `mail` method.
+ */
+class Minz_Mailer {
+	/**
+	 * The view attached to the mailer.
+	 * You should set its file with `$this->view->_path($path)`
+	 *
+	 * @var Minz_View
+	 */
+	protected $view;
+
+	/**
+	 * Constructor.
+	 *
+	 * If PHP version is < 5.5, a warning is logged.
+	 */
+	public function __construct () {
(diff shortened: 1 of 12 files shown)


Frenzie (commented):
Looks pretty straightforward. Btw, phpcs has a couple of minor formatting complaints. ;-)

marienfressinaud:
Indeed! I just disabled the linter on the `PHPMailer` lib since it's not our code, let's see how it goes (edit: it goes fine :))

marienfressinaud:
For some reason, when PHPMailer has a problem, it creates a file under `p/i` folder :/ I need to investigate that…

marienfressinaud:
Fixed, it's just that I don't know how to concatenate two strings in PHP :D (I used a `,` instead of a `.` when calling the `Minz_Log::error` method).

Frenzie:
It's probably because with `echo()` you can just pass multiple strings separated by commas, and concatenating them first would be less efficient. (Although presumably seldom in a way that'd be remotely noticeable.)

jkinable:
@marienfressinaud what's the status of this PR? Is it ready to be merged into dev?

lib/Minz/Mailer.php
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Constructor.
+	 *
+	 * If PHP version is < 5.5, a warning is logged.
+	 */
+	public function __construct () {
+		if (version_compare(PHP_VERSION, '5.5') < 0) {
+			Minz_Log::warning('Minz_Mailer cannot be used with a version of PHP < 5.5.');
+		}
```
Alkarex, line 47: Should we maybe return and fail here?

lib/Minz/Mailer.php
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * The view file is not determined automatically, so you have to select one
+ * with, for instance:
+ *
+ * ```
+ * $this->view->_path('user_mailer/email_need_validation.txt')
+ * ```
+ *
+ * Minz_Mailer uses the PHPMailer library under the hood. The latter requires
+ * PHP >= 5.5 to work. If you instantiate a Minz_Mailer with PHP < 5.5, a
```
Alkarex, line 25: We should add a note in the main readme(s) + documentation

lib/Minz/Mailer.php
marienfressinaud: Well, in fact I think it *might* work if PHPMailer is configured to use the `mail` function. I just don't want to support this case officially, but I'm ok to let a possibility to admins to force the feature despite the warnings.

lib/Minz/Mailer.php
marienfressinaud: I added documentation in [the other PR](https://github.com/FreshRSS/FreshRSS/pull/2481/files#diff-058bbce9aa164d0e426a3e742165070e) since this one is just technical (there is no impact for admins yet)

PR #2481, opened 2019-08-06
Provide email address verification feature

### LINKED ISSUE #2469: Adding email verification support
I'd like to add features to help administrators to manage multi-users instances. I'm currently working on an "email verification" feature but I'm facing an annoying problem.

Context: I need to send emails via SMTP and not using localhost. The only built-in function in PHP to send emails is `mail`, which needs to be configured [via the `php.ini` file](https://www.php.net/manual/en/mail.configuration.php) and needs sendmail to be configured. I'm not even sure we can pass credentials ([some say no, some others say yes but not documented](https://stackoverflow.com/questions/112190/php-ini-smtp-how-do-you-pass-username-password)).

Almost everyone recommends to use [PHPMailer](https://github.com/PHPMailer/PHPMailer) and I think we should go with it. Problem: PHPMailer only supports PHP 5.5+, while FreshRSS works with PHP 5.3+ (according to our [`travis.yml` file](https://github.com/FreshRSS/FreshRSS/blob/master/.travis.yml)). My questions are:

- should we re-evaluate our minimal requirements? Even if some distributions still support these old versions, maybe it's time to move on? (it doesn't mean we will break everything without proper warning to the administrators)
- if the answer to the previous question is "no", is a solution with "no email validation feature" for the older PHP versions would be ok for you?
- or maybe we can use "PHPMailer" for PHP 5.5+ and `mail` for older versions? (but it means more work for me)

Follow-up of my progress:

- [x] add an email field on the profile page and the CLI (reuse the existing `mail_login` config element)
- [x] add a `force_email_validation` element to system config (`false` by default) and add a field to the system config page
- [x] ask for email during registration if `force_email_validation` is `true`
- [x] create a page to validate the user email address (params: `username` and `token`)
- [x] send an email with the URL of the validation page when email is changed (or user create its account)
- [x] block access to FreshRSS if the `email_validation_token` is not `null`
- [x] allow user to resend the validation email
- [x] allow user to change its email while she's blocked
- [x] document the feature
- [x] (nice to have) improve the look of the "validation feedback page" (maybe using https://github.com/FreshRSS/FreshRSS/pull/2467)

(more elements to come while I'm working on it)

marienfressinaud:
This PR adds the support for email address verification.

It is based on #2476 (**so this one needs to be merged first**) and better reviewed commit by commit.

Closes #2469 

## Nice to have

These are improvements that are not "necessary" but could be nice to have. I added the reason why I didn't implement these in parenthesis:

- [ ] give easy access to validation url in development mode, when SMTP isn't configured (useful?)
- [ ] rename `mail_login` into `email` so it's more intuitive (need to migrate current configuration)
- [ ] add `force_email_validation` to the CLI (useful?)
- [x] change profile layout when email is not validated (more work than expected)

Depending on your feedback, I can implement them in this PR, in other PRs, or not at all ;)

## Basic workflow

An email field is added to the profile page.

If the `force_email_validation` boolean is set to `true`:

- the field also appears during registration

Capture d’écran de 2019-08-06 14-33-52

- an email to validate the user is sent when the email address changes (including registration)

Capture d’écran de 2019-08-06 14-35-21

- FreshRSS is blocked on a page which asks the user to validate its email address

Capture d’écran - 2019-08-06 à 14 34 39

- the user can ask to resend the email or access the "profile" page (the other pages are blocked)

Capture d’écran - 2019-08-06 à 14 34 53

- once the user clicks on the link, its account is enabled

Capture d’écran - 2019-08-06 à 14 35 44


Frenzie (commented):
Offer to resend as well?

Edit: d'oh, next commit.

marienfressinaud:
I made a bunch of improvements and added a bit of documentation. I think it's now ready to be reviewed, but it's still blocking by #2476 (from which the two first commits come). Also added the "nice to have" improvements that I didn't implement in the first message.

app/Controllers/userController.php
```diff
@@ -36,12 +36,23 @@ public static function deleteFeverKey($username) {
...
 	}
 
-	public static function updateUser($user, $passwordPlain, $apiPasswordPlain, $userConfigUpdated = array()) {
+	public static function updateUser($user, $email, $passwordPlain, $apiPasswordPlain, $userConfigUpdated = array()) {
 		$userConfig = get_user_configuration($user);
 		if ($userConfig === null) {
 			return false;
 		}
 
+		if ($email !== null && $userConfig->mail_login !== $email) {
```
Alkarex, line 45: `$email != ''` would be safer here, especially since it might come from `Minz_Request::param('email', '');`

app/Controllers/userController.php
```diff
@@ -36,12 +36,23 @@ public static function deleteFeverKey($username) {
...
+			$userConfig->mail_login = $email;
```
Alkarex, line 46: Somewhere we need to sanitize this user input. Either here or in `profileAction()`.
I suggest looking at https://php.net/filter.filters.validate `FILTER_VALIDATE_EMAIL / FILTER_SANITIZE_EMAIL`

app/Controllers/userController.php
marienfressinaud: That was on purpose since the email should not be required when `force_email_validation` is `false`. But I realize that I should enforce the value if it's `true`.

app/Controllers/userController.php
marienfressinaud: I would simply check that the email contains a `@`. Maybe we can strip the string too. But emails are hard to validate and it's almost always badly done (more [here](https://hackernoon.com/the-100-correct-way-to-validate-email-addresses-7c4818f24643)). And I'm not comfortable with that:

>  In general, this validates e-mail addresses against the syntax in RFC 822, with **the exceptions that comments and whitespace folding and dotless domain names are not supported.**

I don't understand why they would respect the RFC, but in fact not…

The feature itself is about validating that the address is valid by sending an email so I don't think we have to be too strict.

marienfressinaud:
Just to let you know @Alkarex, I saw you merged `dev` into my branch but I prefer to rebase my branches to keep the history clean (easier to read, easier to find a bug in a specific commit)

app/Controllers/userController.php
Frenzie: That's definitely *very* odd about dotless domains. The whitespace & comments things is more common. For example, the W3C recommends [this](https://www.w3.org/TR/html51/sec-forms.html#email-state-typeemail) for input type=email:

> This requirement is a willful violation of RFC 5322, which defines a syntax for e-mail addresses that is simultaneously too strict (before the "@" character), too vague (after the "@" character), and too lax (allowing comments, whitespace characters, and quoted strings in manners unfamiliar to most users) to be of practical use here.

```
/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
```

Alkarex: There are still some classes of characters, which should be discarded, and potentially also a min and max length. I think it is dangerous to pass some user input to the backend without any form of sanitization. If you do not like the PHP built-in validation (which I personally think would be reasonable - I an not even sure the rest of the back-end is able to handle what the PHP validation rejects anyway), then at least the sanitize filter `FILTER_SANITIZE_EMAIL` https://php.net/manual/filter.filters.sanitize

Alkarex:
@marienfressinaud I just used GitHub's built-in tool to resolve the conflict :-)
In general, I think forced pushes and other forms of history rewriting should be kept to rare occasions to fix errors. They make collaboration quite difficult. For instance, if I merge this branch in my own `Alkarex/dev` branch to test it (where I also simultaneously test other PRs), and your branch history is rewritten, then I cannot pull your new changes anymore and there is a big work to do to get a working branch again (potentially having to reset my test branch and re-merge all the ones I want to test at the same time). Even when testing a single branch, a rewritten history is annoying as it breaks simple pulls to get newer updates.
It also breaks reviews, make some commits disappear (e.g. my conflict resolution is gone), make some comments on old commits disappear, breaks the ability to see the changes since the last review, etc.
Finally, I do not think the clean history is an issue at all: all git clients that I know distinguish commits made in the active branch from commits made in other branches (e.g. merged), and we can then *squash & merge* when the PR is ready anyway :-)

Frenzie:
Btw, perhaps you already know but I recently discovered that on GitHub you can do this:

```bash
# assuming your own fork is origin and the one here is upstream
git fetch upstream pull/2481/head
git checkout FETCH_HEAD
```

marienfressinaud:
That's interesting and I strongly agree with some of your arguments. In the mean time, I consider my branches as my personal workspaces and nobody should base their work on it nor modify them. In this context, force-push should not be a problem. I would not recommend to test different work-in-progress branches at the same time neither, but I can understand that can be useful.

I'm not a huge fan of squash because you lose how the feature has been constructed. I always try to build my PRs logically and they are reviewable commit by commit. I personally prefer a merge with no fast-forward to keep the history. It can also help a lot to understand why and how a bug got introduced.

In the mean time, GitHub is definitely not adapted to this workflow: as you said, comments are lost, it's difficult to see the last changes, commits are not ordered correctly in the "Commits" tab, etc.

Considering that, I'll try an "in-between": I'll commit my fixes with `commit --fixup=<commit to fix>` and don't rebase until the PR is approved. I'll have to refrain myself because I worked this way for the three last years ^^ I found it powerful and it helped me to understand better how Git works and how to fix an error… but it's definitely advanced usage of Git and collaboration is more important than some technical tricks :)

app/Controllers/userController.php
marienfressinaud: Ok, I'll take a look to clean at least a bit this param!

Alkarex:
It is not a big deal in any case :-) A bit more feedbacks:
Regarding the *squash & merge*, the history is kept in the PR itself, together with the discussions, reviews, etc.
Rewriting history makes a branch more difficult to test (`pull` is not working anymore), even when you are the only one making changes.
Rewriting history also makes the exact construction progress disappear (e.g. missing commits, out of context commits).
By the way, when you submit the PR, you can decide whether you allow changes to your branch by others or not.

Alkarex:
* [ ] It looks like updating the e-mail field on `?c=user&a=profile` leads to a 404
* [ ] I get `Got error 'PHP message: Invalid address:  (From): noreply@localhost\n'` (I have other PHP systems such as Friendica, which are sending e-mails fine on that machine)
* [ ] The page `?c=user&a=validateEmail` does not seem to indicate that there was an error when sending the e-mail
* [ ] Once we are on `?c=user&a=validateEmail` , it is not possible to change user (log-out does not work, home page does not work)

app/Controllers/userController.php
marienfressinaud: `FILTER_SANITIZE_EMAIL` doesn't support internationalized email addresses :( (which [are valid](https://en.wikipedia.org/wiki/Email_address#Internationalization)) For instance, it removes accentuated characters such as `é` or `à`. One solution would be to convert the email to punycode with `idn_to_ascii` and then back to utf8 but the PHP internationalisation module needs to be installed and I'm not even sure it will be enough. `FILTER_SANITIZE_EMAIL` seems no longer relevant to the today standards. The only thing that I'm quite sure is [the limit of 254 characters](https://stackoverflow.com/a/574698) and I can trim the value too…

Frenzie: @marienfressinaud I noticed that in the regex I posted, and curiously Firefox indeed complains if I try a perfectly normal character like é. They've been with us as a proper standard for some 15 years and have been well supported for over a decade…

Btw, I noticed in the PHPMailer changelog:

> Simplification of address validation - now uses PHP's FILTER_VALIDATE_EMAIL pattern by default, retains advanced options

Those advanced options might be worth investigating? Otherwise it seems like we're talking about a difference without distinction if we end up with FILTER_VALIDATE_EMAIL anyway.

@Alkarex I disagree, just about any backend can handle more than what PHP rejects for Reasons™. Also see [here](https://flameeyes.blog/2018/09/30/software-systems-and-institutional-xenophobia/) about the more general problem.

Frenzie: @marienfressinaud 
https://github.com/PHPMailer/PHPMailer/blob/6c74e4b33f815928dac7b230148bc0a4d1a136d5/src/PHPMailer.php#L1227-L1250

The `pcre(8)` filter looks especially interesting:
https://github.com/PHPMailer/PHPMailer/blob/6c74e4b33f815928dac7b230148bc0a4d1a136d5/src/PHPMailer.php#L1266-L1281

Alkarex: I agree that we should do some reasonable efforts to support most valid e-mail addresses. But bad sanitizing is literally the number one security mistake (top 1 OWASP) https://www.owasp.org/index.php/Top_10-2017_A1-Injection 
I think that doing it properly (which is indeed difficult) is precisely the job of FILTER_VALIDATE_EMAIL (which should be improved - example of another problem https://bugs.php.net/bug.php?id=53722 -) but if it is not good enough for the time being, then we could at least check that it is valid UTF-8.

Frenzie: What do (in)valid emails have to do with SQL injection? Not only is that what `mysqli::real_escape_string()` and friends are for (we use PDO here for prepared statements), but <code>`</code> is one of the original valid email characters to boot! I suppose SQL injection might potentially still be averted due to PHP's incorrect quoting behavior (related to whitespace and ;) but it'd be pretty bizarre to depend on that imho.

> but if it is not good enough for the time being, then we could at least check that it is valid UTF-8.

Sure, sounds like a good idea if we don't do it yet (doesn't setting the DB/connection to UTF-8 in PDO like we do take care of this all by itself?), but the connection to email seems a bit tenuous.

Any backend will validate (and reject) email addresses on better criteria, won't it?

Alkarex: The link is not only for SQL injections, and is an illustration for a major class of problems caused by improper input validation. The point is that except by testing / verifying the rest of the pipeline on different platforms and with different backends, one cannot be sure of where the proper validation will be done.

Alkarex: Example https://www.cvedetails.com/cve/CVE-2002-1337/
> Buffer overflow in Sendmail 5.79 to 8.12.7 allows remote attackers to execute arbitrary code via certain formatted address fields, related to sender and recipient header comments as processed by the crackaddr function of headers.c. 

CVSS Score 10.0 (Critical)

Alkarex: Another example https://www.cvedetails.com/cve/CVE-2003-0694/
> The prescan function in Sendmail 8.12.9 allows remote attackers to execute arbitrary code via buffer overflow attacks, as demonstrated using the parseaddr function in parseaddr.c. 

And https://www.cvedetails.com/cve/CVE-2003-0161/
> The prescan() function in the address parser (parseaddr.c) in Sendmail before 8.12.9 does not properly handle certain conversions from char and int types, which can cause a length check to be disabled when Sendmail misinterprets an input value as a special "NOCHAR" control value, allowing attackers to cause a denial of service and possibly execute arbitrary code via a buffer overflow attack using messages, a different vulnerability than CVE-2002-1337.

Alkarex: As @Frenzie found, let's use one of the options proposed by PHPMailer. The HTML5 version looks reasonable to me and would have the advantage of being consistent with the input field type in the HTML front-end as well (edit: which Marien already does, by the way).

Alkarex: Related https://bugs.php.net/bug.php?id=73176
> Sanitize/Validate Filters on URL don't work for international domains

Alkarex:
@marienfressinaud I did not explain why I sometimes have multiple PRs in my own test branch: when I submit a PR, I merge it immediately in my own Alkarex/dev branch, which I run on a few test servers, including my main instance (which I use intensively). Likewise, when there is a PR of significance submitted by someone else, I do the same. In both cases, I let them there for a little while (depending on how potentially problematic they are) before merging, with the hope of spotting any bug / problem before merging the PR, ensuring that our FreshRSS/dev branch is OK-stable. It is therefore quite frequent to have more than one PR to test at a time. If any of these PR rewrites its history, it breaks my test branch :-)

app/Controllers/userController.php
Frenzie: @Alkarex I wasn't suggesting to perform no input validation in any case. But checking if something is valid UTF-8, for example, I just can't see an email related argument in there. :-) Same for a length check of some sort; it's just that 254 chars is specific to email.

What I am saying is that the PHP validation (and also the HTML5 one) results in a tremendous quantity of false-positive rejections. The impact of dropping accents is perhaps relatively small for most of us Europeans, but for Cyrillic, Turkish, the East Asian scripts... well, they're not happy about it. ;-)

I'd say the main advantage of the built-in PHP pattern is that 7.x/8.x will hopefully improve on it soon.

Alkarex: Ah, I did not realise that the HTML5 version does not allow UTF-8 before the `@` 😢

Alkarex: So, it seems that the conclusion is to use the PHPMailer's `pcre8` option as @Frenzie suggested higher up

marienfressinaud: Thanks for digging into the problem! (I was finishing my holidays :)) I didn't thought that PHPMailer could have a verification function (sounds obvious now!) I'll try to finish this tomorrow :+1:

marienfressinaud: It looks like none of the patterns allow internationalized email addresses (not even `pcre8`). I'm also very surprised that Firefox or Thunderbird don't allow UTF-8 before the `@` either… Is there any server or client that fully support UTF-8 addresses?!

I found that PHPMailer provides a `punyencodeAddress` function which convert only the domain, that's a start. So I'll convert the address to punycode first and validate the address against this value. Then I can either store the value as punycode (but I'll have to transform it again when displaying the address) or store the utf-8 version (easier, and it should be ok since PHPMailer transforms all the addresses before sending the email)

Frenzie: > Is there any server or client that fully support UTF-8 addresses?!

Sure, Microsoft has announced support, this one does http://www.postfix.org/SMTPUTF8_README.html

There's an incomplete list on Wikipedia:
https://en.wikipedia.org/wiki/Extended_SMTP#SMTPUTF8

Edit: to be clear, that list is from like 4 years ago and even then I doubt it was complete.

marienfressinaud:
Normally, email addresses are now validated (I used `html5` pattern to match with the browser).

Concerning your four points @Alkarex, I think I fixed the first (404 error): it happened when `force_email_validation` was `false`. I'll need more information for the second: how is it configured? `smtp` or `mail` mailer? Is it inside a Docker container?

I'll take a look to the last feedback tomorrow.

Alkarex:
@marienfressinaud Tested on native Ubuntu with default PHP, postfix. Here is the relevant `php.ini` section (should be the same as the default one):

```ini
[mail function]
; For Win32 only.
; http://php.net/smtp
SMTP = localhost
; http://php.net/smtp-port
smtp_port = 25

; For Win32 only.
; http://php.net/sendmail-from
;sendmail_from = me@example.com

; For Unix only.  You may supply arguments as well (default: "sendmail -t -i").
; http://php.net/sendmail-path
;sendmail_path =

; Force the addition of the specified parameters to be passed as extra parameters
; to the sendmail binary. These parameters will always replace the value of
; the 5th parameter to mail().
;mail.force_extra_parameters =

; Add X-PHP-Originating-Script: that will include uid of the script followed by the filename
mail.add_x_header = Off

; The path to a log file that will log all mail() calls. Log entries include
; the full path of the script, line number, To address and headers.
;mail.log =
; Log mail to syslog (Event Log on Windows).
;mail.log = syslog
```

I think you just need to specify a `from` address

marienfressinaud:
You can set the `from` in the `data/config.php` configuration file. I don't know what would be a valid address, maybe `root@localhost`?

```php
...
  'smtp' => array (
    'hostname' => '',
    'host' => 'localhost',
    'port' => 25,
    'auth' => false,
    'auth_type' => '',
    'username' => '',
    'password' => '',
    'secure' => '',
    'from' => 'root@localhost',
  ),
...

**Comment by Alkarex** [comment]
That should be good. Otherwise, we do have the `base_url` from which we could extract a domain to provide a better default.

**Comment by marienfressinaud** [comment]
I'll need to finish this PR one day! I think there is not so much to do:

- [x] reorganize i18n keys (not happy with my choices)
- [x] finish to translate in English and French
- [x] add the keys to other locales
- [x] change the layout of the profile page when user is blocked (I found an easy solution to my problem)
- [x] check email format during creation as well
- [x] let admin to specify email during user creation

I still have a problem with the feedback if email is not sent: it was easy to add on `sendValidationEmailAction`, but the code is not well organized in `profileAction` and I don't know how it could be added without reorganizing a lot of code (which is more work). I think we should go (for the moment) with the current solution which implies:

- no feedback if email is not sent (when address is changed on the profile page)
- but feedback is given if email is not sent when clicking on the "Resend" button

Tell me if you see anything else that I should do for this PR!

**Comment by Alkarex** [comment]
That sounds good @marienfressinaud . Just let us know when you would like another round of testing / reviewing

**Comment by marienfressinaud** [comment]
I think it's all good now!

File: app/Controllers/userController.php [hunk 3]
```diff
@@ -275,6 +338,122 @@ public static function deleteUser($username) {
...
+		}
+
+		if (!FreshRSS_UserDAO::exists($username) || $user_config === null) {
+			Minz_Error::error(404);
+		}
+
+		if ($user_config->email_validation_token === '') {
+			Minz_Request::good(
+				_t('user.email.validation.feedback.unnecessary'),
+				array('c' => 'index', 'a' => 'index'),
```
```suggestion
				array('c' => 'index', 'a' => 'index')
```

Alkarex: Syntax error, and more of the same type below (also caught by Travis)

Alkarex: This seems to be OK from PHP 7.3+, which I did not know

Frenzie: New in 7.3, cool. :-)

Alkarex: https://wiki.php.net/rfc/trailing-comma-function-calls

Alkarex:
And still have the following error, preventing any e-mail from being sent:
```
[error] --- Minz_Mailer cannot send a message: Invalid address:  (From): root@localhost
AH01071: Got error 'PHP message: Invalid address:  (From): root@localhost\n'
```

P.S.: No difference between `root` and `noreply`. But changing for `postmaster@[.....].fr` worked

Alkarex:
It is because PHPMailer calls PHP's FILTER_VALIDATE_EMAIL which does not accept `@dotless`

Alkarex:
I suggest adding:
```php
PHPMailer::$validator = 'pcre8';
```
Just after
https://github.com/flusio/FreshRSS/blob/2cde74db503adf581ece09a17924da378edcb660/lib/Minz/Mailer.php#L77

Alkarex:
With the syntax fix, and the validator fix, it should be good for me.

marienfressinaud:
I used `html5` validator, like in the rest of the code. It should work now!

Alkarex (approved):

Purexo:
Uh,

I see on fediverse that FreshRSS has been updated, so i the update from web admin interface has alway !

1. after update (complete I hope) redirect to `https://<my instance>/i/?c=error`
2. on network navigator debugger many `https://<my instance>/i/?c=user&a=validateEmail`
3. I try to log with my email instead username, classic error login page
4. I connect to my VPS, configs seems OK
5. Little search on github issue / pr I fall here
6. Inspect some code and found this line : https://github.com/flusio/FreshRSS/blob/a3031d21bf7ffbeae10683ad333dc5c355fa09bb/app/Controllers/userController.php#L128
7. check my account config file : no key `email_validation_token` in this file (not found in example user config file too but i'm on a terminal maybe i don't see it)
8. So i add this key with `''` value
9. Yeah i'm able to connect

I recommend you to use php empty function instead strongly typed equality :
```php
$email_not_verified = !empty(FreshRSS_Context::$user_conf->email_validation_token);
```
Check https://www.php.net/manual/en/function.empty.php
maybe if you support very old php version you should store `FreshRSS_Context::$user_conf->email_validation_token` in a variable and use this var in empty

More safe in context of checking config that can be from very old version of app my instance running without update issue from many years (4-5years I think)

At this time, I'm happy to be a Developper or I will not hase solve my issue without help ^^.

Best regards

Alkarex:
@Purexo Yes, this line is indeed bad, especially the strict inequality. Could you please send a pull request to fix it? I have not checked much the e-mail flow ( ping @marienfressinaud  ), but couldn't you log with your normal username? Please open another issue if that was not the case.
Even when the parameters are not in your own account config file, default parameters are supposed to be loaded from https://github.com/FreshRSS/FreshRSS/blob/dev/config-user.default.php so I am not entirely sure of what the exact problem was.

marienfressinaud:
I confirm that the `email_validation_token` should be taken from the default user config if not set in your own config file. I also checked that `config-user.default.php` is updated during a ZIP update (it [should](https://github.com/FreshRSS/update.freshrss.org/blob/master/scripts/update_util.php#L186-L187) but maybe it failed?) Could you post the content of this file to check if it's actually up-to-date?

marienfressinaud:
Side note about `empty` function: it's probably what should be used here in PHP, but I'm really annoyed with this kind of solution since it basically says: I don't really know what's the value of this variable. Here, it would even have hidden a real bug.

Alkarex:
When some data comes from somewhere that is not fully controlled, the code should allow for some typical variations. In this case, I do not think it makes sense to strictly test for an empty string and fail for the other nullable values. In this case, I would argue for `!= ''`. Not `empty()` though, because the value comes back from a wrapper (which might return null):

https://github.com/FreshRSS/FreshRSS/blob/03208771f7d38771f7b9c965b79877d97cce311a/lib/Minz/Configuration.php#L139-L163

(Which I think we should remove at some point in favour of native code, but that is another story :-))

Purexo:
> but couldn't you log with your normal username? Please open another issue if that was not the case.

I was not able to connect with my username nor email.
Since I've add `email_validation_token` in my user config file, I am able to connect with my username. Not with my email but I don't care.
I think when tried to login with username before modify the config, the form works well, and I was connected but the system was checking `validateEmail` rule, the rules did not passed so redirect (badly infinite, Firefox was detect the redirect loop and abort it.)
Should I create an issue ?

> Even when the parameters are not in your own account config file, default parameters are supposed to be loaded from https://github.com/FreshRSS/FreshRSS/blob/dev/config-user.default.php

I've looked into the `config-user.default.php` file (on my server): no `email_validation_token` key in.

Yes I've update with zip (available in web-admin interface), it work well from many years and really convenient.

I can make a fast PR from github if you wan't. Should I checkout from `dev` branch ?

Alkarex:
@Purexo Yes please, /dev branch 👍

PR #2483, opened 2019-08-08
Docker readme cleanups

SuperSandro2000:


Frenzie (commented):
None of the wording on that page suggests it's recommended to add the user to the docker group though, just that it's something you can do for your own convenience.

SuperSandro2000:
> None of the wording on that page suggests it's recommended to add the user to the docker group though, just that it's something you can do for your own convenience.

And something most people probably have cause it is just way more convenient than removing it from every command or typing in your password all the time.

Frenzie (approved):
Of course I use it this way on my local dev machine for convenience; whether it's necessarily a good idea to go for the convenience on a server I'm not sure but I'll defer to a second opinion by e.g. @Alkarex or @aledeg on that one.

PS https://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel/

SuperSandro2000:
We don't promote it anywhere in the docs to add your user to the docker group. We just assume the probably most common case and if your setup differs you need to adjust the commands to your setup.

Docker/README.md
```diff
@@ -153,17 +153,16 @@ Note that prebuilt images are less recent and only available for x64 (Intel, AMD
 # First time only
 git clone https://github.com/FreshRSS/FreshRSS.git
 
-cd ./FreshRSS/
+cd FreshRSS/
 git pull
-sudo docker pull ubuntu:18.10
```
Alkarex, line 158: When building the image locally, pulling the base image makes sure it is up to date.

SuperSandro2000: Maybe this variant of the build command ensures we don't need to update the version number at two places and there is only one truth.
````
docker build --help
.....
--pull                    Always attempt to pull a newer version of the image
````

So I would change the following line to:
````
docker build --pull --tag freshrss/freshrss -f Docker/Dockerfile .
````

Is this okay for you?

Comment (Alkarex) [reply Docker/README.md:158]: Yes, that would be even better 👍

**Comment by SuperSandro2000** [comment]
The hadolint job fails cause it misses the GitHub token. https://travis-ci.org/FreshRSS/FreshRSS/jobs/572382545

Not sure how to fix that.

File: .gitignore [hunk 2]
```diff
@@ -15,3 +15,4 @@ constants.local.php
 *.autosave
 *.sass-cache
 sass.sh
+Docker/key.gpg
```
Alkarex: What is this key for?

Frenzie:
> The hadolint job fails cause it misses the GitHub token. https://travis-ci.org/FreshRSS/FreshRSS/jobs/572382545

Travis makes sure the token doesn't leak in the log.

.gitignore
SuperSandro2000: Apparently it was my key and I have no idea how it ended up their. I though it was a key from the build process or something related to FreshRSS. Gonna drop that commit.

SuperSandro2000:
> > The hadolint job fails cause it misses the GitHub token. [travis-ci.org/FreshRSS/FreshRSS/jobs/572382545](https://travis-ci.org/FreshRSS/FreshRSS/jobs/572382545)
> 
> Travis makes sure the token doesn't leak in the log.

Yeah I know but all PRs are going to be red and hadolint won't lint it properly. I am unsure how to solve that problem.

Alkarex:
Thanks :-)

PR #2492, opened 2019-08-16
Add Makefile and Docker configuration for development

marienfressinaud:
It is a simple configuration which support only SQLite database for the moment. It just setup an environment with Docker so developers can start working in just a few commands.

Let me know what you think about :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,38 @@
+.DEFAULT_GOAL := help
+
+ifndef TAG
+	TAG=dev-alpine
+endif
+
+ifeq ($(findstring alpine,$(TAG)),alpine)
+	DOCKERFILE=Dockerfile-Alpine
+else ifeq ($(findstring arm,$(TAG)),arm)
+	DOCKERFILE=Dockerfile-QEMU-ARM
+else
+	DOCKERFILE=Dockerfile
+endif
+
+.PHONY: build
+build: ## Build a Docker image
+	docker build \
+		--pull \
+		--tag freshrss/freshrss:$(TAG) \
+		-f Docker/$(DOCKERFILE) .
+
+.PHONY: start
+start: ## Start the development environment (use Docker)
+	docker run \
+		--rm \
+		-v $(shell pwd):/var/www/FreshRSS:z \
+		-p 8080:80 \
+		-e FRESHRSS_ENV=development \
+		--name freshrss-dev \
+		freshrss/freshrss:$(TAG)
+
+.PHONY: stop
+stop: ## Stop FreshRSS container if any
+	docker stop freshrss-dev
+
+.PHONY: help
+help:
+	@grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
--- a/docs/en/developers/01_First_steps.md
+++ b/docs/en/developers/01_First_steps.md
@@ -1,6 +1,59 @@
-# Environment configuration
+# Environment configuration (Docker)
 
-**TODO**
+FreshRSS is built with PHP and uses a homemade framework, Minz. The dependencies are directly included in the source code, so you don't need Composer.
+
+There are various ways to configure your development environment. The easiest and most supported method is based on Docker, which is the solution documented below. If you already have a working PHP environment, you probably don't need it.
+
+We assume here that you use a GNU/Linux distribution, capable of running Docker. Otherwise, you'll have to adapt the commands accordingly.
+
+The commands that follow have to be executed in a console. They start by `$` when commands need to be executed as normal user, and by `#` when they need to be executed as root user. You don't have to type these characters. A path may be indicated before these characters to help you identify where they need to be executed. For instance, `app$ echo 'Hello World'` indicates that you have to execute `echo` command in the `app/` directory.
+
+First, you need to install [Docker](https://docs.docker.com/install/linux/docker-ce/ubuntu/).
+
+Once you're done, clone the repository with:
+
+```console
+$ git clone https://github.com/FreshRSS/FreshRSS.git
+$ cd FreshRSS
+```
+
+Note that, if you want to contribute, you have to fork the repository first and clone your fork instead of the "root" one. Adapt the commands in consequence.
+
+Then, the only command you need to know is the following:
+
+```console
+$ make start
+```
+
+This might take some time while Docker downloads the image. If your user isn't in the `docker` group, you'll need to prepend the command with `sudo`.
+
+**You can now access FreshRSS at [http://localhost:8080](http://localhost:8080).** Just follow the install process and select the SQLite database.
+
+You can stop the containers by typing <kbd>Control</kbd> + <kbd>c</kbd> or with the following command, in another terminal:
+
+```console
+$ make stop
+```
+
+If you're interested in the configuration, the `make` commands are defined in the [`Makefile`](/Makefile).
+
+If you need to use a different tag image (default is `dev-alpine`), you can set the `TAG` environment variable:
+
+```console
+$ TAG=dev-arm make start
+```
+
+You can find the full list of available tags [on the Docker hub](https://hub.docker.com/r/freshrss/freshrss/tags).
+
+You might want to rebuild the Docker image locally. You can do it with:
+
+```console
+$ make build
+$ # or
+$ TAG=dev-arm make build
+```
+
+The `TAG` variable can be anything (e.g. `dev-local`). You can target a specific architecture by adding `-alpine` or `-arm` at the end of the tag (e.g. `dev-local-arm`).
 
 # Project architecture
 


marienfressinaud:
BTW, I don't understand why Travis keeps failing on Hadolint :/

Frenzie:
I didn't look at the specifics much (no time/energy atm) but the principle is great. For example I recently wanted to do something with Wordpress and it only took me a minute or two to find & run the Docker image, which almost makes it something you can do on impulse.

Alkarex:
I am not sure to get the benefit compared with the existing Docker images we have. For development, one can for instance map a local development directory:
```
docker run ... -v  /path-to-local/FreshRSS-dev:/var/www/FreshRSS ... --name freshrss_dev freshrss/freshrss:dev
```
I would suggest tuning the existing commands / images to allow additional usecases if needed, rather than making new ones.

marienfressinaud:
My main motivation was to have a simpler configuration to help developers to understand what's behind a working environment. I find current Docker images a bit difficult to assimilate and I think that using directly PHP and Nginx images with a minimum of "tuning" can help to understand how it works… but there are other ways to explain the environment to developers so I don't really care to use existing images indeed (it's just more work to explain ;)). Also, I did it this way because I had existing configuration for two other projects and it was really little work to adapt to FreshRSS and it allowed me to reuse Docker images that I already downloaded (but FreshRSS ones are quite small so it's not a big deal!)

Alkarex:
I suggest we work on adding a section in https://github.com/FreshRSS/FreshRSS/blob/dev/Docker/README.md on how to use the existing images for FreshRSS development. Here is a full working example:

```sh
cd /path-to-local/FreshRSS/
docker run -it --rm -p 8080:80 -e TZ=Europe/Paris \
  -v $(pwd):/var/www/FreshRSS \
  freshrss/freshrss:dev
```

Then browse to 8080.

marienfressinaud:
I still like the approach of the Makefile to not have to remember the command (or we should rely on the terminal history but I find this approach less pleasant)

Frenzie:
Also you could add a `make push` target. Cf. https://github.com/koreader/virdevenv/blob/9b41d9f36d9b20fdf9b58d04104cc9133fc15ce5/docker/ubuntu/baseimage/Makefile

Alkarex:
Adding a make file, maybe, for those that are familiar with it, but I am unsure of the value (edit: maybe to run PHP tests, and build tests, for instance). The push example is not something even developers should do as we have a continuous integration for that.

Really, the parameters that remain are user-specific (where is your local directory, what port to use, what timezone). We can have a simple shell script or an alias to start the command.  We can even strip some of those parameters out and use default values :-)

```sh
cd /path-to-local/FreshRSS/
docker run --rm -v $(pwd):/var/www/FreshRSS freshrss/freshrss
```

I think that bringing `make`, Docker Compose (which is as far as I can see on the way out with Kubernetes), and especially nginx (no value here, has it?) to the table only increases the complexity.

Frenzie:
@Alkarex Interesting, how's that CI set up exactly? I know those things are there besides Travis but I didn't realize it also auto-pushed.

> We can have a simple shell script or an alias to start the command.

Sure, "Makefile" in this sense is functionally almost completely identical to "shell script". The difference is just that Make is a standardized tool that will do tab autocomplete out of the box. Some random shell script won't, and whatever its name is, it'll be ever so slightly harder to remember than something more standard.

Alkarex:
@Frenzie So it shows that it is nice to talk about those things once in a while :-) Here is it in Docker Hub
image
Automatically implemented as a Webhook in GitHub:
image

Frenzie:
Thanks, I found it!

I'll have to investigate that a bit more. It doesn't really seem like it'd save me much time though. I change the (KOReader) Docker images, I test them, I Git commit & push them… and then I Docker push it too. Occasionally Docker and Git reversed, but that's pretty much besides the point. There's one other person who might do so, and they do the same.

I suppose it'd be helpful to know that the Docker image itself builds if someone else proposes a change, but that by itself is not normally a concern. Still, it's a minor convenience and it'd guarantee consistency.

Alkarex:
With this CI, it also automatically tests the Docker builds on the PRs. What we do not have yet, and should have, is a set of basic tests, which can autorun to check that the basic features are still working after a build.

aledeg:
I think the Makefile is a great addition to the project. For a long time, I had the same thoughts as @Alkarex about it. But now, I am convinced that it will ease development, especially for new contributors.
To be truly useful though, we need to add rules. For instance, at work, I have a rule to start the whole project (downloading docker images, starting them, running composer to get the dependencies, populating the database), a rule to launch unit test, a rule to launch coding standard validation, and so on. This way the whole team use the same configuration everywhere and travis use also the same configuration because it's using also the makefile.
As a bonus, you don't have/need to remember every command and their parameters but if you do need to see what is really happening, you can always check the file content.
If it's completely automated and self-explanatory, the documentation can lighten as well since you don't need to explain every details in the configuration.

So I advocate to keep it but also to improve it!

Frenzie:
In that case I'd say you probably want two Makefiles, one in the repo root with commands like `make test` and `make lint`, as well as perhaps `make run`. The latter would be a phony that does something like `$(MAKE) -C Docker run`.

PS I think up/down and start/stop would make better pairs than run/down.

Alkarex:
I am not against a Makefile as long as it adds value and does not contain hardcoded data that users would have to manually edit in the Makefile itself.
My main concern is not about the Makefile, but about the rest of the PR :-)

Frenzie:
Sure, I agree that adding an extra Docker image this doesn't seem to make much sense. That being said, for development purposes one might want `display_errors = on`, `E_ALL` and the like which you wouldn't want in deployment. So if an extra Docker image makes sense it should probably be a minor modification with the other one as a parent.

aledeg:
@Frenzie To address that, it is better to play with envvar. It's quite easy to inject specific values when starting the container. To ease that kind of things, you can imagine that there is a rule in the makefile which change those envvar when run.

Frenzie:
@aledeg I actually didn't know you could use environment variables in php.ini, awesome! I imagine we'd still need a custom php.ini to allow for that, but a generic one for the container would suffice.

aledeg:
FYI: [PHP: The configuration file - Manual](https://www.php.net/manual/en/configuration.file.php)

Alkarex:
@Frenzie We do not need to edit php.ini for that, we already have the following, which we could make reactive to an environment variable:
https://github.com/FreshRSS/FreshRSS/blob/3f8804f54f8426961a31287fb8e9a3d8f8f84b4d/lib/Minz/FrontController.php#L125-L128

Frenzie:
@Alkarex Alright, probably no changes required to the Docker image in and of itself then. :+1:

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
-# Environment configuration
+# Environment configuration (Docker)
 
-**TODO**
+FreshRSS is built with PHP and use a homemade framework, Minz. The dependencies are directly included in the source code and so you don't need Composer.
```
Frenzie:
```suggestion
FreshRSS is built with PHP and uses a homemade framework, Minz. The dependencies are directly included in the source code, so you don't need Composer.
```

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+
+There are various ways to configure your development environment, the easiest and most supported one is based on Docker. This is the solution documented below. If you already have a working PHP environment, you probably don't need it.
```
Frenzie:
```suggestion
There are various ways to configure your development environment. The easiest and most supported method is based on Docker, which is the solution documented below. If you already have a working PHP environment, you probably don't need it.
```

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+
+We assume here that you use a GNU/Linux distribution, capable to run Docker. Otherwise, you'll have to adapt the commands to your case.
```
Frenzie:
```suggestion
We assume here that you use a GNU/Linux distribution, capable of running Docker. Otherwise, you'll have to adapt the commands accordingly.
```

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+
+The commands that follow have to be executed in a console. They start by `$` when commands need to be executed as normal user, and by `#` when they need to be executed as root user. You don't have to type these characters. A path may be indicated before to help you to identify where they need to be executed. For instance, `app$ echo 'Hello World'` indicates that you have to execute `echo` command under `app/` directory.
```
Frenzie: Or rather, you shouldn't type these characters? ;-)

```suggestion
The commands that follow have to be executed in a console. They start by `$` when commands need to be executed as normal user, and by `#` when they need to be executed as root user. You don't have to type these characters. A path may be indicated before these characters to help you identify where they need to be executed. For instance, `app$ echo 'Hello World'` indicates that you have to execute `echo` command in the `app/` directory.
```

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+```
+
+Note that, if you want to contribute, you have to fork the repository first and clone your fork instead of the "root" one. Adapt the commands in consequence.
+
+Then, the only command you need to know is the following:
+
+```console
+$ make start
+$ # or if your user is not in the `docker` group
+$ SUDO=sudo make start
```
Frenzie: Is there a particular reason not just to use `sudo make start`?

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+
+We assume here that you use a GNU/Linux distribution, capable to run Docker. Otherwise, you'll have to adapt the commands to your case.
+
+The commands that follow have to be executed in a console. They start by `$` when commands need to be executed as normal user, and by `#` when they need to be executed as root user. You don't have to type these characters. A path may be indicated before to help you to identify where they need to be executed. For instance, `app$ echo 'Hello World'` indicates that you have to execute `echo` command under `app/` directory.
+
+First, you need to install [Docker](https://docs.docker.com/install/linux/docker-ce/ubuntu/).
+
+Once you're done, clone the repository with:
+
+```console
```
Frenzie, line 15: I wasn't aware GH had this one. Clever.

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+$ make start
+$ # or if your user is not in the `docker` group
+$ SUDO=sudo make start
+```
+
+This might take some time while Docker download the image.
+
+**You can now access FreshRSS at [http://localhost:8080](http://localhost:8080).** Just follow the install process and select the SQLite database.
+
+You can stop the containers by typing `CTRL + C` or with the following command, in another terminal:
```
Frenzie:
```suggestion
You can stop the containers by typing <kbd>Control</kbd> + <kbd>c</kbd>` or with the following command, in another terminal:
```

Using `<kbd>` looks better (plus it provides semantics), lowercase c for consistency, see https://github.com/FreshRSS/FreshRSS/tree/125a83efc9e6cafb5eb274b07d1d006b0abb3031/Docker#development-mode

docs/en/developers/01_First_steps.md
marienfressinaud: It's because in the previous version, I used the current user UID and GID to set on the PHP containers. By calling `sudo` directly, the values weren't the correct ones:

```console
[marien@pizza FreshRSS]$ id
uid=1000(marien) gid=1000(marien) groupes=1000(marien),10(wheel)
[marien@pizza FreshRSS]$ sudo id
uid=0(root) gid=0(root) groupes=0(root)
```

Now it's not required anymore indeed, but I do not like the current behavior which change the ownership of the files in the `Docker/entrypoint.sh` file. I don't have the courage to change that though: it works correctly like that and there are more important things to work on :).

Edit: In conclusion, I don't mind to change `SUDO=sudo` in `sudo` :)

docs/en/developers/01_First_steps.md
marienfressinaud: I discovered that on Gitlab, I didn't even check that it was working on GitHub :sweat_smile:

docs/en/developers/01_First_steps.md
marienfressinaud: I saw that in the other PR and told myself I should do the same here… but then I forgot about it!

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+$ make start
+$ # or if your user is not in the `docker` group
+$ SUDO=sudo make start
+```
+
+This might take some time while Docker download the image.
+
+**You can now access FreshRSS at [http://localhost:8080](http://localhost:8080).** Just follow the install process and select the SQLite database.
+
+You can stop the containers by typing <kbd>Control</kbd> + <kbd>c</kbd>` or with the following command, in another terminal:
```
Alkarex:
```suggestion
You can stop the containers by typing <kbd>Control</kbd> + <kbd>c</kbd> or with the following command, in another terminal:
```

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,47 @@
...
+We assume here that you use a GNU/Linux distribution, capable of running Docker. Otherwise, you'll have to adapt the commands accordingly.
+
+The commands that follow have to be executed in a console. They start by `$` when commands need to be executed as normal user, and by `#` when they need to be executed as root user. You don't have to type these characters. A path may be indicated before these characters to help you identify where they need to be executed. For instance, `app$ echo 'Hello World'` indicates that you have to execute `echo` command in the `app/` directory.
+
+First, you need to install [Docker](https://docs.docker.com/install/linux/docker-ce/ubuntu/).
+
+Once you're done, clone the repository with:
+
+```console
+$ git clone git@github.com:FreshRSS/FreshRSS.git
```
Alkarex: I would remove this one, for simplicity, and keep only the HTTPS version, which is more robust (e.g. with respect to firewalls, etc.)

Makefile
```diff
@@ -0,0 +1,18 @@
+.DEFAULT_GOAL := help
+
+.PHONY: start
+start: ## Start the development environment (use Docker)
+	$(SUDO) docker run \
+		--rm \
+		-v $(shell pwd):/var/www/FreshRSS:z \
+		-p 8080:80 \
+		--name freshrss \
+		freshrss/freshrss:dev-alpine
```
Alkarex: This hard-coded image name currently does not work on e.g. ARM. If we want to support ARM automatically, we have a few options:
1. add a metafile to Docker Hub (best)
2. use only locally-built Docker images when using Makefile (probably easiest - I think I would go for that one)
3. include a logic here to pull the correct one (not very portable). Example of logic on Debian (but not very portable):
```sh
arch=`dpkg --print-architecture`
if [[ $arch =~ ^arm ]]; then arch="arm";
elif [[ $arch =~ ^amd64 ]]; then arch="amd64";
else echo "Unsupported architecture" >&2; exit 1; fi
echo "Architecture: $arch"
```

Makefile
```diff
@@ -0,0 +1,18 @@
+.DEFAULT_GOAL := help
+
+.PHONY: start
+start: ## Start the development environment (use Docker)
+	$(SUDO) docker run \
+		--rm \
+		-v $(shell pwd):/var/www/FreshRSS:z \
```
Alkarex, line 26: I think you could take advantage of https://github.com/FreshRSS/FreshRSS/pull/2508 and use `-e FRESHRSS_ENV=development`

docs/en/developers/01_First_steps.md
Alkarex: It seems to work fine https://github.com/flusio/FreshRSS/blob/tec/improve-development-environment/docs/en/developers/01_First_steps.md

docs/en/developers/01_First_steps.md
marienfressinaud: Yes, I put the two of them because I tend to prefer SSH, but let's keep it simple :+1:

Makefile
```diff
@@ -0,0 +1,18 @@
...
+		-p 8080:80 \
+		--name freshrss \
```
Alkarex: Maybe that should either be with a more unique name (in the case the user already has a freshrss instance running), or even no name at all (in this mode without `-d` the command does not return anyway, and deletes the container at the end)

marienfressinaud: I prefer to use a name because it lets us to connect to the container more easily (i.e. without getting the output of `docker ps`) or even to build other tools around the container (e.g. other make commands). What about `freshrss-dev`?

Makefile
marienfressinaud: I started to add it and it came to my mind a developer still can change its `data/config.php` file and might want to test FRSS in production mode. I agree it would be a minority of use cases and it's better for developers to start immediately in dev mode, so I'll probably set the environment variable, but I wanted to share my thoughts first ;)

Makefile
marienfressinaud: I don't know about "metafile" on Docker Hub and can't find any information on it… what have you in mind with that?

Also I didn't thought that developers would develop FRSS on ARM architecture… but I'm still at the "ARM is used essentially for mobile phones" phase and I don't follow actuality about ARM :sweat_smile:

marienfressinaud: We could also choose the image tag via an environment variable. Example:

```console
$ TAG=dev-arm make start
```

I would opt for this simple solution for the moment.

docs/en/developers/01_First_steps.md
Frenzie: I tend to think of HTTPS as read-only (it's not, of course).

Makefile
Frenzie: @marienfressinaud I've seen PRs on GitHub by people who apparently developed *on Android*. I mean, I'd maybe possibly consider using ARM something or other at some indeterminate point in the future but only if I could run a proper OS.

docs/en/developers/01_First_steps.md
```diff
@@ -1,6 +1,49 @@
...
+
+Note that, if you want to contribute, you have to fork the repository first and clone your fork instead of the "root" one. Adapt the commands in consequence.
+
+Then, the only command you need to know is the following:
+
+```console
+$ make start
+```
+
+This might take some time while Docker download the image. If your user isn't in the `docker` group, you'll need to prepend the command by `sudo`.
```
Frenzie:
```suggestion
This might take some time while Docker downloads the image. If your user isn't in the `docker` group, you'll need to prepend the command with `sudo`.
```

marienfressinaud:
I think this PR is good too. It'll need to be rebased on dev and squashed though but nothing complicated.

Makefile
```diff
@@ -0,0 +1,23 @@
+.DEFAULT_GOAL := help
+
+ifndef TAG
```
Frenzie, line 3: Why not a simple `?=`?

Alkarex:
@marienfressinaud 
image ;-)

Makefile
Alkarex: Yes, sounds good

Makefile
marienfressinaud: Because I don't know the syntax of Makefile good enough :D in the mean time, I find this form more comprehensible for newcomers!

Frenzie: I find it much harder to read this way (very verbose) but eh, whatever. :-P

Alkarex:
I think the Makefile should have a rule to build an image. Something like:
```sh
docker build --pull --tag freshrss/freshrss-local-dev -f Docker/Dockerfile .
```
Ideally, the start rule should call the build rule if the local build does not exist

marienfressinaud:
I added a `build` target which also accepts the `TAG` variable (it automatically choose the correct Dockerfile based on a `-arm` or `-alpine` at the end of the tag). There is a note at the end of the documentation to explain this. I decided to keep the `start` rule simple. By default, it will work fine since it will look for an existing image on the Docker hub, but if the user wants a specific image, he will have to build it explicitly. I don't think that building a custom image automatically is worse a more complex `start` target.

Edit: and the branch is rebased + squashed, it can be merged if the `build` target is good enough for you

Alkarex:
Looks good to me :-)

Alkarex (approved):

PR #2205, opened 2019-01-01
Docker image alternative using Ubuntu

Alkarex:
Ubuntu seems to be faster, cf. e.g. PHPBench on:
* https://www.phoronix.com/scan.php?page=article&item=freebsd-12-windows&num=5
* https://www.phoronix.com/scan.php?page=article&item=docker-summer-2018&num=4

This image is also different from our Alpine-based image as it is using Apache Event and PHP-FPM as FastCGI. They are also supposed to be faster for some types of workload than mod_php but with other drawbacks (poorer control of flush, more processes running...).
So some concreate performance tests would be needed to check whether this is relevant at all.
It is quite usual for Docker images to offer both a Debian/Ubuntu and an Alpine version, so we could also do the same if there is any need.

Follow-up of https://github.com/FreshRSS/FreshRSS/pull/2202

PR #2495, opened 2019-08-17
Require PHP 5.5+

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2469#issuecomment-522255093
I think it would be reasonable to require PHP 5.5+ for the core of FreshRSS after all.

As Frenzie said, WordPress currently requires PHP 5.6.20+, and it is the most popular PHP application.

We would loose about 20% of the PHP servers according to https://w3techs.com/technologies/details/pl-php/5/all but I expect this number to drop fast after the release of CentOS 8 (CentOS accounts for 17% of Linux servers https://w3techs.com/technologies/details/os-linux/all/all ).

Distributions:
* no impact on Ubuntu, Fedora, Alpine, OpenWRT, FreeBSD, OpenSuze, Mageia, as all active versions have PHP ≥ 7
* no impact on Synology, as all active versions have PHP ≥ 5.6
* we drop Debian 8 Jessie (-2020) - we keep supporting Debian 9 Stretch (2017-06) - current is Debian 10 Buster
* we drop Red Hat 7 (-2024) - we keep supporting RHEL 8 (2019-05)
* we drop CentOS 7 (-2024) - we will support CentOS 8 (to be released soonish)

When dropping older versions, I can better like when it is for a good reason, and there is actually one with PHP 5.5, namely generators (yield) https://php.net/language.generators.overview which I consider
using.


Alkarex:
Previous discussions:
* https://github.com/FreshRSS/FreshRSS/pull/1618
* https://github.com/FreshRSS/FreshRSS/issues/789#issuecomment-73878076

lib/lib_rss.php
```diff
@@ -1,9 +1,9 @@
 <?php
-if (version_compare(PHP_VERSION, '5.3.8', '<')) {
-	die('FreshRSS error: FreshRSS requires PHP 5.3.8+!');
+if (version_compare(PHP_VERSION, '5.5.0', '<')) {
+	die('FreshRSS error: FreshRSS requires PHP 5.5.0+!');
 }
 
-if (!function_exists('json_decode')) {
+if (!function_exists('json_decode')) {	//PHP bug #63520 < PHP 7
```
Alkarex, line 6: https://github.com/FreshRSS/FreshRSS/issues/306

.travis.yml
```diff
@@ -30,12 +30,7 @@ env:
 matrix:
   fast_finish: true
   include:
-    # PHP 5.3 only runs on Ubuntu 12.04 (precise), not 14.04 (trusty)
-    - php: "5.3"
-      dist: precise
-    # PHP 5.4 & 5.5 only run on Travis in 14.04 (trusty), not 16.04 (xenial)
-    - php: "5.4"
-      dist: trusty
+    # PHP 5.5 only run on Travis in 14.04 (trusty), not 16.04 (xenial)
```
Frenzie:
```suggestion
    # PHP 5.5 only runs on Travis in 14.04 (trusty), not 16.04 (xenial)
```
Stupid nit but the old sentence referred to two versions. ;-)

README.fr.md
```diff
@@ -219,7 +219,6 @@ Tout client supportant une API de type Fever ; Sélection :
 * [bcrypt.js](https://github.com/dcodeIO/bcrypt.js)
 * [phpQuery](https://github.com/phpquery/phpquery)
 * [Services_JSON](https://pear.php.net/pepr/pepr-proposal-show.php?id=198)
-* [password_compat](https://github.com/ircmaxell/password_compat)
```
Frenzie, line 222: I don't think `git rm lib/password_compat.php` is part of this PR yet?

Frenzie: My bad, GH hides big changes so at a glance it looks like there are none.

Alkarex: Yes, `password_compat` is removed because it is not needed from PHP 5.5+

Alkarex:
Actually, moving to PHP 5.6+ would not be much harder than 5.5+ based on the statistics and versions available in the distributions (which have either a version <= 5.4 or >= 5.6).
I could not find any big distribution on DistroWatch that is maintaining a PHP 5.5, e.g. https://distrowatch.com/table.php?distribution=Debian

Frenzie:
No real objections here. Anything specific though?

https://www.php.net/manual/en/migration56.new-features.php

A few of those sound somewhat interesting (e.g., variadic functions) but at the same time a minor syntactic convenience isn't much of a reason.

Alkarex:
Ok. So let's stick to 5.5 now with a note that if we need 5.6 for anything, we can update without delay.

marienfressinaud:
I'm ok with this decision too :+1:

I don't know how relevant the stats from w3techs.com are for our users, it may be interesting to do a survey to ask users where and how they host their instance of FreshRSS. Or even to add an  option in FRSS (disabled by default) to send anonymous information automatically (but it's more work and I don't have time to work on this for the moment)

lib/Minz/Configuration.php
```diff
@@ -198,7 +198,7 @@ public function save() {
 			return false;
 		}
 
-		// Clear PHP 5.5+ cache for include
+		// Clear PHP cache for include
 		if (function_exists('opcache_invalidate')) {
```
marienfressinaud, line 202: Do we still need to check if the function exists?

marienfressinaud (approved):

lib/Minz/Configuration.php
Alkarex: Some distributions do not have `opcache` by default, e.g. https://pkgs.alpinelinux.org/packages?name=php*opcache*&branch=edge&arch=x86_64 , which makes me realise that:
* We should add that do the documentation
* We should add it in the Dockerfile for Alpine
* Maybe that might explain a part of why Alpine was so much slower than Ubuntu in my last test https://github.com/FreshRSS/FreshRSS/pull/2205

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/2498

Alkarex:
Actually, I am refactoring a bit our PDO code (https://github.com/FreshRSS/FreshRSS/pull/2522 ) and I do have a wish to move to PHP 5.6+ to be able to use constant arrays
https://www.php.net/manual/en/migration56.new-features.php

Instead of the current `global` approach, which is not so nice, e.g.:
https://github.com/FreshRSS/FreshRSS/blob/8dcdde6251ae4dfc690b1a014488df125c5e5cdc/app/Models/UserDAO.php#L21-L22

https://github.com/FreshRSS/FreshRSS/blob/c76a318193cda63064625b2d92c719b7150d7d64/app/SQL/install.sql.pgsql.php#L4-L9

What do you think? Is it worth it?

Frenzie:
It's a real feature, so I don't see why not. :+1:

marienfressinaud:
It would be a nice simplification :) and https://github.com/FreshRSS/FreshRSS/pull/2495#issuecomment-522538468 was already a sufficient reason to move on

Alkarex:
CentOS 8 will be released next week with PHP 7.2 https://wiki.centos.org/About/Building_8

PR #2506, opened 2019-08-21
Prevent window opener vulnerability with space shortcut

ArthurHoaro:
This change fixes a vulnerability introduced by `window.open()` on untrusted sources. It reproduces the effect of `rel="noreferrer"` with JS.

Cross browser solution from: https://stackoverflow.com/a/40593743

Related to #1245

## Reproduction

> tested with Firefox 68

  1. Add this RSS feed https://www.ecyseo.net/feed/rss
  2. Open the 2nd link "À propos de la faille de sécurité liée à target="_blank" **using the space key shortcut**.
  3. Click on the first of three links "http://bookmarks.ecyseo.net"

Current behaviour: the FreshRSS tab changes.
Expected behaviour: no effect on FreshRSS

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -803,7 +803,9 @@ function init_shortcuts() {
 				if (context.auto_mark_site) {
 					mark_read(document.querySelector('.flux.current'), true, false);
 				}
-				window.open(document.querySelector('.flux.current a.go_website').href);
+				const newWindow = window.open();
+                                newWindow.opener = null;
+				newWindow.location = document.querySelector('.flux.current a.go_website').href;
 				return false;
 			}
 			if (k === s.skip_next_entry) { next_entry(true); return false; }
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -804,8 +804,10 @@ function init_shortcuts() {
 					mark_read(document.querySelector('.flux.current'), true, false);
 				}
 				const newWindow = window.open();
-                                newWindow.opener = null;
-				newWindow.location = document.querySelector('.flux.current a.go_website').href;
+				if (newWindow) {
+					newWindow.opener = null;
+					newWindow.location = document.querySelector('.flux.current a.go_website').href;
+				}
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -803,7 +803,9 @@ function init_shortcuts() {
 				if (context.auto_mark_site) {
 					mark_read(document.querySelector('.flux.current'), true, false);
 				}
-				window.open(document.querySelector('.flux.current a.go_website').href);
+				const newWindow = window.open();
```
Frenzie, line 806: I assume you mean this?

```suggestion
				const newWindow = window.open;
```

Edit: Huh, you can store a reference to the opened window? Interesting. Any particular reason to do this instead of the obvious though? Something like `window.open(url, 'noreferrer', 'noopener')` seems better. IE I assume?

ArthurHoaro: Yes, the SO thread I linked seemed hesitant about browser compatibility, and according to your README IE11 is supported, so I went with the old fashion method. Let me know if you prefer the other one, I'll update the PR.

Alkarex: For reference https://developer.mozilla.org/docs/Web/API/Window/open

Alkarex:
This looks fine. I have just made a little change in the case of popup blockers, in which case `window.open()` returns null.
Tested in IE11, Edge, Firefox

Alkarex:
I have just found a bug in IE11 + Edge with the Space shortcut. ~~Fix coming~~
Edit: fix https://github.com/FreshRSS/FreshRSS/pull/2507

Alkarex:
@ArthurHoaro Thanks again, and please add yourself to https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/6862

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/7077

PR #2507, opened 2019-08-21
IE11 / Edge keyboard compatibility

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2506#issuecomment-523544684

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -720,8 +720,14 @@ function init_shortcuts() {
 				return true;
 			}
 
-			const s = context.shortcuts,
-				k = (ev.key.trim() || ev.code).toUpperCase();
+			const s = context.shortcuts;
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -720,8 +720,14 @@ function init_shortcuts() {
 				return true;
 			}
 
-			const s = context.shortcuts,
-				k = (ev.key.trim() || ev.code).toUpperCase();
+			const s = context.shortcuts;
+			let k = (ev.key.trim() || ev.code || 'Space').toUpperCase();
+
+			//IE11
```
Frenzie, line 726: Should probably specify that it's also Edge like the PR title.
```suggestion
			//IE11 / Edge
```

Alkarex: Actually, that part is solely IE11. Edge does not specify `ev.code` for a space, which is covered by the previous line, while also preventing a risk of null

PR #2515, opened 2019-09-03
add: Allow admins to configure default feeds

### LINKED ISSUE #2514: Configuring default feeds
I'd like to allow admins to configure the default feeds. The feeds will be defined in the `data/config.php` file instead of the `app/SQL/install.sql.*.php` files (`SQL_INSERT_FEEDS`).

It's pretty straightforward to implement:

- [x] transform MySQL instructions to work with an array instead of a single SQL string in order to homogenize the code between databases (see [this lines](https://github.com/FreshRSS/FreshRSS/blob/1.14.3/app/Models/UserDAO.php#L32-L47))
- [x] move actual default feeds from the SQL string to the default configuration
  - transform `SQL_INSERT_FEEDS` into a single string which accept `url`, `name`, `website` and `description` "arguments"
  - add a `default_feeds` array to the system configuration
  - iterate on `default_feeds` instead of the `$SQL_INSERT_FEEDS` instructions in the`userDAO->createUser` method
- [x] remove the FRSS website feed since it's now a redirection to the GitHub releases feed
- [x] update the install script

I don't plan to let admins to configure these feeds from the interface yet.

If you see more default feeds to add by default, feel free to suggest them!

References:
- initial feature suggestion: https://github.com/FreshRSS/FreshRSS/issues/859
- FRSS website feed redirection: https://github.com/FreshRSS/freshrss.org/issues/38

marienfressinaud:
Closes #2514

### DIFF
--- a/app/Models/UserDAO.php
+++ b/app/Models/UserDAO.php
@@ -30,19 +30,11 @@ class FreshRSS_UserDAO extends Minz_ModelPdo {
 				}
 			}
 			if ($ok && $insertDefaultFeeds) {
-				if (defined('SQL_INSERT_FEEDS')) {	//E.g. MySQL
-					$sql = sprintf(SQL_INSERT_FEEDS, $bd_prefix_user);
+				global $SQL_INSERT_FEEDS;
+				foreach ($SQL_INSERT_FEEDS as $instruction) {
+					$sql = sprintf($instruction, $bd_prefix_user);
 					$stm = $userPDO->bd->prepare($sql);
-					$ok &= $stm && $stm->execute();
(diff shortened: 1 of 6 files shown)


Alkarex:
Not tested yet, but let's go for this approach for now. I think a future evolution could be to auto-load an OPML file if it exists, or a default OPML file otherwise, using the existing import functionality that we have.

Alkarex:
Looks good at a quick glance

marienfressinaud:
Indeed, it is not in replacement of a better onboarding system! It's just a quick improvement to an existing feature to help administrators who want to host several users and to adapt the default feeds :)

Issue #2491, opened 2019-08-16
Being realistic about documentation


marienfressinaud:
I open a ticket to discuss about documentation (and I created in the mean time a label to identify "discussion" tickets).

FreshRSS had poor documentation for years, and it went more evident to me since I didn't work on it for 3 years. Here's 3 ideas to improve the situation and that I had successfully applied to other projects:

- [x] keep English documentation only, we are not able to maintain French version in parallel (maybe except the "user" documentation?) and it's more difficult for non-French people to contribute (see https://github.com/FreshRSS/FreshRSS/pull/2584)
- [ ] remove the "blank" sections, there are a lot of "TODOs" and it's just discouraging ourselves
- [x] add a "[PR template](https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository)" containing a list of points to check before merging, including "Document changes" (here's [an example](https://github.com/lessy-community/lessy/blob/master/docs/pull_request_template.md))

I keep this list to 3 tasks since it's always difficult to engage ourselves to more, but feel free to suggest more ideas.

marienfressinaud:
The first point was transformed into "improving our tools to help translating" (https://github.com/FreshRSS/FreshRSS/pull/2590) and, while I was a bit dubious at first, it works pretty well thanks to the work of @Frenzie, great job! :clap: 

I'll find time in the next days to work on the two other points :)

math-GH:
@marienfressinaud @Frenzie @Alkarex 
How do we want proceed here?
Is this topic solved from your side? 

I like the current documentation. (A documentation is never 'complete', but it is growing)

Frenzie:
I wasn't able to quite automate po4a at the time for CI (because I'd have had to compile the latest version every time, give or take some cache) but I'd close this ticket.

Alkarex:
@Frenzie I have forgotten how to sync / update the po4a. Could you please refresh my memory and/or making the corresponding PR?

Frenzie:
You can run update-translations.sh from the docs folder iirc but at the moment it'll require a fair bit of manual effort to fix up because they became out of sync again. (It looks like no one's bothered to do it when updating the translation :-) ).

You need at least 0.56.

PR #2164, opened 2018-12-04
Docs update

pattems:
As promised in https://github.com/FreshRSS/FreshRSS/pull/2158#issuecomment-444164782, I have done a substantial rewrite & reorganization to the English administration documentation. This pulls stuff from the Readme, the documentation I'm proposing replacing, comments on pull requests, and my own personal experience with using and administering my instance of FreshRSS.

I haven't actually had the chance to test the commands on Debian 9 or Ubuntu, so that needs to be verified. I can do that, I just haven't had the chance to yet. The only todo is Apache config in 06_LinuxInstall.md... and honestly it might be the better course of action to just rip that section out and link to a good external tutorial for installing a LAMP stack.

The only other thing I'm unsure of is why footnotes, which I used in a couple of places, don't seem to be behaving correctly; they work fine in https://markdown-it.github.io/, but don't seem to actually convert to proper footnotes in the repository.

PR #2520, opened 2019-09-13
Add support for terms of service

marienfressinaud:
This feature is optional. It is based on the presence of a
`data/tos.html` file that an administrator can create. If this file
exists, FreshRSS will automatically add a "ToS" checkbox on the
registration page that users must check to be able to create their
account.


app/Controllers/indexController.php
```diff
@@ -259,6 +259,23 @@ public function aboutAction() {
...
 
+	/**
+	 * This action displays the EULA page of FreshRSS.
+	 * This page is enabled only if admin created a data/tos.html file.
+	 * The content of the page is the content of data/tos.html.
+	 * It returns 404 if there is no EULA.
+	 */
+	public function tosAction() {
+		$tos_filepath = DATA_PATH . '/tos.html';
+		if (!file_exists($tos_filepath)) {
```
Alkarex: I suggest to test for the return value of file_get_contents() instead of an additional test (which is less robust for e.g. access rights and race conditions)

app/i18n/fr/gen.php
```diff
@@ -23,6 +23,7 @@
 		'update' => 'Mettre à jour',
 	),
 	'auth' => array(
+		'accept_tos' => 'Vous acceptez les <a href="%s">Conditions Générales d’Utilisation</a>.',
```
Alkarex: Maybe "Vous devez accepter..." ?

data/.gitignore
```diff
@@ -5,3 +5,4 @@ force-https.txt
 last_update.txt
 no-cache.txt
 update.php
+tos.html
```
Alkarex, line 9: Maybe consider providing a tos.example.html or something

Alkarex (approved):

data/.gitignore
marienfressinaud: I think giving an example file is quite complicated since ToS can be very different across different instances while being too important to be based on some "default" document. I wanted to add a bit of documentation on this feature but I'm waiting for #2164 and #2491. I'm keeping a list of what I want to document after that. Do you still prefer that I add a note somewhere?

app/i18n/fr/gen.php
marienfressinaud: I think it's quite strange to have "Vous devez accepter…" next to the checkbox, but so is "Vous acceptez…" now that I think about it. What about "Accepter…"?

marienfressinaud: Or "J'accepte…"

Frenzie: Then the same would apply to English. (I accept the TOS/I agree to the TOS…)

Alkarex: Either "Accepter" (my first choice in French), or "J'accepte" (although a bit more "infantilisant" - I cannot find a translation of this concept in English, something like talking to someone who is not fully educated / grown up) sound fine

data/.gitignore
Alkarex: What about a *lorem ipsum* just to provide a template?

marienfressinaud:
It should be good now :)

marienfressinaud:
I added a `required` to the checkbox input + used `join_path`

app/Controllers/authController.php
```diff
@@ -205,7 +205,7 @@ public function registerAction() {
 			Minz_Error::error(403);
 		}
 
-		$this->view->show_tos_checkbox = file_exists(DATA_PATH . '/tos.html');
+		$this->view->show_tos_checkbox = file_exists(join_path(DATA_PATH, 'tos.html'));
```
Alkarex, line 208: I am not sure this adds any benefit, does it? (and it is slower, longer, and custom)
No need to change it back, but just for future cases. Windows does not need a special case, at least.
P.S.: Just to be sure, if you see this comment out of context, I am talking about `join_path()`

Alkarex (approved):

Alkarex:
Looks good :-) Just merge when you are fully done

app/Controllers/authController.php
marienfressinaud: Oh, I didn't know that! I take note for the next time :)

PR #2522, opened 2019-09-15
PDO refactoring for code simplification

Alkarex:
(As promised in https://github.com/FreshRSS/FreshRSS/pull/2496 . It may seem like a lot, but it is mainly some fairly automatic changes)
* Automatic prefix when using the syntax `_tableName` (with backticks)
  * This allows writing:

```sql
SELECT * FROM `_entry`;
```

instead of:

```sql
SELECT * FROM `' . $this->prefix . 'entry`; 
```

As well as writing:

```php
return $this->pdo->lastInsertId('`_tag_id_seq`');
```

instead of:

```php
return $this->bd->lastInsertId('"' . $this->prefix . 'tag_id_seq"');
```

* Uniformity: MySQL is now using `PDO::ATTR_EMULATE_PREPARES = false` just like SQLite and PostgreSQL, with consequences such as only one statement per query. See e.g. https://phpdelusions.net/pdo#emulation
* Use PDO methods exec(), query(), prepare() + execute() in a more efficient way
* Remove auto-update SQL code for versions older than FreshRSS 1.5 (3 years old)
* The name of the default category is set in PHP instead of in the DB (simplies SQL and allows changing the name according to the FreshRSS language)
* Rename `->bd` to `->pdo` (less of a frenshism, and more informative)
* Fix some requests, which were not compatible with MySQL prepared statements


Alkarex:
~~No ready yet: I lack some work in `install.php` (to try to reuse the main code instead of duplicating it), a syntax update for PostgreSQL `nextval()` and `lastInsertId()`, and a bit for MySQL in the SQL file.~~

app/Models/Category.php
```diff
@@ -68,8 +68,11 @@ public function hasFeedsWithError() {
 		return $this->hasFeedsWithError;
 	}
 
-	public function _id($value) {
-		$this->id = $value;
+	public function _id($id) {
+		$this->id = $id;
+		if ($id == FreshRSS_CategoryDAO::DEFAULTCATEGORYID) {
```
Alkarex, line 73: Approach to set the name of the default category when using it (simplifies the SQL, and makes the i18n possible)

lib/Minz/ModelPdo.php
```diff
@@ -102,58 +88,76 @@ public function __construct($currentUser = null, $currentPrefix = null, $current
...
+    }
+
+	abstract public function dbType();
+
+	private $prefix = '';
+	public function prefix() { return $this->prefix; }
+	public function setPrefix($prefix) { $this->prefix = $prefix; }
+
+	private function autoPrefix($sql) {
+		return str_replace('`_', '`' . $this->prefix, $sql);
```
Alkarex, line 129: The magic for the auto-prefix is here

Alkarex:
@marienfressinaud Maybe you could have a look at this PR before you leave for vacation, at least its general principles. `install.php` is not completely finished (not working fully at the moment) but when deployed over an existing installation, this PR should already work more or less (more testing needed though).

app/Models/Category.php
marienfressinaud: Also, I'm not sure it's still possible to rename the default category after that…

app/Models/EntryDAO.php
```diff
@@ -1001,68 +909,67 @@ public function updateLastSeen($id_feed, $guids, $mtime = 0) {
...
+		WHERE e2.is_favorite = 1
+		  AND e2.is_read = 0
+		  AND f2.priority >= :priority_normal2
+	) u
 ORDER BY o
 SQL;
-		$stm = $this->bd->prepare($sql);
-		$stm->execute(array(':priority_normal' => FreshRSS_Feed::PRIORITY_NORMAL));
+		$stm = $this->pdo->prepare($sql);
+		//Binding a value more than once is not standard and does not work with native prepared statements (e.g. MySQL) https://bugs.php.net/bug.php?id=40417
```
marienfressinaud, line 965: Oo this is very not intuitive! Does it mean previous code wasn't working correctly?

app/install.php
```diff
@@ -17,24 +17,10 @@
 	define('STEP', 0);
 }
 
-if (STEP === 3 && isset($_POST['type'])) {
+if (STEP === 2 && isset($_POST['type'])) {
```
marienfressinaud, line 20: Uh… was it necessary to inverse these two steps in this PR? There are so many changes that I'm not sure of what changed exactly :/

marienfressinaud (commented):
Most of these changes are very welcomed :+1: I don't have a lot to say except that I'm not totally comfortable with my review since the PR is quite big. Smaller dedicated PRs or even more focused commits would help a lot :)

app/Models/EntryDAO.php
Alkarex: It means the previous code was working just by chance, because the current PDO default behaviour for MySQL is to use emulated prepared statements instead of native, and because the PHP implementation accidentality allowed that. It has never been standard

app/install.php
Alkarex: The reason for changing the order is that you cannot create the default user fully if the database is not in place first. It becomes similar than for the CLI, were one first create the DB and general config, and then create the default user and other users.

Alkarex: In general, I am trying to remove the custom code in install.php and instead use the code we already have in the PDO classes, like for the CLI install.

app/Models/Category.php
Alkarex: Right, I will have a look

Alkarex: I am looking at another way to implement that. But would you like to allow renaming the default category? I was thinking about disabling this possibility, to simplify and also because it is a special category (a bit like favourites, etc.), but I have no strong opinion.

marienfressinaud: Maybe it's not that important :) feel free to choose what solution you think is the best, I will not be able to review the PR anyway :p

Alkarex: Ok, let me then polish/finalise the PR without the possibility to change the name of the default category, and we can re-add this option when you are back if you like (which will be easier than what we have now, just renaming the category after creation). Enjoy your trip!

PR #2526, opened 2019-09-17
API Reeder compatibility

### LINKED ISSUE #2513: Frequent authentication errors when trying to access Google Reader API via Reeder
After initially logging in fine, I keep getting authentication errors later when trying to mark articles as read or sync my feeds via Reeder. Initial sync seems to work well, but later it doesn't. Are there known issues using Reeder with FreshRSS via Google Reader API or is there an additional setting I need to change in FreshRSS to make it work?

Alkarex:
Fixes #2513

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -214,8 +214,10 @@ function token($conf) {
 function checkToken($conf, $token) {
 //http://code.google.com/p/google-reader-api/wiki/ActionToken
 	$user = Minz_Session::param('currentUser', '_');
-	if ($user !== '_' && $token == '') {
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Bit weird (from Feeder) but looks okay.

marienfressinaud:
So one can authenticate by sending an empty token (or "x")? :scream:

Frenzie:
You still need to have authenticated with user+pass, right?

Alkarex:
As @Frenzie writes, one still needs the full credentials if one does not provide a token

marienfressinaud:
Oh, ok then! :)

PR #2527, opened 2019-09-17
Require PHP 5.6+

Alkarex:
Discussion
https://github.com/FreshRSS/FreshRSS/pull/2495#issuecomment-531911473
Needed for constant arrays
https://www.php.net/manual/en/migration56.new-features.php which will be
used in https://github.com/FreshRSS/FreshRSS/pull/2522

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,6 @@ env:
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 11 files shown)


marienfressinaud (approved):
:ship:

Alkarex:
SimplePie now also requires PHP 5.6+ https://github.com/simplepie/simplepie/pull/625

PR #2533, opened 2019-09-25
Explicit Traefik 1.7

Alkarex:
Traefik 2.0 has some breaking changes.
Making the use of version 1.7 explicit while waiting for upgrading to 2.0
https://community.containo.us/t/traefik-release-v2-0-0/1628


Alkarex:
https://docs.traefik.io/v2.0/migration/v1-to-v2/

Issue #2434, opened 2019-07-03
[FR] Optionally display allways read and unread articles in Favorites


Offerel:
I have configured FreshRSS in that way, that only unread articles are displayed and starting with Main Stream. With that way, i see every time only the newest and unread articles. But sometimes i read a article and mark it as "Favorite". In that case, the article is marked as read by opening the article. The article is saved as unread and favorite. 

If i now go to Favorites, i didn't see my saved article at first, since its marked as read and i filtered read articles. I must additionally "unfilter" read articles to see my saved ones.

With that usecase, i think it would be nice, if clicking on "Favorites" disables also the unread filter. If i click another folder/label, the normal default behavior should be used.

This automatic toggling should be a option in the settings, so that user can disable it by itself.

PR #2535, opened 2019-09-25
Fix for FR to display unread in Favorites

Offerel:
This hardcodes state=3 for favorites. This should fix the FR https://github.com/FreshRSS/FreshRSS/issues/2434


marienfressinaud:
Duplicate of https://github.com/FreshRSS/FreshRSS/pull/2766 which is now merged :)

PR #2542, opened 2019-09-28
Use dc:creator instead of author for RSS 2.0 outputs

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2538#issuecomment-536227043

### DIFF
--- a/app/views/index/rss.phtml
+++ b/app/views/index/rss.phtml
@@ -17,7 +17,7 @@ foreach ($this->entries as $item) {
 				$authors = $item->authors();
 				if (is_array($authors)) {
 					foreach ($authors as $author) {
-						echo "\t\t\t" , '<author>', $author, '</author>', "\n";
+						echo "\t\t\t" , '<dc:creator>', $author, '</dc:creator>', "\n";
 					}
 				}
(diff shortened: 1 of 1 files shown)


app/views/index/rss.phtml
```diff
@@ -17,7 +17,7 @@ foreach ($this->entries as $item) {
 				$authors = $item->authors();
 				if (is_array($authors)) {
 					foreach ($authors as $author) {
-						echo "\t\t\t" , '<author>', $author, '</author>', "\n";
+						echo "\t\t\t" , '<dc:creator>', $author, '</dc:creator>', "\n";
```
Frenzie, line 20: Is `$author` here supposed to be a (pen) name or was this merely a trickle-down effect?

Alkarex:
In RSS 2.0, `<author>` expects an e-mail address https://cyber.harvard.edu/rss/rss.html#ltauthorgtSubelementOfLtitemgt
In most feeds I see, this is not the case, but rather a name

Frenzie:
Right, that's what I mean by trickle-down effect. So it comes from the original feed then.

Alkarex:
Yes

Issue #2448, opened 2019-07-19
MySQL: Article not imported (long content)


nicofrand:
Hi,

it seems an article cannot be imported if its content is too large:

**STR**

1. Import http://www.madd.fr/feed/atom/

**What happens:**

The article named "Bankass" has a long content (due to embedded inline images) and is rejected due to the following error:

```
[Fri, 19 Jul 2019 12:13:59 +0200] [error] --- SQL error addEntry: 22001: 1406 Data too long for column 'content_bin' at row 1 while adding entry in feed 4 with title: Bankass
```

**What I expect:**:

The article is imported (maybe increase the content_bin column size?).

Alkarex:
Hello,
It looks like you are using MySQL, right? (Alternatively, it works with SQLite and PostgreSQL)
That column is declared as `blob` without any further restriction, so I am not sure what the problem is.

https://github.com/FreshRSS/FreshRSS/blob/caeeeb52cafe105c157f838397ade70bc8609900/app/SQL/install.sql.mysql.php#L72

I would have to investigate further. At a quick glance, the max length can be tuned in the MySQL parameters https://dev.mysql.com/doc/refman/8.0/en/blob.html but the defaults (16MB) are high enough, since your article is only about 374kB.

https://github.com/FreshRSS/FreshRSS/blob/caeeeb52cafe105c157f838397ade70bc8609900/app/Models/EntryDAO.php#L155-L211

nicofrand:
Yes I do use MySQL. On another instance (this one is a dev instance, the other is on a yunohost install still with mysql) the article was imported but without content, but it was imported a long ago with probably an older FreshRSS version.

I thought a `BLOB` was 64Kb and `MEDIUMBLOB` 16MB…

I might have an old mysql version too: `mysql  Ver 14.14 Distrib 5.7.26, for Linux (x86_64) using  EditLine wrapper`.

I'll try to see what length my version allows.

Alkarex:
@nicofrand Ah, I think you are correct. We would need to move to `MEDIUMBLOB` for such long articles. https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html

Alkarex:
While waiting for a change, you should be able to manually update your `*entry` and `*entrytmp` tables

nicofrand:
I confirm setting both to `MEDIUMBLOB` fixes it.

Thanks!

Alkarex:
[...]

PR #2551, opened 2019-10-04
MySQL entry content MEDIUMBLOB

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2448
~~Need testing~~


Frenzie (approved):
lgtm

<del>Do you have an example of a something so large it wouldn't fit btw? Just curious. ;-)</del>

Edit: silly question.

Issue #2549, opened 2019-10-02
Postgres SSL connection string


xoxys:
I've tried to set up FreshRSS with Postgres SSL connection. Therefore, I have added following to the config file:
```
'host': postgres.example.com;port=5432;sslmode=verify-full;sslrootcert=/etc/ssl/certs/ca-certificates.crt
```

but this does not work, because the complete path is removed by some filter functions. Running e.g. `/usr/bin/php -f /var/www/app/app/actualize_script.php` results in:
```
PHP Fatal error:  Uncaught Minz_PDOConnectionException: Access to database is denied for `pgfreshrss` (`pgsql:host=postgres.example.com;port=5432;sslrootcert=;sslmode=verify-full;dbname=freshrss`) in /var/www/app/lib/Minz/ModelPdo.php:95
```

As you can see, the path to the root ca was completely dropped (for whatever reason). As a workaround I created a symlink of /etc/ssl/certs/ca-certificates.crt into FreshRSS root dir and simply pass `sslrootcert=ca-certificates.crt` to the connection string. 

While this workaround works as expected, it would be cool to get a better Postgres ssl support.

Alkarex:
For reference: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS
I think we should expose a new parameter (not `host`) to add additional connection parameters, both in the Web install and CLI install.

xoxys:
Cool! Thanks for your fast response 🙂

Alkarex:
@xoxys Could you please test this fix? https://github.com/FreshRSS/FreshRSS/pull/2552
It should allow you to add in your `data/config.php` something like:

```php
...
db => [
...
'connection_uri_params': 'sslmode=verify-full;sslrootcert=/etc/ssl/certs/ca-certificates.crt',
...
],
...
```

Alkarex:
P.S.: I can merge in our /dev branch if it makes it easier to test for you

xoxys:
@Alkarex cool. No thats fine, no merge needed. I have to rebuild my own docker container and will patch the file manually. Thanks for your work!

xoxys:
[...]

PR #2552, opened 2019-10-05
Add optional database connection URI parameters

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2549


Alkarex:
Note for later: We are not yet exposing additional PDO or URI database options during the install process (neither Web nor CLI), but we should do so.
Currently, one has to edit `config.default.php` prior to install to use those advanced options.

Issue #2557, opened 2019-10-05
Using systemd timer to refresh feeds


Offerel:
Hi there. Some system didn't use a cron service, but most of them use systemd. It's easy to use systemd for such cron like features. As example here are 2 services which work together. First you need a freshrss.timer file in /etc/systemd/system/ with the following content:

```
[Unit]
Description=FreshRSS get new content

[Timer]
OnBootSec=30s
OnCalendar=*:0/15

[Install]
WantedBy=timers.target
```
Additionally you need the freshrss.service file in the same directory. Its started 30 seconds after boot and then in a 15 minute interval:

```
[Unit]
Description=FreshRSS get new content
Wants=freshrss.timer

[Service]
User=www-data
Type=simple
ExecStart=/usr/bin/php /path/to/freshrss/app/actualize_script.php > /var/log/FreshRSS.log 2>&1
```
After creating both files, you should reload systemd with `systemctl daeamon-reload` and enable it with `systemctl enable freshrss.timer`. And please make sure, to change the path to your installation in freshrss.service. Maybe this is useful for some people.

Alkarex:
@Offerel Thanks for the info 👍 
Would you feel like making a PR for our documentation with the same content, for instance in the following file or a file next to it with a reference?
https://github.com/FreshRSS/FreshRSS/blob/dev/docs/en/admins/01_Index.md

Offerel:
PR is done. I have tried to make also a PR some days before for my FR #2434

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/2605

PR #2561, opened 2019-10-06
Added informations for systemd service

Offerel:


Alkarex:
@Offerel Could you please make a new PR for the icon? So that we keep this one for systemd?
By the way, the reason I have not merged yet is because I would like this info to be in our docs https://github.com/FreshRSS/FreshRSS/issues/2557#issuecomment-538683148 as apposed to our readme (it would be fine with a link from the readme, though)

Alkarex:
And please remember to checkout our /dev branch for making changes, not /master

Offerel:
Sorry for the confusion I've made. I did the whole thing about the GUI of GitHub on the web and unfortunately also on the master branch. I have now checked out the dev branch and will implement future changes there. Could it be that the WebGUI always summarizes later commits in the previous PR? Or am I just doing something wrong? At home I mostly use Visual Studio Code and let it do the commits for my own projects, I only use this WebGUI on the road.

Alkarex:
No worries. You first need to make a new branch (for instance, branching from /dev into /systemd-doc) , and then you can make as many edits there as you want, which will be in the same PR.

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/2605

Issue #1890, opened 2018-05-20
Import of exported subscriptions fail.


bartlomiejcieszkowski:
```
PHP message: PHP Fatal error:  Uncaught Error: Call to a member function id() on null in /opt/freshrss/freshrss/app/Controllers/importExportController.php:428
Stack trace:
#0 /opt/freshrss/freshrss/app/Controllers/importExportController.php(93): FreshRSS_importExport_Controller->importJson('{\n    "id": "us...', true)
#1 /opt/freshrss/freshrss/app/Controllers/importExportController.php(143): FreshRSS_importExport_Controller->importFile('freshrss_2018-0...', '/tmp/phpTQ99qz')
#2 /opt/freshrss/freshrss/lib/Minz/Dispatcher.php(118): FreshRSS_importExport_Controller->importAction()
#3 /opt/freshrss/freshrss/lib/Minz/Dispatcher.php(47): Minz_Dispatcher->launchAction('importAction')
#4 /opt/freshrss/freshrss/lib/Minz/FrontController.php(84): Minz_Dispatcher->run()
#5 /opt/freshrss/freshrss/p/i/index.php(46): Minz_FrontController->run()
#6 {main}
```

Steps to reproduce:
1. Import attached zip (simple export  with both checkboxes - favourites and list of feeds)

[freshrss_2018-05-20_export.zip](https://github.com/FreshRSS/FreshRSS/files/2020410/freshrss_2018-05-20_export.zip)

It failed on import of json with starred - content is:
```
{
    "id": "user/bart/state/org.freshrss/starred",
    "title": "List of favourite articles",
    "author": "bart",
    "items": [

]}
```

Alkarex:
Thanks for the report, @bartlomiejcieszkowski

upils:
Same issue here with a `.opml` file directly exported from TTRSS.

Alkarex:
Thanks @upils 
Will fix ASAP.

Alkarex:
Should be fixed in https://github.com/FreshRSS/FreshRSS/pull/1901
Tests / feedback appreciated.

Alkarex:
@upils Thanks for the test. It is a related error, but not the same :-) Could you please tell a bit more about your setup? In particular: what database? Empty installation or existing data? Would you have the possibility of sharing a subset of your OPML triggering the problem?

Alkarex:
[...]

PR #2563, opened 2019-10-08
Increase import size

Alkarex:
This is merely a temporary workaround to allow at least some medium size
imports
https://framateam.org/freshrss/pl/7wbt4tcyetrfmris9xdcbq7uuw
The import module should be rewritten to process files one by one and as
data streams instead of loading multiple copies of the whole dataset in
memory as is the case now :-(
https://github.com/FreshRSS/FreshRSS/issues/1890
Note that the new SQLite export/import is distinct from this case.

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -29,7 +29,24 @@ class FreshRSS_importExport_Controller extends Minz_ActionController {
 		Minz_View::prependTitle(_t('sub.import_export.title') . ' · ');
 	}
 
+	private static function megabytes($size_str) {
+		switch (substr($size_str, -1)) {
+			case 'M': case 'm': return (int)$size_str;
+			case 'K': case 'k': return (int)$size_str / 1024;
+			case 'G': case 'g': return (int)$size_str * 1024;
+		}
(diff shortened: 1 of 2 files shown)


app/Controllers/importExportController.php
```diff
@@ -29,7 +29,24 @@ public function indexAction() {
 		Minz_View::prependTitle(_t('sub.import_export.title') . ' · ');
 	}
 
+	private static function megabytes($size_str) {
+		switch (substr($size_str, -1)) {
+			case 'M': case 'm': return (int)$size_str;
+			case 'K': case 'k': return (int)$size_str / 1024;
+			case 'G': case 'g': return (int)$size_str * 1024;
```
Frenzie, line 36: This is the biggest type it supports? No terabytes or something? (Admittedly less likely for RAM.) :-)

Frenzie (approved):
lgtm

app/Controllers/importExportController.php
Alkarex: Apparently, according to https://php.net/function.ini-get

Alkarex:
@squromiv

Issue #2568, opened 2019-10-15
Reading view Icon broken in Firefox (Swage Theme)


Offerel:
It seems that the Reading view icon in Firefox is broken:

![Firefox 69.0.3](https://user-images.githubusercontent.com/14041522/66806599-3ea14f00-ef28-11e9-8b56-3eb5a70caa68.jpg "Firefox 69.0.3")
![Chrome](https://user-images.githubusercontent.com/14041522/66806642-57116980-ef28-11e9-992f-aaed308193c2.jpg "Chrome")

Alkarex:
@pattems Could you maybe have a look?

Offerel:
Since im not at my home pc, i could fix the icon with http://petercollingridge.appspot.com/svg-optimiser

i make a PR

PR #2376, opened 2019-04-16
Fix Sage theme SVG refresh

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2375
`h-2,c` is not OK: It should be either `h-2 c` or `h-2c`

PR #2569, opened 2019-10-15
Fix for broken reading view icon in Firefox

Offerel:
Fix for #2568


Alkarex:
Similar bug than https://github.com/FreshRSS/FreshRSS/pull/2376

Alkarex:
Thanks! Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md

PR #2571, opened 2019-10-16
Fix for broken read icon in Firefox

Offerel:
It seems that the read icons shape has also some error. I fixed them like same way as with the reader view icon. Now the icon looks identical in Chrome and other browsers.


Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2568

PR #2578, opened 2019-10-19
Reeder added native support

stoinov:
Added Reeder as supported in Reader page
Clarified how to use the Fever API with Reeder
Fixed some typos on both pages


stoinov:
@Alkarex I only updated the documentation, but I got those PHP errors... any suggestions?

Alkarex:
Thanks 👍
Please remember to target the /dev branch and not /master

docs/en/users/06_Fever_API.md
```diff
@@ -6,7 +6,7 @@ and general aspects of API access.
 ## RSS clients
 
 There are many RSS clients existing supporting Fever APIs but they seem to understand the Fever API a bit differently.
-If your favourite client does not work properly with this API, create an issue and we will have a look.
+If your favorite client does not work properly with this API, create an issue and we will have a look.
```
Alkarex: That was not a typo, just UK spelling 🤓

stoinov: I know, I just made it more international :) You can revert this if you want to keep it consistent with other documentation.

docs/en/users/06_Fever_API.md
```diff
@@ -26,10 +26,10 @@ Tested with:
 - iOS
   - [Fiery Feeds](https://itunes.apple.com/app/fiery-feeds-rss-reader/id1158763303)
   - [Unread](https://itunes.apple.com/app/unread-rss-reader/id1252376153)
-  - [Reeder-4](https://itunes.apple.com/app/reeder-4/id1449412357)
+  - [Reeder 4](https://itunes.apple.com/app/reeder-4/id1449412357) when using self-hosted Fever option
```
Alkarex: I have not tried myself, but assuming it is possible, we should have a note telling to favour the Google  Reader API option

stoinov: I've been using it for years and it works great. There is one bug in the current update - the native FreshRSS implementation, if I add just the host, it's converting to full URL with /r/api/greader.php which does not work. I already wrote an email to fix this.

So Fever still works, it's just that the native uses greader.

Alkarex:
Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/dev/CREDITS.md

docs/en/users/06_Fever_API.md
Alkarex: Reeder's developer has promised a fix for the URL autocompletion.

aliceout: I'm trying "FreshRSS implementation" in Reeder4 since yesterday, and I have no major problems to report!

I confirm tha Reeder convert itself URL!
But on my version, it convert "domaine.com" in `http://domain.com/p/api/greader.php`
Note that for me is /**p**/api//greader.php and not  /**r**/api/greader.php

A problem comes from the fact that it adds "http" and not "https", which can lead to a refusal (depending on the server configuration).
Mine refused while it is set to redirect by itself. 

But for the rest, I have no problems to report. 
Titles, texts, images and links can be loaded without any problems.
The categories are respected and I can edit my feeds.

PR #2584, opened 2019-10-22
docs: Remove french documentation

marienfressinaud:
As suggested in https://github.com/FreshRSS/FreshRSS/issues/2491, this PR removes the French documentation so we have less to (not) maintain. My feeling is that it will be easier to have English-only doc, but I'm opened to the discussion about keeping some of the files (e.g. README.fr.md or user doc)


Alkarex:
I think it would be a pity to remove it all. I regularly hear IT-interested French people with difficulties in English, and who see that as a barrier. But reducing it would be fine. I suggest:
* To keep the readme
* To keep the user documentation
  * And add links to the English version for missing sections.
* To delete the part for administrators
* To delete the part for developers

The documentation is also something some non-technical contributors can do, and it is nice to offer such tasks to people eager to contribute

Frenzie:
I maintain what I wrote [here](https://github.com/FreshRSS/FreshRSS/pull/2164#issuecomment-527372185). My obstacle is the current manual effort involved in keeping them in sync, not the step of translating. One proven solution is po4a. I'll gladly translate more of it to French. I suppose I should open a counter-PR to put words into actions.

The added value for devs (and perhaps admins) is much smaller, because without a certain basic comprehension of English you'd be missing out on a lot. However, the PHP dev docs are currently available in:

*   Brazilian Portuguese
*    Chinese (Simplified)
*    French
*    German
*    Japanese
*    Romanian
*    Russian
*    Spanish
*    Turkish

So in principle there's little holding back a French monolingual from contributing.

(Romanian is interesting to see there, but I'm guessing it's all volunteer-based.)

Frenzie:
I did a quick search to see if someone had already written a script for this kind of thing. For a presumably relevant discussion see https://github.com/tldr-pages/tldr/pull/2793

Alkarex:
Shall we then start by implementing https://github.com/FreshRSS/FreshRSS/pull/2584#issuecomment-545358412 and then await @Frenzie 's proposition of smarter system (if it is simple enough, that sounds good)?
Maybe we could take this discussion in parallel with FreshRSS UI translation (there, my main concern will be performance - the current solution is indeed very fast)

marienfressinaud:
I'm all inclined to have a more intelligent system, both for doc and UI. My concern for the moment is to improve the quality of the current documentation and to set up a simple workflow that would keep in sync the doc with the state of FreshRSS :) Then, indeed, having a more intelligent system to sync our doc in different languages is something I would be very happy to see coming!

I'll keep the readme and the user doc since I was a bit reluctant to remove them at first and you seem to agree to keep them ;) Maybe we could add for the moment a header to the pages to indicate that the doc might be not up to date and to take the English doc as reference?

Frenzie:
Here's a quick script based on the one I linked:

```bash
#!/bin/bash
# This script performs the following:
# - Generate configuration file for po4a (can be configured in CONFIGFILE)
# - Generate POT file from pages/<DIRECTORY>/*.md
# - Update PO files in i18n directory with POT file
# - Generate localized pages.XX/<DIRECTORY>/*.md (where XX is the language code)
# - Remove unneeded new lines from generated pages

# Name of the po4a configuration file
CONFIGFILE='po4a.conf'

# List of supported languages
LANGS=(fr)

# Check if po4a is installed
if [ -z "$(command -v po4a)" ]; then
	echo 'It seems that po4a is not installed on your system.'
	echo 'Please install po4a to use this script.'
	exit 1
fi

# Generate po4a.conf file with list of TLDR pages
echo 'Generating configuration file for po4a...'
{
	echo '# WARNING: this file is generated with translation-update.sh'
	echo '# DO NOT modify this file manually!'
	echo "[po4a_langs] ${LANGS[*]}"
	echo '[po4a_paths] i18n/templates/freshrss.pot $lang:i18n/freshrss.$lang.po'
} > $CONFIGFILE

for FILE in $(tree -f -i | grep ".md")
do
	echo [type: text] $FILE \$lang:\$lang/$(echo $FILE | cut -d"/" -f 2)/$(echo $FILE | cut -d"/" -f 3) opt:\"-o markdown\" opt:\"-M utf-8\" >> $CONFIGFILE
done

# Generate POT file, PO files, and pages.XX pages
echo 'Generating POT file and translated TLDR pages...'
po4a -k 0 --msgid-bugs-address 'https://github.com/FreshRSS/FreshRSS/issues' $CONFIGFILE
```

It doesn't have the paths right, but the basic idea is that as a translator you use [one of these](https://wiki.archlinux.org/index.php/list_of_applications#Translation_and_localization) to immediately see which strings need translating. Then you use one of the listed tools, which looks like this:

Screenshot_2019-10-23_15-28-35

For non-technical users the translation files can be synced with Transifex/Weblate, making it much easier for anyone to contribute. (Of course this requires some helper scripts in the CI.)

> Maybe we could take this discussion in parallel with FreshRSS UI translation (there, my main concern will be performance - the current solution is indeed very fast)

The current solution doesn't deal well with plurals. Something gettext-ish seems like a fairly obvious approach, also see <https://github.com/oscarotero/Gettext>. Where by gettext-ish I mean use xgettext with pot/po for translation tools, while the actual implementation can work however you want.

Of course there are other formats with similar tooling advantages like XLIFF, and to a lesser extent the Qt Linguist format. See [here](https://docs.weblate.org/en/latest/formats.html) for all formats supported by Weblate, but I shouldn't think that anything other than what we use and the three I already mentioned is of much interest.

Frenzie:
PS It looks like GetText has by far the best and most implementations in PHP.

Frenzie:
As a sidenote, I've found the French original to be quite helpful for improving the words and phrasing in the English translation. On top of which I've already encountered various todos that didn't exist in English yet. It's all quite straightforward when you can use a proper tool like Lokalize. ;-)

French is mostly still in sync anyway. It was just hard to keep it up manually so you couldn't focus on the actually fun part.

NB #2590 is just step one. The ultimate goal is to make it easier for less technical users to contribute. I've got a couple dozen active translators on KOReader many of whom wouldn't exactly appreciate the whole Git commit/GH pull request procedure. Another major advantage of Transifex/Weblate is that it sends you a notification about once a day (or less often if you configure it that way) for continuous localization. That would be a great boon, because tbh I always forget to update the Dutch translation here.

Anyway, like I wrote yesterday I think it should be possible to set up Weblate for the UI strings without changing much of anything about our current UI translation system. That could be a very attractive option.

marienfressinaud:
I close this one in favor of https://github.com/FreshRSS/FreshRSS/pull/2590

PR #2588, opened 2019-10-23
Better performance with yield

Alkarex:
Largely decrease the time to first byte, and memory consumption.
Before, we used to make several copies in memory of the whole list of
articles before sending them to the client. Now streamed as they are
processed.

Taking advantage of PHP 5.5+ https://github.com/FreshRSS/FreshRSS/pull/2495

⚠️ Remove the `min_posts_per_rss` configuration, which is a bit annoying to implement in this context, and not very useful anyway.


Alkarex:
Not very statistically robust, but just as an illustration.
Increase the number of articles per page and/or use a more modest server and/or a slower network to highlight the differences:

Before:
image

After:
image

Alkarex:
Let's continue testing this in /master
There should be no functional difference, besides a performance improvement on the Web UI.

PR #2589, opened 2019-10-23
Greener Travis?

Alkarex:
With PHP 7.4 just around the corner, I suggest to test only the minimum
and maximum PHP versions we support in Travis, to save resources
(greener).
For the translation check, with can test with max version -1 (or min version + 1, or median :-P) to run on
yet another PHP version.
It makes also the Travis tests faster to complete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: php
 php:
   - 5.6
-  - 7.0
-  - 7.1
-  - 7.2
   - 7.3
 
 install:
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
For the current tests there should be very little lost, although it seems like it might be possible `php -l` could miss something that's only problematic in, e.g., PHP 7.1.

Definitely not necessary for phpcs of course; that was more of a proof of concept for testing different versions and that was the only thing available at the time.

marienfressinaud (approved):
good idea :+1:

PR #2591, opened 2019-10-24
New environment variable COPY_LOG_TO_SYSLOG

Alkarex:
False by default.
Makes it easy to monitor all logs from Syslog or STDERR (e.g. `docker logs -f freshrss`).

### DIFF
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -15,10 +15,10 @@ class Minz_Log {
 	 * NOTICE erreurs mineures ou messages d'informations
 	 * DEBUG Informations affichées pour le déboggage
 	 */
-	const ERROR = 2;
-	const WARNING = 4;
-	const NOTICE = 8;
-	const DEBUG = 16;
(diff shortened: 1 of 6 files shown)


lib/Minz/Log.php
```diff
@@ -15,10 +15,10 @@ class Minz_Log {
 	 * NOTICE erreurs mineures ou messages d'informations
 	 * DEBUG Informations affichées pour le déboggage
 	 */
-	const ERROR = 2;
-	const WARNING = 4;
-	const NOTICE = 8;
-	const DEBUG = 16;
+	const ERROR = LOG_ERR;
```
Alkarex: @marienfressinaud I suggest removing those custom constants and use PHP native ones instead https://php.net/function.syslog
What do you think?

Alkarex: Merging to do my own final test before the call to testing, but feel free to reintroduce them if you miss them :-)

PR #2592, opened 2019-10-25
Add category order

aledeg:
Each category has a new 'priority' attribute. It is used to sort categories in
views. Categories with the same priority are sorted alphabetically. Categories
with no priority are displayed after those with one.

For example, if we have the following categories:
- A (priority: 2)
- B (no priority)
- C (priority: 1)
- D (priority: 2)
- E (no priority)
- F (priority: 1)

They will be displayed in the following order:
- C
- F
- A
- D
- B
- E

See #190

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -98,14 +98,14 @@ class FreshRSS_Category extends Minz_Model {
 	}
 
 	public function _attributes($key, $value) {
-		if ($key == '') {
+		if ('' == $key) {
 			if (is_string($value)) {
 				$value = json_decode($value, true);
 			}
 			if (is_array($value)) {
 				$this->attributes = $value;
 			}
-		} elseif ($value === null) {
+		} elseif (null === $value) {
 			unset($this->attributes[$key]);
 		} else {
 			$this->attributes[$key] = $value;
--- a/app/layout/aside_feed.phtml
+++ b/app/layout/aside_feed.phtml
@@ -63,11 +63,12 @@
 		<?php
 			foreach ($this->categories as $cat) {
 				$feeds = $cat->feeds();
+				$position = $cat->attributes('position');
 				if (!empty($feeds)) {
 					$c_active = FreshRSS_Context::isCurrentGet('c_' . $cat->id());
 					$c_show = $c_active || FreshRSS_Context::$user_conf->display_categories;
 		?>
-		<li class="tree-folder category<?= $c_active ? ' active' : '' ?>" data-unread="<?= $cat->nbNotRead() ?>">
+		<li class="tree-folder category<?= $c_active ? ' active' : '' ?>"<?= null === $position ? '' : "data-position='$position'" ?> data-unread="<?= $cat->nbNotRead() ?>">
 			<div class="tree-folder-title">
 				<a class="dropdown-toggle" href="#"><?= _i($c_show ? 'up' : 'down') ?></a>
 				<a class="title<?= $cat->hasFeedsWithError() ? ' error' : '' ?>" data-unread="<?= format_number($cat->nbNotRead()) ?>" href="<?= _url('index', $actual_view, 'get', 'c_' . $cat->id()) ?>"><?= $cat->name() ?></a>
--- a/app/Controllers/indexController.php
+++ b/app/Controllers/indexController.php
@@ -173,7 +173,7 @@ class FreshRSS_index_Controller extends Minz_ActionController {
 	private function updateContext() {
 		if (empty(FreshRSS_Context::$categories)) {
 			$catDAO = FreshRSS_Factory::createCategoryDao();
-			FreshRSS_Context::$categories = $catDAO->listCategories();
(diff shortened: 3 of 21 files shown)


app/layout/aside_feed.phtml
```diff
@@ -63,11 +63,12 @@
 		<?php
 			foreach ($this->categories as $cat) {
 				$feeds = $cat->feeds();
+				$priority = $cat->attributes('priority');
 				if (!empty($feeds)) {
 					$c_active = FreshRSS_Context::isCurrentGet('c_' . $cat->id());
 					$c_show = $c_active || FreshRSS_Context::$user_conf->display_categories;
 		?>
-		<li class="tree-folder category<?= $c_active ? ' active' : '' ?>" data-unread="<?= $cat->nbNotRead() ?>">
+		<li class="tree-folder category<?= $c_active ? ' active' : '' ?><?= null === $priority ? '' : " priority_$priority"?>" data-unread="<?= $cat->nbNotRead() ?>">
```
Alkarex: I have not checked in details, but I suggest using `data-priority="..."`

app/Models/Category.php
```diff
@@ -98,14 +98,14 @@ public function _feeds($values) {
 	}
 
 	public function _attributes($key, $value) {
-		if ($key == '') {
+		if ('' === $key) {
```
Alkarex: Probably `==` is better there. It would still be a problem with null and 0 and false

Alkarex:
Looks good and straightforward 👍

app/layout/aside_feed.phtml
aledeg: The idea here is to be able to target them with a CSS rule. But I can also add some data.

aledeg:
@Alkarex I think I should do something more user friendly by modifying the priority with drag&drop instead of manual entry.
What do you think?

app/layout/aside_feed.phtml
aledeg: @Frenzie I am not sure I follow you thinking. Could you explain me what you have in mind?

Frenzie: Sorry, I figured you knew. That's a CSS attribute selector.

https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

It's been supported by all browsers for close to 15 years.

aledeg: @Frenzie I knew but I didn't understand what you meant. Now I do :)

Alkarex:
@aledeg It would surely be better, if you are motivated enough :-P
Maybe we could have a first version for the 1.15 release and some improvements in 1.16.
To prepare for that, maybe `priority` should be renamed to something more related to ordering.

Frenzie:
Sort(ing) order? Position(ing)?

app/layout/aside_feed.phtml
Alkarex: We are already using CSS rules targetting data properties, e.g. https://github.com/FreshRSS/FreshRSS/blob/dev/p/themes/base-theme/template.css#L1023

aledeg:
@Alkarex 
I'll make it use drag&drop and change the naming. When do you plan to release v1.15? I would like to make it work fully before shipping.
Let me know

@Frenzie I'll go with *position* then

Frenzie:
Sounds good!

Alkarex:
I would like to call for testing FreshRSS 1.15 tomorrow and release it by the end October. Therefore a first simple version for now would be nice (i.e. more or less what you already have now), keeping the drag & drop improvements for the next release

aledeg:
The thing is that with drag&drop, there will be no use for the category configuration field in UI. Thus making the translation obsolete. We can still remove it later but we have to keep that in mind.
I'll make the few changes required tonight and I'll continue to work on it later. Hopefully not like my other PR :)

aledeg:
@Alkarex @Frenzie 
Everything is fixed except for the drag&drop

Alkarex:
Nice, I will try tomorrow morning :-)

Alkarex:
Looks all good :-) I have just shorten a bit the help text as it was not looking so good on different layouts

app/layout/aside_feed.phtml
```diff
@@ -63,11 +63,12 @@
 		<?php
 			foreach ($this->categories as $cat) {
 				$feeds = $cat->feeds();
+				$position = $cat->attributes('position');
 				if (!empty($feeds)) {
 					$c_active = FreshRSS_Context::isCurrentGet('c_' . $cat->id());
 					$c_show = $c_active || FreshRSS_Context::$user_conf->display_categories;
 		?>
-		<li class="tree-folder category<?= $c_active ? ' active' : '' ?>" data-unread="<?= $cat->nbNotRead() ?>">
+		<li class="tree-folder category<?= $c_active ? ' active' : '' ?>"<?= null === $position ? '' : "data-position='$position'" ?> data-unread="<?= $cat->nbNotRead() ?>">
```
Alkarex, line 71: Missing space before `data-position` fixed in https://github.com/FreshRSS/FreshRSS/pull/5203

PR #2540, opened 2019-09-28
Document Reeder compatibility

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2513
https://github.com/FreshRSS/FreshRSS/pull/2526

PR #2594, opened 2019-10-25
Reeder client for MacOs and iOS

aliceout:
- Adding "Reeder" clients for MacOS (Feever et Google-Reader API)
- Changing "Reeder" version (and link) for iOS


aledeg (approved):

Alkarex:
Thanks. I have resolved the conflicts with the /dev branch,
I suggest we only keep Reeder in the Google Reader API section, as that API is better than Fever API, even though Reeder can do both API, and both MacOS and iOS.
I do not have the possibility of testing Reeder myself, but feedback welcome, especially using the Google Reader API / native FreshRSS option.
See:
* https://github.com/FreshRSS/FreshRSS/pull/2578#discussion_r336729291
* https://github.com/FreshRSS/FreshRSS/pull/2540

Issue #2600, opened 2019-10-26
Max articles should be disabled by default


Frenzie:
Cf. https://github.com/FreshRSS/FreshRSS/pull/2599#issuecomment-546606015

Most important: it should be disabled for existing installations when upgrading. It's an extremely destructive setting that should not be enabled without user interaction. This goes together with the new 3 months similarly destructive setting.

I'm also strongly opposed making it a default for new installations since it's secretly destructive, but that's a slightly separate discussion from not harming existing installations.

Alkarex:
I agree for existing installations.
I think we should have default max/min/date limits for new installations, but we can discuss the values.

Alkarex:
Fixed by https://github.com/FreshRSS/FreshRSS/pull/2602

Frenzie:
Back in '05 I curiously tried Google Reader. I didn't care for it (besides that it was available everywhere), but the cherry on top was that 30 days later, it had started removing the first entries. I celebrated its demise.

Anyway, it's more about information than about the setting as such.

Alkarex:
Should be good this time https://github.com/FreshRSS/FreshRSS/pull/2607

Luxtech:
Is it possible to disable the max articles limit on a existing installation?

PR #2599, opened 2019-10-26
FreshRSS 1.15

Alkarex:
Hi fellow FreshRSS users!
After 3 months of quite intense development, the next [version 1.15 of FreshRSS](https://github.com/FreshRSS/FreshRSS/milestones/1.15.0) is finally ready for testing before release.

Several big new features this time, and even more under the hood.

In particular, it is now possible to export/import any FreshRSS account to/from a SQLite file. This also allows you to migrate database (e.g. from MySQL to PostgreSQL) easily, and do better backups or migration from one server to another. Please try it and report whether that works well for you.

The archiving strategy has been reworked and now offers the possibility to control a maximum number of articles per feed. Archiving settings (which are always applied at feed level) can be define for each feed, or by category, or globally. A few additional rules are hidden (for instance always keep the articles seen at the last feed refresh). Dates are to be understood as when an article has last been seen in the original RSS feed, not the publication date. Give it a shot!

Our iOS users have one more client to choose from (the first implementing the Google Reader API, which is more powerful than the clients using the Fever API), and a lot more, see below.

Thanks to the 19 contributors who made this new version :-)

Full [changelog](https://github.com/FreshRSS/FreshRSS/blob/dev/CHANGELOG.md):
* CLI
	* Command line to export/import any database to/from SQLite [#2496](https://github.com/FreshRSS/FreshRSS/pull/2496)
* Features
	* New archiving method, including maximum number of articles per feed, and settings at feed, category, global levels [#2335](https://github.com/FreshRSS/FreshRSS/pull/2335)
	* New option to control category sort order [#2592](https://github.com/FreshRSS/FreshRSS/pull/2592)
	* New option to display article authors underneath the article title [#2487](https://github.com/FreshRSS/FreshRSS/pull/2487)
	* Add e-mail capability [#2476](https://github.com/FreshRSS/FreshRSS/pull/2476), [#2481](https://github.com/FreshRSS/FreshRSS/pull/2481)
	* Ability to define default user settings in `data/config-user.custom.php` [#2490](https://github.com/FreshRSS/FreshRSS/pull/2490)
		* Including default feeds [#2515](https://github.com/FreshRSS/FreshRSS/pull/2515)
	* Allow recreating users if they still exist in database [#2555](https://github.com/FreshRSS/FreshRSS/pull/2555)
	* Add optional database connection URI parameters [#2549](https://github.com/FreshRSS/FreshRSS/issues/2549), [#2559](https://github.com/FreshRSS/FreshRSS/pull/2559)
[...]

PR #2602, opened 2019-10-26
keep_max disabled by default for existing installations

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2600

### DIFF
--- a/app/Models/Context.php
+++ b/app/Models/Context.php
@@ -59,13 +59,13 @@ class FreshRSS_Context {
 		}
 
 		$keepMin = (int)FreshRSS_Context::$user_conf->param('keep_history_default', -5);
-		if ($keepMin != 0 && $keepMin > -5) {	//Freshrss < 1.15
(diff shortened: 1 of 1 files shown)


Alkarex:
@Frenzie would you be able to test / double-check? Otherwise I suggest that we merge

Frenzie:
I can probably upgrade another test instance tonight.

Edit: not fully sure though.

Alkarex:
Ok, then let's merge for additional testing (it is in any case less risky than what is in /dev right now)

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2599

Frenzie:
I just updated 1.14 to the latest dev and it still looks like this:

Screenshot_2019-10-27 Archiving · FreshRSS

Alkarex:
Did not work well. Second attempt https://github.com/FreshRSS/FreshRSS/pull/2607

Frenzie:
Seems to be okay now.

PR #2616, opened 2019-10-31
Fix categories sorting

marienfressinaud:
I had a bug when some articles in a category were unread. Sometimes,
filtering the category displayed all the articles, even with the "Adjust
showing" option.

A reasonably good guess was to look the last commits that impacted the
categories and, indeed, it was introduced by the new category sorting
system.

The sort function (`usort`) doesn't keep the index association, which is
important here. I must admit I don't understand exactly why (I didn't
search), but the keys are probably used somewhere. Another fix would be
to not rely on these keys anymore.

The bug was introduced in 042fcd5e93884690efdd2f957ca02751958c0724

- Ref PR: https://github.com/FreshRSS/FreshRSS/pull/2592
- Ref usort: https://www.php.net/manual/en/function.usort.php
- Ref uasort: https://www.php.net/manual/en/function.uasort.php

### DIFF
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -208,7 +208,7 @@ class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable 
 			return $categories;
 		}
 
-		usort($categories, function ($a, $b) {
+		uasort($categories, function ($a, $b) {
 			$aPosition = $a->attributes('position');
 			$bPosition = $b->attributes('position');
 			if ($aPosition === $bPosition) {


Frenzie (approved):
Cool, thanks.

Alkarex:
It would be nice to find the details of the problem

Alkarex:
Ah, sure, the keys were the category ID, and `usort` broke that.

https://github.com/FreshRSS/FreshRSS/blob/4058898de2006cb4ee3b4565bd3e0235b4d0f0d9/app/Models/CategoryDAO.php#L407

Used e.g. by:

https://github.com/FreshRSS/FreshRSS/blob/4058898de2006cb4ee3b4565bd3e0235b4d0f0d9/app/Models/Context.php#L215-L216

Issue #2619, opened 2019-11-01
Data deletion during 1.15.0 update


RemiCab:
Hello,

After receiving the rss post about the 1.15.0 update, I decided to do it (using the UI). After clicking on the "update" button, I was redirected to the 403 error file. According to the nginx logs, I discovered that the users directories (in data/users) and the config.php file had been deleted during the process.

I have restored these files using my backup and regardless this point the update seems to be correct.

Nevertheless, is this deletion a normal part of this particular update? If yes, it seems important to inform administrators of this behavior. Especially because the UI gives the impression of a smooth process and in this case it was not really the case.

If it is a bug, here are the technical details:

- Raspbian 10
- Nginx 1.14.2
- php 7.3.11-1
- PostgreSQL 11.5
- update FreshRSS 1.14.3 -> 1.15.0

I can't provide any FreshRSS logs since they were in the deleted directories.

Nginx logs:

```
2019/11/01 15:18:51 [error] 6309#6309: *9798 FastCGI sent in stderr: "PHP message: PHP Warning:  touch(): Unable to create file /srv/FreshRSS/data/users/***_adm/log.txt because No such file or directory in /srv/FreshRSS/lib/lib_rss.php on line 317" while reading response header from upstream, client: 188.239.152.230, server: rss.xxx.yy request: "GET /i/?c=update&a=apply&post_conf=1 HTTP/2.0", upstream: "fastcgi://unix:/var/run/php/php7.3-fpm.sock:", host: "rss.xxx.yy"
```

Frenzie:
I'm tagging this as important because it definitely sounds like a bug. However, nothing of the sort happened to me when I tested updating a few different systems.

Could there potentially have been a lack of diskspace or something of the kind involved?

RemiCab:
I have checked it and the disk where FreshRSS is located is full at 51% so that is not the issue.

Could this be a git related problem? I mean, as the user data and config files are not tracked by git, is it possible that the update process has removed all to apply the repository state of the data directory?

Frenzie:
It seems to execute a `git clean` since #2036 when using Git , which I believe would indeed remove untracked files.

https://github.com/FreshRSS/FreshRSS/blob/3aa66f317b496ccd9a2df914bbc747c52081a7ad/app/Controllers/updateController.php#L35

Alkarex:
@Reman38 No, those files are not supposed to be deleted.

Did you install by git?
If yes, could you please verify that you have the following files?
[...]

PR #2625, opened 2019-11-02
Git fetch+reset instead of clean+fetch+merge

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2619
Avoid potentially dangerous git clean, and use more robust fetch + reset strategy instead

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -29,20 +29,22 @@ class FreshRSS_update_Controller extends Minz_ActionController {
 	public static function gitPull() {
 		$cwd = getcwd();
 		chdir(FRESHRSS_PATH);
-		$output = array();
+		$output = '';
 		$return = 1;
 		try {
-			exec('git clean -f -d -f', $output, $return);
+			exec('git fetch', $output, $return);
 			if ($return == 0) {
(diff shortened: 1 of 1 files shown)


Alkarex:
Just tested with various conflicts and seems to work fine

app/Controllers/updateController.php
```diff
@@ -29,20 +29,22 @@ public static function hasGitUpdate() {
...
-			Minz_Log::warning('git pull error:' . $e->getMessage());
+			Minz_Log::warning('Git error:' . $e->getMessage());
+			if ($output == '') {
+				$output = $e->getMessage();
+			}
+			$return = 1;
 		}
 		chdir($cwd);
+		deleteInstall();
 		$line = is_array($output) ? implode('; ', $output) : '' . $output;
```
marienfressinaud, line 48: `$output` is no longer an array here if I'm not wrong

marienfressinaud (approved):
Besides my comment, it looks good to me :+1:

app/Controllers/updateController.php
Alkarex: It risks being an array from `exec()`

PR #2630, opened 2019-11-04
Cookie same-site

Alkarex:
* https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
* https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7
* https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/
* https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

We can discuss whether it should be *Lax* or *Strict*

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -320,7 +320,7 @@ server {
 	# Other SSL stuff goes here
 
 	# Needed for Freshrss cookie/session :
-	proxy_cookie_path / "/; HTTPOnly; Secure; SameSite=Lax";
+	proxy_cookie_path / "/; HTTPOnly; Secure; SameSite=Strict";
 
 	location / {
 		try_files $uri $uri/ =404;
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -320,7 +320,7 @@ server {
 	# Other SSL stuff goes here
 
 	# Needed for Freshrss cookie/session :
-	proxy_cookie_path / "/; HTTPOnly; Secure; SameSite=Strict";
+	proxy_cookie_path / "/; HTTPOnly; Secure; SameSite=Lax";
 
 	location / {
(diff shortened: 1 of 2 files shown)


Alkarex:
Assuming we use form-based login, with a cookie:
* With *Lax* , we can link from another domain to any FreshRSS page, e.g. direct link to a category. It is a relatively minor change compared to what we have now. I believe it will be the default from Chrome 80.
* With *Strict*, we can only link to FreshRSS root (otherwise one has to log-in again), but it is a safer policy. Being able to link to the root at all is made possible thanks to the cookie isolation we have with `/i/` and its redirection

Thoughts?

Alkarex:
After thinking a bit, I believe that *Lax* would be safer for a x.1 release, and we can discuss the possibly upgrade to *Strict* for the next bigger release, e.g. 1.16.0

marienfressinaud:
I don't have the motivation to think about it tonight… but it is important to have for the next 1.15.1?

Alkarex:
*Lax* should be quite safe, especially since it is becoming the default value anyway. 
But it can wait a bit 🙂

marienfressinaud:
Ok, I gave a thought about this. If I understand, in _Strict_ mode, if any external site link to, let's say the configuration page, the user will have to log in again. If so, I'm not really for the _Strict_ mode and I would stay with _Lax_.

Besides this opinion, _Lax_ looks safe indeed and I'm now inclined to merge this PR for the 1.15.1 ;)

Frenzie:
If I understand @Alkarex's summary correctly (don't have time to read the provided links atm) then strict would break functionality like the bookmarklet?

Alkarex:
@Frenzie That is correct, I did not think about it

PR #2627, opened 2019-11-02
Default or custom OPML

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2075
Replaces https://github.com/FreshRSS/FreshRSS/pull/2515 / https://github.com/FreshRSS/FreshRSS/issues/2514

Uses the local `./data/opml.xml` if it exists, otherwise `./opml.default.xml`

PR #2632, opened 2019-11-04
fix: Gitignore opml.xml

marienfressinaud:
This file is used to add default feeds to new users. It was added in
7819a43197d34ef7a6c5626e9e48d7db075c37c9

Problem is that the file is not ignored by Git and risk to be deleted
during a git update.


Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/2627

Issue #2618, opened 2019-11-01
Favorite affects whole article display


Frenzie:
v1.15

Same as #2604. There might also be a minor issue with hover in some circumstances.

Alkarex:
@Frenzie Let's issue a [fast release 1.15.1](https://github.com/FreshRSS/FreshRSS/milestones/1.15.1) for fixing a couple of regressions. It would be nice if that could include this issue

Frenzie:
Sure thing, this shouldn't take long. Not this afternoon though. :)

marienfressinaud:
Do you think you'll have time tonight? :) Otherwise, I'll work on it tomorrow morning so we can release the 1.15.1 in the afternoon or evening

Frenzie:
@marienfressinaud My spare time is fairly occupied until Wednesday (Bill Bailey tomorrow ;-) ), so feel free to go ahead if you want it done tomorrow.

PR #2477, opened 2019-08-04
[CI] Run stylelint

Frenzie:
Perform some basic CSS sanity checking and style enforcement. Closes #2475.

I removed vendor prefixed linear-gradient and transform because those are from the IE9 era. With IE11 as a minimum and soon to be obsolete requirement it doesn't make much sense anymore.

<hr>

Some files were a bit of a mess indentation-wise (tabs mixed with spaces, inconsistent, etc.) which is why their diffs show up so large. The basic formatting (e.g. a space after `:` or not) can be changed at will by running stylelint with the `--fix` flag.

PR #2634, opened 2019-11-05
Fix favorite display for "current" articles

marienfressinaud:
Favorite articles have, in most of the themes, a gold background to
distinguish them from the other articles. However, it can be distracting
to have such a background when reading the articles, so we should turn
them back to the "default" background when articles are opened (class
`.current`).

Bug introduced in e9ce27d8d991d7806ca2c2af7e5282279e378885

I didn't apply the fix to the Ansum and Mapco designs because they
already had this behaviour before the commit.

Related PRs:

- https://github.com/FreshRSS/FreshRSS/pull/2477
- https://github.com/FreshRSS/FreshRSS/pull/2611
- https://github.com/FreshRSS/FreshRSS/pull/2612

Closes https://github.com/FreshRSS/FreshRSS/issues/2618

### DIFF
--- a/p/themes/Ansum/_list-view.scss
+++ b/p/themes/Ansum/_list-view.scss
@@ -20,24 +20,19 @@
 		border-left-color: $main-first;
 	}
 
-	&.not_read {
+	&.not_read:not(.current) {
 		background: $unread-bg; //--------------------
-		// border-left-color: #FF5300;
-		&:hover {
-			background: $unread-bg-light; //--------------------
+
+		&:hover .item.title {
+			background: $unread-bg;
 		}
+	}
 
-		&:not(.current):hover .item.title {
-			background: $unread-bg-light;
-
-
-		}
-
+	&.not_read {
 		.item.title {
 			a {
 				color: $unread-font-color; //--------------------
 			}
-
 		}
 
 		.item.website {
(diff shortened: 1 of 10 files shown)


aledeg (approved):

Frenzie:
> I didn't apply the fix to the Ansum and Mapco designs because they
already had this behaviour before the commit.

I very much doubt it's on purpose. I'd change it. :-)

Anyway, looks fine. Did you also test the hover behavior? It doesn't really matter if you didn't since it'll be a positive commit either way, but otherwise someone would have to do a minor follow-up still. (I already volunteered myself a few days ago.)

marienfressinaud:
I fixed Ansum and Mapco as well.

I wasn't sure of what should be the behaviour on hover: for the themes I patched, there is no change when the article is unread or favorite. I find it better for Ansum and Mapco because the background of the line and the background of the title are a bit de-synchronized due to the animation (at least on Firefox)

Frenzie:
> I wasn't sure of what should be the behaviour on hover: for the themes I patched, there is no change when the article is unread or favorite.

Whatever the behavior was in 1.14 should be fine. The most important thing is that you don't want to see anything ugly like this on hover:

Screenshot_2019-11-05_15-52-06

marienfressinaud:
Yep, I fixed that :)

PR #2635, opened 2019-11-05
Fix database autocreate at install

Alkarex:
Several bugs prevented the auto-creation of the database in Web and CLI installs.
Fix https://github.com/YunoHost-Apps/freshrss_ynh/issues/84#issuecomment-549818408

### DIFF
--- a/app/Models/FeedDAO.php
+++ b/app/Models/FeedDAO.php
@@ -102,7 +102,9 @@ class FreshRSS_FeedDAO extends Minz_ModelPdo implements FreshRSS_Searchable {
 				'httpAuth' => $feed->httpAuth(),
 				'attributes' => $feed->attributes(),
 			);
-			if ($feed->mute() || $feed->ttl() != FreshRSS_Context::$user_conf->ttl_default) {
+			if ($feed->mute() || (
+				FreshRSS_Context::$user_conf != null &&	//When creating a new user
+				$feed->ttl() != FreshRSS_Context::$user_conf->ttl_default)) {
 				$values['ttl'] = $feed->ttl() * ($feed->mute() ? -1 : 1);
 			}
 
--- a/app/install.php
+++ b/app/install.php
@@ -164,7 +164,7 @@ function saveStep2() {
 		$ok = false;
 		try {
 			Minz_Session::_param('currentUser', $config_array['default_user']);
-			$ok = checkDb();
+			$ok = initDb();
 			Minz_Session::_param('currentUser');
 		} catch (Exception $ex) {
 			$_SESSION['bd_error'] = $ex->getMessage();
--- a/app/install.php
+++ b/app/install.php
@@ -132,7 +132,7 @@ function saveStep2() {
 		$config_array = [
 			'salt' => generateSalt(),
 			'base_url' => $base_url,
-			'default_user' => 'admin',
+			'default_user' => '_',
 			'db' => [
 				'type' => $_SESSION['bd_type'],
 				'host' => $_SESSION['bd_host'],
@@ -154,12 +154,18 @@ function saveStep2() {
 		@unlink(DATA_PATH . '/config.php');	//To avoid access-rights problems
 		file_put_contents(DATA_PATH . '/config.php', "<?php\n return " . var_export($config_array, true) . ";\n");
 
+		if (function_exists('opcache_reset')) {
+			opcache_reset();
+		}
+
 		Minz_Configuration::register('system', DATA_PATH . '/config.php', FRESHRSS_PATH . '/config.default.php');
 		FreshRSS_Context::$system_conf = Minz_Configuration::get('system');
 
 		$ok = false;
 		try {
+			Minz_Session::_param('currentUser', $config_array['default_user']);
 			$ok = checkDb();
+			Minz_Session::_param('currentUser');
 		} catch (Exception $ex) {
 			$_SESSION['bd_error'] = $ex->getMessage();
 			$ok = false;
--- a/cli/do-install.php
+++ b/cli/do-install.php
@@ -93,7 +93,7 @@ Minz_Session::_param('currentUser', $config['default_user']);
 
 $ok = false;
 try {
-	$ok = checkDb();
+	$ok = initDb();
 } catch (Exception $ex) {
 	$_SESSION['bd_error'] = $ex->getMessage();
 	$ok = false;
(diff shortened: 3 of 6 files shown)


Alkarex:
Tested by @plopoyop

lib/lib_install.php
```diff
@@ -85,14 +85,21 @@ function checkDb() {
 		$db['pdo_options'] = [];
 	}
 	$db['pdo_options'][PDO::ATTR_ERRMODE] = PDO::ERRMODE_EXCEPTION;
-	$dbBase = isset($db['base']) ? $db['base'] : '';
+	$conf->db = $db;	//TODO: Remove this Minz limitation "Indirect modification of overloaded property"
```
marienfressinaud, line 88: I'm not sure why you need to re-set the $db in conf, but I would avoid this side effect. But maybe we can't do differently due to the actual architecture?

marienfressinaud (commented):
To complete my comment, I would appreciate a more detailed commit message to explain the changes (at least what bugs are you patching?) I'm pretty ok with the PR, but in the same time I don't feel like my review is really useful since I'm just guessing the problem(s) :(

lib/lib_install.php
Alkarex: Exactly: `FreshRSS_Context::$system_conf->db['key'] = 'value';` does not work due to the fact that Minz introduces a barrier with `__get()`.
The error message is in the comment: *Indirect modification of overloaded property*

app/Models/FeedDAO.php
```diff
@@ -102,7 +102,9 @@ public function addFeedObject($feed) {
 				'httpAuth' => $feed->httpAuth(),
 				'attributes' => $feed->attributes(),
 			);
-			if ($feed->mute() || $feed->ttl() != FreshRSS_Context::$user_conf->ttl_default) {
```
Alkarex, line 105: `FreshRSS_Context::$user_conf` is not available when creating a new user

app/install.php
```diff
@@ -132,7 +132,7 @@ function saveStep2() {
 		$config_array = [
 			'salt' => generateSalt(),
 			'base_url' => $base_url,
-			'default_user' => 'admin',
+			'default_user' => '_',
```
Alkarex, line 135: So that we can *touch* e.g. an SQLite database in the `./data/users/_/` folder, which is the only one we have so early in the install process

app/install.php
```diff
@@ -154,12 +154,18 @@ function saveStep2() {
 		@unlink(DATA_PATH . '/config.php');	//To avoid access-rights problems
 		file_put_contents(DATA_PATH . '/config.php', "<?php\n return " . var_export($config_array, true) . ";\n");
 
+		if (function_exists('opcache_reset')) {
+			opcache_reset();
```
Alkarex, line 158: We should always clear opcache before reading a dynamically generated PHP file.

app/install.php
```diff
@@ -154,12 +154,18 @@ function saveStep2() {
...
+		}
+
 		Minz_Configuration::register('system', DATA_PATH . '/config.php', FRESHRSS_PATH . '/config.default.php');
 		FreshRSS_Context::$system_conf = Minz_Configuration::get('system');
 
 		$ok = false;
 		try {
+			Minz_Session::_param('currentUser', $config_array['default_user']);
```
Alkarex, line 166: This is before we actually have the name of a real user, so use default one

cli/do-install.php
```diff
@@ -82,7 +82,24 @@
 	fail('FreshRSS could not write configuration file!: ' . join_path(DATA_PATH, 'config.php'));
 }
 
-if (!checkDb()) {
+if (function_exists('opcache_reset')) {
+	opcache_reset();
+}
+
+Minz_Configuration::register('system', DATA_PATH . '/config.php', FRESHRSS_PATH . '/config.default.php');
+FreshRSS_Context::$system_conf = Minz_Configuration::get('system');
```
Alkarex, line 90: The configuration was not loaded

lib/Minz/ModelPdo.php
```diff
@@ -28,6 +28,9 @@ public function __construct($currentUser = null, $currentPdo = null) {
 		if ($currentUser === null) {
 			$currentUser = Minz_Session::param('currentUser');
 		}
+		if ($currentUser == '') {
+			throw new Minz_PDOConnectionException('Current user must not be empty!', '', Minz_Exception::ERROR);
```
Alkarex, line 32: Another barrier to prevent logic from running without a proper user. The other bugs fixed in the PR allowed coming here with a blank user

lib/lib_install.php
```diff
@@ -85,14 +85,21 @@ function checkDb() {
...
 
-	$db['base'] = '';	//First connection without database name to create the database
-	Minz_ModelPdo::$usesSharedPdo = false;
-	$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
-	$databaseDAO->create();
+	if ($db['type'] !== 'sqlite') {
```
Alkarex, line 90: There is no global database to create for SQLite

Alkarex:
@marienfressinaud I have added some comments

marienfressinaud (approved):
Thanks for the comments! Even if I made some comments, it's generic and out of the scope.

app/install.php
Alkarex: Yes, maybe by using only the Context object, and if so it would not be a big change

lib/lib_install.php
Alkarex: Indeed for changing the name. And yes for changing the configuration, as we are not really using its advantages (setters and getters) but we pay its drawbacks (layers that make the syntax more complex and error prone).
The comment was more to explain the strange syntax

Alkarex: Name changed https://github.com/FreshRSS/FreshRSS/pull/2635/commits/41179bef9f70e233baadd3b8c4bc6974d56e02e4

lib/Minz/ModelPdo.php
X-dark: This breaks cli (cf. #2644 ) Removing this two lines is enough to make cli works again.

Alkarex: @X-dark Thanks for the heads up

Alkarex: Fixed by https://github.com/FreshRSS/FreshRSS/pull/2646

Issue #2637, opened 2019-11-06
Category disappeared


Tealk:
Since the update to 1.15.x I am missing the category "Changelog" if I want to add it again I get the message "The category name already exists".

The feeds that were in "Changelog" are now in "Uncategorized".

Debian 10, Apache2, PHP7.3
downloaded/install via git

Alkarex:
Do you maybe have a tag with that same name?

Tealk:
I never have use Tag's
On the Reeder4 app I see the category, but not on the website.

Alkarex:
Did you maybe change the name of the default category before? ("Uncategorized")
In any case, try to organise the feeds and categories so that they are correct on the Web site.
If this is not synchronised to Reeder, then there is a bug to report to Reeder, but you can try to log out / log in.

Tealk:
So I have these categories, since I have been using FreshRSS for quite some time now, I can't tell if I have renamed a category.
![image](https://user-images.githubusercontent.com/12276250/68298896-69e40d80-0092-11ea-8c84-8daebe1c3bdc.png)

//Edit
relog has no effect.
Reeder4 Shows the category, I didn't want to report it as an error but as information that the API seems to pass it correctly.

Alkarex:
Right, we have removed the possibility of renaming the default category (id=1), but we should have cleared the name of that category, which causes your issue.
I will try to make a fix, but for now you can manually edit the name of category 1 to "Uncategorized" (not important what precise name it is, as long as it does not collide with something else)

Tealk:
Great thanks for the fast help

Alkarex:
Addressed by https://github.com/FreshRSS/FreshRSS/pull/2638

PR #2638, opened 2019-11-06
Fix name of default category

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2637

### DIFF
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -45,6 +45,12 @@ class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable 
 				} else {
 					$this->pdo->exec('DROP INDEX IF EXISTS feed_keep_history_index');	//SQLite at least drop index
 				}
+
+				$stm = $this->pdo->prepare('UPDATE `_category` SET name = :name WHERE id = :id');
+				$stm->bindValue(':id', self::DEFAULTCATEGORYID, PDO::PARAM_INT);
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
:+1:

app/Models/CategoryDAO.php
```diff
@@ -45,6 +45,12 @@ protected function addColumn($name) {
 				} else {
 					$this->pdo->exec('DROP INDEX IF EXISTS feed_keep_history_index');	//SQLite at least drop index
 				}
+
+				$stm = $this->pdo->prepare('UPDATE `_category` SET name = :name WHERE id = :id');
+				$stm->bindValue(':id', self::DEFAULTCATEGORYID, PDO::PARAM_INT);
+				$stm->bindValue(':name', 'Uncategorized');
+				$stm->execute();
```
marienfressinaud, line 52: In `addColumn`? :thinking:

Alkarex: Yes, this could indeed be refactored or renamed, but it is only a trigger based on a lacking column, with a set of related updates

marienfressinaud: well, looking at the rest of the code, it's not that surprising (`DROP` :p) but still a bit odd :upside_down_face:

marienfressinaud (approved):

app/Models/CategoryDAO.php
Alkarex: Indeed...

Alkarex:
Update tested from 1.14.3 (tested with SQLite)

Issue #2509, opened 2019-08-27
blank page on sharing page


Tealk:
Hello,

if i go to the share settings page(/p/i/?c=configure&a=sharing) in the settings, i only get a blank page.
I don't see an error message (console, log, ...).

Frenzie:
Do you have any sharing already set up?

Tealk:
No, I have currently set up a Mastodon instance and just wanted to see if I can add it.

marienfressinaud:
It's probably due to an adblocker. I had the same issue because the url contains "sharing" :/

Tealk:
I didn't even think about it. Thank you!
uBlock blocked it

aledeg:
I think we should modify something in the code base so it does not trigger adblocker. Or at least, document it.

Frenzie:
Only some of those are "sharing", others are more saving/adding (e.g., Wallabag, Pocket, Pinboard). Perhaps something more generic like `integrations`?

Tealk:
`integration` sounds good

Frenzie:
And in any case what the URL says and what the text says doesn't need to be the same. `?c=configure&a=integration` will do just fine either way.

Frenzie:
To put all info together, copying https://github.com/FreshRSS/FreshRSS/issues/2419#issuecomment-504958716 here.

> One addition: You can leave easylist:fanboy-addon/fanboy_social_general_block.txt active, but deactivate "sharing" for your domain where you installed FreshRSS. So your own installation is excluded from the block. To do so, the user must go to _uBlock Origin_ > _Options_ > _My Filters_ > Add a new line with:
`*sharing,domain=~yourdomain.com`

aledeg:
So what do we do to address that? Do we document or do we change the url?

Alkarex:
I think we should:
1) Try to catch such an error so that we do not show a white page
2) change the URL to something less problematic

Frenzie:
But it's the page itself that's blocked? I suppose you could see if it loads through fetch or something…

aledeg:
@Tealk Please have a look at #2643 to check if I covered the original problem. Thank you.

Tealk:
I downloaded the latest dev branch but it's still called share?

![image](https://user-images.githubusercontent.com/12276250/68477827-4f429d80-0226-11ea-956a-d1fa88e47044.png)

Frenzie:
[...]

Issue #2419, opened 2019-06-21
Sharing options empty in dev


Offerel:
If i open the options > sharing, the page is empty in dev branch

Alkarex:
Hello,
Could you please attach a screenshot, and also check the logs for any error?

Offerel:
It seems, that this is client specific. In the meantime im back at home from work, and there it works. I see the page as it should. Nothing was in the logs at server side. On Monday, when I'm back at work, i check there the log in Firefox inspector, if i found something.

Offerel:
The problem was uBlock Origin. There was a list active easylist:fanboy-addon/fanboy_social_general_block.txt. If i deactivate that list, the sharing page works.

aledeg:
I've reopened it to add that in the documentation. It might help others.
Thank you for sharing.

Offerel:
One addition: You can leave easylist:fanboy-addon/fanboy_social_general_block.txt active, but deactivate "sharing" for your domain where you installed FreshRSS. So your own installation is excluded from the block. To do so, the user must go to _uBlock Origin_ > _Options_ > _My Filters_ > Add a new line with:
`*sharing,domain=~yourdomain.com`

Frenzie:
Duplicate with #2509, so closing now.

aledeg:
@Offerel Please have a look at #2643 to check if I covered the original problem. Thank you.

Offerel:
I will check this in the dev branch later today. Generally it should work, if there is no http request with "share" or "sharing" from my understanding.

PR #2590, opened 2019-10-24
Translate docs with po4a

Frenzie:
This is my counter-proposal to #2584.

The way this proof of concept currently works is that you execute the `translation-update.sh` script from the `docs` folder. It then takes care of updating everything.

As currently written, the POT/PO files are put in an i18n subfolder.

Screenshot_2019-10-24_15-35-21

Screenshot_2019-10-24_15-36-25

I will have to manually add in all the translations because the documents are insufficiently in sync to auto-generate translations. That makes this PR deceptively large for the moment. It also provides a great opportunity to go through the English translation as I've wanted to do for years but never got around to. (The way a lot of this reads strongly suggests an original French → English translation.)

Minor changes to translation keys will normally automatically show up as fuzzy matches with the differences highlighted. Then you only have to approve the match.

Screenshot_2019-10-24_15-46-08

This script is based on the one provided by @urbalazs for the tldr project. I could rewrite it in PHP from scratch if desired.

---
This change is [Reviewable](https://reviewable.io/reviews/freshrss/freshrss/2590)

PR #2643, opened 2019-11-08
Rename sharing action

aledeg:
Before, some ad-blocker rules were preventing to display the sharing page
because of its name. I've renamed that page to integration but kept every
thing else identical. So it's still called sharing through out the
application.
If needed, this can be addressed as well.

See #2509 and #2419


Alkarex (approved):

docs/en/users/07_Frequently_Asked_Questions.md
```diff
@@ -52,3 +52,14 @@ Some Linux distribution like Fedora or RedHat Enterprise Linux have SELinux syst
 semanage fcontext -a -t httpd_sys_rw_content_t '/usr/share/FreshRSS/data(/.*)?'
 restorecon -Rv /usr/share/FreshRSS/data
 ```
+
+## Why do I have a blank page while trying to configure the sharing options?
+
+The `sharing` word in the URL is a trigger word for some an-blocker rules. Starting at version 1.16, `sharing` has been replaced by `integration` in the faulty URL while keeping the exact same wording through out the application.
```
Frenzie, line 58: Typo. Should I take it up in #2590 or would you prefer to update this separately?

Alkarex: Fine in your existing PR for something that tiny :-)

Issue #2644, opened 2019-11-08
cli failing with Current user must not be empty


X-dark:
Hi,
I am having some difficulties to use the cli with 1.15.1.

Commands involving specific user return errors similar to 
```
PHP Fatal error:  Uncaught Minz_PDOConnectionException: Access to database is denied for `` (`Current user must not be empty!`) in /usr/share/webapps/freshrss/lib/Minz/ModelPdo.php:32
```

For the user-info cli, this fix is enough:
```patch
--- cli/user-info.php.orig      2019-11-08 12:46:30.386059564 +0100
+++ cli/user-info.php   2019-11-08 12:36:37.834287942 +0100
@@ -43,7 +43,7 @@
 foreach ($users as $username) {
        $username = cliInitUser($username);
 
-       $catDAO = FreshRSS_Factory::createCategoryDao();
+       $catDAO = FreshRSS_Factory::createCategoryDao($username);
        $feedDAO = FreshRSS_Factory::createFeedDao($username);
        $entryDAO = FreshRSS_Factory::createEntryDao($username);
        $tagDAO = FreshRSS_Factory::createTagDao($username);
```

But for export-sqlite-for-user it seems less obvious. Here the full error in this case:
```
FreshRSS exporting database to SQLite for user “girardc”…
PHP Fatal error:  Uncaught Minz_PDOConnectionException: Access to database is denied for `` (`Current user must not be empty!`) in /usr/share/webapps/freshrss/lib/Minz/ModelPdo.php:32
Stack trace:
#0 /usr/share/webapps/freshrss/app/Models/Factory.php(6): Minz_ModelPdo->__construct(false)
#1 /usr/share/webapps/freshrss/app/Models/DatabaseDAO.php(201): FreshRSS_Factory::createUserDao()
#2 /usr/share/webapps/freshrss/cli/export-sqlite-for-user.php(26): FreshRSS_DatabaseDAO->dbCopy('/tmp/freshrss-g...', 1)
#3 {main}
  thrown in /usr/share/webapps/freshrss/lib/Minz/ModelPdo.php on line 32
```

Is this related to a specific config error on my part or a recent change to Models classes that break behavior?

X-dark:
Found the culprit (cf my comment on #2635 )

Alkarex:
@X-dark Could you please try this candidate fix? https://github.com/FreshRSS/FreshRSS/pull/2646

PR #2646, opened 2019-11-08
Fix CLI bug regarding an empty user

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2644

### DIFF
--- a/cli/_cli.php
+++ b/cli/_cli.php
@@ -37,7 +37,7 @@ function cliInitUser($username) {
 	if (FreshRSS_Context::$user_conf == null) {
 		fail('FreshRSS error: invalid configuration for user: ' . $username . "\n");
 	}
-	new Minz_ModelPdo($username);
+	Minz_Session::_param('currentUser', $username);
 
 	return $username;
 }
--- a/lib/Minz/ModelPdo.php
+++ b/lib/Minz/ModelPdo.php
@@ -28,12 +28,12 @@ class Minz_ModelPdo {
 		if ($currentUser === null) {
 			$currentUser = Minz_Session::param('currentUser');
 		}
-		if ($currentUser == '') {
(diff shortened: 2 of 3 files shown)


cli/_cli.php
```diff
@@ -37,7 +37,7 @@ function cliInitUser($username) {
 	if (FreshRSS_Context::$user_conf == null) {
 		fail('FreshRSS error: invalid configuration for user: ' . $username . "\n");
 	}
-	new Minz_ModelPdo($username);
+	Minz_Session::_param('currentUser', $username);
```
Alkarex, line 40: We now check later to ensure that the user is set when working with a user database

lib/Minz/ModelPdo.php
```diff
@@ -28,13 +28,13 @@ public function __construct($currentUser = null, $currentPdo = null) {
...
 			$currentUser = Minz_Session::param('currentUser');
 		}
-		if ($currentUser == '') {
-			throw new Minz_PDOConnectionException('Current user must not be empty!', '', Minz_Exception::ERROR);
-		}
 		if ($currentPdo != null) {
 			$this->pdo = $currentPdo;
 			return;
 		}
+		if ($currentUser == '') {
```
Alkarex, line 35: That check is only relevant when we do not already have a user PDO, so change order of checks

cli/_cli.php
```diff
@@ -37,7 +37,7 @@ function cliInitUser($username) {
 	if (FreshRSS_Context::$user_conf == null) {
 		fail('FreshRSS error: invalid configuration for user: ' . $username . "\n");
 	}
-	new Minz_ModelPdo($username);
```
Alkarex, line 40: This was an old, even less good-looking, work-around

X-dark (approved):
It does fix #2644 for me. Thanks

PR #2647, opened 2019-11-08
Fix DB optimize for MySQL

Alkarex:
`pdo->exec()` is not appropriate for MySQL `OPTIMIZE` because `OPTIMIZE` returns some data and not only a code and then fails.
Adds some error log for SQLite and PostgreSQL at the same time.

### DIFF
--- a/app/Models/DatabaseDAO.php
+++ b/app/Models/DatabaseDAO.php
@@ -156,7 +156,12 @@ class FreshRSS_DatabaseDAO extends Minz_ModelPdo {
 
 		foreach ($tables as $table) {
 			$sql = 'OPTIMIZE TABLE `_' . $table . '`';	//MySQL
-			$ok &= ($this->pdo->exec($sql) !== false);
+			$stm = $this->pdo->query($sql);
+			if ($stm == false || $stm->fetchAll(PDO::FETCH_ASSOC) === false) {
+				$ok = false;
+				$info = $stm == null ? $this->pdo->errorInfo() : $stm->errorInfo();
+				Minz_Log::warning(__METHOD__ . ' error: ' . $sql . ' : ' . json_encode($info));
+			}
 		}
 		return $ok;
 	}
--- a/app/Models/DatabaseDAOPGSQL.php
+++ b/app/Models/DatabaseDAOPGSQL.php
@@ -79,7 +79,11 @@ class FreshRSS_DatabaseDAOPGSQL extends FreshRSS_DatabaseDAOSQLite {
 
 		foreach ($tables as $table) {
 			$sql = 'VACUUM `_' . $table . '`';
-			$ok &= ($this->pdo->exec($sql) !== false);
+			if ($this->pdo->exec($sql) === false) {
+				$ok = false;
+				$info = $this->pdo->errorInfo();
+				Minz_Log::warning(__METHOD__ . ' error: ' . $sql . ' : ' . json_encode($info));
+			}
 		}
 		return $ok;
 	}
--- a/app/Models/DatabaseDAOSQLite.php
+++ b/app/Models/DatabaseDAOSQLite.php
@@ -66,6 +66,11 @@ class FreshRSS_DatabaseDAOSQLite extends FreshRSS_DatabaseDAO {
 	}
 
 	public function optimize() {
-		return $this->pdo->exec('VACUUM') !== false;
+		$ok = $this->pdo->exec('VACUUM') !== false;
+		if (!$ok) {
+			$info = $this->pdo->errorInfo();
+			Minz_Log::warning(__METHOD__ . ' error: ' . $sql . ' : ' . json_encode($info));
+		}
+		return $ok;
 	}
 }


app/Models/DatabaseDAO.php
```diff
@@ -156,7 +156,12 @@ public function optimize() {
 
 		foreach ($tables as $table) {
 			$sql = 'OPTIMIZE TABLE `_' . $table . '`';	//MySQL
-			$ok &= ($this->pdo->exec($sql) !== false);
+			$stm = $this->pdo->query($sql);
+			if ($stm == false || $stm->fetchAll(PDO::FETCH_ASSOC) === false) {
+				$ok = false;
+				$info = $stm == null ? $this->pdo->errorInfo() : $stm->errorInfo();
+				Minz_Log::warning(__METHOD__ . ' error: ' . $sql . ' : ' . json_encode($info));
```
aledeg, line 163: Can't you use string templates instead of string concatenation? String concatenation is really memory intensive.

Something like
```php
Minz_Log::warning(sprintf('%s error: %s : %s', __METHOD__, $sql, json_encode($info)));
```

Alkarex: Last time I looked at it, it was the opposite :-) E.g. https://stackoverflow.com/a/16366498 (might have to be tested again with newer PHP versions, but I really expects that to be a compiler's job)
And besides, I find `sprintf` harder to read and more error prone in those cases, but I do use it in several other situations when it makes sense to have one single string with many variables

Alkarex: From my memory, but would need to find additional references, concatenation of single quotes is most of the time the fastest and most memory efficient (especially for small strings), then approaches based on double quotes, and then string formats (which makes sense, as it is quite a complex function call)

aledeg: The comment you linked is from 2013. Since then, PHP7 arrived and changed that. I've been to a conference where one of the PHP core dev explained how the engine changed between PHP5 and PHP7. One of the thing that he mentioned was that string concatenation is really gready because you need to create a lot of slots in memory. For 2 strings, you have 3 slots. For 3 strings, you 5 slots. For 4 strings, you have 7 slots and so on.

Alkarex: Yes, it would be nice to find a newer benchmark, both for speed and memory

Alkarex: Otherwise I will make one tomorrow :-) I am curious of the results, but my money is still on concatenation of single quotes, at least for CPU, and probably also for memory
Edit: for a case like this line, i.e. a few tiny strings

aledeg: FYI https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html

aledeg: FYI http://blog.jpauli.tech/2015-09-18-php-string-management-html/

Alkarex: From what I can quickly find, the question is more whether to use single vs. double quotes, while function calls to e.g. `sprintf` come last https://jonczyk.me/2016/10/23/phpyths-buster-great-string-performance-test/
That seems to require a little dedicated test :-)

aledeg: Ok, so maybe not `sprintf` ^^.

Alkarex: Very quick tests:

```php
<?php
const N = 10000000;

function test1()
{
	$sql = ' OPTIMIZE TABLE `_alice_category`';

	$info = [ 'HY000', 1, 'near "bogus": syntax error' ];
	$infos = json_encode($info);

	$startmem = memory_get_usage();
	$a_start = microtime(true);

	for ($i = N; $i >= 0; $i--) {
		$s = __METHOD__ . ' error: ' . $i . $sql . ' : ' . $infos;
	}

	$a_end = microtime(true);
	$a_mem = memory_get_usage();

	$timeused = $a_end - $a_start;
	$memused = $a_mem - $startmem;

	echo "TEST 1: concat single()\n";
	echo "TIME: {$timeused}\nMEMORY: $memused\n\n\n";
}

function test2()
{
	$sql = ' OPTIMIZE TABLE `_alice_category`';

	$info = [ 'HY000', 1, 'near "bogus": syntax error' ];
	$infos = json_encode($info);

	$startmem = memory_get_usage();
	$a_start = microtime(true);

	for ($i = N; $i >= 0; $i--) {
		$s = sprintf('%s error: %s%s : %s', __METHOD__, $i, $sql, $infos);
	}

	$a_end = microtime(true);
	$a_mem = memory_get_usage();

	$timeused = $a_end - $a_start;
	$memused = $a_mem - $startmem;

	echo "TEST 1: sprintf()\n";
	echo "TIME: {$timeused}\nMEMORY: $memused\n\n\n";
}

test2();
test1();
test2();
test1();
```

On Raspberry Pi 4 (ARMv8):

```
TEST 1: sprintf()
TIME: 5.3001220226288
MEMORY: 320

TEST 1: concat single()
TIME: 3.7047109603882
MEMORY: 112

TEST 1: sprintf()
TIME: 5.2856578826904
MEMORY: 320

TEST 1: concat single()
TIME: 3.7021839618683
MEMORY: 112
```

(Similar results on AMD64)

In order to get `sprintf` to win on speed, one needs to add several more concatenations, while keeping the parameters very short (otherwise concat wins again)

```php
$s = '> ' . $i . ' = ' . __METHOD__ . ' error: ' . $i . ' > ' . $sql . ' : ' . $infos;
//vs.
$s = sprintf('> %s = %s error: %s > %s : %s', $i, __METHOD__, $i, $sql, $infos);
```

```
TEST 1: sprintf()
TIME: 7.460667848587
MEMORY: 320


TEST 1: concat single()
TIME: 8.2111639976501
MEMORY: 128


TEST 1: sprintf()
TIME: 7.4196529388428
MEMORY: 320


TEST 1: concat single()
TIME: 8.2311270236969
MEMORY: 128
```

But `sprintf` still looses on memory in all cases I tried, even with variables containing 1000 times more text (a case I would have bet `sprintf` would win):

```php
<?php
const N = 10000000;
const R = 1000;

function test1()
{
	$sql = ' OPTIMIZE TABLE `_alice_category`';
	$sql = str_repeat($sql, R);

	$info = [ 'HY000', 1, 'near "bogus": syntax error' ];
	$infos = json_encode($info);
	$infos = str_repeat($infos, R);

	$startmem = memory_get_usage();
	$a_start = microtime(true);

	for ($i = N; $i >= 0; $i--) {
		$s = '> ' . $i . ' = ' . __METHOD__ . ' error: ' . $i . ' > ' . $sql . ' : ' . $infos;
	}

	$a_end = microtime(true);
	$a_mem = memory_get_usage();

	$timeused = $a_end - $a_start;
	$memused = $a_mem - $startmem;

	echo "TEST 1: concat single()\n";
	echo "TIME: {$timeused}\nMEMORY: $memused\n\n\n";
}

function test2()
{
	$sql = ' OPTIMIZE TABLE `_alice_category`';
	$sql = str_repeat($sql, R);

	$info = [ 'HY000', 1, 'near "bogus": syntax error' ];
	$infos = json_encode($info);
	$infos = str_repeat($infos, R);

	$startmem = memory_get_usage();
	$a_start = microtime(true);

	for ($i = N; $i >= 0; $i--) {
		$s = sprintf('> %s = %s error: %s > %s : %s', $i, __METHOD__, $i, $sql, $infos);
	}

	$a_end = microtime(true);
	$a_mem = memory_get_usage();

	$timeused = $a_end - $a_start;
	$memused = $a_mem - $startmem;

	echo "TEST 1: sprintf()\n";
	echo "TIME: {$timeused}\nMEMORY: $memused\n\n\n";
}

test2();
test1();
test2();
test1();
```


```
TEST 1: sprintf()
TIME: 116.68087601662
MEMORY: 126976


TEST 1: concat single()
TIME: 120.04346203804
MEMORY: 77824


TEST 1: sprintf()
TIME: 116.99526000023
MEMORY: 126976


TEST 1: concat single()
TIME: 120.40587306023
MEMORY: 77824
```

I have not tested double-quotes, because I am not sure there is a syntax allowing constants and function calls without using an intermediate variable, but suggestions welcome.

To summarize, I think there is a good value of `sprintf` for longer string models, in particular when they need to be reused / translated, but for shorter cases with short models, few parameters, concat is faster & less memory intensive, and in those cases I personally find concat easier to read (linear instead of jumping back and forth) and less error prone (e.g. to mismatch between the placeholders and the parameters)

Alkarex: Since this line is not specific to this PR, I suggest we keep it for now, and further discuss how to potentially change existing and/or future similar constructs, if you like

Issue #2645, opened 2019-11-08
Encoding problem with cli export/import


X-dark:
Hi,

After exporting my user database to sqlite from MySQL with the cli, I am trying to import it back into PostgreSQL, still with the cli.

It fails with the following error in PostgreSQL log:

```
Nov 08 14:38:18 htpc postgres[815]: 2019-11-08 14:38:18.987 CET [314058] ERROR:  invalid byte sequence for encoding "UTF8": 0xe9 0x20 0x69
```

Is it related to utf8mb4 in MySQL? Is there any way to clean my data to allow import or just skip the invalid value?

Alkarex:
Hum, I did not observe such a problem during my tests. Would you be able to find the more precise data, which makes it fail? We could add some Unicode sanitizing before insertion

Alkarex:
Would you be able to produce a small version of your SQLite file, exhibiting the issue?

X-dark:
Invalid entries where limited to a feed that is no longer updated: http://www.hardware.fr/backend/news.xml. I have been able to import my database, removing all entries from that feed.

Here is the stripped down db (just one article of this feed): https://owncloud.cedricgirard.com/s/dtmHp66jqW5bfSS/download

Alkarex:
It is this comment, apparently in ISO-8859-x in the middle of some UTF-8, which seems to cause the problem, only with PostgreSQL:
```xml
<!-- Page actualit� individuelle -->
```

Alkarex:
@X-dark Could you please check this candidate patch? https://github.com/FreshRSS/FreshRSS/pull/2649

X-dark:
Sorry for not having been able to test this sooner. It does fix this bug. Thank you.

PR #2649, opened 2019-11-09
Sanitize UTF-8 before commiting entries

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2645

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -81,6 +81,14 @@ function safe_ascii($text) {
 	return filter_var($text, FILTER_DEFAULT, FILTER_FLAG_STRIP_LOW | FILTER_FLAG_STRIP_HIGH);
 }
 
+if (function_exists('mb_convert_encoding')) {
+	function safe_utf8($text) { return mb_convert_encoding($text, 'UTF-8', 'UTF-8'); }
+} elseif (function_exists('iconv')) {
+	function safe_utf8($text) { return iconv('UTF-8', 'UTF-8//IGNORE', $text); }
+} else {
+	function safe_utf8($text) { return $text; }
+}
+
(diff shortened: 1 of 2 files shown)


lib/lib_rss.php
```diff
@@ -81,6 +81,14 @@ function safe_ascii($text) {
 	return filter_var($text, FILTER_DEFAULT, FILTER_FLAG_STRIP_LOW | FILTER_FLAG_STRIP_HIGH);
 }
 
+if (function_exists('mb_convert_encoding')) {
+	function safe_utf8($text) { return mb_convert_encoding($text, 'UTF-8', 'UTF-8'); }
+} elseif (function_exists('iconv')) {
+	function safe_utf8($text) { return iconv('UTF-8', 'UTF-8//IGNORE', $text); }
+} else {
+	function safe_utf8($text) { return $text; }
```
Frenzie, line 89: Should it also write a little warning in the logs about the fact that it's not actually doing anything at all?

Alkarex: Maybe in the *check installation* page. Otherwise this would give one line in the logs per call

Frenzie: True, that'd become excessive quickly.

Alkarex: We already have a check there (in the *check install* page https://github.com/FreshRSS/FreshRSS/blob/91cb165829badde07c42a002215cf52779d891b6/lib/lib_install.php#L44

Frenzie: They don't say what kind of problem you might get from not having it though, just a fairly vague "for Unicode". Assuming it shouldn't simply be required if you get such curious hard to track down issues from it?

Alkarex: Indeed, we could improve the message and/or documentation

Issue #673, opened 2014-10-18
RTL support


mat-mo:
Hello, thanks for this great application. Can you please add Right-to-left support? It used used for languages written from right to left such as Arabic, Hebrew, Parsi, and more.

rezad1393:
any progress for this or do I need to open another bug report?

Alkarex:
@rezad1393 I personally do not have experience with RTL, so help welcome. 
In particular, I am not sure how it is supposed to look like, how much the `dir` attribute can do, and what requires changes in the logic or themes

Frenzie:
There's no point to opening another bug report, but adding some information could potentially be helpful. I don't know if @marienfressinaud still has his notes & branch around but I wouldn't mind seeing a link to an RTL newsfeed, for example.

I'm thinking the solution described above is too limited; isn't the language set using an attribute like this?

```xml
`xml:lang="ar"`
```

This language should simply be saved and transferred to the feed item. Then the browser will take care of displaying it correctly without us having to maintain any logic.

mat-mo:
https://www.smashingmagazine.com/2017/11/right-to-left-mobile-design/ might be useful.

Frenzie:
Sure, luckily that's 90 % a matter of sticking `dir=rtl` on the HTML element. Just some slightly off details as you can see.

<img src=https://user-images.githubusercontent.com/202757/68656865-e5323d00-0532-11ea-8d5d-08ac9adb4a3d.png width=30%>

Would you happen to know any examples of feeds? Because as [written above](https://github.com/FreshRSS/FreshRSS/issues/673#issuecomment-59640198):

> the important one, is to display the content correctly, so word don't show on the wrong side of the text. This is mostly a problem with mixed text in RTL and LTR text in the same sentence.

mat-mo:
You can try with:
https://www.geektime.co.il/feed

Frenzie:
And it's left bad, right good?

![Screenshot_2019-11-12_10-11-02](https://user-images.githubusercontent.com/202757/68657882-c339ba00-0534-11ea-8c9a-22a64fabf479.png)

mat-mo:
> And it's left bad, right good?
> 
> ![Screenshot_2019-11-12_10-11-02](https://user-images.githubusercontent.com/202757/68657882-c339ba00-0534-11ea-8c9a-22a64fabf479.png)

Yes, exactly.

rezad1393:
[...]

PR #2656, opened 2019-11-12
Automatic mostly-right RTL with dir="auto"

Frenzie:
See <https://github.com/FreshRSS/FreshRSS/issues/673#issuecomment-552802523>.


Frenzie:
This screenshot illustrates before/after.

Screenshot_2019-11-12_10-22-43

Frenzie:
And for good measure, here's an illustration of before/after when the app itself is in RTL (cf. #2658).

Screenshot_2019-11-12_12-54-28
Screenshot_2019-11-12_12-54-54

PR #2658, opened 2019-11-12
[UI] Add RTL support

### LINKED ISSUE #673: RTL support

Frenzie:
Fixes #673.

<hr>

Now obviously this only fixes one thing when there are *at least* a dozen similar issues, but this principle is my proposed solution to all of them.

Edit: to clarify, basically anything that says something like `position: absolute` or `fixed` is likely affected. So are `border-left`/`right`. The good news is that anything that isn't absolutely positioned should largely Just Work™.

---
This change is [Reviewable](https://reviewable.io/reviews/freshrss/freshrss/2658)

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -681,7 +681,7 @@ case 5:
 ?>
 <!DOCTYPE html>
 <html<?php
-if(_t('gen.dir') === 'rtl') {
+if (_t('gen.dir') === 'rtl') {
 	echo ' dir="rtl" class="rtl"';
 }
 ?>>
--- a/app/install.php
+++ b/app/install.php
@@ -680,7 +680,11 @@ case 5:
 }
 ?>
 <!DOCTYPE html>
-<html>
+<html<?php
+if(_t('gen.dir') === 'rtl') {
+	echo ' dir="rtl" class="rtl"';
+}
+?>>
 	<head>
 		<meta charset="UTF-8" />
 		<meta name="viewport" content="initial-scale=1.0" />
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -1,7 +1,7 @@
 <?php FreshRSS::preLayout(); ?>
 <!DOCTYPE html>
 <html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>"<?php
-if(_t('gen.dir') === 'rtl') {
+if (_t('gen.dir') === 'rtl') {
 	echo ' dir="rtl" class="rtl"';
 }
 ?>>
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -1,6 +1,10 @@
 <?php FreshRSS::preLayout(); ?>
 <!DOCTYPE html>
-<html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>">
+<html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>"<?php
+if(_t('gen.dir') === 'rtl') {
+	echo ' dir="rtl" class="rtl"';
+}
+?>>
 	<head>
 		<meta charset="UTF-8" />
 		<meta name="viewport" content="initial-scale=1.0" />
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -355,6 +355,11 @@ a.btn {
 	transform: rotate(45deg);
 }
 
+html.rtl .dropdown-menu::after {
+	left: 13px;
+	right: auto;
+}
+
 .dropdown-header {
 	padding: 0 5px 5px;
 	color: #888;
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -141,8 +141,8 @@ html.rtl .stick input:first-child {
 }
 
 html.rtl .stick .btn-important:first-child {
-	border-left: 1px solid #06f;
 	border-right: none;
+	border-left: 1px solid #06f;
 }
 
 .stick .btn:last-child,
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -140,6 +140,11 @@ html.rtl .stick input:first-child {
 	border-right: 1px solid #06f;
 }
 
+html.rtl .stick .btn-important:first-child {
+	border-left: 1px solid #06f;
+	border-right: none;
+}
+
 .stick .btn:last-child,
 .stick input:last-child {
 	border-radius: 0 3px 3px 0;
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -710,6 +710,11 @@ html.rtl .aside {
 	left: 2px;
 }
 
+html.rtl .aside_feed .tree-folder-items .dropdown-menu::after {
+	left: auto;
+	right: 2px;
+}
+
 .aside_feed .tree-folder-items .item .dropdown-target:target ~ .dropdown-toggle > .icon,
 .aside_feed .tree-folder-items .item.active .dropdown-toggle > .icon {
 	background-color: #fff;
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -680,6 +680,13 @@ html.rtl .aside {
 	line-height: 1.5rem;
 }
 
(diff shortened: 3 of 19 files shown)


Frenzie:
@mat-mo Does the previous/next article navigation in the bottom right look as expected? (I imagine it doesn't. ;-) )

Screenshot_2019-11-12 Your RSS feeds · FreshRSS

Frenzie:
Besides the act of adding the relevant attributes to the HTML, this PR should pretty much be in a decent enough shape to be merged for testing in the dev branch. It resolves most (hopefully all of course ;-) ) issues.

## LTR
Screenshot_2019-11-12 (1) Your RSS feeds · FreshRSS

## RTL
Screenshot_2019-11-12 (1) Your RSS feeds · FreshRSS
Screenshot_2019-11-12 (1) ההזנות שלך · FreshRSS hebrew

p/themes/Origine/origine.css
```diff
@@ -140,27 +145,52 @@ form th {
...
-.stick input + .dropdown > .btn,
-.stick .dropdown + .btn,
-.stick .dropdown + input,
-.stick .dropdown + .dropdown > .btn {
+html.rtl .stick .btn:last-child,
+html.rtl .stick input:last-child {
+	border-radius: 3px 0 0 3px;
+}
+
+html:not(.rtl) .stick .btn + .btn,
```
Alkarex, line 158: Can't we avoid the `:not(.rtl)` thanks to the `.rtl` below, overriding the properties?

Alkarex (approved):

p/themes/Origine/origine.css
Frenzie: Not quite elegantly. Options include `border-left-width: 0;` instead of none, so you can easily restore it to some `px` value (probably the best alternative) and possibly also some invalid value.

Alkarex: Ok

Alkarex:
So we lack adding `class="rtl"`, right? Should this be automatic based on the language selected?

Frenzie:
Indeed. For testing I used this simple change, that should also be put in install.php.
```diff
diff --git a/app/layout/layout.phtml b/app/layout/layout.phtml
index 498cc447..d08261d7 100644
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -1,6 +1,6 @@
 <?php FreshRSS::preLayout(); ?>
 <!DOCTYPE html>
-<html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>">
+<html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>" dir="rtl" class="rtl">
        <head>
                <meta charset="UTF-8" />
                <meta name="viewport" content="initial-scale=1.0" />
```

I was thinking it could perhaps be put in the `app/i18n/<lang>` folder as some kind of metadata, or alternatively it could simply be hardcoded in an array somewhere.

p/themes/Origine/origine.css
```diff
@@ -131,6 +131,11 @@ form th {
 	border-radius: 3px 0 0 3px;
 }
 
+html.rtl .stick .btn:first-child,
+html.rtl .stick input:first-child {
+	border-radius: 0 3px 3px 0;
+}
+
 .stick .btn-important:first-child {
 	border-right: 1px solid #06f;
```
Frenzie, line 140: Oh, here's one I missed.

p/themes/Origine/origine.css
Frenzie: The alternative looks like this:

```diff
diff --git a/p/themes/Origine/origine.css b/p/themes/Origine/origine.css
index 86210940..5f90fb1b 100644
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -150,16 +150,16 @@ html.rtl .stick input:last-child {
        border-radius: 3px 0 0 3px;
 }
 
-html:not(.rtl) .stick .btn + .btn,
-html:not(.rtl) .stick .btn + input,
-html:not(.rtl) .stick .btn + .dropdown > .btn,
-html:not(.rtl) .stick input + .btn,
-html:not(.rtl) .stick input + input,
-html:not(.rtl) .stick input + .dropdown > .btn,
-html:not(.rtl) .stick .dropdown + .btn,
-html:not(.rtl) .stick .dropdown + input,
-html:not(.rtl) .stick .dropdown + .dropdown > .btn {
-       border-left: none;
+.stick .btn + .btn,
+.stick .btn + input,
+.stick .btn + .dropdown > .btn,
+.stick input + .btn,
+.stick input + input,
+.stick input + .dropdown > .btn,
+.stick .dropdown + .btn,
+.stick .dropdown + input,
+.stick .dropdown + .dropdown > .btn {
+       border-left-width: 0;
 }
 
 html.rtl .stick .btn + .btn,
@@ -171,7 +171,8 @@ html.rtl .stick input + .dropdown > .btn,
 html.rtl .stick .dropdown + .btn,
 html.rtl .stick .dropdown + input,
 html.rtl .stick .dropdown + .dropdown > .btn {
-       border-right: none;
+       border-left-width: 1px;
+       border-right-width: 0;
 }
 
 .stick input + .btn {
```

Alkarex:
I suggest we add a line 
```php
'dir' => 'rtl',
```
in https://github.com/FreshRSS/FreshRSS/blob/dev/app/i18n/en/gen.php
were `/en/` is to be replaced by RTL languages

Frenzie:
@Alkarex Something like this? https://github.com/FreshRSS/FreshRSS/pull/2658/commits/0c06acbdb8c813dd55951cc5c8d1ff2caf0755de

app/layout/layout.phtml
```diff
@@ -1,6 +1,10 @@
 <?php FreshRSS::preLayout(); ?>
 <!DOCTYPE html>
-<html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>">
+<html lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>"<?php
+if(_t('gen.dir') === 'rtl') {
```
```suggestion
if (_t('gen.dir') === 'rtl') {
```

app/install.php
```diff
@@ -680,7 +680,11 @@ function printStep5() {
 }
 ?>
 <!DOCTYPE html>
-<html>
+<html<?php
+if(_t('gen.dir') === 'rtl') {
```
Frenzie: Copied the typo
```suggestion
if (_t('gen.dir') === 'rtl') {
```

p/themes/Origine/origine.css
Alkarex: It looks a bit lighter, maybe? No strong preference

Alkarex:
Yes @Frenzie , looks good 👍

p/themes/Origine/origine.css
Frenzie: I went with the current version because I figured it's slightly easier to parse (if not ltr then border-left:none, elseif ltr then border-right: none) but I'm fine with any approach. A similar option is to add `html.ltr` as well.

Frenzie:
I did this manually, but I just noticed https://github.com/cssjanus/cssjanus https://github.com/cssjanus/php-cssjanus

Which opens up an easy path to a completely different approach.

Frenzie:
I'm thinking this CSSJanus approach may be better than this PR, mainly because it should require (almost) no effort to maintain, not to mention manually going through all the themes would add up even if it can be slowly spaced out.

Using [the PHP port](https://github.com/cssjanus/php-cssjanus/releases/tag/v1.3.0) it could potentially even be integrated into the program itself.

The basic idea is a bit different though. Instead of localized overrides in the same CSS file, you'd generate a `filename.ltr.css` (or whatever you want to call it), and you'd load that one instead of the normal CSS file. There's an article about the principle illustrated with SASS [here](https://www.sitepoint.com/manage-rtl-css-sass-grunt/).

When minor touchups are required, the "current" approach (`html.rtl etc`) could still be used to perform specific overrides.

There are actually also newer CSS attributes that may mostly avoid the issue, but they're not supported by Edge yet.

https://drafts.csswg.org/css-writing-modes-4/#flow-relative
https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-start

marienfressinaud:
Can we close this PR in favor of #2678?

Frenzie:
I believe the other approach is much better in principle, yes.

Issue #2654, opened 2019-11-11
why does freshrss change my feed urls?


rezad1393:
I have some feeds that are from  https://feeds.feedburner.com
like https://feeds.feedburner.com/TheArtOfManliness
but it seems that freshrss changes the url from https to http.
when I fix it again it gets changed again later.
maybe related to websub?

Alkarex:
Hello,
Looking at the source https://feeds.feedburner.com/TheArtOfManliness?format=xml , there is a "self" attribute, which is referring to the HTTP version, and this "self" attribute is used by FreshRSS to find the canonical address.

```xml
<atom10:link xmlns:atom10="http://www.w3.org/2005/Atom" rel="self" type="application/rss+xml" href="http://feeds.feedburner.com/TheArtOfManliness" />
```

Feedburner is already in our list of forced HTTPS domains https://github.com/FreshRSS/FreshRSS/blob/dev/force-https.default.txt but we should double-check that the queries are actually done over HTTPS.

rezad1393:
I have checked again and it does change the update feed url to http.
at least that is what I see when I open manage option of the feed.

and also the blogspot feeds get changed to the https://www.blogger.com/feeds/MANYNUMBERS/posts/default format.

how do I stop this.
I dont like that a program changes my feeds automatically.
how do I stop freshrss saving me from myself?

Alkarex:
The `self` attribute is a way for the feeds to tell what the correct/best/canonical address is.
Likewise, we are obeying `HTTP/1.x 301 Moved Permanently` responses.
Both are standard behaviours, expected from compliant clients.

(Forced) HTTPS is another issue, and we need to double-check in the query logs, not in the feed configuration page.

You can chose to disable WebSub in your `./data/config.php` on the following model:
Cf. https://github.com/FreshRSS/FreshRSS/blob/09c088c62ece3b17615fb4e2addfa16663bb334f/config.default.php#L72-L74

We could consider disabling the logic around `self` when `WebSub` is disabled though.

rezad1393:
I understand some of what you say.
but I dont understand why the application changes what I configed it to use.
when I use https://somefeed and then it uses the http://sameaddress .
cant I force the freshrss not to touch my feeds addresses?
this is the first time ever (after using good reader and inoreader and a lot of desktop apps) that I encounter this issue.

Frenzie:
It's not FreshRSS doing it, it's the feed doing it. If those others don't then they've got a bug imho.

[...]

PR #2659, opened 2019-11-12
Do not obbey rel=self if WebSub is disabled

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/2654
https://github.com/pubsubhubbub/PubSubHubbub/wiki/Moving-Feeds-or-changing-Hubs

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -415,7 +415,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 				$entryDAO->commit();
 			}
 
-			if ($feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
+			if ($pubsubhubbubEnabledGeneral && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
 				if ($feed->selfUrl() !== $url) {	//https://code.google.com/p/pubsubhubbub/wiki/MovingFeedsOrChangingHubs
 					$selfUrl = checkUrl($feed->selfUrl());
 					if ($selfUrl) {
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -416,7 +416,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 			}
 
 			if ($pubSubHubbubEnabled && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
-				if ($feed->selfUrl() !== $url) {	//https://code.google.com/p/pubsubhubbub/wiki/MovingFeedsOrChangingHubs
+				if ($feed->selfUrl() !== $url) {	// https://github.com/pubsubhubbub/PubSubHubbub/wiki/Moving-Feeds-or-changing-Hubs
 					$selfUrl = checkUrl($feed->selfUrl());
 					if ($selfUrl) {
 						Minz_Log::debug('WebSub unsubscribe ' . $feed->url(false));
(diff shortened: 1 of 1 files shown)


Frenzie (commented):
Not sure what I think of the change but the code is fine of course.

marienfressinaud (commented):
I'm not totally comfortable with this since it basically provides a workaround to a problem which is not about WebSub but about "why FRSS changes automatically feed urls"… We can discuss about the interest of this change, but I don't think it satisfies the initial issue

Frenzie:
I agree, I think that a fix to the (supposed?) issue would be some kind of `Prevent automatic URL update` checkbox or dropdown, similar to a setting like `Always clear cache` or `Verify SSL security`.

Edit: It's probably better to phrase it positively rather than negatively, along the lines of `Apply/Follow/Obey URL updates`.

Alkarex:
This patch only indirectly addresses the cases of some automatic URL redirections. Indeed, the `self` part is an element of the WebSub protocol. As such, when disabling WebSub, it makes sense not to process it.

marienfressinaud:
Indeed, it makes sense, but shouldn't we use `$feed->pubSubHubbubEnabled()` instead?

app/Controllers/feedController.php
```diff
@@ -415,7 +415,7 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 				$entryDAO->commit();
 			}
 
-			if ($feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
+			if ($pubsubhubbubEnabledGeneral && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
```
marienfressinaud:
```suggestion
			if ($feed->pubSubHubbubEnabled() && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
```

I think `$feed->pubSubHubbubEnabled()` returns `false` when `$pubsubhubbubEnabledGeneral` is `false` and I find it's more logical this way

Alkarex: `$feed->pubSubHubbubEnabled()` would not work, but you are right that it was not the correct variable :-)
Fix coming using 
https://github.com/FreshRSS/FreshRSS/blob/1ddf98c50b2e0388c76386202bd8be3442640418/app/Controllers/feedController.php#L279

Alkarex:
Any final comment? :-)

app/Controllers/feedController.php
```diff
@@ -415,7 +415,7 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 				$entryDAO->commit();
 			}
 
-			if ($feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
+			if ($pubSubHubbubEnabled && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
 				if ($feed->selfUrl() !== $url) {	//https://code.google.com/p/pubsubhubbub/wiki/MovingFeedsOrChangingHubs
```
Frenzie: This URL no longer works.

Alkarex:
```suggestion
				if ($feed->selfUrl() !== $url) {	// https://github.com/pubsubhubbub/PubSubHubbub/wiki/Moving-Feeds-or-changing-Hubs
```

Alkarex: Fixed :-)

marienfressinaud (approved):

app/Controllers/feedController.php
Alkarex: Reverted to the first version https://github.com/FreshRSS/FreshRSS/pull/2922

PR #2660, opened 2019-11-12
Allow easy Docker port override

Frenzie:
Because I use 8080 for other things. ;-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ ifndef TAG
 	TAG=dev-alpine
 endif
 
+PORT ?= 8080
+
 ifeq ($(findstring alpine,$(TAG)),alpine)
 	DOCKERFILE=Dockerfile-Alpine
 else ifeq ($(findstring arm,$(TAG)),arm)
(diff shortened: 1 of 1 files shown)


Alkarex (approved):

Makefile
```diff
@@ -4,6 +4,8 @@ ifndef TAG
 	TAG=dev-alpine
 endif
 
+PORT ?= 8080
```
aledeg, line 7: I would use a less generic envvar to avoid collision. Maybe something like ```FRSS_PORT```.

aledeg (dismissed):

Makefile
marienfressinaud: The command is usually used as `PORT=8000 make start` (i.e. no need to export) so I don't mind having a generic variable name. If we change to `FRSS_PORT`, we should change `TAG` as well to be coherent

aledeg: That makes sense

marienfressinaud (approved):

Makefile
Frenzie: Or `make start PORT=23423` (doesn't change the point). :-)

PR #2661, opened 2019-11-13
Adding namespaces

aledeg:
Adding namespaces because they are supported since PHP5.3. It's about time I think. It will simplify the name of classes.

Let me know what you think. Do I continue or do I stop?


aledeg:
It's far from being complete, but I'd like to know before I go too deep into the code if I should continue.

Alkarex:
Last time I checked, this had a performance penalty, and required a not-so-elegant `\` syntax in front of almost everything to mitigate the speed toll.
P.S.: So an update on that point would be nice

aledeg:
There is only a subset of methods which needs a `\` in front of it.
Regarding the performance. It might be a little bit slower but I think it will be easier to handle complexity and autoloading.
If it's really slugish, it might be something hidden so far but we probably need to address it anyway. One of the solution use is huge project is to create a cache to improve performance.

marienfressinaud:
If we choose to add namespaces, I would opt for another strategy: add them gradually in the code over time. A PR with one commit changing 365 files cannot be reviewed and will most likely introduce regressions. The problem with my strategy is it takes time and we'll probably stay in a situation where some of the files use namespaces while some others don't.

So now my point of view: not in FreshRSS 1.x. It's too way too many files to be done properly. I know it can be disappointing to not use these cool features but I don't think it's worth a huge migration of the codebase. However, we can imagine a FreshRSS v2, starting from zero, with a different vision of what the interface should be, using our experience from FRSS v1.x… and where we could use all the new PHP features. But it's another story ;)

aledeg:
Ok, so let's put that aside for a while. So sad :'(

Frenzie:
@aledeg I support you. ;-) (How much depends a bit on that performance impact thing.)

aledeg:
@Frenzie Thanks. The impact can be mitigated with a proper cache. At work, we have thousand of classes and we don't experience performance issues. The performance impact comes from the code itself.

marienfressinaud:
I probably didn't insist too much on this, but it was just my opinion: it doesn't mean it should be the finale decision of the community! I'm saying this because I hope my arguments were good enough for you to change your opinion, I'm all opened to change my mind otherwise :)

Frenzie:
@aledeg Naively speaking I don't understand how there could be any performance impact at all? It seems that the distinction with the `FreshRSS_`/`Minz_` prefix for looking up the right method should be effectively zero and that it just makes the code organization a bit easier in the long run.

@marienfressinaud Your arguments are certainly valid, but one could also turn them around. Since @aledeg is willing to put in the effort, maybe that just means it should have a bigger version bump than usual. ^_^

Alkarex:
Just for the record when we discuss namespaces again, here is a related issue https://github.com/FreshRSS/FreshRSS/pull/1847 (e.g. with performance and `\`)

Issue #2663, opened 2019-11-13
Improve logs when an error on profile page happen


JonathanMM:
On profile page, I tried to define an api password, to use an android application. But, the password was not set, and only a generic error displayed on the page.
After investigation (thanks to Mattermost server!), the issue comes from the folder ./data/fever, which haven't write rights.
It should be great if this information will add on /user/*/log* file, to help to debug :)

aledeg:
There is a log in the ```data\users\_``` folder. It does not show in the interface though.
I've addressed that issue in #2664. If you could test it, it will be appreciated.

PR #2664, opened 2019-11-14
Add log in fever api saving process

aledeg:
Before, there was no user log when the fever api credential saving process
was failing. There was one though for the admin user but it did not appear
in the interface.
Now, there is a user log showing what the problem is. The admin log is still
there but catch only unknown errors.

See #2663

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -59,14 +59,22 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 			$apiPasswordHash = self::hashPassword($apiPasswordPlain);
 			$userConfig->apiPasswordHash = $apiPasswordHash;
 
-			@mkdir(DATA_PATH . '/fever/', 0770, true);
-			self::deleteFeverKey($user);
-			$userConfig->feverKey = strtolower(md5($user . ':' . $apiPasswordPlain));
-			$ok = file_put_contents(DATA_PATH . '/fever/.key-' . sha1(FreshRSS_Context::$system_conf->salt) . '-' . $userConfig->feverKey . '.txt', $user) !== false;
+			$feverPath = DATA_PATH . '/fever/';
 
-			if (!$ok) {
-				Minz_Log::warning('Could not save API credentials for fever API', ADMIN_LOG);
-				return $ok;
+			if (!file_exists($feverPath)) {
+				@mkdir($feverPath, 0770, true);
+			}
+
+			if (!is_writable($feverPath)) {
+				Minz_Log::error("Could not save Fever API credentials. The directory does not have write access.");
+			} else {
+				self::deleteFeverKey($user);
+				$userConfig->feverKey = strtolower(md5("{$user}:{$apiPasswordPlain}"));
+				$ok = file_put_contents($feverPath . '.key-' . sha1(FreshRSS_Context::$system_conf->salt) . '-' . $userConfig->feverKey . '.txt', $user) !== false;
+
+				if (!$ok) {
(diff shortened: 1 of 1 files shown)


app/Controllers/userController.php
```diff
@@ -59,13 +59,23 @@ public static function updateUser($user, $email, $passwordPlain, $apiPasswordPla
...
-			@mkdir(DATA_PATH . '/fever/', 0770, true);
+			$feverPath = DATA_PATH . '/fever/';
+
+			if (!file_exists($feverPath)) {
+				@mkdir($feverPath, 0770, true);
+			}
+
+			if (!is_writable($feverPath)) {
+				Minz_Log::error("Could not save fever API credentials. '{$feverPath}' does not have write access.");
+				return false;
```
Alkarex: Should we make that a hard error and give up saving the rest (as we do now), or a warning, which will just prevent Fever API from working but not the rest?

aledeg: Either way is fine as long as we know what is happening.

Alkarex: I think logging a warning and continue would be better

aledeg: OK, I'll modify the logic later.

aledeg: I've modified the logic but I am not convinced it's a good thing :/
It feels odd.

aledeg: @Alkarex I was using it locally without the error and it's really odd because it says that the user has been updated but then there are some errors you are not aware of since they are buried in the logs. I think it's better to have the hard error. What do you think?

marienfressinaud:
It seems to me it's hard to return the correct feedback to the user because the `updateUser` method does too many things. Maybe its time to start splitting it in smaller methods? What about moving the "API password" field in a separate form (it doesn't really belong to the "profile" anyway), targeting a new `apiController->updatePasswordAction`? We could make it failing with an explicit error message to the user easily then, and add precise logs.

It's a bit more work but I'm pretty sure it would enhance the code and the user experience :)

app/Controllers/userController.php
```diff
@@ -59,14 +59,22 @@ public static function updateUser($user, $email, $passwordPlain, $apiPasswordPla
...
+
+			if (!is_writable($feverPath)) {
+				Minz_Log::error("Could not save fever API credentials. '{$feverPath}' does not have write access.");
+			} else {
+				self::deleteFeverKey($user);
+				$userConfig->feverKey = strtolower(md5("{$user}:{$apiPasswordPlain}"));
+				$ok = file_put_contents($feverPath . '.key-' . sha1(FreshRSS_Context::$system_conf->salt) . '-' . $userConfig->feverKey . '.txt', $user) !== false;
+
+				if (!$ok) {
+					Minz_Log::warning('Could not save fever API credentials. Unknown error.', ADMIN_LOG);
```
```suggestion
					Minz_Log::warning('Could not save fever Fever API credentials. Unknown error.', ADMIN_LOG);
```

Alkarex:
While waiting for @marienfressinaud 's suggestion, I think this PR is fine.
Indeed, a problem with Fever API, which is minor compared to the Web (and I would like to say also minor compared to Google Reader API), should not make the whole update to fail

Alkarex:
Furthermore, like @marienfressinaud did e.g. for the install, we could report warnings in the Web UI

marienfressinaud (approved):
It's good enough for me :+1: I'll work on my suggestion then

aledeg:
Log message has been updated

marienfressinaud (changes requested):
Just realized we should not log the `feverPath` to normal users! It can reveal how FRSS is installed

aledeg:
done

marienfressinaud (approved):

PR #2665, opened 2019-11-14
Fix link to installation instructions on first steps

RyanNielson:


Alkarex:
Thanks 👍

docs/en/users/02_First_steps.md
```diff
@@ -2,7 +2,7 @@ Learning how to handle a new application is not always easy. We build FreshRSS t
 
 This section guides you to the pages you need as a new comer.
```
Frenzie, line 3: Note to self:
```suggestion
This section guides you to the pages you need as a newcomer.
```

Frenzie: Was already fixed in #2590, so I'd prefer to move toward merging that unless there are specific concerns. :-)

PR #2669, opened 2019-11-15
Fix sharing/integration

Alkarex:
Arg, it looks like the submit button in sharing/integration page is
broken in 1.15.2 :-(
Fix https://github.com/FreshRSS/FreshRSS/pull/2643


marienfressinaud (approved):
It seems to be the last one

Frenzie (approved):

aledeg:
good catch!

PR #2335, opened 2019-04-03
Feature/new archiving

aledeg:
See #1167

PR #2672, opened 2019-11-17
Fix adding categories in MySQL 5.5

### LINKED ISSUE #2670: addCategory error
I get an error on the latest version when i try to add a category:
SQL error addCategory: ["21S01",1136,"Column count doesn't match value count at row 1"]

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2670 (from https://github.com/FreshRSS/FreshRSS/pull/2335 )
Tested with MySQL 5.5.40, the oldest version on Docker Hub
Debian 8.0 Jessie ships MySQL 5.5.43

### DIFF
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -81,7 +81,7 @@ class FreshRSS_CategoryDAO extends Minz_ModelPdo implements FreshRSS_Searchable 
 
 	public function addCategory($valuesTmp) {
 		$sql = 'INSERT INTO `_category`(name, attributes) '
-		     . 'SELECT * FROM (SELECT TRIM(?), ?) c2 '	//TRIM() to provide a type hint as text for PostgreSQL
+		     . 'SELECT * FROM (SELECT TRIM(?) AS name, TRIM(?) AS attributes) c2 '	//TRIM() to provide a type hint as text
 		     . 'WHERE NOT EXISTS (SELECT 1 FROM `_tag` WHERE name = TRIM(?))';	//No tag of the same name
 		$stm = $this->pdo->prepare($sql);
 


Alkarex:
Tested with SQLite and PostgreSQL as well

Alkarex:
~~Arg, I have visibly been affected by a bug with the changes in the branches...~~
Edit: fixed

Alkarex:
Are you aware of any distribution supporting MySQL 5.5 besides Debian?
* https://distrowatch.com/table.php?distribution=debian (5.5.43, until 2020-05)
* https://hub.docker.com/_/mysql/?tab=tags&name=5.5.40

If not, I suggest we update our documentation to refer to MySQL 5.5.40+ (I dot not personally have an easy way of testing older versions)

Frenzie:
Isn't CentOS always older?

Alkarex:
Indeed @Frenzie but CentOS 6 ships MySQL 5.1, which we had to drop a long time ago due to key features missing, for instance around Unicode support
https://distrowatch.com/table.php?distribution=centos
I am not sure which MySQL version is in CentOS 7
Edit: apparently 5.5.52 https://pkgs.org/download/mysql55-mysql-server

Frenzie:
I think that's 5.7.

Edit: thought wrong, I guess. :-P

Frenzie:
@Alkarex No, I don't think I'm wrong. If I don't misunderstand your site you can choose between 5.5, 5.6, and 5.7. (Not sure why one would pick older? Were the newer ones added later?)

https://centos.pkgs.org/7/centos-sclo-rh-x86_64/rh-mysql57-mysql-server-5.7.24-1.el7.x86_64.rpm.html

Alkarex:
@Frenzie I am not sure either. Probably added later during the continuous maintenance of CentOS 7, I imagine. I have not found an equivalent of https://packages.ubuntu.com/disco/mysql-server for CentOS

Frenzie:
Hold on, it goes much further? Here's 8. I'm guessing they just keep adding new ones.

https://centos.pkgs.org/7/centos-sclo-rh-x86_64/rh-mysql80-3.2-2.el7.x86_64.rpm.html

Seems like it shouldn't be much of a concern then.

Alkarex:
But if someone installed CentOS 7 when it was released, and applied all (security) updates, what MySQL version do they have today?

Frenzie:
It looks like the answer is actually MariaDB 5.5. See https://linuxize.com/post/install-mysql-on-centos-7/ and https://linuxize.com/post/install-mariadb-on-centos-7/

marienfressinaud (approved):

PR #2673, opened 2019-11-18
Update i18n cli tools

aledeg:
Few things were bugging me when using the cli tool for i18n. So I've modified
the tools to be easier to use.

First, the tool automatically adds missing keys to all languages. This way, we
always have all keys in all languages.
Second, the tool detects all untranslated keys and adds automatically the todo
comment after the value.
Third, when adding a new key, the key is pushed to all languages at once. There
is no need to duplicate it manually. Thus making the duplication process obsolete.
Fourth, translation and ignore keys are manipulated at the same time. Thus we don't have obsolete ignored strings anymore.

On top of that, I've added some rules in the makefile for the most common actions on i18n files and I've changed short options to long options for readability.


aledeg:
More things to come in the future.

Frenzie:
Yeah, that todo thing made it nigh useless. The rest was less bothersome but nice to have. ;-)

aledeg:
Thanks!

aledeg:
The cli tool update is finished. I used it to format once for all i18n files.
That make a big commit but now we can reproduce the format anytime. We can split that PR in 2 if you think that it's unmanageable.
Anyway, tests are more than welcome.

aledeg:
@Frenzie I am not sure I understand what you mean. Could you elaborate?

Frenzie:
Don't care about the PR, but I'd prefer to rebase rather than squash the commits.

Frenzie:
@aledeg It probably can't be (easily) helped, but a bunch of things were changed to ASCII order (1, 10, 11, 2) rather than natural order. ASCII order also does silly things like sorting lowercase letters after all the uppercase ones.

aledeg:
I'll see what I can do.

cli/i18n/I18nFile.php
```diff
@@ -80,7 +79,7 @@ private function flatten($translation, $prefix = '') {
 	private function unflatten($translation) {
 		$a = array();
 
-		ksort($translation);
+		ksort($translation, SORT_NATURAL | SORT_FLAG_CASE);
```
aledeg, line 82: @Frenzie I've sorted using natural order.

Frenzie: Requires PHP 5.4 (which is all perfectly fine now). :-)

app/i18n/cz/conf.php
```diff
@@ -85,17 +84,17 @@
 		'state_6' => 'Zobrazit oblíbené nepřečtené články',
 		'state_7' => 'Zobrazit oblíbené články',
 		'state_8' => 'Zobrazit všechny články vyjma oblíbených',
-		'state_9' => 'Zobrazit všechny přečtené články vyjma  oblíbených',
```
Frenzie, line 89: What's happening here with the spaces vs tabs? Is that intentional or a side effect?

Makefile
```diff
@@ -92,6 +92,6 @@ endif
 ## HELP ##
 ##########
 help:
-	@grep -h -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+	@grep --extended-regexp '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
Frenzie, line 99: Looks like you skipped --no-filename

aledeg: You're right

app/i18n/cz/conf.php
aledeg: That's a side effect. It's weird to have a tab in here as well as 2 spaces.

Frenzie: I'm guessing the two spaces is just a typo, but the replacement seems a bit odd here.

Makefile
aledeg: I've checked and this parameter has no effect in our use case

aledeg: It's useful only when we have multiple files. Here we always have only one.

Frenzie: Do you recall if you added it for a specific reason @marienfressinaud ? (Like some future usage.)

aledeg:
I think we can consider this PR as completed.

Frenzie (approved):

Makefile
aledeg: It looks like a copy-&-paste from an internet resource. I've done the same thing at work :)

Alkarex:
I have not followed carefully, so please tell if it should be safe enough for 1.15.3 (which we need to issue soon due to a couple of additional regressions) or if we should wait for 1.16.0

aledeg:
I think it's pretty safe to include it in 1.15.3 but I think it's useless to rush it since it's a cli tools that only devs should use. I would ship it in 1.16.0.
I think it could be a good idea to include it as soon as possible though (after 1.15.3 is released) so translating for 1.16.0 will be easier.

Frenzie:
^ Yup, that. No reason to wait, no reason to hold off either. Even if it were totally broken (which it's not ;-) it's still completely self-contained.

Makefile
marienfressinaud: ha ha, yes it is a copy-paste from what I used at work ^^ I think it was initially from [a Marmelab article](https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html). I added the `-h` to avoid problems when including `.env` files in another project though (see in the article comments). I think it is safe to drop it here since we don't include any files.

Makefile
```diff
@@ -14,27 +15,83 @@ else
...
+	@exit 10
+endif
+ifndef key
+	@echo To ignore a key, you need to provide one in the "key" variable.
+	@exit 10
+endif
+	@$(PHP) ./cli/manipulate.translation.php -a ignore -k $(key) -l $(lang)
+	@echo Key ignored.
+
+.PHONY: i18n-format i18n-add-language i18n-add-key i18n-remove-key i18n-ignore-key
```
marienfressinaud: I prefer to add `.PHONY` next to the commands because it helps to don't forget them when adding a new target

Makefile
Frenzie: That article led me to this:

https://github.com/dergachev/screengif/blob/a26c9880a09fa4c782abdf61ca496270b055d8e8/Makefile#L20-L21

```
changelog:
		git log `git describe --tags --abbrev=0`..HEAD --oneline |  awk 'BEGIN { print "Commits since last tag:" }; {print "- " $$0}' | vim - -R +"vs CHANGELOG.md" +"set noro"
```

Not really sure I quite see the advantage of all that, but it makes me wonder if a `make log` for `git log --oneline` would be useful. :-)

(Probably not.)

aledeg: @Frenzie There is a better tool to make that and it's called git aliases.
You can add an alias this way:
```
git config --global alias.oneline 'log --oneline --abbrev-commit --graph'
```
And use it this way:
```
git oneline
```
That's what I use everyday :)

Makefile
aledeg: Done

Makefile
Frenzie: Maybe. I'd be more inclined say there's a better tool to do that and it's called something like gitk, QGit or git-dag. :-P

But I just realized they mean "changelog" as a tool to help prepare a manual changelog. :man_facepalming:

Makefile
```diff
@@ -14,27 +15,85 @@ else
 	DOCKERFILE=Dockerfile
 endif
 
-.PHONY: build
+############
+## Docker ##
+############
+.PHONY build
```
marienfressinaud: I think you forgot the `:`

aledeg: Right! It's fixed

Makefile
```diff
@@ -35,6 +39,61 @@ start: ## Start the development environment (use Docker)
...
+i18n-add-language: ## Add a new supported language
+ifndef lang
+	@echo To add a new language, you need to provide one in the "lang" variable.
+	@exit 10
+endif
+	@$(PHP) ./cli/manipulate.translation.php -a add -l $(lang)
+	@echo Language added.
+
+.PHONY: i18n-add-key
+i18n-add-key: ## Add a key
```
marienfressinaud: I wonder if we should be more explicit, something like "Add a I18N key to all languages" (maybe i18n is too much)

Makefile
```diff
@@ -35,6 +39,61 @@ start: ## Start the development environment (use Docker)
...
+endif
+ifndef value
+	@echo To add a key, you need to provide its value in the "value" variable.
+	@exit 10
+endif
+	@$(PHP) ./cli/manipulate.translation.php -a add -k $(key) -v $(value)
+	@echo Key added.
+
+.PHONY: i18n-remove-key
+i18n-remove-key: ## Remove a key
```
marienfressinaud: Same here : "Remove a I18N key from all languages"

marienfressinaud (commented):
I didn't look too much into the implementation details, but it looks pretty good :+1: I just made a small comment about Makefile documentation

Makefile
aledeg: Done

marienfressinaud:
I think we can merge it now that 1.15.3 is published :) You'll just have to reformat german translations

Alkarex (approved):

aledeg:
Done

Alkarex:
I have not read everything, but looks fine for me 👍

Alkarex:
Arg, now we have conflicts (I was too slow to merge 🐌). Would you mind resolving them @aledeg ?

aledeg:
@Alkarex done

PR #2675, opened 2019-11-19
Separate the update API password endpoint

marienfressinaud:
As I mentioned in https://github.com/FreshRSS/FreshRSS/pull/2664, it would be great to simplify the `userController` endpoints, especially to enhance the feedback to the user.

This PR aims to extract the "update API password" endpoint. It should be more easily reviewed commit by commit.

First step is to extract some methods which don't belong to the controller but are more or less "utility" functions. So I added a new `app/Utils` folder for these methods. I needed to do this because some are used in two different controllers then, and I don't want to call controller methods from another controller.

The second step is to actually move the endpoint in a new API controller. There are two consequences:

- I moved the API password input in another form
- I removed the possibility to update the API password from the CLI (it could be done quite easily with a bit more code, but I'm not sure it's really useful)

Then, I updated the locales (nothing fancy but it would be easier with https://github.com/FreshRSS/FreshRSS/pull/2673 :smile:)

The last commit is optional, I just tried to simplify the controller a bit more by hiding the presence check of the fever directory. I'm not totally satisfied with this because it makes the fever `updateKey` method a bit more complex. Let me know what you think about this!

It looks like this:

Capture d’écran de 2019-11-19 10-26-38

Last point I want to mention: I initially wanted to move the form in a new "Mobile & desktop" page, with the "Allow API access" checkbox from the "Authentication" page. It could also contain the list of the supported mobile applications and some explanations on how to configure them (e.g. giving the correct URL directly on this page). It was too much code for this PR though, so I stopped.

### DIFF
--- a/app/Utils/passwordUtil.php
+++ b/app/Utils/passwordUtil.php
@@ -1,0 +1,27 @@
+<?php
+
+class FreshRSS_password_Util {
+	// Will also have to be computed client side on mobile devices,
+	// so do not use a too high cost
+	const BCRYPT_COST = 9;
+
+	/**
+	 * Return a hash of a plain password, using BCRYPT
+	 *
+	 * @param string
+	 * @return string
+	 */
+	public static function hash($passwordPlain) {
+		$passwordHash = password_hash(
+			$passwordPlain,
+			PASSWORD_BCRYPT,
+			array('cost' => self::BCRYPT_COST)
+		);
+		$passwordPlain = '';
+
+		// Compatibility with bcrypt.js
+		$passwordHash = preg_replace('/^\$2[xy]\$/', '\$2a\$', $passwordHash);
+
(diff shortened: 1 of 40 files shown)


app/Utils/passwordUtil.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+class FreshRSS_password_Util {
```
Frenzie, line 3: Could be an opportunity for that gradual namespace thing you mentioned?

marienfressinaud: It could if we decided to go with namespaces, but we finally didn't take any decisions about this. Also as I explained, I'm not really a fan of a half done transition, that's why I'm keeping the old way :)

Alkarex (approved):

Issue #446, opened 2014-03-05
Minify CSS and JS files


aledeg:
To continue the optimization process started here => #341, I think it will be a good idea to minify CSS and JS files.

There is many tools available online, here is a few:
- http://closure-compiler.appspot.com/home
- http://jscompress.com/
- http://refresh-sf.com/yui/
- http://cssminifier.com/

math-GH:
Could we include a minifier into our pipeline?

Alkarex:
> Could we include a minifier into our pipeline?

I am not so found of it. Our JavaScript / CSS code is relatively short (compared to other applications / Web sites nowadays), and the build-in gzip compress already does a good job, supplemented by HTTP caching.
It would only increase obscurity, barrier to entry, and difficulty to debug.

math-GH:
You are right. Then I vote for to close this ticket

PR #2678, opened 2019-11-19
[UI] Add RTL support with CSSJanus

### LINKED ISSUE #673: RTL support

Frenzie:
This PR supersedes #2658 and fixes #673. It's a slightly different (automated) principle, which requires significantly less manual attention. The entire file is automatically processed and adjusted. It adds the new 19 kB library CSSJanus, but that's pretty much it.

I didn't really know where to put the auto-generator code; it should presumably be its own method somewhere. It sticks the generated files in themes/rtl and regenerates them if the original was modified more recently than the generated RTL file.

Screenshot_2019-11-19_16-23-08

Screenshot_2019-11-19 (10) ההזנות שלך · FreshRSS
Screenshot_2019-11-19 (10) ההזנות שלך · FreshRSSasf
Screenshot_2019-11-19 (10) ההזנות שלך · FreshRSSl;kj
Screenshot_2019-11-19 תצוגה · FreshRSS23423

---
This change is [Reviewable](https://reviewable.io/reviews/freshrss/freshrss/2678)


app/FreshRSS.php
```diff
@@ -98,13 +98,30 @@ public static function loadStylesAndScripts() {
...
+
+				if (_t('gen.dir') === 'rtl') {
+					if (!file_exists($path_rtl) || filemtime($path_rtl) < $filetime) {
+						// generate RTL CSS
+						if (!class_exists('CSSJanus')) {
+							require_once LIB_PATH . '/lib_CSSJanus.php';
+						}
+						$content = file_get_contents(PUBLIC_PATH . '/themes/' . $theme_id . '/' . $filename_ltr);
+						$rtl_content = CSSJanus::transform($content);
+						file_put_contents($path_rtl, $rtl_content);
```
Alkarex, line 121: If this requires write access, we might consider treating the results as cache, and put them somewhere under `./data/`
We have `ext.php` to serve files that are not directly exposed publicly, which we could generalise.

Frenzie: Sure, that makes sense.

Frenzie: @Alkarex Unfortunately I just noticed a problem.

Ansum, for example, contains things like this:

```css
background-image: url(icons/magnifier.svg);
```

The less elegant approach of just generating it locally and committing it all would be an easy way around that, but of course I wrote this PR to avoid such a thing. It's an action you can forget and it adds extra noise in commits/diffs.

One possible approach would be to copy the whole thing, along these lines (except in PHP instead of shell for Windows compatibility):

```diff
diff --git a/app/FreshRSS.php b/app/FreshRSS.php
index 14995d5e..4ca026e1 100644
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -104,23 +104,27 @@ class FreshRSS extends Minz_FrontController {
                                        $filename_ltr = $file;
                                }
                                $filename_rtl = $filename_ltr . '.rtl.css';
-                               $path_ltr = PUBLIC_PATH . '/themes/' . $theme_id . '/' . $filename_ltr;
-                               $path_rtl = DATA_PATH . '/rtl/'. $theme_id . '.' . $filename_rtl;
+                               $path_ltr = PUBLIC_PATH . '/themes/' . $theme_id;
+                               $path_rtl = DATA_PATH . '/rtl/'. $theme_id;
+                               $path_full_ltr = $path_ltr . '/' . $filename_ltr;
+                               $path_full_rtl = $path_rtl . '/' . $filename_rtl;
 
-                               $filetime = @filemtime($path_ltr);
+                               $filetime = @filemtime($path_full_ltr);
                                $url = '/themes/' . $theme_id . '/' . $filename_ltr . '?' . $filetime;
 
                                if (_t('gen.dir') === 'rtl') {
-                                       if (!file_exists($path_rtl) || filemtime($path_rtl) < $filetime) {
+                                       if (!file_exists($path_full_rtl) || filemtime($path_full_rtl) < $filetime) {
+                                               `rm -rf $path_full_rtl`;
+                                               `cp -r $path_ltr $path_rtl`;
                                                // generate RTL CSS
                                                if (!class_exists('CSSJanus')) {
                                                        require_once LIB_PATH . '/lib_CSSJanus.php';
                                                }
-                                               $content = file_get_contents($path_ltr);
+                                               $content = file_get_contents($path_full_ltr);
                                                $rtl_content = CSSJanus::transform($content);
-                                               file_put_contents($path_rtl, $rtl_content);
+                                               file_put_contents($path_full_rtl, $rtl_content);
                                        }
-                                       $url = '/ext.php?f='. $theme_id . '.' . $filename_rtl . '&t=css&c=rtl&' . $filetime;
+                                       $url = '/ext.php?f='. $theme_id . '/' . $filename_rtl . '&t=css&c=rtl&' . $filetime;
                                }
 
                                Minz_View::prependStyle(Minz_Url::display($url));
```

That principle works fine, setting aside that the icon URLs still don't actually work at this point.

I looked around a bit and [this PR](https://github.com/PrestaShop/PrestaShop/pull/8657/files) simply puts them in the theme directory, but I figure having those writable isn't exactly desirable.

Also pinging @aledeg @marienfressinaud for their thoughts.

marienfressinaud: There are two things that bother me.

The first one is to put CSS theme files under the `./data` folder (actually, there are a lot of files that I would not put in this folder, it's another non-critical subject). I was even considering dropping the `ext.php` and `f.php` mechanisms by replacing them by a dynamic system of "assets collection", to move all CSS and JS under the `/p` folder and to allow pre-processing. It's a long-term idea that I will not develop soon, but just to say that I don't mind if the server write files in the public folder: I find it's more logical.

The second thing is to do that in `loadStylesAndScripts`. Writing these files is a huge side-effect of the method. I would have done it at least in a different method of this class (and yet I don't think it's the ideal solution, but I can't find one for the moment ^^)

Last thing: I would like to have the feedback of someone who is used to RTL interfaces. I find the interfaces look pretty weird (but hey, it's normal since I never use RTL interfaces ^^) and I'm wondering if the solution is to simply reverse the rules or if there are trickier things to think about.

Alkarex: A note regarding were to place files: key aspects include whether the folder is writeable by the Web server, and whether the whole folder can be erased and replaced when there is a new FreshRSS version.

marienfressinaud: True, I didn't took that in account, so it's probably the best approach at the moment. Yet I think we should have a better distinction between user-generated data (config files) and automatically-generated data (cache).

Frenzie: > just to say that I don't mind if the server write files in the public folder: I find it's more logical.

I also find it more logical and it would definitely be easier because there'd literally be nothing to it. Just stick a filename-rtl.css next to the original, done. Barely any effort involved, all correct & working.

Security probably isn't no effort though, hence why I'm asking.

> The second thing is to do that in loadStylesAndScripts. 

That's a non-issue. ;-) As I said in the OP I didn't put it in there because I think it should go there, but as a basic proof of concept of the actions that need to happen. Starting with boilerplate methods is something I find incredibly off-putting to getting things done for simple stuff. :-)

That doesn't mean I want to keep it this way! But like I said or implied in the OP, it didn't seem quite right to just toss it in another method right here. I was thinking more in the direction of a method on a localization class like Minz_Translate, or a new RTL helper class. I'd also add another method such as `isRTL()` that simply does `return _t('gen.dir') === 'rtl'` even if that's currently only repeated twice.

> Last thing: I would like to have the feedback of someone who is used to RTL interfaces

They look completely normal, except for some obvious minor details like the arrows and text not showing in the second & last screenshots, if that's what you're talking about. The script does 95 % of the work but it's not magic. ;-) See [here](https://itwadi.com/comment/4790) for KDE in RTL, for example.

Pinging @WaseemAlkurdi.

Frenzie: Apologies, GH failed to update the comments.

@marienfressinaud 
> Yet I think we should have a better distinction between user-generated data (config files) and automatically-generated data (cache).

Is that distinction sufficient? There's internal cache like SimplePie, and externally reachable cache including minified JS/CSS as well as generated RTL CSS.

WaseemAlkurdi: @Frenzie Yes, the "light mode" screenshot looks fine, except for the missing arrows you talked about (the sidebar ones mostly).
I'm a little concerned about the arrows on the bottom of the screen in the "dark mode" screenshot. They seem to be in the wrong directions, but I can't really tell because the arrows are not there in the "light" screenshots, nor is there any explanation as to what they do.

As to what RTL user interfaces look like: the "visuals" of the app are simply flipped around.
A close European friend of mine who's studying Arabic once gave me a brilliant analogy: it's like driving right-hand if you come from a left-hand-drive country.
To see a live example yourself, try the `Arabic (XB)` language (might be called `Arabic (Pseudo-Bidi)` or `cibarA ([XB])` ) on an Android phone. Don't worry, your phone would still be English (sp you could set it back, but the user interface is flipped around, and English strings start from the right, to give you a feel of how RTL users read and see.
Alternatively, use the `Force RTL layout direction` option in Developer Options on Android. This flips the interface around, but the strings are still LTR.

Frenzie: > I'm a little concerned about the arrows on the bottom of the screen in the "dark mode" screenshot. They seem to be in the wrong directions

Actually that's what I meant by arrows. Not completely sure what you mean by missing arrows, probably the downward pointing ones as part of the row as icons? :-)

Frenzie: Oh right, I see what I wrote, haha. My apologies. I meant this:

Minor details like the arrows (which are in the wrong direction)
and
some missing UI elements.

Frenzie: @WaseemAlkurdi Btw, you can see for yourself over on https://demo.freshrss.org/ if you're curious.

WaseemAlkurdi: @Frenzie I've taken a look. Very awesome project! I'm thinking of deploying it sometime.
You're correct: the bottom (previous/next) arrows are indeed reversed. Other than that, no issues at all that I can see in the screenshots above vs. the demo.

Edit: Missing UI elements? I can't see any in the screenshots above.

Frenzie: @WaseemAlkurdi I didn't see your edit. It's not the UI elements that are missing per se, but the icons on them (on the second and fourth screenshot).

WaseemAlkurdi: @Frenzie I had to stare at the screenshots for a couple minutes to even know what was missing! :-)
But true, the icons are missing (the toolbar and search bar, as far as I can see).

Alkarex:
Let me know when this is ready for another review. As I am not familiar with RTL, I am only looking at potential negative side effects (e.g. performance) on LTR.

Frenzie:
@Alkarex Unfortunately I've hit a roadblock on this approach as detailed [here](https://github.com/FreshRSS/FreshRSS/pull/2678#discussion_r348444939). So I suppose the main question is if we want it dynamically generated on the server or not.

Alkarex:
For (future?) inspiration https://css-tricks.com/css-logical-properties/
We could also have some themes, which drop support for IE/Edge
https://caniuse.com/#feat=css-logical-props

Frenzie:
I was aware of those when I started, yes. :-) Also see https://github.com/FreshRSS/FreshRSS/pull/2658#issuecomment-553327280

Frenzie:
@Alkarex @marienfressinaud @aledeg For the short term I think it would be best to change this into a simple CSSJanus call on the development machine, for example through `make rtl`, and to commit the result. Reworking the cache system to facilitate the likes of https://github.com/FreshRSS/FreshRSS/issues/446 and this (as described by @marienfressinaud in https://github.com/FreshRSS/FreshRSS/pull/2678#discussion_r349868639) is a time investment beyond the scope of what I'm interested in right now, and otherwise it'd just delay RTL a lot longer.

Since that shouldn't take long to draft up I'll probably open a third PR to show what that looks like, leaving this one as is. All of my effort here was intended to avoid putting it in the repo, but I don't see a clear weekend-hack path toward actually doing so right now.

Frenzie:
Merged #2776 instead.

PR #2680, opened 2019-11-20
Old ICU: INTL_IDNA_VARIANT_2003 fallback

### LINKED ISSUE #2676: Failed to add RSS address
I can't add an RSS source at https://www.cnbeta.com/backend.php. And I successfully added it to one of your test sites. Is it the database? I use SQLite.
LOG: ' ' is not a valid URL.

### LINKED ISSUE #2677: Support request: cannot add any feed
Hello,

First of all, I'm new to this, please excuse my potential lack of experience in how the app should work.
I've been trying this with the latest master, currently on the dev branch.

When I'm trying to add a feed, all I get is URL (feed URL) is invalid for all of the followings:
https://appleinsider.com/rss/news/
http://feeds.feedburner.com/Clamav
http://feeds.feedburner.com/CssTricks
https://www.exploit-db.com/rss.xml

Pretty much any feed I try result in this.
I'm not seeing anything flagged up in nxinx and phpfpm logs only the access log hit comes through with a 200 ok.
curl -v -A "SimplePie/1.7" <any_of_the_above> will return the feed just fine.

Found the user log, it says:
[Tue, 19 Nov 2019 14:39:54 +0100] [warning] --- `` is not a valid URL

At first I thought this is UI specific, but OPML import results the same:
[Tue, 19 Nov 2019 14:45:54 +0100] [warning] --- `` is not a valid URL
[Tue, 19 Nov 2019 14:45:54 +0100] [warning] --- Error during OPML feed import from URL: https://realpython.com/atom.xml in category 24

What's the best way for me to debug this further?

Running on LAN (e.g. not exposed to the interwebs, on an internal news.local domain).
Nginx: 1.16.1
PHP 7.2.22 (yeah modcurl in).
MariaDB: 10.2.26
(Base is a DDWRT router, above are the latest available from opkg/entware).

Thanks,
mb

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2676
Fix https://github.com/FreshRSS/FreshRSS/issues/2677
ICU version < 4.6: When INTL_IDNA_VARIANT_UTS46 is not available, use INTL_IDNA_VARIANT_2003
Cf. https://github.com/PrestaShop/PrestaShop/pull/11995

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -49,9 +49,8 @@ spl_autoload_register('classAutoloader');
 
 function idn_to_puny($url) {
 	if (function_exists('idn_to_ascii')) {
-		$parts = parse_url($url);
-		if (!empty($parts['host'])) {
-			$idn = $parts['host'];
+		$idn = parse_url($url, PHP_URL_HOST);
+		if ($idn != '') {
 			// https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003
 			if (defined('INTL_IDNA_VARIANT_UTS46')) {
 				$puny = idn_to_ascii($idn, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -52,7 +52,13 @@ function idn_to_puny($url) {
 		$parts = parse_url($url);
 		if (!empty($parts['host'])) {
 			$idn = $parts['host'];
-			$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
+			if (defined('INTL_IDNA_VARIANT_UTS46')) {
+				$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
+			} elseif (defined('INTL_IDNA_VARIANT_2003')) {
+				$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_2003);
+			} else {
+				$puny = idn_to_ascii($idn);
+			}
 			$pos = strpos($url, $idn);
 			if ($pos !== false) {
 				return substr_replace($url, $puny, $pos, strlen($idn));
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -52,6 +52,7 @@ function idn_to_puny($url) {
 		$parts = parse_url($url);
 		if (!empty($parts['host'])) {
 			$idn = $parts['host'];
+			// https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003
 			if (defined('INTL_IDNA_VARIANT_UTS46')) {
 				$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
 			} elseif (defined('INTL_IDNA_VARIANT_2003')) {
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -54,15 +54,15 @@ function idn_to_puny($url) {
 			$idn = $parts['host'];
 			// https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003
 			if (defined('INTL_IDNA_VARIANT_UTS46')) {
-				$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
+				$puny = idn_to_ascii($idn, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
 			} elseif (defined('INTL_IDNA_VARIANT_2003')) {
-				$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_2003);
+				$puny = idn_to_ascii($idn, IDNA_DEFAULT, INTL_IDNA_VARIANT_2003);
 			} else {
 				$puny = idn_to_ascii($idn);
 			}
 			$pos = strpos($url, $idn);
-			if ($pos !== false) {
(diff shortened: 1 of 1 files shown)


Frenzie (commented):
lgtm, might be useful to add some explanation in the comments?

Like that this one's going away: https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003

Frenzie (approved):

aledeg (approved):

lib/lib_rss.php
```diff
@@ -52,7 +52,14 @@ function idn_to_puny($url) {
 		$parts = parse_url($url);
 		if (!empty($parts['host'])) {
 			$idn = $parts['host'];
-			$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
+			// https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003
+			if (defined('INTL_IDNA_VARIANT_UTS46')) {
+				$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
```
marienfressinaud: Why don't you pass `IDNA_DEFAULT` constant as second argument instead of `0`? Here we don't know to which constant corresponds `0` unless by guessing or logging the `IDNA_DEFAULT` constant.

fxzxmicah:
This problem seems to have solved a large part. Now I can use freshrss, but it seems that there is still a problem: the image of RSS is not displayed?

lib/lib_rss.php
Alkarex: Done 👍

Alkarex:
@fxzxmicah 1) Please try the new version of the patch. 2) Please open another issue for the favicon problem (and check whether this is for one feed or all feeds)

fxzxmicah:
Maybe this is not an issue, because when I click this button on the right-click menu of the browser (picture at bottom), the image can be loaded, and it needs to be clicked one by one. But I didn't encounter this situation on your preview site with the same browser before.

Frenzie:
Could be some leftover cache from your earlier attempts that needed refreshing?

fxzxmicah:
I don't know. It's strange that I haven't seen this option on the right-click menu before.

Frenzie:
Presumably your browser only offers it when an image fails to load. Some browsers always offer "Reload image" in the image context menu. It looks like Firefox 68 doesn't; there you have to choose "View image" (or "Copy image location") and press F5.

Alkarex:
@fxzxmicah and @mb-0 Could you please test one last time the patch, which has just been updated, to ensure that it works in your case when adding new feeds?

fxzxmicah:
In fact, my problem was solved in the first commit, so I didn't feel the difference. Moreover, the last commit did not find any problems for the time being.

mb-0:
@Alkarex tested with a fresh clone, looks all OK on my end, thanks again!

mb

Alkarex:
Thanks for the feedbacks :-)

Alkarex:
@marienfressinaud You might have a similar issue in the e-mail part

marienfressinaud:
This part looks fine: we use the function from PHPMailer which uses `INTL_IDNA_VARIANT_UTS46`

Alkarex:
@marienfressinaud Exactly, and `INTL_IDNA_VARIANT_UTS46` might not be available (this is what we fix in the current PR)

Alkarex:
By the way, we should add a test for `INTL_IDNA_VARIANT_UTS46` (modern ICU) in the install check.

marienfressinaud:
If I'm not wrong, you fixes the absence of `INTL_IDNA_VARIANT_2003`. The PHP documentation states that `INTL_IDNA_VARIANT_UTS46` exists from ICU 4.6 which has been released [in March 2011](http://site.icu-project.org/download/46). I think it is safe to consider it'll exist, isn't it?

Alkarex:
The current PR is mainly to fix the absence of `INTL_IDNA_VARIANT_UTS46`, using the old/deprecated `INTL_IDNA_VARIANT_2003` as the first fallback

marienfressinaud:
Aah, you're right, I was confused because I saw `INTL_IDNA_VARIANT_2003` in the title! Edit: also, it's the morning :p

Alkarex:
There is a great summary on https://github.com/PrestaShop/PrestaShop/pull/11995 (January 2019)

lib/lib_rss.php
```diff
@@ -49,13 +49,19 @@ function classAutoloader($class) {
...
-			$puny = idn_to_ascii($idn, 0, INTL_IDNA_VARIANT_UTS46);
+		$idn = parse_url($url, PHP_URL_HOST);
+		if ($idn != '') {
+			// https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003
+			if (defined('INTL_IDNA_VARIANT_UTS46')) {
+				$puny = idn_to_ascii($idn, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
+			} elseif (defined('INTL_IDNA_VARIANT_2003')) {
+				$puny = idn_to_ascii($idn, IDNA_DEFAULT, INTL_IDNA_VARIANT_2003);
+			} else {
+				$puny = idn_to_ascii($idn);
```
marienfressinaud, line 60: @Alkarex I'm just realizing that you didn't pass the options parameter here. Is there any reasons for that or did you forget?

Alkarex: That was on purpose, as the last fallback, using any default parameter a future version of PHP might have. It is the same strategy than https://github.com/PrestaShop/PrestaShop/pull/11995

Issue #2385, opened 2019-04-25
Add an option to change/expand behaviour of shift+j


stevanov:
Currently Shift+J moves to the next feed, however when i have all articles in that next feed marked read and have 'Show unread' selected then its somewhat useless to go to the next 'empty' feed.

Could be solved by the following solutions;

- when 'Shown only unread' is selected, shift+j automatically skips to the next feed that has unreads
- an additonal shortcut such as alt+j or something that skips to the next feed that has unread articles
- a checkbox option in settings that is able to turn off or on the described behaviour of shift+j

I'm a former inoreader user and this is one of the larger differences in user experience i've noticed thusfar.

For me this would be an especially useful function (and probably one that i could solve with the custom JS addon), as long as issue #1979 isn't solved yet

(For clarification, pressing 'Mark all as read' already follows this behavior of switching to the next feed with unread articles)

stevanov:
Not sure if this is good form, but I propose changing line 483 in main.js (the next_feed function)
https://github.com/FreshRSS/FreshRSS/blob/d008b6c1a7488b99e77287ba14f1fff3ec1a116b/p/scripts/main.js#L483
AND 466
https://github.com/FreshRSS/FreshRSS/blob/d008b6c1a7488b99e77287ba14f1fff3ec1a116b/p/scripts/main.js#L466

to something like

`if (found && feed.attributes["data-unread"].textContent !== "0") {  `

I've tested this to work on my local instance of FreshRSS

Ideally, we should add a function to the settings menu so that end users have the choice between the skipping and non-skipping behaviour.

Alkarex:
Sorry for the delay. Your suggestion looks reasonable. `feed.getAttribute('data-unread')` should work. Could you make a pull request?

stevanov:
> Sorry for the delay. Your suggestion looks reasonable. `feed.getAttribute('data-unread')` should work. Could you make a pull request?

No problem, feed.getAttribute('data-unread') seems like a neater solution than mine. 
That just leaves one design choice, adding this hard coded will force this behaviour onto everyone which i don't think is desirable.

So i either need to add a checkmark in the options or add a new modifier key for J and K (ctrl+shift?). I think i personally would prefer the checkmark in the settings but i'd appreciate your input on this decision. 

And if so, i would appreciate a pointer to what are all the files i need to change for the addition of a new settings menu option. I've had a look into those files to create a pull request before but i wasnt able to figure out the things i needed to change.

Thanks

[...]

PR #2687, opened 2019-11-25
Update navigation on empty feeds

aledeg:
When using feed navigation, the previous behavior was to cycle through all
available feeds regardless of their content. To match the behaviour of the
first feed and last feed navigation shortcuts, the navigation now skips
empty feeds. Now it's consistent through out the application.
When using feed navigation with only empty feeds, the new behavior is to
cycle through all available feeds.

See #2385

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -461,34 +461,58 @@ function next_entry(skipping) {
 
 function prev_feed() {
 	let found = false;
+	let adjacent = null;
 	const feeds = document.querySelectorAll('#aside_feed .feed');
 	for (let i = feeds.length - 1; i >= 0; i--) {
 		const feed = feeds[i];
-		if (found && getComputedStyle(feed).display !== 'none') {
-			delayedClick(feed.querySelector('a.item-title'));
-			break;
-		} else if (feed.classList.contains('active')) {
+		if (feed.classList.contains('active')) {
 			found = true;
+			continue;
+		}
+		if (!found) {
+			continue;
+		}
+		if (getComputedStyle(feed).display === 'none') {
+			continue;
+		}
+		if (feed.dataset.unread != 0) {
+			return delayedClick(feed.querySelector('a.item-title'));
+		} else if (adjacent === null) {
+			adjacent = feed;
 		}
 	}
-	if (!found) {
+	if (found) {
+		delayedClick(adjacent.querySelector('a.item-title'));
+	} else {
 		last_feed();
 	}
 }
 
 function next_feed() {
 	let found = false;
+	let adjacent = null;
 	const feeds = document.querySelectorAll('#aside_feed .feed');
 	for (let i = 0; i < feeds.length; i++) {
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -464,11 +464,16 @@ function prev_feed() {
...
 		const feed = feeds[i];
-		if (found && getComputedStyle(feed).display !== 'none') {
+		if (feed.classList.contains('active')) {
+			found = true;
+			continue;
+		}
+		if (!found) {
+			continue;
+		}
+		if (feed.dataset.unread != 0 && getComputedStyle(feed).display !== 'none') {
```
Frenzie: I'm only taking a quick glance, but is this even reachable? The first if statement sets found to true and immediately continues to the next iteration, while the one directly following continues if found isn't true?

aledeg: This is reachable. I've tested it with many different configurations and they all worked.
The first if statement catches the current feed. The second if statement is there to not process feeds until we found the current feed. The third if statement does the job of switching feeds.

I hope this makes sense!

Alkarex (approved):

marienfressinaud (commented):
The shortcuts don't work anymore if there is no unread articles, isn't it? So we can't navigate when all articles are read

aledeg:
@marienfressinaud that's right. We can add something to fix that. I don't know if it's useful and/or needed.

marienfressinaud:
I never use these shortcuts so I may not be the best person to take this decision, but I think it's strange to not allow navigation in this situation. Also, it would be not consistent with categories and entries navigation.

aledeg:
@marienfressinaud now, shortcuts cycle through all feeds when all feeds are read.

marienfressinaud (approved):
I'm ok with the behaviour.

(nitpick) About the code, I think you can refactor a bit by using another variable than `adjacent` (I would call it `next_feed` or `prev_feed` but it's already the names of the functions). A simplified algo is:

```js
// ...
for(...) {
    // ...
    if (getComputedStyle(feed).display === 'none') { 
        continue
    }

    if (feed.unread !== 0 || next_feed === null) {
        next_feed = feed
    }
}

if (next_feed) {
    delayedClick(next_feed)
} else {
    last_feed()
}
```

You also can distinguish unread test from the null test to benefit of the early return in the first case.

aledeg:
@marienfressinaud I've made some changes but there is something that's not working in what you mentioned. I need a way of separating the next feed (read or not) from the next not read feed.

> ```js
> if (feed.unread !== 0 || next_feed === null) {
>     next_feed = feed
> }
> ```

I think this wont

Alkarex:
@aledeg I have not understood your last comment :-) Do you want to make additional changes?

aledeg:
@Alkarex I want pointing out that the snippet that @marienfressinaud wrote was probably broken. I took into consideration some of the comments and refactor the code. I think this PR is complete. :)

Issue #2697, opened 2019-12-02
get_magic_quotes_gpc() is deprecated in PHP 7.4


candrews:
PHP 7.4 has deprecated `get_magic_quotes_gpc()`:
https://www.php.net/manual/en/function.get-magic-quotes-gpc.php
https://www.php.net/manual/en/security.magicquotes.php

FreshRSS uses this function, resulting in this warning when running `actualize_script.php`:
```
PHP Deprecated:  Function get_magic_quotes_gpc() is deprecated in /var/www/localhost/htdocs/FreshRSS/lib/Minz/Request.php on line 303
```

aledeg:
I'll add that magicquotes are deprecated since PHP 5.4. It's time to make some cleaning.

Alkarex:
We should just delete the `magicQuotesOff()` function

Frenzie:
Yup, apparently completely obsolete since 5.4. Good riddance. ;-)

Alkarex:
Fixed by https://github.com/FreshRSS/FreshRSS/pull/2698

PR #2698, opened 2019-12-02
Remove magicQuotesOff (#2697)

candrews:
Magic quotes have been deprecated since PHP 5.4 and as of PHP 7.4 `get_magic_quotes_gpc()` displays a warning.


Frenzie (commented):
It looks like `Minz_Helper::stripslashes_r` exists exclusively as a helper for this file, so it seems to me that should also be removed.

This PR fixes #2697.

Alkarex (approved):

PR #2703, opened 2019-12-06
Add user language display

aledeg:
When managing users, I've added the language they selected to be able
to communicate with them with the selected language (if I know it).
It could be useful when managing a community or a paid service.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ ifndef value
 	@echo To add a key, you need to provide its value in the "value" variable.
 	@exit 10
 endif
-	@$(PHP) ./cli/manipulate.translation.php -a add -k $(key) -v $(value)
+	@$(PHP) ./cli/manipulate.translation.php -a add -k $(key) -v "$(value)"
 	@echo Key added.
 
 .PHONY: i18n-remove-key
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -535,10 +535,14 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 		$entryDAO = FreshRSS_Factory::createEntryDao($username);
 		$databaseDAO = FreshRSS_Factory::createDatabaseDAO($username);
 
+		$userConfiguration = get_user_configuration($username);
+
 		return array(
 			'feed_count' => $feedDAO->count(),
 			'article_count' => $entryDAO->count(),
 			'database_size' => $databaseDAO->size(),
+			'language' => $userConfiguration->language,
+			'mail_login' => $userConfiguration->mail_login,
 		);
 	}
 }
--- a/app/i18n/cz/admin.php
+++ b/app/i18n/cz/admin.php
@@ -188,6 +188,7 @@ return array(
 		'create' => 'Vytvořit nového uživatele',
 		'database_size' => 'Database size',	// TODO - Translation
 		'delete_users' => 'Delete user',	// TODO - Translation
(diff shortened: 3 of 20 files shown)


app/views/user/manage.phtml
```diff
@@ -62,6 +62,7 @@
 		<thead>
 			<tr>
 				<th><?= _t('admin.user.username') ?></th>
+				<th><?= _t('admin.user.language') ?></th>
```
marienfressinaud, line 66: I'm not sure it's useful to have it in the list, we should pick only the most useful information to avoid a too big table. What do you think?

aledeg: I don't mind removing it. I feel it could be useful in the list unlike other information. I agree that only useful information should be displayed in the list.

marienfressinaud: I guess it depends on the use case ^^

Alkarex:
I think showing the e-mail address would be useful, in which case the language would be informative

aledeg:
@Alkarex where can I find the email address?

**Edit** Is the `mail_login` value the email address?

aledeg:
I found it and I've updated the list

app/Controllers/userController.php
```diff
@@ -535,10 +535,18 @@ private function retrieveUserDetails($username) {
 		$entryDAO = FreshRSS_Factory::createEntryDao($username);
 		$databaseDAO = FreshRSS_Factory::createDatabaseDAO($username);
 
+		Minz_Configuration::register("user-{$username}",
+		                             join_path(USERS_PATH, $username, 'config.php'),
+		                             join_path(FRESHRSS_PATH, 'config-user.default.php'),
+		                             new FreshRSS_ConfigurationSetter());
+		$userConfiguration = Minz_Configuration::get("user-{$username}");
```
marienfressinaud: You can use the `get_user_configuration($username)` function here

app/i18n/cz/admin.php
```diff
@@ -191,6 +191,7 @@
 		'feed_count' => 'Feeds',	// TODO - Translation
 		'language' => 'Jazyk',
 		'list' => 'User list',	// TODO - Translation
+		'mail' => 'Email',	// TODO - Translation
```
marienfressinaud: I would use `email` instead of `mail` to be coherent with what I use in the email validation feature. Renaming `mail_login` in `email` is one of the task I plan to do later (but I have to do some other things before ^^)

app/Controllers/userController.php
aledeg: Oh I didn't know that function exists!

app/i18n/cz/admin.php
aledeg: Sure!

aledeg:
@marienfressinaud Done

Makefile
```diff
@@ -66,7 +66,7 @@ ifndef value
 	@echo To add a key, you need to provide its value in the "value" variable.
 	@exit 10
 endif
-	@$(PHP) ./cli/manipulate.translation.php -a add -k $(key) -v $(value)
+	@$(PHP) ./cli/manipulate.translation.php -a add -k $(key) -v "$(value)"
```
marienfressinaud, line 69: Nitpick: it would have been better to have it in a separate commit to explain why you changed that ;)

marienfressinaud (approved):
lgtm :+1:

Makefile
aledeg: Done

PR #2704, opened 2019-12-06
extend hooks

cn-tools:
I added two new hooks to have more options for influencing with extensions.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -47,6 +47,12 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 
 		$url = trim($url);
 
+		/** @var $url */
+		$url = Minz_ExtensionManager::callHook('check_url_before_add', $url);
+		if ($url === null) {
+			throw new FreshRSS_FeedNotAdded_Exception($url, $title);
+		}
+
 		$cat = null;
 		if ($new_cat_name != '') {
 			$new_cat_id = $catDAO->addCategory(array('name' => $new_cat_name));
@@ -274,6 +280,12 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		$updated_feeds = 0;
 		$nb_new_articles = 0;
 		foreach ($feeds as $feed) {
+			/** @var FreshRSS_Feed $feed */
+			$feed = Minz_ExtensionManager::callHook('feed_before_actualize', $feed);
+			if ($feed === null) {
+				continue;
+			}
+
 			$url = $feed->url();	//For detection of HTTP 301
 
 			$pubSubHubbubEnabled = $pubsubhubbubEnabledGeneral && $feed->pubSubHubbubEnabled();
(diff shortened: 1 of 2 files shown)


aledeg:
Thank you for your contribution.
Could you explain what are the use case for those new hooks? I guess that if you add those, you want new features for your extensions.

cn-tools:
Dear @aledeg !
I develop a new extension who should change the url in the add process. My idea is to change a YouTube Channel URL into an YouTube RSS Feed URL. (hook->'check_url_before_add')

The hook 'feed_before_actualize' is only a abstract idea :)

Have a look:
https://github.com/cn-tools/cntools_FreshRssExtensions

lib/Minz/ExtensionManager.php
```diff
@@ -23,10 +23,18 @@ class Minz_ExtensionManager {
 			'list' => array(),
 			'signature' => 'OneToOne',
 		),
+		'check_url_before_add' => array(  // function($url) -> Url | null
+			'list' => array(),
+			'signature' => 'OneToOne',
+		),
```
aledeg: Move that definition above to keep the list ordered.

lib/Minz/ExtensionManager.php
```diff
@@ -23,10 +23,18 @@ class Minz_ExtensionManager {
...
 		'feed_before_insert' => array(  // function($feed) -> Feed | null
 			'list' => array(),
 			'signature' => 'OneToOne',
 		),
+		'feed_before_actualize' => array(  // function($feed) -> Feed | null
+			'list' => array(),
+			'signature' => 'OneToOne',
+		),
```
aledeg: Same here

app/Controllers/feedController.php
```diff
@@ -274,6 +280,12 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 		$updated_feeds = 0;
 		$nb_new_articles = 0;
 		foreach ($feeds as $feed) {
+			/** @var FreshRSS_Feed $feed */
+			$feed = Minz_ExtensionManager::callHook('feed_before_actualize', $feed);
+			if ($feed === null) {
```
aledeg: Use `null === $feed` instead

app/Controllers/feedController.php
```diff
@@ -47,6 +47,12 @@ public static function addFeed($url, $title = '', $cat_id = 0, $new_cat_name = '
 
 		$url = trim($url);
 
+		/** @var $url */
+		$url = Minz_ExtensionManager::callHook('check_url_before_add', $url);
+		if ($url === null) {
```
aledeg: Use `null === $feed` instead

aledeg (changes requested):
Apart from minor changes, this looks good to me.

cn-tools:
thx 4 comments
change typography ... done

aledeg (approved):

Alkarex:
@Frenzie I am not up to date on how to change the documentation now; we should add the new hooks to the documentation in https://github.com/FreshRSS/FreshRSS/blob/master/docs/en/developers/03_Backend/05_Extensions.md#the-hooks-system

Frenzie:
@Alkarex The English docs are the source, then there's a little helper script to generate the rest. Unfortunately you need po4a 0.56 which may not be readily available yet. That's just these two? I can submit a PR with that.

Alkarex:
Yes please @Frenzie 👍

PR #2707, opened 2019-12-07
tec: Add ~ID~ placeholder in share system

marienfressinaud:
I'm currently playing on a small extension to improve the email sharing
system. It allows to create a message and send an email directly through
FreshRSS. For that, I need to retrieve the shared article. I could use
the link or the title but it would be less robust and less efficient.

Reference: https://github.com/FreshRSS/FreshRSS/issues/300


Frenzie (approved):
Seems fine. :-)

PR #2708, opened 2019-12-07
tec: Allow to change CSP header from controllers

marienfressinaud:
For an extension, I needed to call a script from an external domain.
Unfortunately, the CSP headers didn't allow this domain and I had to
patch manually the FreshRSS FrontController for my extension. It's
obviously not a long-term solution since it has nothing to do in the
core of FRSS, and I don't want to apply this patch manually at each
update.

With this patch, I allow changing the CSP header from inside the
controller actions. It allows extensions to modify headers. It's also an
opportunity to remove a bit of code from the FrontController. I wasn't
happy with the previous implementation anyhow.

Reference: https://github.com/flusio/xExtension-Flus/commit/ed12d56#diff-ff12e33ed31b23bda327499fa6e84eccR143

### DIFF
--- a/app/Controllers/indexController.php
+++ b/app/Controllers/indexController.php
@@ -83,6 +83,13 @@ class FreshRSS_index_Controller extends Minz_ActionController {
 				Minz_Error::error(404);
 			}
 		};
+
+		$this->_csp([
+			'default-src' => "'self'",
+			'frame-src' => '*',
+			'img-src' => '* data:',
+			'media-src' => '*',
+		]);
 	}
 
 	/**
@@ -121,6 +128,13 @@ class FreshRSS_index_Controller extends Minz_ActionController {
 			$title = '(' . FreshRSS_Context::$get_unread . ') ' . $title;
 		}
 		Minz_View::prependTitle($title . ' · ');
+
+		$this->_csp([
+			'default-src' => "'self'",
+			'frame-src' => '*',
+			'img-src' => '* data:',
+			'media-src' => '*',
+		]);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


app/Controllers/indexController.php
```diff
@@ -210,6 +215,30 @@ private function updateContext() {
...
+		];
+
+		$urls_to_authorize = array_filter(array_map(function ($a) {
+			if (isset($a['method']) && $a['method'] === 'POST') {
+				return $a['url'];
+			}
+		}, FreshRSS_Context::$user_conf->sharing));
+		if ($urls_to_authorize) {
+			$urls = implode(' ', $urls_to_authorize);
+			$csp_policies['connect-src'] = "'self' " . $urls;
```
Alkarex: I cannot remember why we need this logic. Could you please refresh my memory @aledeg ?
`connect-src` is not used for Form POST, is it?
https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src

https://github.com/FreshRSS/FreshRSS/blob/82851d2039f619f1b2558e06b04a9e47fceeea54/p/scripts/main.js#L909-L915

aledeg: @Alkarex I cannot remember either :'(

marienfressinaud: It was added for Mastodon in afffbfce0, outdated by d3942b296 (there is no longer "POST" share methods).

marienfressinaud: Maybe we can just clean up a bit this part?

Alkarex: Yes indeed, I think we should remove this part, and re-consider how to add it next time it becomes needed.

marienfressinaud: Done. I also removed CSP declaration in `rssAction`, I think it's useless

Alkarex (dismissed):

PR #2709, opened 2019-12-07
Add an admin flag on users

aledeg:
Now FRSS supports more than one admin. Admins have the same rights as
the default user. Admins can promote or demote other users. The default
user is considered as an admin even if it does not have the admin flag
enabled.

See #2096


app/views/user/details.phtml
```diff
@@ -1,9 +1,11 @@
 <?php $this->partial('aside_configure'); ?>
 
+<?php $isAdmin = $this->details['is_admin']; ?>
+
 <div class="post">
     <a href="<?= _url('user', 'manage'); ?>"><?= _t('admin.user.back_to_manage'); ?></a>
 
-    <legend><?= $this->username; ?></legend>
+    <legend><?= $this->username ?><?php if ($isAdmin) echo ' &#x2015; ', _t('admin.user.admin'); ?></legend>
```
Alkarex:
```suggestion
    <legend><?= $this->username ?><?php if ($isAdmin) echo ' ― ', _t('admin.user.admin'); ?></legend>
```

I prefer Unicode than entities when possible :-)

app/views/user/manage.phtml
```diff
@@ -74,6 +82,7 @@
 			<?php foreach ($this->users as $username => $values) : ?>
 				<tr>
 					<td><?= $username ?></td>
+					<td><?= $values['is_admin'] ? '&#x2714;' : '&nbsp;' ?></td>
```
Alkarex:
```suggestion
					<td><?= $values['is_admin'] ? '✔' : ' ' ?></td>
```

Alkarex (dismissed):

app/views/user/details.phtml
aledeg: I'll change that

aledeg:
@Alkarex done!

app/Controllers/userController.php
```diff
@@ -516,6 +526,30 @@ public function deleteAction() {
 		Minz_Request::forward($redirect_url, true);
 	}
 
+	public function switchAdminAction() {
+		if (!FreshRSS_Auth::hasAccess('admin')) {
```
Alkarex, line 540: Only POST should be allowed there, to avoid CSRF attacks

app/Controllers/userController.php
```diff
@@ -516,6 +526,30 @@ public function deleteAction() {
...
+		}
+
+		$username = Minz_Request::param('username');
+		if (!FreshRSS_UserDAO::exists($username)) {
+			Minz_Error::error(404);
+		}
+
+		$userConfig = get_user_configuration($username);
+
+		$userConfig->_param('is_admin', !$userConfig->is_admin);
```
Alkarex: I have not checked the UI, but can't you pass a boolean instead of toggling?

app/Controllers/userController.php
aledeg: Noted. I'll add that.

app/Controllers/userController.php
aledeg: @Alkarex I am not sure I understand your request!

Alkarex: This code seems to be toggling the status, right? (i.e. false it it was true, true if it was false)
However, this means that requests are not idempotent, meaning that if a request is sent again (e.g. due to partial connectivity problem), it might not produce the intended effect.
So I am suggesting to pass the desired state as parameter

aledeg: Ok, but the thing is I've wanted to make a toggle. So you mean that the toggle should occur in the call not in the application of the call. Right?

Alkarex: Yes, that would be fine :-)

aledeg: I'll change that.

aledeg:
@Alkarex done

app/Controllers/userController.php
aledeg: I've made some changes. I am hoping that I understood correctly what you meant

app/Models/Auth.php
```diff
@@ -115,14 +115,15 @@ public static function giveAccess() {
...
+		$systemConfiguration = Minz_Configuration::get('system');
+		$userConfiguration = Minz_Configuration::get('user');
+		$default_user = $systemConfiguration->default_user;
 		$ok = self::$login_ok;
 		switch ($scope) {
 		case 'general':
 			break;
 		case 'admin':
-			$ok &= Minz_Session::param('currentUser') === $default_user;
+			$ok &= $default_user === Minz_Session::param('currentUser') || $userConfiguration->is_admin;
```
marienfressinaud, line 126: It would be nice to get rid of the concept of default user here, but we would need to set the `is_admin` attribute to true for default user in some kind of migration. Several options:

- you find a quick solution and you implement it
- you wait for me to implement a migration system (I'll start today :) )
- we merge as it is, we'll perform a migration later

I have no real preference for any of these options

app/views/user/manage.phtml
```diff
@@ -26,6 +26,13 @@
 			</div>
 		</div>
 
+		<div class="form-group">
+			<label for="new_user_is_admin" class="group-name"><?= _t('admin.user.is_admin') ?></label>
+			<div class="group-controls">
+				<input type="checkbox" name="new_user_is_admin" id="new_user_is_admin">
+			</div>
```
marienfressinaud, line 35: Maybe we can put the label at the right of the checkbox, as it is for most of the other checkboxes?

app/Controllers/userController.php
```diff
@@ -516,6 +528,42 @@ public function deleteAction() {
...
+		$this->switchAdminAction(false);
+	}
+
+	private function switchAdminAction($isAdmin) {
+		if (!FreshRSS_Auth::hasAccess('admin')) {
+			Minz_Error::error(403);
+		}
+
+		if (!Minz_Request::isPost()) {
+			return;
```
Alkarex: Maybe that should return a 403 as well

app/views/user/manage.phtml
aledeg: Ok, I'll work on that

app/Controllers/userController.php
aledeg: Ok, I'll work on that

app/views/user/manage.phtml
aledeg: Done

app/Controllers/userController.php
aledeg: Done

app/Models/Auth.php
Alkarex: We also need to make sure that a non-admin user cannot self-promote to admin, for instance during an export/import

app/Controllers/userController.php
```diff
@@ -516,6 +528,42 @@ public function deleteAction() {
...
+		if (!Minz_Request::isPost()) {
+			Minz_Error::error(403);
+		}
+
+		$username = Minz_Request::param('username');
+		if (!FreshRSS_UserDAO::exists($username)) {
+			Minz_Error::error(404);
+		}
+
+		$userConfig = get_user_configuration($username);
```
```suggestion
		$userConfig = get_user_configuration($username);
		if ($userConfig == null) {
			Minz_Error::error(500);
		}
```

Alkarex (approved):

app/Controllers/userController.php
Alkarex: `get_user_configuration()` might return null

aledeg: Done in a different way

Alkarex:
Two minor comments for future improvements:
1. It would be nice to add an indication of which is the default user in the list of users
2. We need to `FreshRSS_UserDAO::touch($username);` after changing the user's properties to invalidate the targetted user's cache

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/7783

PR #2721, opened 2019-12-21
tec: Fix test suite

marienfressinaud:
I realized that unit tests weren't executed on Travis. While working on
this file to enable these tests, I started to think we could simplify
it.

I separated jobs so:

- PHP linter and tests are only performed on PHP 7.3
- Translations are tested separatly so they can fail (it was already the
  case but it was hard to understand the way it was done)
- PHP 5.6 only checks syntax issues
- the last job checks css, js, etc. (it didn't change)

PHPUnit is not executed on 5.6 because only the version 5 is available
while the latest version is the 8 (https://phpunit.de/supported-versions.html).

I think it's easier to read (each job is more explicit) but I'm not a
Travis expert so maybe there's some room for improvements.


marienfressinaud:
~~Ah, my `allow_failures` system doesn't work anymore!~~ Fixed

app/Models/Category.php
```diff
@@ -84,7 +84,8 @@ public function _id($id) {
 		}
 	}
 	public function _name($value) {
-		$this->name = trim($value);
+		// regression of b323ed084?
+		$this->name = mb_strcut(trim($value), 0, 255, 'UTF-8');
```
marienfressinaud, line 87: @Alkarex I need a check here because of https://github.com/FreshRSS/FreshRSS/commit/b323ed084#diff-4618a3f4bd008df547c9bbc41e2bd145L71-R71 I don't see why it was changed and it doesn't seem to be related to the commit

Frenzie:
> PHPUnit is not executed on 5.6 because only the version 5 is available

Is there a particular reason this is an issue? The tests are from 2015.

marienfressinaud:
Yes, `createMock` and (I think) `expectException` don't exist. Another solution is to use PHPUnit 5 instead of the 8 for PHP 7.3

Frenzie:
> Another solution is to use PHPUnit 5 instead of the 8 for PHP 7.3

Something along those lines makes more sense to me, because PHP 7.2+ is probably what we run while writing code.

By along those lines I mean that for example PHPUnit 6/PHP 7.0 could be an interesting compromise for using features like createMock. (I can't seem to find when these methods were added.)

app/Models/Category.php
Alkarex: Indeed, that looks like a mistake

Alkarex (approved):

marienfressinaud:
> By along those lines I mean that for example PHPUnit 6/PHP 7.0 could be an interesting compromise for using features like createMock.

`createMock` is not blocking, we still can switch back to `getMock`. What bother me is that PHPUnit 5 and 6 are no longer supported. In the mean time, it's only used for tests, it's not like it introduces security issues in FRSS (at least, I don't think so…)

Question is: for which versions do we want to run the test suite? I implicitly answered by "PHP 7.3" because it was an easy solution: the syntax is checked for PHP 5.6 and 7.3, and the FRSS logic doesn't differ between PHP versions. Problems can happen if default values change in PHP functions from a version to another though, or using a non-existing functions. My preference was to use an up-to-date version of PHPUnit, considering that our test coverage is so low that it wouldn't make a big difference to test on several PHP versions. It's a short term opinion but we'll probably stop support of PHP 5.6 and 7.0 in few months/years (official support [stopped 1 year ago](https://www.php.net/supported-versions.php)).
→ Yet, I'm all opened to change to use PHPUnit 5 or 6 or 7, it's really not a big deal and it has its advantages too :)

Frenzie:
It doesn't matter too much for now, but in principle I think the tests should be run on multiple PHP versions. It's about behavior, not about syntax.

marienfressinaud:
I've setup tests for 7.1 and 7.3. I think it's more important to test latest supported versions than oldest. Also it allows to use PHPUnit 7 which is still supported for few weeks.

Frenzie (approved):
Lgtm

Issue #2722, opened 2019-12-21
Add make lint and make test targets


marienfressinaud:
While I was working on https://github.com/FreshRSS/FreshRSS/pull/2721, I tried to add a `make lint` and a `make test` targets but I don't know what could be the best solution. I wanted to run them in Docker since I don't have PHP installed directly on my machine and to ease contributions, but we need to install PHPUnit and PHP_CodeSniffer first. I see 4 options:

- since these programs are each contained in a single file, we could add them to the repository and add simple targets like that:
```makefile
.PHONY: test
test: ## Run the test suite
       docker run \
               --rm \
               --volume $(shell pwd):/var/www/FreshRSS:z \
               --env FRESHRSS_ENV=test \
               --name freshrss-test \
               freshrss/freshrss:$(TAG) \
               php ./bin/phpunit --bootstrap ./tests/bootstrap.php ./tests
```

I don't like this solution because admins would download those files while it's only necessary for developers. Also, Alpine image doesn't work because it misses Phar extension. And it's a lot of new lines not related to FRSS!

- second option is to create a dedicated Docker image containing the programs used in test mode, maybe lighter and based on a PHP image instead of a FreshRSS one
- introduce Composer so we add it to the FRSS images and we can use it to install PHPUnit and PHP_CodeSniffer
- last option is to keep things as they are and let people install the programs by themselves. It's a bit unfriendly for newcomers

I see no "perfect" solutions and each has its drawbacks. Maybe someone can enlight my thougts? :)

Frenzie:
> second option is to create a dedicated Docker image containing the programs used in test mode, maybe lighter and based on a PHP image instead of a FreshRSS one

I'm mostly in favor of that one if you don't see much added value in testing multiple PHP versions. Admittedly the image is very small and even double the size will still be very small, but it's just a bunch of cruft for deployment scenarios. You could just take the deployment image and add a dev layer or something.

Besides making development slightly easier, doing so also facilitates migration from Travis to another equivalent CI system.

Alkarex:
Very quick feedback without having thought about it much:
* In Alpine, adding the PHAR extension would not cost that much https://pkgs.alpinelinux.org/package/edge/community/x86_64/php7-phar
* Can't we have a pre-test rule that downloads the needed files if they do not already exist?

marienfressinaud:
> Can't we have a pre-test rule that downloads the needed files if they do not already exist?

[...]

PR #2725, opened 2019-12-22
tec: Add a test target to Makefile

marienfressinaud:
Following of #2722 

It works pretty good but my main issue with this PR is that I don't check the `phpunit` binary. Instructions are given in the documentation, but it doesn't seem to be so easy: https://phpunit.readthedocs.io/en/latest/installation.html#verifying-phpunit-phar-releases

The composer solution inside a Docker image looks more secure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,18 @@ endif
 PORT ?= 8080
 PHP := $(shell sh -c 'which php')
 
+ifdef NO_DOCKER
+	PHP = php
+else
+	PHP = docker run \
+		--rm \
+		--volume $(shell pwd):/var/www/FreshRSS:z \
+		--env FRESHRSS_ENV=development \
+		--name freshrss-php-cli \
+		freshrss/freshrss:$(TAG) \
+		php
+endif
+
 ifeq ($(findstring alpine,$(TAG)),alpine)
 	DOCKERFILE=Dockerfile-Alpine
 else ifeq ($(findstring arm,$(TAG)),arm)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -39,6 +39,24 @@ start: ## Start the development environment (use Docker)
 stop: ## Stop FreshRSS container if any
 	docker stop freshrss-dev
 
+###########
+## Tests ##
+###########
+.PHONY: install-test
+install-test: ## Install required packages for test purposes
+	mkdir -p bin/
+	wget -O bin/phpunit https://phar.phpunit.de/phpunit-7.phar
```
Frenzie: Good point, this way my minor objection doesn't really apply.

Frenzie (approved):

Makefile
marienfressinaud: In the mean time, it's not very secure if we don't verify the binary, I'm a bit annoyed by this solution.

Frenzie: Just stick an md5/sha1 on it?

Alkarex: If we ship a hash, then it must be for a precise release

Frenzie: Naturally, but that's just https://phar.phpunit.de/phpunit-7.5.9.phar with SHA256 `5404288061420c3921e53dd3a756bf044be546c825c5e3556dea4c51aa330f69`. See https://phar.phpunit.de/ for the full list.

marienfressinaud: Done, good idea :+1:

The file is removed if the checksum fails. I also added a bit of documentation.

Frenzie (approved):
Lgtm

Makefile
```diff
@@ -39,6 +39,24 @@ start: ## Start the development environment (use Docker)
...
+	mkdir -p bin/
+	wget -O bin/phpunit https://phar.phpunit.de/phpunit-7.5.9.phar
+	echo '5404288061420c3921e53dd3a756bf044be546c825c5e3556dea4c51aa330f69 bin/phpunit' | sha256sum -c - || rm bin/phpunit
+
+.PHONY: test
+test: ## Run the test suite
+	docker run \
+		--rm \
+		--volume $(shell pwd):/var/www/FreshRSS:z \
+		--env FRESHRSS_ENV=test \
```
Alkarex:
```suggestion
		--env FRESHRSS_ENV=development \
```

@marienfressinaud Was that on purpose? I cannot see any matching rule later in the code

marienfressinaud: It was on purpose, but you're right, there's no matching rule at the moment so let's stick to `development`

marienfressinaud:
I've updated the PR so `bin/phpunit` is downloaded when running `make test` and to make Docker optional when running the test suite.

Makefile
```diff
@@ -39,6 +51,18 @@ start: ## Start the development environment (use Docker)
 stop: ## Stop FreshRSS container if any
 	docker stop freshrss-dev
 
+###########
+## Tests ##
+###########
+bin/phpunit: ## Install PHPUnit for test purposes
```
aledeg, line 57: @marienfressinaud Here you assumed that the user use a Linux machine. I am ok with that but it could be nice to have some validation mecanism to notify the user.

aledeg: You can also run that in the docker container if it exists

PR #2727, opened 2019-12-22
Fix labels dropdown not scrollable after a while

romibi:
At least with Firefox the tags/labels dropdown can't be scrolled after some seconds.

The culprit was the close button that was stretched to the whole screen and firefox does not put child elements behind parent elements even if z-index is lower.

This commit should fix this issue by using a different close button for the scrollable dropdown.

Wasn't sure if the original close button is needed for screen-readers or similar situations therefore i just hid that one.


Alkarex (approved):

Alkarex:
Well spotted and thanks for the fix @romibi 👍

PR #2730, opened 2019-12-23
Update zh-cn (i18n)

Felix2yu:
php cli/check.translation.php -d -l=zh-cn -r
Language: zh-cn

zh-cn - Translation is 100.0% complete.


app/i18n/zh-cn/install.php
```diff
@@ -33,89 +33,89 @@
...
 		'xml' => array(
-			'nok' => '找不到用于 XML 解析库。',
-			'ok' => '已找到 XML 解析库。',
+			'nok' => '找不到用于 XML 解析库',
+			'ok' => '已找到 XML 解析库',
 		),
 		'_' => '检查',
 	),
 	'conf' => array(
-		'ok' => '常规配置已保存。',
```
Frenzie, line 101: Just curious, is that `。` just a regular `.` (from our Latin perspective)?

Frenzie (approved):
Don't read Chinese, but looks fine. ;-)

app/i18n/zh-cn/install.php
9uanhuo: Yes, that mean the end of a sentence. So it is used just like the ' . ' in Latin in most time.

Alkarex:
Thanks @Felix2yu !
Could you please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md ?

PR #2735, opened 2019-12-28
Fix PostgreSQL install

### LINKED ISSUE #2732: Installing with postgresql ignores database name
I tried to install with postgresql username 'nginx' and database name 'freshrss' (through the web interface), but it wouldn't work: the error given in the web interface was unhelpful, implying that the user nginx was denied access, but looking at the postgres logs I got `nginx@nginx FATAL:  database "nginx" does not exist`. Trying it all again with the username and database name matching worked. Guessing there's a mistake in the code where the username is used for the database name and the database name entered is dropped?

(Apologies for the poorly documentation — I only realised I should tell you about this after I had finished the installation)

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2732
When username is different from database name

### DIFF
--- a/lib/lib_install.php
+++ b/lib/lib_install.php
@@ -87,16 +87,19 @@ function initDb() {
 	$db['pdo_options'][PDO::ATTR_ERRMODE] = PDO::ERRMODE_EXCEPTION;
 	$conf->db = $db;	//TODO: Remove this Minz limitation "Indirect modification of overloaded property"
 
+	//Attempt to auto-create database if it does not already exist
 	if ($db['type'] !== 'sqlite') {
 		Minz_ModelPdo::$usesSharedPdo = false;
 		$dbBase = isset($db['base']) ? $db['base'] : '';
-		//Use default database for PostgreSQL, empty database for MySQL / MariaDB:
+		//For first connection, use default database for PostgreSQL, empty database for MySQL / MariaDB:
 		$db['base'] = $db['type'] === 'pgsql' ? 'postgres' : '';
 		$conf->db = $db;
 		//First connection without database name to create the database
 		$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
+		//Restore final database parameters for auto-creation and for future connections
 		$db['base'] = $dbBase;
 		$conf->db = $db;
+		//Perfom database auto-creation
 		$databaseDAO->create();
 	}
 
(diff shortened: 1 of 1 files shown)


Alkarex:
@PetraOleum

lib/lib_install.php
```diff
@@ -90,7 +90,8 @@ function initDb() {
 	if ($db['type'] !== 'sqlite') {
 		Minz_ModelPdo::$usesSharedPdo = false;
 		$dbBase = isset($db['base']) ? $db['base'] : '';
-		$db['base'] = '';
+		//Use default database for PostgreSQL, empty database for MySQL / MariaDB:
+		$db['base'] = $db['type'] === 'pgsql' ? 'postgres' : '';
```
marienfressinaud, line 95: This will have no effect: `$db['base']` is set line 97 with the value from line 92. Also, `initDb` is called after checking the `base` parameter is not empty ([here](https://github.com/FreshRSS/FreshRSS/blob/master/app/install.php#L117)) so I don't understand how the value can be empty.

I think we should be explicit instead of setting the value in the back of the user: what do you think of prefilling the `base` input with value `postgres` if the selected database is pgsql?

Alkarex: This does have an effect on the targetted line, which is `$databaseDAO = FreshRSS_Factory::createDatabaseDAO();`
It is not in the back of the user: using `postgres` as the database name is the recommended (as far as I know) approach to create a new PostgreSQL database.
I will add more comments to clarify this section.
What we *could* do, is to add an option to allow users to autocreate the database or not. But I am not sure it is a good idea, as it will add confusion and noise in the user interface, and I better like testing whether the database exists or not (what we already do).
But what we *should* do, is to test what happens with different access rights, with the target database already created or not, as I am not sure we are catching all errors properly.

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/2735/commits/ac5dcc98a7f300e799f77a55f024efcaac5c1ac7

Alkarex: Ah, I think I now understand your suggestion. No, we should not use the default `postgres` database for our own data.

marienfressinaud: Ah, I see. Maybe a better solution would be to not use the `base` argument from the configuration when creating the database instead of switching the value here

Alkarex: This could probably be refactored, but maybe that this a job for another PR :-)

marienfressinaud: Of course ;) and for another time too!

Alkarex: Any remaining comment then? :-)

marienfressinaud: Nope! (I don't know why, I thought it was already merged :))

marienfressinaud (approved):

PR #2745, opened 2019-12-30
Fix logging environment constants

Alkarex:
* COPY_LOG_TO_SYSLOG was only working when Syslog was used
* FRESHRSS_ENV was not properly used to set logging level

### DIFF
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -63,7 +63,7 @@ class Minz_Log {
 			     . ' --- ' . $information . "\n";
 
 			if (defined('COPY_LOG_TO_SYSLOG') && COPY_LOG_TO_SYSLOG) {
-				syslog($level, '[' . $username . '] ' . $log);
+				syslog($level, '[' . $username . '] ' . trim($log));
 			}
 
 			self::ensureMaxLogSize($file_name);
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -33,11 +33,11 @@ class Minz_Log {
 		if (! ($env === 'silent'
 		       || ($env === 'production'
 		       && ($level >= LOG_NOTICE)))) {
+			$username = Minz_Session::param('currentUser', '');
+			if ($username == '') {
+				$username = '_';
+			}
 			if ($file_name === null) {
-				$username = Minz_Session::param('currentUser', '');
-				if ($username == '') {
-					$username = '_';
-				}
 				$file_name = join_path(USERS_PATH, $username, 'log.txt');
 			}
 
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -20,11 +20,14 @@ class Minz_Log {
 	 * @throws Minz_PermissionDeniedException
 	 */
 	public static function record ($information, $level, $file_name = null) {
-		try {
-			$conf = Minz_Configuration::get('system');
-			$env = $conf->environment;
-		} catch (Minz_ConfigurationException $e) {
-			$env = 'production';
+		$env = getenv('FRESHRSS_ENV');
+		if ($env == '') {
+			try {
+				$conf = Minz_Configuration::get('system');
+				$env = $conf->environment;
+			} catch (Minz_ConfigurationException $e) {
+				$env = 'production';
+			}
 		}
 
 		if (! ($env === 'silent'
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -189,7 +189,14 @@ function html_only_entity_decode($text) {
 }
 
 function prepareSyslog() {
-	return COPY_SYSLOG_TO_STDERR ? openlog("FreshRSS", LOG_PERROR | LOG_PID, LOG_USER) : false;
+	if (defined('SYSLOG_PREPARED')) {
+		return null;
+	}
+	define('SYSLOG_PREPARED', true);
+	return COPY_SYSLOG_TO_STDERR ? openlog('FreshRSS', LOG_PERROR | LOG_PID, LOG_USER) : false;
+}
+if (COPY_LOG_TO_SYSLOG) {
+	prepareSyslog();
 }
 
 function customSimplePie($attributes = array()) {
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -9,6 +9,10 @@ if (!function_exists('mb_strcut')) {
 	}
 }
 
+if (COPY_SYSLOG_TO_STDERR) {
+	openlog('FreshRSS', LOG_PERROR | LOG_PID, LOG_USER);
+}
+
 /**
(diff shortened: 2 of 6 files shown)


lib/Minz/Log.php
```diff
@@ -20,11 +20,14 @@ class Minz_Log {
 	 * @throws Minz_PermissionDeniedException
 	 */
 	public static function record ($information, $level, $file_name = null) {
-		try {
-			$conf = Minz_Configuration::get('system');
-			$env = $conf->environment;
-		} catch (Minz_ConfigurationException $e) {
-			$env = 'production';
+		$env = getenv('FRESHRSS_ENV');
```
Alkarex, line 23: @marienfressinaud You might want to refactor that at some point, to avoid FreshRSS logic in Minz. And if I remember correctly, your said you would re-think a bit the Minz configuration logic, so I leave that to later :-)

Alkarex: Related to:

https://github.com/FreshRSS/FreshRSS/blob/6cedeeeae51f5e5220d7f786fb801951811bb743/lib/Minz/FrontController.php#L118-L122

marienfressinaud: Indeed, and I've started [something in Webubbub](https://github.com/flusio/Webubbub/blob/master/lib/Minz/src/Configuration.php). I plan to import/adapt some of the code I wrote for this project to FRSS (it's a kind of v2 for Minz, it allowed me to test other approaches :))

lib/lib_rss.php
```diff
@@ -189,7 +189,14 @@ function html_only_entity_decode($text) {
 }
 
 function prepareSyslog() {
-	return COPY_SYSLOG_TO_STDERR ? openlog("FreshRSS", LOG_PERROR | LOG_PID, LOG_USER) : false;
+	if (defined('SYSLOG_PREPARED')) {
+		return null;
+	}
+	define('SYSLOG_PREPARED', true);
+	return COPY_SYSLOG_TO_STDERR ? openlog('FreshRSS', LOG_PERROR | LOG_PID, LOG_USER) : false;
```
marienfressinaud: I'm not sure what's the purpose of this environment var, shouldn't we use `COPY_LOG_TO_SYSLOG` instead? (and drop this one)

lib/lib_rss.php
```diff
@@ -189,7 +189,14 @@ function html_only_entity_decode($text) {
 }
 
 function prepareSyslog() {
-	return COPY_SYSLOG_TO_STDERR ? openlog("FreshRSS", LOG_PERROR | LOG_PID, LOG_USER) : false;
+	if (defined('SYSLOG_PREPARED')) {
+		return null;
+	}
+	define('SYSLOG_PREPARED', true);
```
marienfressinaud: Why do you use a constant? I think we should remove other calls to `prepareSyslog` and keep only the one you added in this PR (not sure if it's the best place for that btw, but it's nitpicking). This way we can remove the use of `SYSLOG_PREPARED`

lib/Minz/Log.php
Alkarex: Please try to keep it light and refrain from premature abstractions ;-)

lib/lib_rss.php
Alkarex: It is for different cases. FreshRSS has normal text logs, and syslogs (e.g. for SimplePie, favicons...).

* `COPY_LOG_TO_SYSLOG`: Depending on whether you are monitoring FreshRSS text logs or not, this might be relevant, or in the contrary create annoying duplicates.

* `COPY_SYSLOG_TO_STDERR`: In a Docker deployment for instance, copying to stderr is convenient so that those lines are visible in Docker logs, but for other systems with a native syslog (like when FreshRSS is installed natively), then we should not do it.

In our Docker image (where there is no native syslog), we use both, but any combination can be relevant in my opinion.

lib/lib_rss.php
Alkarex: Yes indeed. It was a performance consideration to avoid the cost of `openlog()` when it is not used (i.e. most of the time).

lib/lib_rss.php
marienfressinaud: Ok, right. But we should at least call `openlog` with `COPY_LOG_TO_SYSLOG`: on my server there's a strange behavior where FRSS logs are identified as another app that called `openlog` before.

lib/lib_rss.php
Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/2745/commits/4b93ced3039031dc0d4c7c74118970accca2b15e

Alkarex: P.S. Not that bad thanks to
> 

LOG_ODELAY | (default) delay opening the connection until the first             message is logged
-- | --

lib/lib_rss.php
Alkarex: Ah, really? Can you explain a bit more the situation?

Alkarex: Ok, should be better with https://github.com/FreshRSS/FreshRSS/pull/2745/commits/48ab4ee05fe66989ff5c5d0ad878db4319d5bcb1

marienfressinaud: Well, I had `COPY_LOG_TO_SYSLOG=On` and `COPY_SYSLOG_TO_STDERR=Off`, so the `openlog` wasn't called in FRSS. And for some reasons, the `syslog` calls were considering the `ident` string set by another application, so my logs were all mixed up :upside_down_face:

Alkarex: Ok, strange, but is it better now? :-)

marienfressinaud: It looks good, yes :)

lib/Minz/Log.php
```diff
@@ -60,7 +63,7 @@ public static function record ($information, $level, $file_name = null) {
 			     . ' --- ' . $information . "\n";
 
 			if (defined('COPY_LOG_TO_SYSLOG') && COPY_LOG_TO_SYSLOG) {
-				syslog($level, '[' . $username . '] ' . $log);
+				syslog($level, '[' . $username . '] ' . trim($log));
```
marienfressinaud, line 66: Oh right! I saw these empty lines but didn't understand the reason

lib/Minz/Log.php
```diff
@@ -20,21 +20,24 @@ class Minz_Log {
...
 		       && ($level >= LOG_NOTICE)))) {
+			$username = Minz_Session::param('currentUser', '');
+			if ($username == '') {
+				$username = '_';
+			}
 			if ($file_name === null) {
-				$username = Minz_Session::param('currentUser', '');
-				if ($username == '') {
-					$username = '_';
-				}
```
marienfressinaud, line 37: I was going to open a PR about this one :D

marienfressinaud (approved):

Issue #2748, opened 2019-12-31
Unable to Configure FreshRSS


ProbablyBread:
Weird issue, can't find anyone who has encountered this issue (or maybe my GoogleFu is just weak). 

So I've successfully gotten here:
![image](https://user-images.githubusercontent.com/50959479/71614176-2ebf1000-2be5-11ea-8617-91334d49cae1.png)

After filling in the options and clicking Submit, the fields just get reset and it doesn't move on to the next page. There's no error message in both the Web UI and inside /var/log/nginx/error.log either.

But checking .../data/config.php, the fields for the database does get populated: 
![Untitled](https://user-images.githubusercontent.com/50959479/71614271-a42ae080-2be5-11ea-884c-3b866532e6ae.png)

Any help would be appreciated!

Frenzie:
Duplicate of #2732?

ProbablyBread:
Dropped the freshrss DB, created a new freshrss DB along with a new user of the same name. No luck. 

![Untitled](https://user-images.githubusercontent.com/50959479/71614959-9c6d3b00-2be9-11ea-8ab7-7cb642f8e840.png)

Alkarex:
PostgreSQL is picky about localhost. Try 127.0.0.1 and make sure you have the latest FreshRSS version from our Master branch in which a PostgreSQL bug has just been fixed

Alkarex:
Ups, and I can see an unrelated bug in base_url

ProbablyBread:
I already have the latest version. 

Same issue after changing to 127.0.0.1 as well. Also tried to use 127.0.0.1:5432, but no luck with that either. 

For the record, I'm using Nginx and PHP 7.3 for this.

Alkarex:
And if you drop the DB without manually creating it? (while giving the freshrss user rights to create a database)

ProbablyBread:
Fixed it. I was trying to use nginx reverse proxy to get into freshrss.

Saw this issue (#1958  ) regarding nginx reverse proxy and tried accessing directly into http://URL/freshrss instead of http://URL/proxy-rss and it worked immediately.

Got the reverse proxy working with this as well, not sure if this is good practice: 
https://www.reddit.com/r/selfhosted/comments/dw83ew/freshrss_apache_reverse_proxy/

Thanks for your assistance!

Alkarex:
[...]

PR #2749, opened 2019-12-31
Fix wrong getHeader refactoring

Alkarex:
Fix regression introduced by https://github.com/FreshRSS/FreshRSS/pull/2373
The refactoring required a call to init() even for static functions, which is most of the time not done.
Removed premature abstraction of `$_SERVER`, which was the root cause of the bug.
https://github.com/FreshRSS/FreshRSS/issues/2748#issuecomment-569898931

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -11,7 +11,6 @@ class Minz_Request {
 	private static $controller_name = '';
 	private static $action_name = '';
 	private static $params = array();
-	private static $headers = array();
 
 	private static $default_controller_name = 'index';
 	private static $default_action_name = 'index';
@@ -101,7 +100,6 @@ class Minz_Request {
 	 * Initialise la Request
 	 */
 	public static function init() {
-		static::$headers = $_SERVER;
 		self::initJSON();
 	}
 
@@ -348,10 +346,7 @@ class Minz_Request {
 	 * @return mixed
 	 */
 	public static function getHeader($header, $default = null) {
-		if (isset(static::$headers[$header])) {
-			return static::$headers[$header];
-		}
-		return $default;
+		return isset($_SERVER[$header]) ? $_SERVER[$header] : $default;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


lib/Minz/Request.php
```diff
@@ -227,6 +225,42 @@ public static function getBaseUrl() {
...
+	 *
+	 * Note: for the moment it tests only if address is corresponding to a
+	 * localhost address.
+	 *
+	 * @param $address the address to test, can be an IP or a URL.
+	 * @return true if server is accessible, false otherwise.
+	 * @todo improve test with a more valid technique (e.g. test with an external server?)
+	 */
+	public static function serverIsPublic($address) {
+		if (strlen($address) < strlen('http://a.bc')) {
```
Frenzie, line 239: We're sure it will always includes a protocol at this point?

Alkarex: If it does not, then it is not a valid public address

aledeg (commented):
I am not a big fan of the method used to check the url. Can't you use `filter_var` with a built-in filter instead?

- [PHP: Validate filters - Manual](https://www.php.net/manual/en/filter.filters.validate.php)
- [PHP: filter_var - Manual](https://www.php.net/manual/en/function.filter-var.php)

Alkarex:
@aledeg Which line(s) more precisely?

aledeg:
This part
```
if (strlen($address) < strlen('http://a.bc')
```

and also the one with the `preg_match`

I am not sure if it's doable though

Alkarex:
Sorry for the delay. I will try to look at it tomorrow, but I cannot immediately think of a better way.
The one with `strlen($address)` is just to relax a previous `== ''` now that we might have an empty string or something like `http:`

Alkarex:
@aledeg I do agree that we should improve the way we guess and/or test whether an URL is public/routable. Suggestions welcome. However, I am not sure how to do that better at the moment, and this PR is urgent, as all current installations based on /master are broken

Alkarex:
Sorry to be a bit pushy, but this is quite an urgent bug. So let me merge and let's continue improvements in another PR if needed

PR #2760, opened 2020-01-11
Add a migration system

marienfressinaud:

- Add a Minz_Migrator class to help developers to migrate database or application structure automatically
- Apply migrations when users hit index

**For now, I'd like some feedback on the general architecture. What do you think of this system?**

I mentioned this system [few weeks ago](https://github.com/FreshRSS/FreshRSS/pull/2709#discussion_r360638791).

I tried to keep it simple, based on the approach of Ruby on Rails (not as powerful though) but maybe you'll find it too complicated, let me know :) ~~The main drawback of this system is the migration files need to be chronologically ordered by filename. We might have some surprises if we merge two PRs adding migrations in the wrong order (I can detail how it could go wrong if you want to), but it would impact only people on rolling release and we can add a check on the CI.~~ This is now partially addressed but I still suggest to order filenames chronologically.

In the version 1.17, I plan to drop the `data/do-install.txt` file to only rely on the existence of `data/applied_migrations.txt`. It will make the deployment safer because the install page will no longer risk to appear in front of end users (and so some attackers) during an update.

1. Go to FreshRSS (index)
2. Check there's now an empty `data/applied_migrations.txt` file
3. You can also create a temporary migration file (examples given in the PR files) under `app/migrations` (e.g. `cp tests/fixtures/migrations/2019_12_22_FooBar.php app/migrations/`)
4. Go to FreshRSS (index)
5. Check the `data/applied_migrations.txt` file now contains the last migrations version
6. After that, make sure to drop both the migration AND the `data/applied_migrations.txt` files

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


marienfressinaud:
@Alkarex @aledeg @Frenzie do you have any quick thought on this system? I still want to complete the documentation a bit so a review is not required yet, but I'd like to have your opinion first :) It would be nice to merge it for the next version so I can finish https://github.com/FreshRSS/FreshRSS/pull/1760 then

Frenzie:
At a glance the general principle looks fine and clearer to me.

Alkarex:
It looks fine at a first glance. One comment is that until now, I have tried to always make the updates developer/tester friendly, meaning that someone could test and use PRs in a random order, without relying on a global version number for e.g. database updates. In other words, feature detection (based on errors, to avoid any performance penalty) instead of version detection. It needs not necessarily have to stay this way, but I have liked the approach so far.

marienfressinaud:
Ok, I agree with you. I can list all the applied migrations in the `data/migrations_version.txt` file (probably renamed into `data/applied_migrations.txt`), and only the missing ones will be applied. With this solution, we should let developers to declare dependencies between migrations (if a migration create a table, and another one add a column to this table, the later must always be executed in second). The consequence of this is we'll have to be a bit more precocious when writing migrations, but I'm personally fine with it.

Sounds good to you?

aledeg:
Sorry, I don't have much time to review it :'(

marienfressinaud:
This PR is now ready to be reviewed. I partially addressed [@Alkarex comment](https://github.com/FreshRSS/FreshRSS/pull/2760#issuecomment-575887490) by keeping a trace of all the applied versions in a `data/applied_migrations.txt` file. If a migration is present in this file, it is skipped.

I did NOT implemented the dependency system ([mentioned here](https://github.com/FreshRSS/FreshRSS/pull/2760#issuecomment-575890867)) and kept the recommended chronological system (i.e. migrations are sorted by filenames). It means if a PR depends on another PR, developer must make sure it will be named correctly. By prepending the filename by the current datetime, it actually should always be the case because a migration relying on another one will logically always be created after the later. A different dependency system would make the code harder to understand and the migrations order more difficult to understand.

Last note: I'd love to have it in the next FRSS version so I can remove the `data/do-install.txt` file after.

Frenzie (commented):
Could the number of migrations in the file become a concern?

docs/en/developers/Minz/migrations.md
```diff
@@ -0,0 +1,39 @@
+# How to manage migrations with Minz
+
+Migrations are the way to modify the database or the structure of files under the `data/` path.
+
+## How to write a migration?
+
+Migrations are placed under the `app/migrations` folder.
+
+Good practice is to prepend the filename by the current date and time and explain what does the migration do in few words (e.g. `20200111_163020_CreateFooTable.php`).
```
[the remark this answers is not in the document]
marienfressinaud: I'm not sure if your question is about having both date and time, or about the word ^^

marienfressinaud:
> Could the number of migrations in the file become a concern?

It might on the long term if the number of migrations becomes high, but we'll have several options:

- disable the automatic updates and apply them via the CLI
- remove the oldest migrations

In my opinion, we'll have time before it becomes a real problem.

docs/en/developers/Minz/migrations.md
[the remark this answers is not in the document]
Frenzie: About having both. Seems like it's mainly just harder to read. Especially with those seconds.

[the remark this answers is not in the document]
marienfressinaud: Maybe it's too much indeed, especially if we consider how many migrations we wrote these last years ^^ I'll remove the time and add underscores between year, month and day

Alkarex:
Sorry for the delay. At a quick glance, it looks fine. What is the status?

lib/Minz/Migrator.php
```diff
@@ -0,0 +1,281 @@
...
+	public static function execute($migrations_path, $applied_migrations_path) {
+		if (!file_exists($applied_migrations_path)) {
+			return "{$applied_migrations_path} file does not exist";
+		}
+
+		$applied_migrations_file = fopen($applied_migrations_path, 'r+');
+
+		// we need to acquire a lock to avoid multiple migrations at the same
+		// time if several users try to access the application.
+		if (!flock($applied_migrations_file, LOCK_EX)) {
```
Alkarex: I am a bit worried about this part. I have not had the energy to test on my side, but we should ensure that it works (at least in theory) with different Web servers, OS, mounts (e.g. network drives...), especially when used at such a critical place.
Acquiring a lock is also expensive and error prone (at least last time I checked), and this worries me for scaling (so a two-step approach should be considered: testing a cheaper condition such as file existence, and then go for the lock).
We should also release the lock fast, even when the request is aborted (with the current code, I have not checked but we might end up in an indeterminate state - see e.g. https://php.net/function.ignore-user-abort ).
Finally, it would be good to have a fail safe approach, i.e. that we anyway try to run FreshRSS instead of returning an error message.
Many years ago, I did a few experiments with different methods, still visible in https://alexandre.alapetite.fr/doc-alex/compteur/ and it was not trivial to get right.

Alkarex: P.S. A database request can also be considered

marienfressinaud: You're right, this is not trivial and plus, I didn't think about a potential user abort. Retrospectively I wonder if it shouldn't be better to force the admin to apply the migration himself (via the CLI, or during the automatic update) so we can simply get rid of this lock-approach. What do you think?

>  Finally, it would be good to have a fail safe approach, i.e. that we anyway try to run FreshRSS instead of returning an error message.

I would rather show a maintenance page to normal users, and explain to administrators how to apply the update. I don't like the idea of bypassing an error because it's often the cause of more errors. If the update doesn't apply, it's almost certain something else will not work then, and FRSS will look broken.

marienfressinaud: I changed the system in https://github.com/FreshRSS/FreshRSS/pull/2760/commits/8e05d9115cac334790b0d1f133785a15879e6bfc but it brings different problems. I'm afraid that there's no perfect system if we want to let the migrations being applied on user requests.

The more robust would be to apply them on an update and show a maintenance page while there are not.

Alkarex: I do not have much better solution to propose right now, and as long as it does not introduce a potential bottleneck or issue with some platforms, I think that would be fine.

marienfressinaud:
It should be good now!

docs/en/developers/Minz/migrations.md
```diff
@@ -0,0 +1,39 @@
...
+		return true;
+	}
+}
+```
+
+## How to apply migrations?
+
+They are automatically applied one by one when a user accesses FreshRSS.
+
+Before being applied, migrations are sorted by filenames (see the [`strnatcmp`](https://www.php.net/manual/en/function.strnatcmp.php) function). Already applied migrations are skipped (the list can be found in the `data/applied_migrations.txt` file).
```
Alkarex:
```suggestion
Before being applied, migrations are sorted by filenames (see the [`strnatcmp`](https://php.net/strnatcmp) function). Already applied migrations are skipped (the list can be found in the `data/applied_migrations.txt` file).
```

Very minor: shorter link, which will also do language negotiation instead of hardcoded English

lib/Minz/Migrator.php
```diff
@@ -0,0 +1,279 @@
...
+	 *
+	 * @return boolean|string Returns true if execute succeeds to apply
+	 *                        migrations, or a string if it fails.
+	 */
+	public static function execute($migrations_path, $applied_migrations_path) {
+		if (!file_exists($applied_migrations_path)) {
+			return "{$applied_migrations_path} file does not exist";
+		}
+
+		$applied_migrations = file_get_contents($applied_migrations_path);
```
Alkarex:
```suggestion
		$applied_migrations = @file_get_contents($applied_migrations_path);
```

A bit less IO

data/.gitignore
```diff
@@ -8,3 +8,4 @@ no-cache.txt
 update.php
 tos.html
 opml.xml
+applied_migrations.txt
```
Alkarex, line 11:
```suggestion
applied_migrations.txt
applied_migrations.txt.lock
```

Alkarex: I cannot remember if gitignore would also match that one, but at least this way it is explicit

lib/Minz/Migrator.php
```diff
@@ -0,0 +1,279 @@
...
+			if ($filename[0] === '.') {
+				continue;
+			}
+
+			$filepath = $directory . '/' . $filename;
+			$migration_version = basename($filename, '.php');
+			$migration_class = APP_NAME . "_Migration_" . $migration_version;
+			$migration_callback = $migration_class . '::migrate';
+
+			$include_result = @include_once($filepath);
```
Alkarex, line 147: It looks like we run this `foreach ... include` at each request, also when we do not need a migration?

Alkarex: It would be great to be able to return faster the information about the need to migrate or not.
Just a few quick ideas:
* Compare version in `constants.php` with a file `data/version.txt` and run through the migration code if they are not equal, then update `data/version.txt`
* Or provide a text file `available_migrations.txt` with the list of available migrations, to avoid having to scan a directory and include irrelevant code;
  * Improvement option: If timestamp of `applied_migrations.txt` is 0 (i.e. does not exist) or is older than `available_migrations.txt`, proceed.

What do you think @marienfressinaud ?

p/i/index.php
```diff
@@ -42,14 +42,43 @@
...
+	// if it is missing, it means the application is not installed. But we
+	// should also take care of applications installed before the new
+	// migrations system (<1.16). Indeed, they are installed but the migrations
+	// version file doesn't exist. So for now, we continue to check if the
+	// application is installed with the do-install.txt file: if yes, we create
+	// the version file. Starting from version 1.17, all the installed systems
+	// will have the file and so we will be able to remove this temporary line
+	// and stop using the do-install.txt file to check if FRSS is already
+	// installed.
+	touch($applied_migrations_path);
```
Alkarex, line 59: We might not need that. If the file does not exist, then we try to apply the migrations and finally create the file

Alkarex:
Besides the fact that I would like a cheaper / faster way to return the info about whether we need to migrate anything or not (see my suggestions - I believe we should be able to find something), this is nice 👍

p/i/index.php
marienfressinaud: Yep, with your suggestion above, it might make sense :+1: I just need to take a bit of time to reconsider this first (I expect to work on it this week), this line had a high importance in my mind ^^

lib/Minz/Migrator.php
marienfressinaud: I'll work on it this afternoon: the easiest solution that would require the less of work is to move this `include_once` call to the `migrate` function

marienfressinaud: The easiest solution wasn't easy finally: it would make the whole class and tests more complicated.

I chose another solution: the versions names are extracted from the migration filenames, so I simply compare these names with the versions in the `applied_migrations.txt`.

p/i/index.php
marienfressinaud: So indeed, this line has its importance. Well it might not immediately, but I prefer to keep the current design and to make the Migrator failing if the file doesn't exist since the install process will have to start if `applied_migrations.txt` doesn't exist in a later version of FRSS.

marienfressinaud (commented):
I took the last feedbacks in account, I think I'm done here. Sorry I rebased my branch mechanically

Alkarex (approved):

Alkarex:
I would like to issue soon version 1.16.1 to fix some regressions and a few other bugs. What do you prefer @marienfressinaud , should we merge this one now or wait for next release?

Alkarex:
A quick test on my Raspberry Pi 4 (with SSD disk) shows that in average on 170 requests, this costs 0.4ms (but the migration folder is empty at the moment).

marienfressinaud:
We should wait for the next release 1.16.1 ;)

Alkarex:
> We should wait for the next release 1.16.1 ;)

@marienfressinaud I have not understood :-) That means we merge now, right? (Which I think is fine). But maybe you should have a final look at the version numbers included in the PHP comments and Markdown documentation

marienfressinaud:
Ha ha, sorry I mixed up two messages, I wanted to say "We should wait for the next release" (without the "1.16.1"). In other words: we should wait :p

marienfressinaud:
I'll merge this one after #3092 and maybe #3081 ;)

PR #2763, opened 2020-01-12
Pass the entries to 'entry_before_display' extension hook

### LINKED ISSUE #2762: The greater API doesn't pass content by 'entry_before_display' extension hook.
When we access RSS articles from greater API, then the entries are not passed by the 'entry_before_display' extension hook, so we miss some potential interesting changes done by some extensions.

Is it on purpose ? By example, it can have a great interest for YouTube feeds & YouTube extension, to have in-line video playing.

Perhaps it can be an option for extensions to tell where it need to be enabled ? (WebUI, greader, etc.).

javerous:
Closes #2762

- Pass feed entries processed by greader API to the 'entry_before_display' extension hook.

1. Enable some extension like YouTube, which put YouTube video in-line.
2. Add a YouTube feed to your FreshRSS instance.
3. Check that you see the in-line YouTube video in-line in your RSS client.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Note: I'm not a PHP developer. Feel free to reject if you don't want this change, or if you want to elaborate a more complete solution.

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -920,10 +920,9 @@ $user = authorizationToUser();
 FreshRSS_Context::$user_conf = null;
 if ($user !== '') {
 	FreshRSS_Context::$user_conf = get_user_configuration($user);
-	Minz_Translate::init(FreshRSS_Context::$user_conf->language);
-
 	Minz_ExtensionManager::init();
-
+	Minz_Translate::init(FreshRSS_Context::$user_conf->language);
+	
 	if (FreshRSS_Context::$user_conf != null) {
 		Minz_ExtensionManager::enableByList(FreshRSS_Context::$user_conf->extensions_enabled);
 	}
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -922,7 +922,7 @@ if ($user !== '') {
 	FreshRSS_Context::$user_conf = get_user_configuration($user);
 	Minz_ExtensionManager::init();
 	Minz_Translate::init(FreshRSS_Context::$user_conf->language);
-	
+
 	if (FreshRSS_Context::$user_conf != null) {
 		Minz_ExtensionManager::enableByList(FreshRSS_Context::$user_conf->extensions_enabled);
 	}
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -481,7 +481,11 @@ function entriesToArray($entries) {
 	}
 
 	$items = array();
-	foreach ($entries as $entry) {
+	foreach ($entries as $item) {
+		$entry = Minz_ExtensionManager::callHook('entry_before_display', $item);
+		if ($entry == null) {
+			continue;
+		}
 		$f_id = $entry->feed();
 		if (isset($arrayFeedCategoryNames[$f_id])) {
 			$c_name = $arrayFeedCategoryNames[$f_id]['c_name'];
@@ -917,6 +921,12 @@ FreshRSS_Context::$user_conf = null;
 if ($user !== '') {
 	FreshRSS_Context::$user_conf = get_user_configuration($user);
 	Minz_Translate::init(FreshRSS_Context::$user_conf->language);
+
+	Minz_ExtensionManager::init();
+
+	if (FreshRSS_Context::$user_conf != null) {
+		Minz_ExtensionManager::enableByList(FreshRSS_Context::$user_conf->extensions_enabled);
+	}
 } else {
 	Minz_Translate::init();
 }


p/api/greader.php
```diff
@@ -917,6 +921,12 @@ function markAllAsRead($streamId, $olderThanId) {
 if ($user !== '') {
 	FreshRSS_Context::$user_conf = get_user_configuration($user);
 	Minz_Translate::init(FreshRSS_Context::$user_conf->language);
+
+	Minz_ExtensionManager::init();
```
Frenzie, line 923: Unless it makes a functional difference I think I'd just stick this on the line where it goes alphabetically (and in any case without newlines).

Some other part of the code seems to agree with me:
https://github.com/FreshRSS/FreshRSS/blob/03ddcf0772f77f8e67edcc43a71cb8d01c18a3ff/p/api/pshb.php#L119-L120

p/api/greader.php
```diff
@@ -917,6 +921,12 @@ function markAllAsRead($streamId, $olderThanId) {
...
+
+	if (FreshRSS_Context::$user_conf != null) {
+		Minz_ExtensionManager::enableByList(FreshRSS_Context::$user_conf->extensions_enabled);
+	}
 } else {
```
Frenzie, line 929: Would `$system_conf->allow_anonymous)` be relevant to the GReader API?

Frenzie (commented):
Makes sense to me.

p/api/greader.php
javerous: Like this ? (sorry, I don't have the same code-style, not even to say the same dev language, so it's not easy to adapt myself ^^)

```
	FreshRSS_Context::$user_conf = get_user_configuration($user);
	Minz_ExtensionManager::init();
	Minz_Translate::init(FreshRSS_Context::$user_conf->language);
	if (FreshRSS_Context::$user_conf != null) {
		Minz_ExtensionManager::enableByList(FreshRSS_Context::$user_conf->extensions_enabled);
	}
```

p/api/greader.php
javerous: It seems relevant, I guess (even if I wouldn't have the usage, personally), but it's a bit out of context of this PR, no ? (or I didn't understand the point…).

Frenzie: If it's relevant then the extension hook is relevant, along these lines:

https://github.com/FreshRSS/FreshRSS/blob/7802fd33a627dc7c582df871cfa613d9be8f8788/app/FreshRSS.php#L51-L55

Frenzie: Basically I mean that this else path seems to be about users who aren't logged in. I haven't looked into this, mind you, I'm just looking at the diff here.

javerous: Okay, I see. I think it’s a bit out of my understanding of the sources :s

How `FreshRSS_Context::$user_conf` can't be null if there is no `$user` ? We have to find the default user, in this case ? Or `$user_conf` already point to the default user before we set it to NULL (by security, perhaps ?) ?

I think I don't know the code enough (and the consequences) to play around that part…

That being said, apparently the `allow_anonymous` is not handled at all by greader, it just fails if a user is not set & valid, so the question of extension loading, in this condition, will have to be handled when `allow_anonymous` will be handled in a whole for greader, no ?

```
if ($user == '') {
		unauthorized();
	}
```

p/api/greader.php
Frenzie: Personally I'd stick in a newline before the if statement but something more like that, yes. ;-)

p/api/greader.php
Frenzie: That would mean the answer to my question is that it's *not* relevant. ;-)

javerous: > Basically I mean that this else path seems to be about users who aren't logged in. I haven't looked into this, mind you, I'm just looking at the diff here.

You mean
```
} else {
	Minz_Translate::init();
}
```
?

Yes, I'm not sure why it was done, greader doesn't seems to allow anything, currently, if there is no authenticated user.

Frenzie: Yes, that's why I commented on the else line. That wasn't just a random comment, or I wouldn't have attached it to a line. ^_^

javerous: > That would mean the answer to my question is that it's _not_ relevant. ;-)

Ok, yes. I better understand your point now.

It **would** be relevant to load default user extensions **if** `allow_anonymous` was handled in a whole by this API, but as it doesn't seems to be the case, then I guess it's not relevant, at least for now.

p/api/greader.php
```diff
@@ -481,7 +481,11 @@ function entriesToArray($entries) {
 	}
 
 	$items = array();
-	foreach ($entries as $entry) {
+	foreach ($entries as $item) {
+		$entry = Minz_ExtensionManager::callHook('entry_before_display', $item);
```
Frenzie, line 485: More of a note for other people reading this, because at least at present it doesn't seem too important, but in the issue there was the suggestion of differentiating between `entry_before_display` and `entry_before_api` (or something along those lines).

> Perhaps it can be an option for extensions to tell where it need to be enabled ? (WebUI, greader, etc.)

marienfressinaud: I have no objections to distinguish `entry_before_display` and `entry_before_api` (maybe different naming like `entry_before_api_display`?) but I would stick with current solution for now. It makes the PR easier and we'll be able to merge it quicker :)

marienfressinaud (approved):
Thanks for your contributions @javerous, really appreciated! Don't hesitate to add yourself to the [credits file](https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md) ;)

Frenzie (approved):

Alkarex:
We should do the same for the fever API in `fever.php`

javerous:
@Alkarex It's already partially done in Fever:

```
		// Load list of extensions and enable the "system" ones.
		Minz_ExtensionManager::init();

		foreach ($entries as $item) {
			/** @var FreshRSS_Entry $entry */
			$entry = Minz_ExtensionManager::callHook('entry_before_display', $item);
			if ($entry == null) {
				continue;
			}
```

Would just be nice to add user extension loading, if possible. I can take a look, if you want. Apparently, Reeder is compatible with Fever API, so I can test.

Not sure what is the difference between Fever & Greader, by the way (I mean, what are the advantages of each one). Anyway, it's not the subject.

Alkarex:
The Google Reader API is much more feature-rich and efficient than the Fever one

PR #2766, opened 2020-01-13
Show Favorites as Unread

### LINKED ISSUE #2434: [FR] Optionally display allways read and unread articles in Favorites

Offerel:
Closes #2434

- added a setting in "display" to show starred/favorites as unread

1. got to settings > display
2. a new checkbox is available: Show Favorites as "Unread" 
3. enable the checkbox
4. click on favorites/starred > url parameter state=3 is added to the link
5. if you click the favorites > read and unread articles in favorites are displayed

**Pull request checklist:**
- [ x ] code manually tested

**Question:**
Anyway, what does "show as unread" mean, with the unread background color or something?

**Answer:**
Color isn't changed for the articles. All will stay the same. But currently, if you click Favorites/Starred, only unread articles are displayed. In my workflow, when i star/favorite a article, it will be marked as read. Clicking now Favorite/Starred will not show that article, unless i also enable "unread" from the toolbar. This save that step and displays favorites/starred with state=3 as default. If the option is unchecked, the current behavior is used.

_Hopefully, this time i get it right :-) It's a bit frickely to do that only in the WebGUI, but i didn't have my PC at hand with git installed_

### DIFF
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -37,6 +37,7 @@ return array(
 			'timeout' => 'HTML5 notification timeout',
 		),
 		'show_nav_buttons' => 'Show the navigation buttons',
+		'show_fav_unread' => 'Show Favorites as "Unread"',
 		'theme' => 'Theme',
 		'title' => 'Display',
 		'width' => array(
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -97,7 +97,7 @@ return array(
 	),
 	'reading' => array(
 		'after_onread' => 'After “mark all as read”,',
-		'always_show_favorites' => 'Show by default all articles in favorites',
+		'always_show_favorites' => 'Show all articles in favorites by default',
 		'articles_per_page' => 'Number of articles per page',
 		'auto_load_more' => 'Load more articles at the page bottom',
 		'auto_remove_article' => 'Hide articles after reading',
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -37,7 +37,6 @@ return array(
 			'timeout' => 'HTML5 notification timeout',
 		),
 		'show_nav_buttons' => 'Show the navigation buttons',
-		'show_fav_unread' => 'Always show favorites',
 		'theme' => 'Theme',
 		'title' => 'Display',
 		'width' => array(
@@ -98,6 +97,7 @@ return array(
 	),
 	'reading' => array(
 		'after_onread' => 'After “mark all as read”,',
+		'always_show_favorites' => 'Always show favorites',
 		'articles_per_page' => 'Number of articles per page',
 		'auto_load_more' => 'Load more articles at the page bottom',
 		'auto_remove_article' => 'Hide articles after reading',
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -37,7 +37,7 @@ return array(
 			'timeout' => 'HTML5 notification timeout',
 		),
 		'show_nav_buttons' => 'Show the navigation buttons',
-		'show_fav_unread' => 'Show Favorites as "Unread"',
+		'show_fav_unread' => 'Always show favorites',
 		'theme' => 'Theme',
 		'title' => 'Display',
 		'width' => array(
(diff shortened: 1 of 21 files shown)


Frenzie:
The proposed setting looks like this (see at the bottom):
Screenshot_2020-01-13 Display · FreshRSS

I think it should go in reading instead. Both logically and because the "Articles to display" setting is already right there. This is a refinement of that.

I'd name the setting itself something like `Always show favorites`.

Offerel:
Sure. Feel free to move that to any location you want.

Frenzie:
Alright, since you don't have access to Git right now I can push a couple of follow-up commits if you don't mind. It's also important to add the translation as a todo to all languages. (There's a PHP helper script for that.)

Offerel:
Nice. Many thx for that.

Frenzie:
Alright, for the moment I made it look like this:
Screenshot_2020-01-13_12-10-50

What I didn't realize before I looked at the code, however, is that is only affects the favorites link. So really it's more like `Automatically show all articles in favorites` or something along those lines. This may not be an objection to the functionality, but it currently makes the option text incorrect.

marienfressinaud (commented):
You'll need to define the `show_fav_unread` option in the [default configuration file](https://github.com/FreshRSS/FreshRSS/blob/master/config-user.default.php) as well.

And I agree with @Frenzie that a better label can be found. Maybe "Show all articles in favorites by default" if not too long?

If you want/need to, I can make the changes myself. Also I think we can close https://github.com/FreshRSS/FreshRSS/pull/2535 :wink: (sorry for never responding by the way :/)

Frenzie:
`Show all articles in favorites by default` sounds good to me. Please push the changes. :+1:

marienfressinaud (approved):
Good for me now. I just changed a bit the label from what I first suggested: "Show by default all articles in favorites". Having "by default" at the end sounded a bit like the "favorites" were "by default" and it didn't make any sense.

app/i18n/en/conf.php
```diff
@@ -97,6 +97,7 @@
 	),
 	'reading' => array(
 		'after_onread' => 'After “mark all as read”,',
+		'always_show_favorites' => 'Show by default all articles in favorites',
```
Frenzie: Sorry but no. That's ungrammatical.

> Having "by default" at the end sounded a bit like the "favorites" were "by default" and it didn't make any sense.

Not really. "To X (verb) blah blah blah by default" is regular, idiomatic English. The "by default" is about the verb (to X), not about the "blah blah blah." (But I don't think that's any different in French?)

Anyway, you could always phrase it like this. The "articles" business is to remain consistent with the other settings, not by necessity.

```suggestion
		'always_show_favorites' => 'Show all favorites by default',
```
```suggestion
		'always_show_favorites' => 'Show read and unread favorites by default',
```

Frenzie (changes requested):

app/i18n/en/conf.php
marienfressinaud: Ha ha, I was certain to make a mistake with this last-minute change :D

> Not really. "To X (verb) blah blah blah by default" is regular, idiomatic English. The "by default" is about the verb (to X), not about the "blah blah blah." (But I don't think that's any different in French?)

I think French is a bit more permissive here. First form is right, but can be understood in different ways. That's why I moved the "par défaut" after "Afficher", which is also right in French and avoid the confusion. I changed the English version after that.

I'll pick your first suggestion :)

marienfressinaud: Edit: finally not, "Show all favorites by default" makes it sound like the first "Always show favorites". "Show all articles in favorites" was pointing the location where the option has an impact. I'll stay with my first suggestion and just adapt for French ;)

Frenzie: You're right, the original suggestion was best. :+1:

Frenzie (approved):

marienfressinaud (approved):

Issue #2344, opened 2019-04-04
Multiline titles


tiborepcek:
Hi, i would like to ask, if there is some option to enable multiline titles in RSS feeds, so that i can read the whole (long) title of an article. Thank you.

Frenzie:
Not presently, but it may not be too hard to do with some custom CSS. I think I may have once said I'd look into it but I never found the time to.

Part of the problem is that there's no colspan in CSS display:table, so instead a different approach like flexbox needs to be used. A problem in the sense that you probably need at least a good hour to play around with it, not in the sense that it's particularly difficult.

Frenzie:
Quick proof of concept that doesn't look half-bad:

```css
.horizontal-list{display:flex;flex-wrap:wrap}
.flux .item > a {white-space:normal}
.flux:not(.current):hover .item.title {position:relative;max-width:inherit}
.item:last-child{margin-left:auto}
```

![Screenshot_2019-04-04_15-04-37](https://user-images.githubusercontent.com/202757/55557817-fab4ce00-56ea-11e9-8e76-caff1c77665e.png)

aledeg:
See  #1938

tiborepcek:
> Quick proof of concept that doesn't look half-bad:
> 
> ```css
> .horizontal-list{display:flex;flex-wrap:wrap}
> .flux .item > a {white-space:normal}
> .flux:not(.current):hover .item.title {position:relative;max-width:inherit}
> .item:last-child{margin-left:auto}
> ```

Thanks @Frenzie , it works via [Custom CSS](https://github.com/FreshRSS/Extensions/tree/master/xExtension-CustomCSS) user extension on desktop and on mobile too.

aledeg:
Maybe you could add a PR on the extension repository to add your snippet as a tip for other users.

Frenzie:
Not that quick 10 minute proof of concept. But with another 15 minutes, the result is probably about right:

```css
.horizontal-list {
 display: flex;
 /* uncomment to allow title to expand onto its own line */
 /*flex-wrap: wrap;*/
}
.horizontal-list.bottom {
 display: table;
}
.flux .item {
 flex-shrink: 0;
 line-height: normal;
 padding: .3em 0;
}
.flux .item > a {
 white-space: normal;
}
.flux:not(.current):hover .item.title {
 position: relative;
 max-width: inherit;
}
.flux_header .title {
 flex: auto;
}
```

[...]

PR #2771, opened 2020-01-16
fix: #2701 Fix look of authors

### LINKED ISSUE #2701: Article list  displayed with flicker
I recently instlaled FreshRSS via docker on my home server and reading via a mobile app (Reeder4 on iOS in this case) works fine, but when reading via the web interface, there is some nasty flickering and/or overlay issues going on. I tried with Firefox, Chrome & Vivaldi and with the Chrome-based browsers, there is heavy flickering, whereas with Firefox there are overly issues for the article list - each time while hovering the mouse over the article list.

FreshRSS version is 1.15.3, running via linuxserver/freshrss docker on Ubuntu 18.04. Browsers were the respective latest version on Windows-10. The theme was changed with no effects.

image

freshrss_flicker

marienfressinaud:
Closes #2701

- Place "author" element in the "title" block of entries

This fix makes the authors part of the "title item", so it's not
considered as another block. This is not perfect since authors will
disappear on small screens, but we can discuss of putting titles on
multi-lines (see https://github.com/FreshRSS/FreshRSS/issues/2344).

1. Enable "author" display
2. Disable the multiline CSS fix :stuck_out_tongue: 
3. Check the author element is placed next to the title (and not under)

Capture d’écran de 2020-01-16 16-58-38

Capture d’écran - 2020-01-16 à 16 58 56

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -28,9 +28,9 @@
 		}
 	}
 	?><li class="item website"><a href="<?= _url('index', 'index', 'get', 'f_' . $this->feed->id()) ?>"><img class="favicon" src="<?= $this->feed->favicon() ?>" alt="✇" /> <span><?= $this->feed->name() ?></span></a></li>
-	<li class="item title" dir="auto"><a target="_blank" rel="noreferrer" href="<?= $this->entry->link() ?>"><?= $this->entry->title() ?></a><?php
+	<li class="item title" dir="auto"><a target="_blank" rel="noreferrer" href="<?= $this->entry->link() ?>"><?= $this->entry->title() ?><?php
 		if ($topline_display_authors):
-			?><div class="item author"><?php
+			?><span class="author"><?php
 			$authors = $this->entry->authors();
 			if (is_array($authors)) {
 				$first = true;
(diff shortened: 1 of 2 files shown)


Frenzie:
Could you also include before screenshots?

Frenzie (approved):
lgtm but I don't use the feature

marienfressinaud:
Sure! The main issue was the "flickering" behavior (better illustrations are given in the original ticket)

Capture d’écran de 2020-01-16 17-11-48

Capture d’écran - 2020-01-16 à 17 12 01

Frenzie:
The whitespace on the before seems a bit excessive. Is that on all themes? Makes the after look like more of an improvement than it is. :-P

marienfressinaud:
Yes it was on all themes! I don't think it was intended but since I don't use this feature, I didn't notice

triatic:
I'm not sure about this one, I prefer the Author being in the same position for every post when I'm skim-reading the feed list.

marienfressinaud:
It surely can be improved, but the previous version was problematic and hard to fix. I'll let you open another ticket if you think of a better solution, but it will have to wait for https://github.com/FreshRSS/FreshRSS/issues/2344 to be finished.

PR #2761, opened 2020-01-12
Use a real timestamp in greader API.

javerous:
Closes #2759

Changes proposed in this pull request:

- Use a real timestamp in greader API.

How to test the feature manually:

1. Connect to a FreshRSS instance with Reeder (macOS or iOS), and see that dates are related to RSS `pubDate` dates.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

Note: I'm not a PHP developer. Feel free to reject if you don't want this change, or if you want to elaborate a more complete solution.

PR #2773, opened 2020-01-18
Add the ability to customize greader by extension

### LINKED ISSUE #2759: Problem with dates with Reeder / greader API ?
Hello.

So I noticed something weird with dates shown in my feed in Reeder (macOS).

For example, for this feed `https://www.macg.co/news/feed`, there is this article (item):

```
	<item>
		<title>Mozilla corrige une faille critique dans Firefox 72</title>
		<link>https://www.macg.co/logiciels/2020/01/mozilla-corrige-une-faille-critique-dans-firefox-72-111243</link>
		<description>
Gueule de bois en ce début d'année pour Mozilla. Firefox 72, dont la version finale est disponible depuis le 7 janvier, doit être mis à jour de toute urgence en raison de la découverte d'une faille de sécurité « 0 Day » qui a déjà été exploitée par des malandrins. Tous les utilisateurs ayant installé la version 72 du navigateur libre doivent télécharger toutes affaires cessantes la mise à jour fraichement mise en ligne.



L'affaire est manifestement de la plus haute importance : la CISA, l'agence sur la sécurité informatique américaine, a très officiellement alerté les internautes de cette vulnérabilité qui permet à un forban de prendre le contrôle de l'ordinateur infecté. Les versions 72.0.1 de Firefox et 68.4.1 de Firefox ESR (la mouture « grandes organisations » du navigateur) corrigent la faille.

À noter que la version iOS de Firefox n'est pas concerné en raison de l'utilisation du moteur WebKit.
</description>
		<pubDate>Fri, 10 Jan 2020 17:00:29 +0100</pubDate>
		<dc:creator>Mickaël Bazoge</dc:creator>
		<guid isPermaLink="true">https://www.macg.co/logiciels/2020/01/mozilla-corrige-une-faille-critique-dans-firefox-72-111243</guid>
	</item>
```
Which is related to this article: https://www.macg.co/logiciels/2020/01/mozilla-corrige-une-faille-critique-dans-firefox-72-111243

So the date seems to be ok (17h00 for Paris time, i.e. +1). It's displayed the right way in the Web UI (my PHP is configured with the right Timezone).

Then, if I check the `https://[redacted]/freshrss/api/greader.php/reader/api/0/stream/items/contents` associated to this article fetch, I see this content:

```
{
	"id": "user/-/state/com.google/reading-list",
	"updated": 1578673145,
	"items": [{
		"id": "tag:google.com,2005:reader/item/00059bcb73fa3a68",
		"crawlTimeMsec": "1578673009998",
		"timestampUsec": "1578673009998440",
		"published": 1578672000,
		"title": "Mozilla corrige une faille critique dans Firefox 72",
		"summary": {
			"content": "Gueule de bois en ce début d'année pour Mozilla. Firefox 72, dont la version finale est disponible depuis le 7 janvier, doit être mis à jour de toute urgence en raison de la découverte d'une faille de sécurité « 0 Day » qui a déjà été exploitée par des malandrins. Tous les utilisateurs ayant installé la version 72 du navigateur libre doivent télécharger toutes affaires cessantes la mise à jour fraichement mise en ligne.\n\n\n\nL'affaire est manifestement de la plus haute importance : la CISA, l'agence sur la sécurité informatique américaine, a très officiellement alerté les internautes de cette vulnérabilité qui permet à un forban de prendre le contrôle de l'ordinateur infecté. Les versions 72.0.1 de Firefox et 68.4.1 de Firefox ESR (la mouture « grandes organisations » du navigateur) corrigent la faille.\n\nÀ noter que la version iOS de Firefox n'est pas concerné en raison de l'utilisation du moteur WebKit."
		},
		"alternate": [{
			"href": "https://www.macg.co/logiciels/2020/01/mozilla-corrige-une-faille-critique-dans-firefox-72-111243"
		}],
		"categories": ["user/-/state/com.google/reading-list", "user/-/label/Informatique"],
		"origin": {
			"streamId": "feed/51",
			"title": "MacGeneration"
		},
		"author": "Mickaël Bazoge"
	}]
}
```

"updated " 1578673145 → `10/01/2020 - 17:19:05`
"published": 1578672000 → `10/01/2020 - 17:00:00`

So the timestamp seems to be right too.

But then Reeder show the article date `17:16`, which is the exact date when the `https://[redacted]/freshrss/api/greader.php/reader/api/0/stream/items/contents` request was done (so when it fetched the article content).

So do you think it's a bug in Reeder, or is it something invalid in the greader protocol which makes Reeder to be lost ?

Thank you.

Note: I put the whole investigation, so if it appears it's indeed a Reeder bug, I can point this ticket to Reeder dev.

javerous:
Closes #2759 (take 2)

- Use `dateAdded()` in greader API to forge the `crawlTimeMsec` & `timestampUsec` field, so it enforces the fact it's a timestamp + make the ability to change the value independently of the id.
- Add the ability to change the `dateAdded` of a `FreshRSS_Entry`, so an extension can change it in an extension hook (`entry_before_display` by example).

Can be used, by example, to exchange the `crawlTimeMsec` & `timestampUsec`, to workaround some RSS client usage of this field (like Reeder or EasyRSS - see the long ticket discussion…).

1. The best is probably to do an extension like I plan to add on my instance, which is basically:
```
public function init() {
		$this->registerHook('entry_before_display', array($this, 'handle_entry'));
	}
	public function handle_entry($entry) {
		if (isset($_SERVER['REQUEST_URI']) && strpos($_SERVER['REQUEST_URI'], '/api/greader.php') !== FALSE) {
			$entry->_dateAdded($entry->date(true));
		}
		return $entry;
	}
```

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -14,6 +14,7 @@ class FreshRSS_Entry extends Minz_Model {
 	private $content;
 	private $link;
 	private $date;
+	private $date_added = 0; //In microseconds
 	private $hash = null;
 	private $is_read;	//Nullable boolean
 	private $is_favorite;
@@ -68,11 +69,15 @@ class FreshRSS_Entry extends Minz_Model {
 			return timestamptodate($this->date);
 		}
 	}
-	public function dateAdded($raw = false) {
-		$date = intval(substr($this->id, 0, -6));
+	public function dateAdded($raw = false, $microsecond = false) {
 		if ($raw) {
-			return $date;
+			if ($microsecond) {
+				return $this->date_added;
+			} else {
+				return intval(substr($this->id, 0, -6));
+			}
 		} else {
+			$date = intval(substr($this->id, 0, -6));
 			return timestamptodate($date);
 		}
 	}
@@ -119,6 +124,9 @@ class FreshRSS_Entry extends Minz_Model {
 
 	public function _id($value) {
 		$this->id = $value;
+		if ($this->date_added == 0) {
+			$this->date_added = $value;
+		}
 	}
 	public function _guid($value) {
 		if ($value == '') {
@@ -160,6 +168,13 @@ class FreshRSS_Entry extends Minz_Model {
 		$this->hash = null;
 		$value = intval($value);
 		$this->date = $value > 1 ? $value : time();
+	}
+	public function _dateAdded($value, $microsecond = false) {
+		if ($microsecond) {
+			$this->date_added = $value;
+		} else {
+			$this->date_added = $value * 1000000;
+		}
 	}
 	public function _isRead($value) {
 		$this->is_read = $value === null ? null : (bool)$value;
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -496,8 +496,8 @@ function entriesToArray($entries) {
 		}
 		$item = array(
 			'id' => 'tag:google.com,2005:reader/item/' . dec2hex($entry->id()),	//64-bit hexa http://code.google.com/p/google-reader-api/wiki/ItemId
-			'crawlTimeMsec' => substr($entry->id(), 0, -3),
-			'timestampUsec' => '' . $entry->id(),	//EasyRSS
+			'crawlTimeMsec' => substr($entry->dateAdded(true, true), 0, -3),
+			'timestampUsec' => '' . $entry->dateAdded(true, true), //EasyRSS & Reeder
 			'published' => $entry->date(true),
 			'title' => escapeToUnicodeAlternative($entry->title(), false),
 			'summary' => array('content' => $entry->content()),


javerous:
So, @Frenzie & @Alkarex here is an implementation of my proposal, as quickly discussed.

I hope you will consider it, I think it somehow clean and non-intrusive. It keep the original greader API behavior / specs, but allows extension to use the alternative behavior, like proposed in the PR https://github.com/FreshRSS/FreshRSS/pull/2761.

I tested the code, it works fine. And I tested an extension which use it, I'm able to exchange the value as wanted (and as expected by Reeder).

app/Models/Entry.php
```diff
@@ -68,11 +69,15 @@ public function date($raw = false) {
...
 	}
-	public function dateAdded($raw = false) {
-		$date = intval(substr($this->id, 0, -6));
+	public function dateAdded($raw = false, $microsecond = false) {
 		if ($raw) {
-			return $date;
+			if ($microsecond) {
+				return $this->date_added;
+			} else {
+				return intval(substr($this->id, 0, -6));
```
Frenzie, line 77: I realize it already said this, but it still kind of tickles.

Frenzie (commented):
I've always thought something like this was the way to go.

Alkarex (approved):

Alkarex:
Thanks @javerous , that looks good.
For your extension, I suggest you filter on `$_GET['client']` to only target Reeder

Alkarex:
Any remaining comment @Frenzie ?

Frenzie (approved):

javerous:
> Thanks @javerous , that looks good.

Thanks for accepting it :)

> For your extension, I suggest you filter on `$_GET['client']` to only target Reeder

If I decide to publish it on my GitHub (not sure if it can interest lot of people ?), I will elaborate it a bit more, yes. Add a configuration page to configure the User-Agent to which it should apply, by example.

app/Models/Entry.php
javerous: "kind of tickles" ? 🤔 Are your talking about `substr` ? Yes, I was a bit surprised too. I wanted to do an integer divide, but apparently is not that easy in PHP ! There is `intdiv`, but it's available only since PHP 7.

By the way, whith your comment, I note I'm still using `$this->id` while I wanted to use `$this->date_added`, when not `$microsecond`, for coherency / homogeneity of this getter 😩🤦‍♂️

In practice, it doesn't change anything (we are never using a changed `date_added` when `$microsecond == false`), but can be good to fix it, no ? Do I open a little PR for that ?

Frenzie: Yes, please fix it. :+1: 

It kind of tickles because it's just chomping off part of the string rather than rounding. I doubt it matters, but it feels funny. ;-)

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/2774

ybbond:
sorry to comment in old thread

@javerous I read the #2759 issue (skimming actually) and ended up here, but cannot found the solution.

is there currently an **extension** to make Reeder display published date? I just purchased it and setup my FreshRSS selfhosted app (v1.16.0), and I found the issue in Reeder still remain

thanks before!

Frenzie:
The extension is at https://github.com/javerous/freshrss-greader-redate

javerous:
@ybbond So yes, I created this little extension. I use it every day since some months : it works well, as much as I can tell. You will need FreshRSS 1.16.0 to make it work (so: good for you).

It's "compatible" with EasyRSS too, but because this client doesn't send specific headers as Reeder do, it's far weaker (i.e it can impact others clients if activated).

[Edit]
Side note: Reeder keep a local cache, so the extension will not have any impact on articles already fetched and cached.

I didn't find any way to remove the cache on iOS version (except… uninstall and reinstall the whole app), and on macOS you can remove those files

```
~/Library/Containers/com.reederapp.macOS/Data/Library/Application Support/default.realm
~/Library/Containers/com.reederapp.macOS/Data/Library/Application Support/default.realm.lock
```

Alkarex:
Ideally a bug should be escalated back to Reeder, for them to pick the appropriate JSON field in the responses: https://github.com/FreshRSS/FreshRSS/issues/2759#issuecomment-575350382

javerous:
@Alkarex And EasyRSS, which have the exact same problem (at least the last time I checked), or at least the same interpretation of what this field is supposed to be…

But I agree, the best would be to fetch the right fields. But Reeder dev is never answering, so… It's why I preferred to create the extension ;)

ybbond:
Thanks! The extension works!

Yes, the extension that acts like middleware is not ideal, more that FreshRSS itself has been listed as "oficially supported" in Reeder.

javerous:
@ybbond I suspect that Reeder is just reusing common code for Google Reader when setting up a FreshRSS account, as it used to be compatible with Google Reader. And I suspect that Google Reader was setting this field in a way which used to make more sense in Reeder way to handle it.

I don't know who is wrong there : Google Reader, Reeder & EasyRSS, of the reverse engineering of the API.

Alkarex:
See my comment https://github.com/FreshRSS/FreshRSS/issues/2759#issuecomment-575350382 for an example of how it was in Google Reader. The field names are also quite clear

javerous:
@Alkarex Yes yes, I read them when you originally posted it. But it's all result of reverse engineering, right (i.e. lot of interpretation) ? If I well understand, Google never gave official doc of its API.

Alkarex:
The payload I gave as example was provided by Google Reader, not reconstructed.
As an example, I have just checked that [Readrops](https://github.com/readrops/Readrops) and [FeedReader](https://jangernert.github.io/FeedReader/) do it correctly. I will try to find time to fix EasyRSS

javerous:
@Alkarex Ah yes, indeed. I missed the "output from the **original Google Reader**". So yes, I'm not sure why Reeder & EasyRSS did that this way 🤷‍♂️

Alkarex:
@javerous Maybe not to have to maintain two sets of dates (the publication date is problematic for anything else than being informative for the user and cannot be reliably used for sync, paging, etc.)

javerous:
@Alkarex The publication date is also important for article sorting, from my point of view (but I remember there was some disagreement on this point ^^)

Alkarex:
Fixed in EasyRSS 😃 https://github.com/Alkarex/EasyRSS/pull/49

PR #2778, opened 2020-01-20
Enhance content path feature

### LINKED ISSUE #2772: Being able to flush / recompute entry content when changing "article CSS path".
Hello,

I'm playing a bit with "Articles CSS path on original website" advanced features on Feed configuration.

I have some difficulties to use it. Apparently, it's impossible to apply this setting to feed content already fetched, which means:
- It's impossible to fix old articles with their website content.
- It's impossible to fix content fetched when subscribing to a new feed, as articles are fetched right away.

Aside the fact we miss the feature for pre-existing articles, it make it difficult to check if the path is correct, as we need to wait for new articles to be fetched to verify the result.

I didn't find any way to force FreshRSS to "recompute" content, so it updates fetched article according to this setting change. I tried to set to NULL the content field in the database, but it didn't help (FreshRSS doesn't try to re-fetch content in this case).

Wouldn't be nice to be able to force flush content fetched, so this setting had a chance to be re-evaluated ? Or automatically force reapplying this setting to fetched content when it changes ?

Thanks you !

javerous:
Closes #2772

- Add ability to show a whole screen pop-up which don't automatically disappear (instead of notifications).
  - `/p/themes/Origine/origine.css`
  - `/p/scripts/main.js`
  - `/app/layout/layout.phtml`
- Add ability to reload a whole feed and clear cache
  - `/app/Controllers/feedController.php`
  - `/app/views/helpers/feed/update.phtml`
  - `/p/themes/icons/screwdriver.svg`
  - `/app/Models/Themes.php`
- Add ability to preview content path on first article of a feed, to check if it's correct
  - `/app/Controllers/feedController.php` // this is probably not the right place
  - `/p/scripts/extra.js`
  - `/app/views/helpers/feed/update.phtml`
  - `/p/themes/icons/look.svg`
  - `/app/Models/Themes.php`

1. Use the new UI elements in feed configuration side panel.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Screenshots:
- _The preview button next to path field (the "eye" icon) + help moved on next line_
Screenshot 2020-01-22 at 22 53 31

- _The popup UI which preview the result of the current content-path (no need to be saved to be able to have a preview, of course)_
Screenshot 2020-01-22 at 22 54 35

- _The new Maintenance section with 2 actions - help is perhaps redoundant, or perhaps need to be enhanced_
Screenshot 2020-01-22 at 22 52 36


javerous:
For now, it's mainly a draft, to see if the core team is interested in the change. If yes, there is still a bunch of things to do (propagate translation, propagate CSS things, etc.)

- I'm not a PHP / JS developer, so my JS changes are perhaps not the best way to do what I wanted to do. Though, I tried to follow what already exists.
- There is a lot of security restriction with JS I didn't want to touch (and which I wasn't even aware it exists…) like onclick which can't be random JS anymore, apparently. It complexifies usage of JS, so as previous point, I did my best to make it work, without being (too) dirty.
- The content path preview is generated by `feedController.php`, but it's probably not the right place to do it. Feel free to suggest a better place for this kind of helper / tools, without having to re-implement an authentication system, like with APIs. I added some suggestions in the code, under `XXX` comment.
- I thought it would be a good idea to add a new `Maintenance` section in the feed configuration, which group a bunch of maintenances buttons. It's useful for this ticket, but I'm not 100% sure it's the best way to handle it.

About this last point, some extra thinking:
- Once someone do "Delete all articles" in the archiving section, the entries are then very slowly (i.e. after a very long moment) refetched, because of the cache, at least on my tests. So this feature by itself was not very practical to reapply content-path to articles. Plus, it forces to remove all articles already fetched by FreshRSS, and some of them are not in the remote server feeds anymore, so it's a bit unfortunate to lost articles to be able to reapply content-path.
- I wanted to add a checkbox next to the content-path field to force reapply the path to articles, but it looked a bit weird: it would have been just a temporary check box, as an extra parameter to the Submit, but for a UX point of view, it looked like a setting, which was confusing. So I thought it would be nice to have a dedicated button for that in a new "Maintenance" section: "Reload Articles". This button re-fetch everything (without removing articles already fetched), and then reapply path-content to every article in the feed. It can be time and bandwidth consuming, so perhaps this feature can be limited to admins ? At least, this was perfect for me.
- I thought it would be fine to be able to clear the cache of feed (and not only automatic clearing), so I added a button in "Maintenance". It's not specifically linked to the ticket (even if it can be used to speed up re-fetch after an "Delete all articles"), but looked nice to have this kind of feature too.

p/scripts/main.js
```diff
@@ -1215,6 +1215,80 @@ function init_notifications() {
 }
 // </notification>
 
+// <popup>
+var popup = null,
+	popup_working = false;
+	
+function openPopupWithMessag(msg) {
```
Alkarex: Typo at "message" ;-)

Alkarex:
Thanks @javerous 
At a quick first glance, it seems to look good 👍 
I will try to test soon. Do not hesitate to post screenshots to get faster feedback from more people on the general concepts.

app/Controllers/feedController.php
```diff
@@ -647,6 +647,126 @@ public function deleteAction() {
...
+	 * This action force to clear cache of a feed.
+	 *
+	 * Parameters are:
+	 *   - id (mandatory - no default): Feed ID
+	 *
+	 */
+	public function clearcacheAction() {
+		//Get Feed.
+		$id = Minz_Request::param('id');
+
```
Alkarex: Please trim whitespace at the end of all lines.
Cf. https://travis-ci.org/FreshRSS/FreshRSS/jobs/639279131

p/scripts/main.js
```diff
@@ -1215,6 +1215,84 @@ function init_notifications() {
...
+}
+
+function init_popup() {
+	// Fetch elements.
+	popup = document.getElementById('popup');
+	popup_iframe = document.getElementById('popup-iframe');
+	popup_p = document.getElementById('popup-p');
+
+	//Configure clode-button.
+	document.getElementsByClassName("popup-close")[0].onclick = function() {
```
```suggestion
	document.getElementsByClassName('popup-close')[0].onclick = function() {
```

Also this kind of thing may be a bit undependable, although nowadays the DOM between browsers is probably more stable than it used to be. It's better practice to get the relevant child node more dependably than just a random mystery index.

p/scripts/main.js
```diff
@@ -1215,6 +1215,84 @@ function init_notifications() {
...
+	popup_working = false;
+}
+
+function init_popup() {
+	// Fetch elements.
+	popup = document.getElementById('popup');
+	popup_iframe = document.getElementById('popup-iframe');
+	popup_p = document.getElementById('popup-p');
+
+	//Configure clode-button.
```
```suggestion
	//Configure close button.
```

As an aside, I'm not sure if we have a agreed upon standard for this, but I find it much nicer to write `// Comment.`. (I wasn't going to say anything about it since I'm already completely failing to focus on substantial code logic but then you wrote `// Fetch elements.` :-P)

Frenzie (commented):
Not sure what I think of the icons, especially because it looks a bit like an illustration. just a textual `Preview` button might be preferable.

Frenzie:
Btw, you can just paste the JS into <https://jshint.com/>. You don't need to install it locally or wait for the CI.

p/scripts/main.js
javerous: Not sure how to do that :s

p/scripts/main.js
```diff
@@ -1215,6 +1215,84 @@ function init_notifications() {
...
+  	//Configure close-on-click.
+	window.onclick = function(event) {
+		if (event.target == popup) {
+			closePopup();
+		}
+	}
+
+	//Configure iframe size.
+	popup_iframe.onload = function(event) {
+		var size = document.documentElement.clientHeight - 250;
```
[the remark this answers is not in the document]
javerous: I tried some percent thing, but nothing was working (either de iframe was not cliped by the block, either the block was too large).

I think I reach my non-web developer capability there. How you would do that ?

p/scripts/main.js
Frenzie: It's more of a general comment really. Since it's 2020 you can get away with just doing `document.querySelector('.popup-close').addEventListener('click', …` which is really just the same thing as the 0 index anyway.

app/layout/layout.phtml
```diff
@@ -66,5 +66,14 @@
 	<span class="msg"><?= $msg ?></span>
 	<a class="close" href=""><?= _i('close') ?></a>
 </div>
+
+<!-- Popup-->
+<div id="popup" class="popup">
```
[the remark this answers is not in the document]
javerous: Isn't what is already done for notification ? `<div id="notification" class="notification"…`

app/i18n/en/sub.php
```diff
@@ -72,6 +72,13 @@
 		'validator' => 'Check the validity of the feed',
 		'website' => 'Website URL',
 		'websub' => 'Instant notification with WebSub',
+		'maintenance' => array(
+			'title' => 'Maintenance',
+			'clear_cache' => 'Clear Cache',
+			'clear_cache_help' => 'Clear the cache of this feed on disk',
+			'reload_articles' => 'Reload Articles',
```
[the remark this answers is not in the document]
javerous: In english, the capitalization is not the norm ? I'm not an english expert. That being said, it's right it's not the case on other UI elements, so we need to be at least homogeneous. 👍

app/Controllers/feedController.php
```diff
@@ -647,6 +647,126 @@ public function deleteAction() {
...
+		$entryDAO->commit();
+
+		//Give feedback to user.
+		Minz_Request::good(_t('feedback.sub.feed.reloaded', $feed->name()), array(
+			'params' => array('get' => 'f_' . $feed->id())
+		));
+	}
+
+	/**
+	 * This action forge a preview of a content path.
```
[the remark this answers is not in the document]
javerous: Okay, it's an habit naming from my dev work.

app/layout/layout.phtml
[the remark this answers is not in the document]
Frenzie: Same comment for that one then. Anyway, let's ignore this then. ^_^

p/scripts/extra.js
```diff
@@ -3,6 +3,23 @@
 /* globals context, openNotification, xmlHttpRequestJson */
 /* jshint esversion:6, strict:global */
 
+function fix_popup_preview_path() {
+	var links = document.getElementsByClassName('popup-preview-path');
+
+	for (var i = 0; i < links.length; i++) {
+		var link = links[i];
+
+		link.onclick = function (event) {
```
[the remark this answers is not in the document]
javerous: Ah yes ? Why that (it's real question, as said, I'm not a really web developer).

p/scripts/main.js
javerous: Okay, I will use that.

app/i18n/en/sub.php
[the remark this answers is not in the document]
Frenzie: There is no real norm. If you look at places where they've actually thought about it they'll just tell you to be consistent with whatever you choose. For example [this](https://developer.apple.com/library/archive/documentation/FinalCutProX/Conceptual/FxPlugHIG/TextStyleGuidelines/TextStyleGuidelines.html#//apple_ref/doc/uid/TP40013782-CH6-SW1) is what Apple says:

> Use either title-style or sentence-style capitalization for all interface element labels.

p/scripts/extra.js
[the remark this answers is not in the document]
javerous: By the way, why Travis is telling me that
```
Functions declared within loops referencing an outer scoped variable may lead to confusing semantics. (link)
```
Not sure if it's an error, but I don't understand how it can be a problem, it's common to refer a variable defined on parent block, no ? How do you fix that ?

p/scripts/main.js
```diff
@@ -1215,6 +1215,84 @@ function init_notifications() {
 }
 // </notification>
 
+// <popup>
+var popup = null,
```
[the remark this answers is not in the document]
javerous: Okay, I just copied what is done for notification. I'm not sure of the difference between let & var (well, we have them in Swift, and it's mutable vs immutable, not sure it have the same semantic in javascript). I will take a look.

p/scripts/main.js
```diff
@@ -1215,6 +1215,84 @@ function init_notifications() {
...
+	if (popup_working === true) {
+		return false;
+	}
+
+	popup_working = true;
+
+	popup_p.innerHTML = msg;
+	popup_p.removeAttribute('hidden');
+
+	popup.style.display = "block";
```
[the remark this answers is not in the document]
javerous: Mainly because I googled a lot about how to create pop-up. Plus lake of habit to play with different manner to declare a string ^^'

I will fix all of them.

p/scripts/extra.js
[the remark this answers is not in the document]
Frenzie: If you're using .onclick you're overriding everything else that may have been set.

If you're using addEventListener, you can attach (and easily remove) multiple listeners as desired.

p/scripts/main.js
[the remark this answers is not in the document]
Frenzie: No, `let` is a matter of scope working as you expect it to. Basically JavaScript `var` is kind of broken.

The program was written pre-`let`, or at least prior to `let` being something that one could merrily use outside of Firefox.

p/scripts/main.js
javerous: Yes, It's my abit to use a space, but I saw that most common doesn't have, so I tried to follow current code-style. I failed on some places :/ ;)

javerous:
> Not sure what I think of the icons, especially because it looks a bit like an illustration. just a textual `Preview` button might be preferable.

The textual button are so big… Perhaps can be a button with this icon but no text ? ^^'

What others are thinking about that ?

javerous:
> Btw, you can just paste the JS into https://jshint.com/. You don't need to install it locally or wait for the CI.

Ah yes, think you ! It will help.

p/scripts/extra.js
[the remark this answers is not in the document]
javerous: Yes, okay. Good point, thanks.

Frenzie:
> The textual button are so big… Perhaps can be a button with this icon but no text ? ^^'

Note that my main point is that you should make it look like a button, like these ones:

Screenshot_2020-01-20_11-31-37

Edit: In English

Screenshot_2020-01-20_11-33-03

p/scripts/main.js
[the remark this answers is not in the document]
Frenzie: It doesn't work out if you simply give it a max-height of 90% or something?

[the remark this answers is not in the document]
javerous: Yes, it's what I tried. The block was at the right size, but then the iframe content was not clipped anymore (I tested in Safari & Chrome). Didn't make sense for me, but I have no idea how to solve it…

I will take a look this evening.

javerous:
> Note that my main point is that you should make it look like a button, like these ones:

Okay, I will try this.

p/scripts/extra.js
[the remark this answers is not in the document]
javerous: That being said, I had to revert the change. We need to cancel the event by returning false (to prevent the browser to actually redirect to the link), which is not possible with a listener.

app/Controllers/feedController.php
javerous: Done.

javerous:
> > The textual button are so big… Perhaps can be a button with this icon but no text ? ^^'
> 
> Note that my main point is that you should make it look like a button, like these ones:

Done (screenshot updated).

javerous:
(Note: sorry for the force push, it was to change my e-mail)

p/scripts/extra.js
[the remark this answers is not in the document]
Frenzie: I beg your pardon?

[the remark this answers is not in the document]
Frenzie: I'm pretty sure "return false" should do that just fine. But in any case, also see https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault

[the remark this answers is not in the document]
Frenzie: I was probably confused between proper JS and jQuery. In any case preventDefault is the right way to do it.

javerous:
So I added a side proposal in commit 41e8461. The idea is to have a new controller (like the ActionController) which is in charge of controller generating only raw content, like for the path-content preview.

If you feel this proposal change too many things just for this feature, then I can revert it. If you think that naming is not good, I can change it (`Content`, `cont`, etc.).

As usual, I tested manually, and it works well.

@marienfressinaud If I well understood, you are the one who created the Minz dispatching system in the first place. What do you think of this proposal ? I know that dispatching was mainly expecting an Action Controller, but I feel that another type of controller can make sense, and don't create too much mess...

javerous:
So at this point, it remains mainly "details" (adding translation & propagate to other languages, propagate popup to other themes, etc.). I will do it if I have confirmation the core team is okay with the whole feature idea (and last changes).

@marienfressinaud I'm conscious that this content controller idea break a bit the MVC concept. That being said, I think it's okay to be flexible with it, and accept some "side channel", if it doesn't destroy the whole model.
A possible alternative would (perhaps) be to have custom Minz_View which generates the actual content, but not sure if it would be actually better.

app/Controllers/previewController.php
```diff
@@ -0,0 +1,59 @@
...
+			echo 'no-feed-for-entry'; // FIXME: translate.
+			return;
+		}
+
+		$feed->_pathEntries($content_path);
+		$entry->loadCompleteContent(true);
+
+		echo '<!DOCTYPE html><html><body>';
+		echo $entry->content();
+		echo '</body></html>';
```
aledeg: Is there a reason why you echoed the content instead of using a template?

app/Models/Themes.php
```diff
@@ -95,6 +96,7 @@ public static function alt($name) {
 			'rss' => '☄',
 			'unread' => '☐',
 			'refresh' => '🔃',	//↻
+			'screwdriver' => '🔩',
```
aledeg: I wonder if this is appropriated. I understand that you want to display it along with actions that update the cache. My concern is about how users will understand that icon.

aledeg (commented):
Don't forget to add the other languages before shipping that feature! Use the Makefile to do that.

javerous:
> Don't forget to add the other languages before shipping that feature! Use the Makefile to do that.

Yes yes. As said, I'm waiting to have an "agreement in principle" of the whole feature and structural choices I did, before going into "annoying details" ;) Thank for the Makefile indication btw, I thought I would have to do all by myself :)

app/Models/Themes.php
javerous: Yes, I agree. I didn't find any screwdriver emoticons, and as it was just alt (I added a real screwdriver svg icon)…

Do you have better idea ? Perhaps use the `⚙︎` or `⚙️` as alternative ? Or remove `screwdriver` in a whole, and use `configure` instead ?

app/Controllers/previewController.php
javerous: Not completely sure of your point but the content, there, is going to be (most of the time) raw partial HTML structure extracted from the website itself (and not the feed content). So it doesn't fit well with our templating system (?).
And anyway, the style while depends on the RSS client (FreshRRS Web UI, Reeder, etc.) showing it, so I'm not sure it makes sense to force a style ?

Or, if you mean that the enclosure (`<html><body>…`) should be in a template, then why not. No special reason, except I wasn't aware of this mechanism.

Or, if you mean that using the `echo` function is not a good idea for some reason, then same: no special reason, except that I wasn't aware of a better mechanism.

marienfressinaud:
I don't have much time to review your PR in details this week (I struggle to organize myself and I have a conference on Thursday and Friday, the days I dedicate to work on FRSS :/). Just few comments:

- this PR is big and hard to review (for me at least). If you extract new actions `clearcache` and `reload` in a different PR (I think it shouldn't be too hard), we could probably merge it quickly (note I didn't made my opinion on this code yet :p)
- I don't understand the purpose of this new `Minz_ContentController` class. I'm pretty sure you can achieve what you want with the current system. Looking at the commit https://github.com/FreshRSS/FreshRSS/commit/41e846140a8d21e23538c846869d674e85583c8a, I would say you just had to do something like that:

```php
public function previewContentPathAction() {
	// [lines from 738 to 765]

	$this->view->_layout(false);
	$this->view->content = $entry->content();
}
```

and in the view file (`app/views/feed/previewContentPath.phtml`):

```html
<!DOCTYPE html>
<html>
	<body>
		<?= $this->content ?>
	</body>
</html>
```

Even if Minz is not perfect, adding this new controller concept is really not needed, and worse: it introduces fresh new code (so less tested) in the core of the framework, which increase the risks of bugs.

javerous:
@marienfressinaud Thank you for your answer. My main concern was it doesn't fit well the semantic of an `action`.  Anyway, I will take a deeper look at your suggestion, it seems interesting, even if it prevent this preview controller to produce any kind of preview (it will be limited only for raw HTML content).

For `reload`, it's actually part of the path enhancement, as it *can* be used to apply a new path to the whole feed.

For `cache` clearing, yes indeed. That being said, I'm not in an hurry (but perhaps you are to release 1.6).

marienfressinaud:
> My main concern was it doesn't fit well the semantic of an action.

I understand, but I think it's ok. Maybe the word "action" wasn't the one I should choose initially ^^ But don't mind about this, I even [render PDFs](https://github.com/flusio/xExtension-Flus/blob/master/Controllers/billingController.php#L358) with actions in my own extension, so if you tell me that you don't succeed to render a bit of HTML, I'm not going to believe you :p

Ok with the rest of your comment :)

javerous:
> so if you tell me that you don't succeed to render a bit of HTML, I'm not going to believe you :p

Ha ha, yes. Actually I had it working with a view, with a bit different method, but it looked hacky. 

https://github.com/javerous/FreshRSS/blob/967cf9d9a9c2ca9332bad7501d350ea6100bd0b2/app/Controllers/feedController.php#L734

I will try your solution.

app/Controllers/previewController.php
Frenzie: The source may be more useful than the rendered result (not saying you can't have both), as in something like `htmlspecialchars($entry->content())`.

Frenzie:
@javerous It's not quite working as expected for me giving it a quick try, but perhaps this is an issue with the original functionality?

If I enter something nonsensical like `bla` I expect it to show nothing, but instead it just seems to show the entry as in the feed. Only a selector that actually matches something seems to affect the result.

javerous:
> @javerous It's not quite working as expected for me giving it a quick try, but perhaps this is an issue with the original functionality?
> 
> If I enter something nonsensical like `bla` I expect it to show nothing, but instead it just seems to show the entry as in the feed. Only a selector that actually matches something seems to affect the result.

Yes, it's the result of `loadCompleteContent`. If the path (selector) doesn't match anything (produce empty result), it fallback to the original feed content, which is probably a good thing.

I can add an option to `loadCompleteContent` to prevent this fallback for the preview, but I **think** it makes sense to show what the path entered will *actually*  produce when seeing an article managed by FreshRSS.

app/Controllers/previewController.php
javerous: @Frenzie I'm not sure to understand :s

Anyway, I guess that suggestion from @marienfressinaud will fix this concern ?

marienfressinaud: If I understand, the suggestion of @Frenzie is to show the raw HTML instead of the rendered, but I don't agree. We might have a toggle to switch (a bit like the GitHub comment editor), but as a user I largely prefer to "see" what I'll get instead of some HTML

javerous: Ah yes, I see. So well, having the raw HTML can have interest to help write the path, indeed. But yes, having the actual rendered result look better for an end-user.

I will check if I can add the two (like the rendered first, and the raw after), or even having some web UI to switch between the two à la GitHub comment.

Frenzie:
> I can add an option to loadCompleteContent to prevent this fallback for the preview, but I think it makes sense to show what the path entered will actually produce when seeing an article managed by FreshRSS.

That's a very black & white way of looking at it. ;-)

There are two facts:

1. You want to know that your selector didn't match anything.
2. You want to know what FreshRSS will actually show.

At its most basic you could resolve that by displaying something like:

```html
<p class="warning">The selector didn't match anything. As a fallback the original feed text will be displayed instead.</p>
```

javerous:
> That's a very black & white way of looking at it. ;-)

No no, I didn't say it should be like this, it was just my point of view (so the emphasized **think** ;)).

> <p class="warning">The selector didn't match anything. As a fallback the original feed text will be displayed instead.</p>

Yes, look like a good idea.

app/Models/Themes.php
aledeg: The thing is I do not have a better idea. :/
I am just wondering if we won't end in the unknown icon hell.

app/Controllers/previewController.php
Frenzie: @marienfressinaud I explicitly didn't say instead. ;-)

app/Models/Themes.php
javerous: Yes, I'm aware of this unknown icon hell. And I'm not a UX expert… Perhaps `🛠` or `🔧`, and rename the icon as `maintenance`, `repair`, `housekeeping` or `tune-up` (and keep the screwdriver as the svg icon ?).

Look a bit more coherent, generic but still somehow self-explanatory ? For sure, the icon can't be used alone, but…

Frenzie: What exactly is the goal you wish to achieve by adding this screwdriver icon to these two buttons?

javerous: Decoration ? Relief ? Like all button with icon…

In this case, to denot it's all maintenance, housekeeping…

[Edit] I mean, it's not that I'm an extremist of icons in buttons, but a bunch of other buttons have icons, so in sake of homogeneity and coherency (even if it's not very homogeneous in FreshRSS, currently)…

If all of you have concerns with it, then I can remove it, but perhaps can be good to some "audit" to make things homogeneous, or at least define clear rules (i.e. not based on by-case feeling) so we know when we should add an icon, and when we should not.

Frenzie: All button icons where? The only one I see is the standard refresh icon on the endearingly named actualize button. ;-) (Actualize is French for refresh; it's English for making something actual/real.) HIG guidelines typically say something like [this](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/#using-symbols-in-menus):

> Avoid using custom symbols in menus. People are familiar with the standard symbols. Using nonstandard symbols introduces visual clutter and could confuse the user.

Frenzie: > we know when we should add an icon, and when we should not.

When in doubt, don't. ^_^

javerous: My doubt was not that strong :p 

I agree from Apple's HIG guidelines, but most of the time, I try to follow what is currently done, at least for homogeneity.

By example, in this panel:
- There are icons for side actions (like the pointer icon for Website URL & Feed URL, and Actualize)
- There are icons for help text (I thought it was a button or a mouse-over tooltip, the first time).
- There is not icons for main actions, like submit or cancel.
- And there is some intruder, like "Check the validity of the feed" or "Delete all articles".

My decision was just trying to homogenize by "category".

Frenzie: > There are icons for side actions (like the pointer icon for Website URL & Feed URL, and Actualize)

These are a different category.

> There are icons for help text (I thought it was a button or a mouse-over tooltip, the first time).

I believe it's a lifebuoy. Definitely not very clear.

> And there is some intruder, like "Check the validity of the feed" or "Delete all articles".

Not an intruder, perfectly consistent.

Those new buttons need to be blue to be consistent with the current UI. (For better or worse.)

javerous: > Not an intruder, perfectly consistent.

You created them ? :p

> Those new buttons need to be blue to be consistent with the current UI. (For better or worse.)
 
But then, why "Check the validity of the feed" is not blue ? 

Well, we don't have the same categorizing algorithm, I think. But that's okay, I will turn my buttons blue, and remove the icons. But I will stay the preview button as-it, I think an eye is better than a textual "Preview".

Frenzie: > You created them ? :p

Nope, you might say I created surprisingly little. But I didn't get involved until post 1.0. ;-)

(You have me to thank for all the automated CI comments on your PHP, JS and CSS, for example. :-D)

> But then, why "Check the validity of the feed" is not blue ?

The way I see it, blue signifies an action that will have lasting consequences. (It will submit the form.)

javerous: > (You have me to thank for all the automated CI comments on your PHP, JS and CSS, for example. :-D)

Thank you ! (no, seriously, we have a lot of disagreement, but I really appreciate all your comments, and the effort you put in the review !).

javerous: Fixed.

javerous:
On a little side note:
- All around the code, the term "path" is used (and it ends in `$selector` parameter of `phpQuery::find($selector...)`)
- In the UI, the text is `Articles CSS path on original website`.
- @Frenzie is using the term "selector" (which seem coherent with phpQuery naming).

Without changing "path" all around the code, it would perhaps be nice to take the opportunity of this PR to change the Web UI name ? I'm not sure why it's "CSS" (perhaps some historical reason), but doesn't seem to be right anymore (?).
 Something like `Article HTML selector` (something shorter than before), and for the help text something like `Extract content from original website in place of RSS feed (caution, requires more time!)` + a link to some selector format doc, if anyone knows one (I didn't find anything clean for phpQuery, so perhaps we can put a link to a jQuery doc ?).

Frenzie:
It's a CSS selector, implemented in PHP. jQuery has nothing whatsoever to do with it, except that once upon a time it implemented CSS selectors in JS when browsers didn't. THe right term is CSS selector.

javerous:
Okay. I said that because I saw some site saying that "phpQuery is a jQuery for PHP" (e.g. there https://www.berejeb.com/2010/07/phpquery-le-portage-jquery-en-php/), and that the doc for https://api.jquery.com/find/ seems to work with `phpQuery::find()` (it's how I created my own selectors).

By the way, what is the relation with CSS ? I mean, I have a selector which is `article:first`, and it have nothing to do with CSS… (or it has a relation, and I don't understand the relation between standard HTML nodes and CSS ?).

Frenzie:
Fair enough, it's jQuery for PHP. But "CSS selector" is what you'll want to put into a search engine if you don't know what this stuff is. "By the way, it technically also supports various jQuery extensions" isn't something that should be reflected in the primary label, imho.

> or it has a relation, and I don't understand the relation between standard HTML nodes and CSS ?

It's a jQuery extension from long before `:first-child` was widely supported. A bit like how `filter` was a non-standard Internet Explorer property before it was standardized.

Frenzie:
Oh, my bad, I was wrong about what it does based on the name. When writing JS you should probably use standard `querySelector` instead.

https://api.jquery.com/first-selector/
> Selects the first matched DOM element.

But anyway ;-)
> Because :first is a jQuery extension and not part of the CSS specification …

javerous:
I just changed `path` -> `selector` in the English version of the Web UI. For some reason, all other languages seem to use "selector". I will not touch more than that this text, as it's already translated in all languages…

p/scripts/main.js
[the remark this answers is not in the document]
javerous: Okay, I tried something, but it's more "voodoo" than anything else. I definitively don't understand CSS (and I barely understand how people can enjoy using this…).

At least, the notification cliping for iframe seems to works as expected, now (tested only on Chrome and Safari).

aledeg:
@javerous How was the use of the Makefile to generate the missing translation strings? I am curious to hear if it is useful or not, if it needs some improvements.

javerous:
@aledeg The point is that, as I didn't know this script, I added my strings manually to the English language files, with the idea I will add to other files at the very end, once texts are stabilized.

So, when I wanted to propagate my new keys to other languages, now being aware of your script, my idea was "I'm going to use `make i18n-add-key` to add each keys / values I manually added in the English file to other languages, and it will just replace existing key with the same values in my English file".

It failed because of key duplication, so I commented `throw new Exception('The selected key already exist.');` (line 145) in data file. Then it indeed propagated my key as expected, but also proposed other keys, which was not really expected (and not sure if the intended behavior). But it's exactly what I wanted.

So it worked, but I didn't use the standard path. Perhaps having a “propagate" command can help, for cases like mine. By example, something like `make i18n-propage-lang lang=en`, which would just copy new keys (and remove removed keys) from the English file to other language files (it can even detect keys marked as “TODO” in other languages file to replace them with existing key value of the English file — it can facilitate sync for value which changes, and which are still not fixed in other languages) .

(And yes, I'm unable to write short answer…)

Frenzie:
I think the format action will do that, won't it? (Talking about calling `manipulate.translation.php` directly here; didn't check the Makefile command.)

marienfressinaud:
You could also have removed the key (manually or with the `make` command), and re-add it with `make`. Something can be improved here but commenting the exception wasn't necessary ^^

javerous:
@Frenzie Possible, I didn't test.
@marienfressinaud Yes. But there were multiple keys to pass, and I didn't know the first one would also create other keys, so, by laziness…

javerous:
@aledeg Another point: it seems the shorting algorithm is not stable (perhaps some libs, or the PHP version): it re-order the `date` fields of `gen.php` on my side. I suspect the upper-case comparison.

app/Controllers/previewController.php
javerous: Okay, I did something which works. Not sure it's the best way to do it, but look somehow ok to me, for a preview at least.

javerous:
Okay, I think it's done on my side. Hope the whole PR will be okay for all of you ;)

javerous:
@Frenzie:
> I think the format action will do that, won't it? (Talking about calling `manipulate.translation.php` directly here; didn't check the Makefile command.)

I just tested, and no, it's not like a "sync" or "propagate". It just reformat the files (and remove all the `// TODO - Translation`…).

app/i18n/en/sub.php
```diff
@@ -34,7 +34,7 @@
 		),
 		'clear_cache' => 'Always clear cache',
 		'css_help' => 'Retrieves truncated RSS feeds (caution, requires more time!)',
-		'css_path' => 'Articles CSS path on original website',
+		'css_path' => 'Articles CSS selector on original website',
```
Frenzie:
```suggestion
		'css_path' => 'CSS selector for article on original website',
```
```suggestion
		'css_path' => 'Article CSS selector on original website',
```

p/themes/base-theme/template.css
```diff
@@ -772,6 +772,73 @@ br {
...
+	display: none;
+	position: fixed;
+	z-index: 1;
+	left: 0;
+	top: 0;
+	width: 100%;
+	height: 100%;
+	overflow: auto;
+	background-color: rgb(0,0,0);
+	background-color: rgba(0,0,0,0.4);
```
Frenzie, line 786: Since the idea here is something a lot closer to "gray" than to "black", the fallback on the preceding line should probably be something closer to `#eee` than to `#000`?

Frenzie (approved):
I have no pressing concerns remaining. Nice job!

p/themes/base-theme/template.css
javerous: Yes, you are right. Fixed.

javerous:
> I have no pressing concerns remaining. Nice job!

Thank you! I hope the feature will be validated by others ;)

app/views/helpers/feed/update.phtml
```diff
@@ -354,5 +358,27 @@
...
+				<br/>
+				<div class="stick">
+					<a class="btn btn-important" href="<?= _url('feed', 'reload', 'id', $this->feed->id()) ?>">
+						<?= _t('sub.feed.maintenance.reload_articles') ?>
+					</a>
+				</div>
+				<br/>
+				<?= _i('help') ?> <?= _t('sub.feed.maintenance.reload_articles_help') ?>
+			</div>
+		</div>
```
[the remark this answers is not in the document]
marienfressinaud: Also I think I saw discussions about what kind of buttons should be used here but I had no time to read arguments at this moment. As I understand the choice of this important class, it's to highlight the fact the buttons have a strong impact on the database (and it makes sense). My opinion changed over time but today I would keep the `important` class only for the main action of the page; ideally, only one per page. 

Yet as I said, I prefer consistency over the application and important buttons are not used this way so there's no need to change.

p/scripts/main.js
```diff
@@ -1215,6 +1215,72 @@ function init_notifications() {
...
+
+	popup_txt = document.getElementById('popup-txt');
+
+	//Configure close button.
+	document.getElementById('popup-close').addEventListener('click', function (ev) {
+  		closePopup();
+  	});
+
+  	//Configure close-on-click.
+	window.onclick = function(event) {
```
marienfressinaud: `addEventListener`?

app/Controllers/feedController.php
```diff
@@ -647,6 +647,150 @@ public function deleteAction() {
...
+		//Configure.
+		$this->view->fatalError = '';
+		$this->view->selectorSuccess = false;
+		$this->view->htmlContent = '';
+
+		$this->view->_layout(false);
+
+		//XXX 'script-src' allow possible script inlined in the previewed page to be executed.
+		//If we want to allow only our script, perhaps we should create a file, and reference it, instead of having it in-line.
+		//This view will be in an iframe, without any context, is it that dangerous? What a script can do?
```
marienfressinaud: Actually the risk is pretty low since we sanitize the content. I would avoid this anyway since it doesn't seem required (if my previous suggestion works :)). Here I would reuse the policies of the index page to be consistent:

```php
		$this->_csp([
			'default-src' => "'self'",
			'frame-src' => '*',
			'img-src' => '* data:',
			'media-src' => '*',
		]);
```

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,71 @@
...
+		<?php
+				}
+		?>
+		<center>
+			<div>
+				<input type="checkbox" id="freshrss_rendered" name="freshrss_rendered" checked>
+				<label for="freshrss_rendered"><?= _t('sub.feed.selector_preview.show_rendered') ?></label>
+
+				<input type="checkbox" id="freshrss_raw" name="freshrss_raw">
+				<label for="freshrss_raw"><?= _t('sub.feed.selector_preview.show_raw') ?></label>
```
marienfressinaud: I would use a radio button to switch between the two views. Maybe not perfect either, but the problem of checkboxes is:

- if you uncheck both, the screen is empty
- if both are checked, the source code is not immediately visible and user might think it doesn't work

marienfressinaud (commented):
I made few comments, hopefully it'll not require a lot of changes. I would have done it quite differently but not that much and your contribution is highly appreciated! :) I tried to stay focus on what matters to me (when it didn't, I explicitly said change is not needed). A lot has already been discussed and I have to admit I'm not motivated enough to dig into the 107 comments (maybe one of the most discussed PR ^^) so please excuse me if my points were already discussed.

The last thing that would require more work in my opinion is the design of the popup: it's a bit raw like this, but I can work on it if you want (I can make a different PR so we can merge yours and profit of your work).

app/layout/layout.phtml
```diff
@@ -66,5 +66,18 @@
...
 </div>
+
+<!-- Popup-->
+<div id="popup">
+	<div id="popup-content">
+		<div id="popup-close" class="popup-row">&times;</div>
+		<div id="popup-txt" class="popup-row"></div>
+		<div id="popup-iframe-container" class="popup-row">
+			<div id="popup-iframe-sub">
+				<iframe id="popup-iframe" frameborder="0" sandbox="allow-scripts"></iframe>
```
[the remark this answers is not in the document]
aledeg: Maybe we need to use the new security header
https://scotthelme.co.uk/a-new-security-header-feature-policy/

javerous:
> I made few comments, hopefully it'll not require a lot of changes. I would have done it quite differently but not that much and your contribution is highly appreciated! :)

Thanks. The point is that I'm not a web dev at all. Last time, more than 15 years ago, it was to code my personal little website with old PHP, HTML 4, and table everywhere, so, well…

> I tried to stay focus on what matters to me (when it didn't, I explicitly said change is not needed).

Okay, I will pass on your review today or this week-end. Thanks !

> A lot has already been discussed and I have to admit I'm not motivated enough to dig into the 107 comments (maybe one of the most discussed PR ^^) so please excuse me if my points were already discussed.

Yeah ! I don't know how we ended up in this situation ^^'

> The last thing that would require more work in my opinion is the design of the popup: it's a bit raw like this, but I can work on it if you want (I can make a different PR so we can merge yours and profit of your work).

Yes, I agree, I'm not that much a UI guy, and I discovered that I deeply hate CSS, so yes, if you are motivated to enhance things, it's perfectly OK for me ;)

app/i18n/ru/sub.php
```diff
@@ -34,7 +34,7 @@
 		),
 		'clear_cache' => 'Always clear cache',	// TODO - Translation
 		'css_help' => 'Retrieves truncated RSS feeds (caution, requires more time!)',	// TODO - Translation
-		'css_path' => 'Articles CSS path on original website',	// TODO - Translation
+		'css_path' => 'Articles CSS selector on original website',	// TODO - Translation
```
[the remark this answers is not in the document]
javerous: Yes, it's a problem I pointed with the translation tool integrated in the Makefile. Once you propagated a translation, it's hard to updates the TODO-ed ones.

Fixed.

app/layout/layout.phtml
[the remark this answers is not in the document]
javerous: Just by curiosity, why do you prefer not to use iframe ?

So yes, it was for simplicity + I thought it would be better to don't interact with current page style. I was not sure of what was covered by `sanitizeHTML`, btw.

And it's why I created a double mechanism for the Popup: one to show content by iframe, and one to show content by changing a simple "p" node (`<div id="popup-txt" class="popup-row"></div>`), so the pop-up is a bit more flexible. Not sure it's the right way to do it, but…

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,71 @@
+<!DOCTYPE html>
+<html>
+	<body>
+		<?php
+			if ($this->fatalError != '') {
+		?>
```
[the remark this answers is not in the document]
javerous: Yes, it seems better than what I did. I wanted to respect indentation "by-area", but it make the whole things more difficult to read.

Fixed.

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,71 @@
...
+		<hr>
+
+		<div id="freshrss_rendered_view"> <?= $this->htmlContent ?></div>
+		<hr id="freshrss_separator_view" hidden>
+		<pre id="freshrss_raw_view" hidden> <?= htmlspecialchars($this->htmlContent) ?></pre>
+		<?php
+			}
+		?>
+
+	<script>
```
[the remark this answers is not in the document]
javerous: Hmm no, it's not going to work, because this page is a "virgin" view (no standard scripts included) in an iframe 😞

app/views/helpers/feed/update.phtml
```diff
@@ -274,7 +274,11 @@
 		<div class="form-group">
 			<label class="group-name" for="path_entries"><?= _t('sub.feed.css_path') ?></label>
 			<div class="group-controls">
-				<input type="text" name="path_entries" id="path_entries" class="extend" value="<?= $this->feed->pathEntries() ?>" placeholder="<?= _t('gen.short.blank_to_disable') ?>" />
+				<div class="stick">
+					<input type="text" name="path_entries" id="path_entries" class="extend" value="<?= $this->feed->pathEntries() ?>" placeholder="<?= _t('gen.short.blank_to_disable') ?>" />
+					<a id="popup-preview-selector" class="btn" href="<?= _url('feed', 'contentSelectorPreview', 'id', $this->feed->id(), 'selector', 'selector-token') ?>"><?= _i('look') ?></a>
+				</div>
+				<br/>
```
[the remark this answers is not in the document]
javerous: Why not, but my intend was, in fact, to open another PR once this one would be validated (?) to fix other fields and make things coherent. Because the current results (help on the same line) can look really broken:
Screenshot 2020-01-30 at 22 12 46

Note, thought, that some existing fields already put the help on a new line:
Screenshot 2020-01-30 at 22 13 07

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
javerous: - I guess the br is the same than the previous comment ?
- For the `important`, it was a suggestion from @Frenzie

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
javerous: Here is the result without the `<br/>` :s

Screenshot 2020-01-30 at 22 50 13

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
javerous: Here is the result without the `<br/>` :s

Screenshot 2020-01-30 at 22 50 18

p/scripts/main.js
javerous: Ah yes. As said, I'm very old-school for web dev :s

Fixed.

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
javerous: Fixed, by the way ;)

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
javerous: Fixed, by the way ;)

app/views/feed/contentSelectorPreview.phtml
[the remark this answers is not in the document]
Frenzie: You say that as if you don't have full control over this "virgin" view. :-P

app/views/feed/contentSelectorPreview.phtml
javerous: Yes, actually it's what I thought about the two box unchecked. I wanted to add some code to force to have at least one box checked, but laziness…

And I agree this UI is a bit raw. I'm not skilled enough to do something better (so my comment about your proposal to take care of this after merging).

Fixed.

app/layout/layout.phtml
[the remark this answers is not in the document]
Frenzie: > I thought it would be better to don't interact with current page style.

Wait, that was a conscious choice? I thought of it as a for now perfectly functional (and possibly for years to come) but I'd just style it with the existing .flux_content .content classes.

Frenzie:
PS I agree with most or possibly all @marienfressinaud wrote but playing a bit with the UI in a follow-up PR or two is small fry.

app/layout/layout.phtml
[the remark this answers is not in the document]
javerous: The simplicity was an important point in my choice too ;p

But I guess it would simplify the javascript problem. And I see that `XMLHttpRequest` is already used in `init_slider_observers`, so I guess I can guet inspiration from there...

But how do you deal with `.flux_content .content` when there is an error ?

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
Frenzie: Wait, can you link me to that? It seems unlikely that I would suggest !important.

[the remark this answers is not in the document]
Frenzie: Oh, .important, not !important. Never mind.

app/views/feed/contentSelectorPreview.phtml
Frenzie: Please never write code to make a checkbox behave like a radiobox. ;-)

app/views/helpers/feed/update.phtml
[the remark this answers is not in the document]
javerous: 😅

app/views/feed/contentSelectorPreview.phtml
javerous: Well, I agree, but it was not exactly the behavior of radiobox, as it would have been possible to select the two box at the same time. The idea was just to make impossible to deselect the two altogether.

Frenzie: Oh, just add a `disabled` attribute to the one. Just a couple lines of JS. :-P

app/layout/layout.phtml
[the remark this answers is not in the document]
Frenzie: We can probably use fetch now btw.

https://caniuse.com/#feat=fetch

Don't style the error with the same class? Not sure I quite follow.

app/Controllers/feedController.php
```diff
@@ -647,6 +647,150 @@ public function deleteAction() {
...
+		$this->_csp([
+			'default-src' => "'self'",
+			'script-src' => "'unsafe-inline'",
```
Alkarex: If there is no good reason for this line, this permission should be removed. Check @marienfressinaud 's comment above.

app/Controllers/feedController.php
```diff
@@ -647,6 +647,150 @@ public function deleteAction() {
...
+		if (!$content_selector) {
+			$this->view->fatalError = _t('feedback.sub.feed.selector_preview.selector_empty');
+			return;
+		}
+
+		//Check Feed ID validity.
+		$entryDAO = FreshRSS_Factory::createEntryDao();
+		$entries = $entryDAO->listWhere('f', $feed_id);
+
+		if (count($entries) == 0) {
```
```suggestion
		if (empty($entries)) {
```
This also catches the case when `listWhere` might return `false`

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html>
+	<body>
+		<?php if ($this->fatalError != '') { ?>
+			<p class="warning"><?= $this->fatalError ?></p>
+		<?php } else { ?>
+			<?php if ($this->selectorSuccess === false) { ?>
+				<p class="warning"><?= _t('feedback.sub.feed.selector_preview.no_result') ?></p>
+				<hr>
```
Alkarex:
```suggestion
				<hr />
```
Favour strict syntax

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+				<div>
+					<input type="radio" id="freshrss_rendered" name="freshrss_type" checked>
+					<label for="freshrss_rendered"><?= _t('sub.feed.selector_preview.show_rendered') ?></label>
+
+					<input type="radio" id="freshrss_raw" name="freshrss_type">
+					<label for="freshrss_raw"><?= _t('sub.feed.selector_preview.show_raw') ?></label>
+				</div>
+			</center>
+
+			<hr>
```
Alkarex:
```suggestion
			<hr />
```

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+		<?php if ($this->fatalError != '') { ?>
+			<p class="warning"><?= $this->fatalError ?></p>
+		<?php } else { ?>
+			<?php if ($this->selectorSuccess === false) { ?>
+				<p class="warning"><?= _t('feedback.sub.feed.selector_preview.no_result') ?></p>
+				<hr>
+			<?php } ?>
+			<center>
+				<div>
+					<input type="radio" id="freshrss_rendered" name="freshrss_type" checked>
```
Alkarex:
```suggestion
					<input type="radio" id="freshrss_rendered" name="freshrss_type" checked="checked" />
```

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+					<label for="freshrss_rendered"><?= _t('sub.feed.selector_preview.show_rendered') ?></label>
+
+					<input type="radio" id="freshrss_raw" name="freshrss_type">
```
Alkarex:
```suggestion
					<input type="radio" id="freshrss_raw" name="freshrss_type" />
```

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+<html>
+	<body>
+		<?php if ($this->fatalError != '') { ?>
+			<p class="warning"><?= $this->fatalError ?></p>
+		<?php } else { ?>
+			<?php if ($this->selectorSuccess === false) { ?>
+				<p class="warning"><?= _t('feedback.sub.feed.selector_preview.no_result') ?></p>
+				<hr>
+			<?php } ?>
+			<center>
```
Alkarex: Please consider replacing (`<center>` is obsolete) https://developer.mozilla.org/docs/Web/HTML/Element/center

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+
+					<input type="radio" id="freshrss_raw" name="freshrss_type">
+					<label for="freshrss_raw"><?= _t('sub.feed.selector_preview.show_raw') ?></label>
+				</div>
+			</center>
+
+			<hr>
+
+			<div id="freshrss_rendered_view"> <?= $this->htmlContent ?></div>
+			<pre id="freshrss_raw_view" hidden> <?= htmlspecialchars($this->htmlContent) ?></pre>
```
Alkarex:
```suggestion
			<pre id="freshrss_raw_view" hidden="hidden"><?= htmlspecialchars($this->htmlContent) ?></pre>
```

app/views/feed/contentSelectorPreview.phtml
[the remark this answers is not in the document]
Alkarex: Yes, moving this script so that is not inline would be good (and we should try to prevent inline scripts with CSP whenever possible).

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+		<?php } ?>
+
+	<script>
+		var rendered_node = document.getElementById("freshrss_rendered");
```
Alkarex:
```suggestion
		const rendered_node = document.getElementById('freshrss_rendered');
```
And the same for following variables

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -0,0 +1,55 @@
...
+		var raw_node = document.getElementById("freshrss_raw");
+
+		var rendered_view = document.getElementById("freshrss_rendered_view");
+		var separator_view = document.getElementById("freshrss_separator_view");
+		var raw_view = document.getElementById("freshrss_raw_view");
+
+		rendered_node.addEventListener("click", function(ev){
```
Alkarex:
```suggestion
		rendered_node.addEventListener('click', function (ev) {
```
Minor whitespace preference (case of anonymous function). Same below.

app/views/feed/contentSelectorPreview.phtml
Frenzie: Hm, this must have been added more recently; I distinctly recall saying to use `let` and `const` instead of `var`. :-)

app/Controllers/feedController.php
```diff
@@ -647,6 +647,150 @@ public function deleteAction() {
...
 	}
 
+	/**
+	 * This action force clears the cache of a feed.
+	 *
+	 * Parameters are:
+	 *   - id (mandatory - no default): Feed ID
+	 *
+	 */
+	public function clearcacheAction() {
```
```suggestion
	public function clearCacheAction() {
```
And corresponding calls

app/views/feed/contentSelectorPreview.phtml
javerous: They are wrong there ? https://www.w3schools.com/tags/att_input_checked.asp

It's the first time I see `checked="checked"`…

app/views/feed/contentSelectorPreview.phtml
javerous: Same comment… https://www.w3schools.com/tags/att_hidden.asp

app/views/feed/contentSelectorPreview.phtml
Alkarex: Not wrong, check again your link in the XHTML section, a note about attribute minimisation. The form I suggest is the full, non-minimised one, which happens to be XML compatible and more strict

app/layout/layout.phtml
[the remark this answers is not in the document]
javerous: So, I tried to remove the iframe and use this weird `fetch` API to put the result in the innerHTML of a div.

For some reason, it seems to works… (JavaScript is a language which doesn't make any sense to me, I don't know how all of you are able to understand what you are doing with it!).

But I feel that the result is worst than with an iframe, in fact. I have tons of errors because of CSP (not able to load remote images, not able to load inline images, not able to load movies, etc.)

The problem is because this popup content (i.e. part of the article) now "run" in the CSP of the "subscription" controller (i.e. default CSP) which is very restricive. I'm not sure we want to open it more just for the preview...

The iframe had the advantage to better control the request / result context, and limit the CSP opening strictly to the preview content.

Screenshot 2020-02-01 at 19 47 37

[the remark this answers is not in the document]
Frenzie: Ftr, I don't think there's anything wrong with iframes in this context but I'm open to arguments to the contrary. But not "interacting" with the page style struck me as a bug, not a feature. :-)

p/scripts/preview.js
```diff
@@ -20,16 +20,16 @@ function update_ui() {
...
+	raw_node = document.getElementById('freshrss_raw');
+	raw_view = document.getElementById('freshrss_raw_view');
 
 	rendered_node.addEventListener('click', function (ev) {
 		update_ui();
 	});
 
-	raw_node.addEventListener("click", function (ev) {
+	raw_node.addEventListener('click', function (ev) {
 		update_ui();
```
Frenzie: Is it actually necessary for this to be an anonymous function?

javerous: That's a… good question ? 🤷‍♂️

Frenzie: In that case, I'd say simplify it to `'click', update_ui` (careful, no `()`) and see if it works. Otherwise that's a lot of wasted lines. ;-)

marienfressinaud:
What's the status of this PR? The few remaining remarks I see don't look too complicated (nor really too important). I'll take time to improve the style once I'm sure you're ready (doing it in the same PR seems better).

javerous:
@marienfressinaud Oh, sorry, I missed the last comments. I will do it this evening or this week-end.

For CSS, I'm not that much motivated :s

Can you edit directely in my repo ? I'm not sure if GitHub allow that btw…

marienfressinaud:
If you didn't uncheck the option when you created your PR, I should be able :) If I can't, don't worry, I'll find another way!

javerous:
I think I saw the option, and checked it (or at least didn't unchecked) so all of you can edit, but it's more related to the PR edit itself (like comments, title, etc.), no ?

In doubt, I added you as a collaborator on my fork. Feel free to commit directly from there, if it's OK for you, I think it will be easier than pointing me each lines I need to edit :s

Be assured I will not remove the PR and the branch.

marienfressinaud:
I'll work on it tomorrow if you've finished on your side :)

javerous:
Okay, I think that this time, it's all done ! (the tests are failing because of some time-out, I don't think it's related to my last changes…)

cc @marienfressinaud

app/views/feed/contentSelectorPreview.phtml
```diff
@@ -1,30 +1,36 @@
 <?php FreshRSS::preLayout(); ?>
 <!DOCTYPE html>
-<html class="preview_background">
+<html class="preview_background" lang="<?= FreshRSS_Context::$user_conf->language ?>" xml:lang="<?= FreshRSS_Context::$user_conf->language ?>">
 	<head>
 		<?= self::headStyle() ?>
-		<script src="<?= Minz_Url::display('/scripts/preview.js?' . @filemtime(PUBLIC_PATH . '/scripts/main.js')) ?>"></script>
+		<script src="<?= Minz_Url::display('/scripts/preview.js?' . @filemtime(PUBLIC_PATH . '/scripts/preview.js')) ?>"></script>
```
javerous, line 6: Oh ! It's why my script had difficulties to be re-downloaded by my browser :/

marienfressinaud:
I fixed some minor details I found while working on the style. The improved popup looks like this (don't pay attention to the small scrollbar glitch at the bottom of the popup, it was caused by the devtools):

Capture d’écran - 2020-02-07 à 18 15 55

Capture d’écran - 2020-02-07 à 18 16 12

I'm not totally satisfied with the controls box (maybe too grey?), but I think it's enough for now.

app/views/feed/contentSelectorPreview.phtml
marienfressinaud: That would be a good reason ^^ We could imagine a system that would calculate the `filemtime` automatically to avoid such problems (e.g. by introducing a `Minz_Url::static('/scripts/preview.js')` method), but it's very low in my priorities

javerous:
@marienfressinaud Yes, I was testing the changes right now. It's definitely better!

For the gray, I don't have special advice, it looks good to me (I'm not accustomed to see source code with white text on a dark background, but I think it's the most common).

[Edit] Oh, you were speaking about the gray of the **controls box**. It's perfect from my point of view.

marienfressinaud:
I didn't style specifically the source code, just reused what was done in the rest of the application :) If we change the style here, we should change it for code in the articles.

marienfressinaud:
Fixed for the `dir="auto"` :+1:  Concerning the language, the same issue happens in the rest of the application. It will be at least true for the error messages and the controls ^^

Frenzie:
Not exactly, the rest of the application has an actual application going on. :-P (But fair enough.)

marienfressinaud:
If there's no new comments, I'll merge this PR before the end of the day :)

Frenzie:
@marienfressinaud Like I said before, I could nitpick various things but I think it's good enough to merge. ;-)

javerous:
🎉

Alkarex:
Sorry for the late feedback. This is a great work :-)
One suggestion of improvement (I have not read the whole thread): If I understand correctly, the preview mode test against the Web site's home page, right?
However, the home page might not be the page on which to test the CSS path. An improvement could be to provide to either: a) expose a field on which to type an example of article URL; or b) automatically test on the latest article retrieved.

javerous:
@Alkarex Thank you, it was a bit "laborious" (I don't know if it translates well), but mainly because I'm not a web developer ;p

> One suggestion of improvement (I have not read the whole thread): If I understand correctly, the preview mode test against the Web site's home page, right?

No. I followed the suggestion from @Frenzie : I do the preview on the last article (i.e. the last retrieved) of the feed, so it should be okay (it's your "b" proposal, I think). I used the feature a bunch of time since, and it's very practical (okay, I'm perhaps not very impartial, but it helped me for sure).

A possible enhancement would be to be able to update the css path while we are inside the preview pop-up, but it seems too complex for me ;p

Alkarex:
Great, testing more carefully on another feed indeed shows that it works as intended :-)

Alkarex:
Slight refactoring in https://github.com/FreshRSS/FreshRSS/pull/4291

PR #2794, opened 2020-02-06
Improve login and registration pages

marienfressinaud:
I noticed a bunch of issues on the login/registration forms, causing a bad experience. I tried to address most of them in this PR.

- Keep the user on login page on failure 
- Show an error if username already exists
- Check the password format in the backend
- Return a better message if username is invalid
- Add a title to the login page
- Improve look of login and register pages

The last point is still WIP because I'll have to update all the themes. I'd like your feedback before going forward. It's nothing fancy: it mainly adds more space to the forms (so they can breathe and labels stay on a single line). I also changed the look of the "cancel" button back to a real link since it's its behavior. I'll change its label to "Back to home" in a future PR which will add a proper home page.

Before (tell me if you want better quality images):

before

After:

after

1. Log in with a wrong password, check you stay on the login page and an error appears
2. Check the title of the login page is prepended by "Login" (it was not set before)
3. Try to register with an exiting username, check you stay on registration page and you get a correct error
4. Try to register with an invalid username (e.g. with a space inside), check you get an explicit message
5. Try to register with a bad password (less than 7 characters), check you get an error

Note you might need to edit the DOM to remove frontend restrictions on username and password formats.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -109,6 +109,7 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 	public function formLoginAction() {
 		invalidateHttpCache();
 
+		Minz_View::prependTitle(_t('gen.auth.login') . ' · ');
 		Minz_View::appendScript(Minz_Url::display('/scripts/bcrypt.min.js?' . @filemtime(PUBLIC_PATH . '/scripts/bcrypt.min.js')));
 
 		$conf = Minz_Configuration::get('system');
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -122,7 +122,10 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 
 			$conf = get_user_configuration($username);
 			if ($conf == null) {
-				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
+				Minz_Request::bad(
+					_t('feedback.auth.login.invalid'),
+					array('c' => 'auth', 'a' => 'login')
+				);
 				return;
 			}
 
@@ -151,7 +154,10 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 				                  ' user=' . $username .
 				                  ', nonce=' . $nonce .
 				                  ', c=' . $challenge);
-				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
+				Minz_Request::bad(
+					_t('feedback.auth.login.invalid'),
+					array('c' => 'auth', 'a' => 'login')
+				);
 			}
 		} elseif (FreshRSS_Context::$system_conf->unsafe_autologin_enabled) {
 			$username = Minz_Request::param('u', '');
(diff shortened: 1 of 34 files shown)


app/i18n/de/user.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 return array(
+	'already_exists' => 'The user %s already exists.',	// TODO - Translation
```
Frenzie: User or username?

(And possibly: exists or taken?)

app/i18n/cz/user.php
```diff
@@ -38,4 +38,7 @@
 			'invalid' => 'You must accept the Terms of Service to be able to register.',	// TODO - Translation
 		),
 	),
+	'username' => array(
+		'invalid' => 'The username is invalid.',	// TODO - Translation
```
Frenzie, line 41: I've noticed what's presumably a security practice of not visibly differentiating between invalid user & password. (As a user I'd find it annoying; just putting it out there for the moment.)

Frenzie (commented):
The pictures are a bit too low-res imo. No need for new ones; they're legible (if barely :-P).

It says "maximum…" (lowercase m) and "At least…" (uppercase A).

> Keep the user on login page on failure

I should've done that years ago, but it only ever seems to bother me when I'm on my phone far away from a proper PC or some such and then I forget.

app/i18n/cz/user.php
marienfressinaud: This message is used during registration, not during login :) The message during login is actually "Login is invalid" which is maybe not good either. I'm not sure about the security impact of clearly state which of the username or the password is wrong, but I agree with you that it's annoying from a user perspective. I'll ask some friends who work in security for their insight.

marienfressinaud: Considering that an attacker can test the existence of a username by trying to create an account, I would say there are no reasons to hide the exact information

Frenzie: That's a great point. :+1:

Alkarex (approved):

Alkarex:
Good for me

Frenzie (approved):
lgtm too, I only had those minor spelling nitpicks :+1:

app/Controllers/authController.php
```diff
@@ -151,7 +155,10 @@ public function formLoginAction() {
 				                  ' user=' . $username .
 				                  ', nonce=' . $nonce .
 				                  ', c=' . $challenge);
-				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
+				Minz_Request::bad(
```
Alkarex, line 158: The new code returns 200, and not 403 anymore, right? It could be good to reintroduce the 403 at some point.
Indeed, error codes are used by other systems (such as fail2ban) e.g. to detect attacts

marienfressinaud: To be exact it returns a 302. I think there's work to do on how we handle responses, and errors in particular but I didn't want to make this PR too complicated. Here I wanted to keep the users on the same URL (`?c=auth&a=login`) for consistency and because there's a small condition that would be skipped otherwise (nothing dramatic).

It wouldn't be too complicated to rework this part, though I'm not sure how important it is.

Alkarex: https://github.com/FreshRSS/FreshRSS/issues/2901

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/2903

PR #2808, opened 2020-02-25
No pre for enclosure-description

### LINKED ISSUE #2807: <media:description> tag transforms to <pre class="enclosure-description">
Feeds from the Toronto Star like this one: http://www.thestar.com/content/thestar/feed.RSSManagerServlet.articles.news.gta.rss 

caption photos with a &lt;media:description> tag. FreshRSS translates that to &lt;pre class="enclosure-description"> which is fine for a regular browser, but is problematic when I read my feeds via FeedMe on my phone.

The &lt;pre> is rendered in one big long line causing the caption to scroll horizontally way off the right edge of the screen. This isn't ideal because swipe left is the next article gesture so it takes several swipes to get to the end of the caption and then one more to get to the next article.

I have included a screenshot showing the caption under the photo going off the edge of the screen.
Screenshot_20200225-120505708

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2807
In browsers, the styling `white-space:pre-line` will be applied by CSS,
but no styling will be done for clients via API.

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -1135,7 +1135,7 @@ input:checked + .slide-container .properties {
 	margin-left: .8em;
 }
 
-pre.enclosure-description {
+.enclosure-description {
 	white-space: pre-line;
 }
 
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -401,7 +401,7 @@ class FreshRSS_Feed extends Minz_Model {
 						$content .= $enclosureContent;
 
 						if ($enclosure->get_description() != '') {
-							$content .= '<pre class="enclosure-description">' . $enclosure->get_description() . '</pre>';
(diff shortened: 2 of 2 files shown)


p/themes/base-theme/template.css
```diff
@@ -1135,7 +1135,7 @@ input:checked + .slide-container .properties {
 	margin-left: .8em;
 }
 
-pre.enclosure-description {
+.enclosure-description {
 	white-space: pre-line;
```
Frenzie, line 1139: Does this style still make sense if you're just using a regular `<p>`?

Frenzie (approved):
lgtm

p/themes/base-theme/template.css
Alkarex: I think this can be applied to any block, including div, p, etc.

Frenzie: Sure, I just meant to double check if it was providing added value. :+1:

PR #2809, opened 2020-02-26
Fix wrong foreach in applyFilterActions

shn7798:
$filterAction->actions()  returns Array.

Current works fine because the php switch behavior.
See: https://stackoverflow.com/questions/2259266/0-in-switch-case


Alkarex:
Thanks 👍

PR #2813, opened 2020-02-29
Fallback to GUID when entry title is empty

Alkarex:
When entries have an empty title, our UI was a bit broken. In this
(rare) case, we revert to GUID, which is always non-empty.


Frenzie (approved):
Seems like it'd look a bit weird, but no weirder than empty I suppose. :-)

Alkarex:
Example before:
image
Example after:
image

Alkarex:
@Frenzie It is often a number or a URL

PR #2821, opened 2020-03-08
Option to show/hide favicons

### LINKED ISSUE #2819: High number of requests coming from FreshRSS gets me blocked from hosting
Hi,
I don't have a clear view of what is going on, but my FreshRSS install gets me temporarily blocked by the hosting service due to a high number of requests (more than 1000 in 10 secs).
The requests are in the form: "GET /f.php?xxxxxx HTTP/1.1"
Where xxxx is different every time, but for the rest it's the same.
This seem to happen when I'm not even refreshing the feeds. I get the feeling it has to do with the favicons of the feeds / articles because when I open FreshRSS they are not there and they appear progressively.
But could it be that it's checking the favicon of every article (not feed)? I don't have more than 1000 feeds in my instance, so I can only assume it's linked to articles?
If I can do something to help pinpoint the problem, let me know.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2819
New option `'show_favicons' => true,` or `'show_favicons' => false,` in `./data/users/alice/config.php` to show or hide favicons

### DIFF
--- a/app/layout/aside_feed.phtml
+++ b/app/layout/aside_feed.phtml
@@ -94,7 +94,8 @@
 						<a class="dropdown-toggle" data-fweb="<?= $feed->website() ?>"><?= _i('configure') ?></a>
 						<?php /* feed_config_template */ ?>
 					</div>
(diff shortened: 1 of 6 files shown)


Frenzie (approved):
Seems fine as an option, not sure I quite agree with the word "fix" given the misconfigured cache though. ;-)

ltnzr:
Confirm this fixed #2819 for me. Agree it might be a misconfiguration of the server in the first place, but in my case, I don't have access to the server configuration, so this was a nice solution. Thanks again!

PR #2837, opened 2020-03-22
New core-extensions to allow Docker volumes for third-party extensions

### LINKED ISSUE #2650: Extension not working
CustomCSS is not working when I load it to the extensions folder. I know after the major version update it was recommended to disable the extensions, which I did, but now I would love to have my extensions back. Are all of these working because at my end they do not load...

Alkarex:
#Fix https://github.com/FreshRSS/FreshRSS/issues/2650
Split our extensions directory into two:
1) Core extensions shipped with FreshRSS in `./lib/core-extensions/`
2) Third-party extensions modified by end-users in `./extensions/` which can easily be mounted as a Docker volume

Additional testing needed.


Frenzie (approved):
Makes sense.

Alkarex:
Arg, some extensions such as CustomCSS are using `EXTENSIONS_PATH` in their code
https://github.com/FreshRSS/Extensions/blob/ade2979857e91d838c1835f773797fc70cd7145b/xExtension-CustomCSS/extension.php#L24-L33

Alkarex:
We should start pushing the use of `./data/extensions-data/` as introduced in https://github.com/FreshRSS/FreshRSS/pull/1547 to avoid requiring write access in the extensions folder

Alkarex:
I have re-introduced `EXTENSIONS_PATH` as a legacy constant for back-compatibility, while waiting for updating the extensions to avoid mixing code and data.

Alkarex:
Tested, also with Docker. Existing extensions should not be affected.

Alkarex:
Merging a bit fast, so that I can send another PR depending on it. But feedback still welcome :-)

Issue #2836, opened 2020-03-22
Mark Above As Read


yjfguhd:
I'm experimenting switching from [TT-RSS](https://tt-rss.org/) to FreshRSS, and I miss the "Mark articles above as read" feature. You could scroll through articles, right-click on one and set all articles above the current one as read. I'm currently using "Mark articles as read while scrolling", but this has a problem when either the page is reloaded (by the browser), or the browser shuts down unexpectedly and restores the page when started up again. In both cases, Firefox has remembered how far down I was scrolling, and goes straight to the same scroll position when it loads the page. This has the effect of setting all the newly displayed articles above wherever the scroll view lands as read. That can be reproduced easily:

1. Enable "Mark article as read by scrolling"
1. Disable "Hide articles after reading"
1. For the purpose of this example, optionally set "Mark articles as unfolded by default", in order to facilitate a larger scrolling area
1. Go to a feed that has many articles
1. Scroll down a few pages worth, until the scroll bar is some distance from the top. Note the currently displayed article, and that all articles above have been marked as read as intended
1. Reload the page in the browser (ctrl-r)
1. Observe that the browser has scrolled down the same distance as before, and all new articles above, that you haven't yet seen, have been marked as read

That's expected behaviour due to the "Mark all articles as read" feature, and the behaviour of the browser, but it's a gotcha that I don't like (I have to then scroll up reading through articles that I haven't yet read, but have been marked as such). Having the ability to tell the site to not mark any articles as read until I say, "all the above", has been a good workaround, and I wish it was available in FreshRSS.

It's easy to scroll through articles; I don't want to have to specifically tell each article that it's been read, by clicking "Next", "Next", "Next"…

FreshRSS 1.15.3
Firefox 68.6.0esr

Alkarex:
Hello @yjfguhd 
I believe we can address your needs with a series of improvements:
* Mitigate the issue with the scrollbar when restoring the Web page. Already done. Could you please help testing https://github.com/FreshRSS/FreshRSS/pull/2842 ?
* We can relatively fast introduce a keyboard shortcut to mark previous articles as read.
  * Edit: done in https://github.com/FreshRSS/FreshRSS/pull/2843 and tests welcome
* Implement a customisable history for the number of read articles to keep while strolling, instead of keeping none / all. See https://github.com/FreshRSS/FreshRSS/issues/1695
[...]

PR #2842, opened 2020-03-23
Reset scroll

Alkarex:
Address some of https://github.com/FreshRSS/FreshRSS/issues/2836
This patch resets the scrollbar to the top when a FreshRSS page is
initialised.
Makes sure that when a FreshRSS page is restored in a Web browser
session, we do not hit the problematic situation when the content is
refreshed but the scrollbar moved back to where it was, potentially
erroneously marking as read many as read, of landing on the wrong focus.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1542,6 +1542,7 @@ function init_normal() {
 }
 
 function init_beforeDOM() {
+	document.scrollingElement.scrollTop = 0;
 	if (['normal', 'reader', 'global'].indexOf(context.current_view) >= 0) {
 		init_normal();
 	}


p/scripts/main.js
```diff
@@ -1542,6 +1542,7 @@ function init_normal() {
 }
 
 function init_beforeDOM() {
+	document.scrollingElement.scrollTop = 0;
```
Frenzie, line 1545: This actually works? Interesting. :-)

Frenzie (approved):

p/scripts/main.js
Alkarex: Apparently yes, but more tests welcome :-)

yjfguhd:
Work for me in Firefox 68.6.0esr!

PR #2843, opened 2020-03-23
Shortcut mark previous articles as read

Alkarex:
Edit: Updated by https://github.com/FreshRSS/FreshRSS/pull/2881

Address part of https://github.com/FreshRSS/FreshRSS/issues/2836
The shortcut <kbd>⇧ Shift</kbd>+<kbd>R</kbd> now marks previous articles as read.
The previous action of marking all articles as read is now achieved by <kbd>Alt ⎇</kbd>+<kbd>r</kbd>
So this is a slight breaking change, but I like the logic of the upward-facing <kbd>⇧ Shift</kbd> arrow to hint at previous articles, while <kbd>Alt ⎇</kbd> is more like "all", at least to my eyes :-)


app/i18n/en/conf.php
```diff
@@ -180,7 +180,7 @@
 		'reading_view' => 'Switch to reading view',
 		'rss_view' => 'Open RSS view in a new tab',
 		'see_on_website' => 'See on original website',
-		'shift_for_all_read' => '+ <code>shift</code> to mark all articles as read',
+		'shift_for_all_read' => '+ [Shift] to mark previous articles as read<br />+ [Alt] to mark all articles as read',
```
Frenzie: `<kbd>Shift</kbd>`?

app/i18n/nl/conf.php
```diff
@@ -180,7 +180,7 @@
 		'reading_view' => 'Schakel naar leesaanzicht',
 		'rss_view' => 'Open RSS-aanzicht in een nieuwe tab',
 		'see_on_website' => 'Bekijk op originale website',
-		'shift_for_all_read' => '+ <code>shift</code> om alle artikelen als gelezen te markeren',
+		'shift_for_all_read' => '+ [Shift] to mark previous articles as read<br />+ [Alt] om alle artikelen als gelezen te markeren',	// TODO - Translation
```
```suggestion
		'shift_for_all_read' => '+ [Shift] om voorgaande artikelen als gelezen te markeren<br />+ [Alt] om alle artikelen als gelezen te markeren',
```

app/i18n/en/conf.php
Alkarex: `<kbd>Shift</kbd>` was not looking so good, at least with my font rendering, but let me know if you prefer it.
I considered using ⇧ and ⎇ but did not find them self-explanatory enough. We could combine them with the text, though, like [Shift ⇧] and [Alt ⎇], what do you think?

Alkarex: It looks like [⇧ Shift] is more common (with the arrow first)

Frenzie: The font rendering looks fine to me, but without additional styles it'll just take the default monospace font face from the browser. Additionally it could be styled cutely like on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd

Screenshot_2020-03-23_21-54-15

I don't think I've ever seen ⎇ before, except for maybe something similar on a road sign. I doubt it'd be very helpful here.

Alkarex: image 🤓
For reference https://fr.wikipedia.org/wiki/Touche_alt

Alkarex:
The new styling also affects the rendering of `<kbd>` in articles, such as https://www.raspberrypi.org/blog/how-to-play-sound-and-make-noise-with-your-raspberry-pi/ , but I think it is a nice side-effect, e.g.:

image

app/i18n/en/conf.php
Frenzie: On my laptop keeb it's <kbd>shift ⇧</kbd>, btw.

Frenzie:
> but I think it is a nice side-effect

It's actually been a very minor mental to-do list item of mine for like two years now to style it like pretty much like that. :-P

Frenzie (approved):

PR #2844, opened 2020-03-23
More gitignore and dockerignore

Alkarex:
Add to gitignore and dockerignore some additional files created by the
different tests we have


.dockerignore
```diff
@@ -1,3 +1,6 @@
 **/.git/
+**/bin/
```
Alkarex, line 2: @marienfressinaud Do you need that when using `make` in your local Docker image?

aledeg (approved):

.dockerignore
marienfressinaud: I don't think so since I'm mounting the root directory in the container

PR #2845, opened 2020-03-23
Loader animation while waiting for first articles

Alkarex:
Especially useful when a slow search is being performed.
Takes advantage of https://github.com/FreshRSS/FreshRSS/pull/2588
Thanks to one new flush at the right time, the left column and top menu will be there, waiting for the articles to come. This avoids layout flashes (the left column used to take the full width while waiting for articles).
For optimal use, this assumes that the deployment has a proper buffering configuration for the whole pipeline, such as with our official Docker image (also when using Træfik as a reverse proxy in front). Might not work just as well with other Apache or nginx setups.

Here is how it looks like during a slow DB query.

Before:
image

After:
image


p/themes/Alternative-Dark/template.css
```diff
@@ -868,7 +868,21 @@ br {
...
+	position: relative;
+	z-index: 5;
+}
+
+#first_load {
+	min-height: 40px;
+	min-width: 40px;
+	position: fixed;
+	top: 200px;
+	width: calc(100% - 300px);
```
Frenzie: Note to self:
https://developer.mozilla.org/en-US/docs/Web/CSS/calc#Browser_compatibility

Frenzie (approved):
Did you also test on small screen (mobile)?

Alkarex:
> 
> 
> Did you also test on small screen (mobile)?

@Frenzie There is a slight glitch on mobile, which I will fix tonight (and BTW, in general, we should improve the top view of most themes in their mobile version).

Alkarex:
Now works well on mobile as well. CSS simplified a bit at the same time.

PR #2855, opened 2020-03-29
API return OK for an empty request

Alkarex:
Fix https://framateam.org/freshrss/pl/yozcezrnufno9eut3173d4wpry


Frenzie (approved):
lgtm, don't know if the behavior is any better or worse as such

PR #2861, opened 2020-03-30
Simplify CSS remove z-index

### LINKED ISSUE #2860: Dark Theme navigation buttons on mobile not working
https://github.com/FreshRSS/FreshRSS/tree/master/p/themes/Dark

they're visible, but they seem to be 'behind' the article list

previously, the buttons were on top of the article list.  now, when i try to press the 'next article' button, it doesnt hit the next article button, i hit the title of the article that's in the way of the button.

seriously love freshrss i use it every day!  thanks so much for all the hard work!  please let me know what other information i can provide :)

Thanks!

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2860
Improvement of https://github.com/FreshRSS/FreshRSS/pull/2845

https://github.com/FreshRSS/FreshRSS/commit/22916d6a3ed259b5a5dca5cddb4a58433729384f#r38143737

### DIFF
--- a/p/themes/Alternative-Dark/template.css
+++ b/p/themes/Alternative-Dark/template.css
@@ -869,18 +869,13 @@ br {
 }
 
 /*=== "Load" parts */
-#stream {
-	position: relative;
-	z-index: 5;
-}
 
 /* Only matches while waiting for div#stream to be received */
 #first_load:nth-last-of-type(2) {
 	margin: -40px auto 0 auto;
 	height: 40px;
 	position: relative;
-	top: 250px;
-	z-index: 2;
+	top: 290px;
 }
 
 #load_more {
--- a/p/themes/Alternative-Dark/template.css
+++ b/p/themes/Alternative-Dark/template.css
@@ -871,7 +871,7 @@ br {
 /*=== "Load" parts */
 
 /* Only matches while waiting for div#stream to be received */
-#first_load:nth-last-of-type(2) {
+#first_load:nth-last-of-type(1), #first_load:nth-last-of-type(2) {
 	margin: -40px auto 0 auto;
 	height: 40px;
 	position: relative;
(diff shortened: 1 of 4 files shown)


p/themes/Alternative-Dark/template.css
```diff
@@ -869,18 +869,13 @@ br {
...
-}
 
 /* Only matches while waiting for div#stream to be received */
 #first_load:nth-last-of-type(2) {
 	margin: -40px auto 0 auto;
 	height: 40px;
 	position: relative;
-	top: 250px;
-	z-index: 2;
+	top: 290px;
```
marienfressinaud, line 878: I don't see why the position relative and `top` are needed either :thinking:

marienfressinaud: I think the problem comes from the margin `-40px`, you just have to use a positive value :)

Alkarex: For that to work, we would need to move `#first_load` to after `.nav_menu` in the DOM, which increases the risk of not being able to benefit from the load effect, because `.nav_menu` is quite large and therefore significant in the buffering (especially important to avoid the layout flash addressed by https://github.com/FreshRSS/FreshRSS/pull/2845 )

Positive margin, worse:
image

Example without `.nav_menu` loaded yet:
image

marienfressinaud: I hadn't thought of that!

I think a cleaner option would be to use the CSS grid system, but it's less supported and would require more work, so let's go with yours for now :)

marienfressinaud (approved):

p/themes/Alternative-Dark/template.css
Alkarex: And for some reason, Chrome does not behave as I believe it should (not rendering the animation like Firefox or Edge). I am looking at it as part of https://github.com/FreshRSS/FreshRSS/pull/2909

Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/2909/commits/1e627dd5554cf80c3bac480ef261a4e6d5bfa093

Issue #2806, opened 2020-02-24
Filter actions for feeds seems not to work properly


xoxys:
The `Filter actions` input filed on the feed management page suggest to `Write one search filter per line`. This works for the first filter but after adding a second condition to a new line **all** future articles will be marked as read automatically.

I use filters as described here: https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html#filtering-articles

Alkarex:
Hello,
I cannot immediately reproduce the problem. You could please show a more precise example (with the filters you are using)?
Maybe related to the bug https://github.com/FreshRSS/FreshRSS/pull/2809 which has just been fixed in the /master branch

xoxys:
Sorry for the late response. I've switched my server to latest (/master). Lets see, if it works now. I'll report back :)

xoxys:
Hmm seems not to work as expected...
FreshRSS version: 1.15.4-dev
Feed: http://osbn.de/feed.rss
Filteractions:
```
author:bitblokes.de
author:soeren-hentzschel.at
```

**Expected result**
Only articles matching these filters will be automatically marked as readed.

**Current result**
All articles of this feed will be automatically marked as readed.

edequalsawesome:
I'm running into the same issue -- for the feed on Gizmodo (`https://gizmodo.com/index.xml`), for instance, I had the following filters active (in the Manage > Filter Actions area):

```
author:on+io9
author:on+lifehacker
author:on+Kinja+Deals
author:Kinja!+on+Kinja+Roundup
```

When I do a search in the feed for those items, only the things that I'd expect to be filtered show up, but after adding it to the Manage > Filter Actions area, it's marking all new articles as read. (It appears to be adding an extra line after the filters once I hit the Submit button -- not sure if that is relevant, but thought I'd throw it out there anyway!)

I've also tried adding the filters with `OR` between them on one line, but it's seeming to do the same thing.

I'm using version 1.15.4-dev as well. If there are any other details I can provide, please let me know!

Alkarex:
Alright, bug confirmed when using an `author` filter

Alkarex:
Should be fixed by https://github.com/FreshRSS/FreshRSS/pull/2863
Additional tests welcome

Alkarex:
P.S. Testing should also be easier now with this fix https://github.com/FreshRSS/FreshRSS/pull/2862

xoxys:
[...]

PR #2862, opened 2020-03-30
Allow feed to be actualized after being truncated

Alkarex:
Before, on had to wait for the cache to expire before being able to refresh a feed that had been truncated via the Web interface.
Now, one can "delete all articles" and hit "actualize" immediately after without problem.
Useful for testing filters, debugging e.g. https://github.com/FreshRSS/FreshRSS/issues/2806


Alkarex:
Useful for https://github.com/FreshRSS/FreshRSS/pull/2863

Frenzie (approved):
Sounds useful. :+1:

PR #2863, opened 2020-03-30
Fix action filter when filtering on author

### LINKED ISSUE #2806: Filter actions for feeds seems not to work properly

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2806
Bug: array instead of string


Alkarex:
Merging to ease user feedback

Issue #947, opened 2015-08-16
[improvement] full article content by fivefilter rules


DoumeDuSud:
Hello 
_Merci beaucoup pour tout le travail de dévellopement, Fresh RSS devient très sympa :+1:_

Is it possible to implement the same tool than use FiveFilter or Wallabag to take the full article content ?

It's more fine tune with the possibility to remove some part of the page. And most of all they are already a huge quantity of site config file (https://github.com/fivefilters/ftr-site-config ) and for my side I make my one rules when needed for my walabag 

(issue related to #199 & #785 )

marienfressinaud:
Of course it's something wanted. Note wallabag will use Graby, a fork of Full-Text RSS (https://github.com/j0k3r/graby).

I should work on it one day but never found time for it...

SecT0uch:
This feature would be awesome (in a User Friendly way).

Issue #785, opened 2015-02-05
full article content


Fshamri:
Hi,
i found it great application..... but does it support full article content extraction? 

Thanks

Alwaysin:
Hi,

Yes it does! The process is described in https://github.com/FreshRSS/FreshRSS/issues/199#issuecomment-26408922 but it is in french ; should be the time to translate it!

Waiting for someone to do so, you can play with the setting "Articles CSS path on original website" at the bottom of the configuration of a feed, if you have some knowledge you might be able to manage making it working :)

aledeg:
There is an other option which is to use [rss-bridge](https://github.com/sebsauvage/rss-bridge) project. This way, you can separate your concerns.
That's what I do for some feeds.
I think it is more flexible

Alwaysin:
Or also http://code.fivefilters.org/full-text-rss

mbe-financial-com:
Flym is one of the better-known RSS readers on Android, and I noticed it does a pretty good job of fetching full articles without any extra configuration, which would be pretty near impossible anyway for a source such as Google News that links to all sorts of publications. It's in these two pieces of code that basically apply a whole bunch of regexes:
https://github.com/FredJul/Flym/blob/master/app/src/main/java/net/frju/flym/service/FetcherService.kt (function `mobilizeAllEntries()`) and
https://github.com/FredJul/Flym/blob/master/app/src/main/java/net/frju/flym/utils/HtmlUtils.kt (function `improveHtmlContent()`).
Not pretty but fairly effective. I might try porting that, what do you think?

Frenzie:
That first one doesn't really seem to do anything at a glance, it just passes it on to `Readability4JExtended()`? An equivalent library will doubtless be somewhere in Wallabag, possibly too tightly integrated, as well as [here](https://github.com/andreskrey/readability.php) (andreskrey/readability.php).

Alkarex:
@mbe-financial-com Our current way to fetch full content is pretty obsolete, so new possibilities would be much welcome. As @Frenzie writes, we need to clarify whether those methods depend on third-party services or not, and we could even offer more than one option

sbrl:
Sounds like a readability thing you're talking about @mbe-financial-com. Firefox has a reader view built in - perhaps the code there could be used as a base? or maybe someone has already done this and published on packagist?

simon-wessel:
[...]

PR #2867, opened 2020-03-31
Fix fetch content for complex HEAD

### LINKED ISSUE #2864: Encoding issue with CSS selector
Hi !

For some time now I have a encoding issue with full text content retrival.
It only happens with France 3 Régions feeds, like [https://france3-regions.francetvinfo.fr/nouvelle-aquitaine/vienne/rss]()

The selector I use is : `.article-style`
Title looks okay but article content has wrong encoding.

image

Thanks for your help !

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2864

*  Fix `contentSelectorPreviewAction()` https://github.com/FreshRSS/FreshRSS/pull/2778 , which was not compatible with `yield` #2588
* Fix lib_phpQuery:
  * Replace `split()` by `explode()` for PHP7 compatibility https://php.net/str-split
  * Fix for documents with a `<head>` tag more complicated, like `<head profile="http://www.w3.org/1999/xhtml/vocab">`

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -753,14 +753,20 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		//Check Feed ID validity.
 		$entryDAO = FreshRSS_Factory::createEntryDao();
 		$entries = $entryDAO->listWhere('f', $feed_id);
+		$entry = null;
 
-		if (empty($entries)) {
+		//Get first entry (syntax robust for Generator or Array)
+		foreach ($entries as $myEntry) {
+			$entry = $myEntry;
+			break;
+		}
+
+		if ($entry == null) {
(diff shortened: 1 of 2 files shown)


Alkarex:
P.S. We should also patch this library to support HTML5 charset declaration like `<meta charset="UTF-8" />`

Alkarex:
Actually, we should find a replacement for `lib_phpQuery` , which is abandoned https://github.com/TobiaszCudnik/phpquery

Time to look at the related issues, e.g.
* https://github.com/FreshRSS/FreshRSS/issues/947
* https://github.com/FreshRSS/FreshRSS/issues/785

Any volunteer? :-)

Alkarex:
Merging a bit fast to allow our user to test and provide feedback https://github.com/FreshRSS/FreshRSS/issues/2864#issuecomment-606892871
Feedback still welcome :-)

PR #2873, opened 2020-04-03
[i18n] Update Occitan language

Quent-in:
New strings translated into Occitan

### DIFF
--- a/app/i18n/oc/admin.php
+++ b/app/i18n/oc/admin.php
@@ -122,7 +122,7 @@ return array(
 		'title' => 'Extensions',
 		'update' => 'Mesa a jorn disponibla',
 		'user' => 'Extensions utilizaire',
-		'version' => 'Version',	// TODO - Translation
+		'version' => 'Version',
 	),
 	'stats' => array(
 		'all_feeds' => 'Totes los fluxes',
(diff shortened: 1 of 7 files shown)


aledeg (approved):

aledeg:
Thank you for your contribution

app/i18n/oc/admin.php
```diff
@@ -122,7 +122,7 @@
 		'title' => 'Extensions',
 		'update' => 'Mesa a jorn disponibla',
 		'user' => 'Extensions utilizaire',
-		'version' => 'Version',	// TODO - Translation
+		'version' => 'Version',
```
Frenzie, line 125: So many difficult ones. ;-)

app/i18n/oc/user.php
```diff
@@ -30,15 +30,15 @@
 		),
 	),
 	'password' => array(
-		'invalid' => 'The password is invalid.',	// TODO - Translation
+		'invalid' => 'Lo senhal es invalid.',
```
Frenzie, line 33: Even invalid is straight Occitan? Wow. :+1:

Frenzie (approved):

Alkarex:
Thanks @Quent-in 👍 
Could you please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md ?

app/i18n/oc/user.php
Quent-in: Yes and even stronger, look:
The password is empty => Lo senhal es void
Easy to decode when you're developer, right? ^^

PR #2879, opened 2020-04-08
gitattributes: Enforce Unix line endings

Alkarex:
https://www.git-scm.com/docs/gitattributes
Disable automatic end-of-line conversion from Unix to Windows, and enforces LF (Unix line endings).
We do not have any file type in the project that requires Windows line endings (e.g. PHP works fine on Windows with Linux line endings), but the automatic conversion makes it annoying to work on the same git local repository from different systems (e.g. hybrid Linux / Windows development)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


Frenzie:
> We do not have any file type in the project that is sensitive to line
ending

I don't think that's accurate. Shellscripts at least are sensitive to it, possibly Makefiles too. Which makes Docker sensitive to it as well.

https://github.com/FreshRSS/FreshRSS/blob/bd036425aa495824b13a17891698a14d52ec93ba/Docker/entrypoint.sh#L5-L6

I'm reasonably certain that'd say `chmod -R g+w ./data/\r`, which would then say "no such file or directory."

But isn't the main point of automatic conversion to prevent giant diffs for no apparent reason, regardless if it literally causes scripts to fail or not?

Alkarex:
When there is automatic conversion, one will indeed observe a giant diff when looking at the same git directory from e.g. Linux or Windows. This patch solves that.
Examples of files, which cannot run on Windows with Unix endings are `.bat` files, but we do not have any. The other files such as Makefile and Docker should be fine with strict Unix endings, aren't they?
If there is any file extension that does require a Windows line ending on Windows platform, we can exclude it.

Alkarex:
> I'm reasonably certain that'd say `chmod -R g+w ./data/\r`, which would then say "no such file or directory."

@Frenzie This patch makes sure that we will only have Unix endings on all platforms, not the other way around

Frenzie (approved):
Ah, this enforces LF? Well excellent then :-)

Frenzie:
I misunderstood your text which made me think that `*` meant "don't do this" ^_^

Alkarex:
I have updated the title and the description

SuperSandro2000:
Isn't autocrlf invented for exactly that? Have CRLF locally and LF always in git.

Alkarex:
@SuperSandro2000 Even on Windows, the environment is increasingly hybrid:
* Development tools are being moved to WSL (Linux on Windows)
* Virtual machines or dual-boot sharing a partition
* Local Windows copy and remote Linux server

In all those cases, changing the line endings makes the workflows less efficient:
* It creates gigantic git diff and slow git operations
* Manually moving a file from one environment to the other is problematic
* Quick remote diffs (e.g. based on file size) do not work

On the other hand, having Unix endings everywhere works just fine. There are only a few cases for which Windows endings are necessary, such as `.bat` files, which we do not use.

PR #2880, opened 2020-04-08
Markdownlint

Alkarex:
I have started to use [markdownlint](https://github.com/DavidAnson/markdownlint/) to find typos in Markdown.
Let's discuss which rules to enforce / disable / customize


Frenzie (commented):
Are the changes in this PR the full extent of what would be caused by the proposed ruleset?

Alkarex:
@Frenzie No, there would be more fixes in other Markdown files - a work that is not done yet -. I should have written that indeed :-)
Some of the fixes will be for pure errors (e.g. wrong syntax for links, wrong depth for title, etc.), while some others are more about personal preferences

Alkarex:
I suggest we merge that and patch the rules later as needed. So far, the rules are not enforced anyway, and are just there to give warnings in compatible text editors. This has already helped me spot some typos, wrong syntax for links, etc.

PR #2887, opened 2020-04-10
Fix search input styling in Safari

### LINKED ISSUE #2885: Search input box display issue in Safari
The search input box doesn't display as expected in Safari, adding the following style rule fixes this issue:

```
.header > .item.search input {
	-webkit-appearance: none;
}
```
Screen Shot 2020-04-09 at 10 28 14 pm
Screen Shot 2020-04-09 at 10 27 23 pm

kinoushe:
Closes #2885 

- Disable browser-native search input styling in Safari to be consistent with other web browsers

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -573,7 +573,9 @@ a.btn {
 }
 
 .header > .item.search input {
-	-webkit-appearance: none; /* csslint allow: -webkit-appearance */
+	/* stylelint-disable property-no-vendor-prefix */
+	-webkit-appearance: none;
+	/* stylelint-enable property-no-vendor-prefix */
 }
 
 /*=== Body */
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -572,6 +572,10 @@ a.btn {
 	width: 100px;
 }
 
+.header > .item.search input {
+	-webkit-appearance: none; /* csslint allow: -webkit-appearance */
+}
+
 /*=== Body */
 #global {
 	display: table;
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -572,7 +572,7 @@ a.btn {
 	width: 100px;
 }
 
-.header > .item.search input {
+input[type="search"] {
 	/* stylelint-disable property-no-vendor-prefix */
 	-webkit-appearance: none;
 	/* stylelint-enable property-no-vendor-prefix */
--- a/p/themes/base-theme/template.rtl.css
+++ b/p/themes/base-theme/template.rtl.css
@@ -573,7 +573,9 @@ a.btn {
 }
 
 .header > .item.search input {
-	-webkit-appearance: none; /* csslint allow: -webkit-appearance */
+	/* stylelint-disable property-no-vendor-prefix */
+	-webkit-appearance: none;
+	/* stylelint-enable property-no-vendor-prefix */
 }
 
 /*=== Body */
--- a/p/themes/base-theme/template.rtl.css
+++ b/p/themes/base-theme/template.rtl.css
@@ -572,6 +572,10 @@ a.btn {
 	width: 100px;
 }
 
+.header > .item.search input {
+	-webkit-appearance: none; /* csslint allow: -webkit-appearance */
+}
+
 /*=== Body */
 #global {
 	display: table;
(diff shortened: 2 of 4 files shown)


Alkarex:
Thanks @kinoushe 👍 
Could you please put those rules in
* https://github.com/FreshRSS/FreshRSS/blob/master/p/themes/base-theme/template.css
* https://github.com/FreshRSS/FreshRSS/blob/master/p/themes/base-theme/template.rtl.css

Those files are used by other themes, while `base.css` is just an example of basic theme to actually used.

Alkarex:
For reference https://developer.mozilla.org/docs/Web/CSS/appearance
@Frenzie Could you please check what we can do with Travis regarding the CSS linting error here?

Frenzie:
I think I'd prefer not to make a global exception but you can just stick `/* csslint allow: prefix-whatever-its-called */` behind the line (or behind the `{` for the whole block I think).

kinoushe:
Sorry @Alkarex I couldn't find those selectors in the `template.css` files, so have now created them and reverted the `base.css` changes.

@Frenzie I've added the `/* csslint allow: -webkit-appearance */` comment, please let me know if this is incorrect.

Thanks for your patience, I'm very new to this!

p/themes/base-theme/template.css
```diff
@@ -572,6 +572,10 @@ a.btn {
 	width: 100px;
 }
 
+.header > .item.search input {
+	-webkit-appearance: none; /* csslint allow: -webkit-appearance */
```
Alkarex, line 576:
```suggestion
.header > .item.search input {
	/* stylelint-disable property-no-vendor-prefix */
	-webkit-appearance: none;
	/* stylelint-enable property-no-vendor-prefix */
}
```

There might be a better syntax, but this one seems to work

Alkarex:
```suggestion
	/* stylelint-disable property-no-vendor-prefix */
	-webkit-appearance: none;
	/* stylelint-enable property-no-vendor-prefix */
```

There might be a better syntax, but this one seems to work

p/themes/base-theme/template.rtl.css
```diff
@@ -572,6 +572,10 @@ a.btn {
 	width: 100px;
 }
 
+.header > .item.search input {
+	-webkit-appearance: none; /* csslint allow: -webkit-appearance */
```
Alkarex:
```suggestion
	/* stylelint-disable property-no-vendor-prefix */
	-webkit-appearance: none;
	/* stylelint-enable property-no-vendor-prefix */
```

p/themes/base-theme/template.css
```diff
@@ -572,6 +572,12 @@ a.btn {
 	width: 100px;
 }
 
+.header > .item.search input {
```
Alkarex: Maybe we should go for a faster & better selector. Could you try that one:

```suggestion
input[type="search"] {
```

p/themes/base-theme/template.rtl.css
```diff
@@ -572,6 +572,12 @@ a.btn {
 	width: 100px;
 }
 
+.header > .item.search input {
```
Alkarex:
```suggestion
input[type="search"] {
```

p/themes/base-theme/template.css
kinoushe: Yes, works as expected in Safari and would cover any other search inputs.

Alkarex:
Is it only Safari that is affected, or other Webkit-based browsers? And which platforms?

kinoushe:
Safari is the only Webkit browser requiring the style fix that I have tested. I've checked the latest versions of Chrome and Opera and they're unaffected by the fix. I can only test on a Mac.

kinoushe:
It also corrects the display in Safari on iPadOS/iOS where the search input appears as a pill shape otherwise:

IMG_5177
IMG_5178

Alkarex:
Ok, thanks, and please put a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md in another PR 👍

p/themes/base-theme/template.css
Frenzie: I meant `-webkit-appearance: none; /* csslint allow: property-no-vendor-prefix */` btw.

Not that it matters or should be changed or anything, I just think it looks slightly neater. ;-)

Alkarex: Please send a fix if you have a better syntax @Frenzie , so that it can also be used as a reference for future needs

PR #2892, opened 2020-04-11
Filter by multiple feed IDs

### LINKED ISSUE #1688: Introduce a new filter for filtering feeds
Hello,
Currently, I can display one feed by clicking on the left menu. If I want to display multiple feeds, I can put them in a category. But I can't put a feed in more than one category.
So it would be useful for me to be able to use filters to display only some feeds.
For example the filter | feed:"Se coucher moins bête" feed:"Le moment Meurice" would only display "Se coucher moins bête" and "Le moment Meurice" based on feed name.
Right now it's possible as a workaround to use the "inurl" filter when all links have the same domain name (example : inurl:secouchermoinsbete.fr works), but it's not the case for example for Shaarli RSS feed (ex : https://husimo.fr/shaarli/rss?do=atom) because link1 is github.com, link2 is devhints.io.
Thank you for your great work :+1:

Alkarex:
Add the possibility to filter by feed ID like `f:123 more-search`
or multiple feed IDs, like `f:123,234,345 more-search` or an exclusion like `!f:456,789 more-search`


Alkarex:
This is a part of a series of technical blocks needed to improve automatic actions (tagging / marking-as-read / ...).

If hope you will like the `f:123,456,789` syntax, which enables transforming the expression into an efficient SQL query of the form `SELECT ... WHERE ... IN (?, ?, ?)` or `SELECT ... WHERE ... NOT IN (?, ?, ?)`

We could easily add the same for other operators such as `author:` `date:` etc. for improved consistency and expressiveness.

The current `OR` boolean mechanism we have is great but does not allow this type of expression, because we do not have parentheses (which would be much more complicated to implement, and probably also much more difficult to transform into an efficient SQL query).

Relying only on the current `OR` mechanism would mean something like `word1 f:123 OR word1 f:234 OR word1 f:345` which is both annoying to write and inefficient to run.

Alkarex:
@Frenzie I have forgotten whether you might have done something better for the translations of the doc? Sorry I am not up to date on the process...

Frenzie:
@Alkarex I wrote a script that uses po4a. Unfortunately the current version in Ubuntu 20.04 is still too old (0.55) for some reason, so you have to compile your own (0.56 or higher); it's kind of annoying.

Anyway, just run the script from the `docs` subfolder and it generates the relevant PO files and output. You can edit the PO files by hand but the great advantage comes from using a program like Lokalize to assist with the translation. ^_^

Alkarex:
@Frenzie Is it still possible to modify the `.md` files by hand (especially for small changes)?

Frenzie:
Yes, it's multidirectional.

Alkarex:
No other comment @Frenzie besides the translation? :-)

Frenzie (approved):
code looks sane, didn't test

PR #2896, opened 2020-04-12
Fix JS for statistics repartition select form

Alkarex:
Error was: `s.form is null in extra.js:166:7`
At the same time, disable the form and clears the value while waiting for the next page to load, to better understand what is happening.

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -163,9 +163,11 @@ function init_select_observers() {
 					const opt = s.options[s.selectedIndex],
 						url = opt.getAttribute('data-url');
 					if (url) {
-						s.form.querySelectorAll('[type=submit]').forEach(function (b) {
-								b.disabled = true;
-							});
+						s.disabled = true;
+						s.value = '';
+						if (s.form) {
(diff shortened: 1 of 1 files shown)


Alkarex:
The code is only used for this form (which was not working anymore):
image

Frenzie (approved):
Seems fine, not sure I fully understood how you'd be in a situation where form is null though?

Alkarex:
We do not have any `<form>` at all for this `<select>` (which should probably be improved, for accessibility, by the way)
https://github.com/FreshRSS/FreshRSS/blob/14fa3f27817d36a2add3c985ccbf553a9ec0e0a0/app/views/stats/repartition.phtml#L8

Frenzie:
Hm, sounds more like an HTML bug than a JS bug then?

Alkarex:
This form was designed to work with JavaScript: the `onchange` is handled automatically.
But yes, I would have better liked a classic form-based approach, which can also function without JavaScript, but that would require a bit more work than this fix

Alkarex:
HTML / accessibility improvements welcome :-)

PR #2897, opened 2020-04-12
Fix double HTML-encoding in category names

Alkarex:
Category names are already HTML-encoded when reaching the view.


Alkarex:
See the category `Vidéo &amp; Audio` :
image

Frenzie (approved):

PR #2899, opened 2020-04-13
Reduce DB locks

Alkarex:
Do not lock the DB to update it when there is nothing to commit.
This should reduce the risk of seeing some `database is locked` errors.


Alkarex:
Here is an example of a WebSub server sending me twice the same update (I am not sure why) almost at the same time, which sometimes produced a DB lock error despite having properly skipped the feed:

image

Frenzie (approved):
Makes sense. Odd cause though.

PR #2903, opened 2020-04-16
A wrong login must return HTTP 403

### LINKED ISSUE #2901: [ authentification ] bannir IP
Bonjour,

Je cherche un moyen pour bannir les ip qui font une mauvaise identification.
Je n'ai rien trouvé, j'ai peut-être mal cherché…

Cela est-il possible avec fail2ban ?

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2901
https://github.com/FreshRSS/FreshRSS/pull/2794/files#r389319248

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -116,17 +116,18 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 		$limits = $conf->limits;
 		$this->view->cookie_days = round($limits['cookie_duration'] / 86400, 1);
 
-		if (Minz_Request::isPost()) {
+		$isPOST = Minz_Request::isPost() && !Minz_Session::param('POST_to_GET');
+		Minz_Session::_param('POST_to_GET');
+
+		if ($isPOST) {
 			$nonce = Minz_Session::param('nonce');
 			$username = Minz_Request::param('username', '');
 			$challenge = Minz_Request::param('challenge', '');
 
 			$conf = get_user_configuration($username);
 			if ($conf == null) {
-				Minz_Request::bad(
-					_t('feedback.auth.login.invalid'),
-					array('c' => 'auth', 'a' => 'login')
-				);
+				//We do not test here whether the user exists, so most likely an internal error.
+				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
 				return;
 			}
 
@@ -155,10 +156,15 @@ class FreshRSS_auth_Controller extends Minz_ActionController {
 				                  ' user=' . $username .
 				                  ', nonce=' . $nonce .
 				                  ', c=' . $challenge);
-				Minz_Request::bad(
-					_t('feedback.auth.login.invalid'),
-					array('c' => 'auth', 'a' => 'login')
-				);
+
+				header('HTTP/1.1 403 Forbidden');
+				Minz_Session::_param('POST_to_GET', true);	//Prevent infinite internal redirect
+				Minz_View::_param('notification', [
+					'type' => 'bad',
+					'content' => _t('feedback.auth.login.invalid'),
+				]);
+				Minz_Request::forward(['c' => 'auth', 'a' => 'login'], false);
+				return;
 			}
 		} elseif (FreshRSS_Context::$system_conf->unsafe_autologin_enabled) {
 			$username = Minz_Request::param('u', '');
--- a/app/Models/Auth.php
+++ b/app/Models/Auth.php
@@ -213,6 +213,7 @@ class FreshRSS_Auth {
 class FreshRSS_FormAuth {
 	public static function checkCredentials($username, $hash, $nonce, $challenge) {
 		if (!FreshRSS_user_Controller::checkUsername($username) ||
+				!ctype_graph($hash) ||
 				!ctype_graph($challenge) ||
 				!ctype_alnum($nonce)) {
 			Minz_Log::debug('Invalid credential parameters:' .


app/Models/Auth.php
```diff
@@ -213,6 +213,7 @@ public static function isCsrfOk($token = null) {
 class FreshRSS_FormAuth {
 	public static function checkCredentials($username, $hash, $nonce, $challenge) {
 		if (!FreshRSS_user_Controller::checkUsername($username) ||
+				!ctype_graph($hash) ||
```
Alkarex, line 216: Just for consistency

Frenzie (approved):
lgtm

app/Controllers/authController.php
```diff
@@ -155,10 +156,15 @@ public function formLoginAction() {
 				                  ' user=' . $username .
 				                  ', nonce=' . $nonce .
 				                  ', c=' . $challenge);
-				Minz_Request::bad(
-					_t('feedback.auth.login.invalid'),
-					array('c' => 'auth', 'a' => 'login')
-				);
+
+				header('HTTP/1.1 403 Forbidden');
```
marienfressinaud, line 160: we should call `http_response_code(403)` here instead. My server is using HTTP/2 for instance.

app/Controllers/authController.php
```diff
@@ -155,10 +156,15 @@ public function formLoginAction() {
...
+				Minz_Session::_param('POST_to_GET', true);	//Prevent infinite internal redirect
```
marienfressinaud, line 161: That's pretty weird in my opinion, but fine for now. Maybe I should take time to improve the whole way to return responses to make it easier and more clear. But what I have in mind would take a lot of time and ask a lot of changes, and so is error-prone.

app/Controllers/authController.php
Alkarex: I also use HTTP/2 and that works fine. HTTP/2 still uses the same status codes. Furthermore, PHP might not know anyway what the Web server and/or proxy are running. So I believe `http_response_code( )` only adds the proper text description.
image

marienfressinaud: Ok then, but it still would be less confusing to not specify the protocol and to use a function that is dedicated to this purpose

Alkarex: Mja, as you want :-) This version is explicit and well known for years. I am not a fan of having too many abstractions instead of a well-established standard (HTTP, SQL...). For instance, I had to look up [`http_response_code()`](https://php.net/http-response-code) to find out what it does, and the spec does not say much, except in the comments telling that it does not support all error codes (which?) so you still need `header()` anyway :-P

According to the PHP source code, it seems that this function is doing exactly that: if it was not already provided, then use a lookup table (implemented with a `while (err->code != 0) {` loop 😲 ) to provide the text corresponding to the status code, and an error if it cannot find the status code (undocumented):
https://github.com/php/php-src/blob/a1eaaa692e91fa28ee1ba050150547cd1317be87/sapi/fpm/fpm/fpm_main.c#L355-L369
https://github.com/php/php-src/blob/a1eaaa692e91fa28ee1ba050150547cd1317be87/ext/standard/head.c#L398-L429
https://github.com/php/php-src/blob/master/main/http_status_codes.h

But with that said, do feel free to change if you prefer, I do not mind. A quick search shows 29 occurrences in our source code ;-)

Alkarex: For completeness, the `HTTP/x` part is transformed by each layer (Web server, proxy), so here is a raw response after it went through Apache and Traefik:

```sh
$ curl -I -X POST 'https://freshrss.example.net/i/?c=auth&a=login'
HTTP/2 403
cache-control: private, must-revalidate, max-age=0
content-security-policy: default-src 'self'
content-type: text/html; charset=UTF-8
date: Fri, 17 Apr 2020 11:00:07 GMT
etag: "983b81b9a93c73d15c6a86d4d6df6330"
last-modified: Fri, 17 Apr 2020 11:00:07 GMT
server: Apache/2.4.41 (Ubuntu)
set-cookie: FreshRSS=oeefb7ittu38ef17db5tpqa8na; path=/i/; secure; HttpOnly; SameSite=Lax
strict-transport-security: max-age=31536000
vary: Accept-Encoding
x-content-type-options: nosniff
```

marienfressinaud: > But with that said, do feel free to change if you prefer, I do not mind. A quick search shows 29 occurrences in our source code ;-)

Ha ha, I'm not that attached to this function :p But I maintain it would be less confusing if we didn't specify the protocol, and `http_response_code()` is the best candidate, while non perfect. For instance, [the first comment](https://www.php.net/manual/en/function.header.php#92305) on the `header` function warn against hard-coding the protocol (10 years ago, right :D). I wouldn't expect the Web server to change this value if already set.

> the spec does not say much, except in the comments telling that it does not support all error codes (which?) so you still need header() anyway :-P

I'm obviously not arguing against the use of `header` which is really convenient, just its current use. By comparing the list of supported status code with a [more complete list](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes), I can't tell which unsupported ones would make sense to FreshRSS (except maybe the 418 ;p)

Finally, I'm afraid that if you want to consider the implementation of the language, PHP might not be the best choice at first ^_^ In the mean time, compilers are often able to optimize this kind of operation, so I wouldn't overlook this.

And again, I just think it's disturbing but it's not a big deal to keep it as it is. It would be if the servers didn't override this value, but you already confirmed it was ok. We'll change if it becomes a problem ;)

Alkarex: Hehe, no worries :-)

PR #2906, opened 2020-04-17
Fix CSS Edge glitch in archiving options

Alkarex:
With Edge, only setting `.hidden = false` does not make the element and sub-elements properly visible.
Should not make any difference for other browsers.

## Example of glitch in Edge
image

## Fixed
image


Alkarex:
Ping @aledeg maybe good to know for other projects (even though EdgeHTML is going away)

Frenzie (approved):
Weird

PR #2909, opened 2020-04-18
Fix nav_menu mark-as-read

### LINKED ISSUE #2905: Mark as read unseen articles
I don't know if this is the same as in #2429 but the effect is the same. The issue occurs after updating to the latest release version 1.16.  

For example, i have 7 new articles at the left side in "main feeds" but only 5 displayed in the mainframe. After clicking "Mark all as read" button at the top, all 7 articles marked as read, not only the 5 in the main frame...

Alkarex:
#Fix https://github.com/FreshRSS/FreshRSS/issues/2905
Fix regression from https://github.com/FreshRSS/FreshRSS/pull/2588

~~We need info about the first item (id_max) before being able to output `nav_menu`. Before https://github.com/FreshRSS/FreshRSS/pull/2588 we used to dump all articles in memory before starting to produce an output. Now that we stream the output, we need a temporary buffer until we have received the first item/article.~~

~~At the same time, I have disabled the "mark as read" menus when there are no articles to mark as read:~~
image


Alkarex:
Now the loading animation works in Chrome as well

marienfressinaud:
Sorry but the code becomes less and less clear to me. At a certain point, I prefer to have a code that performs less efficiently but being able to read it. Here I see an output buffering starting in a controller, and cleaned for some reasons in a callback declared in a view. Also some `flush`, sometimes next to a `ob_flush`, etc. It doesn't make any sense to me while I'm sure there is one ;)

It seems to me we should stop (or maybe merge this one as it seems to fix a bug) and reconsider the whole way of displaying this page. Preferably without any `ob_start` and `flush` and all these things, but if they appear to be necessary, at least we should find a better way to architecture things…

Alkarex:
Indeed @marienfressinaud , it is a bit messy between the controller and the view, and I will try to come back to it at some point. It is also possible to use some functions that hide a bit some aspects from the programmer, such as https://php.net/ob-implicit-flush (because `ob_flush` and `flush` are two distinct things).

Alkarex:
Another note: several of those hacky things would not be needed if we went for JavaScript client side instead of a working HTML page enhanced with JavaScript

Alkarex:
I think I might have an idea to reduce the complexity of the code in the view (e.g. `callbackFirstEntry`). One reason it is like it is in this patch is because PHP Generators do not offer peek or similar operations (for a good reason), only linear streaming https://stackoverflow.com/a/662925 . But we could have a mini wrapper that keeps the last entry in memory and allows peek, which could make the view code more straightforward without impacting performance

app/layout/aside_feed.phtml
```diff
@@ -109,6 +109,9 @@
 	</form>
 </div>
 
+<div id="first_load" class="loading"></div>
```
Alkarex, line 112: Moved as high as possible

app/Controllers/indexController.php
```diff
@@ -48,6 +48,8 @@ public function normalAction() {
 		}
 		Minz_View::prependTitle($title . ' · ');
 
+		FreshRSS_Context::$id_max = time() . '000000';
```
Alkarex, line 51: Simplified logic relying solely on current time instead of the actual newest element in DB. Should not matter because inserted elements after this request will have a newer ID.

app/Controllers/indexController.php
```diff
@@ -66,29 +68,20 @@ public function normalAction() {
...
+		$this->view->callbackBeforePagination = function ($view, $nbEntries, $lastEntry) {
 			if ($nbEntries >= FreshRSS_Context::$number) {
 				//We have enough entries: we discard the last one to use it for the next pagination
 				ob_clean();
 				FreshRSS_Context::$next_id = $lastEntry->id();
 			}
 			ob_end_flush();
-
-			FreshRSS_Context::$id_max = $firstEntry === null ? (time() - 1) . '000000' : $firstEntry->id();
-			if (FreshRSS_Context::$order === 'ASC') {
```
Alkarex, line 85: Use same logic for ASC as for DESC

Alkarex:
@marienfressinaud New proposition, with much simplified logic.
@Offerel Could you please test this new version of the fix?

Alkarex:
~~Ah, there might be a bug with paging in ASC, will check a bit later~~

Alkarex:
> ~~Ah, there might be a bug with paging in ASC, will check a bit later~~

Scratch that, should be good after all :-)

app/layout/aside_feed.phtml
marienfressinaud: Can't we move these two lines between `$this->partial('aside_feed');` and `$this->partial('nav_menu');` (in normal and reader files?) It looks unnatural to have them in this file (i.e. they are not part of the aside), and so it's more difficult to understand their purpose.

marienfressinaud (approved):
I prefer this version a lot more :) I made only one comment but it's not critical so feel free to do what you want about it

app/layout/aside_feed.phtml
Alkarex: The best would be to change the CSS rule for `div#aside_feed` to ensure that it does not take the full width when it is the sole element while loading (a consequence of  `display:table-cell`), which is responsible for a layout flash in particular in Chrome. This `#first_load` is thus a helper for its previous sibling `div#aside_feed`.
Moving it between `$this->partial('aside_feed');` and `$this->partial('nav_menu');` would have two small drawbacks:
* It would need to be repeated for `normal.phtml` and for `reader.phtml`
* It would appear ~1.6KB later which, depending on where the buffering cuts, reduces a bit its chances of being visible

I suggest to leave it as it is and come back to it next time we optimise the CSS layout

marienfressinaud: Ok then :)

PR #2911, opened 2020-04-19
Enhance the way help elements are displayed

### LINKED ISSUE #2886: Enhance the way help elements are displayed
Hello,

I already talked a bit about that on a previous PR, but I think the way the help elements are displayed need to be enhanced.

Currently (at least on Safari), they can look very bogus (not to say messy) :

Screenshot 2020-04-09 at 15 42 31

Screenshot 2020-04-09 at 15 42 35

Screenshot 2020-04-09 at 15 42 44

→ They are wrapped in an unexpected place.

I would suggest to just show them (in a whole) on the next "line" each time we are using them. I'm not sure it would be the best solution, but for the advantages I see :
- They wouldn't be "stuck" in the right, without any place, specially on side panel : less wrapping.
- They wouldn't wrap in an unexpected place anymore (i.e. they would wrap with alignment to the "help" icon, instead of the associated element as it's currently the case).

Or perhaps the text can just be hidden, and shown as a real tooltip, when the cursor wait a bit (or click) on the "help" icon ?

javerous:
Closes https://github.com/FreshRSS/FreshRSS/issues/2886

- Move help under elements instead of being at the right. I feel it's more elegant, specially for view very limited in size, like side panel.
- Homogenize some help placement in some views (register, add-feed-by-get…)

1. Just… go to the views. 

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

Here is some comparative screenshot (doesn't include all changes). I updated the whole UI, and not just the side panels. I felt it's more homogeneous this way (and, personally, I like the result), but I can limit this change only to side panels if wanted.

→ _app/views/helpers/feed/update.phtml_
- Before
feed_update_before
feed_reload_before

- After
feed_update_after
feed_reload_after

→ _app/views/user/profile.phtml_ & _app/views/user/details.phtml_
- Before 
user_before

- After
user_after

→ _app/views/auth/register.phtml_
- Before
register_before

- After
register_after

### DIFF
--- a/app/views/auth/register.phtml
+++ b/app/views/auth/register.phtml
@@ -5,8 +5,9 @@
 		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
 
 		<div class="form-group">
-			<label for="new_user_name"><?= _t('gen.auth.username'), '<br />', _i('help'), ' ', _t('gen.auth.username.format') ?></label>
+			<label for="new_user_name"><?= _t('gen.auth.username') ?></label>
 			<input id="new_user_name" name="new_user_name" type="text" size="16" required="required" autocomplete="off" pattern="<?= FreshRSS_user_Controller::USERNAME_PATTERN ?>" autocapitalize="off" />
+			<p class="help"><?= _i('help') ?> <?= _t('gen.auth.username.format') ?></p>
 		</div>
 
 		<?php if ($this->show_email_field) { ?>
@@ -19,12 +20,14 @@
 		<?php } ?>
 
 		<div class="form-group">
-			<label for="new_user_passwordPlain"><?= _t('gen.auth.password'), '<br />', _i('help'), ' ', _t('gen.auth.password.format') ?></label>
+			<label for="new_user_passwordPlain"><?= _t('gen.auth.password') ?></label>
 			<div class="stick">
 				<input type="password" id="new_user_passwordPlain" name="new_user_passwordPlain" required="required" autocomplete="new-password" pattern=".{7,}" />
 				<a class="btn toggle-password" data-toggle="new_user_passwordPlain"><?= _i('key') ?></a>
 			</div>
 			<noscript><b><?= _t('gen.js.should_be_activated') ?></b></noscript>
+			<p class="help"><?= _i('help') ?> <?= _t('gen.auth.password.format') ?></p>
+
 		</div>
 
 		<?php if ($this->show_tos_checkbox) { ?>
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -59,6 +59,10 @@ p {
 	font-size: 1em;
 }
 
+p.help {
+	margin: 5px 0 0.5em !important;
+}
+
 sup {
 	line-height: 25px;
 	position: relative;
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -59,8 +59,8 @@ p {
 	font-size: 1em;
 }
 
-p.help {
-	margin: 5px 0 0.5em !important;
+p.help, .prompt p.help {
+	margin: 5px 0 0.5em;
 }
 
 sup {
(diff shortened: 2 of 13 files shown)


p/themes/base-theme/template.css
```diff
@@ -59,6 +59,10 @@ p {
 	font-size: 1em;
 }
 
+p.help {
+	margin: 5px 0 0.5em !important;
```
Alkarex: Could you try to find a rule without `!important` ?

p/themes/base-theme/template.rtl.css
```diff
@@ -59,6 +59,10 @@ p {
 	font-size: 1em;
 }
 
+p.help {
+	margin: 5px 0 0.5em !important;
```
Alkarex: Could you try to find a rule without `!important` ?

Alkarex:
Thanks, that looks good 👍

p/themes/base-theme/template.css
javerous: I'm not good enough in CSS for that, sorry :s

It's needed because of the `prompt` class (used in `register`, and defined in some templates). I saw that some other rules are using it, so I thought it was okay.

p/themes/base-theme/template.rtl.css
javerous: Same.

p/themes/base-theme/template.css
Alkarex: You could try something like
```css
* p.help {
```

or

```css
p.help, .prompt p.help {
```

Frenzie: The * p.help suggestion's a bit ugly to me. (Not objectionable like !important, just dispreferred.)

Edit: also a bit similar to !important in that it seems to use too much specificity.

app/views/auth/register.phtml
```diff
@@ -5,8 +5,9 @@
 		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
 
 		<div class="form-group">
-			<label for="new_user_name"><?= _t('gen.auth.username'), '<br />', _i('help'), ' ', _t('gen.auth.username.format') ?></label>
+			<label for="new_user_name"><?= _t('gen.auth.username') ?></label>
 			<input id="new_user_name" name="new_user_name" type="text" size="16" required="required" autocomplete="off" pattern="<?= FreshRSS_user_Controller::USERNAME_PATTERN ?>" autocapitalize="off" />
+			<p class="help"><?= _i('help') ?> <?= _t('gen.auth.username.format') ?></p>
```
Frenzie, line 10: Is getting rid of the label a conscious decision? (Not an objection, I think it might make the most sense this way.)

javerous: @Frenzie Yes, I thought it would make sense to be homogeneous with what we are doing in other places with this help couple. That being said, I'm not 100% sure if there was a specific reason to encapsulate the help in the label with a `<br>` (I supposed it was old code which was not updated, as it doesn't look like a very modern CSS-ish way to do things).

Frenzie: To be clear, by "getting rid of the label" I meant not using `<p><label for=>` on the help, nothing weird.

p/themes/base-theme/template.css
javerous: @Alkarex So I tested:
- `* p.help {`, it doesn't work (still using prompt `p` marging):
Screenshot 2020-04-20 at 12 59 34

- `p.help, .prompt p.help {` seems to works (I see the right marging):
Screenshot 2020-04-20 at 13 00 13

So I guess we can retain the second solution. But isn't breaking a bit encapsulation / CSS hierarchy ? (i.e. it's a bit fragile and it doesn't scale that much: it needs to be updated each time a new class defining a `p` is used by an HTML element which encapsulate an `help`, like with this prompt)

app/views/auth/register.phtml
javerous: Yep, it's what I understood ;)

p/themes/base-theme/template.css
Alkarex: This is because `.prompt p` was a bit too general for a rule

PR #2913, opened 2020-04-19
Fix for "Flash Of Unstyled Content" and slow loading in the Firefox

### LINKED ISSUE #2912: "Flash Of Unstyled Content" and slow loading in the Firefox because of <script defer>
FreshRSS version: 1.15.3 (Yunohost build)
Hardware: Raspberry Pi 3B+
Browser: Firefox 75

There has been lot of loading and display issues when browsing FreshRSS with Firefox.

Firefox loads content without styling first, causes lot of stress to Yunohost and often logs out after this happens.

I found the culprit: Firefox doesn't like script tags with "defer" value in them, so it tries to load scripts first then CSS.

There's stupidly easy but slightly hacky solution: add <script> with comment **only** into HTML after CSS styles but before the actual scripts.

Source: https://bugzilla.mozilla.org/show_bug.cgi?id=1404468#c78

Not only this fixes FOUT, it fixes slow loading issues (at least on my side).

Submitting pull request soon.

krisu5:
Closes #2912

- Adds commented but otherwise empty <script> tag to layout.phtml HTML between styles and actual scripts. This stops Firefox to loading scripts first, fixes FOUC and improves performance.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -9,6 +9,7 @@ if (_t('gen.dir') === 'rtl') {
 		<meta charset="UTF-8" />
 		<meta name="viewport" content="initial-scale=1.0" />
 		<?= self::headStyle() ?>
+		<script>/* Firefox FOUC Fix */</script>
 		<script id="jsonVars" type="application/json">
 <?php $this->renderHelper('javascript_vars'); ?>
 		</script>


Alkarex:
We should remember to track https://bugzilla.mozilla.org/show_bug.cgi?id=1404468#c78 to see the conclusions

Alkarex:
@krisu5 Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

Frenzie:
I was actually able to reproduce a FOUC today (without the fix) on my laptop while another tab(s) was eating a lot of CPU. But normally it simply appears instantaneously. I'll have to see if I can reproduce it on my old netbook or artificially slowed down in some way.

Alkarex:
Ah, good idea @Frenzie ,
I only tried to slow down network, but not CPU

Alkarex:
I have made a quick test with 1 CPU reduced at 30% speed, and with/without network throttling at 2G-speed.

With that setup, I sometimes see
image

And when disabling cache:
image

It was only a 5min test, but I was not able to produce a significant different with/without the fix. Additional tests welcome

krisu5:
> @krisu5 Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

Done. Thanks for merge!

Frenzie:
@Alkarex Indeed, my further findings are the same.

Alkarex:
It could be related to having the developer tools open or not https://bugzilla.mozilla.org/show_bug.cgi?id=1404468#c81
@krisu5 Could you check whether you observe any difference with/without the fix when the developer tools are closed?

krisu5:
> It could be related to having the developer tools open or not https://bugzilla.mozilla.org/show_bug.cgi?id=1404468#c81
> @krisu5 Could you check whether you observe any difference with/without the fix when the developer tools are closed?

No, same thing happens.

My guess would be that in the Yunohost build, removing the quick access to SSO via the "Yunohost tile" caused this. I think it didn't have Firefox FOUC issue before removal, but there was more non-intentional log outs, so that's why they got rid of it.

PR #2922, opened 2020-04-22
WebSub move fixes

Alkarex:
Small fixes related to when WebSub changes address:
* When unsubscribing, one must use the current/old address, not the newly found selfUrl;
* This change https://github.com/FreshRSS/FreshRSS/pull/2659#discussion_r347263068 was wrong, so reverted to the first version. We must obey the rules also for feeds for which the initial WebSub enabling is not finished, otherwise we never have a chance to redirect the feed to the proper selfUrl.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -427,7 +427,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 				$entryDAO->commit();
 			}
 
-			if ($pubSubHubbubEnabled && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
+			if ($pubsubhubbubEnabledGeneral && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
 				if ($feed->selfUrl() !== $url) {	// https://github.com/pubsubhubbub/PubSubHubbub/wiki/Moving-Feeds-or-changing-Hubs
 					$selfUrl = checkUrl($feed->selfUrl());
 					if ($selfUrl) {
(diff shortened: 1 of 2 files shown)


app/Controllers/feedController.php
```diff
@@ -427,7 +427,7 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 				$entryDAO->commit();
 			}
 
-			if ($pubSubHubbubEnabled && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
+			if ($pubsubhubbubEnabledGeneral && $feed->hubUrl() && $feed->selfUrl()) {	//selfUrl has priority for WebSub
```
Alkarex, line 430: Example: We register to `https://push-pub.appspot.com/feed` which `self` is `http://push-pub.appspot.com/feed` so we cannot never complete the WebSub enabling for that feed if we do not enter this condition

app/Models/Feed.php
```diff
@@ -695,7 +695,11 @@ public function pubSubHubbubPrepare() {
 
 	//Parameter true to subscribe, false to unsubscribe.
 	public function pubSubHubbubSubscribe($state) {
```
marienfressinaud, line 697: not a big deal and not directly related to this PR, but wouldn't it better to call this method `setPubSubHubbubSubscriptionState`? or `setWebsubSubscriptionState`?

marienfressinaud (approved):
it looks good but also complicated to be sure we are considering all the edge cases simply by looking at the code :s

app/Models/Feed.php
Alkarex: Yes, a bit of renaming at a few places could be good. I will try to find time to make another PR for better handling the case when a user deletes a feed, in a multi-user context

PR #2933, opened 2020-04-30
2914 show last login date

### LINKED ISSUE #2914: Feature Request: Show last login (use) date of a user.
I have a couple of users in my instance. It would be nice to identify inactive users for a cleanup.

equinoxmatt:
Closes #2914

- Adds last login date to default user configuration file
- Displays last login date on manage users and user detail screen
- Writes current timestamp to user configuration when user logs in.

1. Create user.
2. Confirm 'Last Login' column on manage users screen and detail screen is 'Never Logged In'
3. Login as new user
4. Confirm 'Last Login' displays correct timesamp.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [] unit tests written (optional if too hard)
- [ ] documentation updated

I was wondering about time formatting. I have stuck with the basic ISO 8601 format, but it isn't the most user friendly display. How are dates handled generally within the project?

Screenshot from 2020-04-30 10-33-28

Screenshot from 2020-04-30 10-33-52

### DIFF
--- a/app/views/user/manage.phtml
+++ b/app/views/user/manage.phtml
@@ -77,6 +77,7 @@
 				<th><?= _t('admin.user.feed_count') ?></th>
 				<th><?= _t('admin.user.article_count') ?></th>
 				<th><?= _t('admin.user.database_size') ?></th>
+                <th><?= _t('admin.user.last_login') ?></th>
 				<th>&nbsp;</th>
 			</tr>
 		</thead>
@@ -90,6 +91,7 @@
 					<td><?= format_number($values['feed_count']) ?></td>
 					<td><?= format_number($values['article_count']) ?></td>
 					<td><?= format_bytes($values['database_size']) ?></td>
+                    <td><?= $values['last_login'] ? $values['last_login'] : _t('admin.user.never_loggedin') ?></td>
 					<td><a href="<?= _url('user', 'details', 'username', $username) ?>">Details</a></td>
 				</tr>
 			<?php endforeach ?>
--- a/app/views/user/manage.phtml
+++ b/app/views/user/manage.phtml
@@ -77,7 +77,7 @@
 				<th><?= _t('admin.user.feed_count') ?></th>
 				<th><?= _t('admin.user.article_count') ?></th>
 				<th><?= _t('admin.user.database_size') ?></th>
-                <th><?= _t('admin.user.last_login') ?></th>
+				<th><?= _t('admin.user.last_login') ?></th>
 				<th>&nbsp;</th>
 			</tr>
 		</thead>
@@ -91,7 +91,7 @@
 					<td><?= format_number($values['feed_count']) ?></td>
 					<td><?= format_number($values['article_count']) ?></td>
 					<td><?= format_bytes($values['database_size']) ?></td>
-                    <td><?= $values['last_login'] ? $values['last_login'] : _t('admin.user.never_loggedin') ?></td>
+					<td><?= $values['last_login'] ? $values['last_login'] : _t('admin.user.never_loggedin') ?></td>
 					<td><a href="<?= _url('user', 'details', 'username', $username) ?>">Details</a></td>
 				</tr>
 			<?php endforeach ?>
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -616,6 +616,7 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 			'language' => $userConfiguration->language,
 			'mail_login' => $userConfiguration->mail_login,
 			'is_admin' => $userConfiguration->is_admin,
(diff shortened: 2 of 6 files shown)


Alkarex:
That looks good @equinoxmatt 👍 
I will have a look at bit more tonight.

There will be at least some minor things:
* Remove the docker-compose in the commit;
* Minor whitespace consistency
* i18n for other languages (just putting a TODO for the other languages)

app/views/user/manage.phtml
```diff
@@ -77,6 +77,7 @@
 				<th><?= _t('admin.user.feed_count') ?></th>
 				<th><?= _t('admin.user.article_count') ?></th>
 				<th><?= _t('admin.user.database_size') ?></th>
+                <th><?= _t('admin.user.last_login') ?></th>
```
Alkarex:
```suggestion
				<th><?= _t('admin.user.last_login') ?></th>
```

Minor whitespace consistency

app/views/user/manage.phtml
```diff
@@ -90,6 +91,7 @@
 					<td><?= format_number($values['feed_count']) ?></td>
 					<td><?= format_number($values['article_count']) ?></td>
 					<td><?= format_bytes($values['database_size']) ?></td>
+                    <td><?= $values['last_login'] ? $values['last_login'] : _t('admin.user.never_loggedin') ?></td>
```
Alkarex:
```suggestion
					<td><?= $values['last_login'] ? $values['last_login'] : _t('admin.user.never_loggedin') ?></td>
```

Minor whitespace consistency

Alkarex:
I am not sure the method for finding out last login is good.
We might have people using solely the API for instance

See my suggestion https://github.com/FreshRSS/FreshRSS/issues/2914#issuecomment-616465778

Using:

```php
FreshRSS_UserDAO::mtime($username)
```

i.e. last user action that has an impact on database (e.g. marking an article as read, adding a feed...)

app/views/user/manage.phtml
equinoxmatt: Oops good spot, tabs vs spaces :sweat_smile:

equinoxmatt:
@Alkarex  - I did make a comment on that issue.  The problem that I see, is an admin can modify a user, which in turns modifies the configuration file. If you use the last modified time of the configuration file, this does not reflect the time the user was last active.

Also, from my tests (I could be wrong), simply logging in does not modify the configuration file, therefore it is not a reliable indicator of when the user was last 'active'.

Alkarex:
We could call `FreshRSS_UserDAO::touch();` after an active login to catch this case.
You are right if an admin makes a manual configuration change in the user configuration file.
I would like to move the user configuration file in a new database table, but in the meantime, I would better like to avoid reading/overwriting the user configuration file so intensively, which I find a bit too risky.

equinoxmatt:
Ahhh, I didn't know about the touch function. 

That seems like the best solution then.

1. Add touch event upon login (api and application)
2. Revert to using `FreshRSS_UserDAO::mtime($username)` (remove new field from user config file)

That sound correct?

Alkarex:
Yes, and when we end up making a new database table for user information and/or config, we will have a better field to store this information

equinoxmatt:
@Alkarex the only other issue I see, is as it stands, there will never be a state where the user hasn't been logged in, as the file will always have a modified date. I guess this will have to be resolved by moving user configuration to the database. Just a though.
 
EDIT: I accidentally closed the PR, can you re-open please?

equinoxmatt:
@Alkarex this has been updated now:

- Translations updated with TODOs
- Last login time now uses `reshRSS_UserDAO::mtime($username)`
- Successful login now touches the config file to record last login.

Could you re-open the pull request? I can't for some reason. I suspect I messed up my merge history by merging from master and then removing it and this has confused github. If you can't re-open it i will create a new PR.

Also, just a thought about translations.  Rather than having a bunch of TODOS on translations, it would be nice to have a designated fallback language. So if the chosen language doesn't have a translation for the string, it uses the default one.

Alkarex:
Hum, it does not look like I can reopen the PR on my side :-/
Can you resubmit?

Frenzie:
> Also, just a thought about translations. Rather than having a bunch of TODOS on translations, it would be nice to have a designated fallback language. So if the chosen language doesn't have a translation for the string, it uses the default one.

That is what happens, isn't it? Todos are for translators. You can search for them. ;-)

equinoxmatt:
@Frenzie Ahh, that makes sense. Apologies, when asked to create the TODOS i assumed that was because there was no fallback. I don't know enough about the codebase... yet ;)

equinoxmatt:
> Hum, it does not look like I can reopen the PR on my side :-/
> Can you resubmit?

#2936

PR #2935, opened 2020-04-30
API Use string type for Usec

### LINKED ISSUE #2934: [greader-api] mixed string and number
when making a get request to `/reader/api/0/unread-count` the following is returned:
```
...
    {
      "id": "feed/6",
      "count": 0,
      "newestItemTimestampUsec": 1588269040490890
    },
...
```
when looking at other documentations about google reader api: The `newestItemTimestampUsec` should be a string and not a number.
sources:
- https://feedhq.readthedocs.io/en/latest/api/reference.html#unread-count
- https://web.archive.org/web/20120608040458/http://undoc.in/unread-count.html

Alkarex:
#Fix https://github.com/FreshRSS/FreshRSS/issues/2934

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -421,7 +421,7 @@ function unreadCount() {	//http://blog.martindoms.com/2009/10/16/using-the-googl
 			$unreadcounts[] = array(
 				'id' => 'feed/' . $feed->id(),
 				'count' => $feed->nbNotRead(),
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Weird but okay.

Alkarex:
@Frenzie I believe the original intent was to be compatible with 32-bit platforms

Issue #1876, opened 2018-04-30
Mobile : No access to parameters


Glaived:
Hi,

on a smartphone, there is no icon to access the settings, would it be possible to fix it?

![capture-20180430220226-375x387](https://user-images.githubusercontent.com/1835170/39447501-3b042cb4-4cc2-11e8-8515-d86d089a60ea.png)

seboss666:
Hello,
I see this issue is still opened but not much activity, so i'll add my two cents just in case.
With the current behaviour of feed management settings, I can understand that convert the settings interface is not a trivial task so mask it at first place is not necesarily bad. But it's not good when your main use of FreshRSS is from mobile (and not a third party app like easyrss that does not work correctly on my phone).
Do you have at least some ideas about how you plan to do that ? Maybe at first just expose the feeds managements settings (categories, feed addition/modifications), as I think it's the main feature freshrss may manipulate in a day-to-day usage. Sadly I'm not skilled enough particularly in CSS to help it, aside from testing.

seboss666:
I've tried it from the bookmarklet to see how it copes. It is not broken as we could have expected, but it's clearly not suitable for mobile viewport. I've posted a screen on my last blog post (the reason I'm getting here in fact :) ) : https://blog.seboss666.info/2018/12/les-flux-rss-sur-youtube-la-fin-de-mon-compte-google/

Frenzie:
Oh, I guess I only looked at the subscriptions management window.

There's the padding in .post.

There's the left in `#slider.active` and the left margin in .form-group .group-controls which takes care of the worst of it.
```css
.form-group .group-controls {
    min-width: 250px;
    /* margin: 0 0 0 220px; */
}
```

And there's the width of textarea:

```css
textarea {
    /* width: 360px; */
    height: 100px;
}
```

Not quite right, but anyway, that's just a few minutes of preliminary investigation. ;-)

![screenshot_2018-12-24_10-52-34](https://user-images.githubusercontent.com/202757/50396404-20d0e680-076a-11e9-8d02-f3f7fcf748a9.png)

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1929

Issue #2883, opened 2020-04-08
Theme Alternative-Dark consider using same base template


Alkarex:
@fabianski7 Could you please try to use the same base template https://github.com/FreshRSS/FreshRSS/blob/master/p/themes/base-theme/template.css than other themes, instead of a slightly modified copy https://github.com/FreshRSS/FreshRSS/blob/master/p/themes/Alternative-Dark/template.css ?

I suggest you override what you want in https://github.com/FreshRSS/FreshRSS/blob/master/p/themes/Alternative-Dark/adark.css instead

This makes it easier to push small changes for all themes at once, such as:
* https://github.com/FreshRSS/FreshRSS/pull/2882
* https://github.com/FreshRSS/FreshRSS/pull/2877
* https://github.com/FreshRSS/FreshRSS/pull/2861
* https://github.com/FreshRSS/FreshRSS/pull/2845
* https://github.com/FreshRSS/FreshRSS/pull/2843

Follow-up of https://github.com/FreshRSS/FreshRSS/pull/2811

Alkarex:
Done in https://github.com/FreshRSS/FreshRSS/pull/3000
Tests welcome

PR #2938, opened 2020-05-01
Add/mobile feed configuration

marienfressinaud:
References #1876

Add a button to the feed management page and improve the look of this page and panel on mobile. I'll open another PR later for the configuration section.

Capture d’écran - 2020-05-01 à 10 40 24

Capture d’écran - 2020-05-01 à 10 40 39

1. On mobile or by enabling mobile mode in devtools (<kbd>ctrl+maj+m</kbd> on Firefox), open the feed management page
2. Check everything is working fine
3. Repeat with each theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Alkarex:
Looks good @marienfressinaud 👍

marienfressinaud:
Btw, I don't remember if there's a process to generate RTL files or if I need to change the files manually? cc @Frenzie

Frenzie:
@marienfressinaud I added a `make rtl` command or something to that extent. The requirements should be fairly self-evident from that but I can give more detail later if you want. You have to install the relevant Node.JS module, probably RTLCSS but perhaps it was the other one. (There are two that are basically the same and currently in the progress of merging.)

marienfressinaud:
I think I'm done with this PR. If you want to understand what I did, the first commits are the step-by-step process applied to the Origine theme and to the template file. Then I fixed themes one by one and I finished by applying the `make rtl` command (didn't check it was working fine though).

Few thoughts about this PR:

- there was less work than what I expected (but keep in mind the "configuration" section is not included here)
- it's very painful to maintain so many themes and I guess some of them could be merged (no needs to have two "dark" themes, also Origine, Origine-compact and Dark themes could be one theme with options)
- we should not accept themes that doesn't rely on the default `template.css`: its purpose is to reduce the maintenance complexity, so the Alternative-Dark theme must be adapted
- the whole CSS isn't well organized and I'm starting to think Sass could come to the rescue. It could be helpful to create new themes as well. The problem is it adds complexity to our development workflow, but in the mean time it is required to work on Ansum, Mapco and Swage themes.

Alkarex:
Great, @marienfressinaud 😃 

> we should not accept themes that doesn't rely on the default template.css: its purpose is to reduce the maintenance complexity, so the Alternative-Dark theme must be adapted

Indeed https://github.com/FreshRSS/FreshRSS/issues/2883

I think the next big work on CSS should be done at the same time than another significant UI change. For instance, several aspects would be much easier / faster / greener with a Web app - without changing much of our controllers, and we could still keep a simpler no-JavaScript HTML

Alkarex:
There is a little bug when scrolling down and then up on mobile (Firefox Beta on Android), but I do not think it is a big deal (see bottom):
Screenshot_20200505-170821_Firefox Beta

Alkarex (approved):

Frenzie:
As an aside, I'm not all that much in favor of SASS/preprocessors. See https://www.456bereastreet.com/archive/201603/why_i_dont_use_css_preprocessors/ for what's also my general POV.

> That is, I don’t really find the “problems” with CSS that preprocessors are intended to solve, problems. The scale of the site I’m building does not matter, be it a tiny site with just a few static pages or a humongous corporate intranet. I simply have never felt the need for mixins, nesting or extends.

>  to me the solution is worse than the problem.

> However, me not using Sass or other CSS preprocessors like cssnext does not mean I don’t use CSS processors. The difference, as I see it, is whether or not your CSS requires compilation before browsers can understand it, which I really want to avoid.

(Agreed wrt template.css of course. :-) )

Frenzie (approved):
Seems fine, didn't check in great detail (especially for all the individual themes).

marienfressinaud:
> I think the next big work on CSS should be done at the same time than another significant UI change. For instance, **several aspects would be much easier / faster / greener with a Web app** - without changing much of our controllers, and we could still keep a simpler no-JavaScript HTML

Remember some others would be more difficult ;) I believe you would rely on an existing API so this part would not be such a big deal, but then we will have to maintain the state of the app in the browser, which can become a real problem. Also we would add NPM or Yarn to manage our JS dependencies, which tend to be mysterious to me (`npm install` had a tendency to break too often on my previous projects without clear explanation). The biggest benefit I can see in a Web app is to make it working offline, but then we have to deal with [service workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) which are still very painful to set up correctly and to debug.

I'm not saying we should not go with a Web app (there are other benefits, of course), but keep in mind it's not a silver-bullet: it adds a lot of hidden complexity while there might be different simpler approaches to simplify the current code. On my new project, I'm going to try a different approach based on a small JS framework, [Stimulus](https://github.com/stimulusjs/stimulus), and [Turbolinks](https://github.com/turbolinks/turbolinks) to improve performance during the navigation. I hope it will be a lot easier to maintain than a Web app, but I can't say yet if it would be a good fit with FRSS ^^

> As an aside, I'm not all that much in favor of SASS/preprocessors.

I kind of agree with you: with modern CSS, I believe good conventions and coding style can solve a lot of problems people try to solve with Sass, without adding complexity. But there is still one benefit that is only partially addressed by CSS: splitting the code in different files. The current `template.css` file is ~1400 lines long which is impossible to maintain correctly. I'd like to split it in different files, as it is done in Ansum and Mapco themes (but slightly differently though). There is obviously the CSS [`@import` rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@import), but I think the performance would become disastrous without HTTP2.

A small script concatenating the files would be enough though… but since Sass is already there for 3 themes, we might as well make it available for the other themes. This doesn't mean we have to use all the complexity from Sass ;)

kinoushe:
In Safari, the search button no longer vertically aligns with the search input field.

It appears that the change to `display: inline-flex` from `display: inline-block` for `.stick` in the `template.css` files has caused this:

```
/*=== Buttons */
 .stick {
  display: inline-flex;
  ...
} 
```

In my browser this results in the `button` rendering larger at 16 x 29px high, instead of 16 x 25px as before (25px is the height of the input field):

Screen Shot 2020-05-22 at 11 14 42 am

marienfressinaud:
Thanks for the report @kinoushe! I cannot test on Safari but I was able to reproduce on Gnome Web (finally this browser proved it can be useful ^^), I guess they both use an older version of Webkit. I made a fix: https://github.com/FreshRSS/FreshRSS/pull/2995
Could you give it a try?

kinoushe:
Haha, excellent! I've just manually tested your new CSS rules and it fixes the button height in Safari (13.1) also. Thanks @marienfressinaud!

PR #2942, opened 2020-05-02
Table overview of compatible clients (Help welcome)

Alkarex:
I found it difficult - especially for newcomers - to find and pick a client, and I believe a table would help.
Help welcome to fill-in the blanks.

PR #2944, opened 2020-05-02
Fever integer type for last_refreshed_on_time

### LINKED ISSUE #2940: [fever-api] mixed string and number in time
now the exact opposite to #2934 :
when looking at https://feedafever.com/api the returned `last_refreshed_on_time` should always be a number. FreshRSS-Fever is always returning a `String`.

Alkarex:
#Fix https://github.com/FreshRSS/FreshRSS/issues/2940
https://feedafever.com/api

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -288,7 +288,7 @@ class FeverAPI
 		$arr = array('api_version' => self::API_LEVEL, 'auth' => $status);
 
 		if ($status === self::STATUS_OK) {
-			$arr['last_refreshed_on_time'] = (string) $this->lastRefreshedOnTime();
+			$arr['last_refreshed_on_time'] = $this->lastRefreshedOnTime();
 			$arr = array_merge($arr, $reply);
 		}
 


p/api/fever.php
```diff
@@ -288,7 +288,7 @@ public function wrap($status, array $reply = array())
 		$arr = array('api_version' => self::API_LEVEL, 'auth' => $status);
 
 		if ($status === self::STATUS_OK) {
-			$arr['last_refreshed_on_time'] = (string) $this->lastRefreshedOnTime();
```
Alkarex, line 291: @kevinpapst It was [a long time ago](https://github.com/FreshRSS/FreshRSS/pull/1836), but do you happen to remember why this was casted to a string?
https://github.com/FreshRSS/FreshRSS/issues/2940

kevinpapst: Hahaha, can you remember why you did things last months or even two years ago? 😁 

Short answer: no, unfortunately I can't 😞 looks correct when checking the API definition.
I would question if there was a bug at any point or if that change is only happening to be compatible with the "official documentation". I never had troubles with any clients I tested.

Offtopic: I'd really like to help more, but I don't have the time to take care of any other projects that my ones currently... sorry that you have to take care of the API now! I know its a burden (but still great for the iOS users).

kevinpapst (approved):

p/api/fever.php
Alkarex: No worries, and thanks for the quick answer :-)
And if you are still using it, could you try that it still works with your favourite app after this fix? I can merge in the master branch if that can help.
Also, I am making a table overview of clients, and feedback welcome (I have nothing iOS to try myself) https://github.com/FreshRSS/FreshRSS/pull/2942

Alkarex:
Second take https://github.com/FreshRSS/FreshRSS/pull/2945

Issue #2953, opened 2020-05-06
CRON_MIN -> Real cron


Pofilo:
Hello,

In the Docker image, we have a `CRON_MIN` parameter to enable the cron task.
However, I want it to run at 7AM / 12PM and 19PM, and there is no way to do such a thing right now.
The cron looks like: `5 7,12,21 * * *` (5 is random for minutes (to not overload servers at 0)).

I'm not confortable with `crontab -l | sed -r "\\#FreshRSS#s#^[^ ]+ #$CRON_MIN #" | crontab -`.
Is there a way to use CRON with this line with hours too ? (the best would be to be compatible of course).

With this line, I actually get:
```
freshrss-cron    | Result: success
freshrss-cron    | "-":0: bad command
freshrss-cron    | errors in crontab file, can't install.
```

Thanks.

Alkarex:
While waiting for something better, here are a few comments:

* Run your cron much more often (e.g. 30min), but use FreshRSS global settings as well as feeds' settings to define how often they actually need to be refreshed.

* Users are randomized before actualizing the feeds, and the order of the feeds is not always the same, so that we do not always hit the servers at the same time.

* Use `sleep` in front of your cron command to give additional time offset. Side note: you can even use it to randomize further, e.g.

```sh
sleep `shuf -i 1-59 -n 1`
```

aledeg:
What should we do about that issue? What do we want to implement and/or document?

Alkarex:
@aledeg Both things can be considered. We could consider exposing another environment variable with the full cron config, with precedence over `CRON_MIN`. On the negative side, I am not sure how many would use that, and it might be confusing to have several cron settings. We also have the possibility of running more advanced cron job on the host, which are already documented https://github.com/FreshRSS/FreshRSS/tree/master/Docker#cron-job-to-automatically-refresh-feeds

Alkarex:
Implemented by a volume with https://github.com/FreshRSS/FreshRSS/pull/3927

PR #2954, opened 2020-05-06
Fix tty error on cron with docker

Pofilo:
In #2953, I reported that we can't use crons with hours. But there is still the solution to use the crontab of the host.
This PR is a correction of a very small typo causing: `the input device is not a TTY`.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -213,7 +213,7 @@ Remember not pass the `CRON_MIN` environment variable to your Docker run, to avo
 Example on Debian / Ubuntu: Create `/etc/cron.d/FreshRSS` with:
 
 ```
-7,37 * * * * root docker exec --user www-data -i freshrss php ./app/actualize_script.php > /tmp/FreshRSS.log 2>&1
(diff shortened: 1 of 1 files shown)


Docker/README.md
```diff
@@ -213,7 +213,7 @@ Remember not pass the `CRON_MIN` environment variable to your Docker run, to avo
 Example on Debian / Ubuntu: Create `/etc/cron.d/FreshRSS` with:
 
 ```
-7,37 * * * * root docker exec --user www-data -it freshrss php ./app/actualize_script.php > /tmp/FreshRSS.log 2>&1
+7,37 * * * * root docker exec --user www-data -i freshrss php ./app/actualize_script.php > /tmp/FreshRSS.log 2>&1
```
Alkarex:
```suggestion
7,37 * * * * root docker exec --user www-data freshrss php ./app/actualize_script.php > /tmp/FreshRSS.log 2>&1
```

I believe we can remove the `-i` as well

Pofilo: True, nobody is waiting in front of a cron routine :)

Alkarex (approved):

Alkarex:
Please put a little line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

PR #2955, opened 2020-05-06
Attempt to reduce max memory usage during actualize

### LINKED ISSUE #2952: actualize_script.php is overloading the server
Hello,

I have a small Debian GNU/Linux 10 (buster) server (RAM : 512Mo / SWAP : 512Mo).

Since I have updated from FreshRSS 1.14.2 to FreshRSS 1.16.0, the `actualize_script.php` is overloading my server : full memory and full swap. The server is unusable.

The script is killed by the kernel, after some times :

```
kernel: [  653.943405] Out of memory: Kill process 2226 (php) score 786 or sacrifice child
kernel: [  653.943423] Killed process 2226 (php) total-vm:950028kB, anon-rss:356572kB, file-rss:0kB, shmem-rss:0kB
kernel: [  654.240137] oom_reaper: reaped process 2226 (php), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

The strange thing is that with the actualize button on the web interface, everything is working (feeds are correctly actualized) and the server is not overused at all.

For my bigger user, I have 550 feeds, 50000 articles and 300 MiB of data.

I also noticed a lot of FreshRSS debug lines now in /var/log/syslog :

```
FreshRSS[2521]: FreshRSS GET https://www.xxx
FreshRSS[2514]: SimplePie MD5 cache match for https://www.xxx
FreshRSS[2518]: SimplePie GET https://www.xxx
```

I temporarily disabled the cron task until I got a solution.

I can reproduce the problem just by running `php app/actualize_script.php`. I can see the memory growing and when full, the swap is growing until being also full.

How can I solve this problem ?

Alkarex:
#Fix https://github.com/FreshRSS/FreshRSS/issues/2952

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -333,9 +333,9 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 
 			$needFeedCacheRefresh = false;
 
-			// We want chronological order and SimplePie uses reverse order.
-			$entries = array_reverse($feed->entries());
-			if (count($entries) > 0) {
+			$entries = $feed->entries();
+			$nbEntries = count($entries);
+			if ($nbEntries > 0) {
 				$newGuids = array();
 				foreach ($entries as $entry) {
 					$newGuids[] = safe_ascii($entry->guid());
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
I hope all the unsetting won't slow it down too much?

Alkarex:
> I hope all the unsetting won't slow it down too much?

I do not think it will make any measurable difference. We unset about two variables per entry, and the processing speed of each feed and entries is mostly defined by the HTTP request for the feed, followed by DB inserts for each entry. This is something the garbage collector is supposed to do anyway - we do it just a bit earlier and aggressively.

PR #2958, opened 2020-05-08
Add lang and mail in CLI user-info

Alkarex:
And update documentation after
https://github.com/FreshRSS/FreshRSS/pull/2296
Related to https://github.com/FreshRSS/FreshRSS/pull/2936

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -65,7 +65,7 @@ cd /usr/share/FreshRSS
 # -h is to use a human-readable format
 # --header outputs some columns headers
 # --user indicates a username, and can be repeated
-# Returns: 1) a * iff the user is admin, 2) the name of the user,
+# Returns: 1) a * if the user is admin, 2) the name of the user,
 #  3) the date/time of last user action, 4) the size occupied,
 #  and the number of: 5) categories, 6) feeds, 7) read articles, 8) unread articles, 9) favourites, 10) tags,
 #  11) language, 12) e-mail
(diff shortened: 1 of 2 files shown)


Alkarex:
```
cli/user-info.php -h --header --user alex --user alice
default | user                 | last user activity        | space used      | categories | feeds      | reads      | unreads    | favourites | tags       | lang  | email
*       | alex                 | 2020-05-08T08:34:03+02:00 | 396.06 MiB      | 20         | 195        | 124175     | 3202       | 275        | 18         | fr    | alexandre@alapetite.fr
        | alice                | 2020-01-06T20:40:41+01:00 | 216.00 KiB      | 1          | 1          | 2          | 9          | 0          | 0          | fr    |
Result: success
```

cli/README.md
```diff
@@ -61,12 +61,14 @@ cd /usr/share/FreshRSS
 ./cli/list-users.php
 # Return a list of users, with the default/admin user first
 
-./cli/user-info.php -h --user username
+./cli/user-info.php -h --header --user username1 --user username2 ...
 # -h is to use a human-readable format
-# --user can be a username, or '*' to loop on all users
+# --header outputs some columns headers
+# --user indicates a username, and can be repeated
 # Returns: 1) a * iff the user is admin, 2) the name of the user,
```
aledeg: there is an extra ''f'' in ''iff''. You may consider to change that as well in your PR

aledeg (approved):

cli/README.md
Alkarex: It was originally on purpose https://en.wikipedia.org/wiki/Iff
but you are not the first one thinking it was an error, so it might be best to remove indeed

Frenzie: I probably only know about if and only if because my high school math teacher mentioned it at some point. I wouldn't be inclined to think it was on purpose, generally speaking.

aledeg: I didn't know about that! Thank you for pointing that out.

Alkarex: Done :-)

Issue #2960, opened 2020-05-08
Reeder 4 not syncing in latest docker build (1.16.1-dev)


lx-s:
Reeder 4 on iOS doesn't sync unread or unstarred items anymore (but also doesn't show an error and can log in normally). Reverting to `1.16.0` fixes this behaviour.

I've looked through commits since 1.16.0 and noticed https://github.com/FreshRSS/FreshRSS/commit/f0db70dcb03cef0c53db0ee0b678a64b79c941df and https://github.com/FreshRSS/FreshRSS/commit/0fc71fdf42d537d996609f2d6a19803f7274a1bd. Could one of those two be the culprit?

Do you need any further information from me?

Thanks for your help!

Alkarex:
Do you connect through the Google Reader API?
Could you please ensure that you have the very latest Docker build? (you can check the image ID)
I cannot test myself, but another user just reported that it works https://github.com/FreshRSS/FreshRSS/issues/2956#issuecomment-625314921

lx-s:
**API:**
Reeder 4 uses the Google-Reader API as backend (I selected "FreshRSS" during account setup and Reeder 4 automatically configured ```https://example.tld/api/greader.php/``` as API backend to my FreshRSS instance).

**Image-ID**
I am not sure, where to find the image ID, might this be it?
```
{
  "freshrss/freshrss": {
    "latest":"ddc53d37f56b27fa39d67410dc2767438acd6e8d0d8caed9fc6bb8bf9d86de91"
  }
}
```
It is the ```freshrss:latest```-image from today though.

Alkarex:
To check the version details, you can try something like:

```sh
docker inspect --format='{{.Name}} {{.Image}} {{index .Config.Labels "org.opencontainers.image.revision"}}' freshrss
```

Alkarex:
Can I send you a test account so you can try with Reeder 4 on my account?

lx-s:
> Docker

Unfortunately I won’t be able to check this weekend.

> Can I send you a test account so you can try with Reeder 4 on my account?

Of course :)

Alkarex:
@lx-s I have sent you a test account by e-mail

Alkarex:
New image available on Docker Hub

Alkarex:
Seems to work again :-) Isn't it @lx-s ?

lx-s:
> Seems to work again :-) Isn't it @lx-s ?

[...]

PR #2961, opened 2020-05-08
Fix bad relogging clients

Alkarex:
Some clients (like EasyRSS) are still using the old HTTP Authorization
header after having logged to log in with another user.
We should not attempt to process Authorization headers during a login
request


Alkarex:
Merging in master to ease debugging , e.g. https://github.com/FreshRSS/FreshRSS/issues/2960

PR #2962, opened 2020-05-10
add lazy loading on favicon

upskaling:
chargement des favicon en différé

### DIFF
--- a/app/layout/aside_feed.phtml
+++ b/app/layout/aside_feed.phtml
@@ -94,7 +94,7 @@
 						<a class="dropdown-toggle" data-fweb="<?= $feed->website() ?>"><?= _i('configure') ?></a>
 						<?php /* feed_config_template */ ?>
 					</div>
-					<?php if (FreshRSS_Context::$user_conf->show_favicons): ?><img class="favicon" src="<?= $feed->favicon() ?>" alt="✇" /><?php endif; ?>
+					<?php if (FreshRSS_Context::$user_conf->show_favicons): ?><img class="favicon" src="<?= $feed->favicon() ?>" alt="✇" loading="lazy" /><?php endif; ?>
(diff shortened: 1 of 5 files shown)


app/layout/aside_feed.phtml
```diff
@@ -94,7 +94,7 @@
 						<a class="dropdown-toggle" data-fweb="<?= $feed->website() ?>"><?= _i('configure') ?></a>
 						<?php /* feed_config_template */ ?>
 					</div>
-					<?php if (FreshRSS_Context::$user_conf->show_favicons): ?><img class="favicon" src="<?= $feed->favicon() ?>" alt="✇" /><?php endif; ?>
+					<?php if (FreshRSS_Context::$user_conf->show_favicons): ?><img class="favicon" src="<?= $feed->favicon() ?>" alt="✇" loading="lazy" /><?php endif; ?>
```
Frenzie, line 97: Does it actually make a difference here given how it doesn't depend on display:none? I ask mainly because the articles themselves are already lazy loaded.

Alkarex:
For reference:
* https://caniuse.com/#feat=loading-lazy-attr
* https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes
* https://github.com/scott-little/lazyload
* https://developer.mozilla.org/docs/Web/HTML/Element/iframe#attr-loading

app/layout/aside_feed.phtml
Alkarex: In the mobile view, this part is now shown, for instance. And in the main view, when categories are collapsed, it might work as well (not tested yet)

Alkarex (approved):
Looks good, and I see no drawback of starting experimenting with this attribute

Alkarex:
Merci @upskaling 
Est-ce vous pourriez ajouter une ligne pour vous dans https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md ?

PR #2964, opened 2020-05-10
Improve logs during cron actualization

Alkarex:
For
https://github.com/FreshRSS/FreshRSS/issues/2952#issuecomment-626218921
And fix a little bug writing logs with the wrong user name

### DIFF
--- a/app/actualize_script.php
+++ b/app/actualize_script.php
@@ -2,17 +2,27 @@
 <?php
 require(__DIR__ . '/../cli/_cli.php');
 
+/**
+ * Writes to FreshRSS admin log, and if it is not already done by default,
+ * writes to syslog (only if simplepie_syslog_enabled in FreshRSS configuration) and to STDERR
+ */
+function notice($message) {
+	Minz_Log::notice($message, ADMIN_LOG);
+	if (!COPY_LOG_TO_SYSLOG && SIMPLEPIE_SYSLOG_ENABLED) {
+		syslog(LOG_NOTICE, $message);
+	}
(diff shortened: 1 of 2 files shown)


lib/Minz/Log.php
```diff
@@ -55,7 +55,8 @@ public static function record ($information, $level, $file_name = null) {
 				$level_label = 'debug';
 				break;
 			default :
-				$level_label = 'unknown';
+				$level = LOG_INFO;
```
Alkarex, line 58: Not something we actually use, but just for code robustness, as we use `$level` a few lines below, and it is important that it is a valid value.

app/actualize_script.php
```diff
@@ -43,44 +54,35 @@
...
-		fwrite(STDOUT, 'Actualize ' . $user . "...\n");	//Unbuffered
-	}
-	echo $user, ' ';	//Buffered
-
 
 	Minz_Session::_param('currentUser', $user);
 	new Minz_ModelPdo($user);	//TODO: FIXME: Quick-fix while waiting for a better FreshRSS() constructor/init
 	FreshRSS_Auth::giveAccess();
 	$app->init();
+	notice('FreshRSS actualize ' . $user . '...');
```
Alkarex, line 65: This line is moved further down, after user initialisation, otherwise the log line was prefixed with the wrong username (from the previous loop)

Alkarex:
Requires this fix https://github.com/FreshRSS/FreshRSS/pull/2963 to work fully well

PR #2976, opened 2020-05-15
Reformat i18n correctly

marienfressinaud:
**This PR should take precedence over the others that modify i18n files** (there's none at the moment).

Changes proposed in this pull request:

- Reformat i18n files (they were all badly formatted and the i18n make commands were useless)
- Add a simple `make i18n-update-key` command (it was useful to myself during this PR so I included the command in the PR)
- Make i18n formatting command case-sensitive so it stops to invert lines at each call
- Make Travis to check the files are correctly formatted

Some translations are lost, in particular the `shift_for_all_read` which was only half translated due to the last change. Unfortunately, since the strings were different, the formatter considered the strings to be completely translated.

There are still few things to do such as few `// TODO`s that should be ignored, some unused keys and documentation to write, but I propose to merge this PR quickly to avoid merge conflicts with potential upcoming PRs.

How to test the feature manually:

There's nothing to really test, but you can verify the application is still correctly translated ;).

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated: I'll do it this afternoon, but I think I'll open a different PR

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/master/docs/en/developers/04_Pull_requests.md).

PR #2978, opened 2020-05-15
Add documentation for i18n

marienfressinaud:
Follow-up of #2976. **Only the last commit needs to be reviewed**, I didn't want the documentation to block the merge of the other PR ;).

Add documentation about internationalization. Documentation already existed in `docs/en/developers/03_Backend/05_Extensions.md` but it was hard to find and incomplete.

Read the doc, fix typos and check the links.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/docs/en/internationalization.md
+++ b/docs/en/internationalization.md
@@ -58,7 +58,7 @@ The command will create a new subfolder under `app/i18n/` and copy the strings f
 
 ## Translate the interface
 
-You might have noticed some strings are not translated in the interface. These strings are usually in English while you chose a different language. This is because we mostly speak English or French and it's pretty difficult to us to speak all the different languages!
+You might have noticed some strings are not yet translated from English even though you've selected a different language. This is because we mostly speak English or French and it's pretty difficult to us to speak all the different languages!
 
 To update a string, you just have to open its file, find the string, and change it (without removing the quotes around it!) You might want to remove the comment at the end of the line, but you should prefer to use the following command:
 
--- a/docs/en/internationalization.md
+++ b/docs/en/internationalization.md
@@ -1,0 +1,132 @@
+# Contributing to internationalization (i18n)
+
+Thanks to different contributors, FreshRSS is actually translated in 15 languages. This section will explain you the basis of the internationalization in FreshRSS, either if you want to translate the application in your own language or if you have to make a specific change.
+
+## Overview
+
+It is common (and that's an understatement) to want to show some text to the user. The problem is that FreshRSS has users of different nationalities. It is therefore necessary to be able to manage different languages in order not to remain confined to English or French.
+
+The solution is to use the `Minz_Translate` module, which allows dynamic translation of FreshRSS. Before using this module, it is necessary to know where to find the strings to be translated. Each language has its own subdirectory in a parent directory named `app/i18n/`. For example, English language files are located in [`app/i18n/en/`](/app/i18n/en/). There are seven different files:
+
+* `admin.php` for anything related to FreshRSS administration
+* `conf.php` for configuration
+* `feedback.php` contains translations of feedback messages
+* `gen.php` stores what is global to FreshRSS (`gen` stands for “general”)
+* `index.php` for the main page that lists feeds and the About page
+* `install.php` contains strings related to the installation
+* `sub.php` for subscription management (`sub` stands for “subscription”)
+* `user.php` contains some strings related to the User model
+
+This organization makes it possible to avoid a single huge translation file.
+
+The translation files are quite simple: it's only a matter of returning a PHP array containing the translations. As an example, here's an extract from [`app/i18n/fr/gen.php`](/app/i18n/fr/gen.php):
+
+```php
+<?php
+
+return array(
+	'action' => [
+		'actualize' => 'Actualiser',
+		'back_to_rss_feeds' => '← Retour à vos flux RSS',
+		'cancel' => 'Annuler',
+		'create' => 'Créer',
+		'disable' => 'Désactiver',
+	),
+	'freshrss' => array(
+		'_' => 'FreshRSS',
+		'about' => 'À propos de FreshRSS',
+	),
+	// ...
+];
+```
+
+Each value can be referenced by a key: it consists of a series of identifiers separated by dots. The first identifier indicates from which file to extract the translation, while the following ones indicate array entries. Thus, the `gen.freshrss.about` key is referencing the `about` entry from the `freshrss` entry which is part of the main array returned by the `gen.php` file. This allows us to further organize our translation files.
+
+You should not have to write the array by yourself and we provide several commands to ease the manipulation of these files. Let's see some common use cases.
+
+## Add support for a new language
+
+If you want to add support for your own language which is not supported by FreshRSS yet, you can call this command:
+
+```console
+$ make i18n-add-language lang=[your language code]
+```
+
+You must replace `[your language code]` by the language tag of your language. It must follow the [IETF BCP 47 standard](https://en.wikipedia.org/wiki/IETF_language_tag). For instance, English is `en` and French is `fr`. You can target a specific region with a subtag, for instance `pt-br` for Brazilian Portuguese. If your not sure of the code, Wikipedia might be a good start to find it or you can ask us for help too.
+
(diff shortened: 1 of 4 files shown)


Alkarex (approved):

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
+# Contributing to internationalization (i18n)
+
+Thanks to different contributors, FreshRSS is actually translated in 15 languages. This section will explain you the basis of the internationalization in FreshRSS, either if you want to translate the application in your own language or if you have to make a specific change.
```
Frenzie: Do we really want to mention a specific number? ;-) Also "actually" doesn't seem appropriate, it should be left out or replaced by something like "actively."

```suggestion
Thanks to our contributors, FreshRSS is translated into more than 15 languages. This section will explain the basics of internationalization in FreshRSS, from translating the application to your own language to making a specific change.
```

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Overview
+
+It is common (and that's an understatement) to want to show some text to the user. The problem is that FreshRSS has users of different nationalities. It is therefore necessary to be able to manage different languages in order not to remain confined to English or French.
+
+The solution is to use the `Minz_Translate` module, which allows dynamic translation of FreshRSS. Before using this module, it is necessary to know where to find the strings to be translated. Each language has its own subdirectory in a parent directory named `app/i18n/`. For example, English language files are located in [`app/i18n/en/`](/app/i18n/en/). There are seven different files:
```
Frenzie, line 9: Does that link work from freshrss.github.io?

Frenzie (commented):
Leaving a comment early because GH is bugging me about changing diffs :-P

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+];
+```
+
+Each value can be referenced by a key: it consists of a series of identifiers separated by dots. The first identifier indicates from which file to extract the translation, while the following ones indicate array entries. Thus, the `gen.freshrss.about` key is referencing the `about` entry from the `freshrss` entry which is part of the main array returned by the `gen.php` file. This allows us to further organize our translation files.
+
+You should not have to write the array by yourself and we provide several commands to ease the manipulation of these files. Let's see some common use cases.
+
+## Add support for a new language
+
+If you want to add support for your own language which is not supported by FreshRSS yet, you can call this command:
```
Frenzie:
```suggestion
If you want to add support for a language which isn't supported by FreshRSS yet, you can run this command:
```

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+
+```console
+$ make i18n-add-language lang=[your language code]
+```
+
+You must replace `[your language code]` by the language tag of your language. It must follow the [IETF BCP 47 standard](https://en.wikipedia.org/wiki/IETF_language_tag). For instance, English is `en` and French is `fr`. You can target a specific region with a subtag, for instance `pt-br` for Brazilian Portuguese. If your not sure of the code, Wikipedia might be a good start to find it or you can ask us for help too.
```
Frenzie:
```suggestion
You must replace `[your language code]` by the language tag of your language. It must follow the [IETF BCP 47 standard](https://en.wikipedia.org/wiki/IETF_language_tag). For instance, English is `en` and French is `fr`. You can target a specific region with a subtag, for instance `pt-br` for Brazilian Portuguese. If you're not sure of the code, Wikipedia might be a good start to find it or you can ask us for help too.
```

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+
+The command will create a new subfolder under `app/i18n/` and copy the strings from the reference language (i.e. English). It will also mark all the translations with a special tag represented by a comment: `// TODO - Translation`. We'll see in the next section how to translate the strings.
+
+## Translate the interface
+
+You might have noticed some strings are not translated in the interface. These strings are usually in English while you chose a different language. This is because we mostly speak English or French and it's pretty difficult to us to speak all the different languages!
```
Frenzie:
```suggestion
You might have noticed some strings are not translated in the interface. These strings are usually in English if you've selected a different language. This is because we mostly speak English or French and it's pretty difficult to us to speak all the different languages!
```

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+
+To update a string, you just have to open its file, find the string, and change it (without removing the quotes around it!) You might want to remove the comment at the end of the line, but you should prefer to use the following command:
+
+```console
+$ make i18n-format
+```
+
+It will remove the comments on the lines that you've changed, and will reformat the file correctly. If you've made some mistakes, it will fix them by itself or it will tell you it can't (well… the command will dramatically fail without any damage, don't worry).
```
Frenzie:
```suggestion
It will remove the comments on the lines that you've changed, and will reformat the file correctly. If you've made any mistakes, it will fix them automatically or it will tell you it can't (well… the command will dramatically fail without any damage, don't worry).
```

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+
+This adds the key to all the files. It’ll be in English, waiting for other translators.
+
+Conversely, you may want to remove a key that is no longer used in the application with:
+
+```console
+$ make i18n-remove-key key=the.key.to.remove
+```
+
+Finally, if the English version of a string needs to be changed, you need to consider two cases. If the change doesn't impact the meaning of the sentence, and therefore other languages doesn't need to change (e.g. to fix a typo), you should make the change manually in the file. In the other case, you should use this command:
```
Frenzie:
```suggestion
Finally, if the English version of a string needs to be changed, you need to consider two cases. If the change doesn't impact the meaning of the sentence, and therefore other languages don't need to change (e.g. to fix a typo), you should make the change manually in the file. In any other case, you should use the following command:
```

docs/en/internationalization.md
```diff
@@ -0,0 +1,132 @@
...
+
+To access these translations, you must use the `_t()` function (which is a shortcut for `Minz_Translate::t()`). Code example:
+
+```html
+<p>
+	<?= _t('gen.freshrss.about') ?>
+</p>
+```
+
+The function expects a translation key, but there's a special case that sometimes makes life easier: the `_` identifier. This must necessarily be present at the end of the chain and gives a value to the higher-level identifier. It's pretty hard to explain but very simple to understand. In the example given above, a `_` is associated with the value `FreshRSS`: this means that there is no need to write `_t('gen.freshrss._')` but `_t('gen.freshrss')` suffices.
```
Frenzie:
```suggestion
The function expects a translation key, but there's a special case that sometimes makes life easier: the `_` identifier. This must necessarily be present at the end of the chain and gives a value to the higher-level identifier. It's pretty hard to explain but very simple to understand. In the example given above, an `_` is associated with the value `FreshRSS`: this means that there is no need to write `_t('gen.freshrss._')` but `_t('gen.freshrss')` suffices.
```

Frenzie:
@marienfressinaud It wasn't fully clear to me what was old & new; please ignore my comments if it's all old and then it should ideally be applied afterward so that po4a can handle the update better.

docs/en/internationalization.md
marienfressinaud: > Do we really want to mention a specific number? ;-) 

I didn't want to but since it's translated exactly into 15 languages, I had some remorse to write "more than 15 languages"… But we just need to find someone to add another translation and we'll be good :p

docs/en/internationalization.md
marienfressinaud: I hope so: it works in the GitHub interface so I guess it'll work on github.io too

docs/en/internationalization.md
marienfressinaud: I think this change has not the same sense than my initial sentence. Maybe `whereas` instead of `if`?

marienfressinaud:
> @marienfressinaud It wasn't fully clear to me what was old & new; please ignore my comments if it's all old and then it should ideally be applied afterward so that po4a can handle the update better.

Most of your comments were about new documentation, except at least one comment ("a `_`" to "an `_`" so I think it's not a big deal). Btw, I really appreciate your feedback: some of my mistakes were careless mistakes, but I always learn something :)

docs/en/internationalization.md
Frenzie:
```suggestion
You might have noticed some strings are not yet translated from English even though you've selected a different language. This is because we mostly speak English or French and it's pretty difficult to us to speak all the different languages!
```

PR #2990, opened 2020-05-20
Better hide spinner

### LINKED ISSUE #2986: Dark theme: Loader.gif never disappears
Hello,

I recently setup FreshRSS on my server with the latest Docker image, with the version 1.16.1-dev of FreshRSS.
I notice when I'm on a page where there is at least 1 stream, the animated loading gif is always visible on the list. Inside DOM, the concerned `div` element is `id="first_load"`.
I guess the class `loading` is never changed and so the loading gif stays always.

Tested on Firefox 68.8.0 ESR and Chromium 80.0.x, on a Debian Buster, without AdBlocker.

Regards.

### LINKED ISSUE #2992: small display bug
FreshRSS Version 1.16.1-dev
Firefox and firefox dev (latest version in windows)

bug

since a few updates I have this little bug (only when all items are marked as seen).

Have a nice day.

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/2986#issuecomment-631396735
By CSS, and also by JS 🙈

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1526,6 +1526,13 @@ function faviconNbUnread(n) {
 	}
 }
 
+function removeFirstLoadSpinner() {
+	const first_load = document.getElementById('first_load');
+	if (first_load) {
+		first_load.remove();
+	}
+}
+
 function init_normal() {
 	const stream = document.getElementById('stream');
 	if (!stream) {
@@ -1556,6 +1563,7 @@ function init_beforeDOM() {
 }
 
 function init_afterDOM() {
+	removeFirstLoadSpinner();
 	init_notifications();
 	init_popup();
 	init_confirm_action();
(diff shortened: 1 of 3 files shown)


Alkarex:
@nicofrand Could you please give it a try?

p/scripts/main.js
```diff
@@ -1579,11 +1587,9 @@ init_beforeDOM();	//Can be called before DOM is fully loaded
 if (document.readyState && document.readyState !== 'loading') {
 	init_afterDOM();
 } else {
-	document.addEventListener('DOMContentLoaded', function () {
-			if (window.console) {
-				console.log('FreshRSS waiting for DOMContentLoaded…');
```
Alkarex, line 1584: Minor: Moved the log message outside of the callback so that it is printed immediately

p/themes/base-theme/template.css
```diff
@@ -591,12 +592,17 @@ input[type="search"] {
 
 /*=== Body */
 #global {
+	background: inherit;
```
Alkarex, line 595: The various `background: inherit;` are to avoid transparency, which make it difficult to hide background elements

p/scripts/main.js
```diff
@@ -1556,6 +1563,7 @@ function init_beforeDOM() {
 }
 
 function init_afterDOM() {
+	removeFirstLoadSpinner();
```
Alkarex, line 1566: Should not be needed for our core themes, but might be needed for other themes or extensions.

Alkarex:
@nicofrand Please consider adding the following CSS rule to your extension (so that the spinner will be hidden by CSS before JS)

```css
#threepanesviewcontainer {
	background: inherit;
}
```

nicofrand:
Thanks, that fixes it for me, with or without extension! (My extension only works with JS anyway, but I'll add the CSS property indeed).

PR #2993, opened 2020-05-21
Fix fetch preview

### LINKED ISSUE #2923: ⓘ Reload articles and fetch complete content - HTTP ERROR 500
I'm trying out selectors to pull truncated content, but I can't really tell if I'm doing it right.

I'm trying to get the complete comment thread from github issues using maybe `discussion_bucket` or `js-discussion`

no matter what I enter, if I click the eye next to that, I just get a blank popout window.. then if I go to the bottom of the page and click "Reload articles" I get http error 500.

any insight would be appreciated
thank you.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2923
In MariaDB / MySQL, we cannot start a new query if we have not consumed
the previous buffered query fully.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -700,7 +700,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		//Extract all feed entries from database, load complete content and store them back in database.
 		$entries = $entryDAO->listWhere('f', $feed_id, FreshRSS_Entry::STATE_ALL, 'DESC', 0);
 
-		//We need another DB connexion in parallel
+		//We need another DB connection in parallel
 		Minz_ModelPdo::$usesSharedPdo = false;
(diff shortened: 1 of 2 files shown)


app/Controllers/feedController.php
```diff
@@ -755,8 +755,9 @@ public function contentSelectorPreviewAction() {
 
 		//Get first entry (syntax robust for Generator or Array)
 		foreach ($entries as $myEntry) {
-			$entry = $myEntry;
-			break;
```
Alkarex, line 759: We must not break because we need to consume the result of the buffered query fully before starting a new one.

Alkarex:
Ah, one more fix needed for "reload"

Alkarex:
Done

Leepic (approved):
It works for me (reload + preview). Thank you!

PR #2996, opened 2020-05-22
tec: Add a make lint command in the Makefile

### LINKED ISSUE #2722: Add make lint and make test targets

marienfressinaud:
Closes #2722 (I almost forget about this ticket :))

Note: this PR is not urgent so it can wait for the 1.16.1 to be released.

- Add `php7-xmlwriter` and `php7-tokenizer` to the Alpine image (required by PHP_CodeSniffer)
- Exclude some config files and extensions path from the linter
- Add `make lint` and `make lint-fix` commands to the Makefile

Execute `make lint` and `make lint-fix` in a console, and verifies it downloads the binaries and executes the linter.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] unit tests written (optional if too hard) (N/A)
- [x] documentation updated

### DIFF
--- a/Docker/Dockerfile-Alpine
+++ b/Docker/Dockerfile-Alpine
@@ -5,7 +5,7 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 RUN apk add --no-cache \
 	apache2 php7-apache2 \
 	php7 php7-curl php7-gmp php7-intl php7-mbstring php7-xml php7-zip \
-	php7-ctype php7-dom php7-fileinfo php7-iconv php7-json php7-opcache php7-phar php7-session php7-simplexml php7-xmlreader php7-zlib \
+	php7-ctype php7-dom php7-fileinfo php7-iconv php7-json php7-opcache php7-phar php7-session php7-simplexml php7-xmlreader php7-xmlwriter php7-tokenizer php7-zlib \
 	php7-pdo_sqlite php7-pdo_mysql php7-pdo_pgsql
 
 RUN mkdir -p /var/www/FreshRSS /run/apache2/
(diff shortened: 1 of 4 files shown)


Frenzie (approved):
Looks good enough to me for basic use.

Alkarex (approved):

PR #2999, opened 2020-05-23
Fix JS show password in install

Alkarex:
The button for showing the password was not working in the install (null exception).
Updated JavaScript file to ES6 at the same time.

### DIFF
--- a/p/scripts/install.js
+++ b/p/scripts/install.js
@@ -1,31 +1,31 @@
 // @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-3.0
 "use strict";
-/* jshint globalstrict: true */
+/* jshint esversion:6, strict:global */
 
 function show_password(ev) {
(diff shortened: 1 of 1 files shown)


p/scripts/install.js
```diff
@@ -1,31 +1,31 @@
 // @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-3.0
 "use strict";
-/* jshint globalstrict: true */
+/* jshint esversion:6, strict:global */
 
 function show_password(ev) {
-	var button = ev.target;
-	var passwordField = document.getElementById(button.getAttribute('data-toggle'));
+	const button = ev.currentTarget;
```
Alkarex, line 6: `ev.target` is the image, while the one we want is the button in `ev.currentTarget`

Alkarex: I had to double-check the doc :-P
https://developer.mozilla.org/docs/Web/API/Event/currentTarget

PR #3003, opened 2020-05-24
Add auto-registration when using http_auth

bpatath:

When using `http_auth`, the app trusts the web server to perform the authentication. Thus, it should also trust the web server to only pass valid users. Currently, users must be manually created by the admin before they can log in. This PR adds an option so that, when a non-existing user authenticate using `http_auth`, the user is automatically created and logged in.

1. Use a working `http_auth` based setup
2. Set `http_auth_auto_register` to `true` in your config file
3. Log in to a non-existing FreshRSS user
4. The user should be created and logged in, instead of showing a permission error

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


Alkarex:
Thanks @bpatath , that sounds fair enough.
I will have a look after the release of FreshRSS 1.16.1 (in the coming days)

app/Models/Auth.php
```diff
@@ -62,7 +62,17 @@ private static function accessControl() {
...
 			$current_user = httpAuthUser();
-			$login_ok = $current_user != '' && FreshRSS_UserDAO::exists($current_user);
+			if ($current_user == '') {
+				return false;
+			}
+			$login_ok = FreshRSS_UserDAO::exists($current_user);
+			if (!$login_ok && $conf->http_auth_auto_register) {
+				$email = null;
+				if ($conf->http_auth_auto_register_email_field !== '') {
+					$email = $_SERVER[$conf->http_auth_auto_register_email_field];
```
Alkarex:
```suggestion
				if ($conf->http_auth_auto_register_email_field != '' &&
					isset($_SERVER[$conf->http_auth_auto_register_email_field])) {
					$email = $_SERVER[$conf->http_auth_auto_register_email_field];
```

We probably need to check that this key exists

Alkarex:
@plopoyop You are using HTTP Auth in YunoHost; do you have any feedback on this user auto-create feature? (probably not useful for YunoHost but you might have an opinion anyway)

Alkarex:
@bpatath Could you please add a little note in our documentation? https://github.com/FreshRSS/FreshRSS/blob/master/docs/en/admins/09_AccessControl.md

plopoyop:
> probably not useful for YunoHost

We are calling freshrss user creation script every time a  user is created in yunohost.
Auto registration would be great.

Alkarex:
Ok @plopoyop so as this pull request is, would it be directly helpful, or do you need anything else?

marienfressinaud:
Do we need an option for that? It seems fine to me to have the feature always enabled

Alkarex:
> Do we need an option for that? It seems fine to me to have the feature always enabled

Indeed, as we can whitelist / blacklist users at Web server level (e.g. `.htaccess`), but I was afraid I might had overlooked an undesirable security implication

Frenzie:
That seems reasonable given the context, but I'm not quite sure.

marienfressinaud:
Maybe if a server has several users but the sysadmin wants to restrict who can access FRSS? Indeed we can keep the option for this reason (but is a sysadmin really manage its users this way? :p) It looks more admin-friendly to switch it on by default though

Edit: sorry, misclick

Alkarex:
> Maybe if a server has several users but the sysadmin wants to restrict who can access FRSS?

For that an `allow` or `deny` style of rule at Web server level can also do the job

Alkarex:
@bpatath I have made a few changes. Could you please double-check that it still works for you?

app/Models/Auth.php
```diff
@@ -47,8 +47,8 @@ public static function init() {
 	 * @return boolean true if user can be connected, false else.
 	 */
 	private static function accessControl() {
-		$conf = Minz_Configuration::get('system');
-		$auth_type = $conf->auth_type;
+		FreshRSS_Context::$system_conf = Minz_Configuration::get('system');
```
Alkarex, line 50: I will make another PR with a better refactoring of this part, but which requires a few global changes.

Alkarex:
Let's continue testing in /master

app/Models/Auth.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3070

PR #3013, opened 2020-05-29
Fix PostgreSQL database auto-create with limited rights

### LINKED ISSUE #3009: Postgres connection error in Heroku
Hi, I could use some help debugging this connection error.

The error at setup for either the CLI or web install is consistently

> Blast! Verify your database information. : Access to database is denied for `[username omitted]`: SQLSTATE[08006] [7] FATAL: permission denied for database "postgres" DETAIL: User does not have CONNECT privilege.

psql command with the same database connection details works. I've also added a `pgtest.php` file with
```
<?php
    $connection = pg_connect (getenv("DATABASE_URL"));
    if($connection) {
       echo 'connected';
    } else {
        echo 'there has been an error connecting';
    } 
?>
```
When I load the page it is successful.

Step 2 of the install is all ok - modules are loaded.

If I understand FreshRSS uses PDO and not the pgsql extension. I have `pdo_pgsql` and `pgsql` modules loaded.

This test `pdotest.php` also returns connected
```
<?php
    $db = parse_url(getenv("DATABASE_URL"));

    $pdo = new PDO("pgsql:" . sprintf(
        "host=%s;port=%s;user=%s;password=%s;dbname=%s",
        $db["host"],
        $db["port"],
        $db["user"],
        $db["pass"],
        ltrim($db["path"], "/")
    ));

    if($pdo) {
       echo 'connected';
    } else {
        echo 'there has been an error connecting';
    }
?>
```

If it helps I can also share access to `phpinfo.php` and whatever else will help project developers. 
Thank you for any information on what may be wrong.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3009
Install would fail if the user is not even allowed to connect to the default `postgres` database, because this is what is done to attempt to auto-create the FreshRSS database, so we need to catch this error.

### DIFF
--- a/app/Models/DatabaseDAO.php
+++ b/app/Models/DatabaseDAO.php
@@ -21,7 +21,7 @@ class FreshRSS_DatabaseDAO extends Minz_ModelPdo {
 		try {
 			$sql = sprintf($SQL_CREATE_DB, empty($db['base']) ? '' : $db['base']);
 			return $this->pdo->exec($sql) !== false;
-		} catch (PDOException $e) {
+		} catch (Exception $e) {
 			$_SESSION['bd_error'] = $e->getMessage();
 			syslog(LOG_DEBUG, __method__ . ' warning: ' . $e->getMessage());
 			return false;
@@ -34,7 +34,7 @@ class FreshRSS_DatabaseDAO extends Minz_ModelPdo {
 			$stm = $this->pdo->query($sql);
 			$res = $stm->fetchAll(PDO::FETCH_COLUMN, 0);
 			return $res != false;
-		} catch (PDOException $e) {
+		} catch (Exception $e) {
 			$_SESSION['bd_error'] = $e->getMessage();
 			syslog(LOG_DEBUG, __method__ . ' warning: ' . $e->getMessage());
 			return false;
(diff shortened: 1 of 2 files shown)


lib/lib_install.php
```diff
@@ -94,13 +94,19 @@ function initDb() {
 		//For first connection, use default database for PostgreSQL, empty database for MySQL / MariaDB:
 		$db['base'] = $db['type'] === 'pgsql' ? 'postgres' : '';
 		$conf->db = $db;
-		//First connection without database name to create the database
-		$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
+		try {
+			//First connection without database name to create the database
+			$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
```
Alkarex, line 99: This raises an exception if the user does not have CONNECT privilege:

> SQLSTATE[08006] [7] FATAL: permission denied for database "postgres"

app/Models/DatabaseDAO.php
```diff
@@ -21,7 +21,7 @@ public function create() {
 		try {
 			$sql = sprintf($SQL_CREATE_DB, empty($db['base']) ? '' : $db['base']);
 			return $this->pdo->exec($sql) !== false;
-		} catch (PDOException $e) {
+		} catch (Exception $e) {
```
Alkarex, line 24: At the time of the writing, I visibly got confused by / forgot about the custom Minz_PDOConnectionException type, so the specific native catch was not working (I am personally not a fan of custom Exception types).

Alkarex: Here, we would like to catch any type of error anyway

PR #3015, opened 2020-05-30
Fix alert-warn background

Alkarex:
Fix minor CSS regression from
https://github.com/FreshRSS/FreshRSS/pull/2990

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -384,6 +384,10 @@ a.btn {
 	width: 90%;
 }
 
+.alert-warn {
+	background: inherit;
+}
+
 .group-controls .alert {
 	width: 100%
 }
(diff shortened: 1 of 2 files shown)


p/themes/base-theme/template.css
```diff
@@ -604,7 +608,7 @@ input[type="search"] {
 	table-layout: fixed;
 }
 
-#stream {
+#stream:not(.alert-warn) {
 	background: inherit;
```
Alkarex, line 612: This rule otherwise had priority in other themes overriding `.alert-warn  { background: XXX }`

Alkarex:
Before the fix:
image

After the fix:
image

p/themes/base-theme/template.css
```diff
@@ -384,6 +384,10 @@ a.btn {
 	width: 90%;
 }
 
+.alert-warn {
+	background: inherit;
+}
```
marienfressinaud, line 389: Why do we need these lines? The `#stream:not(.alert-warn)` seems to be enough, isn't it?

Alkarex: Yes indeed. It was just to make sure to have a background in all cases

PR #3018, opened 2020-05-30
Fix display_categories upgrade from 1.16.0

### LINKED ISSUE #2915: User Interface Error with recent Git checkout
When I expand a category with unread articles, I see all feeds with unread articles.

When I click on one feed with unread articles, the category collapses again.

This a new behaviour, before the update I kept seeing all feeds with unread items in the expanded category.

### LINKED ISSUE #3017: 1.16.1: reading a feed folds its category in the left pane
I upgraded my 1.16.0 version to 1.16.1 (server: Debian linux, Client Firefox 76.0.1 unbuntu), and found an UI bug that made me downgrade to 1.16.0:

When unfolding a category in the left column (CSS class: "tree-folder category"), clicking a feed to read it instantly folds back the category.
Reverting back to 1.16.0 restored the normal behaviour of leaving the category unfolded.

It happens with Firefox (76.0.1 (64-bit)) 
and Chrome (Version 81.0.4044.138 (Official Build) (64-bit)), both on Ubuntu Linux.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3017
The property `'display_categories' => boolean,` in 1.16.0 is not properly converted to `'display_categories' => 'string',` in 1.16.1 due to a weak type comparison bug.

### DIFF
--- a/app/Models/ConfigurationSetter.php
+++ b/app/Models/ConfigurationSetter.php
@@ -183,7 +183,7 @@ class FreshRSS_ConfigurationSetter {
 	}
 
 	private function _display_categories(&$data, $value) {
-		if (!in_array($value, [ 'active', 'all', 'none' ])) {
+		if (!in_array($value, [ 'active', 'all', 'none' ], true)) { 
 			$value = $value === true ? 'all' : 'active';
 		}
 		$data['display_categories'] = $value;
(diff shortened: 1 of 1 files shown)


app/Models/ConfigurationSetter.php
```diff
@@ -183,7 +183,7 @@ private function _show_fav_unread(&$data, $value) {
 	}
 
 	private function _display_categories(&$data, $value) {
-		if (!in_array($value, [ 'active', 'all', 'none' ])) {
+		if (!in_array($value, [ 'active', 'all', 'none' ], true)) {
```
Alkarex: Second parameter to enforce strict type comparison

PR #3027, opened 2020-06-02
Change add feed action

### LINKED ISSUE #777: UX très frustrante pour l'ajout d'un nouveau flux
Il faut SEPT clicks pour ajouter un flux en lui attribuant une catégorie.
A la longue c'est très frustrant.
Peut être le principal problème est la taille du champ d'ajout beaucoup trop petit, et qui se redimensionne lorsque il a le focus. C'est assez insupportable et fait qu'on est obligé de cliquer dans le vide pour pouvoir cliquer sur le bouton de validation ou pour dérouler le drop down de choix d'une catégorie...
Peut être que cela a déja changé dans la version dev ? Je ne suis que sur la branche beta.

aledeg:
Closes #777

- Add a dedicated page to add a feed to avoid multiple clicks.
- Change the subscription page by removing the add feed action and reducing the size of the add category action.
- Change main page to add direct access to the add feed action.

1. Add a simple feed
2. Add a category
3. Add a feed with a new category
4. Add a feed with HTTP authentication

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

To ease the use of this feature, there is a new dedicated page to avoid numerous clicks.
Especially for switching categories or using HTTP authentication. The underlying actions
remain identical but the interface is simplified.

### DIFF
--- a/app/layout/aside_feed.phtml
+++ b/app/layout/aside_feed.phtml
@@ -14,7 +14,7 @@
 	<?php if (FreshRSS_Auth::hasAccess()) { ?>
 	<div class="stick configure-feeds">
 		<a id="btn-subscription" class="btn btn-important" href="<?= _url('subscription', 'index') ?>"><?= _t('index.menu.subscription') ?></a>
-		<a id="btn-importExport" class="btn btn-important" href="<?= _url('importExport', 'index') ?>"><?= _i('import') ?></a>
+		<a id="btn-add" class="btn btn-important" href="<?= _url('subscription', 'add') ?>"><?= _i('add-white') ?></a>
 	</div>
 	<?php } elseif (FreshRSS_Auth::accessNeedsLogin()) { ?>
 	<a href="<?= _url('index', 'about') ?>"><?= _t('index.menu.about') ?></a>
--- a/app/views/subscription/add.phtml
+++ b/app/views/subscription/add.phtml
@@ -10,7 +10,7 @@
 		<div class="form-group">
 			<label class="group-name" for="url_rss"><?= _t('sub.feed.url') ?></label>
 			<div class="group-controls">
-				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off"/>
+				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long"/>
 			</div>
 		</div>
 
--- a/app/views/subscription/add.phtml
+++ b/app/views/subscription/add.phtml
@@ -1,0 +1,91 @@
+<?php $this->partial('aside_subscription'); ?>
+
+<div class="post drop-section">
+    <a href="<?= _url('index', 'index') ?>"><?= _t('gen.action.back_to_rss_feeds') ?></a>
+
+	<h2><?= _t('sub.title.add_feed') ?></h2>
+    <form id="add_rss" method="post" action="<?= _url('feed', 'add') ?>" autocomplete="off">
+		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
+
+		<div class="form-group">
+			<label class="group-name" for="url_rss"><?= _t('sub.feed.url') ?></label>
+			<div class="group-controls">
+				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off"/>
+			</div>
+		</div>
+
+		<div class="form-group">
+			<label class="group-name" for="category"><?= _t('sub.category') ?></label>
+			<div class="group-controls">
+                <select name="category" id="category">
+                <?php foreach ($this->categories as $cat) { ?>
+                <option value="<?= $cat->id() ?>"<?= $cat->id() == 1 ? ' selected="selected"' : '' ?>>
+                    <?= $cat->name() ?>
+                </option>
+                <?php } ?>
+                </select>
+			</div>
+		</div>
+
+        <legend><?= _t('sub.feed.auth.configuration') ?></legend>
+		<div class="form-group">
+			<label class="group-name" for="http_user_feed"><?= _t('sub.feed.auth.username') ?></label>
+			<div class="group-controls">
+				<input id="http_user_feed" name="http_user_feed" type="text" autocomplete="off"/>
+			</div>
+		</div>
+
+		<div class="form-group">
+        <label class="group-name" for="http_pass_feed"><?= _t('sub.feed.auth.password') ?></label>
+			<div class="group-controls">
+				<input id="http_pass_feed" name="http_pass_feed" type="text" value=" " autocomplete="new-password"/>
+			</div>
+		</div>
+
+		<?php if (FreshRSS_Auth::hasAccess('admin')) { ?>
+		<legend><?= _t('sub.feed.advanced') ?></legend>
+		<div class="form-group">
+			<label class="group-name" for="timeout"><?= _t('sub.feed.timeout') ?></label>
+			<div class="group-controls">
+				<input type="number" name="timeout" id="timeout" min="3" max="120" value="" placeholder="<?= _t('gen.short.by_default') ?>" />
+			</div>
+		</div>
+
+		<div class="form-group">
+			<label class="group-name" for="ssl_verify"><?= _t('sub.feed.ssl_verify') ?></label>
+			<div class="group-controls">
+				<label class="checkbox" for="ssl_verify">
+					<select name="ssl_verify" id="ssl_verify">
+						<option value=""><?= _t('gen.short.by_default') ?></option>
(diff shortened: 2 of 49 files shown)


aledeg:
UX is definitely not my strong suit so please guide me to have something usable

Alkarex:
I will check more tonight, but looks good 👍 
Maybe you could add a few more advanced parameters, which otherwise might prevent the feed from being added (e.g. timeout, SSL):
image

Furthermore, maybe we should have a checkbox for adding `#force_feed`

There is a minor layout glitch in at least one theme:
image

mlaily:
If you want feedback, here is mine, as the original issue creator.

I like the dedicated "Add a feed" page. 🙂
The option to create a new category directly from this page is not necessary in my opinion, bu it's really not a problem.

I find the "Subscription management" page a bit misleading now, because the only action is the creation of a category, which is rarely used.
It's also immediately under the "Subscription management" header, and not clearly separated from the feeds. (It was in some kind of titled box before. Did you change that because it was taking too much space?)

- Maybe add a link/button to the feed creation page? Or maybe even a simple line of text indicating there is a dedicated page to subscribe to a feed? (I know it's also in the left menu, but I think it would help to see it on the page, especially for long time user that expect to find the feed subscription button here)
- Maybe add a legend or some kind of header to the category creation? I think it would be better to have something like "Add/Create a new category" instead of just the watermark in an isolated text input. It could stay all on one like if you think it's better.
- Maybe add some kind of separation (a simple `<hr/>` ?) between the new category input and the feeds list.

Sorry for the lengthy feedback. I might be a bit too nit-picking, but it could be the only chance I get in 5 years, so I hope you forgive me. 😀

Thank you for your work.

aledeg:
@Alkarex I do not have this problem on my machine. I am using FF on Arch. Even on smaller screens, there is no such thing.
I do not want to include more fields since they are not required to work. Basically, I've only moved the original form in its own page. This new page is only meant to add the feed. Once it's done, you are automatically redirected to its configuration where you can tweak the feed configuration.

@mlaily I hear you. I've moved the category creation in the feed creation page where I've separated the 2 forms. I've also renamed the page to show that there is one page to create the elements.
I've also added a small information box containing a link to the new page.

mlaily:
I like your changes.

Something seems broken when adding a feed though.

image

```
172.17.0.1 - - [02/Jun/2020:19:14:25 +0000] "GET /i/?c=subscription&a=add HTTP/1.1" 200 2672 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:78.0) Gecko/20100101 Firefox/78.0"
[Tue Jun 02 19:14:34.139830 2020] [php7:notice] [pid 18] [client 172.17.0.1:41786] PHP Notice:  Undefined variable: Minz_Request in /var/www/FreshRSS/app/Controllers/feedController.php on line 166
[Tue Jun 02 19:14:34.139936 2020] [php7:error] [pid 18] [client 172.17.0.1:41786] PHP Fatal error:  Uncaught Error: Class name must be a valid object or a string in /var/www/FreshRSS/app/Controllers/feedController.php:166\nStack trace:\n#0 /var/www/FreshRSS/lib/Minz/Dispatcher.php(119): FreshRSS_feed_Controller->addAction()\n#1 /var/www/FreshRSS/lib/Minz/Dispatcher.php(47): Minz_Dispatcher->launchAction('addAction')\n#2 /var/www/FreshRSS/lib/Minz/FrontController.php(84): Minz_Dispatcher->run()\n#3 /var/www/FreshRSS/p/i/index.php(48): Minz_FrontController->run()\n#4 {main}\n  thrown in /var/www/FreshRSS/app/Controllers/feedController.php on line 166
172.17.0.1 - - [02/Jun/2020:19:14:33 +0000] "POST /i/?c=feed&a=add HTTP/1.1" 200 336 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:78.0) Gecko/20100101 Firefox/78.0"
```

aledeg:
Thank you. I'll look at it

aledeg:
@mlaily it's fixed.

mlaily:
I did some more testing.

As I said, I like it like that. The only thing I find strange is that validation (popup messages for an invalid url, an empty category name...) is done on the subscriptions page, and not the page where the inputs are now...

aledeg:
Good catch! I didn't noticed before. I'll look into it

aledeg:
@mlaily it's fixed!

mlaily:
That was quick! :) Everything looks fine by me.

Alkarex:
@aledeg For the layout bug, try with Pafat theme in English.
image
or Blue Lagoon
image

To solve the bug and simplify the UI, I think you could consider dropping one of the icons (e.g. the one for import/export, and keep only the one to add a new feed).

For the additional fields, it can also wait for another PR, because it is not something we have now anyway. The reason for exposing timeout and SSL (and `#force_feed`) at subscription time is that otherwise it is not possible to add a feed that needs those parameters without changing global parameters, because adding the feed will fail before allowing the user to change those parameters.

See this example of feed that takes 30s to load http://test.alapetite.fr/slow.rss.xml.php and try to add it to FreshRSS (which, by default, has 15s timeout)

Minor: Maybe "Add" instead of "Create" for the button label:
image
(It took be a couple of seconds to understand where I should click to proceed with adding the feed)

aledeg:
@Alkarex I've removed the import button and change the action button text. For the other fields, I'll rework it when they are available.

Alkarex:
@aledeg The other fields are already available, but currently only from a feed configuration page (so we have the text, parameters, etc.)

aledeg:
Ok, I'll look into it tomorrow.

aledeg:
@Alkarex I've added timeout and ssl. I could not find the force option.

app/layout/aside_feed.phtml
```diff
@@ -14,7 +14,7 @@
 	<?php if (FreshRSS_Auth::hasAccess()) { ?>
 	<div class="stick configure-feeds">
 		<a id="btn-subscription" class="btn btn-important" href="<?= _url('subscription', 'index') ?>"><?= _t('index.menu.subscription') ?></a>
-		<a id="btn-importExport" class="btn btn-important" href="<?= _url('importExport', 'index') ?>"><?= _i('import') ?></a>
```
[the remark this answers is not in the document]
aledeg: @Alkarex done

Alkarex:
The feed attributes were not applied / used before loading. Fixed by https://github.com/FreshRSS/FreshRSS/pull/3027/commits/52ba7ff4c6912c80d285384d7138e2013fae8a2b

Alkarex (approved):

Alkarex:
@marienfressinaud Do you want to have a look at the new UI?

p/themes/Swage/swage.rtl.css
```diff
@@ -63,7 +63,7 @@ a.btn:hover {
 a#btn-subscription {
   width: 76%;
 }
-a#btn-importExport {
```
Alkarex, line 66: Ah, it needs to be changed in the `*.scss` files as well (Ansum, Mapco, Swage themes)

aledeg: done

app/views/subscription/add.phtml
```diff
@@ -0,0 +1,91 @@
...
+    <a href="<?= _url('index', 'index') ?>"><?= _t('gen.action.back_to_rss_feeds') ?></a>
+
+	<h2><?= _t('sub.title.add_feed') ?></h2>
+    <form id="add_rss" method="post" action="<?= _url('feed', 'add') ?>" autocomplete="off">
+		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
+
+		<div class="form-group">
+			<label class="group-name" for="url_rss"><?= _t('sub.feed.url') ?></label>
+			<div class="group-controls">
+				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off"/>
```
marienfressinaud:
```suggestion
				<input id="url_rss" class="long" name="url_rss" type="url" required="required" autocomplete="off"/>
```

marienfressinaud (commented):
Maybe the "login" and "advanced" sections can be merged together, and hidden by default? It looks like they are required and make the form more complex than it is

marienfressinaud:
I don't have a lot of time this week to spend on the review but it's a nice improvement and I trust you for the rest of the review :)

aledeg:
@marienfressinaud the idea is to keep the layout from the add page and the layout from the update panel with the same fields and headers organization. For sure, the add page is a skimmed version of the update page but I find that hiding elements will be counter-intuitive.

PR #3032, opened 2020-06-04
Return the correct Content-Type for SVG favicons

### LINKED ISSUE #2572: SVG feed favicons aren't displayed correctly
Recently one of sites I'm subscribing updated its favicon, switching from PNG to SVG file. FreshRSS correctly downloaded new favicon:

```
[.../FreshRSS/data/favicons] # file 61f3eba2.ico 
61f3eba2.ico: SVG Scalable Vector Graphics image
```
however the favicon is not displayed correctly. When opening its URL directly, Firefox shows _"The image https://.../freshrss/f.php?61f3eba2 cannot be displayed because it contains errors"_. Opera fails in similar way.

I've also discovered bug cause: it's serving SVG image with "Content-Type: image/x-icon" HTTP header. Changing to "image/svg+xml" in the f.php makes both browsers display favicon correctly (but obviously breaks all other favicon types... at least in Opera). Probably instead of hardcoding the content type header, some sort of dynamic discovery should be implemented. 

PS. feed in question for problem reproduction: https://www.chip.pl/feed/

pcaro90:
Closes #2572

- Before displaying the favicon, check if the file type is "image/svg" using `mime_content_type`.
- If so, change the returned Content-Type HTTP header to "image/svg+xml".

1. Add a feed with a SVG favicon, such as https://github.com/FreshRSS/FreshRSS/releases
2. Before the fix, SVG favicons show up as the default favicon. After the fix, they should appear as the actual favicon.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/f.php
+++ b/p/f.php
@@ -48,5 +48,13 @@ if ($ico_mtime == false || $ico_mtime < $txt_mtime || ($ico_mtime < time() - (mt
 header('Content-Disposition: inline; filename="' . $id . '.ico"');
 
 if (!httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
-	readfile($ico);
+
+    // In case the favicon is a svg file, replace the Content-Type header
+    $ico_content = file_get_contents($ico);
+    if (substr($ico_content, 0, 4) === '<svg') {
+            header('Content-Type: image/svg+xml');
+    }
+
+    echo $ico_content;
+
 }
--- a/p/f.php
+++ b/p/f.php
@@ -5,6 +5,7 @@ require(LIB_PATH . '/favicons.php');
 require(LIB_PATH . '/http-conditional.php');
 
 function show_default_favicon($cacheSeconds = 3600) {
+	header('Content-Type: image/x-icon');
 	header('Content-Disposition: inline; filename="default_favicon.ico"');
 
 	$default_mtime = @filemtime(DEFAULT_FAVICON);
--- a/p/f.php
+++ b/p/f.php
@@ -50,11 +50,11 @@ header('Content-Disposition: inline; filename="' . $id . '.ico"');
 if (!httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
 
     // In case the favicon is a svg file, replace the Content-Type header
-    $ico_content = file_get_contents($ico);
-    if (substr($ico_content, 0, 4) === '<svg') {
-            header('Content-Type: image/svg+xml');
-    }
+	$ico_content_type = mime_content_type($ico);
+	if ($ico_content_type === 'image/svg') {
+		header('Content-Type: image/svg+xml');
+	}
 
-    echo $ico_content;
+	readfile($ico);
 
 }
--- a/p/f.php
+++ b/p/f.php
@@ -49,7 +49,7 @@ header('Content-Disposition: inline; filename="' . $id . '.ico"');
 
 if (!httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
 
-    // In case the favicon is a svg file, replace the Content-Type header
+	// In case the favicon is a svg file, replace the Content-Type header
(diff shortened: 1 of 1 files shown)


Frenzie (commented):
Did you check what happens with PNG favicons (of which there are plenty)?

pcaro90:
> Did you check what happens with PNG favicons (of which there are plenty)?

Sure, they remain unaffected.

aledeg:
Can't you use https://www.php.net/manual/en/function.mime-content-type.php?

pcaro90:
> Can't you use https://www.php.net/manual/en/function.mime-content-type.php?

Actually, that could be a much more complete solution. It could be (maybe?) less efficient than `substr`, but given that favicons are very small files, it wouldn't matter that much.

However, would you just relay on `mime_content_type` for every ico file, sending the result as `Content-Type`, or would you just check if the file is a SVG file?

Frenzie:
@pcaro90 
> Sure, they remain unaffected.

I think you answered a different question than I intended to ask. I wasn't asking if this PR breaks PNG support (which wouldn't make sense), but whether PNG was working at all. So browsers don't care when PNG is given an `image/x-icon` mimetype?

pcaro90:
I just tested `mime_content_type`, which returns `image/svg` for SVG icons, instead of `image/svg+xml`, which should be the correct type for the browsers to handle. So, it's cleaner using `mime_content_type`, but it cannot be passed as the `Content-Type` header as-is. I'll use that function to detect the file type, instead of `substr`.

> I think you answered a different question than I intended to ask. I wasn't asking if this PR breaks PNG support (which wouldn't make sense), but whether PNG was working at all. So browsers don't care when PNG is given an `image/x-icon` mimetype?

Apparently... The only non-working favicons I've found are SVG. Not sure about how `image/x-icon` is handled by browsers.

p/f.php
```diff
@@ -48,5 +48,13 @@ function show_default_favicon($cacheSeconds = 3600) {
...
 if (!httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
+
+	// In case the favicon is a svg file, replace the Content-Type header
+	$ico_content_type = mime_content_type($ico);
+	if ($ico_content_type === 'image/svg') {
+		header('Content-Type: image/svg+xml');
+	}
+
 	readfile($ico);
+
```
Frenzie:
```suggestion
```

p/f.php
```diff
@@ -48,5 +48,13 @@ function show_default_favicon($cacheSeconds = 3600) {
 header('Content-Disposition: inline; filename="' . $id . '.ico"');
 
 if (!httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
+
```
Frenzie:
```suggestion
```

Frenzie (commented):
I'd leave out the extra whitespace. ^_^

Alkarex:
@pcaro90 Thanks 👍 I have made a few more changes https://github.com/FreshRSS/FreshRSS/pull/3032/commits/ad1366c71fba45e2196a4a0ffcf9d33765aaad10
Please double-check that it still works fine for you

pcaro90:
@Alkarex It works fine, except when the default favicon is served. In that case, the returned Content-Type is `text/html; charset=UTF-8` (that's why I left the `image/x-icon` Content-Type header untouched in the first place). Adding the correct Content-Type in the `show_default_favicon` function, it works perfectly.

Alkarex:
I have made a final change https://github.com/FreshRSS/FreshRSS/pull/3032/commits/52a89ddeb099c983f4aabe6fe39034309923130a

Alkarex (approved):

Alkarex:
@pcaro90 Thanks again 👍 Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

PR #3034, opened 2020-06-05
Add missing translations

aledeg:

- Add missing translation on the extension page

1. remove a system extension
2. remove a user extension

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/i18n/en/feedback.php
+++ b/app/i18n/en/feedback.php
@@ -36,7 +36,7 @@ return array(
 	),
 	'extensions' => array(
 		'already_enabled' => '%s is already enabled',
-		'cannot_delete' => '%s cannot be deleted',
+		'cannot_remove' => '%s cannot be removed',
 		'disable' => array(
 			'ko' => '%s cannot be disabled. <a href="%s">Check FreshRSS logs</a> for details.',
 			'ok' => '%s is now disabled',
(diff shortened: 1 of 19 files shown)


app/i18n/en/feedback.php
```diff
@@ -47,6 +48,7 @@
 		'no_access' => 'You have no access on %s',
 		'not_enabled' => '%s is not enabled',
 		'not_found' => '%s does not exist',
+		'removed' => '%s deleted',
```
Frenzie: Remove might be better than delete? But either way since they're new I'd keep the key and the text the same.

aledeg: Sure, I can change that.
The keys are not new, they've never been translated.

aledeg: done

aledeg:
I've added a quick method to be able to check if a key exists

app/i18n/en/feedback.php
```diff
@@ -36,6 +36,7 @@
 	),
 	'extensions' => array(
 		'already_enabled' => '%s is already enabled',
+		'cannot_delete' => '%s cannot be deleted',
```
Frenzie: I left the comment only on one line but I did mean it for both. ;-)

aledeg: What do you have in mind for that one?

Frenzie (approved):
lgtm :-)

PR #3037, opened 2020-06-06
Fix user queries when they contain "

aledeg:

- Fix user query when they contain "

1. search for a string containing a "
2. create a user query from it
3. save the queries created
4. before the fix, the search pattern disappear from the configuration

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/configure/queries.phtml
+++ b/app/views/configure/queries.phtml
@@ -15,7 +15,7 @@
 
 			<div class="group-controls">
 				<input type="hidden" id="queries_<?= $key ?>_url" name="queries[<?= $key ?>][url]" value="<?= $query->getUrl() ?>"/>
-				<input type="hidden" id="queries_<?= $key ?>_search" name="queries[<?= $key ?>][search]" value="<?= $query->getSearch() ?>"/>
+				<input type="hidden" id="queries_<?= $key ?>_search" name="queries[<?= $key ?>][search]" value="<?= urlencode($query->getSearch()) ?>"/>
 				<input type="hidden" id="queries_<?= $key ?>_state" name="queries[<?= $key ?>][state]" value="<?= $query->getState() ?>"/>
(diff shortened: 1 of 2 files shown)


Alkarex:
I believe we need something like `htmlspecialchars`.
Try with e.g. `<"te&st'>` and we get some unescaped chars in the DOM

aledeg:
ok, I'll try to fix that as well. I did not encounter that type of problem before. I'll keep you posted.

aledeg:
I cannot see the unescaped chars you mentioned. I've tried with the string you mentioned and everything is properly urlencoded.

Alkarex:
The HTML behind
image
looks like that:
image

Alkarex:
We can also produce an XSS (blocked thanks to our CSP rules) with `<script>alert('XSS');</script>`
image

aledeg:
I've checked and it did not trigger any XSS. But it was not displayed properly either. I fixed it anyway.

app/views/configure/queries.phtml
```diff
@@ -52,7 +52,7 @@
 
 					<ul>
 						<?php if ($query->hasSearch()) { ?>
-						<li class="item"><?= _t('conf.query.search', $query->getSearch()->getRawInput()) ?></li>
+						<li class="item"><?= _t('conf.query.search', htmlspecialchars($query->getSearch()->getRawInput())) ?></li>
```
Alkarex:
```suggestion
						<li class="item"><?= _t('conf.query.search', htmlspecialchars($query->getSearch()->getRawInput(), ENT_NOQUOTES, 'UTF-8')) ?></li>
```

Explicit encoding

Alkarex:
You need to disable our CSP protection to test XSS ;-)

PR #3045, opened 2020-06-09
tec: Refactor the export feature

### LINKED ISSUE #3035: OPML export has PHP warning at the bottom: "Cannot modify header information"
When I export my list of feeds (under Import/Export > Export list of feeds (OPML)), the resulting file has the following at the bottom:

`<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /var/www/freshrss/app/Controllers/importExportController.php:930) in <b>/var/www/freshrss/lib/Minz/ActionController.php</b> on line <b>63</b><br />
`

- Version 1.16.3-dev
- PHP 7.1.33
- MySQL

marienfressinaud:
Closes https://github.com/FreshRSS/FreshRSS/issues/3035

Even if the issue #3035 seemed pretty simple at a first glance, it was
more complicated than I expected. Because we send CSP headers AFTER
running the controller actions, it means we can't "echo" any content
from the controller. It's in fact a good practice, but it was easier at
the time we developed the feature.

To fix that, the only thing I had to do was to move the `print()` and
`readfile()` function into the view. The problem was that we needed to
output the content from the CLI too. Then, things became more
complicated. I decided to extract the export-related methods in a
`FreshRSS_Export_Service` class, in order to use it from both the
controller and the CLI. It was an opportunity to refactor the whole
feature in order to make it a bit more linear and easy to read.

1. Export OPML, labelled, starred, and feeds files (one by one, then several to generate a Zip) from the browser
2. Run the `php cli/export-zip-for-user.php` and `php cli/export-opml-for-user.php` commands
3. Verify none of the files contain an error and that they can be imported back to FRSS

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] N/A unit tests written (optional if too hard)
- [x] N/A documentation updated

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -721,76 +721,6 @@ class FreshRSS_importExport_Controller extends Minz_ActionController {
 	}
 
 	/**
-	 * Export the files of a user and send them to the user by HTTP
-	 *
-	 * @param string $username
-	 * @param boolean $export_opml
-	 * @param boolean $export_starred
-	 * @param boolean $export_labelled
-	 * @param array|boolean $export_feeds The list of feeds ids to export, true to export all the feeds
-	 * @param integer $maxFeedEntries Limit the number of entries to export (default is 50)
-	 *
-	 * @throws FreshRSS_ZipMissing_Exception if we try to export more than two files and the zip extension doesn't exist
-	 *
-	 * @return integer The number of exported files
-	 */
-	public function exportFile($username, $export_opml = true, $export_starred = false, $export_labelled = false, $export_feeds = array(), $maxFeedEntries = 50) {
-		require_once(LIB_PATH . '/lib_opml.php');
-
-		$this->catDAO = new FreshRSS_CategoryDAO($username);
-		$this->entryDAO = FreshRSS_Factory::createEntryDao($username);
-		$this->feedDAO = FreshRSS_Factory::createFeedDao($username);
-
-		if ($export_feeds === true) {
-			//All feeds
-			$export_feeds = $this->feedDAO->listFeedsIds();
-		}
-		if (!is_array($export_feeds)) {
-			$export_feeds = array();
-		}
-
-		$day = date('Y-m-d');
-
-		$export_files = array();
-		if ($export_opml) {
-			$export_files["feeds_${day}.opml.xml"] = $this->generateOpml();
-		}
-
-		if ($export_starred || $export_labelled) {
-			$export_files["starred_${day}.json"] = $this->generateEntries(
-				($export_starred ? 'S' : '') .
-				($export_labelled ? 'T' : '')
-			);
-		}
-
-		foreach ($export_feeds as $feed_id) {
-			$feed = $this->feedDAO->searchById($feed_id);
-			if ($feed) {
-				$filename = "feed_${day}_" . $feed->category() . '_'
-				          . $feed->id() . '.json';
-				$export_files[$filename] = $this->generateEntries('f', $feed, $maxFeedEntries);
-			}
-		}
-
-		$nb_files = count($export_files);
-		if ($nb_files > 1) {
-			// If there are more than 1 file to export, we need a ZIP archive.
-			$filename = 'freshrss_' . $username . '_' . $day . '_export.zip';
-			try {
-				$this->sendZip($filename, $export_files);
-			} catch (Exception $e) {
-				throw new FreshRSS_ZipMissing_Exception($e);
-			}
-		} elseif ($nb_files === 1) {
-			// Only one file? Guess its type and export it.
-			$filename = key($export_files);
-			$type = self::guessFileType($filename);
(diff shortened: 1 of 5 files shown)


app/Controllers/importExportController.php
```diff
@@ -720,76 +720,6 @@ private function addFeedJson($origin) {
 		return $return;
```
marienfressinaud, line 720: The diff concerning this file is a bit complicated to review, I suggest to only read the new file

app/Services/ExportService.php
```diff
@@ -0,0 +1,189 @@
...
+			$categories[$key]['name'] = $category->name();
+			$categories[$key]['feeds'] = $this->feed_dao->listByCategory($category->id());
+		}
+
+		$view->categories = $categories;
+
+		return [
+			"feeds_{$day}.opml.xml",
+			$view->helperToString('export/opml')
+		];
```
marienfressinaud, line 58: I'm not completely happy with this pattern (i.e. returning an array containing the filename and the content) but it makes the things easier in the controller and the cli

app/Services/ExportService.php
```diff
@@ -0,0 +1,189 @@
...
+		$view->categories = $this->category_dao->listCategories();
+		$day = date('Y-m-d');
+
+		$view->list_title = _t('sub.import_export.starred_list');
+		$view->type = 'starred';
+		$view->entriesId = $this->entry_dao->listIdsWhere(
+			$type, '', FreshRSS_Entry::STATE_ALL, 'ASC', -1
+		);
+		$view->entryIdsTagNames = $this->tag_dao->getEntryIdsTagNames($view->entriesId);
+		// The following is a streamable query, i.e. must be last
```
marienfressinaud, line 85: I kept this comment line, but I didn't understand its meaning

app/views/importExport/export.phtml
```diff
@@ -0,0 +1 @@
+<?= $this->content ?>
```
marienfressinaud, line 1: This is the real fix of #3035 :)

app/Controllers/importExportController.php
```diff
@@ -798,135 +728,113 @@ public function exportFile($username, $export_opml = true, $export_starred = fal
...
-
-		$nb_files = 0;
-		try {
-			$nb_files = $this->exportFile(
-				Minz_Session::param('currentUser'),
-				Minz_Request::param('export_opml', false),
-				Minz_Request::param('export_starred', false),
-				Minz_Request::param('export_labelled', false),
-				Minz_Request::param('export_feeds', array())
+			return Minz_Request::forward(
```
marienfressinaud, line 736: I just remembered I forgot to explain these `return`. It's actually useless because the `forward(..., true)` will immediately trigger a redirection, but it's implicit and not clear for a new developer. I suggest we start to put explicit `return` when we want the method to stop, what do you think?

app/Services/ExportService.php
marienfressinaud: No it's fine, it's just that I wasn't able to spot the difference between this method and the others. But I just realize that `fetchAll()` isn't called, and I didn't know that `PDOStatement` implemented the `Traversable` interface

Alkarex:
Looks good.
Not necessarily for this PR, but:

* Maybe we should have a default limit for the number of exported articles via the CLI when using the ZIP method (which is currently loading everything in memory and therefore only works for tiny cases):

```sh
docker exec --user www-data freshrss cli/export-zip-for-user.php --user alex
FreshRSS exporting ZIP for user “alex”…
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 290816 bytes) in /var/www/FreshRSS/app/Services/ExportService.php on line 175
```

* As it is (also) meant for other tools than FreshRSS, we should apply `trim(..., '; ')` in https://github.com/FreshRSS/FreshRSS/blob/7a748e25ab7187bba53decd2f41bd7b6383440f3/app/views/helpers/export/articles.phtml#L39

Alkarex (approved):

marienfressinaud:
> Maybe we should have a default limit for the number of exported articles via the CLI when using the ZIP method

Do you mean [the `$number_entries` variable](https://github.com/FreshRSS/FreshRSS/pull/3045/files#diff-111058ef1e0e909b6c54f9c3213bb771R25) (default is 100) or something else?

Alkarex:
Yes @marienfressinaud , this is what I mean. I did not check and - due to the memory problem I observed - assumed there was no default limit through the CLI. Now that we have the SQLite export (which uses streams), the ZIP export is less relevant, but I do not like the fact that it uses so much memory and crashes even for *light* exports. This might be out of scope for this PR though

Alkarex:
Suggestion for future work:
* Keep the OPML export as it is;
* Keep the JSON export of favourites / labels almost as it is, but modify it to stream the output;
* Promote the SQLite export/import so that is is accessible from the Web interface (in addition of CLI);
* Remove the ZIP export (but keep the ZIP import for now) as it is a bit redundant and would require more work to re-implement in a memory friendly way.

PR #3051, opened 2020-06-11
Fix API quickadd

### LINKED ISSUE #3049: [greader-api] subscription/quickadd
I think the `subscription/quickadd` is not really returning what it should return.
When looking at [feedhq](https://feedhq.readthedocs.io/en/latest/api/reference.html#subscription-quickadd)

the `quickadd` is not well documented, but currently it returns something like:
```
{
  "numResults": 1,
  "streamId": 225
}
```
looking all over the api-specs returning an integer here does not make sense at all since google reader ids are always like this: e.g. "feed/http://feeds.feedburner.com/alistapart/main"

aside from that, the `query` is completely omitted and should be added as well.

Sorry for being pedantic about the APIs, but currently the integer IDs does not make sense to me at all.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3049

* https://github.com/theoldreader/api#adding-subscription
* https://www.inoreader.com/developers/add-subscription
* https://feedhq.readthedocs.io/en/latest/api/reference.html#subscription-quickadd

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -392,10 +392,15 @@ function subscriptionEdit($streamNames, $titles, $action, $add = '', $remove = '
 function quickadd($url) {
 	try {
 		$url = htmlspecialchars($url, ENT_COMPAT, 'UTF-8');
+		if (substr($url, 0, 5) === 'feed/') {
+			$url = substr($url, 5);
+		}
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

Stunkymonkey:
works perfectly fine. :+1:

PR #3053, opened 2020-06-12
Feature/upgrade extension system

### LINKED ISSUE #2900: Button install extensions
Would it be possible to introduce the feature of downloading extensions in the appropriate folders by hitting a button in the frontend? It is rather cumbersome to install these by SSHing and also the updating of extensions is tedious. It should be easier, esp. as it is just downloading couple of files into a folder.

aledeg:
Closes #2900

- [ ] Update extension from github
- [ ] Install extension from github
- [x] Move extension actions in the table
- [x] Hide system extensions when not admin
- [x] Sort extension list
- [x] Add a status column in extension list

1. install extension
2. update extension
3. delete extension
4. configure extension

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/extensionController.php
+++ b/app/Controllers/extensionController.php
@@ -26,15 +26,45 @@ class FreshRSS_extension_Controller extends Minz_ActionController {
 		);
 
 		$this->view->extensions_installed = array();
+		$user_installed = array();
 
 		$extensions = Minz_ExtensionManager::listExtensions();
 		foreach ($extensions as $ext) {
 			$this->view->extension_list[$ext->getType()][] = $ext;
 			$this->view->extensions_installed[$ext->getEntrypoint()] = $ext->getVersion();
+			if ('user' === $ext->getType()) {
+				$user_installed[] = $ext->getEntrypoint();
+			}
 		}
 
-		$availableExtensions = $this->getAvailableExtensionList();
+		$availableExtensions = $this->getAvailableExtensionSortedList($user_installed);
 		$this->view->available_extensions = $availableExtensions;
+	}
+
+	/**
+	 * Returns the sorted list of extensions.
+	 * The order is based on status (installed or not) and name (alphabetical).
+	 */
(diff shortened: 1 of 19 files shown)


aledeg:
The install and update parts are missing for now. I need some guidance/idea on how to proceed. I am not sure what is the best way to handle that considering our code base.

Alkarex:
That is a nice feature @aledeg 👍 

Here is a bit of preliminary brainstorming:

* Consider where to automatically install those extensions:
  * We have `./extensions/` but only `./data/` is supposed to be writeable by the Web server (in installations, which do not perform Web-based updates).
  * Should we allow non-admin users to install their extensions in a user-specific extensions directory? Like `./data/users/alice/extensions/ `
  * Should we move `./extensions/` to `./data/extensions/` and allow only admins to install extensions there? Alternatively, update our guides and distributions to ask `./extensions/` to be writeable, but I am not so found of it.
* (Not necessarily for this PR): Reconsider where extensions are supposed to save their data
  * Check e.g. the existing `./data/extensions-data/`
  * Different for global or user extensions?
* How to download the extensions:
  * It would be so nice to have a git client guaranteed to be available. In absence of that, I think we need to download a ZIP archive of the whole repository containing the extension, e.g. https://github.com/aledeg/FreshRSS-extensions/archive/master.zip
  * We might have to include a hash of the ZIP in https://github.com/FreshRSS/Extensions/blob/master/extensions.json in addition of the version number.
  * As the ZIP might contain more than one extension, we can either delete the others or keep them not enabled

aledeg:
Thank you for your insight. I'll look into that later!

SuperSandro2000:
> Should we allow non-admin users to install their extensions in a user-specific extensions directory? Like ./data/users/alice/extensions/

This sounds like a security nightmare and extensions might require extra php modules. Enable on per user basis sounds fine but this not.

Alkarex:
@SuperSandro2000 You are right, until we have better permissions, that is premature

PR #3062, opened 2020-06-15
Update rules to use Make syntax

aledeg:

- Update rules to use Make syntax instead of using some kind of shell that could be not portable

1. run the modified rules without parameters

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,8 +93,7 @@ i18n-format: ## Format I18N files
 .PHONY: i18n-add-language
 i18n-add-language: ## Add a new supported language
 ifndef lang
-	@echo To add a new language, you need to provide one in the "lang" variable.
-	@exit 10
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Seems fine but I don't quite follow "some kind of shell that could be not portable" since this is all basic POSIX going back decades.

aledeg:
I might be over-dramatic but sometimes, there are discrepancies between systems. I've experienced it first-hand at work with one colleague using Mac. I don't remember the specific but when using only make syntax, it's more portable.
I might be wrong but that's my feeling.

Frenzie:
That should only be a problem of using shell-specific extensions (i.e., POSIX is common but something Bash-specific might not work in Zsh for example).

PR #3070, opened 2020-06-19
Avoid manual initialisations of system or user configuration

Alkarex:
More consistent use of Context. Code reduction.
Moved the complicated configuration initialisations to the Context object.
Promised in https://github.com/FreshRSS/FreshRSS/pull/3003#discussion_r439742363


lib/Minz/FrontController.php
```diff
@@ -31,9 +31,6 @@ class Minz_FrontController {
 	 */
 	public function __construct () {
 		try {
-			Minz_Configuration::register('system',
-			                             DATA_PATH . '/config.php',
-			                             FRESHRSS_PATH . '/config.default.php');
```
Alkarex, line 36: @marienfressinaud It was not so nice to initialise the system configuration here with a FreshRSS constant in Minz. What do you think?

app/Models/Context.php
```diff
@@ -76,6 +98,14 @@ public static function init() {
 		}
 	}
 
+	/**
+	 * Initialize the context for the global system and the current user
+	 */
+	public static function init() {
```
Alkarex: Deprecated. We can maybe remove it immediately.

lib/Minz/FrontController.php
marienfressinaud: I agree, an alternative would be to rename the constant to something like `ROOT_PATH` (or create one)

Alkarex:
@aledeg A bit of help to finalise this PR would be much welcome, if you are motivated enough :-)

In any case, the idea is simply to use `FreshRSS_Context` everywhere instead of a bit randomly as we did before.
The status is that it is almost done but needs some testing, corresponding fixes, and proof reading. As it spans over many files, the risk of conflicts with new PRs is high, so it would be nice to merge it.

Alkarex:
I am looking a bit more at it now

aledeg:
I've looked quickly at it (did not test) and it looks good to me.
I have a question though. Do you need to keep `FreshRSS_ConfigurationSetter` as there is only one call to it and you removed it?
I also notice that you sometimes use `==` and sometimes `===`. Is there a reason for that?

Apart from that, what help do you need?

Alkarex:
Ok, I will look at `FreshRSS_ConfigurationSetter`. The `== ''` is when I want to catch more largely (empty string, null, false...).
I think it is almost ready for a test (I have just found another little bug in the install process - I will change the status of the PR when fixed).
What would be nice would mainly be proof reading and tests (positive = everything should work, but also negative = denying access when not legitimate) of the different entry points and login types / variants (anonymous access...), and things related to user management (i.e. admin aspects such as creating users).

aledeg:
OK, let me know when ready.

Alkarex:
Ready, but best to merge https://github.com/FreshRSS/FreshRSS/pull/3305 first

app/Controllers/errorController.php
```diff
@@ -33,10 +33,12 @@ public function indexAction() {
 		case 500:
 			header('HTTP/1.1 500 Internal Server Error');
 			$this->view->code = 'Error 500 - Internal Server Error';
+			$this->view->errorMessage = 'Error 500 - Internal Server Error';
```
Alkarex, line 36: The view is using `errorMessage` (should have been another PR)

Alkarex:
I running this on my main instance; seems stable

Alkarex:
Looks all good in my tests. Let's continue testing in master 🤞🏻

PR #3068, opened 2020-06-19
[i18n] Update en-UK grammar and style

Frenzie:
Taken from <https://github.com/FreshRSS/FreshRSS/pull/3060>

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

---
This change is [Reviewable](https://reviewable.io/reviews/freshrss/freshrss/3068)

PR #3072, opened 2020-06-20
Change how updating a key works

aledeg:

- Update referential i18n values without losing translations

1. change a value in the referential with the make rule
2. see it changed in all untranslated files
3. see it stay unchanged in all translated files

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, to update a key from the reference language (en), the key was deleted
then created. This way was unproductive for all keys already translated.
Now, the key is updated in all untranslated languages.

See https://github.com/FreshRSS/FreshRSS/pull/3068#issuecomment-646868894

### DIFF
--- a/cli/i18n/I18nData.php
+++ b/cli/i18n/I18nData.php
@@ -182,7 +182,16 @@ class I18nData {
 		    !array_key_exists($key, $this->data[static::REFERENCE_LANGUAGE][$this->getFilenamePrefix($key)])) {
 			throw new Exception('The selected key does not exist for the selected language.');
 		}
-		$this->data[$language][$this->getFilenamePrefix($key)][$key] = $value;
+		if (static::REFERENCE_LANGUAGE === $language) {
+			$previousValue = $this->data[static::REFERENCE_LANGUAGE][$this->getFilenamePrefix($key)][$key];
+			foreach ($this->getAvailableLanguages() as $lang) {
+				if ($this->data[$lang][$this->getFilenamePrefix($key)][$key] === $previousValue) {
+					$this->data[$lang][$this->getFilenamePrefix($key)][$key] = $value;
+				}
+			}
+		} else {
+			$this->data[$language][$this->getFilenamePrefix($key)][$key] = $value;
+		}
 	}
 
 	/**
--- a/Makefile
+++ b/Makefile
@@ -125,8 +125,7 @@ endif
 ifndef value
 	$(error To update a key, you need to provide its value in the "value" variable)
 endif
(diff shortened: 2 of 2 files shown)


Frenzie (commented):
There seems to be something odd going on with Chinese?

aledeg:
What do you mean? I did not see anything suspicious.

Frenzie:
I mean like:
```sh
php cli/manipulate.translation.php -a add -k admin.user.admin -v bla -l en
```
The todos are updated as expected, but on top of that:
```diff
diff --git a/app/i18n/zh-cn/admin.php b/app/i18n/zh-cn/admin.php
index e880168b..0646aa08 100644
--- a/app/i18n/zh-cn/admin.php
+++ b/app/i18n/zh-cn/admin.php
@@ -182,7 +182,7 @@ return array(
                'title' => '更新系统',
        ),
        'user' => array(
-               'admin' => '管理员',
+               'admin' => 'bla',       // TODO - Translation
                'article_count' => '文章数',
                'articles_and_size' => '%s 篇文章（%s)',
                'back_to_manage' => '← 返回用户列表',
```

aledeg:
I'll check that!

aledeg:
@Frenzie Good catch! I didn't notice it when testing. Now it should work properly.

cli/i18n/I18nData.php
```diff
@@ -182,7 +182,16 @@ public function addValue($key, $value, $language) {
 		    !array_key_exists($key, $this->data[static::REFERENCE_LANGUAGE][$this->getFilenamePrefix($key)])) {
 			throw new Exception('The selected key does not exist for the selected language.');
 		}
-		$this->data[$language][$this->getFilenamePrefix($key)][$key] = $value;
+		if (static::REFERENCE_LANGUAGE === $language) {
+			$previousValue = $this->data[static::REFERENCE_LANGUAGE][$this->getFilenamePrefix($key)][$key];
```
Frenzie, line 186: I think that'd be more in line with the general style, unless I'm reading more into it than there is. (As in camelCase = function, object or method. More of a question for @Alkarex.)
```suggestion
			$previous_value = $this->data[static::REFERENCE_LANGUAGE][$this->getFilenamePrefix($key)][$key];
```

Frenzie (approved):
lgtm, seems to be working fine

cli/i18n/I18nData.php
Alkarex: I personally favour camelCase for new code (except when in the middle of another piece of code, which uses another convention), but we do have a bit of inconsistency :-)

Alkarex (approved):

Issue #3031, opened 2020-06-04
Reeder 4 hiccups and broken login


gruberb:
I am running into a few problems when using FreshRSS on a Ubuntu 18.04 installation with Reeder 4 for iOS.

The initial login and fetching the Feeds works. however:
- Adding a new feed within Reeder fails (error screen "could not add feed"). When I close and open the app, the feed is there
- I can't put new feeds into folder when adding them in Reeder and have to do so manually after I "fail added" (adding, error, closing app, opening app) the new feed
- After every refresh, I have to login manually again because the login fails ("Request Failed, Pleaes re-authorize this account")

There are no logs in `/var/log/apache2`.

Login via the browser, adding feeds etc. works perfectly fine.

Alkarex:
Hello,
Please check the logs in `./FreshRSS/data/users/_/log_api.txt` and share the parts related to those errors if you can.

To enable more logging (also of successful requests), you can uncomment the following two lines:
https://github.com/FreshRSS/FreshRSS/blob/27f0b614189c090028dfb9a56d1b21f1adf263d4/p/api/greader.php#L919-L920

I do not have any Apple product to test by myself.

ishanjain28:
I am running freshrss(commit, 0085b5e0c1eb8d9d50fc264c86865329c09e877e) on arch linux.

I haven't been able to replicate any of the issues. It seems to be working perfectly for me. 

@gruberb, Have you already tried clearing data/reinstalling the app?

gruberb:
Yeah the logs show no errors and I re-installed everything 3 or more times now.

Seems like an internal Reeder problem maybe. No apache logs and no Reeder logs.

Alkarex:
Do you say that you do not have any log at all in your Web server logs and API log? If that is really the case, it might be that the request does not even arrive

maurice-schleussinger:
I can confirm this. I just installed FreshRSS via Docker and added an account to the Reeder 4 Mac app. 

Adding feeds always works, but also always generates an error message.
After basically doing anything that changes feeds, folders and so on I get the alert that I need to login again and it fails. After restarting the app, everything works fine again.
**But**, I believe it only occurs if I am also logged in via web at the same time. I will continue to monitor this and report. 🙂

kour1er:
[...]

PR #3073, opened 2020-06-20
API Compatibility Reeder

Alkarex:
#fix
https://github.com/FreshRSS/FreshRSS/issues/3031#issuecomment-646973334
Reeder sends e.g. `feed/feed/247` instead of `feed/247`


Frenzie (approved):
Seems fine, the number doesn't cause any trouble further down?

Edit: never mind, I overlooked the `feed/feed` part; I thought it was about the `247` part.

PR #3081, opened 2020-06-25
Show feed name inside the article

### LINKED ISSUE #3079: Show website name when opening article
This CSS rule remove the website name, but it seems that no rule display it elsewhere  :

```
.item.website span
{
    display: none;
}
```

Expected : We can see the website name with mobile view

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3079


Alkarex:
Desktop (we could hide it, but it looks OK to me):
image

Mobile:

image

aledeg (approved):

app/views/index/normal.phtml
```diff
@@ -72,20 +72,26 @@ $today = @strtotime('today');
...
 						foreach ($authors as $author):
 							echo $first ? _t('gen.short.by_author') . ' ' : '· ';
 							$first = false;
 				?>
 <em><a href="<?= _url('index', 'index', 'search', 'author:' . str_replace(' ', '+', htmlspecialchars_decode($author, ENT_QUOTES))) ?>"><?= $author ?></a></em>
 				<?php	endforeach; ?>
-				</div><?php
-					endif;
+				</div><?php endif; ?>
+				<div class="text"><?php
```
Alkarex, line 95: I have added a new `<div class="text">...</div>` around the actual text of the article, to ease the layout (inlining) of our other information elements

p/themes/base-theme/template.rtl.css
```diff
@@ -719,7 +719,7 @@ input[type="search"] {
 }
 
 .flux .item.title .author {
-	padding-right: 1rem;
+	padding-left: 1rem;
```
Alkarex, line 722: @Frenzie I am guessing it was a bug (there was no margin in RTL mode)

Frenzie: Possibly (can't check atm), but then it's better to do it with a control directive like `/*rtl:ignore*/` in the original so the fix won't be overwritten if you run `make rtl`.

marienfressinaud (approved):

Alkarex:
A few more related comments https://github.com/FreshRSS/FreshRSS/issues/3079#issuecomment-650109551
Help welcome, otherwise I will try to find time in a couple of days

marienfressinaud:
I didn't see them. I can't work on this today but the suggestions are good :+1:

Alkarex:
Addressed https://github.com/FreshRSS/FreshRSS/issues/3079#issuecomment-650109551 but keeping the same design because authors can be something highly variable, as is typical for e.g. academic papers:
image

app/views/index/normal.phtml
```diff
@@ -72,20 +72,29 @@ $today = @strtotime('today');
...
-						$first = true;
-						foreach ($authors as $author):
-							echo $first ? _t('gen.short.by_author') . ' ' : '· ';
-							$first = false;
-				?>
-<em><a href="<?= _url('index', 'index', 'search', 'author:' . str_replace(' ', '+', htmlspecialchars_decode($author, ENT_QUOTES))) ?>"><?= $author ?></a></em>
-				<?php	endforeach; ?>
-				</div><?php
-					endif;
+				<div class="subtitle">
```
Alkarex, line 75: New subtitle section to ease styling

PR #3092, opened 2020-07-01
[CI] Update GH token authentication method

Frenzie:
See <https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/>.

---
This change is [Reviewable](https://reviewable.io/reviews/freshrss/freshrss/3092)


Alkarex (approved):

Frenzie:
@Alkarex I think there must be something wrong with the token. Maybe it's some kind of catch 22 and quoting the string makes it not interpret the variables?

Frenzie:
@Alkarex Yeah, definitely something wrong with the token. I just realized it works just fine on my own repo with my own token. https://travis-ci.org/github/Frenzie/FreshRSS/jobs/703851514
```
$ curl -H "Authorization: token $GITHUB_TOKEN" https://api.github.com/repos/hadolint/hadolint/releases/latest

{

  "message": "Bad credentials",

  "documentation_url": "https://developer.github.com/v3"

}
```

Alkarex:
What is the status @Frenzie ? Do you need anything?

Frenzie:
@Alkarex On second look, I think I can update the token myself. So I don't think so.

Frenzie:
@Alkarex Yeah no, I don't understand why it's not working here. As you can see it keeps running just fine on my own branch: https://travis-ci.org/github/Frenzie/FreshRSS/jobs/705553584

Frenzie:
As you can see the token's working in principle. (That's when you committed https://github.com/FreshRSS/FreshRSS/commit/cbcbd64acf4663f9f7edd7a7d6f136473cffe7b9) But I'm wary of just merging this in the hopes that it'll Just Work™.

Screenshot_2020-07-06_23-19-32

marienfressinaud:
Why do we need a token here? The Hadolint documentation doesn't mention it https://github.com/hadolint/hadolint/blob/master/docs/INTEGRATION.md#travis-ci

Frenzie:
@marienfressinaud I would be just fine with simply bumping the version number once in a blue moon, but when this was added it was deemed preferable to use the GH API to always grab the latest version. The GH API requires a token or it'll give Travis a rate limit (i.e., sometimes it'll succeed and sometimes it won't).

marienfressinaud:
Ok! Is the token the same between your configuration and the FRSS configuration? Maybe the later contains special characters which break the build?

Note that if you can't find a solution, I don't think there are high risks to pin the version number (rather the contrary). Our Dockerfiles don't change a lot and we probably don't need to have the latest version. Also, always getting the latest version adds a risk of a breaking change in the Hadolint API / configuration, which could break our builds.

Frenzie:
@marienfressinaud I tried the same token, different token, etc. No difference. I think it's Travis keeping the token out of PRs, as per https://travis-ci.community/t/allow-foreign-prs-to-use-secret-variables/8236

I'll update it to pin a version. I prefer that too.

.travis.yml
```diff
@@ -58,7 +58,8 @@ jobs:
         - HADOLINT="$HOME/hadolint"
       install:
         - npm install --save-dev jshint stylelint stylelint-order stylelint-scss stylelint-config-recommended-scss
-        - curl -sLo "$HADOLINT" $(curl -s https://api.github.com/repos/hadolint/hadolint/releases/latest?access_token="$GITHUB_TOKEN" | jq -r '.assets | .[] | select(.name=="hadolint-Linux-x86_64") | .browser_download_url') && chmod 700 ${HADOLINT}
+        - curl -sL -o ${HADOLINT} "https://github.com/hadolint/hadolint/releases/download/v1.18.0/hadolint-$(uname -s)-$(uname -m)"
+    && chmod 700 ${HADOLINT}
```
marienfressinaud: I think this line prevents Travis to parse the file

Frenzie: I didn't notice it was a separate line (I c/p-ed it from your link); that makes no sense at all. ;-)

marienfressinaud (approved):

PR #3096, opened 2020-07-04
Minz allow parallel sessions

### LINKED ISSUE #3093: Minz: allow parallel sessions
freshrss locks up when checking for a feed(that has an error or takes a long time)

so I noticed this when I used the "actualize" button for refreshing a feed that seems to had a timeout error.
when I do that in a new tab, by right clicking on actualize button and opening in new tab, then as long as that tab is opening and trying to fetch the feed, all the freshrss cant operate correctly.

it doesn't actually lock up as in freeze in the former opened tabs. but refreshing the already opened freshrss tabs or opening a new tab to a freshrss instance would just result in a tab that waits to the first (actualizing) tab to finish either successfully or fail with a timeout and then others tabs work normally.

it seems to me that this program just uses one thread? (I am not a programmer so forgive me if I use the wrong names) that lockes up when some feed updating is going on.

so the feed updating thread just locks up whole UI too (not freezing graphically but not like a time-outing server that waits for server response ,which I think what is actually happening).

btw why it is called actualize?
what it is actualizing?

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3093

`session_start()` is blocking until close.

See:
* https://php.net/session-write-close
* https://ma.ttias.be/php-session-locking-prevent-sessions-blocking-in-requests/

P.S.: To avoid problems with API calls, which do not have real Web session, I have introduced the concept of `volatile` session in Minz, which do not use any cookie or local storage on disk for session data (we used to create a lot of garbage)


Alkarex:
@rezad1393 If you want try manually on your current installation, you can manually add the following 3 lines, which are the most important part of the patch:
https://github.com/FreshRSS/FreshRSS/pull/3096/commits/eee547cf9d9ce3b182b6c3092f677255a7f7eb02

app/Models/DatabaseDAO.php
```diff
@@ -22,7 +22,7 @@ public function create() {
 			$sql = sprintf($SQL_CREATE_DB, empty($db['base']) ? '' : $db['base']);
 			return $this->pdo->exec($sql) !== false;
 		} catch (Exception $e) {
-			$_SESSION['bd_error'] = $e->getMessage();
+			Minz_Session::_param('bd_error', $e->getMessage());
```
Alkarex: This in particular requires testing

app/Models/DatabaseDAO.php
```diff
@@ -35,7 +35,7 @@ public function testConnection() {
 			$res = $stm->fetchAll(PDO::FETCH_COLUMN, 0);
 			return $res != false;
 		} catch (Exception $e) {
-			$_SESSION['bd_error'] = $e->getMessage();
+			Minz_Session::_param('bd_error', $e->getMessage());
```
Alkarex: This in particular requires testing

lib/Minz/Session.php
```diff
@@ -35,13 +36,26 @@ public static function param($p, $default = false) {
...
+		session_start();
 		if ($v === false) {
 			unset($_SESSION[$p]);
 		} else {
 			$_SESSION[$p] = $v;
 		}
+		session_write_close();
 	}
 
+	public static function _params($keyValues) {
```
Alkarex, line 84: Optimisation function, to avoid starting/closing the session for several parameters to change in a row

Alkarex:
P.S. Manual global refresh should also be faster now

Frenzie (approved):
@Alkarex Excellent, I've now been able to reproduce both the problem and the fix.

Alkarex:
@Frenzie I have been trying to think whether the fact that sessions can now be accessed from multiple requests in parallel could have any security implication, e.g. changing user in one session while another request is half way. I cannot immediately think of a problem, but we should at least double-check our use of e.g. https://php.net/session-destroy

rezad1393:
thanks for fast replies.
I love that this program's developers answer quickly to user feedback.
hope this lasts.

lib/Minz/Session.php
```diff
@@ -4,18 +4,31 @@
...
 	/**
 	 * Initialise la session, avec un nom
 	 * Le nom de session est utilisé comme nom pour les cookies et les URLs(i.e. PHPSESSID).
-	 * Il ne doit contenir que des caractères alphanumériques ; il doit être court et descriptif
+	 * Il ne doit contenir que des caractères alphanumériques ; il doit être court et descriptif.
+	 * If the volatile parameter is true, then no cookie and not session storage are used.
+	 * Volatile is especially useful for API calls without cookie / Web session.
 	 */
-	public static function init($name) {
+	public static function init($name, $volatile = false) {
```
Alkarex, line 35: New concept of volatile session, to avoid a warning with our API calls, and with a nice optimisation that avoids creating local garbage for unneeded session data on disk.

p/i/index.php
```diff
@@ -27,8 +27,6 @@
 	require(APP_PATH . '/install.php');
 } else {
 	session_cache_limiter('');
-	Minz_Session::init('FreshRSS');
```
Alkarex, line 30: Moved down to be called only once

p/i/index.php
```diff
@@ -38,13 +36,16 @@
...
 		if (httpConditional($dateLastModification, 0, 0, false, PHP_COMPRESSION, true)) {
+			Minz_Session::init('FreshRSS');
+			Minz_Session::_param('keepAlive', 1);	//To prevent the PHP session from expiring
 			exit();	//No need to send anything
 		}
 	}
 
 	try {
 		$front_controller = new FreshRSS();
 		$front_controller->init();
```
Alkarex: This line also calls `Minz_Session::init('FreshRSS');`

lib/Minz/Session.php
```diff
@@ -4,18 +4,31 @@
...
+
 		$cookie = session_get_cookie_params();
 		self::keepCookie($cookie['lifetime']);
 
 		// démarre la session
 		session_name($name);
 		session_start();
+		session_write_close();
+		//Use cookie only the first time the session is started
+		ini_set('session.use_cookies', '0');
```
Alkarex, line 51: Needed, otherwise PHP will attempt to re-send some HTTP headers

marienfressinaud: I don't understand when this is set to 1?

marienfressinaud (commented):
The main issue I can find with this solution is that "notification" messages could start to behave oddly and it will be hard to debug.

Also I think the "volatile" solution could be avoided with a better architecture (i.e. avoiding calling controllers directly from the CLI scripts) but it looks like an acceptable solution for the current situation.

lib/Minz/Session.php
Alkarex: 1 is the default PHP setting

p/i/index.php
Alkarex: That would also be fine. We are calling `new FreshRSS()` from `actualize_script.php` as well (which should be updated or replaced by our CLI approach some day)

lib/Minz/Session.php
```diff
@@ -4,18 +4,31 @@
...
-	 * Il ne doit contenir que des caractères alphanumériques ; il doit être court et descriptif
+	 * Il ne doit contenir que des caractères alphanumériques ; il doit être court et descriptif.
+	 * If the volatile parameter is true, then no cookie and not session storage are used.
+	 * Volatile is especially useful for API calls without cookie / Web session.
 	 */
-	public static function init($name) {
+	public static function init($name, $volatile = false) {
+		self::$volatile = $volatile;
+		if (self::$volatile) {
+			$_SESSION = [];
```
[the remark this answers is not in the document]
Alkarex: Apparently :-)

Alkarex:
Can you pinpoint what the problem with notifications could be, @marienfressinaud ?

lib/Minz/Session.php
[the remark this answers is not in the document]
Alkarex:
```php
<?php
$_SESSION = [];

function test() {
	$_SESSION['Hello'] = 'world';
}

test();
print_r($_SESSION);
```

lib/Minz/Session.php
marienfressinaud: Yes, but it looks like we always set `use_cookies` to `0` :thinking:

Alkarex: We do so **after** having called `session_start()` the first time

marienfressinaud:
If a user opens FRSS in two tabs and sends two requests in parallel, we could have a situation where:

- request A sets a notification
- request B sets a notification
- request A returns and shows the notification set by request B
- request B returns and shows nothing

It's maybe not a big deal, but by removing the lock mechanism, we introduce a risk of race conditions in the code.

lib/Minz/Session.php
marienfressinaud: ha! It means calling `session_start` adds HTTP headers each time we call it. It could be great to have this information directly as a comment ;)

Alkarex:
> If a user opens FRSS in two tabs and sends two requests in parallel, we could have a situation where:
> 
>     * request A sets a notification
> 
>     * request B sets a notification
> 
>     * request A returns and shows the notification set by request B
> 
>     * request B returns and shows nothing

Ah indeed. We could consider tracking better which request to respond with the notification (for instance with a URL parameter, either with the content of the notification itself, or with an ID)

marienfressinaud:
> Ah indeed. We could consider tracking better which request to respond with the notification (for instance with a URL parameter, either with the content of the notification itself, or with an ID)

This solution comes with more difficulties:

- URL are size limited
- the ID solution cannot pass variables
- how to get rid of the parameter so refreshing the page will not show the notification again?

I'm thinking about an alternative solution. We should determine which calls take a long time to finish and for now I can see only one: the feeds refreshing endpoint. So I suggest to close write access only on this call. A generic notification message could be set at the beginning. What do you think?

Alkarex:
Session write access must be closed on all requests, otherwise `session_start()` will block, but we can indeed think of workarounds.
A long request when refreshing a bad feed is likely to need a notification (e.g. timeout...).
Searches can also take several seconds depending on the system.
For the URL parameter approach (only an idea), our messages are not that long, are they? I do not think it is a problem to see the message again when refreshing a page, but if it is a problem, we could also find a solution for that. I can try an implementation if you like.

marienfressinaud:
> Session write access must be closed on all requests, otherwise session_start() will block

Yes, but it would likely be painless since the other requests end a lot quicker. Session write can be closed on index pages too (I don't think we change the session here anyway).

> A long request when refreshing a bad feed is likely to need a notification (e.g. timeout...).

In fact, my "generic notification" idea solves nothing. A better approach would be to call `session_write_close` at the beginning of the call and `session_start` before setting the notifications (but without closing again to avoid the race condition). It means another tab could load during the "refresh" call is ongoing.

> I can try an implementation if you like.

I'm worried about the energy it might take and the big changes it implies. I mean the notification system is not broken and proved to work well for few years now, I'm annoyed to change it now. The real issue is "page doesn't load when a request is ongoing in another tab" and can be detected/painful only with some specific requests (refreshing feeds, searches), let's focus on them instead of changing the whole notification system.

Alkarex:
Now waiting for https://github.com/FreshRSS/FreshRSS/pull/3208

Alkarex:
The case for concurrent requests and their respective notifications is fixed

lib/Minz/Session.php
```diff
@@ -4,18 +4,51 @@
...
 class Minz_Session {
+	private static $volatile = false;
+
+	/**
+	 * For mutual exclusion.
+	 */
+	private static $locked = false;
+
+	public static function lock() {
+		if (!self::$volatile && !self::$locked && session_start()) {
```
marienfressinaud, line 15: It's the morning, but I don't understand why you need to add the `self::$locked` variable: doesn't `session_start` already acts as a lock?

Alkarex: This is to prevent the case of calling several times `lock()` from the same process (already locked), but as we do not do it, it does not change anything at the moment

lib/Minz/Session.php
```diff
@@ -35,13 +68,34 @@ public static function param($p, $default = false) {
 	 * @param $v la valeur à attribuer, false pour supprimer
 	 */
 	public static function _param($p, $v = false) {
+		if (!self::$volatile && !self::$locked) {
+			session_start();
```
Alkarex, line 72: If we wanted to get fresh values from a potential concurrent session, we would need to do the same for `param()`.
I am not sure this is desirable though

marienfressinaud (approved):
Ok then :)

Alkarex:
Let's merge and continue testing in the main branch (I have heard we will soon have to change that branch name to avoid *master*...). Keeping an eye in particular on the install process, CLI, and concurrent requests (all good for me so far).

Alkarex:
@rezad1393 If you are around, please give a try to the latest master branch

rezad1393:
>I have heard we will soon have to change that branch name to avoid _master_...

you dont have to, but it is nice thing to do.
I always find it weird that the world master and slave was used in computer world. like master hdd and slave hdd from the PATA world.

rezad1393:
> @rezad1393 If you are around, please give a try to the latest master branch

can I do that in my personal (kind like production) instance? or is it too dangerous to do so?

Alkarex:
@rezad1393 if you take a backup before, that should be fine. How do you deploy FreshRSS? (git, Docker, zip...)
The current master has not changed anything in the DB structure compared to the last release, so it is easy to to back at the moment

rezad1393:
just zip I guess, I don't remember.
So I guess I will try it and see what happens.

Frenzie:
> I always find it weird that the world master and slave was used in computer world. like master hdd and slave hdd from the PATA world.

I definitely found that terminology odd when I found out about it as a child in the '90s, but I think Git is like a master copy? Which is to say, I don't think the duplicates are "slaves" like they are in HDDs or databases and it's just the word "master" but I could be wrong.

(Anyhoo, main is fine, probably clearer if you're new to Git.)

rezad1393:
oh 
I cant test because I dont have a slow feed to test with,
https://test.alapetite.fr/slow.rss.xml.php doesnt exist anymore

Alkarex:
@rezad1393 Now the test feed should be back

rezad1393:
I thinks it works
and overall feels faster too but don't know why

Alkarex:
Thanks for the test @rezad1393 👍

rezad1393:
it seems that I spoke too early.
sometimes the ui doesnt load completely (missing panes) 
I try to see if I can see what cause it.

rezad1393:
it is even more obvious on my tablet (firefox android) which shows this behavior more.
for example if I click on the fav lis, it shows an incomplete list. sometimes 1 sometimes 4 sometimes 30(or whatever the lazy loading limit is) and then dragging the page to bottom to show more doesn't work and just shows the 3 boxes blinking and not loading anymore items.

lib/Minz/Session.php
```diff
@@ -4,18 +4,51 @@
...
+
+	public static function lock() {
+		if (!self::$volatile && !self::$locked && session_start()) {
+			self::$locked = true;
+		}
+		return self::$locked;
+	}
+
+	public static function unlock() {
+		if (!self::$volatile && session_write_close()) {
```
Alkarex, line 22: Bug with PHP 7.1- fixed in https://github.com/FreshRSS/FreshRSS/pull/3273

p/i/index.php
Alkarex: This prevented getting the name of the current user, thus preventing the HTTP cache from working.
Fixed in https://github.com/FreshRSS/FreshRSS/pull/6719

PR #3099, opened 2020-07-06
Hide feed credentials when adding a new feed

Alkarex:
Avoid showing HTTP Basic Auth credentials in feed error log and in feed
title (when there is no title provided)

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -287,7 +287,7 @@ class FreshRSS_Feed extends Minz_Model {
 				if ((!$mtime) || $simplePie->error()) {
 					$errorMessage = $simplePie->error();
 					throw new FreshRSS_Feed_Exception(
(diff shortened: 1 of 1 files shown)


app/Models/Feed.php
```diff
@@ -287,7 +287,7 @@ public function load($loadDetails = false, $noCache = false) {
 				if ((!$mtime) || $simplePie->error()) {
 					$errorMessage = $simplePie->error();
 					throw new FreshRSS_Feed_Exception(
-						($errorMessage == '' ? 'Unknown error for feed' : $errorMessage) . ' [' . $url . ']'
+						($errorMessage == '' ? 'Unknown error for feed' : $errorMessage) . ' [' . $this->url . ']'
```
Alkarex, line 290: `$this->url` does not contain the `user:password` component, unlike `$url` which does

marienfressinaud (approved):

Issue #3103, opened 2020-07-12
Add Different Time Customizations 


Fogest:
Currently other platforms like tt-rss and other non-self hosted RSS readers will have times in a relative format. Currently I have to look at the time and take a rough guess in my head on how old an article is. Whereas if I look in something like tt-rss I can see that an article was imported "23 minutes ago".

After it is over 1-2 hours old it switches to just showing a time such as "7:03" and "10:21". If the feed was imported from a previous day other than today the date/time would switch to show "Jul 11, 23:49". 

Doing the time in a way similar to tt-rss or other readers allows you to have to do less manual comparing to what your current time is as the website can do it for you. 

I would think a setting like this may be suited for the "Display" configuration.

aledeg:
I think this is probably better as an extension.

aledeg:
@Fogest I've just added an extension to customize how dates are displayed. It's available [here](https://github.com/aledeg/FreshRSS-extensions/tree/master/xExtension-DateFormat). Let me know what you think.

Fogest:
@aledeg Awesome, it didn't work at first but I noticed a bunch of commits that reference this which I assume were needed for the extension. I updated to the `:latest` on docker and the extension works great!

I am going to close this, however as a secondary question, is there a reason when supporting by unread that there is not an order to the unread (see image below)? My `sort order` in the settings is using "Newest first"

![image](https://user-images.githubusercontent.com/1163628/87627510-e4bef700-c6fc-11ea-8905-2658c45bec76.png)

(also not sure why that one is happening in the future)

PR #3106, opened 2020-07-13
Add machine-readable dates in views

aledeg:

- Add machine-readable dates in views

1. Check the DOM

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

This will allow to make some transformations on dates when needed.
See #3103 and #853

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -87,9 +87,11 @@ class FreshRSS_Entry extends Minz_Model {
 	public function date($raw = false) {
 		if ($raw) {
 			return $this->date;
-		} else {
-			return timestamptodate($this->date);
 		}
+		return timestamptodate($this->date);
+	}
+	public function machineReadableDate() {
+		return @date (DATE_ATOM, $this->date);
 	}
 	public function dateAdded($raw = false, $microsecond = false) {
 		if ($raw) {
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -44,6 +44,6 @@
 			?></span><?php
 		endif;
 	?></a></li>
-	<?php if ($topline_date) { ?><li class="item date"><?= $this->entry->date() ?> </li><?php } ?>
+	<?php if ($topline_date) { ?><li class="item date"><time datetime="<?= $this->entry->machineReadableDate() ?>"><?= $this->entry->date() ?></time>&nbsp;</li><?php } ?>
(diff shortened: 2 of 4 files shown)


app/Models/Entry.php
```diff
@@ -87,9 +87,11 @@ public function link() {
 	public function date($raw = false) {
 		if ($raw) {
 			return $this->date;
-		} else {
-			return timestamptodate($this->date);
 		}
+		return timestamptodate($this->date);
+	}
+	public function machineReadableDate() {
+		return @date ('Y-m-d\TH:i:s', $this->date);
```
Alkarex: Wouldn't the standard `date('c', ...)` work, or even `gmdate('c', ...)`?

aledeg: It works and it's easier to understand. But I found an even clearer way :)

Alkarex:
Let me merge https://github.com/FreshRSS/FreshRSS/pull/3081 which has a couple of conflicting lines

Alkarex (approved):

app/views/index/reader.phtml
```diff
@@ -62,8 +62,7 @@ $content_width = FreshRSS_Context::$user_conf->content_width;
 						endforeach;
 						echo ' — ';
 					endif;
-					echo $item->date();
-				?></div><div class="text">
+				?><time datetime="<?= $item->machineReadableDate() ?>"><?= $item->date() ?></time></div><div class="text">
```
aledeg, line 65: @Alkarex Please, check if I didn't forget something while rebasing.

app/views/helpers/index/normal/entry_header.phtml
```diff
@@ -44,6 +44,6 @@
 			?></span><?php
 		endif;
 	?></a></li>
-	<?php if ($topline_date) { ?><li class="item date"><?= $this->entry->date() ?> </li><?php } ?>
```
Alkarex, line 47: Minor: We used to have a space after the date, to avoid being too close to the border, especially when the external link icon is not displayed:
image

Alkarex (approved):

app/views/index/reader.phtml
Alkarex: Looks good 👍

app/views/helpers/index/normal/entry_header.phtml
aledeg: I've fixed that but I feel that this should be handled by CSS instead of relying on spacing. My 2¢

Alkarex: Indeed, and I cannot remember why it was like that, but let's take that separately :-)

PR #3118, opened 2020-07-14
Change machine-readable date format

aledeg:

- Change machine-readable date format to UTC

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, the date was using the server timezone. This can be a problem
when using different timezones.
Now, the date is converted to UTC first. This way, we always have the
date in the same format no matter what the server configuration is.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -90,8 +90,11 @@ class FreshRSS_Entry extends Minz_Model {
 		}
 		return timestamptodate($this->date);
 	}
-	public function machineReadableDate() {
-		return @date (DATE_ATOM, $this->date);
+	public function machineReadableDate($utc = true) {
+		if (true === $utc) {
+			return @gmdate(DATE_ATOM, $this->date);
+		}
+		return @date(DATE_ATOM, $this->date);
 	}
 	public function dateAdded($raw = false, $microsecond = false) {
 		if ($raw) {


Frenzie (approved):
Sounds good to me.

Alkarex:
I proposed `gmdate()` in https://github.com/FreshRSS/FreshRSS/pull/3106#discussion_r453616111 but did not insist, because as in both cases the format includes a time-zone, it should not matter. What is the problem more precisely?

P.S.: The slight advantage of `date()` is that it *also* contains the information of the server time-zone, in cases when it might be relevant to know

P.P.S: What I mean by *not matter* is that parsing those two alternatives using any compliant function (including in JavaScript and PHP) should give the exact same result

aledeg:
I am trying to use the date with momentjs. It has an UTC method to process the date. That's the reason behind that change.

Alkarex:
Hum, it is supposed to work out of the box with formats containing a timezone information. Isn't it the case?
See e.g. https://momentjs.com/docs/#/parsing/parse-zone/

Alkarex:
See also the native https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse
Anyway, not a problem to merge this PR if needed, but I think the problem, if any, might be hidden elsewhere

aledeg:
I slept on that and I might be overreacting with the UTC format. I'll see if I can do what I need with momentjs with what we have right now. I might reopen that PR later on.

PR #3121, opened 2020-07-16
Add tag management page

aledeg:
Closes #3058

Changes proposed in this pull request:

- The new page allows to create, delete and rename tags.

How to test the feature manually:

1. access the tag management page
2. create a tag
3. rename a tag
1. delete a tag

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/master/docs/en/developers/04_Pull_requests.md).

PR #3125, opened 2020-07-23
[lang] Tags management → tag management

Frenzie:
Cf. <https://github.com/FreshRSS/FreshRSS/pull/3121#issuecomment-662922474>.

---
This change is [Reviewable](https://reviewable.io/reviews/freshrss/freshrss/3125)


Alkarex:
Either here or in another PR, I think the two menus (subscriptions and tags) should be moved next to each other
image
And what about fixing *subscriptions* as well?

Alkarex (approved):

Frenzie:
@Alkarex 
> And what about fixing subscriptions as well?

It's already been fixed in en-US and as such also in #3068, but finishing that up is annoyingly time consuming. But the end is in sight; not too much left. :+1:

PR #3164, opened 2020-09-01
Add variable undeclared ($url)

ColonelMoutarde:
Closes #

Add variable undeclared ($url)

### DIFF
--- a/app/Exceptions/ZipException.php
+++ b/app/Exceptions/ZipException.php
@@ -3,8 +3,8 @@
 class FreshRSS_Zip_Exception extends Exception {
 	private $zipErrorCode = 0;
 
-	public function __construct($zipErrorCode, $url = null) {
-		parent::__construct('ZIP error! ' . $url, 2141);
+	public function __construct($zipErrorCode) {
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Thanks :+1:

aledeg (changes requested):
I think it is better to change the error message instead of adding a useless parameter to the method

ColonelMoutarde:
As you want :)

app/Exceptions/ZipException.php
```diff
@@ -3,7 +3,7 @@
 class FreshRSS_Zip_Exception extends Exception {
 	private $zipErrorCode = 0;
 
-	public function __construct($zipErrorCode) {
+	public function __construct($zipErrorCode, $url = null) {
 		parent::__construct('ZIP error! ' . $url, 2141);
```
Alkarex:
```suggestion
	public function __construct($zipErrorCode) {
		parent::__construct('ZIP error!', 2141);
```

I believe this is what @aledeg suggests. You can just click on accept here @ColonelMoutarde

aledeg (approved):

PR #3165, opened 2020-09-01
Fix variable undeclared ($sql)

ColonelMoutarde:
Closes #

|  Line|   app/Models/DatabaseDAOSQLite.php  |
 ------| ---------------------------------- 
 | 72   |  Undefined variable: $sql |


aledeg (changes requested):
There is no need to have this variable in the method definition. But as we have an undefined variable in the logging, we should change the logging message.

ColonelMoutarde:
As you want, you are the boss ;)

aledeg (approved):

PR #3168, opened 2020-09-02
Fix tag in user query

### LINKED ISSUE #3163: Undefine variable
Hello,
I found a problem

 | Line |  Models/UserQuery.php|      
 | ------ | -------------------------- 
  |157  |  Undefined variable: $tag  |

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3163

### DIFF
--- a/app/Models/UserQuery.php
+++ b/app/Models/UserQuery.php
@@ -153,7 +153,7 @@ class FreshRSS_UserQuery {
 		if ($this->tag_dao == null) {
 			throw new FreshRSS_DAO_Exception('Tag DAO is not loaded in UserQuery');
 		}
-		$category = $this->category_dao->searchById($id);
+		$tag = $this->tag_dao->searchById($id);
 		if ($tag) {
 			$this->get_name = $tag->name();
 		} else {
--- a/p/themes/Ansum/_layout.scss
+++ b/p/themes/Ansum/_layout.scss
@@ -38,6 +38,7 @@
 		&.search {
 			// text-align: center;
 			// width: 50%;
+
 			input {
 				width: 230px;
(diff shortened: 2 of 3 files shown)


Alkarex:
@aledeg I cannot remember how to test a `get` parameter in a user query, can you?

Alkarex:
Merging fast to fix build. This part for sure was wrong, and let's make another PR if there is anything left to be fixed

aledeg:
@Alkarex what do you mean by testing a get parameter?

Alkarex:
@aledeg this `parseTag` function is only used by `parseGet`
https://github.com/FreshRSS/FreshRSS/blob/6fac0e40c4d3bc0e51a124015dcde9b7d766f7b5/app/Models/UserQuery.php#L77-L98
which is in turn only used when there is a `get` parameter:
https://github.com/FreshRSS/FreshRSS/blob/6fac0e40c4d3bc0e51a124015dcde9b7d766f7b5/app/Models/UserQuery.php#L33-L35

aledeg:
Thank you for the explanations, but I still can not understand what you are trying to achieve. I am a bit slow today!

Alkarex:
@aledeg :-) It is just that I do not know the use-case, which requires this `?get=XXX` in UserQuery, and I would like to either test it (to check that is still works as intended) or remove it (if it is obsolete)
https://github.com/FreshRSS/FreshRSS/blob/6fac0e40c4d3bc0e51a124015dcde9b7d766f7b5/app/Models/UserQuery.php#L33-L34

aledeg:
ok, I understand. I'll look into it and I'll let you know.

PR #3170, opened 2020-09-02
bump default logout from 30 days to 3 months

### LINKED ISSUE #3169: Suggestion: Make the login timeout larger than 30 days
Currently, login expires after 30 days.
image

Having to login to my own service every 30 days is inconvenient, would it make sense to increase the timeout to be indefinite, or just a very long time? (1 year or longer)

C0rn3j:
Closes #3169

- Change login cookie expiry from 30 to 365 days

1. Try to login, form should say 365 days instead of 30

Pull request checklist:

- [X] clear commit messages
- [ ] code manually tested
- [X] documentation updated


Frenzie (approved):
Fine by me, but I don't know if everyone would agree. ^_^

aledeg:
The code itself is fine. But I think that instead of having to choose the number of seconds, we should have some drop-down to select a more convenient value. One with a number and one with a period (year, month, etc).
But that is for another PR :)

SuperSandro2000:
Why is the same value hardcoded at multiple places? This should be a const.

> we should have some drop-down to select a more convenient value.

I think this is just feature creep.

I think a year is quite high and can leave a lot of sessions open at a time. 3 or maybe 6 months sounds like a better idea.

Alkarex:
Here is an alternative with automatic cookie renewal instead of a longer expire value 🎄 https://github.com/FreshRSS/FreshRSS/pull/3287
Tests welcome!

Frenzie (commented):
@Alkarex Something like this?

Alkarex:
@Frenzie Maybe `DEFAULT_COOKIE_DURATION`, as it can be overridden in `config.php` ? 
But I find it a bit confusing to have it as constant and as a config. So I would suggest defining the default value in `app/Models/Auth.php` instead of in `constants.php`

config.default.php
```diff
@@ -89,7 +89,7 @@
 	'limits' => array(
 
 		# Duration in seconds of the login cookie.
-		'cookie_duration' => 2592000,
+		'cookie_duration' => COOKIE_DURATION,
```
Frenzie: @Alkarex What about this one?

Alkarex: `FreshRSS_FormAuth::DEFAULT_COOKIE_DURATION` should work.
Was that your question?

Frenzie: No, more that it comes across weirder/more confusing to me.

Alkarex: Then maybe leaving it as an integer there would be simpler

Frenzie: I wasn't objecting (much), but also it doesn't seem to work.
```
[Sat Dec 26 21:39:27.542015 2020] [php7:error] [pid 19] [client 127.0.0.1:49252] PHP Fatal error:  Uncaught Error: Class 'FreshRSS_FormAuth' not found in /var/www/FreshRSS/config.default.php:92
```

Alkarex: Ah sorry, should have been in `FreshRSS_Auth::*`

Alkarex: Another option being to extract the `FreshRSS_FormAuth` class in its own PHP file

Frenzie: Drat, I did that first before your other comment but I never committed it.

Alkarex (approved):

PR #3171, opened 2020-09-03
When ssl_verify is false, decrease SSL security

Alkarex:
When `ssl_verify` option is set to false for a feed, allow lower security such as SHA-1 signatures.
Needed for older sites when FreshRSS is running on current Debian or Ubuntu.

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900984
* https://wiki.debian.org/ContinuousIntegration/TriagingTips/openssl-1.1.1
* https://stackoverflow.com/questions/58342699/php-curl-curl-error-35-error1414d172ssl-routinestls12-check-peer-sigalgwr

Fix error of type `cURL error 35: error:1414D172:SSL routines:tls12_check_peer_sigalg:wrong signature`

Example of feeds:
* https://www.version2.dk/it-nyheder/rss
* https://ing.dk/rss/nyheder


Frenzie (approved):
I've seen a couple of feeds with issues but I think they're different, with an outdated certificate somewhere in the chain. Regardless. 👍

PR #3180, opened 2020-09-16
SimplePie: Fix HTTP 301 permanent redirection

Alkarex:
When adding feeds it worked fine, but detecting permanent redirects for existing feeds was sometimes broken (only when PHP open_basedir was not set).

Indeed, using the built-in CURLOPT_FOLLOWLOCATION instead of the manual method in SimplePie hides the list of HTTP redirects along the way, and prevents the distinction of e.g. 301 vs. 302 redirects.

This patch disables CURLOPT_FOLLOWLOCATION in SimplePie, and fixes the manual method at the same time.
The use of CURLOPT_FOLLOWLOCATION was nevertheless not systematic (only when open_basedir was not set), so now there is only one logic independent of open_basedir.

I will send a PR upstream to SimplePie.

How to test: pick a feed with 301 redirection such as HTTP to HTTPS, e.g. http://ing.dk/rss/term/341
Manually change back to previous address (to simulate a feed that is changing address)
Refresh feed and observe whether the 301 redirect is obeyed.

### DIFF
--- a/lib/SimplePie/SimplePie/File.php
+++ b/lib/SimplePie/SimplePie/File.php
@@ -113,11 +113,11 @@ class SimplePie_File
 				curl_setopt($fp, CURLOPT_REFERER, SimplePie_Misc::url_remove_credentials($url));
 				curl_setopt($fp, CURLOPT_USERAGENT, $useragent);
 				curl_setopt($fp, CURLOPT_HTTPHEADER, $headers2);
-				if (!ini_get('open_basedir') && !ini_get('safe_mode') && version_compare(SimplePie_Misc::get_curl_version(), '7.15.2', '>='))
-				{
-					curl_setopt($fp, CURLOPT_FOLLOWLOCATION, 1);
-					curl_setopt($fp, CURLOPT_MAXREDIRS, $redirects);
-				}
+				//if (!ini_get('open_basedir') && !ini_get('safe_mode') && version_compare(SimplePie_Misc::get_curl_version(), '7.15.2', '>='))
+				//{
+				//	curl_setopt($fp, CURLOPT_FOLLOWLOCATION, 1);
+				//	curl_setopt($fp, CURLOPT_MAXREDIRS, $redirects);
+				//}
 				foreach ($curl_options as $curl_param => $curl_value) {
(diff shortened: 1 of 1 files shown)


lib/SimplePie/SimplePie/File.php
```diff
@@ -113,11 +113,11 @@ public function __construct($url, $timeout = 10, $redirects = 5, $headers = null
 				curl_setopt($fp, CURLOPT_REFERER, SimplePie_Misc::url_remove_credentials($url));
 				curl_setopt($fp, CURLOPT_USERAGENT, $useragent);
 				curl_setopt($fp, CURLOPT_HTTPHEADER, $headers2);
-				if (!ini_get('open_basedir') && !ini_get('safe_mode') && version_compare(SimplePie_Misc::get_curl_version(), '7.15.2', '>='))
```
Alkarex, line 116: This logic was used sometimes, in which case it was problematic. Removed.

Alkarex:
Let's merge and continue testing in /master

lib/SimplePie/SimplePie/File.php
```diff
@@ -113,11 +113,11 @@ public function __construct($url, $timeout = 10, $redirects = 5, $headers = null
...
-				{
-					curl_setopt($fp, CURLOPT_FOLLOWLOCATION, 1);
-					curl_setopt($fp, CURLOPT_MAXREDIRS, $redirects);
-				}
+				//if (!ini_get('open_basedir') && !ini_get('safe_mode') && version_compare(SimplePie_Misc::get_curl_version(), '7.15.2', '>='))
+				//{
+				//	curl_setopt($fp, CURLOPT_FOLLOWLOCATION, 1);
+				//	curl_setopt($fp, CURLOPT_MAXREDIRS, $redirects);
+				//}
```
SuperSandro2000, line 120: In git leaving commented code is not a good practise. This should probably be removed.

Alkarex: See https://github.com/simplepie/simplepie/pull/660 (which I will merge back when accepted)

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3206

PR #3185, opened 2020-09-22
Remember open categories

### LINKED ISSUE #2248: [Feature request] Remember opened categories in the left menu
Hi !

First of all, thanks to all the contributors of this project ! I was using Feedly before jump to FreshRSS for few weeks now. I've had issues with Fever API warnings who already have been solved (I use the Reeder iOS app).
I'm pretty happy with FreshRSS now. 

However, I find the behavior of left menu irritating. It will be great if the opened categories could be remembered between pages.
Unfortunately, I'm a sysadmin and not a developer (even less frontend Javascript developer).

If someone find some time ;)

prashanttholia:
Closes #2248 

- Browser session storage based implementation to remember open categories between visits to various pages on the site.

1. Simply test the feature on any browser.

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


prashanttholia:
@Alkarex, I considered using classList, but it is only partially supported in IE 10 and not supported at all in older versions. If it is not a concern, please let me know. I will go ahead and update the code.

Alkarex:
@prashanttholia We already use `classList`, so it is not a problem. We only support back to IE11 (partially) at the moment.

p/scripts/main.js
```diff
@@ -1606,3 +1612,111 @@ if (document.readyState && document.readyState !== 'loading') {
...
+			img.src = img.src.replace('/icons/down.', '/icons/up.');
+			img.alt = '△';
+		});
+	}
+}
+
+//Sidebar scroll handler. Stores sidebar scroll position in Session Storage.
+document.getElementById('sidebar').onscroll = function(event) {	
+	sessionStorage.setItem('FreshRSS_sidebar_scroll_position', event.target.scrollTop);
+};
```
Alkarex: Please keep the new line at the end of the file

Alkarex:
Thanks! I will have a look in more details shortly

p/scripts/main.js
```diff
@@ -1606,3 +1612,111 @@ if (document.readyState && document.readyState !== 'loading') {
...
+			add_class(category_element.querySelector('.tree-folder-items'), 'active');
+			const img = category_element.querySelector('a.dropdown-toggle img');
+			img.src = img.src.replace('/icons/down.', '/icons/up.');
+			img.alt = '△';
+		});
+	}
+}
+
+//Sidebar scroll handler. Stores sidebar scroll position in Session Storage.
+document.getElementById('sidebar').onscroll = function(event) {
```
Alkarex: Personal preference: I like having a space before an anonymous function:
```suggestion
document.getElementById('sidebar').onscroll = function (event) {	
```

p/scripts/main.js
```diff
@@ -1606,3 +1612,111 @@ if (document.readyState && document.readyState !== 'loading') {
 	document.addEventListener('DOMContentLoaded', init_afterDOM, false);
 }
 // @license-end
+
+
+
+
+//Adds CSS class to an HTML element.
```
Alkarex: Please move your code higher up, before the init stuff, and before the license-end tag

prashanttholia: Done.

p/scripts/main.js
prashanttholia: Done.

p/scripts/main.js
prashanttholia: Done.

prashanttholia:
Hi @Alkarex, 

I have made some updates in the main.js file. These updates consist of the following - 
- Feedback points mentioned by you in this thread above. (All 3)
- Moving sidebar scroll event listener inside init_column_categories function so that it gets called only on the normal screen.
- Delaying (using setTimeout) setting the sidebar scroll position in init_column_categories function (gets called on page load). Added delay because, on page load, the browser was also attempting to set the sidebar scroll position to top. This was interfering with the functionality. Now, with delay, the functionality attempts to set the scroll position after the browser is done. I have not had any foolproof considerations to pick the delay duration. Would love to have your feedback on it. 

Thanks.

prashanttholia:
@alkarex, did you get a chance to see the functionality? Looking forward to your feedback.

p/scripts/main.js
```diff
@@ -692,21 +692,113 @@ function init_posts() {
 	}
 }
 
+//Gets URL search parameter value.
+//Does not use URLSearchParams to ensure coverage of older versions of Internet Explorer.
+function get_search_param_value(param_name) {
```
Alkarex: Let's drop IE11 for non-essential features. I suggest you use URLSearchParams instead of this function (testing that URLSearchParams exists otherwise do nothing).

p/scripts/main.js
```diff
@@ -692,21 +692,113 @@ function init_posts() {
...
+}
+
+//Removes category id from session storage when user closes corresponding category in the left sidebar.
+function remove_category_from_session_storage(category_id) {
+	if ((category_id.charAt(0) !== 'c') && (category_id !== 'tags')) {
+		return;
+	}
+
+	let open_categories = sessionStorage.getItem('FreshRSS_open_categories');
+	if (open_categories.indexOf(category_id) !== -1) {
```
Alkarex: Wouldn't a JSON be easier to manipulate than a comma-separated string? Please give it a try.

p/scripts/main.js
```diff
@@ -692,21 +692,113 @@ function init_posts() {
...
+	if ((category_id.charAt(0) !== 'c') && (category_id !== 'tags')) {
+		return;
+	}
+
+	let open_categories = sessionStorage.getItem('FreshRSS_open_categories');
+	if (!open_categories) {
+		open_categories = '';
+	}
+
+	if (open_categories.indexOf(category_id) === -1) {
```
Alkarex: See below comment about using JSON instead of comma-separated string.
Furthermore, there is potentially a bug with multi-digit category IDs, e.g. `1,11,111`

p/scripts/main.js
```diff
@@ -692,21 +692,113 @@ function init_posts() {
...
 	}
 
+	display_open_categories();
+
+	// Set sidebar scroll position. Wait to let browser set the scroll position first.
+	setTimeout(function() { document.getElementById('sidebar').scrollTop = parseInt(sessionStorage.getItem('FreshRSS_sidebar_scroll_position')); }, 100);
+
+	// Sidebar scroll handler. Stores sidebar scroll position in Session Storage.
+	document.getElementById('sidebar').onscroll = function (event) {
+		sessionStorage.setItem('FreshRSS_sidebar_scroll_position', event.target.scrollTop);
```
Alkarex: This looks very expensive to do.
In any case, I suggest removing the scroll position from this PR, and focus in this PR on saving the open/close state of the categories. The scroll position can be discussed in another issue / PR.

Alkarex:
We have a parameter to control whether to unfold categories or not:
image
We need to add a new parameter to allow what your PR does, namely remembering which categories are open.

p/scripts/main.js
prashanttholia: Okay,

p/scripts/main.js
prashanttholia: Sure, I'll make the update.

p/scripts/main.js
prashanttholia: I'll check for the bug.

p/scripts/main.js
prashanttholia: Okay.

prashanttholia:
@Alkarex, yes, I'll add the new parameter to the "Categories to unfold" dropdown.
Thanks for all the feedback.

p/scripts/main.js
Alkarex: P.S. If you can actually move to JSON as suggested, no need to check for this bug, which is specific to the comma-separated approach

prashanttholia: Okay.

p/scripts/main.js
prashanttholia: Need one clarification. Shall I remove the entire feature if URLSearchParams is not available or shall I just put this part of code inside an availability condition? I'm not sure if the latter is desirable. Latter would also result in inconsistent behavior across two sets of browsers. Please advise.

Alkarex: We will probably drop IE11 in the not too distant future, so I would rather like not to introduce additional non-essential code for IE11, especially for non-default options. I suggest a test of whether `URLSearchParams` exists in the `display_open_categories()` function, and simply exit if it does not exist.

prashanttholia: Okay.

prashanttholia:
@Alkarex, I have implemented all feedback points. I request you to review.

At present, for the new 'Categories to unfold' dropdown item, I have set the English translation text as the translation text in all languages. Please let me know whether shall I add appropriate translations using Google Translate or some other translation service.

prashanttholia:
@Alkarex, in an attempt to resolve check failures, I messed up the commit chain in this PR. I apologize. The changes reflected "Files changed" tab are correct. Hope you would not find it problematic to check and merge the updates. I'm working towards acquiring relevant git & github knowledge. Thanks.

prashanttholia:
@Alkarex, in an attempt to resolve check failures, I messed up the commit chain in this PR. I apologize. The changes reflected under the "Files changed" tab are correct. Hope you would not find it problematic to check and merge the changes. I'm working towards acquiring relevant git and github knowledge. Thanks.

Alkarex:
@prashanttholia I have made a first round of changes. Please double-check https://github.com/FreshRSS/FreshRSS/pull/3185/commits/2bfa8a9c8392222ce1cf56669782029431791249

I think we should use localStorage instead of sessionStorage, to keep the values between sessions, which I believe is more useful. What remains to be done is to clear the localStorage when logging out and before logging in.

I have not looked in details yet, but the init code can probably be simplified: I suggest using the default "open" mechanism, and simply open the additional categories, which were previously open.

Last thing, I am not sure looking into URLSearchParams is needed at all. What needs to be saved, I believe, are the categories that have been manually opened.

prashanttholia:
> @prashanttholia I have made a first round of changes. Please double-check [2bfa8a9](https://github.com/FreshRSS/FreshRSS/commit/2bfa8a9c8392222ce1cf56669782029431791249)

@Alkarex, thanks for the review.
 
> I think we should use localStorage instead of sessionStorage, to keep the values between sessions, which I believe is more useful. What remains to be done is to clear the localStorage when logging out and before logging in.

Sure. I'll update the logout and login functionality.
 
> I have not looked in details yet, but the init code can probably be simplified: I suggest using the default "open" mechanism, and simply open the additional categories, which were previously open.
> 
> Last thing, I am not sure looking into URLSearchParams is needed at all. What needs to be saved, I believe, are the categories that have been manually opened.

Yes, I think the suggestions above would simplify the implementation very much. I'll incorporate them as well.

Thanks.

prashanttholia:
@Alkarex, all review points have been implemented. I request you to check them. In particular, I request you to review the changes made in "Delete open categories on login and logout" commit. Thanks.

Alkarex:
I have extracted the part related to the sidebar position into https://github.com/FreshRSS/FreshRSS/pull/3231

Alkarex:
Sorry for the delay. This is not forgotten

Alkarex:
@prashanttholia Sorry for the delay. I have made a few changes and simplifications https://github.com/FreshRSS/FreshRSS/pull/3185/commits/67de805c1fbfc919326790ec7502ff258fdff490
Could you please test?

Alkarex (approved):

prashanttholia:
@Alkarex, thanks for the review! I have tested the functionality and everything is working perfectly.

Alkarex:
Thanks again 👍 Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

prashanttholia:
Thanks, @Alkarex!

PR #3186, opened 2020-09-22
lib_phpQuery support PHP 8

### LINKED ISSUE #3082: Support PHP 8
https://github.com/php/php-src/blob/php-8.0.0alpha1/UPGRADING

In particular:

* [x] Removed `create_function()`. Anonymous functions may be used instead. (Used by `lib_phpQuery.php`)

prashanttholia:
Project updates to support PHP 8.

Closes #3082

- 

1.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


Alkarex:
Thanks @prashanttholia , that looks good. I will have a look shortly.
Did you already test whether your changes are working?

lib/lib_phpQuery.php
```diff
@@ -2115,22 +2132,24 @@ protected function pseudoClasses($class) {
...
 						));
 						return $isHeader
 							? $node
-							: null;')
+							: null;
+					}
 				)->elements;
 //				$this->elements = $this->map(
-//					create_function('$node', '$node = pq($node);
+//					function($node) { $node = pq($node);
```
aledeg: This commented section should be removed completely.

lib/lib_phpQuery.php
```diff
@@ -4701,11 +4724,12 @@ public static function phpToMarkup($php, $charset = 'utf-8') {
 			while (preg_match($regex, $php, $matches)) {
 				$php = preg_replace_callback(
 					$regex,
-//					create_function('$m, $charset = "'.$charset.'"',
-//						'return $m[1].$m[2]
+//					function($m, $charsetParam) use ($charset) {
```
aledeg: Same here

lib/lib_phpQuery.php
```diff
@@ -4737,9 +4761,9 @@ public static function markupToPHP($content) {
 		/* <php>...</php> to <?php...? > */
 		$content = preg_replace_callback(
 			'@<php>\s*<!--(.*?)-->\s*</php>@s',
-//			create_function('$m',
-//				'return "<'.'?php ".htmlspecialchars_decode($m[1])." ?'.'>";'
-//			),
+//			function($m) {
```
aledeg: same here

lib/lib_phpQuery.php
prashanttholia: Done.

lib/lib_phpQuery.php
prashanttholia: Done.

lib/lib_phpQuery.php
prashanttholia: Done.

prashanttholia:
Hi @Alkarex, I had tested separately only the update I made in CallbackBody class, but did not test the other updates. I'll test them too and post about it here.

lib/lib_phpQuery.php
```diff
@@ -1033,7 +1033,22 @@ public function __construct($paramList, $code, $param1 = null, $param2 = null,
 			$param3 = null) {
 		$params = func_get_args();
 		$params = array_slice($params, 2);
-		$this->callback = create_function($paramList, $code);
+
+		$this->callback = function(&...$args) use ($paramList, $code) {
+			$paramListArray = explode(',', $paramList);
```
Alkarex, line 1038: Is this function copied from somewhere else? If yes, please add a reference as a comment. If not, fine :-)

Alkarex:
Let's continue testing in the main branch.
@prashanttholia Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

lib/lib_phpQuery.php
prashanttholia: @Alkarex, this function is not copied, totally self-written.

Alkarex: Ok 👍

prashanttholia:
@Alkarex, I have been working on points to integrate PHP 8 changes. I understand that I have not been as prompt, but I would request you to allow me to do it. I request you to let me know whether shall I continue to use this PR to commit my updates?

Alkarex:
@prashanttholia You are very welcome. Please open a new PR. It was fine with that PR focussing on a single library

prashanttholia:
@Alkarex, okay, thanks a lot. One more point I wanted to mention was that I had tested the changes made in the library. The changes were made at 3 places in the file - 
1. CallbackBodyClass
2. phpQueryObject::pseudoClasses method
3. phpQuery::markupToPHP method

I wanted to mention few points about these.
For 1, I could not find the class being used anywhere in the project. So I have tested the new callback function in the class using sample functions in a separate script.
For 2, I have tested all cases where the changes were made using the FreshRSS management interface. All changes produced same results as the original script for variety of selector values for each case. I would, though, would want to bring into your notice that I found that the selector values "input" & ":enabled" did not produce valid results in both the original and the updated scripts.
For 3, I could not find a source in the project from where this was directly/indirectly being called and I have not tested it yet. I believe the implementation is correct though. I shall it test it any way and post my update here.

Thanks.

Alkarex:
Thanks for the detailed explanation @prashanttholia , and that looks sufficient for now 👍

prashanttholia:
Okay, thanks.

PR #3190, opened 2020-09-24
Add polish localization

wolfpld:
This PR adds polish localisation to FreshRSS. Most of the strings have been translated and verified, with some omissions left untranslated:

- Entries that seem to be no longer used (_seem_ is the key word here).
- User account creation process. I use FreshRSS in a single user configuration.
- Strings used during the installation process.
- Majority of transient notifications.

There are no obvious layout issues, with one exception on the login screen:

obraz


Frenzie (commented):
Could you also add the unchanged strings to the exceptions list? See https://travis-ci.org/github/FreshRSS/FreshRSS/jobs/730096527 ^_^

aledeg:
First, thank you for your contribution.
Second, why did you leave untranslated string?

wolfpld:
> Could you also add the unchanged strings to the exceptions list?

Will do.

> Second, why did you leave untranslated string?

Mainly because I'm interested in making FreshRSS look nice to me, and I don't need to be completionist. This is more a beauty of the language thing, rather than accessibility enablement.

A good example would be the installation process - it is done once, then never used again. I don't see a point translating it. It wouldn't change anything in how the reader is presenting itself during the day-to-day usage.

As for the notifications, I would like to have them translated, but frankly, I can barely notice them. This is how they look in my setup:

obraz

I don't want to translate anything without knowing the context in which the string is used, as that would be a recipe for bad translation. While reversing translation of a sentence back to the original language, to infer its true meaning can be hilarious, it's always annoying. A great example for this is how Bard's Tale 4 was translated to Polish.

obraz

Here the action description says "almost [happened]".

obraz

When the translation was updated in a patch, the hint was changed to "nearby".

The proper translation is of course "zamknij", "close". I don't know if it was ever corrected.

wolfpld:
Unchanged strings are now properly marked.

Frenzie (approved):

wolfpld:
Added three strings I forgot about.

PR #3208, opened 2020-10-03
Minz: Attach a notification to a request

Alkarex:
Notifications should be attached to a request, not to a global session.
Implement https://github.com/FreshRSS/FreshRSS/pull/3096#issuecomment-654891906
Prepare https://github.com/FreshRSS/FreshRSS/pull/3096

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -356,8 +356,11 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 				FreshRSS_Auth::giveAccess();
 			}
 
-			Minz_Request::setNotification($ok ? 'good' : 'bad',
-				_t('feedback.user.created' . (!$ok ? '.error' : ''), $new_user_name));
+			if ($ok) {
+				Minz_Request::setGoodNotification(_t('feedback.user.created', $new_user_name));
+			} else {
+				Minz_Request::setBadNotification(_t('feedback.user.created.error', $new_user_name));
+			}
 		}
 
 		$redirect_url = urldecode(Minz_Request::param('r', false, true));
@@ -543,8 +546,11 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 			}
 			invalidateHttpCache();
 
-			Minz_Request::setNotification($ok ? 'good' : 'bad',
-				_t('feedback.user.deleted' . (!$ok ? '.error' : ''), $username));
+			if ($ok) {
+				Minz_Request::setGoodNotification(_t('feedback.user.deleted', $username));
+			} else {
+				Minz_Request::setBadNotification(_t('feedback.user.deleted.error', $username));
+			}
 		}
 
 		Minz_Request::forward($redirect_url, true);
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -356,11 +356,8 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 				FreshRSS_Auth::giveAccess();
 			}
 
-			$notif = array(
-				'type' => $ok ? 'good' : 'bad',
-				'content' => _t('feedback.user.created' . (!$ok ? '.error' : ''), $new_user_name)
-			);
-			Minz_Session::_param('notification', $notif);
+			Minz_Request::setNotification($ok ? 'good' : 'bad',
+				_t('feedback.user.created' . (!$ok ? '.error' : ''), $new_user_name));
 		}
 
 		$redirect_url = urldecode(Minz_Request::param('r', false, true));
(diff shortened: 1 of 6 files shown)


lib/Minz/Request.php
```diff
@@ -274,6 +311,20 @@ public static function forward($url = array(), $redirect = false) {
...
 		$url = Minz_Url::checkUrl($url);
+		$requestId = self::requestId();
+		$url['params']['rid'] = $requestId;
+
+		//Forward notification
+		$notif = self::getNotification();
+		if ($notif) {
+			//TODO: Will need to ensure non-concurrency when landing https://github.com/FreshRSS/FreshRSS/pull/3096
+			$forwards = Minz_Session::param('forwards', []);
+			$forwards[$requestId] = [
```
Alkarex: Designed to be able to forward other things than notifications as well.

lib/Minz/Request.php
```diff
@@ -274,6 +311,20 @@ public static function forward($url = array(), $redirect = false) {
 		}
 
 		$url = Minz_Url::checkUrl($url);
+		$requestId = self::requestId();
+		$url['params']['rid'] = $requestId;
+
+		//Forward notification
+		$notif = self::getNotification();
+		if ($notif) {
```
Alkarex: Could alternatively be moved to `setNotification()`

aledeg:
I like what you've done here.

I was wondering if it could be a good idea to have more than one setter for that matter. Let me explain!
Instead of using ```Minz_Request::setNotification('good', _t('feedback.conf.updated'));``` use ```Minz_Request::setSuccessNotification(_t('feedback.conf.updated'));```. Instead of using ```Minz_Request::setNotification('bad', $msg);``` use ```Minz_Request::setFailureNotification($msg);```

This way, there is no need to rely on a constant value everywhere in the codebase, which is prone to errors. I've used the words success and failure in the method names but it could message levels like in monolog.
Using such thing will make the code more expressive and easier to check automatically.

Alkarex:
Good idea, here you are @aledeg https://github.com/FreshRSS/FreshRSS/pull/3208/commits/24b846b58a82765e2166ca8d4a846be7803f1c86

Alkarex:
@marienfressinaud Just for the record, if this Minz codebase is used in any other context, the former way of setting notifications is still possible `Minz_View::_param('notification', [ ... ])`

app/Controllers/userController.php
```diff
@@ -356,11 +356,11 @@ public function createAction() {
...
-			$notif = array(
-				'type' => $ok ? 'good' : 'bad',
-				'content' => _t('feedback.user.created' . (!$ok ? '.error' : ''), $new_user_name)
-			);
-			Minz_Session::_param('notification', $notif);
+			if ($ok) {
+				Minz_Request::setGoodNotification(_t('feedback.user.created', $new_user_name));
+			} else {
+				Minz_Request::setBadNotification(_t('feedback.user.created.error', $new_user_name));
+			}
```
marienfressinaud, line 363: A lot better :+1:

lib/Minz/Request.php
```diff
@@ -261,6 +261,50 @@ public static function serverIsPublic($address) {
...
+
+	public static function setBadNotification($content) {
+		self::setNotification('bad', $content);
+	}
+
+	public static function getNotification() {
+		$notif = null;
+
+		//Restore forwarded notifications
+		//TODO: Will need to ensure non-concurrency when landing https://github.com/FreshRSS/FreshRSS/pull/3096
```
marienfressinaud, line 291: This comment would be more useful directly in the https://github.com/FreshRSS/FreshRSS/pull/3096 thread

lib/Minz/Request.php
marienfressinaud: Well, yes :+1: I don't understand why we would save the notification both in session `notification` AND `requests` parameters. It looks to me we can replace the code of `setNotification()`  by the following lines.

marienfressinaud (approved):

marienfressinaud:
> Just for the record, if this Minz codebase is used in any other context […]

I hope it isn't haha :smile:

lib/Minz/Request.php
Alkarex: I will work on https://github.com/FreshRSS/FreshRSS/pull/3096 right after this PR lands, so it does not matter but was mainly intended for the current PR review

lib/Minz/Request.php
Alkarex: OK, here you are https://github.com/FreshRSS/FreshRSS/pull/3208/commits/76dc983942980044b53807b227511edf8d2c0d5d

PR #3211, opened 2020-10-06
tell git .png and .waff files are not text

ravenx99:
When I git-pull 1.17.0 and later, I immediately have "changed files" in the form of .png and .waff files that git wants to CRLF -> LF convert.  I can't stash these changes, can't check out another branch, etc until I have done something with them.
This addition to .gitattributes tells git these files are not text, so it doesn't try to end-of-line convert them.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 * text=auto eol=lf
+*.png -text
+*.waff -text
+*.waff2 -text


Alkarex (approved):

Alkarex:
Hum, while that probably hides another misconfiguration elsewhere, I do not think that would hurt.
And you do not observe the same problem for our other binary files such as .gif, .ico ?

ravenx99:
Alkarex, it's an interesting question.  git did not complain about .gif or .ico files, just .png and .waff.   A quick check shows that the first .gif I found in the repo does contain a CRLF sequence, so I can't say why git ignored gifs but not pngs.  (I wonder why git doesn't recognize them all as binary file types to start with.)  But I do know that this change let me move forward without git trying to make me commit these "changes".

Alkarex:
Thanks 👍 Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md

.gitattributes
```diff
@@ -1 +1,4 @@
 * text=auto eol=lf
+*.png -text
+*.waff -text
```
Alkarex, line 3: I have just noticed a typo: it should be `*.woff`
@ravenx99 Could you make a new PR?

PR #3219, opened 2020-10-10
Enabling url parameter in user queries

knasdk:
Closes #

- When selecting saven user queries, the URL can now be set

1. Goto Configure -> User queries and create a name in the name field, and a search term in the url field, and submit
2. Go back to the frontpage and select the query, and the search is done

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Alkarex:
Here is how it looks like:

image

What do you think @aledeg ?
The URL field is maybe not so user-friendly. Good enough for now, or can you think of a quick improvement?

aledeg:
@knasdk @Alkarex I don't understand why there is a need for that. Could you explain it to me?

Alkarex:
@aledeg It is not something that I need myself, but the idea is to be able to edit an existing user query.
I think it initially started due to the fact that @knasdk thought that the name field was the actual query field.

aledeg:
If this is the case, we might need to rework everything because we had some complaints about that feature being not usable and hard to understand.

Alkarex:
Meanwhile @aledeg , do you believe showing the URL (editable or not) as this PR does could improve the current situation?

aledeg:
I doubt it but we could still add this feature until we have something better.

aledeg:
@knasdk @Alkarex I've put some thoughts for the edit tool. I am not convinced that we should allow the raw URL modification. I think this will be prone to faulty configurations. This will increase the flow of issue.

However, being able to modify a query from here will be a nice addition. I think the best way to do that is to provide tools that does that without typing directly the URL.
We can add some actions to add and remove filters (drop-down for categories, feeds, states, etc).
It will be probably longer but also safer.

My 2¢

knasdk:
@aledeg, @Alkarex - I use FreshRSS as a research library, and what I need is a way to search all news. E.g. `?search=!#sport AND !#soccer` or `?search=Corona* OR Covid-?19` etc. What you propose @aledeg about a tool for editing, would be fine as an additional opportunity, but I think it would be great to leave open the possibility to write raw searches too.

aledeg (changes requested):
@knasdk As the need to edit the raw query is needed, here are my ideas to keep it safe. I don't want to have open issues because someone messed things up without knowing.
First, I will add a notice to the user that he is modifying the query and this might break if there is problem in the syntax.
Second, I will hide it by default and display it only after a user action.

aledeg:
@knasdk I've made some tests and we can not merge that as-is. It's working for your use case but it breaks the original feature.
When I try to edit an existing query, it adds a `search` parameter in the URL.
Before, I had:
Screen Shot 2021-01-06 at 07 18 02
After I pressed the `Submit` button, I have:
Screen Shot 2021-01-06 at 07 18 18
And here is the content of my configuration file:
```php
  'queries' => 
  array (
    0 => 
    array (
      'name' => 'Query n°1',
      'search' => 'test',
      'url' => '?search=./?search=test&amp;rid=5ff5aa5fa175f',
    ),
  ),
```
The other thing that is not working is if I change completely the url, I have discrepancies between the search param and the url param in the configuration file which is reflected on the query page.
Screen Shot 2021-01-06 at 07 25 42
```php
  array (
    0 => 
    array (
      'name' => 'Query n°1',
      'search' => 'test',
      'url' => '?search=helloworld',
    ),
  ),
```

If you want to modify the whole URL, you need to extract those parameters from the URL and push then correctly in the configuration file. If you only want to modify the search then, you should expose only the search and not the whole URL.

knasdk:
@aledeg, @Alkarex - I'm still not sure how you intended this search mechanism to work. I named the label `URL`, because that's how I reasoned from the code, it was supposed to work. 

@aledeg I do not see it clearly what you mean by hiding the editing feature. I mean, when a user clicks "Add a query", it would not be intuitive to hide what he came to do... Maybe I misunderstood you.
@Alkarex I agree with you that it might be better to ad the `?search` query to the URL when outputting the user queries selection list in the menu. Is that what you meant too?

aledeg:
@knasdk I think there is a misunderstanding on how this feature works. That means there is room for improvement to make it work properly.

## What is its purpose
The idea of the user query is to act as a bookmark to get the information you need quickly. The user query is a search you perform on a regular basis so you want to keep it to avoid filtering the same thing over and over.

## How it works
It works as a bookmark like a bookmark in Firefox. That means that it uses the current URL to generate the user query. It takes all the parameters available in the URL (search, state, category, feed, label, order) and convert that in the search query.
So if you need a specific search to be used as a user query, you need to validate that search then add it as a user query.

## What could we do to improve
Probably the wording `Add a query` is misleading. At the time I wrote that, it was the only thing I had in mind. Maybe we should change that to something more explicit like `Save current query` or `Bookmark current query`.
Probably we need a tool to update user queries as I mentioned beforehand but that outside of the scope for now.

## What I think you are trying to achieve
For what I understand, you want to query your feeds with a search parameter. It is possible at the moment to do that, even if the UI is not explicit about that. UI is definitely not my strong suit.

## How you should do what you are trying to achieve
There is only 2 steps:
1. create your perfect search by filtering what you need (ex: `!#sport AND !#soccer`).
2. when you're satisfied with the search result, click the `Add a query` link.
And you're done!

You could even go further with your search by adding sort order, filtering by state, filtering by category, filtering by feed or filtering by label.

## Final thoughts
I really think that what you need does not require this PR. I think we need to improve the wording and to modify the documentation. This will clarify how this feature works.
Later, it will be nice to have a way to modify a query because at the moment, the only option is to drop it and recreate it.
I hope this clarify the use of the user query feature.

knasdk:
@aledeg WHAAAT - I had not figured out that it was working like that. Now I see, and you're correct - that renders this PR pointless. Cool. I'm sorry for taking your time for nothing.

aledeg:
@knasdk I am glad that we cleared things up. You didn't take my time for nothing.
We agree that there is a problem and we need to address that. The problem was not the one that we thought on the first place :)

I keep this PR open until I am done with modifications to make things clearer.

knasdk:
@aledeg - With peace in mind I leave my baby in your custody :smile:

PR #3231, opened 2020-10-18
Remember sidebar position

### LINKED ISSUE #3209: Feed List Jumps to Top When Selecting Feeds Near the Bottom of the Sidebar
I have ~100 feeds in the left sidebar. Sometimes I want to start viewing a feed near the bottom of the list (for example #75),  so I scroll down and select it. The feed articles are displayed properly, but the feed list in the sidebar jumps back to the top. I would expect the selected feed name to remain visible in the sidebar at all times.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3209
Extracted from https://github.com/FreshRSS/FreshRSS/pull/3185

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -697,6 +697,9 @@ function init_column_categories() {
 		return;
 	}
 
+	//Restore sidebar scroll position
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
lgtm

PR #3250, opened 2020-11-11
API fix continuation type

### LINKED ISSUE #3247: Reeder not working with FreshRSS greader API
FreshRSS Version: 1.17.1-dev

When use greader API, Reeder can not get any items.

I have found some decode errors in Console for Reeder.
```
syncer error Optional(Swift.DecodingError.typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "continuation", intValue: nil)], debugDescription: "Expected to decode String but found a number instead.", underlyingError: nil)))
```

but fever API is ok in Reeder.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3247

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -673,7 +673,7 @@ function streamContents($path, $include_target, $start_time, $stop_time, $count,
 	if (count($entries) >= $count) {
 		$entry = end($entries);
 		if ($entry != false) {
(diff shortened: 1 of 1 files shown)


jim3ma (approved):
LGTM

PR #3266, opened 2020-11-25
Fixes a minor jaggy motion of the sidebar.

PAHXO:
Fixes jaggy motion of the sidebar while scrolling on mobile clients.

I didn't try many values aside from %200 cause it worked very well from the first time. I was using a Galaxy S8+

PS: Also, I use a value for the width of 50% instead of 90% for better aesthetics.

Closes #

- Adding the variable "height" with a value of 200% to the sidebar
-
-

1. Any smartphone
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -1296,6 +1296,7 @@ input:checked + .slide-container .properties {
 
 	.aside:target {
 		width: 90%;
+		height: 200%;
 	}
 
 	.aside_feed .configure-feeds {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -1296,7 +1296,7 @@ input:checked + .slide-container .properties {
 
 	.aside:target {
 		width: 90%;
-		height: 200%;
+		height: 100vh;
 	}
 
 	.aside_feed .configure-feeds {


Alkarex:
Thanks! Would you be able to send a screenshot illustrating the differences with and without this fix?

PAHXO:
Sure! Well, sorry that I don't feel like providing a video but hopefully pictures will do.

See here sometimes while scrolling down the URL bar on Android Google Chrome hides itself
example

Then, it becomes like this 
example2
Resulting in a gap that forms here? Though, it mostly shows when you scroll at a higher pace than usual 

PS: It is definitely not caused by my previews modifications to the bar.

Frenzie:
> PS: Also, I use a value for the width of 50% instead of 90% for better aesthetics.

Confused me for a second, not in this PR. But as a general PSA, 50% results in:

Screenshot_2020-11-26_16-02-24

p/themes/base-theme/template.css
```diff
@@ -1296,6 +1296,7 @@ input:checked + .slide-container .properties {
 
 	.aside:target {
 		width: 90%;
+		height: 200%;
```
Frenzie:
```suggestion
		height: 100vh;
```

Frenzie (approved):

PR #3268, opened 2020-11-27
Better Date/Time improvements 

PAHXO:
#853

[Time format with a string](https://github.com/FreshRSS/FreshRSS/pull/3268/commits/325cf7b015ab5f3e88c3d757cfe5aaecc01066bf)

ffed2

A better way to manage the date/time.
Maybe if we could give the user access to change the string somehow?
It could be better if the user can see the time passed since that post 

- Better management for the Date/Time 
  * 12/24 Hours clock, Names of Days, Months, Years.
feed23

- Showing up the time passed for the post. E.g 5 hours ago, 3 months ago.. etc

- Ui settings to make it easily customizable. . _sadly too much for me_

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -44,6 +44,6 @@
 			?></span><?php
 		endif;
 	?></a></li>
-	<?php if ($topline_date) { ?><li class="item date"><time datetime="<?= $this->entry->machineReadableDate() ?>"><?= $this->entry->date() ?></time>&nbsp;</li><?php } ?>
+	<?php if ($topline_date) { ?><li class="item date"><time datetime="<?= $this->entry->machineReadableDate() ?>"><?= date('Y-m-d h:i A',strtotime($this->entry->machineReadableDate())) ?></time>&nbsp;</li><?php } ?>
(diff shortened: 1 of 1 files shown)


PAHXO:
[Time passed](https://github.com/FreshRSS/FreshRSS/compare/master...thamerx:patch-3)
With this one, we can see the time passed.

I got this function from [StackOverflow ](https://stackoverflow.com/questions/2915864/php-how-to-find-the-time-elapsed-since-a-date-time)



                  class CustomFunctions {
	 		public static function humanTiming ($time)
			{

				$time = time() - $time; // to get the time since that moment
				$time = ($time<1)? 1 : $time;
				$tokens = array (
					31536000 => 'year',
					2592000 => 'month',
					604800 => 'week',
					86400 => 'day',
					3600 => 'hour',
					60 => 'minute',
					1 => 'second'
				);

				foreach ($tokens as $unit => $text) {
					if ($time < $unit) continue;
					$numberOfUnits = floor($time / $unit);
					return $numberOfUnits.' '.$text.(($numberOfUnits>1)?'s':'');
				}

			}	
                  }




I already tried those Date/time improvements and they worked great! for most feeds

fffed

aledeg:
I don't think we should add that in the code. The chosen format might not suit everyone. That's why I've written an extension to handle dates. See in the extension list for more information.
The extension modify the date depending on the chosen configuration. If there are other format needed, it might be a better idea to support that in the extension.

PR #3273, opened 2020-12-12
Session compatibility PHP 7.1 and older

### LINKED ISSUE #3239: Password mismatch on new install
Everything installed successfully, but I can't login. I even tried this twice, copied my password and pasted it elsewhere. Even when it and redid my password in the config with bcrypt to reset it. Again I tried several times. I dont know what the issue is.

In the log.txt I have
user=admin, nonce=, c=............(edited this part out)

Using version 1.17.1-dev

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3239
Reason: https://php.net/session-write-close used to return void and not boolean before PHP 7.2

### DIFF
--- a/lib/Minz/Session.php
+++ b/lib/Minz/Session.php
@@ -12,14 +12,16 @@ class Minz_Session {
 	private static $locked = false;
 
 	public static function lock() {
-		if (!self::$volatile && !self::$locked && session_start()) {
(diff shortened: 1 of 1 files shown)


Alkarex:
The regression was introduced by https://github.com/FreshRSS/FreshRSS/pull/3096

PR #3283, opened 2020-12-24
Extract autoloading process

aledeg:

- Change autoloading to accomodate namespaces

1. Use the application to check than everything is working as before.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/lib/autoload.php
+++ b/lib/autoload.php
@@ -1,26 +1,74 @@
 <?php
 
-function classAutoloader($class) {
-	if (strpos($class, 'FreshRSS') === 0) {
-		$components = explode('_', $class);
-		switch (count($components)) {
-			case 1:
-				include(APP_PATH . '/' . $components[0] . '.php');
-				return;
-			case 2:
-				include(APP_PATH . '/Models/' . $components[1] . '.php');
-				return;
-			case 3:	//Controllers, Exceptions
-				include(APP_PATH . '/' . $components[2] . 's/' . $components[1] . $components[2] . '.php');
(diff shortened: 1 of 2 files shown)


lib/lib_rss.php
```diff
@@ -15,6 +15,8 @@ function mb_strcut($str, $start, $length = null, $encoding = 'UTF-8') {
 	openlog('FreshRSS', LOG_CONS | LOG_ODELAY | LOG_PID, LOG_USER);
 }
 
+require_once 'autoload.php';
```
```suggestion
require_once LIB_PATH . '/autoload.php';
```

Just a suggestion to be more similar to what we do elswhere

lib/autoload.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Register path in which to look for a class.
+     */
+    public function registerPath($path) {
+        if (is_string($path)) {
+            $this->searchPath[] = $path;
+        }
+        if (is_array($path)) {
+            array_push($this->searchPath, ...$path);
```
Alkarex, line 14: I have just double-checked that the splat operator is available from PHP 5.6+, so fine

Alkarex (approved):

lib/autoload.php
aledeg: I've checked that before writing it :)

lib/lib_rss.php
aledeg: Fixed

PR #3287, opened 2020-12-26
Auto-renew login cookie

### LINKED ISSUE #3169: Suggestion: Make the login timeout larger than 30 days

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3169#issuecomment-685983797
Alternative to https://github.com/FreshRSS/FreshRSS/pull/3170

When we get the long-term login cookie (i.e. when starting a new session), renew it at the same time for the same duration

### DIFF
--- a/app/Models/Auth.php
+++ b/app/Models/Auth.php
@@ -269,7 +269,23 @@ class FreshRSS_FormAuth {
 		}
 
 		$credentials = @file_get_contents($token_file);
-		return $credentials === false ? array() : explode("\t", $credentials, 2);
+		if ($credentials !== false && self::renewCookie($token)) {
+			return explode("\t", $credentials, 2);
+		}
+		return [];
+	}
+
+	private static function renewCookie($token) {
+		$token_file = DATA_PATH . '/tokens/' . $token . '.txt';
+		if (touch($token_file)) {
+			$conf = Minz_Configuration::get('system');
+			$limits = $conf->limits;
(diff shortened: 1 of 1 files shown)


app/Models/Auth.php
```diff
@@ -269,7 +269,23 @@ public static function getCredentialsFromCookie() {
...
+		}
+		return [];
+	}
+
+	private static function renewCookie($token) {
+		$token_file = DATA_PATH . '/tokens/' . $token . '.txt';
+		if (touch($token_file)) {
+			$conf = Minz_Configuration::get('system');
+			$limits = $conf->limits;
+			$cookie_duration = empty($limits['cookie_duration']) ? 2592000 : $limits['cookie_duration'];
```
Frenzie: I'm still very frequently annoyed by this imho extremely low default value. Almost every time I start Steam, I have to log in again. Excuse me for only playing a game once every month or two…

Speaking as a user, I don't think this should be an "instead" at all, but an "in addition to." I'll leave the exact value in the middle, but I'll repeat that I'd prefer to see at the very least 3 months by default.

Alkarex: Ok @Frenzie , let's re-open https://github.com/FreshRSS/FreshRSS/pull/3170 and change it to 3 months (could you patch it?), in addition of the current PR

PR #3289, opened 2020-12-26
Layout add subscription

Alkarex:
I kept hitting the 'Add' button for a category when adding a new feed, so I have moved the creation of a category (which is small) to the top of the page


Alkarex:
image

aledeg (approved):

PR #3290, opened 2020-12-27
Add Minz namespace

aledeg:

- add namespace for Minz
- tidy Minz classes

1. Nothing should have changed since it involves mostly moving code around.

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


aledeg:
@Alkarex @marienfressinaud Should I continue my changes? It's working for what I could see during my tests but I would like to go further. If you think this is not the right time to do that, I'll stop to focus on something else.

Alkarex:
@aledeg That looks fine to me, if you like it for Minz. But while I see the benefit of using namespaces for distributed libraries, the added value for the core of a software is much less clear, with the risk of being more negative (more code, more complexity, ugly syntax, reduced performance) than positive (I am not sure what the positive aspects are, to be honest, if any).
In particular, I would rather not see a `\` in front of everything 😟
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

> When calling the global function in a non-fully qualified way, it is slower. Running functions inside a namespace are always slower than running global functions.

Frenzie:
> When calling the global function in a non-fully qualified way, it is slower. Running functions inside a namespace are always slower than running global functions.

Ouch! Is that still true in newer PHP 7.x/8 btw?

Speaking purely from a reading perspective I find `Minz_Something` more pleasant to look at than `Minz\Something`, and setting aside performance there seems to be no practical distinction between a namespace and a prefix. I'm inclined to think that as a rule of thumb, using a particular namespace for "simpler" code often obscures what's happening for fake simplicity (i.e., you need to look at the top of the file rather than it just being clear).

Alkarex:
@Frenzie I have not checked with PHP 8, but I would think so, due to the design. It would be nice to find an update on this topic. The *solution* so far being to have everything, including native functions, fully qualified (like `\explode()` instead of just `explode()`), as soon as a file is using namespaces.

aledeg:
So there is no point I continue this.
But here are a few thoughts.

I do not see any advantage to have Minz as a framework. Because we are not using it elsewhere, we could drop it and include it in FreshRss and use it as a real core. I feel that sometimes it drags us more than it helps. I am not saying that we just remove its logic but that we include it in FreshRss classes instead of having them in the lib folder. I find that keeping it that way is kind of awkward because it is part of the core without being in it.

I see advantages to use namespaces especially if we follow PSR recommendations (not limited to namespace). It could be easier to interface with other tools if the need to include libraries arise.
It could be also useful when replacing a class we are using by another one using the same signature. For example, let say than our Request class has some issues and we want to replace it with Symfony Request class, the only thing to do is change the namespace of the `use` directive. On top of my mind, PSRs could be applied also to logging and caching.

For the performances, I used to work with a company and our APIs had millions of hits per days, maybe per hours. The performance issue rarely arose because we were using non-fully qualified functions but because logic or assumption were flawed. If performance is an issue and we need to address it, it would be a better idea to have a great caching system and efficient database queries.

Regarding syntax, I kept `Minz\Something` because I did not want to work for nothing. My idea is to have just `Something` in the code with a `use` directive at the top of the file.

Anyway, we can close this if needed.

Alkarex:
Indeed @aledeg , having Minz in a separate library is not so much needed (anymore) and some aspects could be simplified.
Regarding performance, I have already tried to pay great care to the efficiency of our database queries, and we do have some forms of caching (`op_cache` and HTTP Not Modified) :-)

Frenzie:
> So there is no point I continue this.

At the very least, you shouldn't stop on my account if you think it's worthwhile. I'll get over looking at `\` soon enough. ;-)

> For example, let say than our Request class has some issues and we want to replace it with Symfony Request class, the only thing to do is change the namespace of the use directive.

Although this doesn't seem terribly convincing to me. Roughly speaking either it's designed to be compatible or it's not. ^_^

aledeg:
Spending some time on my extensions, I found a new purpose for namespace.
If I need classes in my extension, I could use a namespace instead of having to name all my classes with a prefix to avoid potential conflicts with other extensions.

marienfressinaud:
Hey! Happy new year since you're all here :)

Using namespace makes sense in extensions and I do use them. I don't think aesthetic is a strong argument since it's mainly personal taste (both styles are fine to me). Performances should be compared before and after, but unless it becomes noticeable to the end user, it should not be a strong argument against using namespaces in my opinion. Performances ARE important, especially for FreshRSS, but it can be counter-balanced by other arguments (readability, maintainability, etc.)

In the other hand, when I see 130 changed files, 1837 new lines / 1600 removed, even if it's mostly find/replace, it raises my warning flag. I did this kind of PR in the past, but it's rarely bug-free and always implies the code to not be reviewed properly. It might be ok, but the added value must be obvious. PSR recommendations are great, but are they worth to spend our energy in them? I think it's totally fine to accept that FRSS has a poor design. My opinion is that FRSS is good enough as it is. It could be a LOT better, but I don't imagine it being radically different.

@aledeg I know it's painful to work on something and to see little or no enthusiasm in the others. So to be clear: I do appreciate your work here. I'm reluctant mostly because I think the balance pros/cons is negative (and also because I'm afraid of too many changes at once ^^). Next time, I suggest that we start a discussion before doing too much work on this kind of big PR. Also I think we all have different priorities or opinions on what should be FRSS in the next months/years. Maybe we can start a discussion so everyone can express his desires? It can be the occasion to create a small roadmap for 2021, with each of us responsible of a part :)

aledeg:
No hard feelings :)
I understand.

Alkarex:
@marienfressinaud thanks and Happy New Year 2021 to you and all 🎆

PR #3274, opened 2020-12-12
Docker image for oldest supported PHP version

Alkarex:
Add a Docker file, which can be used to test FreshRSS with our oldest supported PHP version.

See https://pkgs.alpinelinux.org/packages?name=php5&branch=v3.4&arch=x86_64

PHP 5.6.36-r0

Issue #3191, opened 2020-09-25
Docker: support MySQL 8+


tvlpirb:
I'm on the installation page in the web interface, when entering my database details I get 
`Blast! Verify your database configuration. : Access to database is denied for `freshrss`: SQLSTATE[HY000] [2054] The server requested authentication method unknown to the client`

I am running FreshRSS following the docker instructions and hence used the following commands to create the database:
```
$ docker volume create mysql-data
$ docker run -d --restart unless-stopped --log-opt max-size=10m \
  -v mysql-data:/var/lib/mysql \
  -e MYSQL_ROOT_PASSWORD=example \
  -e MYSQL_DATABASE=freshrss \
  -e MYSQL_USER=freshrss \
  -e MYSQL_PASSWORD=example \
  --net freshrss-network \
  --name mysql mysql
```

On the web interface I have entered the following details host = mysql, username = freshrss, password as the one I changed it to, database = freshrss and database prefix is left blank.

I've tried to exec into the docker container and was able to use mysql cli. Is there anything I'm missing?

Alkarex:
Sorry for the delay, this went under my radar during a busy period. This is a sign of MySQL version (8+) that is too recent compared to the PHP version in our current Docker images (7.3.x), which does not support `caching_sha2_password` (requires PHP 7.4.0+ and best 7.4.4+, which should soon be in Alpine 3.13 and Debian 11 Bullseye).

* https://php.net/ref.pdo-mysql
* https://github.com/FreshRSS/FreshRSS/issues/2998#issuecomment-633116735
* https://www.php.net/ChangeLog-7.php#7.4.3
* https://bugs.php.net/bug.php?id=79011

https://php.net/mysqli.requirements

> This is because MySQL 8 defaults to caching_sha2_password, a plugin that is not recognized by the older PHP (mysqlnd) releases. Instead, change it by setting default_authentication_plugin=mysql_native_password in my.cnf. The caching_sha2_password plugin will be supported in a future PHP release.

Here is an example of configuration for MySQL 8.0:

```sh
docker run -d --restart unless-stopped --log-opt max-size=10m \
  -v mysql-data:/var/lib/mysql \
  -e MYSQL_ROOT_PASSWORD=freshrss_root_pass \
  -e MYSQL_DATABASE=freshrss_db \
  -e MYSQL_USER=freshrss_user \
  -e MYSQL_PASSWORD=freshrss_pass \
  --net freshrss-network \
[...]

PR #3294, opened 2020-12-28
Docker new :newest image

Alkarex:
New optional Docker image to test upcoming software versions, using `alpine:edge`
Similar than https://github.com/FreshRSS/FreshRSS/pull/3274 (oldest image) but for newest.
Useful for e.g. testing PHP8 https://github.com/FreshRSS/FreshRSS/issues/3082 , MySQL 8 https://github.com/FreshRSS/FreshRSS/issues/3191


Docker/Dockerfile-Newest
```diff
@@ -41,7 +41,7 @@ RUN rm -f /etc/apache2/conf.d/languages.conf /etc/apache2/conf.d/info.conf \
 	# https://bugs.php.net/bug.php?id=78681 \
 	sed -r -i "s/php8_module/php_module/" \
 		/etc/apache2/conf.d/php8-module.conf && \
-	ln -s /usr/bin/php8 /usr/bin/php && \
+	ln -s /usr/bin/php8 /usr/bin/php ; \
```
Frenzie: I'm not sure I follow? If it fails we assume all's well and we want to continue with the rest?

Alkarex: Yes indeed. I do expect that php8 will be symlinked from php out of the box at some point

Alkarex: And `ln` fails if the target already exists

Frenzie: Seems (mildly) risky to depend on failure to be intentional in that sense. :-)

Alkarex: It is "only" an image for development, using an unstable base. `php` is not installed if not for what we have ourselves installed higher up, so it cannot be something else. So it is just a way to say *Create symlink if it does not exist already* . Feel free to propose a nicer syntax if you like :-)

Frenzie: Hence mildly.

I don't know if this is a nicer syntax, but this would be the correct behavior:
```suggestion
	if [ ! -f /usr/bin/php ]; then ln -s /usr/bin/php8 /usr/bin/php; else true; fi && \
```

Alkarex: Let's go for it :-)

Frenzie (approved):
lgtm; it's 99 % identical to the regular one after all. ;-)

PR #3297, opened 2020-12-28
Add a file for each PDO class

aledeg:

- split Minz pdo file to have a class per file

1. Use the application, as the modification changes the database access, you'll notice right away if it's not working.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

Before, we had 5 classes in the ModelPdo file. It was bad for 2 reasons.
The first reason is that it is considered bad practice to have multiple
class in one file. This is especially true when using autoloading. On top
of that it is less readable considering the size of the file. The second
reason is that so far we were lucky. Everytime we needed to access the
database, it was through the ModelPdo class which loads all the other
classes. If we want to access directly the connection, it wont be loaded.
On top of that, the system is configured to work on a single database,
but as we have every connection definition in a single file, all classes
were loaded at the same time. Thus using memory and processing time for
nothing.
Now, we have a file for each class. To work with autoloading, classes
were slightly renamed to match autoloading rules.

### DIFF
--- a/app/Models/DatabaseDAO.php
+++ b/app/Models/DatabaseDAO.php
@@ -241,7 +241,7 @@ class FreshRSS_DatabaseDAO extends Minz_ModelPdo {
 		$sqlite = null;
 
 		try {
-			$sqlite = new MinzPDOSQLite('sqlite:' . $filename);
+			$sqlite = new Minz_PdoSqlite('sqlite:' . $filename);
 		} catch (Exception $e) {
 			$error = 'Error while initialising SQLite copy: ' . $e->getMessage();
 			return self::stdError($error);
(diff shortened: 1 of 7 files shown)


Alkarex:
Good idea. See also https://github.com/FreshRSS/FreshRSS/pull/3170#discussion_r549037561 a couple of other classes need the same

Alkarex (approved):

aledeg:
@Alkarex I cannot find those classes you've mentioned. I search for multiple class definitions per file but couldn't find any.

Alkarex:
It is not in Minz: https://github.com/FreshRSS/FreshRSS/blob/ac0d90c1008bce32c56e49cb642d40391e45e0a5/app/Models/Auth.php#L240

PR #3299, opened 2020-12-28
Change the autoload process

aledeg:

- Change autoloading to accomodate namespaces

1. Use the application to check than everything is working as before.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

~~Before, the autoload process was too restricted. It was really dependant on our
code tree. It was hard to add more classes to be loaded automatically. On top
of that, it did not support autoloading classes following the PSR-4 recommendation.~~

~~Now, the autoload process is more open. It supports partially the PSR-4 recommendation,
there is no specific code to load Minz classes or PHPMailer classes. This is the
starting point to reorganize the codebase to introduce long waiting changes as seen
in #789. It would be a nice to later rework the tree, rename classes, and add namespace
in a fashion that follows the PSR-4. Then specific FRSS workarounds in the autoload
could be dropped.~~

### DIFF
--- a/lib/autoload.php
+++ b/lib/autoload.php
@@ -1,0 +1,45 @@
+<?php
+
+const PREFIX_CLI_I18N = 'Cli\\I18n\\';
+const PREFIX_CSSXPATH = 'Gt\\CssXPath\\';
+const PREFIX_PHPMAILER = 'PHPMailer\\PHPMailer\\';
+
+function includeFreshrssClass($class) {
+	$components = explode('_', $class);
+	switch (count($components)) {
+		case 1:
+			include APP_PATH . "/{$components[0]}.php";
+			break;
+		case 2:
+			include APP_PATH . "/Models/{$components[1]}.php";
+			break;
+		case 3:	//Controllers, Exceptions
+			include APP_PATH . "/{$components[2]}s/{$components[1]}{$components[2]}.php";
+			break;
+	}
+}
+
+function requireClass($class, $path, $search, $replace, $prefix = null) {
+	if ($prefix !== null) {
+		$class = substr($class, strlen($prefix));
+	}
+	require $path . str_replace($search, $replace, $class) . '.php';
+}
+
+function classAutoloader($class) {
+	if (str_starts_with($class, 'FreshRSS')) {
+		includeFreshrssClass($class);
+	} elseif (str_starts_with($class, 'Minz')) {
+		requireClass($class, LIB_PATH . '/', '_', '/');
+	} elseif (str_starts_with($class, 'SimplePie')) {
+		requireClass($class, LIB_PATH . '/SimplePie/', '_', '/');
+	} elseif (str_starts_with($class, PREFIX_CLI_I18N)) {
+		requireClass($class, CLI_PATH . '/i18n/', '\\', '/', PREFIX_CLI_I18N);
+	} elseif (str_starts_with($class, PREFIX_CSSXPATH)) {
+		requireClass($class, LIB_PATH . '/phpgt/cssxpath/src/', '\\', '/', PREFIX_CSSXPATH);
+	} elseif (str_starts_with($class, PREFIX_PHPMAILER)) {
+		requireClass($class, LIB_PATH . '/phpmailer/phpmailer/src/', '\\', '/', PREFIX_PHPMAILER);
+	}
+}
+
+spl_autoload_register('classAutoloader');
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -23,6 +23,8 @@ if (COPY_SYSLOG_TO_STDERR) {
 	openlog('FreshRSS', LOG_CONS | LOG_ODELAY | LOG_PID, LOG_USER);
 }
 
+require_once LIB_PATH . '/autoload.php';
+
 /**
  * Build a directory path by concatenating a list of directory names.
  *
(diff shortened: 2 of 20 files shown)


lib/autoload.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+class ClassLoader {
+    private $searchPath = [];
```
Alkarex: @Frenzie Do you know why Travis did not catch the space indenting here?
@aledeg Could you change to tabs like the other files?

lib/autoload.php
```diff
@@ -0,0 +1,70 @@
...
+        if (false === strpos($class, 'FreshRSS_')) {
+            return;
+        }
+
+        // This match FRSS classes
+        $freshrssClass = str_replace(['FreshRSS_', '_'], '', $class);
+        foreach ($this->searchPath as $path) {
+            foreach (glob($path . DIRECTORY_SEPARATOR . '*', GLOB_ONLYDIR) as $subPath) {
+                $file = $subPath . DIRECTORY_SEPARATOR . $freshrssClass . '.php';
+                if (file_exists($file)) {
```
Alkarex: On a second read, I am a bit worried about nested loops and multiple IO operations to load a file for which we are supposed to know the exact path

lib/autoload.php
Frenzie: Isn't lib/* excluded or something? Please ping me in a couple of days if I forget to check.

Frenzie: Not /*. Strange.
https://github.com/FreshRSS/FreshRSS/blob/0a2d9b3b54ee51a3965d79c68409ef045ad5834b/phpcs.xml#L6-L9

lib/lib_rss.php
```diff
@@ -15,6 +15,8 @@ function mb_strcut($str, $start, $length = null, $encoding = 'UTF-8') {
 	openlog('FreshRSS', LOG_CONS | LOG_ODELAY | LOG_PID, LOG_USER);
 }
 
+require_once LIB_PATH . DIRECTORY_SEPARATOR . 'autoload.php';
```
Alkarex: `/` works on all platforms and is shorter :-)

lib/autoload.php
aledeg: I've changed to tabs.

lib/autoload.php
aledeg: I removed the loops by adding manually the path during configuration.

lib/lib_rss.php
aledeg: Except that if you read too fast like I did, you forget the `/` :)

lib/autoload.php
```diff
@@ -0,0 +1,74 @@
...
+	{
+		if ($file = $this->findFile($class)) {
+			require $file;
+		}
+	}
+
+	/**
+	 * Find the file containing the class definition.
+	 */
+	public function findFile($class) {
```
Alkarex: I am not so found of *searching*  for the class on disk, when we have the possibility of directly knowing the class filename 😟

aledeg: I agree. The initial problem was that it was too restricted and too dependent on the structure. It is still but in a different way.
Before, the rule was really strict on how to name the classes (ex: must end with _Controller), which give weird convention name and also discrepancies (ex: `FreshRSS_api_Controller` vs `FreshRSS_AlreadySubscribed_Exception`, search in Controllers for controllers but in Model for class with only one word after freshrss). That make no sense to me.
Now, there is only one rule, the class name has to match its file.

The idea with the namespace was to be able to match directly the file by using the namespace instead of relying on a weird convention on the class name to find the file to include. If you compare the process of including the class `Controller\ApiController` versus the class `FreshRSS_api_Controller`, it will be easier to convert to a path the later than the former.
With that in mind, we could remove most of the disk access to load the classes because we now right away where it is without relying on some convoluted method to extract the path from the class.

aledeg:
@Alkarex I am trying a different approach. But I have a blank page all the time.
And nothing is shown in the log. Do you have any ideas on how I could proceed to see more than nothing?
Thank you

Alkarex:
@aledeg Yes, just before `function classAutoloader($class) {`, add the following three lines:

```php
error_reporting(-1);
ini_set('display_startup_errors', 1);
ini_set('display_errors', 'On');
```

Gives something like that when a class does not exist:

```
sudo docker exec freshrss ./cli/export-opml-for-user.php --user freshrss
FreshRSS exporting OPML for user “freshrss”…

Warning: include(/var/www/FreshRSS/app/Models/EntryDAO.php): Failed to open stream: No such file or directory in /var/www/FreshRSS/lib/lib_rss.php on line 42

Warning: include(): Failed opening '/var/www/FreshRSS/app/Models/EntryDAO.php' for inclusion (include_path='.:/usr/share/php8') in /var/www/FreshRSS/lib/lib_rss.php on line 42

Fatal error: Uncaught Error: Class "FreshRSS_EntryDAO" not found in /var/www/FreshRSS/app/Models/Factory.php:37
Stack trace:
#0 /var/www/FreshRSS/app/Services/ExportService.php(32): FreshRSS_Factory::createEntryDao()
#1 /var/www/FreshRSS/cli/export-opml-for-user.php(19): FreshRSS_Export_Service->__construct()
#2 {main}
  thrown in /var/www/FreshRSS/app/Models/Factory.php on line 37
PHP Warning:  include(/var/www/FreshRSS/app/Models/EntryDAO.php): Failed to open stream: No such file or directory in /var/www/FreshRSS/lib/lib_rss.php on line 42
PHP Warning:  include(): Failed opening '/var/www/FreshRSS/app/Models/EntryDAO.php' for inclusion (include_path='.:/usr/share/php8') in /var/www/FreshRSS/lib/lib_rss.php on line 42
PHP Fatal error:  Uncaught Error: Class "FreshRSS_EntryDAO" not found in /var/www/FreshRSS/app/Models/Factory.php:37
Stack trace:
#0 /var/www/FreshRSS/app/Services/ExportService.php(32): FreshRSS_Factory::createEntryDao()
#1 /var/www/FreshRSS/cli/export-opml-for-user.php(19): FreshRSS_Export_Service->__construct()
#2 {main}
  thrown in /var/www/FreshRSS/app/Models/Factory.php on line 37
```

aledeg:
@Alkarex I can not spot the error.
I am trying to generate a cache file with all FreshRss classes and then load it. This will avoid to make IO calls. The cache file is generated properly but it seems I can not use it.
I do not know why.
If you could help me spot the missing element, that would be great!

aledeg:
I've changed the organization and I've spotted the error!

aledeg:
@Alkarex I think my work is done here. I kept the filesystem search for freshrss classes but I cached it. Now, it is search once and cached. Later, the cached version is retrieved automatically.
Let me know what you think.

aledeg:
@Alkarex I was working on that long overdue PR and I cannot understand why phpstan tests are failing. By branch is up-to-date with the parent branch and the failing tests are somehow unrelated to my changes.
Do you have any ideas?

lib/Autoload/UniversalClassLoader.php
```diff
@@ -0,0 +1,56 @@
...
+		if ($file = $this->findFile($class)) {
+			require_once $file;
+		}
+	}
+
+	/**
+	 * Find the file containing the class definition.
+	 */
+	protected function findFile($class) {
+		foreach ($this->registeredPath as $path) {
```
Alkarex: I am still not a fan of searching on disk, something very slow (especially since it needs to be done for every PHP file we load), when we already know where it is...

aledeg: I can add caching on that as well.

Alkarex: P.S. I know you have tried to explain higher up the rationale, but I still don't get it. It is significantly longer code (harder to understand, at least to me) with loops and disk scans, and it does not look like it solves any current problem, does it?

Alkarex:
> @Alkarex I was working on that long overdue PR and I cannot understand why phpstan tests are failing. By branch is up-to-date with the parent branch and the failing tests are somehow unrelated to my changes. Do you have any ideas?

@aledeg I guess it is because PHPStan fails to use the new autoloading, in particular loading `./data/cache/18dd0b692afe4afac65809e7eb081f9f0269e841.class.php'` leading to many errors

aledeg:
@Alkarex I've refactor the existing code and removed what I have done previously.
I extracted the autoload in its own file so we could include it easily in composer or phpunit configuration if needed.

aledeg:
@Alkarex I've added the use of the autoloading in the I18n section of the cli tool. We do not need to play around with `include` to make the thing working.
I've also added a namespace for said section.

PR #3303, opened 2020-12-28
Improved feed action filters

### LINKED ISSUE #2941: Filter-actions: mark old entries as read
Hi,
for news feeds I am subscribing to I want entries older than two days to be marked as read.
I tried to use `!date:P2D` in filter actions but this does not work (while it works fine when using it in the search field).
Are the filter-actions applied to already downloaded entries at all? Or does the filter only fire while downloading?
Further more with the filter configured, all new entries are marked as read.

### LINKED ISSUE #3142: Some Feeds had double article
I have install FreshRSS on a Plesk Webhosting with PHP 7.4.8

With some feeds I receive the articles twice. They are loaded, I read them and after 1-2 days the feeds appear again. They were not marked as unread. The same article appears twice in the overview and refers to the same blog link.

Some example: https://christophrox.wordpress.com/

Alkarex:
* Option to keep at max n unread articles per feed
#fix https://github.com/FreshRSS/FreshRSS/issues/2941

* Option to auto mark as read existing titles
#fix https://github.com/FreshRSS/FreshRSS/issues/3142

* Re-order some feed options


Alkarex:
Sorry that two features are in the same PR, but they are at least in distinct commits - I ended up developing / testing those two in a quite intertwined process :-)

Yellow: moved
Magenta: new

image

image

Alkarex:
@marienfressinaud  @aledeg @Frenzie Any feedback on those new features so far? I would like to add one more in the same style, based on whether articles are still on the RSS server or not

Frenzie:
Purely on the basis of your screenshot I think using *n* seems a bit awkward. Why not just say `Max articles to keep` or something? But I haven't had time to test it yet.

aledeg:
Are those rules working on a single feed or on all feed at once?
How those rules are applied? Based on the article title?

Alkarex:
@Frenzie Ok, I will try another text.
@aledeg For now, it is purely at feed level, but I do would like later to have 1) default feed options at category and/or global level; 2) similar options applied at category and/or global level. The max number of articles is keeping the newest ones. The filtering on titles is indeed based on perfect title match (in the future, we could potentially introduce a Levenshtein distance).

Regarding the new option I have just mentioned, it will be based on the same mechanism than what we already have to know whether we already have an article in database or not (which is based on either a provided GUID, or multiple parameters).

Alkarex:
@Frenzie Changed by *Max number of articles to keep unread*

aledeg:
@Alkarex I've created a discussion on that matter → #3337

math-GH:
any news to this PR?

Alkarex:
@mathContao Completing this PR is high on my TODO list :-)

Alkarex:
Seems to work fine. Let's merge and continue testing in edge.

I have several ideas for future PRs, such as:
* limiting the *mark as read if title exists* functionality in time (e.g. only apply if the existing title is at max 3 days old)
* have the *mark as read if title exists* functionality to match on a category or global scope, in addition to feed scope

PR #3305, opened 2020-12-29
Fix bugs in anomymous mode

Alkarex:
Login bug (submit button not working) and refresh bug (JS null exception, and then 403).

### DIFF
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -76,8 +76,9 @@ class FreshRSS extends Minz_FrontController {
 			}
 			if (!(FreshRSS_Auth::isCsrfOk() ||
 				(Minz_Request::controllerName() === 'auth' && Minz_Request::actionName() === 'login') ||
-				(Minz_Request::controllerName() === 'user' && Minz_Request::actionName() === 'create' &&
-					!FreshRSS_Auth::hasAccess('admin'))
+				(Minz_Request::controllerName() === 'user' && Minz_Request::actionName() === 'create' && !FreshRSS_Auth::hasAccess('admin')) ||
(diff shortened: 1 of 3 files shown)


app/FreshRSS.php
```diff
@@ -76,8 +76,9 @@ private static function initAuth() {
 			}
 			if (!(FreshRSS_Auth::isCsrfOk() ||
 				(Minz_Request::controllerName() === 'auth' && Minz_Request::actionName() === 'login') ||
-				(Minz_Request::controllerName() === 'user' && Minz_Request::actionName() === 'create' &&
-					!FreshRSS_Auth::hasAccess('admin'))
+				(Minz_Request::controllerName() === 'user' && Minz_Request::actionName() === 'create' && !FreshRSS_Auth::hasAccess('admin')) ||
+				(Minz_Request::controllerName() === 'feed' && Minz_Request::actionName() === 'actualize' && FreshRSS_Context::$system_conf->allow_anonymous_refresh) ||
```
Alkarex, line 80: Exception to permit anonymous refresh (if allowed in the config)

p/scripts/extra.js
```diff
@@ -51,7 +51,7 @@ function init_crypto_form() {
 
 	forgetOpenCategories();
 
-	const submit_button = document.querySelector('button[type="submit"]');
+	const submit_button = document.getElementById('loginButton');
```
Alkarex, line 54: In anonymous mode, we have two submit buttons!

p/scripts/main.js
```diff
@@ -1170,6 +1170,11 @@ function updateFeed(feeds, feeds_count) {
 function init_actualize() {
 	let auto = false;
 
+	const actualize = document.getElementById('actualize');
```
Alkarex, line 1173: In anonymous mode, there is not always an actualize button...

Alkarex:
Let's merge, as it is a mildly urgent fix

PR #3312, opened 2021-01-01
Add temp folder check during install

### LINKED ISSUE #3310: Install check that temp path is writeable
I'm running XAMPP 7.4.11 on Mac OS 10.15.7.

FreshRSS will not load feeds. This appears due to line 62 in the FreshRSS/constants.php file, where:

//Directory used for feed mutex with *.freshrss.lock files. Must be writable.
safe_define('TMP_PATH', sys_get_temp_dir());

directs to a non-writeable hidden system file outside XAMPP - in this case:

/var/folders/7x/sdw9rlr1573fwbdw96lwkg7r0000gp/T.

aledeg:
Closes #3310

- Add temp folder check during install

1. Make a new install with a writable temp folder.
2. Make a new install with a non-writable temp folder.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, the temp path was not check during install. With some configuration,
FRSS was not working because of a non-writable temp directory. It happened
with XAMPP on MacOS X but it might be the case for other platforms.
Now, the temp path is checked during install to make sure it is writable.


Alkarex (approved):

Alkarex:
Great 👍 
As a future improvement, maybe we could factorise *The HTTP server must have write permissions* to avoid having to translate it multiple times.

Alkarex:
Ah, and as another improvement, it would be nice to have the same check result in the *Installation check* view
https://github.com/FreshRSS/FreshRSS/blob/master/app/views/update/checkInstall.phtml

But this is already better than the current status :-)

aledeg:
As I changed the wording by using a parameter, there is a lot of things to factorize. Not only the HTTP part. Actually, the whole message can be modified to keep only one.

PR #3315, opened 2021-01-02
Fix author search link

### LINKED ISSUE #3314: Filtering by author is incomplete
When clicking on the author of an article, it filters by the author but only in the main stream. It is not a problem when the feed is available in the main stream. But when the feed is not available in the main stream, there is no result returned.

What's missing is the current context on the author filter.

aledeg:
Closes #3314

- fix author filter when the feed is not in the main stream

1. Select a feed that is not visible in the main stream
2. Click the author of an article to filter
3. There should be some results

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, when clicking on the author link, the search was done on the
main stream in the normal view. It's fine until the feed is not visible
in the main stream.
Now, the current context is used along with the search.

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -71,6 +71,13 @@ class Minz_Request {
 			'params' => self::$params,
 		);
 	}
+	public static function modifiedCurrentRequest(array $extraParams = null) {
+		$currentRequest = self::currentRequest();
+		if (null !== $extraParams) {
+			$currentRequest['params'] = array_merge($currentRequest['params'], $extraParams);
+		}
+		return $currentRequest;
+	}
 
 	/**
 	 * Setteurs
--- a/app/views/index/normal.phtml
+++ b/app/views/index/normal.phtml
@@ -88,7 +88,7 @@ $today = @strtotime('today');
 								echo $first ? _t('gen.short.by_author') . ' ' : '· ';
 								$first = false;
 					?>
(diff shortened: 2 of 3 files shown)


lib/Minz/Request.php
```diff
@@ -71,6 +71,15 @@ public static function currentRequest() {
 			'params' => self::$params,
 		);
 	}
+	public static function modifiedCurrentRequest(array $extraParams = null) {
```
Alkarex, line 74: Good idea; I had to double-check which type-hinting was available from which PHP version (this info is apparently gone from the current PHP documentation):
http://web.archive.org/web/20200101052646/https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.types
We should remember to use it more, and then even more when we drop PHP 5.x.

lib/Minz/Request.php
```diff
@@ -71,6 +71,15 @@ public static function currentRequest() {
...
+		$currentRequest = self::currentRequest();
+		if (is_array($extraParams)) {
+			foreach ($extraParams as $key => $value) {
```
Alkarex: Wouldn't `array_merge()` (or similar array function) work better here?

aledeg: Of course! I'll change that.

lib/Minz/Request.php
Alkarex: Ah, moved there :-) https://php.net/language.types.declarations

Alkarex: About dropping PHP 5: https://github.com/FreshRSS/FreshRSS/discussions/3321

PR #3285, opened 2020-12-26
Fix user queries with labels

aledeg:
Closes #3215

Changes proposed in this pull request:

- Fix user queries translation when they use labels as filter

How to test the feature manually:

1. Create a label
2. Apply it on articles
3. Select the label to display only its linked articles
4. Create a user query based on that result
5. Check the user query configuration page to check the translation

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/master/docs/en/developers/04_Pull_requests.md).

Before user queries with labels were not translated. Actually, it was not
even processed.
Now those user queries are translated properly.

See #3215

PR #3318, opened 2021-01-02
Update translation remember_categories

rom-1:
string "remember_categories" was not translated to german

Closes #

-
-
-

1.
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/i18n/de/conf.php
+++ b/app/i18n/de/conf.php
@@ -125,7 +125,7 @@ return array(
 			'all_articles' => 'Alle Artikel zeigen',
 			'all_categories' => 'Alle Kategorien',
 			'no_category' => 'Keine Kategorie',
-			'remember_categories' => 'Remember open categories',	// TODO - Translation
+			'remember_categories' => 'Geöffnete Kategorien merken',
 			'unread' => 'Nur ungelesene zeigen',
 		),
 		'sides_close_article' => 'Klick außerhalb des Artikel-Textes schließt den Artikel',


Frenzie (approved):
But if you also want to do this one, please add it to this PR rather than opening a new one. :+1: 

https://github.com/FreshRSS/FreshRSS/blob/b80bda2c95fe25ce96d1ca56c17b0b73efbdcca4/app/i18n/de/conf.php#L73

rom-1:
> But if you also want to do this one, please add it to this PR rather than opening a new one. 👍
> 
> https://github.com/FreshRSS/FreshRSS/blob/b80bda2c95fe25ce96d1ca56c17b0b73efbdcca4/app/i18n/de/conf.php#L73

I would like to, but I don't even know where this shows up in the UI, even when actively looking for this string. Without further context this string is difficult to translate.

Frenzie:
> I would like to, but I don't even know where this shows up in the UI, even when actively looking for this string. Without further context this string is difficult to translate.

It looks like there is no further context. Should we delete it @Alkarex?

Alkarex:
No @Frenzie , this is from a recent PR by @aledeg https://github.com/FreshRSS/FreshRSS/pull/3285
It is used in a dynamic way at:

https://github.com/FreshRSS/FreshRSS/blob/7bc2cc5825547f5b5cf15005fda937e06065b45d/app/views/configure/queries.phtml#L67

Which makes it difficult to find, so we should maybe add a comment there with the list of possible strings

Frenzie:
I see, thanks. At first I thought it might be a case of the GitHub GUI acting up (it seems to be surprisingly bad at finding strings in the source), but a local grep yielded no results either.

PR #3320, opened 2021-01-02
API more BazQux compatibility

Alkarex:
Implement compatibility fix, to allow the streamID / inclusion criteria
to be given as `&s=` parameter instead of in the URL path.
https://github.com/bazqux/bazqux-api#fetching-streams
This form of request is apparently sometimes used by FeedMe:

```

/api/greader.php/reader/api/0/stream/contents?output=json&n=100&xt=user%2F-%2Fstate%2Fcom.google%2Fread&ot=0&s=user%2F-%2Flabel%2FBlogs
```

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -979,8 +979,15 @@ if ($pathInfos[1] === 'accounts') {
 			if (!ctype_digit($continuation)) {
 				$continuation = '';
 			}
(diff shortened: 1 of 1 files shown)


Alkarex:
Merging, as I do not believe many will be interested in testing this edge-case PR :-)

PR #3325, opened 2021-01-03
German Translation, change "Submit" -> "Save" 

### LINKED ISSUE #3319: [SUGGESTION] German Translation: Renaming "Send" Button to "Save"
Reference: https://github.com/FreshRSS/FreshRSS/blob/082ef9854c41d0d4ef9a36ee2fa657311bd2f44f/app/i18n/de/gen.php#L27

Hi,

I just want to throw in a little suggestion: Currently the submit button in the user interface is translated to "send" (Abschicken) in German. While this is technically correct (because it is a html form which is sending a http request) it is very confusing for non-technical end users. In Germany, most of the UI's are working with a "SAVE" Button, so my suggestion would be to rename it to "Speichern" (Saving) which people are VERY familiar with. If this does not fit, alternatively, the button could be renamed to a more general "OK". But "sending" something just feels plain wrong in the german language if you don't really transmitting something particular.

I do not speak french, but the current translation seems to be "valider", which seems to be translating to "Confirm", so it seems like you guys are also not "submitting data" to the webserver in your language. ;)

Because this is a somewhat major change, I am not submitting a pull request, but open an issue for discussion.

Thanks

rom-1:
In German UI, "Submit" (Abschicken) is almost never used. Instead, people are used to have "Save" (Speichern) as text on their buttons. Discussion: https://github.com/FreshRSS/FreshRSS/issues/3319

Closes #3319 

-
-
-

1.
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/i18n/de/gen.php
+++ b/app/i18n/de/gen.php
@@ -24,7 +24,7 @@ return array(
 		'remove' => 'Entfernen',
 		'rename' => 'Umbenennen',
 		'see_website' => 'Website ansehen',
-		'submit' => 'Abschicken',
+		'submit' => 'Speichern',
 		'truncate' => 'Alle Artikel löschen',
 		'update' => 'Aktualisieren',
 	),


app/i18n/de/gen.php
```diff
@@ -24,7 +24,7 @@
 		'remove' => 'Entfernen',
 		'rename' => 'Umbenennen',
 		'see_website' => 'Website ansehen',
-		'submit' => 'Abschicken',
+		'submit' => 'Speichern',
```
Frenzie, line 27: The one place I didn't think of yesterday that's slightly different from the rest is the installation, btw, in case that's relevant.

Frenzie (approved):

app/i18n/de/gen.php
rom-1: I can't recall the installation UI exactly. 
-> If it is a single page with just a form asking for MySQL Credentials and so on, then "Save" (Speichern) would be ok. 
-> If it is a multi-page setup wizard, then a "Continue"/"Next" Button (Fortfahren/Weiter) would be preferred to get to page 2, page 3, etc.

Frenzie: Agreed, it should be changed to continue in parts.

Issue #2159, opened 2018-12-01
Suggestion: Drop support for Internet Explorer


pattems:
I'd initially started writing this in my most recent message on the scrollbar pr, but I figured it would get more visibility for potential discussion as an issue.

If there's not already a timeline in place for dropping support for Internet Explorer 11, I'd suggest that the time has come to start thinking about it. Per Microsoft, when they released Edge, "The latest features and platform updates will only be available in Microsoft Edge. We will continue to deliver security updates to Internet Explorer 11 through its supported lifespan."(1) As Windows 10 has an indefinite End Of Life date and IE11 will (allegedly, anyways) be supported until the EOL date of Windows 10, this leaves FreshRSS indefinitely stuck with either using web technologies available in IE11, circa 2015, or relying on polyfills for that browser, which has a rapidly shrinking userbase.

According to StatCounter's Global Stats, in the last 3 years, Internet Explorer Usage has dropped from 9.65% to 2.8% (2), and Desktop usage has fallen from 16.79% to 5.36%. (3)

I'm not suggesting anything as drastic as immediately dropping support for IE, but I think it's time to consider a timeline for phasing out support in the future.

Thoughts, Questions, Concerns?

(1) https://web.archive.org/web/20160716113141/https://developer.microsoft.com/en-us/microsoft-edge/platform/faq/ (General Platform Questions -> Will Internet Explorer 11 continue to receive updates?)
(2) http://gs.statcounter.com/browser-market-share#monthly-201511-201811
(3) http://gs.statcounter.com/browser-market-share/desktop/worldwide#monthly-201511-201811

Frenzie:
There's a difference between dropping support and accepting it won't get all the fancy new stuff. But I'm not sure if I understand why there should be a timeline? Unless by "timeline" you mean a set of conditions completely unrelated to time.

pattems:
In my mind, supporting a browser means that it should have feature parity with other browsers. If the aim of supported is just being usable even if the UX is different on different browsers, then I'll adjust my definition and close the issue.

Perhaps Roadmap would have been a better choice of terms, because I didn't suggest a condition. It could be time based (at the end of some year), or it could be conditional (when global/desktop usage drops under a certain threshold, when X amount of features would have to be polyfilled to reasonably support IE).  Again, this is only a relevant concern now and in the near future if the aim is feature-parity.

[...]

PR #3330, opened 2021-01-03
Add clipboard sharing

aledeg:
See #3327

- Add a clipboard sharing option

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

When the discussion #3327 started, I thought that the clipboard share
is a good idea. It never occurred to me because I do not use a
smartphone.
It might need some polishing but it seems pretty usable as-is.

### DIFF
--- a/app/i18n/de/gen.php
+++ b/app/i18n/de/gen.php
@@ -185,6 +185,7 @@ return array(
 	'share' => array(
 		'Known' => 'Known-Seite (https://withknown.com)',
 		'blogotext' => 'Blogotext',
+		'clipboard' => 'Zwischenablage',
 		'diaspora' => 'Diaspora*',
 		'email' => 'E-Mail',
 		'facebook' => 'Facebook',
--- a/app/i18n/nl/gen.php
+++ b/app/i18n/nl/gen.php
@@ -185,6 +185,7 @@ return array(
 	'share' => array(
 		'Known' => 'Known-gebaseerde sites',
 		'blogotext' => 'Blogotext',
+		'clipboard' => 'Klembord',
 		'diaspora' => 'Diaspora*',
 		'email' => 'Email',
 		'facebook' => 'Facebook',
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -946,7 +946,7 @@ function init_stream(stream) {
 			return true;
 		}
 
-		el = ev.target.closest('.item.share > a[href="#"]');
+		el = ev.target.closest('.item.share > a[data-type="print"]');
 		if (el) {	//Print
 			const tmp_window = window.open();
 			for (var i = 0; i < document.styleSheets.length; i++) {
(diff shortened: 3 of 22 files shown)


app/i18n/nl/gen.php
```diff
@@ -185,6 +185,7 @@
 	'share' => array(
 		'Known' => 'Known-gebaseerde sites',
 		'blogotext' => 'Blogotext',
+		'clipboard' => 'Clipboard',	// TODO - Translation
```
Frenzie:
```suggestion
		'clipboard' => 'Klembord',
```

p/scripts/main.js
```diff
@@ -960,6 +960,17 @@ function init_stream(stream) {
...
 
+		el = ev.target.closest('.item.share > a[data-type="clipboard"]');
+		if (el) {	//Clipboard
+			const textarea = document.createElement('textarea');
+			textarea.textContent = el.href;
+			document.body.appendChild(textarea);
+			textarea.select();
+			document.execCommand('copy');
+			document.body.removeChild(textarea);
+			return false;
```
Frenzie: You think https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText is too new?

app/i18n/de/gen.php
```diff
@@ -185,6 +185,7 @@
 	'share' => array(
 		'Known' => 'Known-Seite (https://withknown.com)',
 		'blogotext' => 'Blogotext',
+		'clipboard' => 'Clipboard',	// TODO - Translation
```
rom-1: `'clipboard' => 'Zwischenablage',`

rom-1 (changes requested):

p/scripts/main.js
aledeg: I didn't know about it. It seems that it's fine but it is not supported by Internet Explorer.
I was searching if we still support it but couldn't find that information in the doc. If we don't, I'll change that.

aledeg: I found the information and we still support IE.

Frenzie: IE is dead now, cf. what I wrote in https://github.com/FreshRSS/FreshRSS/issues/2159#issuecomment-443409899

Frenzie (approved):
But that IE thing isn't important. :+1:

aledeg:
I'll see what I can do with the clipboard API. I guess it would be better than what I have done.

aledeg:
@Frenzie Thank you! I didn't know about that API.
It works really well and it's really easy to use it.
I know another project where I will use it.

p/scripts/main.js
```diff
@@ -960,6 +960,12 @@ function init_stream(stream) {
 			return false;
 		}
 
+		el = ev.target.closest('.item.share > a[data-type="clipboard"]');
+		if (el) {	//Clipboard
+			navigator.clipboard.writeText(el.href);
```
Alkarex, line 965: Could you please test `if (navigator.clipboard) {}` to fail gracefully with e.g. IE11?

aledeg: done

p/scripts/main.js
Alkarex: Yes, I think it is OK not to have IE11 support for new features. It is better if it does not break/crash though, for now, but that can be tested with feature detection

Alkarex (approved):

Alkarex:
Nice, works fine :-)
As a potential future improvement, maybe auto-closing the menu and/or provide some kind of feedback could be nice.

PR #3331, opened 2021-01-03
Fix share option sort

aledeg:

- Change share option sort in the configuration page

1. Go to the share configuration page.
2. Display the share drop-down list.
3. Shares must be sorted alphabetically.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, share options were displayed in the order defined in the
configuration file. So the order reflected the share history and
was not really usable.
Now, share options are displayed in alphabetical order which makes
much more sense.

### DIFF
--- a/app/Models/Share.php
+++ b/app/Models/Share.php
@@ -43,6 +43,10 @@ class FreshRSS_Share {
 			$share_options['type'] = $share_type;
 			self::register($share_options);
 		}
+
+		uasort(self::$list_sharing, function ($a, $b) {
+			return strcasecmp($a->name(), $b->name());
+		});
 	}
 
 	/**


app/Models/Share.php
```diff
@@ -43,6 +43,10 @@ public static function load($filename) {
 			$share_options['type'] = $share_type;
 			self::register($share_options);
 		}
+
+		uasort(self::$list_sharing, function ($a, $b) {
+			return $a->name() < $b->name() ? -1 : 1;
```
Alkarex:
```suggestion
			return strcmp($a->name(), $b->name());
```

Would the above work? I think it would be cleaner, and takes into account equality (no impact in this case though)

aledeg: I'd rather use this:
```
		uasort(self::$list_sharing, function ($a, $b) {
			return $a->name() <=> $b->name();
		});
```

But it's not possible at the moment ;)

aledeg: @Alkarex I've changed it.

PR #3333, opened 2021-01-03
Change base extension class type

aledeg:

- Change the extension base class to abstract

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, there was some guidelines on how to use the extension class
and how to extend it. Those guidelines were defined as comments.
Now, those guidelines are enforced by the code itself. There is no
need for those comments anymore.

Beware, this change will break every extension that are not following
those guideline. I've made the change for the core extensions but works
need to be done for the others.

### DIFF
--- a/lib/Minz/Extension.php
+++ b/lib/Minz/Extension.php
@@ -3,7 +3,7 @@
 /**
  * The extension base class.
  */
-class Minz_Extension {
+abstract class Minz_Extension {
 	private $name;
 	private $entrypoint;
 	private $path;
@@ -31,11 +31,9 @@ class Minz_Extension {
 	 * - version: a version for the current extension.
 	 * - type: "system" or "user" (default).
 	 *
-	 * It must not be redefined by child classes.
-	 *
 	 * @param $meta_info contains information about the extension.
 	 */
-	public function __construct($meta_info) {
+	final public function __construct($meta_info) {
 		$this->name = $meta_info['name'];
 		$this->entrypoint = $meta_info['entrypoint'];
 		$this->path = $meta_info['path'];
@@ -50,8 +48,6 @@ class Minz_Extension {
 	/**
 	 * Used when installing an extension (e.g. update the database scheme).
 	 *
-	 * It must be redefined by child classes.
-	 *
 	 * @return true if the extension has been installed or a string explaining
 	 *         the problem.
 	 */
@@ -63,8 +59,6 @@ class Minz_Extension {
 	 * Used when uninstalling an extension (e.g. revert the database scheme to
 	 * cancel changes from install).
 	 *
-	 * It must be redefined by child classes.
-	 *
 	 * @return true if the extension has been uninstalled or a string explaining
 	 *         the problem.
 	 */
@@ -75,10 +69,8 @@ class Minz_Extension {
 	/**
 	 * Call at the initialization of the extension (i.e. when the extension is
 	 * enabled by the extension manager).
-	 *
-	 * It must be redefined by child classes.
 	 */
-	public function init() {}
+	abstract public function init();
 
 	/**
 	 * Set the current extension to enable.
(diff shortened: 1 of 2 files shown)


Alkarex:
Could you think of a way to have some kind of soft transition period? I would be worried about breaking many extensions...
In any case, it would be nice to go through https://github.com/FreshRSS/Extensions and see the status

aledeg:
I'll see what I can do.

aledeg:
@Alkarex I've added a clear error message in the logs if there are extensions that are not defined correctly.

aledeg:
@Alkarex Here is a comment from @kevinpapst 
> BTW. there is an abstract(?) extension - these changes are completely unnecessary.
Wasted time for many developers (you and me here and all the others involved everywhere else) without ANY benefit.
These methods should have been introduced in the extensions base class.

What do you think?

kevinpapst:
Sounds a bit harsh after reading it again 😁 but I stand by it (keep in mind that this is an outside view):
without knowing anything about your intentions ... just from looking at the code ... what is is the problem that you are solving?

This change seems to me like a typical developer decision: truly good intentions, but cost intense without a REAL benefit!? 
Most systems I know have an interface for a plugin and some kind of (abstract) base class, that implements all optional methods with an empty body. 

If the extension is fine after redeclaring the method the exactly same way than it was in the base class before - then the change is not necessary from MY perspective. Especially when you think about the amount of extensions existing, the required hours for all changes (fork, branch, change, push, PR, review, release - for every extension and often 2 developers involved) - AND then comes the pain for the admins/users during the update.

My extensions are already upgraded thanks to @aledeg - so I have no strong feelings for both ways, but I thought I add some context to my above comment.

Frenzie:
I didn't really have time to look at this earlier, but I agree with @kevinpapst. Why require extensions to implement more annoying boilerplate, which is only relevant when you're doing something like altering the database schema? The comments are simply wrong. Only an extension like TTRSS_API must define all those methods. Many extensions "must" only define `init()`.

I'd keep that `final` in there for the __construct and possibly the abstract `init()`. The other methods "must" only be defined if you're actually doing something with them. The fact that now you're adding these few lines everywhere proves @kevinpapst right.

```php
	public function install() {
		return true;
	}

	public function uninstall() {
		return true;
	}

	public function handleConfigureAction() {
	}
```

lib/Minz/Extension.php
```diff
@@ -50,35 +48,25 @@ public function __construct($meta_info) {
 	/**
 	 * Used when installing an extension (e.g. update the database scheme).
 	 *
-	 * It must be redefined by child classes.
-	 *
```
Frenzie, line 54:
```
	 * It must be redefined by child classes IFF (if and only if) the extension desires to perform some installation action.
	 *
```

Frenzie:
>  these changes are completely unnecessary.
Wasted time for many developers (you and me here and all the others involved everywhere else) without ANY benefit.

tl;dr Completely this. Not no benefit. NEGATIVE benefit.

aledeg:
Ok, so let's change!

lib/Minz/Extension.php
```diff
@@ -75,10 +69,8 @@ public function uninstall() {
 	/**
 	 * Call at the initialization of the extension (i.e. when the extension is
 	 * enabled by the extension manager).
-	 *
-	 * It must be redefined by child classes.
 	 */
-	public function init() {}
+	abstract public function init();
```
Frenzie, line 73: NB Assuming that an extension would be useless if it didn't define this, anyway. ;-)

aledeg:
@Alkarex We can merge that without side-effect since the scope has been reduced.

@kevinpapst Feel free to revert the changes I added earlier to your extensions.

Alkarex (approved):

Alkarex:
I was a bit away, but I am pleased to see that arguments have been expressed respectfully, so thanks for that, please keep it this way :-)
The outcome looks good and reasonable to me

PR #3334, opened 2021-01-04
Add autofocus on subscription page

aledeg:

- autofocusing on the URL input when landing on the subscription page

1. Open the subscription page
2. Check that the focus is on the feed URL input

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, there was no autofocusing on the subscription page. It feels
a bit off since most of the time, when you're landing on that page
you want to add a new feed.
Now, the focus is on the feed URL input to smooth the process of adding
a feed.

### DIFF
--- a/app/views/subscription/add.phtml
+++ b/app/views/subscription/add.phtml
@@ -27,7 +27,7 @@
 		<div class="form-group">
 			<label class="group-name" for="url_rss"><?= _t('sub.feed.url') ?></label>
 			<div class="group-controls">
-				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long"/>
+				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long" autofocus="autofocus" />
 			</div>
 		</div>
 


app/views/subscription/add.phtml
```diff
@@ -27,7 +27,7 @@
 		<div class="form-group">
 			<label class="group-name" for="url_rss"><?= _t('sub.feed.url') ?></label>
 			<div class="group-controls">
-				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long"/>
+				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long" autofocus/>
```
Frenzie: I assume we just want this one:
```suggestion
				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long" autofocus>
```

But otherwise it should be something like this:
```suggestion
				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long" autofocus="autofocus"/>
```
The combination's just weird. ;-)

aledeg: I've changed it

app/views/subscription/add.phtml
```diff
@@ -27,7 +27,7 @@
 		<div class="form-group">
 			<label class="group-name" for="url_rss"><?= _t('sub.feed.url') ?></label>
 			<div class="group-controls">
-				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long"/>
+				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long" autofocus="autofocus"/>
```
Alkarex: Very minor, but while we are at it, the common convention is to put a space before ` />`

```suggestion
				<input id="url_rss" name="url_rss" type="url" required="required" autocomplete="off" class="long" autofocus="autofocus" />
```

Alkarex (approved):

app/views/subscription/add.phtml
Frenzie: Only for compatibility with Netscape 4 or something along those lines. >_>

Alkarex: Indeed I believe the convention came from some parsing-related aspect, to make sure it worked in all cases even in older products :-)

aledeg: I've changed it

PR #3335, opened 2021-01-04
Add a deprecation log error for extensions

aledeg:

- Add error when loading extension which do not override mandatory methods

1. Browse the logs

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Extensions must override some parent methods. The rule is just a
guideline because only described in comments. The idea is to
deprecate that flexibility and add it to the code. To warn users
beforehand, we log error message regarding incomplete extensions.

See #3333

### DIFF
--- a/lib/Minz/ExtensionManager.php
+++ b/lib/Minz/ExtensionManager.php
@@ -172,6 +172,14 @@ class Minz_ExtensionManager {
 			                  '` is not an instance of `Minz_Extension`');
 			return null;
 		}
+		$reflector = new ReflectionClass($extension);
+		$className = $reflector->getName();
+		if ('Minz_Extension' === $reflector->getMethod('handleConfigureAction')->class ||
+			'Minz_Extension' === $reflector->getMethod('install')->class ||
+			'Minz_Extension' === $reflector->getMethod('init')->class ||
+			'Minz_Extension' === $reflector->getMethod('uninstall')->class) {
+			Minz_Log::error("The '{$className}' extension class definition is deprecated. It will continue to work with the current version but will break in the future. The '{$className}' extension class needs to override the 'handleConfigurationAction' method, the 'install' method, the 'init' method, and the 'uninstall' method to work properly.");
+		}
 
 		return $extension;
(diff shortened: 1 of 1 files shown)


lib/Minz/ExtensionManager.php
```diff
@@ -172,6 +172,14 @@ public static function load($info) {
 			                  '` is not an instance of `Minz_Extension`');
 			return null;
 		}
+		$reflector = new ReflectionClass($extension);
+		$className = $reflector->getName();
+		if ('Minz_Extension' === $reflector->getMethod('handleConfigureAction')->class ||
```
Alkarex, line 177: I have not double-checked: Do we have a guarantee from before that the extension we are loading is inheriting from Minz_Extension? If not, this code will raise an exception if one of the methods does not exist

Alkarex (approved):
Looks fine to me (minus my question / comment).
An alternative would be to deprecate the `Minz_Extension` base class, and make a new one with the new signature (I think this last option is more common for breaking changes in an API)

lib/Minz/ExtensionManager.php
aledeg: Extension must be instance of Minz_Extension to be recognized by the manager. Thus we can be sure that the base class is Minz_Extension. So we always have at least the Minz_extension methods available. If the definition of those methods are in Minz_Extension, that means it was never overriden.

aledeg:
We could do the alternate way but that means that we will allow 2 different ways of doing things. I think that it might not be necessary at the moment.

Alkarex:
As you prefer @aledeg , you are more into extensions than I am :-)
But does that mean that we cannot merge https://github.com/FreshRSS/FreshRSS/pull/3333 for now?

aledeg:
You could merge #3333 but it will break all extensions. We could fix all extensions we are managing beforehand though. It's preferable to merge this one first, fix all extensions we can, contact all extension maintainers we know of. We then merge #3333 in the `n+1` release.

Issue #2015, opened 2018-09-10
Improving user queries


marienfressinaud:
User queries are generally hard to understand. First, accessing user queries from the "admin" link is very confusing since you are welcomed with a placeholder and no explanations to create a query or even to start to understand what is a user query.

Then, you can't understand what's the purpose of "Add a query" before clicking on it (and even then, it's not totally understandable).

Finally, it's very confusing to click on the "delete" cross next to the user queries, refreshing the page, and see them again (you need to submit before but it's not clear at all!). Same feedback applies to sharing options. The "cancel" button is not very useful neither and we could imagine to drop it.

Alkarex:
One of the improvements I really want is a new UI for the search, to help with advanced search options. Maybe there could be some synergies with user queries.

Alkarex:
It would also be nice to offer more options in user queries, such as with an option to expose them by API.

zan73:
Modification of existing queries and multi-string searches (with OR/AND syntax in the same manner as major search engines) would be very welcome!

Alkarex:
@zan73 `OR` is already supported :-) `AND` is the default. Double quotes are also working, but only for search expressions like `intitle:"Hello World"` (I thought we had them working everywhere, but apparently not - I will have to look back into it, please open a new issue if there is not one already)

aledeg:
See #2216

PR #3346, opened 2021-01-08
Add user query manual sorting

aledeg:

- Add user query sorting by drag and drop
- Remove useless cancel action on user query configuration page

1. Bookmark some user queries
2. Display the user query configuration page
3. Drag and drop them to reorder them (don't forget to submit your changes to apply them)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

See #2015 and #2216

### DIFF
--- a/p/scripts/user.query.js
+++ b/p/scripts/user.query.js
@@ -1,0 +1,48 @@
+// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-3.0
+"use strict";
+/* jshint esversion:6, strict:global */
+
+function init_draggable() {
+	if (!window.context) {
+		if (window.console) {
+			console.log('FreshRSS user query waiting for JS…');
+		}
+		setTimeout(init_draggable, 50);
+		return;
+	}
+	
+	let source;
+	const configureQueries = document.querySelector('#configureQueries');
+
+	configureQueries.addEventListener('dragstart', event => {
+		source = event.target.closest('[draggable="true"]');
+		event.dataTransfer.setData('text/html', source.outerHTML);
+		event.dataTransfer.effectAllowed = 'move';
+	});
+	configureQueries.addEventListener('dragover', event => event.preventDefault());
+	configureQueries.addEventListener('dragleave', event => event.preventDefault());
+	configureQueries.addEventListener('drop', event => {
+		event.preventDefault();
+		event.stopPropagation();
+		const dropQuery = event.target.closest('[draggable="true"]');
+		if (null === dropQuery) {
+			source.remove();
+			configureQueries.querySelector('legend').insertAdjacentHTML('afterend', event.dataTransfer.getData('text/html'));
+		} else if (source !== dropQuery) {
+			source.remove();
+			dropQuery.insertAdjacentHTML('afterend', event.dataTransfer.getData('text/html'));
+		}
+	});
+
+	// This is needed to work around a Firefox bug → https://bugzilla.mozilla.org/show_bug.cgi?id=800050
+	configureQueries.addEventListener('focusin', event => {
+		event.target.closest('input[id^="queries_"][id$="_name"]').select();
+	});
+}
+
+if (document.readyState && document.readyState !== 'loading') {
+	init_draggable();
+} else if (document.addEventListener) {
+	document.addEventListener('DOMContentLoaded', event => init_draggable(), false);
+}
+// @license-end
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -275,6 +275,8 @@ class FreshRSS_configure_Controller extends Minz_ActionController {
 	 * checking if categories and feeds are still in use.
 	 */
 	public function queriesAction() {
+		Minz_View::appendScript(Minz_Url::display('/scripts/user.query.js?' . @filemtime(PUBLIC_PATH . '/scripts/user.query.js')));
+
 		$category_dao = FreshRSS_Factory::createCategoryDao();
 		$feed_dao = FreshRSS_Factory::createFeedDao();
 		$tag_dao = FreshRSS_Factory::createTagDao();
--- a/app/views/configure/queries.phtml
+++ b/app/views/configure/queries.phtml
@@ -77,7 +77,6 @@
 		<div class="form-group form-actions">
 			<div class="group-controls">
 				<button type="submit" class="btn btn-important"><?= _t('gen.action.submit') ?></button>
-				<button type="reset" class="btn"><?= _t('gen.action.cancel') ?></button>
 			</div>
 		</div>
 		<?php } else { ?>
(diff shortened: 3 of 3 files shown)


p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
+// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-3.0
+"use strict";
+/* jshint esversion:6, strict:global */
+
+function init_draggable() {
```
Alkarex, line 5: I would have put that in `extra.js` but for that page, I do not really mind

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+	});
+}
+
+if (document.readyState && document.readyState !== 'loading') {
+	init_draggable();
+} else if (document.addEventListener) {
+	document.addEventListener('DOMContentLoaded', function () {
+		init_draggable();
+	}, false);
+}
```
Frenzie: I'm not sure I quite understand what all this is for compared to just DOMContentLoaded?

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+	configureQueries.addEventListener('dragleave', event => {
+		event.preventDefault();
+	});
+	configureQueries.addEventListener('drop', event => {
+		event.preventDefault();
+		event.stopPropagation();
+		const dropQuery = event.target.closest('[draggable="true"]');
+		if (null === dropQuery) {
+			source.remove();
+			configureQueries.querySelector('legend').insertAdjacentHTML('afterend', event.dataTransfer.getData('text/html'));
```
Frenzie, line 30: How did I miss out on the introduction of insertAdjacentHTML/Element/Text. :-O Thanks so much!

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+		} else if (source !== dropQuery) {
+			source.remove();
+			dropQuery.insertAdjacentHTML('afterend', event.dataTransfer.getData('text/html'));
+		}
+	});
+
+	// This is needed to workaround a Firefox bug → https://bugzilla.mozilla.org/show_bug.cgi?id=800050
```
Frenzie:
```suggestion
	// This is needed to work around a Firefox bug → https://bugzilla.mozilla.org/show_bug.cgi?id=800050
```

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+			console.log('FreshRSS user query waiting for JS…');
+		}
+		setTimeout(init_draggable, 50);
+		return;
+	}
+	
+	let source;
+	const configureQueries = document.querySelector('#configureQueries');
+
+	configureQueries.addEventListener('dragstart', event => {
```
Alkarex, line 17: Not a big deal because we decided not to care about IE11 for new features, but worth noting that this is the first time we use the arrow function syntax https://caniuse.com/arrow-functions

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+		source = event.target.closest('[draggable="true"]');
+		event.dataTransfer.setData('text/html', source.outerHTML);
+		event.dataTransfer.effectAllowed = 'move';
+	});
+	configureQueries.addEventListener('dragover', event => {
+		event.preventDefault();
+	});
```
Alkarex: Minor: with the arrow function syntax, this can be simplified further to something like:

```suggestion
	configureQueries.addEventListener('dragover', event => event.preventDefault());
```

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+	// This is needed to workaround a Firefox bug → https://bugzilla.mozilla.org/show_bug.cgi?id=800050
+	configureQueries.addEventListener('focusin', event => {
+		event.target.closest('input[id^="queries_"][id$="_name"]').select();
+	});
+}
+
+if (document.readyState && document.readyState !== 'loading') {
+	init_draggable();
+} else if (document.addEventListener) {
+	document.addEventListener('DOMContentLoaded', function () {
```
Alkarex: If you keep the separate file and the arrow function syntax, you might want to use it here as well ;-)

p/scripts/user.query.js
Frenzie: One thing I like about the brackets (and why I don't like optionally leaving them out, at least in the traditional if/else type locations) is that you can just put in another line without any bother. Even if it's just a `console.log(event)`.

Just something to potentially keep in mind. Should be fine here.

Alkarex (approved):

p/scripts/user.query.js
```diff
@@ -0,0 +1,54 @@
...
+		}
+	});
+
+	// This is needed to workaround a Firefox bug → https://bugzilla.mozilla.org/show_bug.cgi?id=800050
+	configureQueries.addEventListener('focusin', event => {
+		event.target.closest('input[id^="queries_"][id$="_name"]').select();
+	});
+}
+
+if (document.readyState && document.readyState !== 'loading') {
```
Frenzie, line 43: @Alkarex I was going to say this doesn't work in IE11 either, but apparently it mostly does.
https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState

Alkarex: Sure it does, otherwise FreshRSS would not have worked at all with IE11 the past several years :-P

Alkarex:
Fine for me as it is. See some suggestions if you want @aledeg 
Another suggestion of (future) improvement: it would be to make it visually more apparent that drag & drop is possible (affordance), and maybe also a visual clue (line) of where it lands, as it is quite hard to move a query to the first position for instance

p/scripts/user.query.js
Frenzie: DOMContentLoaded works in IE9+, so as written here I don't think that's relevant. :-P

Alkarex:
@Frenzie Do you know any Unicode sign that resembles the following? To show the possibility of reordering
image

Alkarex:
Maybe:

```
☰
```

Frenzie:
Not with the arrows, but you have, among others:

𝍤
𝌇
☳
𝄚
𝍨

Alkarex:
Easy to build otherwise, e.g.:

```
△
☰
▽
```

p/scripts/user.query.js
aledeg: I just copied it from another file. It might be some legacy code!

Frenzie:
Maybe. The alignment in your example is terrible though. ^_^ I think they usually only show handles without arrows?

aledeg:
> Fine for me as it is. See some suggestions if you want @aledeg
> Another suggestion of (future) improvement: it would be to make it visually more apparent that drag & drop is possible (affordance), and maybe also a visual clue (line) of where it lands, as it is quite hard to move a query to the first position for instance

I've tried to add a visual clue without luck so far.
To show that the drag & drop is possible, the mouse cursor switch to a hand over the elements

Alkarex:
Ok, let's leave it to that for now then. Future UI improvements welcome :-)

PR #3347, opened 2021-01-09
Do not import feed causing DB error

### LINKED ISSUE #3339: [Feature Request] Remove duplicate feeds
Hey there, I searched around but wasn't sure if this has been suggested yet. I did a search but didn't quite find a match, so forgive me if this has already been suggested somewhere.

I noticed an odd quirk in FreshRSS where if you import multiple OPMLs that have overlapping feeds, they are imported into FreshRSS again, and it leads to multiple copies of the same feed.

As an example, I imported an Inoreader OPML to FreshRSS a few months back. Fast forward to today, where I exported a fresh OPML from Inoreader (which included some additional new feeds), and imported that into FreshRSS only to find that now FreshRSS is loaded with duplicate feeds that contain exactly the same content.

Is it possible to include an option to deduplicate these feeds?

Alkarex:
The DB error might be that the new feed tries to redirect to an already existing feed, in which case
#fix https://github.com/FreshRSS/FreshRSS/issues/3339

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -90,7 +90,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 			'name' => $title != '' ? $title : $feed->name(),
 			'website' => $feed->website(),
 			'description' => $feed->description(),
-			'lastUpdate' => time(),
+			'lastUpdate' => 0,
 			'httpAuth' => $feed->httpAuth(),
 			'attributes' => $feed->attributes(),
 		);
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -103,7 +103,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		$feed->_id($id);
 
 		// Ok, feed has been added in database. Now we have to refresh entries.
-		self::actualizeFeed($id, $url, false, null, true);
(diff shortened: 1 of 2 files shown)


app/Controllers/feedController.php
```diff
@@ -731,8 +740,9 @@ public function reloadAction() {
 		$entryDAO2 = FreshRSS_Factory::createEntryDao();
 
 		foreach ($entries as $entry) {
-			$entry->loadCompleteContent(true);
-			$entryDAO2->updateEntry($entry->toArray());
+			if ($entry->loadCompleteContent(true)) {
```
Alkarex, line 743: Minor unrelated error fix, sorry :-)

Alkarex: When the CSS Path for full content retrieval was empty, the *reload* function would cause errors

app/Controllers/feedController.php
```diff
@@ -325,6 +325,8 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 				continue;
 			}
 
+			$isNewFeed = $feed->lastUpdate() <= 0;
```
Alkarex, line 328: Works better than a manual parameter, especially when we import multiple feeds (e.g. OPML)

Alkarex:
Merging, to allow reporting user to test

PR #3353, opened 2021-01-09
Possiblity to autoinstall in Docker Compose

### LINKED ISSUE #2724: Docker: Environment variables for database instance
In the README.md, it is suggested to simply link an existing MySQL or PostgreSQL instance via `--link` command, however, when the container is stopped and started, it falls back to installation wizard and we have to enter the database credentials once again interactively.

This is not helpful in cases where automated container upgrades are set up (e.g. with watchtower). There should be environment variable support like most other web service containers provide accepting database host, database user, database password and database name environment variables.

### LINKED ISSUE #3105: REQ: allow specifying db info via env variables in docker-compose
It would be convenient if the docker-compose.yml file would automatically configure the freshrss docker image to connect to the postgres database so that the user would not have to enter any database connection info into the FreshRSS installation wizard.  The docker-compose file already lets us specify the database name, username, and password for the "freshrss_postgresql" container, so all the necessary information could be similarly injected into the "freshrss" container, letting docker users skip the database setup page of the wizard.

Example of another project that uses the docker-compose to configure the database settings in the app:
`https://git.tt-rss.org/fox/ttrss-docker-compose/src/master/docker-compose.yml`.  Notice how each container that requires database access has environment variables for the db username and password.  It should also have a variable for the database hostname in case the user wants to use a different database hostname.

### LINKED ISSUE #3349: Bootstrap FreshRSS entirely via environment variables
Copying most of: https://github.com/FreshRSS/FreshRSS/issues/3340#issuecomment-756963805

Since most of the data is in a DB, is there any way to pass in the following values via environment files:

```
Language
DB:
  Host
  Username
  Password
  DB
  Table Prefix
Default User:
  Name
  Auth
  Password
```

These are all asked during the initial setup. I am currently running a centralized database and would love to be able to pass these values in and not have to worry about which host my FreshRSS install lives on since its mostly running in docker with an orchestrator. 

I am able to enter these manually and everything works fine, but that is an annoyance when nomad schedules the docker container on a different host.

Let me know if this is possible, and thanks for the awesome work!

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3349

It is simply calling our existing CLI: do-install.php and create-user.php
https://github.com/FreshRSS/FreshRSS/tree/master/cli
(So all the API options are available)

FreshRSS will typically be ready a few seconds before the database, so introduce a tolerance when the database is not available / up (yet) by trying a few times to connect. Also useful to avoid service interruption when DB service is restarted.

Example:

```yml
version: "3"

services:
  freshrss-db:
    image: postgres:12-alpine
    container_name: freshrss-db
    hostname: freshrss-db
    restart: unless-stopped
    volumes:
      - db:/var/lib/postgresql/data
    environment:
      POSTGRES_USER: freshrss
      POSTGRES_PASSWORD: freshrss
      POSTGRES_DB: freshrss

  freshrss-app:
    image: freshrss/freshrss
    container_name: freshrss-app
    hostname: freshrss-app
    restart: unless-stopped
    ports:
      - "8080:80"
    depends_on:
      - freshrss-db
    volumes:
      - data:/var/www/FreshRSS/data
      - extensions:/var/www/FreshRSS/extensions
    environment:
      CRON_MIN: '*/20'
      FRESHRSS_ENV: development
      FRESHRSS_INSTALL: |-
        --api_enabled
        --base_url https://rss.example.net
        --db-base freshrss
        --db-host freshrss-db
        --db-password freshrss
        --db-type pgsql
        --db-user freshrss
        --default_user admin
        --language en
      FRESHRSS_USER: |-
        --api_password freshrss
        --email user@example.net
        --language en
        --password freshrss
        --user admin
      TZ: Europe/Paris

volumes:
  db:
  data:
  extensions:
```

### DIFF
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -2,12 +2,12 @@
 
 php -f ./cli/prepare.php >/dev/null
 
-find /etc/php*/ -name php.ini -exec sed -r -i "\\#^;?date.timezone#s#^.*#date.timezone = $TZ#" {} \;
-find /etc/php*/ -name php.ini -exec sed -r -i "\\#^;?post_max_size#s#^.*#post_max_size = 32M#" {} \;
-find /etc/php*/ -name php.ini -exec sed -r -i "\\#^;?upload_max_filesize#s#^.*#upload_max_filesize = 32M#" {} \;
+find /etc/php*/ -type f -name php.ini -exec sed -r -i "\\#^;?date.timezone#s#^.*#date.timezone = $TZ#" {} \;
+find /etc/php*/ -type f -name php.ini -exec sed -r -i "\\#^;?post_max_size#s#^.*#post_max_size = 32M#" {} \;
+find /etc/php*/ -type f -name php.ini -exec sed -r -i "\\#^;?upload_max_filesize#s#^.*#upload_max_filesize = 32M#" {} \;
 
 if [ -n "$LISTEN" ]; then
-	find /etc/apache2/ -name FreshRSS.Apache.conf -exec sed -r -i "\\#^Listen#s#^.*#Listen $LISTEN#" {} \;
+	find /etc/apache2/ -type f -name FreshRSS.Apache.conf -exec sed -r -i "\\#^Listen#s#^.*#Listen $LISTEN#" {} \;
 fi
 
 if [ -n "$CRON_MIN" ]; then
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -21,12 +21,13 @@ if [ -n "$CRON_MIN" ]; then
 fi
 
 if [ -n "$FRESHRSS_INSTALL" ]; then
+	# shellcheck disable=SC2046
 	php -f ./cli/do-install.php -- \
 		$(echo "$FRESHRSS_INSTALL" | sed -r 's/[\r\n]+/\n/g' | paste -s -) \
 		1>/tmp/out.txt 2>/tmp/err.txt
 	EXITCODE=$?
-	cat /tmp/out.txt | grep -v 'Remember to'
-	cat /tmp/err.txt | grep -v 'Please use' 1>&2
+	grep -v 'Remember to' /tmp/out.txt
+	grep -v 'Please use' /tmp/err.txt 1>&2
 
 	if [ $EXITCODE -eq 3 ]; then
 		echo 'ℹ️ FreshRSS already installed; no change performed.'
@@ -42,11 +43,12 @@ if [ -n "$FRESHRSS_INSTALL" ]; then
 fi
 
 if [ -n "$FRESHRSS_USER" ]; then
+	# shellcheck disable=SC2046
 	php -f ./cli/create-user.php -- \
 		$(echo "$FRESHRSS_USER" | sed -r 's/[\r\n]+/\n/g' | paste -s -) \
 		1>/tmp/out.txt 2>/tmp/err.txt
 	EXITCODE=$?
-	cat /tmp/out.txt | grep -v 'Remember to'
+	grep -v 'Remember to' /tmp/out.txt
 	cat /tmp/err.txt 1>&2
 
 	if [ $EXITCODE -eq 3 ]; then
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -2,9 +2,6 @@
 
 php -f ./cli/prepare.php >/dev/null
 
-chown -R :www-data .
-chmod -R g+r . && chmod -R g+w ./data/
-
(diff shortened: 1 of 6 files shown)


Frenzie (commented):
While it should be harmless here, it's probably better to avoid the shellcheck warning you're channeling with useless cat & probably also quotation marks around the $() subshell. (And also double check if it has something else to say besides what stands out to me at a glance.)

Alkarex:
I think I will look again at `cli/do-install.php` and `cli/create-user.php` to (optionally) be able to call them multiple times and update the configurations accordingly, but probably in another PR.

Alkarex:
@Frenzie Please give an example, I am not sure what you mean precisely :-)

Docker/entrypoint.sh
```diff
@@ -23,4 +20,50 @@ if [ -n "$CRON_MIN" ]; then
 	crontab -l | sed -r "\\#FreshRSS#s#^[^ ]+ #$CRON_MIN #" | crontab -
 fi
 
+if [ -n "$FRESHRSS_INSTALL" ]; then
+	php -f ./cli/do-install.php -- \
+		$(echo "$FRESHRSS_INSTALL" | sed -r 's/[\r\n]+/\n/g' | paste -s -) \
```
Frenzie, line 26:
```suggestion
		"$(echo "$FRESHRSS_INSTALL" | sed -r 's/[\r\n]+/\n/g' | paste -s -)" \
```

Docker/entrypoint.sh
```diff
@@ -23,4 +20,50 @@ if [ -n "$CRON_MIN" ]; then
...
+		1>/tmp/out.txt 2>/tmp/err.txt
+	EXITCODE=$?
+	cat /tmp/out.txt | grep -v 'Remember to'
+	cat /tmp/err.txt | grep -v 'Please use' 1>&2
```
Frenzie:
```suggestion
	< /tmp/out.txt grep -v 'Remember to'
	< /tmp/err.txt grep -v 'Please use' 1>&2
```

Docker/entrypoint.sh
```diff
@@ -23,4 +20,50 @@ if [ -n "$CRON_MIN" ]; then
...
+		echo '❌ FreshRSS error during installation!'
+		exit $EXITCODE
+	fi
+
+	rm -f /tmp/out.txt /tmp/err.txt
+fi
+
+if [ -n "$FRESHRSS_USER" ]; then
+	php -f ./cli/create-user.php -- \
+		$(echo "$FRESHRSS_USER" | sed -r 's/[\r\n]+/\n/g' | paste -s -) \
```
Frenzie, line 48:
```suggestion
		"$(echo "$FRESHRSS_USER" | sed -r 's/[\r\n]+/\n/g' | paste -s -)" \
```

Docker/entrypoint.sh
Alkarex: Hum, that does not work. We do have multiple options, not a single one ;-)
With the double quotes, it is transformed into a single parameter

Alkarex: What we receive is something like that:
```
--api_enabled --base_url https://rss.example.net --db-base freshrss --db-host freshrss-db --db-password freshrss --db-type pgsql --db-user freshrss --default_user admin --language en
```

Docker/entrypoint.sh
Alkarex: I was not familiar with this syntax

Docker/entrypoint.sh
Frenzie: Are we being POSIX compliant? Otherwise we could switch the shebang to /bin/bash to make it clear we want word splitting and that it's not an oversight. The other alternative is a `shellcheck disable=SC1234`.
```
read -ra args <<(echo "$FRESHRSS_INSTALL" | sed -r 's/[\r\n]+/\n/g' | paste -s -)
php "${args[@]}"
```

I double checked btw, and shellcheck didn't complain about anything worse.

Alkarex: I was trying to stick to POSIX, yes. At the very least, must work with the default shell in Debian and Alpine, which are the ones which will be executing that for now

Alkarex: So what do you suggest, @Frenzie ?

Frenzie: Either shellcheck disable so we know it's on purpose or just nothing. :-)

Docker/entrypoint.sh
Alkarex: For the record: https://github.com/koalaman/shellcheck/wiki/SC2002
I was familiar with a syntax like `wc < tmp.txt` but not the other way around like `< tmp.txt wc`
I have to admit that can better like the cat version for its consistency. For instance, we still need one further below.
I did not know UUOC was a thing, so I have learned something !
Changed; good now?

Frenzie (approved):

PR #3355, opened 2021-01-11
Add drag & drop marker on user query list

aledeg:

- Add drag & drop marker on user query list

1. re-order the user queries to see it in action

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, it was nearly impossible to know exactly where the dragged
item will land when dropping it.
Now, there is a visual marker to show the drop location.

An HR tag is inserted dynamically in the DOM. It's possible to style
it if needed.

### DIFF
--- a/p/scripts/user.query.js
+++ b/p/scripts/user.query.js
@@ -2,7 +2,7 @@
 "use strict";
 /* jshint esversion:6, strict:global */
 
-function init_draggable() {
+const init_draggable = function() {
 	if (!window.context) {
 		if (window.console) {
 			console.log('FreshRSS user query waiting for JS…');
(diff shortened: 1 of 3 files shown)


aledeg:
See #2015 and #2216

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
 		return;
 	}
 	
-	let source;
+	var source;
```
Alkarex: Any reason for `var` ? (just being curious)

Alkarex: P.S. This changes the scope to global

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
...
 	const configureQueries = document.querySelector('#configureQueries');
+	const addMarker = function(position, element) {
```
Alkarex: Personal preference: I like a space before the parenthesis for anonymous functions. But as you are using arrow function syntax  below, you might just as well use it here

p/scripts/user.query.js
Frenzie: Looks accidental to me.

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
...
+		const hr = configureQueries.querySelector('hr.drag-drop-marker');
+		if (null === hr) {
+			element.insertAdjacentHTML(position, '<hr class="drag-drop-marker"/>');
```
Alkarex:
```suggestion
			element.insertAdjacentHTML(position, '<hr class="drag-drop-marker" />');
```
Minor consistency :-)

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
...
+		}
+	}
+	const removeMarker = function() {
```
Alkarex: Same as above

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
...
 	configureQueries.addEventListener('dragstart', event => {
 		source = event.target.closest('[draggable="true"]');
 		event.dataTransfer.setData('text/html', source.outerHTML);
 		event.dataTransfer.effectAllowed = 'move';
 	});
-	configureQueries.addEventListener('dragover', event => event.preventDefault());
-	configureQueries.addEventListener('dragleave', event => event.preventDefault());
+	configureQueries.addEventListener('dragover', event => {
+		event.preventDefault();
+		if ('function' !== typeof event.target.closest) {
```
Alkarex: No need to change it, but testing simply testing `if (event.target.closest)` is probably enough

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
...
+		if (null === hr) {
+			element.insertAdjacentHTML(position, '<hr class="drag-drop-marker"/>');
+		}
+	}
+	const removeMarker = function() {
+		const hr = configureQueries.querySelector('hr.drag-drop-marker');
+		if (null !== hr) {
+			hr.remove();
+		}
+	}
```
Alkarex:
```suggestion
	};
```

Alkarex: A semicolon is also missing line 77 from previous PR

p/scripts/user.query.js
```diff
@@ -11,27 +11,63 @@ function init_draggable() {
...
 	
-	let source;
+	var source;
 	const configureQueries = document.querySelector('#configureQueries');
+	const addMarker = function(position, element) {
+		const hr = configureQueries.querySelector('hr.drag-drop-marker');
+		if (null === hr) {
+			element.insertAdjacentHTML(position, '<hr class="drag-drop-marker"/>');
+		}
+	}
```
Alkarex:
```suggestion
	};
```

p/scripts/user.query.js
Frenzie: PS That's why I set up jshint after all. ^_^

Alkarex: Indeed; and I probably have merged too fast...

Alkarex:
Looks good @aledeg :-)
Little bug with the Swage theme: moving to the first position does not seem to work

aledeg:
@Alkarex The swage theme works but in some weird ways. The reason behind that is the element detection that cannot find the legend DOM element because it does not cover the whole area. I chose this path to avoid mouse position detection but after all, it seems that I will need to :)

Alkarex:
@aledeg Don't embark in something overly complicated for such a small bug. Can't it be fixed by adding more space at the top?

aledeg:
@Alkarex It's no big deal, I already on it.
Even with more space, that won't work because that won't increase the legend box.

aledeg:
It's fixed!

p/scripts/user.query.js
```diff
@@ -13,32 +13,73 @@ function init_draggable() {
...
+		} else {
 			dropQuery.insertAdjacentHTML('afterend', event.dataTransfer.getData('text/html'));
 		}
+		source.remove();
+		removeMarker();
 	});
 
 	// This is needed to work around a Firefox bug → https://bugzilla.mozilla.org/show_bug.cgi?id=800050
 	configureQueries.addEventListener('focusin', event => {
 		event.target.closest('input[id^="queries_"][id$="_name"]').select();
```
Alkarex: I can generate the following error (Firefox 84 / Ubuntu 20.10)
image

Alkarex: When hitting the validate button

aledeg: Even though I have no idea how you get those, I add a fix.

aledeg: My fix is not correct

aledeg: This time, it should be working as intended.

aledeg:
@Alkarex Could you please review this PR? I have a follow-up PR based on this one. Thank you!

Alkarex:
@aledeg Did you see my previous comment? (some are broken due to the forced-pushes) 
image

aledeg:
@Alkarex I did not see your comment. What do you suggest? just testing with:
```js
if (event.target.closest)
```
?

Frenzie:
That wouldn't work as expected if event.target doesn't exist for some reason.

Alkarex:
To be safe, something like: `if (event.target && event.target.closest) {`

I was a bit unsure, but looking at the spec https://developer.mozilla.org/docs/Web/API/EventTarget , https://developer.mozilla.org/docs/Web/API/FocusEvent suggests that `target` is supposed to be always defined (otherwise there would be no event), but the type of `target` can vary, but `focusin` is supposed to always be an `Element`, so the whole test is probably not needed. Just pick your favourite approach then :-)

Alkarex (approved):

aledeg:
@Alkarex I've change the test. It should be good now!

PR #3361, opened 2021-01-11
Fix images in enclosures without MIME type

### LINKED ISSUE #3358: no pictures in display of the feed of 'Le Monde' newspaper
Hi,

The feed 
https://www.lemonde.fr/rss/une.xml
is not fully displayed: only the title and text do appear, but the pictures don't. 

With my previous rss reader, it was working properly.

Thanks for all the good work !

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3358
Assume enclosures without a mime type, without a length, but with either a width or a height are images

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -405,23 +405,23 @@ class FreshRSS_Feed extends Minz_Model {
 						$elinks[$elink] = true;
 						$mime = strtolower($enclosure->get_type());
 						$medium = strtolower($enclosure->get_medium());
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Maybe fall back to file extension depending on potential side effects?

Alkarex:
I thought about file extensions, but then it should be a list for each type, and with a priority for mime over file type, etc. But it might become needed indeed...

PR #3342, opened 2021-01-07
two new hooks

cn-tools:
For my new extension [Copy2Clipboard](https://github.com/cn-tools/cntools_FreshRssExtensions) i need 2 new hooks.

Changes proposed in this pull request:

- Add hook to add some navigation options: Now the developers of extensions have the chance to add additional navigation points. In my case, i add an COPY button.
- Add hook to add some jsonVars: In my case i need some translated texts in the javascript section. With this hook i can add this texts into the `jsonVar` script-tag and in my javascript code i can use this text with `context.extensions.copy2clipboard.msg_good_more`.

Pull request checklist:

- [ ] clear commit messages
- [X] code manually tested
- [ ] unit tests written (optional if too hard)
- [x] documentation updated - EN only!

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/master/docs/en/developers/04_Pull_requests.md).

PR #3362, opened 2021-01-12
fix layout changes since new hook `nav_menu`

cn-tools:
With #3342 layout was changed ... Sorry!
Now layout changing only, if there are `nav_menu` hooks active

### DIFF
--- a/app/layout/nav_menu.phtml
+++ b/app/layout/nav_menu.phtml
@@ -176,7 +176,7 @@
 	</div>
 
 	<?php $nav_menu_hooks = Minz_ExtensionManager::callHook('nav_menu'); ?>
-	<?php if ($nav_menu_hooks != ''){ ?>
+	<?php if ($nav_menu_hooks != '') { ?>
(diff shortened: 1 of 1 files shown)


cn-tools:
i think the problem with travis is in the file [_cli.php](https://github.com/FreshRSS/FreshRSS/blob/master/cli/_cli.php) in line 20

original:
`function fail($message, $exitCode=1) {`
spaces around '=' not available

is this correct?

Frenzie:
That's correct, but not related to this PR.

app/layout/nav_menu.phtml
```diff
@@ -175,9 +175,12 @@
 		</a>
 	</div>
 
+	<?php $nav_menu_hooks = Minz_ExtensionManager::callHook('nav_menu'); ?>
+	<?php if ($nav_menu_hooks != ''){ ?>
```
Alkarex:
```suggestion
	<?php if ($nav_menu_hooks != '') { ?>
```
Minor syntax convention

PR #3366, opened 2021-01-13
Add a query configuration page

### LINKED ISSUE #2015: Improving user queries

aledeg:
Closes #2015 

- Add user query configuration page; it contains the filter link, the delete action and a way to modify the user query itself
- Add position auto-validation after drag&drop

1. Edit, move, delete user queries

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, there was no way to modify a user query. Thus you need to
create a new one and delete the old one afterward.
Now, every user query can be modified if needed. They have their
own configuration page on which it can be modified or deleted.


app/i18n/fr/conf.php
```diff
@@ -65,12 +65,22 @@
 		'_' => 'Filtres utilisateurs',
 		'deprecated' => 'Ce filtre n’est plus valide. La catégorie ou le flux concerné a été supprimé.',
 		'display' => 'Afficher les résultats du filtre',
-		'filter' => 'Filtres appliqués :',
+		'filter' => array(
+			'_' => 'Filtres appliqués :',
+			'categories' => 'Display by category',	// TODO - Translation
```
Alkarex: Un oubli ? :-)

app/Controllers/configureController.php
```diff
@@ -304,9 +305,92 @@ public function queriesAction() {
...
+		$query = new FreshRSS_UserQuery(FreshRSS_Context::$user_conf->queries[$id], $feed_dao, $category_dao, $tag_dao);
+		$this->view->query = $query;
+		$this->view->queryId = $id;
+		$this->view->categories = $category_dao->listCategories(false);
+		$this->view->feeds = $feed_dao->listFeeds();
+		$this->view->tags = $tag_dao->listTags();
+
+		if (Minz_Request::isPost()) {
+			$params = array_filter(Minz_Request::param('query', []));
+			if ($params['search']) {
```
Alkarex:
```suggestion
			if (!empty($params['search'])) {
```
Same below

app/views/configure/queries.phtml
```diff
@@ -14,30 +14,15 @@
 			</label>
 
 			<div class="group-controls">
+				<?= $query->getName() ?>
```
Alkarex, line 17: I have double-checked that this looks properly escaped upstream

Alkarex:
Looks very good @aledeg 👍🏻 
My only comment is simply about the wording for "By keyword", which seems a bit strange when the request is more complicated, e.g.:

image

Alkarex:
P.S.: It would be great to have a similar configuration page for labels ;-)

aledeg:
@Alkarex I fixed the PR.

aledeg:
@Alkarex What do you mean by that?

> It would be great to have a similar configuration page for labels

We do have a label configuration page in the subscription menu!

Alkarex:
Regarding the labels, I would like to add more options for each label (e.g. archiving, automatic actions...) so it would be nice to have a configuration page per label, similarly to what you have done for filters, and also similar to categories and feeds.

Alkarex (approved):

Alkarex:
I still think that the `'search' => 'By keyword',` is strange when it contains a more complicated search query and/or does not contain keywords ( such as `date:P7d author:Alice` ). Maybe something like *By search expression* or something like that, but I have not given it much thoughts.

Likewise, `'order' => 'By date',` (which appears as *by date*) is also a bit strange because it sounds like a filter like the ones just above it, but is in fact some ordering. So maybe *Ordering*, *Sort*, or similar.

Besides that, looks fine :-)

aledeg:
I've changed the wording but I am not convinced :/

Alkarex:
Let's get a third pair of eyes; @Frenzie ? :-)

image

Frenzie:
I think it's calmer to look at and easier to scan if you get rid of all those "by"s. Something like:

## Filter (applied)
* Expression
* State
* Sort by date

aledeg:
@Alkarex @Frenzie I've fixed the PR, this includes:
- changing the wording.
- moving modifications in their own commit.

Alkarex:
Hum, I have not taken the time to check whether it is this PR or the previous related one, but we now have this warning when after manipulating a user query (e.g. its name)
image

Frenzie:
A `chaîne` always makes me think "link" instead of "string".

aledeg:
@Alkarex It looks like a cache problem. I had that error as well. Change few things and revert them. Now it's gone :/

PR #3367, opened 2021-01-13
CURLOPT parameters per feed

### LINKED ISSUE #3271: Per feed proxy configuration
I've seen #675
Is there a way to make just single feed to use proxy or two or more feeds to use separate proxy configuration each?

printfuck:
Closes #3271

 - proxy settings per feed
 - cookie settings per feed

1. set a proxy or cookie
2. check if the result changes

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


app/i18n/de/sub.php
```diff
@@ -35,6 +35,10 @@
 		'clear_cache' => 'Nicht cachen (für defekte Feeds)',
 		'css_help' => 'Ruft gekürzte RSS-Feeds ab (Achtung, benötigt mehr Zeit!)',
 		'css_path' => 'Pfad zur CSS-Datei des Artikels auf der Original-Webseite',
+		'proxy' => 'Verwende einen Proxy, um den Feed abzuholen', 
+		'proxy_help' => 'Wähle ein Protokoll (z.B. SOCKS5) und einen Proxy mit Port (z.B. 127.0.0.1:1080)',
+		'css_cookie' => 'Verwende Cookies beim Herunterladen des Feed-Inhalts mit CSS-Filtern',
+		'css_cookie_help' => 'Beispiel: foo=bar; gdpr_consent=true; cookie=value',
```
aledeg: To add keys, use the Makefile. It will save you some work. You've updated only 3 languages, the others needs to be updated as well.
On the CLI, you just need to run for instance:
```
make i18n-add-key key='sub.feed.proxy' value='Set a proxy for fetching this feed'
```
It will work only if the key is not set. So you'll need to remove your changes in the i18n files then run the command once for each key. Then you'll add your translations again.
The command will take care of ordering the keys, adding the keys in every files, add the todo flag on languages you cannot translate.

printfuck: Thanks. Never seen docker used in a Makefile before 😄

aledeg: It's very handy :)

Frenzie:
Quite!

Alkarex:
Could you please try to resolve the conflicts @printfuck ? Let us know if you face any difficulty

printfuck:
I fixed all the linting hints, but it's still waiting for the Docker builds. Is there something I missed?

Alkarex:
Don't worry about the Docker builds. It looks good. I will test 👍🏻

app/Models/Entry.php
```diff
@@ -376,7 +376,15 @@ public static function getContentByParsing($url, $path, $attributes = array(), $
 			CURLOPT_FOLLOWLOCATION => true,
 			CURLOPT_ENCODING => '',	//Enable all encodings
 		]);
+
 		curl_setopt_array($ch, FreshRSS_Context::$system_conf->curl_options);
+
+		if (isset($attributes['curl_params']) && is_array($attributes['curl_params'])) {
+			foreach ($attributes['curl_params'] as $co => $v) {
```
Alkarex: Could you check whether `curl_setopt_array()` could be used there instead of a loop?

app/i18n/fr/sub.php
```diff
@@ -33,6 +33,8 @@
 			'username' => 'Identifiant HTTP',
 		),
 		'clear_cache' => 'Toujours vider le cache',
+		'css_cookie' => 'Use Cookies when fetching the article content',	// TODO - Translation
+		'css_cookie_help' => 'Example: foo=bar; gdpr_consent=true; cookie=value',	// TODO - Translation
```
```suggestion
		'css_cookie' => 'Utiliser des cookies pour la réception du contenu des articles',
		'css_cookie_help' => 'Exemple : <kbd>foo=bar; gdpr_consent=true; cookie=value</kbd>',
```

app/i18n/fr/sub.php
```diff
@@ -61,6 +63,8 @@
 			'main_stream' => 'Afficher dans le flux principal',
 			'normal' => 'Afficher dans sa catégorie',
 		),
+		'proxy' => 'Set a proxy for fetching this feed',	// TODO - Translation
+		'proxy_help' => 'Select a protocol (e.g: SOCKS5) and enter the proxy address (e.g: 127.0.0.1:1080)',	// TODO - Translation
```
```suggestion
		'proxy' => 'Utiliser un proxy pour télécharger ce flux',
		'proxy_help' => 'Sélectionner un protocole (ex. : SOCKS5) et entrer l’adresse du proxy (ex. : <kbd>127.0.0.1:1080</kbd>)',
```

Alkarex:
@printfuck I have made a few changes; please have a look https://github.com/FreshRSS/FreshRSS/pull/3367/commits/7795d480c0c6ae981c8228c25e61bea43522e5e0

Alkarex:
@printfuck I have made a few changes; please have a look

printfuck:
This looks good to me.

Alkarex:
Ok, let's merge and please test a bit more if you can

Alkarex:
I am wandering whether we should allow normal users to use this feature, or only admins.
Furthermore, I forgot that we will have to deal with cache differences between the users, just like we do when providing a username/password to retrieve a feed.

printfuck:
I agree. Unless 
(1) the proxy servers are known to the users (e.g. published in a wiki for the given instance of freshrss) and are only available to the instance,
(2) or the users are running the proxy servers themselves, limiting access to the specific freshrss instance,
this feature will only ever be useful for admins.

I don't understand the cache thing as of now. After reading the code multiple times, I didn't get why there is need for a cache.

PR #3368, opened 2021-01-14
Refactor requirement check in install

aledeg:

- Add requirement check template during install

1. Make a new FreshRSS install

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

I've extracted the template in the check requirement to
ensure that future modifications will be easier.

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -362,108 +362,39 @@ function printStep0() {
 <?php
 }
 
+function printStep1Template($key, $value, $messageParams = []) {
+	$message = _t("install.check.{$key}.{$value}", ...$messageParams);
+	if ('ok' === $value) {
+		?><p class="alert alert-success"><span class="alert-head"><?= _t('gen.short.ok') ?></span> <?= $message ?></p><?php
+	} else {
+		?><p class="alert alert-error"><span class="alert-head"><?= _t('gen.short.damn') ?></span> <?= $message ?></p><?php
+	}
+}
+
 // @todo refactor this view with the check_install action
 function printStep1() {
 	$res = checkRequirements();
 ?>
 	<noscript><p class="alert alert-warn"><span class="alert-head"><?= _t('gen.short.attention') ?></span> <?= _t('install.javascript_is_better') ?></p></noscript>
 
-	<?php if ($res['php'] == 'ok') { ?>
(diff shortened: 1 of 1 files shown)


app/install.php
```diff
@@ -362,108 +362,39 @@ function printStep0() {
 <?php
 }
 
+function printStep1Template($key, $value, $messageParams = []) {
+	$message = _t("install.check.{$key}.{$value}", ...$messageParams);
```
Alkarex, line 366: Could we generalise the translation to have only one string instead of many?

aledeg: We could but I am not sure it will work. I have to think about it.

Alkarex: While it would be nice to remove the many translation strings we have for that, it can also wait for later

aledeg: Actually, I think the whole install process needs a bit of love. I'll look into it when I have time and feel like it.

Alkarex: Should we then merge this PR as is while waiting for future improvements?

aledeg: I think it's better if you do not want to see this one collect dust

aledeg:
I've messed up the translation in that one. I fixed it in #3373

app/install.php
```diff
@@ -362,108 +362,39 @@ function printStep0() {
...
-
-	<?php if ($res['http_referer'] == 'ok') { ?>
-	<p class="alert alert-success"><span class="alert-head"><?= _t('gen.short.ok') ?></span> <?= _t('install.check.http_referer.ok') ?></p>
-	<?php } else { ?>
-	<p class="alert alert-error"><span class="alert-head"><?= _t('gen.short.damn') ?></span> <?= _t('install.check.http_referer.nok') ?></p>
-	<?php } ?>
+	<?php
+	printStep1Template('php', $res['php'], [PHP_VERSION, FRESHRSS_MIN_PHP_VERSION]);
+	printStep1Template('pdo', $res['pdo']);
+	printStep1Template('curl', $res['curl'], [$version['version']]);
```
Alkarex, line 383: @aledeg Ups, I did not spot it, but the variable `$version` is undefined

aledeg: Good catch, fixed in #3427

PR #3369, opened 2021-01-14
Add constant for PHP requirements

aledeg:

- Add constant for PHP requirements
- Remove PHP version checks for obsolete versions

1.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -358,9 +358,6 @@ class FreshRSS_configure_Controller extends Minz_ActionController {
 			Minz_Error::error(403);
 		}
 
-		$can_enable_email_validation = version_compare(PHP_VERSION, '5.5') >= 0;
-		$this->view->can_enable_email_validation = $can_enable_email_validation;
-
 		if (Minz_Request::isPost()) {
 			$limits = FreshRSS_Context::$system_conf->limits;
 			$limits['max_registrations'] = Minz_Request::param('max-registrations', 1);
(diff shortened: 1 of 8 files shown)


app/Controllers/configureController.php
```diff
@@ -358,9 +358,6 @@ public function systemAction() {
 			Minz_Error::error(403);
 		}
 
-		$can_enable_email_validation = version_compare(PHP_VERSION, '5.5') >= 0;
-		$this->view->can_enable_email_validation = $can_enable_email_validation;
```
Frenzie, line 362: Not currently relevant of course, but I think we were still going for 5.3+ when this one was added (and as such it's not equivalent with min version since it's just the one feature).

aledeg: I know but now, the minimum requirement is 5.6. So there is no need to keep a test that will be always true.

lib/lib_rss.php
```diff
@@ -416,7 +416,7 @@ function check_install_php() {
 	$pdo_pgsql = extension_loaded('pdo_pgsql');
 	$pdo_sqlite = extension_loaded('pdo_sqlite');
 	return array(
-		'php' => version_compare(PHP_VERSION, '5.5.0') >= 0,
+		'php' => version_compare(PHP_VERSION, FRESHRSS_MIN_PHP_VERSION) >= 0,
```
Frenzie, line 419: Would this check be obsolete too? Hard to tell on phone.

aledeg: I think the check has to be kept but the version it was tested against was. I think that was forgotten when changing the requirements. That's the reason why I want to have only have one constant value. Next time we need to change it, we won't face the same issue.

PR #3372, opened 2021-01-15
Change integration configuration page

aledeg:

- Add integration ordering by drag and drop

1.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

I've refactored the user query code to be able to use it also on the
integration configuration page. Now, items can be re-organized as in
the user query list.


p/scripts/integration.js
```diff
@@ -0,0 +1,54 @@
...
+        form.submit();
+    });
+};
+
+if (document.readyState && document.readyState !== 'loading') {
+	init_integration();
+} else if (document.addEventListener) {
+	document.addEventListener('DOMContentLoaded', event => init_integration(), false);
+}
+// @license-end
```
Alkarex: Minor: please new line

p/scripts/extra.js
```diff
@@ -239,6 +199,85 @@ function init_configuration_alert() {
...
+		}
+	};
+	const removeMarker = () => {
+		const hr = draggableList.querySelector('hr.drag-drop-marker');
+		if (null !== hr) {
+			hr.remove();
+		}
+	};
+
+	draggableList.addEventListener('dragstart', event => {
```
Alkarex: Not needed for `integration.js`, but in `main.js` and `extra.js`, could you please keep using an IE11-compatible syntax for a few more months? So no arrow functions here. Otherwise, as this code is not an essential feature, it could also be fine to move it to e.g. `integration.js`.

p/scripts/integration.js
aledeg: done!

p/scripts/extra.js
aledeg: @Alkarex I've extracted the draggable feature in its own file. Later we could merge it back to `extra.js`

PR #3373, opened 2021-01-15
Add username hint for permissions during install

### LINKED ISSUE #2758: Provide user hint for error message "HTTP server must have rights to write into"
I got this eror during installation check: (archlinux arm in a raspberry pi 4b)
```
...
...
Ok! You have the fileinfo library.

Blast! Check permissions on ./data directory. HTTP server must have rights to write into

Blast! Check permissions on ./data/cache directory. HTTP server must have rights to write into

Blast! Check permissions on ./data/users directory. HTTP server must have rights to write into

Blast! Check permissions on ./data/favicons directory. HTTP server must have rights to write into

Ok! Your HTTP REFERER is known and corresponds to your server.

Please fix errors before skipping to the next step.
```

And permissions are already correctly set:
```
[lifeng@alarmpi data]$ ls -alt
total 52
drwxr-xr-x 13 root http 4096 Jan  8 21:16 ..
drwxrwxr-x  9 root http 4096 Jan  8 21:16 .
-rw-rw-r--  1 root http  137 Jan  8 21:16 .gitignore
-rw-rw-r--  1 root http  174 Jan  8 21:16 .htaccess
drwxrwxr-x  4 root http 4096 Jan  8 21:16 PubSubHubbub
drwxrwxr-x  2 root http 4096 Jan  8 21:16 cache
-rw-rw-r--  1 root http    0 Jan  8 21:16 do-install.txt
drwxrwxr-x  2 root http 4096 Jan  8 21:16 extensions-data
drwxrwxr-x  2 root http 4096 Jan  8 21:16 favicons
drwxrwxr-x  2 root http 4096 Jan  8 21:16 fever
-rw-rw-r--  1 root http  309 Jan  8 21:16 index.html
drwxrwxr-x  2 root http 4096 Jan  8 21:16 tokens
-rw-rw-r--  1 root http 2771 Jan  8 21:16 tos.example.html
drwxrwxr-x  3 root http 4096 Jan  8 21:16 users

```
nginx is running using default http user
```
[lifeng@alarmpi data]$ ps aux|grep nginx
root      2296  0.0  0.0   7468   732 ?        Ss   00:07   0:00 nginx: master process /usr/bin/nginx -g pid /run/nginx.pid; error_log stderr;
http      2297  0.0  0.0   8672  3748 ?        S    00:07   0:00 nginx: worker process
lifeng    2465  0.0  0.0   2528   548 pts/1    S+   00:24   0:00 grep nginx

```

the site file:
```shell
server {
        listen 8080;
        listen 443 ssl;

        # HTTPS configuration
        ssl_certificate /etc/nginx/ssl/server.crt;
        ssl_certificate_key /etc/nginx/ssl/server.key;

        # your servers URL(s)
        server_name localhost;

        # the folder p of your FreshRSS installation
        root /usr/share/nginx/html/p/;

        index index.php index.html index.htm;

        # nginx log files
        access_log /var/log/nginx/rss.access.log;
        error_log /var/log/nginx/rss.error.log;

        # php files handling
        # this regex is mandatory because of the API
        location ~ ^.+?\.php(/.*)?$ {
                # fastcgi_pass unix:/var/run/php/php7.0-fpm.sock;
                fastcgi_pass unix:/run/php-fpm/php-fpm.sock;
                fastcgi_split_path_info ^(.+\.php)(/.*)$;
                # By default, the variable PATH_INFO is not set under PHP-FPM
                # But FreshRSS API greader.php need it. If you have a Bad Request error, double check this var!
                fastcgi_param PATH_INFO $fastcgi_path_info;
                include fastcgi_params;
                fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
        }

        location / {
                try_files $uri $uri/ index.php;
        }
}
```
php module:
```
[lifeng@alarmpi FreshRSS]$ php -m
[PHP Modules]
Core
ctype
curl
date
dom
fileinfo
filter
gmp
hash
intl
json
libxml
mbstring
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
Phar
posix
readline
Reflection
session
SimpleXML
SPL
sqlite3
standard
tokenizer
xml
xmlreader
xmlwriter
zip
zlib

[Zend Modules]

```

If i do it in the cli way, it seems fine:
```
[lifeng@alarmpi FreshRSS]$ sudo -u http sh -c 'cli/prepare.php'
��� Remember to re-apply the appropriate access rights, such as:
	sudo chown -R :www-data . && sudo chmod -R g+r . && sudo chmod -R g+w ./data/
Result: success
```

The log dir doesn't provide any thing useful, and I guess freshrss won't write anything to it if it thinks it's not writable

aledeg:
Closes #2758

- Add username hint for permission checks during install

1. Make a new install with the wrong permissions on folders

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

The username is retrieved by 2 different means to support different
configurations and architectures. If there is no way to find the
username, the "unknown" string is displayed instead.

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -363,17 +363,35 @@ function printStep0() {
 }
 
 function printStep1Template($key, $value, $messageParams = []) {
-	$message = _t("install.check.{$key}.{$value}", ...$messageParams);
 	if ('ok' === $value) {
+		$message = _t("install.check.{$key}.ok", ...$messageParams);
 		?><p class="alert alert-success"><span class="alert-head"><?= _t('gen.short.ok') ?></span> <?= $message ?></p><?php
 	} else {
+		$message = _t("install.check.{$key}.nok", ...$messageParams);
 		?><p class="alert alert-error"><span class="alert-head"><?= _t('gen.short.damn') ?></span> <?= $message ?></p><?php
 	}
+}
+
+function getProcessUsername() {
+	if (function_exists('posix_getpwuid') && function_exists('posix_geteuid')) {
+		$processUser = posix_getpwuid(posix_geteuid());
+		return $processUser['name'];
+	}
+
+	if (function_exists('exec')) {
+		exec('whoami', $output);
+		if (!empty($output[0])) {
+			return $output[0];
+		}
+	}
+
+	return _t('install.check.unknown_process_username');
 }
 
 // @todo refactor this view with the check_install action
 function printStep1() {
 	$res = checkRequirements();
+	$processUsername = getProcessUsername();
 ?>
 	<noscript><p class="alert alert-warn"><span class="alert-head"><?= _t('gen.short.attention') ?></span> <?= _t('install.javascript_is_better') ?></p></noscript>
 
@@ -388,11 +406,11 @@ function printStep1() {
 	printStep1Template('xml', $res['xml']);
 	printStep1Template('mbstring', $res['mbstring']);
 	printStep1Template('fileinfo', $res['fileinfo']);
-	printStep1Template('data', $res['data'], [DATA_PATH]);
-	printStep1Template('cache', $res['cache'], [CACHE_PATH]);
-	printStep1Template('tmp', $res['tmp'], [TMP_PATH]);
-	printStep1Template('users', $res['users'], [USERS_PATH]);
-	printStep1Template('favicons', $res['favicons'], [DATA_PATH . '/favicons']);
+	printStep1Template('data', $res['data'], [DATA_PATH, $processUsername]);
+	printStep1Template('cache', $res['cache'], [CACHE_PATH, $processUsername]);
+	printStep1Template('tmp', $res['tmp'], [TMP_PATH, $processUsername]);
+	printStep1Template('users', $res['users'], [USERS_PATH, $processUsername]);
+	printStep1Template('favicons', $res['favicons'], [DATA_PATH . '/favicons', $processUsername]);
 	printStep1Template('http_referer', $res['http_referer']);
 	?>
 
--- a/app/i18n/cz/install.php
+++ b/app/i18n/cz/install.php
@@ -92,6 +92,7 @@ return array(
 			'nok' => 'Zkontrolujte oprávnění adresáře <em>%s</em>. HTTP server musí mít do tohoto adresáře práva zápisu.',
 			'ok' => 'Permissions on the temp directory are good.',	// TODO - Translation
 		),
+		'unknown_process_username' => 'unknown',	// TODO - Translation
 		'users' => array(
 			'nok' => 'Zkontrolujte oprávnění adresáře <em>%s</em>. HTTP server musí mít do tohoto adresáře práva zápisu.',
 			'ok' => 'Oprávnění adresáře users jsou v pořádku.',
--- a/app/i18n/de/install.php
+++ b/app/i18n/de/install.php
@@ -92,6 +92,7 @@ return array(
 			'nok' => 'Überprüfen Sie die Berechtigungen des Verzeichnisses <em>%s</em>. Der HTTP-Server muss Schreibrechte besitzen.',
 			'ok' => 'Die Berechtigungen des Temp Verzeichnisses sind in Ordnung.',
 		),
+		'unknown_process_username' => 'unknown',	// TODO - Translation
 		'users' => array(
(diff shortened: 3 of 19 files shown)


aledeg:
@Frenzie I've updated an existing value for a key. It is already translated in many languages. Do you know how we could process to indicate that we need translations again?

Frenzie:
@aledeg You should probably just delete it using the CLI (best done through `make`), and then re-add the key as if it were new.

app/install.php
```diff
@@ -362,9 +362,24 @@ function printStep0() {
 <?php
 }
 
+function getProcessUsername() {
+	if (function_exists('posix_getpwuid') && function_exists('posix_geteuid')) {
+		$processUser = posix_getpwuid(posix_geteuid());
+		return $processUser['name'];
+	}
+
+	exec('whoami', $output);
```
Alkarex: This function might be disabled, e.g. on shared hosting, like the previous ones, via https://php.net/ini.core#ini.disable-functions
Please check that this does not crash

app/install.php
```diff
@@ -362,9 +362,24 @@ function printStep0() {
...
+	if (!empty($output)) {
```
Alkarex, line 372:
```suggestion
	if ($output != '') {
```
Minor: `empty()` is especially useful on structures / arrays; for scalar variables that are already existing, it is slightly overkill

```suggestion
	if (!empty($output[0])) {
```

app/install.php
aledeg: I've made a test with a dummy function and it does nothing.

app/install.php
aledeg: Your suggestion will crash if `$output` is empty. On the other hand, `$output` will have some data at its first index if not empty.

app/install.php
Alkarex: When `exec` is put in the list of disabled functions in `php.ini` ?

app/install.php
Alkarex: `empty($output[0])` should be safe even if `$output` is null / empty / not array.
This is one of the main roles for `empty()`.
Test:

```php
<?php
$output = null;
echo empty($output[0]) ? 'empty' : 'not empty'; //no crash
```

app/install.php
aledeg: I haven't thought about that! I'll look into it

Alkarex: I have just tested and it solely produces a warning:

```
PHP Warning:  exec() has been disabled for security reasons in Standard input code on line 2
```

But you need to declare `$output` prior to using it

app/install.php
```diff
@@ -362,9 +362,24 @@ function printStep0() {
 <?php
 }
 
+function getProcessUsername() {
+	if (function_exists('posix_getpwuid') && function_exists('posix_geteuid')) {
+		$processUser = posix_getpwuid(posix_geteuid());
+		return $processUser['name'];
+	}
+
```
Alkarex, line 380:
```suggestion

$output = null;
```

app/install.php
aledeg: I've added a test to see whether or not the function exist prior to call it

app/install.php
aledeg: I think there is no need for that with the test I've added to check if the function exists or not

app/install.php
```diff
@@ -362,9 +362,26 @@ function printStep0() {
...
+	if (function_exists('exec')) {
+		exec('whoami', $output);
+		if (!empty($output)) {
```
```suggestion
		if (!empty($output[0])) {
```

My previous suggestion + comment were erased by a forced push, so here they are again, just to finish the conversation.

Does not matter in practice here, but I would still argue that the correct test is `!empty($output[0])`

Three examples:

```php
$output = null;
empty($output[0]); //return true - what we want

$output = [ 1 => 'Test' ];
empty($output); //return false - not what we want
empty($output[0]); //return true - what we want

$output = [ '' ];
empty($output); //return false - not what we want
empty($output[0]); //return true - what we want
```

app/install.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3373/files#r563200885

app/install.php
aledeg: fixed

PR #3377, opened 2021-01-17
Fix form login if user does not exist

Alkarex:
Small bug from https://github.com/FreshRSS/FreshRSS/pull/3070 , leading to an exception due to not being able to log the login error is the user does not exist at all.

Minor: Also add some very primitive mitigation of timing attacks (to find out whether a user exists or not, although I have not checked whether this might be guessed through other means) - before, if the user did not exist, the response was always measurably faster; now it is harder to tell due to the noise


Alkarex:
Merging a bit fast, as this is a regression

PR #3379, opened 2021-01-17
Updated German Translation

rom-1:
completed german translation, 100% as of now


app/i18n/de/conf.php
```diff
@@ -67,20 +67,20 @@
...
-			'search' => 'Expression',	// TODO - Translation
-			'state' => 'State',	// TODO - Translation
-			'tags' => 'Display by tag',	// TODO - Translation
-			'type' => 'Type',	// TODO - Translation
+			'categories' => 'Nach Kategorie anzeigen',
+			'feeds' => 'Nach Feed anzeigen',
+			'order' => 'Nach Datum sortieren',
+			'search' => 'Suchbegriff',
+			'state' => 'Eigenschaft',
+			'tags' => 'Nach Tag anzeigen',
```
Frenzie: That's not confusing? :-)

aledeg (approved):

Frenzie (approved):

aledeg:
@rom-1 Thank you :+1:

app/i18n/de/conf.php
rom-1: in my humble opinion, the whole user query interface is confusing as it is right now, but this might just be me. :P 

I guess you are referring to the expression "Tag" and "Datum". It is, but the user has to get this from context. On all other pages, you are referring to "Tags" instead of Bezeichnungen ("labels"), so I continued to use this expression. using "Tags" is fine in german I guess, it's simply taken from english as we lack a better word.

aledeg: @rom-1 Is the new interface still confusing? I worked on that to make it more understandable. So if it is, I missed my goal. Could you explain why you think it is confusing?

rom-1: @aledeg Oh, no, it is definitely better than before, you improved it! But I guess there is still some room for improvement. ;)

I think my main issue is, that I personally do not see any use case for myself using user queries at all. The idea of user queries seem to be, to have a specific search term, for example "iphone" and "camera", which someone **regularly** (e.g. on a **daily** basis) needs to search for in his feeds. I do not see this as a requirement on my end at all. ;-)

Putting my personal issue aside, just a few suggestions to improve user queries:
- Add a Menu Item for "Edit User Queries" in the Main Menu (#dropdown-query) next to "Bookmark current query" OR put an [edit icon](https://fontawesome.com/v4.7.0/icon/pencil) next to every user query in this dropdown menu
- in the user queries editor: the hyperlink "filter" on the top right below the headline should be a more prominent real **button**, maybe with a text like "run", "show" or "go", or maybe it can stay "filter" if it becomes a real button. It should be placed at the bottom, next to save/cancel.
- in the user queries editor: the "type" box should be a multi-select box. A user might want to search for "iphone" on 2 or 3 feeds at the same time.
- the word "state" and "type" are not fitting, but I lack better words. Maybe leave "state" as it is. But "type" feels plain wrong. Even "nothing" (empty) would be better than "type", as the content of the box is quite self explaining.
- it would be great to have some kind of syntax help link next to every input box, in which the user can input complex user queries. This means on the main page next to the search box, and in the user query editor next to the "expression" input box. Maybe put a small [question mark icon](https://fontawesome.com/v4.7.0/icon/question-circle) there, with a link to: https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html --> "with the search field" (actually, there should be an extra page "query syntax" in documentation).

aledeg: Thank you for your insight. It might be worth a new issue.
I am afraid that if we keep it as a comment, it will be lost!

PR #3381, opened 2021-01-17
Curlopt fixes

### LINKED ISSUE #3378: minor CURLOPT problem
I was testing the CURLOPT merge and found a really dumb flaw. I really should've seen it earlier.

https://github.com/FreshRSS/FreshRSS/blob/ee175dd6169a016fc898fac62d046e22c205dec0/lib/lib_rss.php#L190-L195

line 192 should obviously be:
```
$curl_options[$co] = $v; 
```

printfuck:
Closes #3378

Pull request checklist:

- [ ] clear commit messages
- [x ] code manually tested
- [ ] documentation updated

Everything else should be OK.


lib/lib_rss.php
```diff
@@ -189,7 +189,7 @@ function customSimplePie($attributes = array()) {
 	}
 	if (!empty($attributes['curl_params']) && is_array($attributes['curl_params'])) {
 		foreach ($attributes['curl_params'] as $co => $v) {
-			curl_setopt($ch, $co, $v);
+			$curl_options[$co] = $v;
```
Alkarex, line 192: As a future improvement, could you please check whether a version without loop and using e.g. `array_merge()` could do the job?

printfuck: I already did, it should work.

Alkarex: We need to double check that the array keys are not modified in a wrong way (I am afraid they would be, with `array_merge`)

PR #3383, opened 2021-01-17
Add a theme extension class

aledeg:
Closes #3345

- Add a theme extension class
- Transform themes to extensions
-

1.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


aledeg:
@Alkarex @Frenzie I've started the work on the theme extensions. I'll push all my modification here keeping the history neat. I'll keep one commit per item discussed in #3345.
Let me know what are your thoughts about my work.
At the moment, I've created only the abstract class.

aledeg:
@Alkarex @marienfressinaud @Frenzie I have something usable. At the moment, the only thing that changes is that themes are extensions. Everything should be working as before. The only adjustment is to activate the theme you were using before.
I've added some notice to the user to indicate why it behaves this way.

Could you please test that out and let me know what you think?
Thank you

aledeg:
There is a lot of files involved but it's mostly renaming the themes files. So don't worry, this PR is not that big :)

Alkarex:
I will test tomorrow

app/Models/Themes.php
```diff
@@ -1,124 +1,223 @@
...
+
+		// Load user default theme
+		$defaultTheme = self::$defaultTheme;
+		if (empty($metadata)) {
+			$metadata = self::extractMetadata(new SplFileInfo(THEME_PATH . "/{$current_user}/{$defaultTheme}"), $current_user);
+		}
+
+		// Load system default theme
+		if (empty($metadata)) {
+			$metadata = self::extractMetadata(new SplFileInfo(THEME_PATH . "/{$defaultTheme}"), null);
```
Alkarex:
```
PHP Fatal error:  Uncaught TypeError: FreshRSS_Themes::extractMetadata():
Argument #2 ($current_user) must be of type string, null given,
called in /var/www/FreshRSS/app/Models/Themes.php on line 165
and defined in /var/www/FreshRSS/app/Models/Themes.php:111
Stack trace:
#0 /var/www/FreshRSS/app/Models/Themes.php(165): FreshRSS_Themes::extractMetadata()
#1 /var/www/FreshRSS/app/FreshRSS.php(104): FreshRSS_Themes::load()
#2 /var/www/FreshRSS/app/FreshRSS.php(142): FreshRSS::loadStylesAndScripts()
#3 /var/www/FreshRSS/app/layout/layout.phtml(1): FreshRSS::preLayout()
#4 /var/www/FreshRSS/lib/Minz/View.php(88): include('...')
#5 /var/www/FreshRSS/lib/Minz/View.php(101): Minz_View->includeFile()
#6 /var/www/FreshRSS/lib/Minz/View.php(68): Minz_View->buildLayout()
#7 /var/www/FreshRSS/lib/Minz/Dispatcher.php(54): Minz_View->build()
#8 /var/www/FreshRSS/lib/Minz/FrontController.php(81): Minz_Dispatcher->run()
#9 /var/www/FreshRSS/p/i/index.php(70): Minz_FrontController->run()
#10 {main}
  thrown in /var/www/FreshRSS/app/Models/Themes.php on line 111
```

lib/Minz/ThemeExtension.php
```diff
@@ -0,0 +1,102 @@
...
+
+	public function init() {
+	}
+
+	public function install() {
+		$current_user = Minz_Session::param('currentUser');
+		$themes_path = THEME_PATH . "/{$current_user}";
+
+		// Create the directory for the user themes
+		if (!file_exists($themes_path) && false === mkdir($themes_path)) {
```
Alkarex: What about prior to login?

Alkarex: Is that writing in `./p/themes/.../` ? This is not supposed to be a writeable location for the Web server. And it does not persist (updates, Docker...). Could we try to find another approach?

lib/Minz/ThemeExtension.php
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		// Create the theme directory
+		$theme_path = "{$themes_path}/{$this->getName()}";
+		if (!file_exists($theme_path) && false === mkdir($theme_path)) {
+			return "Cannot create {$theme_path} directory.";
+		}
+
+		// Copy the content of the extension theme to the user theme
+		$filepath = join_path($this->getPath(), 'static');
```
Alkarex: Is a full copy of the theme necessary?

lib/Minz/ThemeExtension.php
aledeg: The user is empty and the theme is copied without the user part

lib/Minz/ThemeExtension.php
aledeg: Probably not. But for that, we need to include a way of knowing which files to include.
There's nothing like that at the moment.

Alkarex: And if some kind of user data is necessary, it would be nice to put it under `./data/users/xxx/...`

app/Models/Themes.php
aledeg: :/ I've never experienced that. Under what conditions do you have this error?

lib/Minz/ThemeExtension.php
aledeg: I am not there yet :)

Alkarex:
@aledeg Could you write a few words about the use-case for copying a theme for each user instead of using it as-is? Hopefully only some optional user-specific parameters would be needed

aledeg:
Sure. I've copied the file in the public area to be able to use them without a lot of modification.
If the files were to sit in the extension repository, we will need to rely on the `ext.php?XXXX` to link css and images. That means a php call for each css file and image.

Alkarex:
@aledeg If the only reason for the copy is to serve the files, I would better like to go the `ext.php?xxx` route, with proper HTTP caching

aledeg:
Oh! I remember the other reason. It's because the theme is loaded before everything else. If we load them with `ext.php?xxx` we might have some weird behavior if the order is not the one intended. For instance if the customCSS is loaded before the theme, it won't work.

Either we copy the files in the theme as it is always linked before the extension or we need to figure out a way of adding a priority on the extension. Maybe by introducing a `theme` type.

Alkarex:
A `theme` type sounds like a good idea, and not only for that specific reason

aledeg:
Ok, I'll change things around. I was so proud of my copying mechanism :'(

aledeg:
> A `theme` type sounds like a good idea, and not only for that specific reason

What other reasons do you have in mind?

app/Models/Themes.php
aledeg: I've found the error!

aledeg: It's fixed

Alkarex: Tested in PHP8

aledeg:
@Alkarex I've changed things around. It's working well.
I only have one concern, it's for customized url calls from the css files (loader for instance). I have no idea how I can make those calls when they are relative to the file but I cannot reach them directly.

The only idea I have, and it feels like it's not a good one, is to preprocess the css files to track the use of url and change those to the real link.
Other ideas are welcomed.

aledeg:
@Alkarex @Frenzie Do you have some ideas regarding the problem I've mentioned earlier? I am stuck and I would like to move forward.

Alkarex:
@aledeg Quick thought (I can come back to it during the week-end):

Use a a convention that is both valid and rare, like search & replace `././` by a full path)

```css
url('././relative-path')
```

And make sure to have proper HTTP caching (I can give an example) like we do for some other PHP calls

aledeg:
I have a new idea on how to handle the things I've mentioned in previous messages.
I am thinking to revert my changes to have an install process of the extension.
My idea is to introduce extensions that are both `system` and `user`. The `system` part will allow to install the files in the filesystem and the `user` part will allow every user to activate it or not.
I will need to find a way to handle the two behaviors in the configuration but it seems easier to work this way.

To make this working, I'll need #3626 

@Alkarex @Frenzie Let me know what you think about that idea.

Frenzie:
That sounds sensible to me.

PR #3397, opened 2021-01-25
Add comprehensive user configuration in extensions

aledeg:

- Add comprehensive user configuration in extensions
-
-

1. Modify an extension code to use the new methods
2. Save the configuration
3. Check that the user configuration file is modified with the extension configuration

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

Before, the extension configuration was handled by its author. There
was discrepancies between extensions on how the configuration was
stored.
Now, we could rely on a single way of storing configuration. This won't
invalidate how the extensions are storing their configuration but will
allow authors to focus on what is important.

### DIFF
--- a/lib/Minz/Extension.php
+++ b/lib/Minz/Extension.php
@@ -11,6 +11,7 @@ abstract class Minz_Extension {
 	private $description;
 	private $version;
 	private $type;
+	private $config_key = 'extensions';
 
 	public static $authorized_types = array(
 		'system',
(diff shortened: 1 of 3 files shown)


lib/Minz/Extension.php
```diff
@@ -195,4 +195,38 @@ public function registerTranslates() {
 	public function registerHook($hook_name, $hook_function) {
 		Minz_ExtensionManager::addHook($hook_name, $hook_function, $this);
 	}
+
+	/**
+	 * @return bool
+	 */
+	private function isUserConfigurationEnabled() {
+		if (!class_exists('FreshRSS_Context', false)) {
```
Alkarex, line 204: What is the use-case for this line?

lib/Minz/Extension.php
```diff
@@ -195,4 +195,38 @@ public function registerTranslates() {
...
+			return [];
+		}
+		return FreshRSS_Context::$user_conf->{$this->getName()};
+	}
+
+	public function setUserConfiguration(array $configuration) {
+		if (!$this->isUserConfigurationEnabled()) {
+			return;
+		}
+		FreshRSS_Context::$user_conf->{$this->getName()} = $configuration;
```
Alkarex: Should it maybe be one lever further down, into a sub-section `->extensions` ?

aledeg: It could be. I found that it was easier this way because there is no need to manipulate the extension content. By that I mean that I cannot to something like:
```php
FreshRSS_Context::$user_conf->extension->{$this->getName()} = $configuration;
```
nor
```php
FreshRSS_Context::$user_conf->extension[{$this->getName()}] = $configuration;
```

I need to get the content of the extension first do the modifications then save.

lib/Minz/Extension.php
aledeg: Just to be sure!

lib/Minz/Extension.php
Alkarex:
```php
FreshRSS_Context::$user_conf->extensions[$this->getName()] = ...;
```
should work

aledeg: I haven't try this one. I'll keep you posted.

aledeg: It does not work :/

aledeg: To make it work, I need to do something like:
```php
		$extensions = FreshRSS_Context::$user_conf->extensions;
		$extensions[$this->getName()] = $configuration;
		FreshRSS_Context::$user_conf->extensions = $extensions;
```

aledeg: For the get method, it works directly though. I will need to do something similar for the remove method. Do you think this is better?

Alkarex: Arg, this is because of the indirection caused by `Minz_Configuration`. Maybe making `$user_conf->extensions` return an https://php.net/class.arrayaccess (object looking like an array) instead of a copy-on-write array (as now) could work

aledeg: @Alkarex I don't mind adding the extensions layer. I think it is a better way to organize things even if the code is less elegant. I made a first draft real quick to validate the idea. Now I can refine it :)

PR #3407, opened 2021-01-30
Better error message on failure

C0rn3j:
Fixes typo and makes the error message better at the same time.

'Check logs files'  is not very user friendly, so now the message suggests to check the systemd logs or /var/log if it's a more weird setup where the units are nonexistent.

image

---

Not a 100% with git, can the commit history be squashed to just the most recent commit so I don't litter the history with my catching up to upstream and old commits even if I already made a MR?


lib/Minz/FrontController.php
```diff
@@ -106,7 +106,7 @@ public function run () {
 	*/
 	private function killApp ($txt = '') {
 		if ($txt == '') {
-			$txt = 'See logs files';
+			$txt = 'See log files for the service "journalctl -xeu freshrss" or check in /var/log if the service is not present';
```
Frenzie: I see no reason for the discrepancy. ;-)
```suggestion
			$txt = 'See log files for the service "journalctl -xeu freshrss" or check in /var/log if the service is not present.';
```

p/i/index.php
```diff
@@ -80,7 +80,7 @@
 		// page to the users (especially non administrators).
 		echo '### Fatal error! ###<br />', "\n";
 		Minz_Log::error($error);
-		echo 'See log files for the service "journalctl -xeu freshrss" or check in /var/log if the service is not present.';
+		echo 'See log files for the service "journalctl -xeu freshrss" or check in /var/log if the service is not present';
```
Frenzie: Seems the wrong way to resolve it to me (also nice timing, lol)

C0rn3j: image

It looks a bit weird with the dot to me, but I do not care either way, I'll add it back to both and mention php-fpm and webserver also while I'm at it

C0rn3j: Added dots and made the messages better

aledeg:
@C0rn3j Thank you for the PR!
I like what you've done. I have some concern though. Their are not with what you've done but your PR revealed some underlying problem. So maybe we could try to tackle that at the same time :)

I feel that we should have only one error message instead of 2. I think that in the future, if we need to modify it, we will forgot there is the other one to modify as well. So I suggest that you extract that in a function.
While you're at it, could you improve the markup as well?
I mean that `### Application problem ###` could be a header like `<h1>Application problem</h1>` and the details you've added could be in a `PRE` tag. This way, we do not need to rely on `BR` tags to show the error.

@Alkarex @Frenzie what do you guys think?

C0rn3j:
@aledeg 
I agree that this should be in one place, but I am not familiar with the codebase(nor that much with PHP).

Where would be a good place for the function?

aledeg:
@C0rn3j You could add a function in the `lib_rss.php` file. That would be a good start. We can move it around later.

Also, instead of writing your string the way you did, you could use the [heredoc string syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc).
That would be easier to maintain because you do not have to escape quotes
```php
$message=<<<MSG
<h1>{$errorTitle}</h1>
<pre>
See log files for the feed update service "journalctl -xeu freshrss".
If you do not use a service to update the feeds, run the script manually "/usr/sbin/php /usr/share/webapps/freshrss/app/actualize_script.php".
If you get no hints from there, see logs for your webserver i.e. "journalctl -xeu nginx".
If you are using php-fpm, also check logs for it "journalctl -xeu php-fpm".
</pre>
<b>Note</b>: Change paths and service names as necessary depending on where and what things are installed. Additional logs may be in "/var/log".
MSG;
```

Here is an example with proper HTML markup:
Screenshot_2021-01-30 Screenshot

C0rn3j:
Here's my attempt.

Without passing optional error
image
Passed optional error (h2)
image

There's a useless `<h2></h2>` tag if the optional error is missing but I guess that's fine.

Frenzie:
@aledeg 
> what do you guys think?

Sounds sensible in principle but I'm not very familiar with this part of the code.

lib/Minz/FrontController.php
```diff
@@ -106,9 +106,10 @@ public function run () {
 	*/
 	private function killApp ($txt = '') {
 		if ($txt == '') {
-			$txt = 'See logs files';
+			exit (errorMessage ("Application problem") );
```
aledeg: Here, you do not need the `if` part anymore since you handle everything in the `errorMessage` function. You can now simplify the method by writing only the following:
```php
private function killApp ($txt = '') {
	exit(errorMessage('Application problem', $txt);
}
```

lib/lib_rss.php
```diff
@@ -578,3 +578,18 @@ function validateShortcutList($shortcuts) {
 	}
 	return $shortcuts_ok;
 }
+
+function errorMessage($errorTitle, $error = '') {
```
aledeg, line 582: As the `$error` variable is optional, it will be better to handle that case to avoid the empty `h2` tag.
I would add something like:
```php
if ('' !== $error) {
	$error = "<h2>{$error}</h2>";
}
```
before the `$message` definition.
Then in the `$message` definition, I would replace `<h2>{$error}<h2>` by `{$error}`.

lib/lib_rss.php
```diff
@@ -578,3 +578,18 @@ function validateShortcutList($shortcuts) {
...
+	$message=<<<MSG
```
aledeg: As you only return a string value, I would omit the use of a variable by using `return` directly:
```php
return <<<MSG
...
...
MSG;
```

lib/lib_rss.php
```diff
@@ -578,3 +578,18 @@ function validateShortcutList($shortcuts) {
...
+	<h2>{$error}<h2>
+	<pre>
+	See log files for the feed update service "journalctl -xeu freshrss".
+	If you do not use a service to update the feeds, run the script manually "/usr/sbin/php /usr/share/webapps/freshrss/app/actualize_script.php".
+	If you get no hints from there, see logs for your webserver i.e. "journalctl -xeu nginx".
+	If you are using php-fpm, also check logs for it "journalctl -xeu php-fpm".
+	</pre>
+	<b>Note</b>: Change paths and service names as necessary depending on where and what things are installed. Additional logs may be in "/var/log".
+	MSG;
+	echo($message);
```
aledeg: FIY, `echo` is not a function, it's a language construct. So you should use it without parentheses.
```php
echo $message;
```
It's really handy because `echo` takes any values without the need to concatenate them.
```php
echo $var1, 'string1', 1234, $var2, ...;
```

p/i/index.php
```diff
@@ -76,11 +76,8 @@
 	}
 
 	if ($error) {
-		// TODO this should be definitely improved to display a nicer error
-		// page to the users (especially non administrators).
-		echo '### Fatal error! ###<br />', "\n";
 		Minz_Log::error($error);
-		echo 'See logs files.';
+		errorMessage ("Fatal error");
```
aledeg: There is no variable in that string so you could switch to single quotes `'`

aledeg (changes requested):
@C0rn3j There are few minor changes to add before we could merge that. Thank you!

lib/lib_rss.php
C0rn3j: Didn't know about the concat bit!

lib/lib_rss.php
C0rn3j: Huh, I didn't know return would do the equivalent of an echo?

aledeg (approved):
LGTM

lib/Minz/FrontController.php
```diff
@@ -105,10 +105,7 @@ public function run () {
 	* Permet d'arrêter le programme en urgence
 	*/
 	private function killApp ($txt = '') {
-		if ($txt == '') {
-			$txt = 'See logs files';
-		}
-		exit ('### Application problem ###<br />'."\n".$txt);
+		exit (errorMessage ('Application problem', $txt) );
```
Alkarex:
```suggestion
		exit (errorMessage('Application problem', $txt));
```
Minor formatting

lib/lib_rss.php
```diff
@@ -578,3 +578,21 @@ function validateShortcutList($shortcuts) {
...
 	return $shortcuts_ok;
 }
+
+function errorMessage($errorTitle, $error = '') {
+	// Prevent empty <h2> tags by checking if error isn't empty first
+	if ('' !== $error) {
+		$error = "<h2>{$error}</h2>";
+	}
+	return <<<MSG
+	<h1>{$errorTitle}</h1>
```
Alkarex, line 590:
```suggestion
	$errorTitle = htmlspecialchars($errorTitle, ENT_NOQUOTES, 'UTF-8');
	return <<<MSG
	<h1>{$errorTitle}</h1>
```

lib/lib_rss.php
```diff
@@ -578,3 +578,21 @@ function validateShortcutList($shortcuts) {
 	}
 	return $shortcuts_ok;
 }
+
+function errorMessage($errorTitle, $error = '') {
+	// Prevent empty <h2> tags by checking if error isn't empty first
+	if ('' !== $error) {
+		$error = "<h2>{$error}</h2>";
```
```suggestion
	if ('' !== $error) {
		$error = htmlspecialchars($error, ENT_NOQUOTES, 'UTF-8');
		$error = "<h2>{$error}</h2>";
```

lib/lib_rss.php
```diff
@@ -578,3 +578,21 @@ function validateShortcutList($shortcuts) {
...
+	}
+	return <<<MSG
+	<h1>{$errorTitle}</h1>
+	{$error}
+	<pre>
```
Alkarex: This is making a lot of assumptions about the user's configuration, OS, namings, and I am afraid it might often-times be more confusing than helpful. I will make a few suggestions

lib/lib_rss.php
```diff
@@ -578,3 +578,21 @@ function validateShortcutList($shortcuts) {
...
+	See log files for the feed update service "journalctl -xeu freshrss".
+	If you do not use a service to update the feeds, run the script manually "/usr/sbin/php /usr/share/webapps/freshrss/app/actualize_script.php".
+	If you get no hints from there, see logs for your webserver i.e. "journalctl -xeu nginx".
+	If you are using php-fpm, also check logs for it "journalctl -xeu php-fpm".
+	</pre>
+	<b>Note</b>: Change paths and service names as necessary depending on where and what things are installed. Additional logs may be in "/var/log".
```
```suggestion
	<h2>Common problems</h2>
	<p>A typical problem leading to this message is wrong file permissions in the <kbd>./FreshRSS/data/</kbd> folder so make sure the Web server can write there and in sub-directories.</p>
	<h2>Common locations for additional logs</h2>
	<p><strong>N.B.:</strong> Adapt names and paths according to your local setup.</p>	
	<ul>
	<li>If using Docker: <kbd>docker logs -f freshrss</kbd></li>
	<li>To check Web server logs on a Linux system using systemd; <kbd>journalctl -xeu apache2</kbd>
	and if you are using php-fpm: <kbd>journalctl -xeu php-fpm</kbd></li>
	<li>Otherwise, Web server logs are typically located in <kbd>/var/log/apache2/</kbd> or similar</li>
	<li>System logs may also contain relevant information in <kbd>/var/log/syslog</kbd></li>
	</ul>
	<p>More logs can be generated by enabling <code>'environment' => 'development',</code> in <kbd>./FreshRSS/data/config.php</kbd></p>
```

A user would not see this message when using cron, so I not not think the cron/service is very relevant here.
We could link to a troubleshooting page with better guidance.

p/i/index.php
```diff
@@ -76,11 +76,8 @@
 	}
 
 	if ($error) {
-		// TODO this should be definitely improved to display a nicer error
-		// page to the users (especially non administrators).
-		echo '### Fatal error! ###<br />', "\n";
 		Minz_Log::error($error);
-		echo 'See logs files.';
+		errorMessage ('Fatal error');
```
Alkarex:
```suggestion
		errorMessage('Fatal error');
```

lib/Minz/FrontController.php
Alkarex: We should not call from Minz a function that does not exist there. I will make a suggestion

lib/lib_rss.php
Frenzie: Something like samp or code would be more appropriate than kbd, wouldn't it?

p/i/index.php
C0rn3j: https://github.com/FreshRSS/FreshRSS/blob/master/lib/Minz/FrontController.php#L70
I did this because it looked like the used code style in the files (this example is even committed by you), should I then change all functions that do this in the three files I'm editing while I'm at it?

lib/lib_rss.php
C0rn3j: >so I do not think the cron/service is very relevant here

It is relevant. In my case, I've had an issue where I didn't have pdo enabled/installed as I was downgrading from PHP 8 to 7, and the only way to figure it out was to look at the actualize_script logs which pointed it out.

```
/usr/sbin/php /usr/share/webapps/freshrss/app/actualize_script.php
FreshRSS starting feeds actualization at 2021-01-30T21:08:25+00:00
FreshRSS install failed requirements:
• pdo-pgsql
• ok
```

And the PHP 8 issue itself was logged there also.

p/i/index.php
Alkarex: Yes please, no space before parenthesis in our current convention

lib/lib_rss.php
C0rn3j: I used your comment but replaced the kbd tags with code tags and kept the actualize script note in

lib/Minz/FrontController.php
C0rn3j: I think I dealt with everything else, but not sure how to fix this problem.

Alkarex: Let me have a look

Alkarex:
I have made a few changes. Please double-check.
See also https://github.com/FreshRSS/FreshRSS/pull/3410

Alkarex (approved):

lib/lib_rss.php
```diff
@@ -578,3 +578,30 @@ function validateShortcutList($shortcuts) {
...
+	<ul>
+	<li>If using Docker: <code>docker logs -f freshrss</code></li>
+	<li>To check Web server logs on a Linux system using systemd: <code>journalctl -xeu apache2</code>
+	and if you are using php-fpm: <code>journalctl -xeu php-fpm</code></li>
+	<li>Otherwise, Web server logs are typically located in <code>/var/log/apache2/</code> or similar</li>
+	<li>System logs may also contain relevant information in <code>/var/log/syslog</code>, or if using systemd: <code>sudo journalctl -xe</code></li>
+	</ul>
+	<p>More logs can be generated by enabling <code>'environment' => 'development',</code> in <code>./FreshRSS/data/config.php</code></p>
+	<p>Running the feed update script (with the same user and PHP version as your Web server) might provide other hints, e.g.:
+		<code>sudo -u www-data /usr/bin/php /var/www/FreshRSS/app/actualize_script.php</code></p>
```
C0rn3j:
```suggestion
		<code>sudo -u www-data /usr/bin/php ./FreshRSS/app/actualize_script.php</code></p>
```

I'd either change to /var/www everywhere, or keep the ./ notation as in the other examples

C0rn3j:
Looks all good to me!

I've skimmed #3410 but didn't get the link between the two issues othen than it touches the actualize dependency message.

C0rn3j:
Thanks to everyone involved, you make it easy to get changes in!

lib/lib_rss.php
```diff
@@ -578,3 +578,30 @@ function validateShortcutList($shortcuts) {
...
+	<li>If using Docker: <code>docker logs -f freshrss</code></li>
+	<li>To check Web server logs on a Linux system using systemd: <code>journalctl -xeu apache2</code>
+	and if you are using php-fpm: <code>journalctl -xeu php-fpm</code></li>
+	<li>Otherwise, Web server logs are typically located in <code>/var/log/apache2/</code> or similar</li>
+	<li>System logs may also contain relevant information in <code>/var/log/syslog</code>, or if using systemd: <code>sudo journalctl -xe</code></li>
+	</ul>
+	<p>More logs can be generated by enabling <code>'environment' => 'development',</code> in <code>./FreshRSS/data/config.php</code></p>
+	<p>Running the feed update script (with the same user and PHP version as your Web server) might provide other hints, e.g.:
+		<code>sudo -u www-data /usr/bin/php ./FreshRSS/app/actualize_script.php</code></p>
+	MSG;
```
Alkarex, line 606: Invalid syntax fixed in https://github.com/FreshRSS/FreshRSS/pull/3426
https://php.net/manual/language.types.string#language.types.string.syntax.heredoc

PR #3413, opened 2021-02-02
Fix extension file search

aledeg:

- Fix extension file search
-
-

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, it was possible to retrieve only the files from extensions. Thus
making core extension files unreachable.
Now, the selected file is search through all extensions folders.

### DIFF
--- a/p/ext.php
+++ b/p/ext.php
@@ -13,6 +13,28 @@ const SUPPORTED_TYPES = [
 	'svg' => 'image/svg+xml',
 ];
 
+/**
+ * @return string
+ */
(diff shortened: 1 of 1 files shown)


aledeg:
This modification is needed for the theme extensions.

Alkarex:
If possible, it would be nice to avoid searching on disk, but that should do it for now

aledeg:
How do you think we should do that?

Alkarex:
For instance by including the type of extension as a prefix of the extension name

PR #3416, opened 2021-02-03
Fix: handle very big feed

Kiblyn11:
I was trying to add a very very huge XML feed but it kept failing because of out of memory exception (using latest docker with 4gb ram).
I figured file could be chunk where it was ooming (only in cleanMd5 and parse functions).
I tested personally for one week and it's working fine.

- In functions which are ooming, read chunk of data instead of whole file and process
- Did that by using php memory
- It is the recommended way to parse big files with xml_parse (https://www.php.net/manual/fr/function.xml-parse.php)

- Unfortunately I can't give away the big feed as it's private one, I can say that it's 14megabytes unminified and 149959 lines long.
1. Try to add very big feed
2. Should not error

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated


lib/SimplePie/SimplePie.php
```diff
@@ -1322,12 +1322,21 @@ public function enable_exceptions($enable = true)
 
 	function cleanMd5($rss)
 	{
-		return md5(preg_replace(array(
-			'#<(lastBuildDate|pubDate|updated|feedDate|dc:date|slash:comments)>[^<]+</\\1>#',
-			'#<(media:starRating|media:statistics) [^/<>]+/>#',
-			'#<!--.+?-->#s',
-			), '', $rss));
-		
+		$data = '';
```
Frenzie: Please also try to submit the fix upstream SimplePie. :-) https://github.com/simplepie/simplepie (After @Alkarex comments.)

Also note the comments we've added to things that aren't upstream.

https://github.com/FreshRSS/FreshRSS/blob/98d7aef49264dea528f89df8b198e4bdc03cd5f1/lib/SimplePie/SimplePie.php#L1514-L1521

lib/SimplePie/SimplePie.php
```diff
@@ -1322,12 +1322,21 @@ public function enable_exceptions($enable = true)
...
+
+		$stream = fopen('php://memory','r+');                             
+		fwrite($stream, $rss);                                           
+		rewind($stream);                                                  
+																		  
+		while (($stream_data = fread($stream, 16384))) {                  
+			$data .= preg_replace(array(
```
Frenzie: What happens if the target of the regular expression happens to be cut in two?

lib/SimplePie/SimplePie.php
Alkarex: This part to allow caching is not (yet) in SimplePie upstream, so this patch cannot directly be sent upstream

Frenzie: No problem then, or is there already a PR for it?

lib/SimplePie/SimplePie/Parser.php
```diff
@@ -181,12 +181,20 @@ public function parse(&$data, $encoding, $url = '')
...
-			{
-				$this->error_code = xml_get_error_code($xml);
-				$this->error_string = xml_error_string($this->error_code);
-				$return = false;
-			}
+			$stream = fopen('php://memory','r+');                             
+			fwrite($stream, $data);                                           
+			rewind($stream);                                                  
+																			  
+			while (($stream_data = fread($stream, 16384))) {
```
Alkarex: But that part could be sent upstream https://github.com/simplepie/simplepie/blob/98ab0fe9f5c04e72c2015eb2ba6a4bedff2dabd1/library/SimplePie/Parser.php#L167

lib/SimplePie/SimplePie.php
Alkarex: @Kiblyn11 In your tests, was this part also crashing, or only the one below regarding `xml_parse()` ?

Kiblyn11: @Frenzie you're right, that case fails.
@Alkarex Yes it was crashing, preg_replace is a memory hog when used with complex regex on fat content

I guess splitting isn't ideal here for that reason, I'm not sure what is the purpose of this cleaning, maybe there is a less expensive way to do it.

lib/SimplePie/SimplePie/Parser.php
Kiblyn11: Alright I will push a PR there

lib/SimplePie/SimplePie.php
Frenzie: The reason is detailed in https://github.com/FreshRSS/FreshRSS/commit/9aab83af115de3b210ea41caae49b70a0f82492a so it looks fairly important to indeed filter them out.

I think https://www.php.net/manual/en/book.xmlreader.php should be able to handle this better (on a per-node basis). It's detailed [here](https://drib.tech/programming/parse-large-xml-files-php).

NB Just brainstorming. I don't have personal experience with it.

Alkarex: Ideally, the filtering should be done after the XML parsing, if not too slow. It is not a big problem if some deletions are not done, as it will only increase the risk slightly of not being able to cache the feed once.

Alkarex:
@Kiblyn11 I have quickly plotted the size of a sample of feeds I use:

image

The current chunking of 16384 bytes looks a bit small to me (i.e. many iterations in the loop). I would feel more comfortable with a larger value. It looks like 256K (262144 bytes) would be a size that allows most feeds (85% in my example) to be processed in a single iteration, without being too large for causing the memory issues you faced.

Could you please try with 262144 instead of 16384 an confirm that it works fine in your case? If you have time, a comparison in terms of total memory and total processing time between the two values would be appreciated (with the example of your big feed).

Frenzie:
Does processing really take that much memory? I realize it'll use exponentially more than 256K, but please keep in mind I run FreshRSS on a VPS with only 512 MB RAM and I've never faced any such issues. I have plenty of feeds in the realm of 300-500K which would prefer to be processed all at once. ;-) For that matter, I think @Alkarex runs it on some weakling Raspberry Pi with probably also 512 MB RAM.

I'm not really sure how to quickly find the largest feed I subscribe to though.

lib/SimplePie/SimplePie/Parser.php
```diff
@@ -181,12 +181,20 @@ public function parse(&$data, $encoding, $url = '')
 			xml_set_element_handler($xml, 'tag_open', 'tag_close');
 
 			// Parse!
-			if (!xml_parse($xml, $data, true))
-			{
-				$this->error_code = xml_get_error_code($xml);
-				$this->error_string = xml_error_string($this->error_code);
-				$return = false;
-			}
+			$stream = fopen('php://memory','r+');
```
Alkarex: @Kiblyn11 Maybe we should consider `php://temp` instead. What do you think? https://php.net/wrappers.php

Alkarex:
@Frenzie Anyway, I think it is nice to fix @Kiblyn11 case. So probably the main remaining point of discussion would be the size of the chunking (could also be 512K or even 1MB). Everything smaller than this chunk size would be unchanged compared to now.

For finding the size of my feeds, I exported to OPML, extracted the URLs to a file, and fetched them with `wget -i`

Alkarex:
P.S. I have upgraded to a 8GB Raspberry Pi :-) But I also still use a cheap 2GB OVH Kimsufi KS-1.

lib/SimplePie/SimplePie/Parser.php
Kiblyn11: @Alkarex yeah sounds better than putting that into memory, have to make sure it's cleaned properly though.
Will look into it.

Kiblyn11:
@Alkarex Will try to update chunk size accordingly and report metrics.
Being a bit busy these days, should be able to update you by next week.

Alkarex:
@Kiblyn11 I have made a few changes, please give it a try https://github.com/FreshRSS/FreshRSS/pull/3416/commits/f88d26a9bbe78d5c504e3526d23afb0d5a556299
* Fixes in error handling (case of the last call to xml_parse, case of
error during fopen, break in case of XML error...)
* Takes advantage of the chunking for computing the cache hash
* Larger chunks of 1MB

Alkarex (approved):

Frenzie (approved):
Should be fine.

Alkarex:
Let's merge to get a bit more testing. @Kiblyn11 feedback welcome, especially some metrics

lib/SimplePie/SimplePie/Parser.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3416

Patriot2407:
+1 I also have this issue

PR #3422, opened 2021-02-05
Add a method to retrieve a configuration value

aledeg:

- Add a method to retrieve a configuration value
-
-

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

This will simplify extension code by removing a lot of logic from
the extension itself when it's not needed. I've tested it on one
of my extension with all the other recent extension modifications
and I could remove half of the code needed before.

### DIFF
--- a/lib/Minz/Extension.php
+++ b/lib/Minz/Extension.php
@@ -12,6 +12,7 @@ abstract class Minz_Extension {
 	private $version;
 	private $type;
 	private $config_key = 'extensions';
+	private $user_configuration;
 
 	public static $authorized_types = array(
 		'system',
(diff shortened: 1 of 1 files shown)


lib/Minz/Extension.php
```diff
@@ -236,6 +237,21 @@ public function getUserConfiguration() {
...
+	 * @param mixed $default
+	 * @return mixed
+	 */
+	public function getUserConfigurationValue(string $key, $default = null) {
+		if (!is_array($this->user_configuration)) {
+			$this->user_configuration = $this->getUserConfiguration();
+		}
+
+		if (array_key_exists($key, $this->user_configuration)) {
+			return $this->user_configuration[$key];
```
Alkarex, line 250: What do you want to return if `$this->user_configuration[$key] === null` ?
`null` or `$default` ?

aledeg: `null` since this is the stored value.

Alkarex: Ok then

PR #3423, opened 2021-02-05
SimplePie: Minor fix return type for broken feeds

Alkarex:
Fix a rare error when an invalid feed is forced to be added again.
FreshRSS code (not upstream)


Alkarex:
Merging to help testing alongside https://github.com/FreshRSS/FreshRSS/pull/3424

PR #3424, opened 2021-02-05
API implement OPML import/export

### LINKED ISSUE #3421: [Feature Request] OPML Import/Export via GReader API
I'm working on adding FreshRSS integration to my RSS Reader, and found that import and export endpoints are not implemented in FreshRSS' `greader.php` file. Just wondering if there is any plan to support either of these.

I can generate a OPML file myself using my app's local cache/db, so I can get away without access to an export endpoint, but, an import endpoint would be nice to have.

I understand different services implement the import endpoint differently (i.e. BazQux vs FeedHQ), so, maybe there is not an agreed upon standard way of doing this. Either way, I just thought I would bring this up, just in case there is any interest.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3421

```
/reader/api/0/subscription/export
/reader/api/0/subscription/import
```

How to test:

```sh
curl 'http://localhost:8080/api/greader.php/accounts/ClientLogin?Email=freshrss&Passwd=freshrss'

curl -v -H "Authorization:GoogleLogin auth=freshrss/6e54ad760d782b9647a2338129e93147e39ad26c" \
  'http://localhost:8080/api/greader.php/reader/api/0/subscription/export'

curl -v -H 'Authorization:GoogleLogin auth=freshrss/6e54ad760d782b9647a2338129e93147e39ad26c' \
  -H  'Content-Type: application/xml' -d '@/home/alexandre/Téléchargements/opml.xml' \
  'http://localhost:8080/api/greader.php/reader/api/0/subscription/import'
```

Check also that the CLI and normal exports/imports are still working.


app/Controllers/importExportController.php
```diff
@@ -213,199 +215,6 @@ private static function guessFileType($filename) {
 		return 'unknown';
 	}
 
-	/**
```
Alkarex, line 216: The whole code below is just moved verbatim to `app/Services/ImportService.php`

app/Services/ImportService.php
```diff
@@ -0,0 +1,221 @@
...
+	 */
+	public function __construct($username) {
+		require_once(LIB_PATH . '/lib_opml.php');
+
+		$this->username = $username;
+		$this->catDAO = FreshRSS_Factory::createCategoryDao($username);
+		$this->feedDAO = FreshRSS_Factory::createFeedDao($username);
+	}
+
+	/**
```
Alkarex, line 29: The whole code below comes verbatim from `app/Controllers/importExportController.php`

Alkarex:
Tested by reporter https://github.com/FreshRSS/FreshRSS/issues/3421

PR #3430, opened 2021-02-07
Check that PHP session is working during install

### LINKED ISSUE #3198: docker-compose: can't connect to DB
I am trying to install FreshRSS on my server using docker-compose, but I get this error when setting up the database:
> **Arf !** Vérifiez les informations d’accès à la base de données. : Access to database is denied for `freshrss`: SQLSTATE[08006] [7] could not connect to server: Connection refused Is the server running on host "localhost" (127.0.0.1) and accepting TCP/IP connections on port 5432? could not connect to server: Cannot assign requested address Is the server running on host "localhost" (::1) and accepting TCP/IP connections on port 5432?

## Steps to reproduce

_I tried the same steps on Fedora 32 (the Linux on my local machine) using `podman-compose` and it worked as expected._

_It is when I installed FreshRSS on my server that I got the error. The differences are that my server is on Ubuntu 18.04 and I use Nginx on my server (while I was using `localhost:8080` on my machine)_

1. Be on Ubuntu 18.04
2. Copy the [`docker-compose.yml`](https://raw.githubusercontent.com/FreshRSS/FreshRSS/master/Docker/docker-compose.yml) file locally
3. Run `docker-compose up -d`
4. Set up nginx using [the conf given in the FreshRSS repo](https://github.com/FreshRSS/FreshRSS/tree/master/Docker#alternative-reverse-proxy-using-nginx) but removing the SSL part for now:

        upstream freshrss {
            server 127.0.0.1:8500;
        }
        
        server {
            server_name feed.example.net;
            listen      80;
            listen [::]:80;
    
            # Needed for Freshrss cookie/session :
            proxy_cookie_path / "/; HTTPOnly; Secure; SameSite=Lax";
    
            location / {
                    proxy_pass http://freshrss/;
                    add_header X-Frame-Options SAMEORIGIN;
                    add_header X-XSS-Protection "1; mode=block";
                    proxy_redirect off;
                    proxy_buffering off;
                    proxy_set_header Host $host;
                    proxy_set_header X-Real-IP $remote_addr;
                    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                    proxy_set_header X-Forwarded-Proto $scheme;
                    proxy_set_header X-Forwarded-Port $server_port;
                    proxy_read_timeout 90;
    
                    # Forward the Authorization header for the Google Reader API.
                    proxy_set_header Authorization $http_authorization;
                    proxy_pass_header Authorization;
            }
        }
5. Restart Nginx: `sudo systemctl reload nginx.service`
6. Go to `http://feed.example.net`
7. Set up the language, pass the verifications, and set up the database like this:
    db type: postgresql
    host: localhost
    user: freshrss
    passwd: freshrss
    database: freshrss
8. Validate
9. See the error appear in red at the top:
    image

# Debugging

I learned that with PostgreSQL we can test the db connection with `pg_isready` so I did that, first in the `db` container and then in the `app` container.

## In freshrss-db

1. Get into the container:

        $ docker exec -it freshrss-db bash

2. Run:

        $ pg_isready -d freshrss -h localhost -p 5432 -U freshrss

3. See the output:

        localhost:5432 - accepting connections

## In freshrss-app

1. Get into the container:

        $ docker exec -it freshrss-app bash

2. Update the packages and install postgresql-client

        $ apt update
        $ apt install postgresql-client

3. Run:

        $ pg_isready -d freshrss -h localhost -p 5432 -U freshrss

4. See the output:

        localhost:5432 - no response

5. Try with the host `freshrss-db` instead:

        $ pg_isready -d freshrss -h freshrss-db -p 5432 -U freshrss

6. See the output:

        freshrss-db:5432 - accepting connections

As you can see, the connections are accepted on `freshrss-db`, but then if I change db host from `localhost` to `freshrss-db` in the form, I get a weird behavior: it reload the DB setup page as if I first arrived on it (with only SQLite selected in the dropdown).

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3198

### DIFF
--- a/app/i18n/cz/install.php
+++ b/app/i18n/cz/install.php
@@ -122,7 +122,7 @@ return array(
 	'not_deleted' => 'Nastala chyba, soubor <em>%s</em> musíte smazat ručně.',
 	'ok' => 'Instalace byla úspěšná.',
 	'session' => array(
-		'nok' => 'The web server seems to be incorrectly configured for cookies required for PHP sessions! ',	// TODO - Translation
+		'nok' => 'The web server seems to be incorrectly configured for cookies required for PHP sessions!',	// TODO - Translation
 	),
 	'step' => 'krok %d',
 	'steps' => 'Kroky',
--- a/app/i18n/cz/install.php
+++ b/app/i18n/cz/install.php
@@ -121,6 +121,9 @@ return array(
 	),
 	'not_deleted' => 'Nastala chyba, soubor <em>%s</em> musíte smazat ručně.',
 	'ok' => 'Instalace byla úspěšná.',
+	'session' => array(
+		'nok' => 'The web server seems to be incorrectly configured for cookies required for PHP sessions! ',	// TODO - Translation
+	),
 	'step' => 'krok %d',
 	'steps' => 'Kroky',
 	'this_is_the_end' => 'Konec',
--- a/app/i18n/de/install.php
+++ b/app/i18n/de/install.php
@@ -121,6 +121,9 @@ return array(
 	),
 	'not_deleted' => 'Etwas ist schiefgelaufen; Sie müssen die Datei <em>%s</em> manuell löschen.',
 	'ok' => 'Der Installationsvorgang war erfolgreich.',
+	'session' => array(
+		'nok' => 'The web server seems to be incorrectly configured for cookies required for PHP sessions! ',	// TODO - Translation
+	),
 	'step' => 'Schritt %d',
 	'steps' => 'Schritte',
 	'this_is_the_end' => 'Das ist das Ende',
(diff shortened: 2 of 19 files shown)


Frenzie (approved):

Alkarex:
image

aledeg (changes requested):
If I understand correctly, you set a value in session to `ok` then read it and check if you have `ok` in session.
I find that it could be fragile if the two are not in sync (one can forgot to change it). I'd rather have the value in a constant instead of having to rely on people to not forget to sync those values.

Alkarex:
Ups, sorry @aledeg , I had just merged when I saw your comment

aledeg:
Don't worry :)

Alkarex:
We use `'ok'` everywhere in that file :-)

aledeg:
I know! And that bugs me :)
We should use boolean instead for the other values. But it's out of the scope I was mentioning.

Here, it makes no sense to use a boolean but a string. I would have envisioned a `SESSION_TEST` constant with whatever data in it. `dummy content` for instance.

Anyway, it's not a big deal :)

PR #3433, opened 2021-02-09
Add support for extension user files

### LINKED ISSUE #1709: Minz_Extensions and accessing files
While working on https://github.com/FreshRSS/Extensions/pull/23 I stumbled upon this comment https://github.com/FreshRSS/FreshRSS/issues/1456#issuecomment-348797338 by @Alkarex regarding the `./data/extensions-data/` directory.

Some thoughts about that:
- What about adding a second method besides Minz_Extension::getFileUrl() which uses the new path and a deprecation warning for the old one? By now, its unclear where files should be stored.
- What about a helper method like Minz_Extension::getUserPath() which could be consumed by extension authors?
- Maybe some sentences about these directories in the extension docu?

But my main issue is:
I cannot use the new directory due to security limitations in the p/ext.php:
https://github.com/FreshRSS/FreshRSS/blob/master/p/ext.php#L20

The script only allows to access files in the extensions static/ sub-directories

Edit: Probably the old behaviour should be removed in 2.0 completely, as the data/ directory seems to be the place where such file changes should happen exclusively.

aledeg:
Closes #1709 (possibly)

- Add support for extension user files
-
-

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Extension user files can be stored easily in the user folder instead of the
static folder.

### DIFF
--- a/p/ext.php
+++ b/p/ext.php
@@ -32,10 +32,15 @@ function get_absolute_filename(string $file_name) {
 		return $third_party_extension;
 	}
 
+	$user = realpath(USERS_PATH . '/' . $file_name);
+	if (false !== $user) {
+		return $user;
+	}
+
 	return '';
 }
 
-function is_valid_path_extension($path, $extensionPath) {
+function is_valid_path_extension($path, $extensionPath, $isStatic = true) {
 	// It must be under the extension path.
 	$real_ext_path = realpath($extensionPath);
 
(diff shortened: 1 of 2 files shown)


aledeg:
Please let me know if something is missing.

p/ext.php
```diff
@@ -48,7 +53,12 @@ function is_valid_path_extension($path, $extensionPath) {
 		return false;
 	}
 
-	// File to serve must be under a `ext_dir/static/` directory.
+	// User files do not need further validations
+	if (!$isStatic) {
```
Alkarex, line 57: We should not allow arbitrary paths in the filesystem, so `$path` needs to be constrained somehow, at least preventing things such as `../`

aledeg: This validation is skipped only for files generated by the extension configuration in the user folder. And the validation specifies that it must be in the user folder.
I am confused why we should be testing static for those files. Because in the event of the use of `../`, it has to be contained in the user folder. So I guess the risks are really low.

Alkarex:
Security regression https://github.com/FreshRSS/FreshRSS/pull/4928
@aledeg We need to move this functionality to a normal controller to check that we are serving a file from the proper user.
`ext.php` should be limited to files not requiring login.

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/4930

Alkarex:
Taking advantage of this PR: https://github.com/FreshRSS/FreshRSS/pull/6267

PR #3440, opened 2021-02-11
Add user maintenance hooks

### LINKED ISSUE #3398: [Feature request] Add maintenance hooks
I think it could be useful to introduce new hooks for the extension system.
I am thinking hooks that can be use for maintenance purposes. It could be at the system level (perform a vacuum on sqlite databases), at the user level (remove useless configuration), or at the feed level (modify existing article content).

At the moment, we have the `feed_before_insert` hook which could be used for the feed level but it is triggered (as far as I understand) only when the feed is updated.
We also have the `freshrss_init` hook which could be used for the system level but it is trigger for every calls. It's probably too much for a maintenance task.

What I envisioned is hooks that are triggered only during the `actualize_script.php` process. In the file itself, we could have the hook at the system level and the hook at the user level. The hook at the feed level would have to be in the `feedController.php` process.

I suggest the following names:
- `freshrss_system_maintenance`
- `freshrss_user_maintenance`
- `freshrss_feed_maintenance`

aledeg:
Closes #3398 (partially)

- Add user maintenance hooks

1. Add an extension using a new hook.
2. Run the `make refresh` command.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

The new hook allows to add maintenance extensions at the user level.

### DIFF
--- a/app/actualize_script.php
+++ b/app/actualize_script.php
@@ -72,6 +72,9 @@ foreach ($users as $user) {
 	}
 
 	FreshRSS_Auth::giveAccess();
+
+	Minz_ExtensionManager::callHook('freshrss_user_maintenance');
+
 	$app->init();
 	notice('FreshRSS actualize ' . $user . '...');
 	echo $user, ' ';	//Buffered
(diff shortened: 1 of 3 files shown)


app/actualize_script.php
```diff
@@ -72,6 +72,9 @@ function notice($message) {
 	}
 
 	FreshRSS_Auth::giveAccess();
+
+	Minz_ExtensionManager::callHook('freshrss_user_maintenance');
```
Alkarex, line 76: Would be good to put it in the CLI version as well https://github.com/FreshRSS/FreshRSS/blob/master/cli/actualize-user.php

aledeg: I've tried but it seems that it is not triggered. :/

Alkarex: Could you please try with https://github.com/FreshRSS/FreshRSS/pull/3443 ?

aledeg: I've tried without success

Alkarex: Have you tested with a system or user extension?

Alkarex: Could you please try again with https://github.com/FreshRSS/FreshRSS/pull/3443 ?

aledeg: I tried with a system extension

aledeg: So far, no success either :/

Alkarex: Let's merge here and come back to it when https://github.com/FreshRSS/FreshRSS/pull/3443 works

PR #3446, opened 2021-02-14
Change confusing naming for label

aledeg:
Closes #3444

Changes proposed in this pull request:

- Change confusing naming for label

How to test the feature manually:

1. Check label management page

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/master/docs/en/developers/04_Pull_requests.md).

Before, labels were referred to as tag which was confusing since
tags are content generated (retrieved from the feed) and label
are user generated.
Now, labels are properly named in their management page.

PR #3448, opened 2021-02-14
Updates DE Translation: tags -> labels

rom-1:
in addition to PR3446. see https://github.com/FreshRSS/FreshRSS/pull/3446

Edit: There is no real translation for 'labels' in german, as this word is 1:1 used in our language for technical 'labels'. btw google does the same with gmail: they are simply using the word "labels" in german.


app/i18n/de/sub.php
```diff
@@ -124,10 +124,10 @@
 		'add' => 'Feed oder Kategorie hinzufügen',
 		'add_category' => 'Kategorie hinzufügen',
 		'add_feed' => 'Feed hinzufügen',
-		'add_tag' => 'Tag hinzufügen',
-		'delete_tag' => 'Tag löschen',
+		'add_label' => 'Label hinufügen',
```
aledeg, line 127: It seems that there is a `z` missing?

app/i18n/de/index.php
```diff
@@ -53,12 +53,12 @@
...
 		'stats' => 'Statistiken',
 		'subscription' => 'Abonnementverwaltung',
-		'tags' => 'Meine Tags',
+		'tags' => 'Meine Labels',
 		'unread' => 'Nur ungelesene zeigen',
 	),
 	'share' => 'Teilen',
 	'tag' => array(
-		'related' => 'Verwandte Tags',
+		'related' => 'Artikel-Labels',
```
aledeg, line 61: This is actually an article tag not a label :)

app/i18n/de/sub.php
rom-1: correct, sorry, "hinzufügen"

app/i18n/de/conf.php
```diff
@@ -28,7 +28,7 @@
 			'display_authors' => 'Autoren',
 			'entry' => 'Artikel-Symbole',
 			'publication_date' => 'Datum der Veröffentlichung',
-			'related_tags' => 'Verwandte Tags',
+			'related_tags' => 'Artikel-Labels',
```
aledeg, line 31: article tag as well

app/i18n/de/index.php
rom-1: whoops, sorry :)

app/i18n/de/sub.php
aledeg: Perfect!

app/i18n/de/conf.php
rom-1: are you sure about this? I am refering to Settings Display -> in german, this looks like this right now:
2021-02-14_15h46_24

So if I check this for "bottom line", it shows "my labels" in the bottom line. therefore, this seem to relate to labels?

aledeg:
@rom-1 I've updated my PR with your translations. Please review!
If it's correct, I think you can close this one.

app/i18n/de/conf.php
Frenzie: Labels are added by the user and tags come from the feed. Are they both toggled by the same setting?

rom-1: Update: ok, I don't get it. Even if I "uncheck" the "verwandte tags" (related tags) option in the bottom line (fußzeile), the "my labels" is showing in the bottom line in the main view. so this setting does not seem to have any effect at all.

aledeg: This toggle show article tags (from feed) in the article itself.
Labels are user generated. You cannot turn off labels. But you can turn off tags.

Frenzie: No, that's what I was (more or less) expecting to happen. That means you can't turn them off in the display there, although one might wonder why not.

For tags you need an article that actually has some. They come from the feed.
Screenshot_2021-02-14_15-53-15

rom-1: oh. ok, then I mixed it up. I was not aware that you cannot turn of labels, sorry. 
seems I simply do not have any feeds which support article tags then.

aledeg: You explain it better than I did :)
I was trying to say the same thing but in the worst possible way

rom-1:
Changes have been implemented to #3446

PR #3453, opened 2021-02-16
Add a content action parameter to work with CSS selector

### LINKED ISSUE #3451: [Feature request] Add a parameter to either replace or update article content while using a CSS selector
By default, the parameter would replace the content.

Some feeds have valuable information even though it's partial. The idea is then to use the CSS selector to grab the complete data. But if the data provided is slightly different from the feed itself, the data from the feed is lost.

To work around that limitation, my only solution is to have an extension grabbing the content for those feeds. It's not really scalable since I have to write an extension per feed.
A parameter would be easier.

aledeg:
Closes #3451

- Add a content action parameter to work with CSS selector

1. Change feed configuration while using a CSS selector to retrieve the content
2. Refresh the feed
3. See if the content has changed according to the configuration (cache need to be cleaned!)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, when retrieving article contents with CSS selector, the content
of the RSS feed was replaced by the content of the CSS selector. It works
well in most of the cases but if there is a different content in the feed
and in the selector, the former is lost.
Now, there is a parameter to decide which action is performed after retrieving
the content. By default, the previous behavior is kept. But now, it is
possible to append or prepend the CSS selector content to the feed content.


Alkarex (approved):

aledeg:
@Alkarex This does not work properly. When selecting `append` or `prepend`, and reloading articles, the content is retrieved again and append to the already retrieved content. This might be getting out of control. It works fine on the first use but then it keeps appending or preprending content.
I think we should revert it.

aledeg:
Or we need to find a way to identify the content being added and remove that content beforehand. Either way, there is something to do before shipping.

Alkarex:
Ok, well spotted

aledeg:
I'll tried to do something about it. Probably by adding some HTML comments in the content to be able to clean the content before updating it. Let's not revert it now except if you want to ship the next release today :)

Alkarex:
1.8 is not ready for today (unfortunately) - there are still a few things to close before a RC (the top ~5 of https://github.com/FreshRSS/FreshRSS/milestone/46 )

PR #3455, opened 2021-02-17
Update DE Translation

rom-1:
Updating DE Translation, addition to https://github.com/FreshRSS/FreshRSS/pull/3453

### DIFF
--- a/app/i18n/de/sub.php
+++ b/app/i18n/de/sub.php
@@ -34,10 +34,10 @@ return array(
 		),
 		'clear_cache' => 'Nicht cachen (für defekte Feeds)',
 		'content_action' => array(
-			'_' => 'Content action when fetching the article content',	// TODO - Translation
-			'append' => 'Add after existing content',	// TODO - Translation
-			'prepend' => 'Add before existing content',	// TODO - Translation
-			'replace' => 'Replace existing content',	// TODO - Translation
+			'_' => 'Behandlung von Feed-Inhalt beim Herunterladen von Artikelinhalt',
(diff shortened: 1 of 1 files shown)


app/i18n/de/sub.php
```diff
@@ -34,10 +34,10 @@
 		),
 		'clear_cache' => 'Nicht cachen (für defekte Feeds)',
 		'content_action' => array(
-			'_' => 'Content action when fetching the article content',	// TODO - Translation
-			'append' => 'Add after existing content',	// TODO - Translation
-			'prepend' => 'Add before existing content',	// TODO - Translation
-			'replace' => 'Replace existing content',	// TODO - Translation
+			'_' => 'Behandlung von Feed-Inhalt beim Herunterladen von Artikelinhalt',
+			'append' => 'Artikelinhalt hinter Feed-Inhalt platzieren',
```
Frenzie: Not something like einfügen?

Frenzie (approved):
lgtm

app/i18n/de/sub.php
rom-1: you are right, it's too early in the morning for my brain. platzieren is not the wrong word, but I guess hinzufügen simply fits best in this case.

https://github.com/FreshRSS/FreshRSS/pull/3455/commits/8a2b424b6f073632eef2cf851287ce3fbea584b3

Frenzie: Haha, I actually thought you made a conscious decision against hinzufügen.

aledeg (approved):

PR #3461, opened 2021-02-18
SQLite: fix updateEntry

### LINKED ISSUE #3130: SQLite: Reloading article after having added a selector does not work
**How to reproduce :**

- Add a new feed
- Edit this feed
- Add a selector
- Check that the selector gives the good content with the new "eye" button
- Click on Validate ("Valider", in French)

Then go again to edit the feed :
- Click on Reload articles ("Recharger les articles", in French)

**Actual :**

The reloading is ***very slow***  but the feed does not have the website content expected.

**Expected :**

All the feeds contain the content from the original website

**Note :**

For the new articles published after, the website content will be well retrieved.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3130


app/Controllers/feedController.php
```diff
@@ -734,19 +734,14 @@ public function reloadAction() {
 
 		//Extract all feed entries from database, load complete content and store them back in database.
 		$entries = $entryDAO->listWhere('f', $feed_id, FreshRSS_Entry::STATE_ALL, 'DESC', 0);
-
-		//We need another DB connection in parallel
-		Minz_ModelPdo::$usesSharedPdo = false;
-		$entryDAO2 = FreshRSS_Factory::createEntryDao();
+		$entries = iterator_to_array($entries);	//TODO: Improve
```
Alkarex: For debugging: This is to disable streaming = load all articles in memory. But we might exhaust memory by doing so.

app/Models/EntryDAO.php
```diff
@@ -183,13 +183,14 @@ public function updateEntry($valuesTmp) {
...
 
+		$this->updateEntryPrepared = null;	//TODO: Remove this line after debugging
 		if ($this->updateEntryPrepared === null) {
 			$sql = 'UPDATE `_entry` '
 				. 'SET title=:title, author=:author, '
 				. ($this->isCompressed() ? 'content_bin=COMPRESS(:content)' : 'content=:content')
 				. ', link=:link, date=:date, `lastSeen`=:last_seen, '
 				. 'hash=' . $this->sqlHexDecode(':hash')
-				. ', ' . ($valuesTmp['is_read'] === null ? '' : 'is_read=:is_read, ')
+				. ', ' . ($valuesTmp['is_read'] === null ? '' : 'is_read=:is_read, ')	//FIXME: Likely error in the logic here, if this gets cached for next entry
```
Alkarex: Probable (unrelated?) bug

app/Models/EntryDAO.php
```diff
@@ -183,13 +183,14 @@ public function updateEntry($valuesTmp) {
 			$valuesTmp['is_read'] = null;
 		}
 
+		$this->updateEntryPrepared = null;	//TODO: Remove this line after debugging
```
Alkarex: Disable caching of prepared statement for debugging

app/Controllers/feedController.php
```diff
@@ -734,19 +734,14 @@ public function reloadAction() {
...
 
 		foreach ($entries as $entry) {
 			if ($entry->loadCompleteContent(true)) {
-				$entryDAO2->updateEntry($entry->toArray());
+				$entryDAO->updateEntry($entry->toArray());
```
aledeg: I've used only that modification and it works. There might be something fishy about the locking mechanism.

app/Controllers/feedController.php
aledeg: The tested feed retrieves only 10 articles. We're good for the memory consumption :)

Alkarex: I think we need to introduce an explicit limit there. The line a bit higher up retrieves all articles for that feed, which can be several thousands...

Alkarex:
Working on the issue. For the record, here is a script to generate an arbitrary long RSS feed to test article retrieval:

Use it with a parameter such as `https://example.net/long.rss.php?n=500`

```php
<?php
	$url = 'https://example.net/long.rss.php';

	$n = intval($_GET['n'] ?? '10');
	$i = intval($_GET['i'] ?? '0');
	$timestamp = strtotime('2021-01-01T01:01:01Z');

if ($i <= 0):
	header('Content-Type: application/atom+xml; charset=UTF-8');
?>
	<title>Long RSS</title>
	<subtitle>A long RSS feed</subtitle>
	<link href="<?= $url ?>?n=<?= $n ?>" rel="self" />
	<id><?= $url ?>?n=<?= $n ?></id>
	<updated><?= date('c', $timestamp + 0) ?></updated>

	<?php for ($i = 1; $i <= $n; $i++): ?>

	<entry>
		<title>Entry <?= $i ?></title>
		<link href="<?= $url ?>?i=<?= $i ?>" />
		<id><?= $url ?>?i=<?= $i ?></id>
		<summary>A summary for entry <?= $i ?>.</summary>
		<content type="xhtml">
			<div xmlns="http://www.w3.org/1999/xhtml">
				Some content for entry <?= $i ?>

			</div>
		</content>
		<updated><?= date('c', $timestamp + ($i * 10)) ?></updated>
		<author>
			<name>Some One</name>
			<email>some.one@example.net</email>
		</author>
	</entry>
	<?php endfor; ?>
</feed>
<?php
else:
	header('Content-Type: text/html; charset=UTF-8');
?>
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" lang="en-GB">
<head>
<meta charset="UTF-8" />
<title>Entry <?= $i ?></title>
<meta name="robots" content="none" />
</head>

<body>
	<h1>Entry <?= $i ?></h1>
	<p>
		Some longer original content for entry <?= $i ?>.<br />
		This entry was produced at <?= date('c', $timestamp + ($i * 10)) ?>.
	</p>
	<blockquote lang="la">
		Lorem ipsum dolor sit amet, consectetur adipiscing elit.
		Phasellus ex est, placerat in efficitur vel, tincidunt sit amet eros.
		Nullam fringilla id odio non semper.
		Suspendisse mattis gravida nunc eget faucibus.
		Mauris sollicitudin euismod libero varius elementum.
		Duis varius tellus nec vulputate pellentesque.
		Praesent nec venenatis sapien.
		Cras sodales bibendum nulla vitae lobortis.
		Nulla quis elit nec risus rhoncus mollis eget quis diam.
	</blockquote>
</body>
</html>
<?php
endif;
?>
```

app/Controllers/feedController.php
Alkarex: ~~This line enough does not seem enough for SQLite when dealing with a larger number of entries~~

Alkarex: Update: scratch that - seems to work after all :-)

app/Models/EntryDAO.php
Alkarex: Will fix in another PR

Alkarex:
Tested with 500 articles with SQLite, MySQL, PostgreSQL (latest versions, PHP 8.0.2)

Alkarex:
Also tested with SQLite with PHP 5.6.36

app/Models/EntryDAO.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3500

app/Models/EntryDAO.php
```diff
@@ -189,7 +189,7 @@ public function updateEntry($valuesTmp) {
 				. ($this->isCompressed() ? 'content_bin=COMPRESS(:content)' : 'content=:content')
 				. ', link=:link, date=:date, `lastSeen`=:last_seen, '
 				. 'hash=' . $this->sqlHexDecode(':hash')
-				. ', ' . ($valuesTmp['is_read'] === null ? '' : 'is_read=:is_read, ')
+				. ', ' . ($valuesTmp['is_read'] === null ? '' : 'is_read=:is_read, ')	//FIXME: Likely error in the logic here, if this gets cached for next entry
```
Alkarex:
```suggestion
				. ', ' . ($valuesTmp['is_read'] === null ? '' : 'is_read=:is_read, ')
```

aledeg (approved):
Seems fair.
But I wonder why we need a second connection for other databases.

Alkarex:
> But I wonder why we need a second connection for other databases.

@aledeg ~~You are correct - it looks like it is actually working with a single connection with other databases as well! Maybe that is after we are using native prepared statements instead of emulated...~~

Alkarex:
I will test some more a bit later

Alkarex:
Update:
* SQLite requires the single connection
* MySQL requires the second connection, as we are using an unbuffered query:

> HY000: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

* PostgreSQL works with both options, but is not taking advantage of the unbuffered streaming at the moment (TODO for later)

PR #3463, opened 2021-02-18
Add full content delimiter and action

aledeg:

- Add full content delimiter and action

1. Reload many times a feed with a content retrieved by a CSS selector (append or prepend mode only)
2. Check that the retrieved content is not repeated

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, when appending or prepending the content of the CSS selector
content, it was added to the content. It was working fine for the
first call but every subsequent calls were pilling the retrieved
content on top of the already retrieved content. Thus we had an ever
growing content with a lot of duplication.
Now, the CSS selector content is identified by an HTML comment which
is used to remove the content for every subsequent calls.

The bug was introduced in #3453

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -447,12 +447,14 @@ class FreshRSS_Entry extends Minz_Model {
 						$feed->attributes()
 					);
 					if ('' !== $fullContent) {
+						$fullContent = "<!-- FULLCONTENT start //-->{$fullContent}<!-- FULLCONTENT end //-->";
+						$originalContent = preg_replace('#<!-- FULLCONTENT start //-->.*<!-- FULLCONTENT end //-->#s', '', $this->content());
 						switch ($feed->attributes('content_action')) {
 							case 'prepend':
-								$this->content = $fullContent . $this->content();
+								$this->content = $fullContent . $originalContent;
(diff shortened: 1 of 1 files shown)


app/Models/Entry.php
```diff
@@ -447,12 +447,14 @@ public function loadCompleteContent($force = false) {
 						$feed->attributes()
 					);
 					if ('' !== $fullContent) {
+						$fullContent = "<!-- FULLCONTENT start //-->{$fullContent}<!-- FULLCONTENT end //-->";
+						$originalContent = preg_replace('#<!-- FULLCONTENT start //-->.*<!-- FULLCONTENT end //-->#s', '', $this->content());
```
Alkarex, line 451: There might be a risk in some situations that the end of a very large article might be truncated.
I therefore suggest searching for `<!-- FULLCONTENT start //-->`, and getting the rest.
Why is `<!-- FULLCONTENT end //-->` needed?

aledeg: The `<!-- FULLCONTENT end //-->` is needed when you preprend the content. The `<!-- FULLCONTENT start //-->` is needed when you append the content.

PR #3466, opened 2021-02-19
Fix feed sort order

### LINKED ISSUE #3228: Feed title sorting in LEFT COLUMN shoudn't be case sensitive
Feed title sorting in LEFT COLUMN shouldn't be case sensitive.
As you can see all uppercase feeds are sorted before the lowercase ones. It would be best to keep them all sorted as case insensitive, respecting alphabetical order.

This is what I meant in https://github.com/FreshRSS/FreshRSS/issues/3144, but being a new user I didn't know correct terminology. I thought that fixing one would fix the other, but apparently that's not the case.

FreshRSS 1.17.0

aledeg:
Closes #3228

- Fix feed sort order

1. check that the feed order does not change when changing the case of its name

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, the feeds in the list weren't properly ordered. Uppercase
values were before lowercase values.
Now, the feed order is forced to ignore the case.

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -62,6 +62,10 @@ class FreshRSS_Category extends Minz_Model {
 			}
 		}
 
+		usort($this->feeds, function ($a, $b) {
+			return strnatcasecmp($a->name(), $b->name());
+		});
+
 		return $this->feeds;
 	}
 


Alkarex:
This was due to difference in our collations for some databases. Could be fixed there as well

PR #3470, opened 2021-02-20
Change zip handling methods

### LINKED ISSUE #3460: Update Zip functions to use the Object API
While looking into updating the tests for PHP 8.0 (#3458) The first set of tests ran for PHP 8 and the deprecation of the `zip_` functions was flagged. 

```
FILE: FreshRSS/FreshRSS/app/Controllers/importExportController.php

--------------------------------------------------------------------------------
FOUND 6 ERRORS AFFECTING 6 LINES
--------------------------------------------------------------------------------
 68 | ERROR | Function zip_open() has been deprecated
 73 | ERROR | Function zip_read() has been deprecated
 78 | ERROR | Function zip_entry_name() has been deprecated
 80 | ERROR | Function zip_entry_read() has been deprecated
 82 | ERROR | Function zip_entry_filesize() has been deprecated
 87 | ERROR | Function zip_close() has been deprecated
--------------------------------------------------------------------------------
```

I did limited research but it appears these should be replaced with the Object API:

For example: `zip_open()` replaced with `ZipArchive::open` 

`zip_open` : https://www.php.net/manual/en/function.zip-open.php
`ZipArchive::open` : https://www.php.net/manual/en/ziparchive.open.php

aledeg:
Closes #3460

- Change zip handling methods

1. Export a user subscriptions
2. Import for another user the generated file

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, we were using zip function but they are marked as deprecated as of
PHP 8.0. It's not safe to use them anymore since they can be removed at any
given time.
Now, we are using the ZipArchive class to handle our zip methods. It's safe
to use it since it's available for PHP 5.2 and higher.

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -64,31 +64,24 @@ class FreshRSS_importExport_Controller extends Minz_ActionController {
 
 		// We try to list all files according to their type
 		$list = array();
-		if ($type_file === 'zip' && extension_loaded('zip')) {
-			$zip = zip_open($path);
-			if (!is_resource($zip)) {
+		if ('zip' === $type_file && extension_loaded('zip')) {
+			$zip = new ZipArchive();
+			$result = $zip->open($path);
+			if (true !== $result) {
 				// zip_open cannot open file: something is wrong
-				throw new FreshRSS_Zip_Exception($zip);
+				throw new FreshRSS_Zip_Exception($result);
 			}
-			while (($zipfile = zip_read($zip)) !== false) {
-				if (!is_resource($zipfile)) {
-					// zip_entry() can also return an error code!
-					throw new FreshRSS_Zip_Exception($zipfile);
-				} else {
-					$type_zipfile = self::guessFileType(zip_entry_name($zipfile));
-					if ($type_file !== 'unknown') {
-						$list_files[$type_zipfile][] = zip_entry_read(
-							$zipfile,
-							zip_entry_filesize($zipfile)
-						);
-					}
+			for ($i = 0; $i < $zip->numFiles; $i++) {
+				$type_zipfile = self::guessFileType($zip->getNameIndex($i));
+				if ('unknown' !== $type_zipfile) {
+					$list_files[$type_zipfile][] = $zip->getFromIndex($i);
 				}
 			}
-			zip_close($zip);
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Seems fine.

app/Controllers/importExportController.php
```diff
@@ -64,31 +64,24 @@ public function importFile($name, $path, $username = null) {
 
 		// We try to list all files according to their type
 		$list = array();
-		if ($type_file === 'zip' && extension_loaded('zip')) {
-			$zip = zip_open($path);
-			if (!is_resource($zip)) {
+		if ('zip' === $type_file && extension_loaded('zip')) {
+			$zip = new ZipArchive();
+			$result = $zip->open($path, ZipArchive::RDONLY);
```
Alkarex: The constant `ZipArchive::RDONLY` is first available from PHP 7.4.3 https://php.net/zip.constants

Alkarex:
```suggestion
			$result = $zip->open($path, 16);	//ZipArchive::RDONLY from PHP 7.4.3+
```

Alkarex:
Great @aledeg 👍🏻

app/Controllers/importExportController.php
aledeg: Thanks, I haven't pay attention to that. We might as well drop it.

app/Controllers/importExportController.php
```diff
@@ -64,31 +64,24 @@ public function importFile($name, $path, $username = null) {
 
 		// We try to list all files according to their type
 		$list = array();
-		if ($type_file === 'zip' && extension_loaded('zip')) {
-			$zip = zip_open($path);
-			if (!is_resource($zip)) {
+		if ('zip' === $type_file && extension_loaded('zip')) {
+			$zip = new ZipArchive();
+			$result = $zip->open($path);
```
Alkarex, line 69:
```suggestion
			$result = $zip->open($path, 16);	//ZipArchive::RDONLY from PHP 7.4.3+
```

What about this suggestion?

aledeg: I feel that it's not needed. I can add it if you prefer though

PR #3472, opened 2021-02-20
"Show all" option on labels

### LINKED ISSUE #3263: [Feature Request] "Show all" default option for labels
There's already an option to default to showing all articles when viewing favorites. Would it be possible to add a similar option for labels as well? Coming from Feedly, I like having unread-only as the default view and use favorites to save articles, but then I categorize those favorites using labels, or what Feedly calls "boards".

aledeg:
Closes #3263

- "Show all" option on labels

1. Modify the configuration and check the label content

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

It uses the favorite option to keep a similar experience through out
the application and also to limit the number of options.

### DIFF
--- a/app/Models/Context.php
+++ b/app/Models/Context.php
@@ -292,9 +292,10 @@ class FreshRSS_Context {
 			self::$get_unread = $tag->nbUnread();
 			break;
 		case 'T':
+			$tagDAO = FreshRSS_Factory::createTagDao();
 			self::$current_get['tags'] = true;
 			self::$name = _t('index.menu.tags');
-			self::$get_unread = 0;
+			self::$get_unread = $tagDAO->countNotRead();
 			break;
 		default:
 			throw new FreshRSS_Context_Exception('Invalid getter: ' . $get);
--- a/app/Models/TagDAO.php
+++ b/app/Models/TagDAO.php
@@ -259,12 +259,19 @@ class FreshRSS_TagDAO extends Minz_ModelPdo implements FreshRSS_Searchable {
 		return $res[0]['count'];
 	}
 
-	public function countNotRead($id) {
+	public function countNotRead($id = null) {
 		$sql = 'SELECT COUNT(*) AS count FROM `_entrytag` et '
 			 . 'INNER JOIN `_entry` e ON et.id_entry=e.id '
-			 . 'WHERE et.id_tag=? AND e.is_read=0';
+			 . 'WHERE e.is_read=0';
+		$values = null;
+
+		if (null !== $id) {
+			$sql .= ' AND et.id_tag=?';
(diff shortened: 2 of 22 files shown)


app/Models/Context.php
```diff
@@ -292,9 +292,10 @@ public static function _get($get) {
 			self::$get_unread = $tag->nbUnread();
 			break;
 		case 'T':
+			$tagDAO = FreshRSS_Factory::createTagDao();
 			self::$current_get['tags'] = true;
 			self::$name = _t('index.menu.tags');
-			self::$get_unread = 0;
+			self::$get_unread = $tagDAO->countNotRead();
```
Alkarex, line 298: Is this additional request really needed?
(It is a bit expensive: ~120ms on my test system)

aledeg: If you don't do that, the filter does nothing when you select "My labels"

Alkarex: I have only quickly tested but could not see a problem. Could you please post screenshots with/without ?

aledeg: The problem arise when you select "Adjust showing" in the reading configuration page.
The code that messes things up is this one:
https://github.com/FreshRSS/FreshRSS/blob/fd803875410cf056242a786260fb8ed62a733c06/app/Controllers/indexController.php#L199-L204

aledeg: The steps to reproduce is to select "Adjust showing", to add labels on multiple articles (both read and unread) and to select "My labels".
It works fine while selecting individual labels though!

aledeg: If it's not clear enough, I'll add some screenshots.

aledeg: There might be a better way to address that issue. I haven't dig into that.

Alkarex: Yes please for a screenshot, as I cannot reproduce the problem when keeping the original value of zero there:

image

aledeg: Screenshot_2021-02-26 My labels · FreshRSS
Here I can see 1 unread article and 2 read articles. I am expecting to see only 1 unread article when using the "Adjust showing" option.
This screenshot has been taken from the master branch.

Alkarex: Ah OK, the difference is visible when the option "Show all articles in favourites by default (Applies also on labels)" is disabled. When this option is enabled, it could stay to zero.  Let's leave it like that.

Alkarex:
For another PR, maybe when the option "Show all articles in favourites by default (Applies also on labels)" is enabled, we should also show all the labels, to avoid the empty drop-down menu (example with two labels, but without any unread articles):

image

PR #3480, opened 2021-02-26
Optimize theme Pafat for the mobile view

myuki:
It is better for the mobile view to delete the button's `min-width`.

Before:

Before

After:

After

### DIFF
--- a/p/themes/Pafat/pafat.css
+++ b/p/themes/Pafat/pafat.css
@@ -120,6 +120,7 @@ form th {
 .stick {
 	vertical-align: middle;
 	font-size: 0;
+	min-width: 215px;
 }
 
 .stick input,
@@ -1167,6 +1168,7 @@ a.btn {
 
 	.nav_menu .stick {
 		margin: 0 10px;
+		min-width: max-content;
 	}
 
 	.nav_menu .stick .btn {
--- a/p/themes/Pafat/pafat.css
+++ b/p/themes/Pafat/pafat.css
@@ -1168,7 +1168,7 @@ a.btn {
 
 	.nav_menu .stick {
 		margin: 0 10px;
-		min-width: max-content;
+		min-width: 0;
 	}
 
 	.nav_menu .stick .btn {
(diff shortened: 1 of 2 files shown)


Alkarex:
Thanks @Myuki 👍🏻 
I suggest you make the change in a `@media` section to only apply it when the width of the window is small.
Either in the existing one (see below), or in a new one to be applied even when the width is a bit larger.

https://github.com/FreshRSS/FreshRSS/blob/9ea3e7710ad54e76ff1f84bafb208da865816f93/p/themes/Pafat/pafat.css#L1131

Frenzie:
Around 1280 and 1600 are other common breakpoints, presumably based on typical device screensizes.

myuki:
> Thanks @Myuki 👍🏻
> I suggest you make the change in a `@media` section to only apply it when the width of the window is small.
> Either in the existing one (see below), or in a new one to be applied even when the width is a bit larger.
> 
> https://github.com/FreshRSS/FreshRSS/blob/9ea3e7710ad54e76ff1f84bafb208da865816f93/p/themes/Pafat/pafat.css#L1131

It's a better way. I will change later.

p/themes/Pafat/pafat.css
```diff
@@ -1168,6 +1168,7 @@ a.btn {
 
 	.nav_menu .stick {
 		margin: 0 10px;
+		min-width: max-content;
```
Alkarex:
```suggestion
		min-width: 0;
```

Should achieve the same, while being lighter and more back-compatible

p/themes/Pafat/pafat.rtl.css
```diff
@@ -1168,6 +1168,7 @@ a.btn {
 
 	.nav_menu .stick {
 		margin: 0 10px;
+		min-width: max-content;
```
Alkarex:
```suggestion
		min-width: 0;
```

Alkarex:
@Myuki If you could be motivated enough to look at some of the other themes for this specific problem, please do so :-)
Thanks again 👍🏻

myuki:
> 
> 
> @Myuki If you could be motivated enough to look at some of the other themes for this specific problem, please do so :-)
> Thanks again 👍🏻

I had checked other themes, no one use `min-width` as the same way. But some button is too big for mobile view.

Alkarex:
Ok, thanks!

PR #3488, opened 2021-02-27
Travis: Enforce phpcs line length + whitespace

Alkarex:
* Give up on allowing a mixture of tabs and spaces, in order to be able to easily enforce line length and whitespace rules.
See https://github.com/FreshRSS/FreshRSS/pull/2317#discussion_r583260405

* Report warnings in Travis, and not only errors

* Lines longer than lineLimit=165 are reported as a warning (this is the size of a GitHub window)
* Lines longer than absoluteLineLimit=190 are now an error (that works also for our .phtml template files)

* Fixed the lines that did not comply with those rules. Mainly some lines that were much too long, but also a some wrong whitespace

* Now also include our tests
* Incidently, works also for our .css files

Now we can run:

```sh
phpcs -p --extensions=phtml,php,css .
```

My motivation came from giving a try to https://marketplace.visualstudio.com/items?itemName=ikappas.phpcs that provides instant feedback in the editor.
Before this patch, there were warnings all over the place.


aledeg:
>     * Give up on allowing a mixture of tabs and spaces, in order to be able to easily enforce line length and whitespace rules.
>       See [#2317 (comment)](https://github.com/FreshRSS/FreshRSS/pull/2317#discussion_r583260405)

Yes :tada:

phpcs.xml
```diff
@@ -20,19 +20,15 @@
...
-	<!-- One line must never exceed 120 characters -->
+	<!-- Maximum line length -->
 	<rule ref="Generic.Files.LineLength">
 		<!-- For language strings maximum line lengths make little sense. -->
 		<exclude-pattern>./app/i18n/</exclude-pattern>
 		<!-- Don't enforce line length on the HTML; the point is to improve legibility, not reduce it -->
-		<exclude-pattern>./app/install.php</exclude-pattern>
-		<!-- @todo remove test exclusion -->
-		<exclude-pattern>./tests/app/</exclude-pattern>
+		<exclude-pattern>.phtml</exclude-pattern>
```
Alkarex: @Frenzie Do you know if it is possible to have different rules depending on the file extension?

app/Models/CategoryDAO.php
```diff
@@ -80,9 +80,13 @@ protected function autoUpdateDb($errorInfo) {
...
-		$sql = 'INSERT INTO `_category`(name, attributes) '
-		     . 'SELECT * FROM (SELECT TRIM(?) AS name, TRIM(?) AS attributes) c2 '	//TRIM() to provide a type hint as text
-		     . 'WHERE NOT EXISTS (SELECT 1 FROM `_tag` WHERE name = TRIM(?))';	//No tag of the same name
+		// TRIM() to provide a type hint as text
+		// No tag of the same name
+		$sql = <<<'SQL'
+INSERT INTO `_category`(name, attributes)
+SELECT * FROM (SELECT TRIM(?) AS name, TRIM(?) AS attributes) c2
+WHERE NOT EXISTS (SELECT 1 FROM `_tag` WHERE name = TRIM(?))
+SQL;
```
aledeg, line 89: This is so much clearer :+1:

app/Models/DatabaseDAOPGSQL.php
```diff
@@ -58,12 +58,12 @@ public function size($all = false) {
...
-			     . "pg_total_relation_size('`{$this->pdo->prefix()}entry`') + "
-			     . "pg_total_relation_size('`{$this->pdo->prefix()}entrytmp`') + "
-			     . "pg_total_relation_size('`{$this->pdo->prefix()}tag`') + "
-			     . "pg_total_relation_size('`{$this->pdo->prefix()}entrytag`')";
+				. "pg_total_relation_size('`{$this->pdo->prefix()}category`') + "
+				. "pg_total_relation_size('`{$this->pdo->prefix()}feed`') + "
+				. "pg_total_relation_size('`{$this->pdo->prefix()}entry`') + "
+				. "pg_total_relation_size('`{$this->pdo->prefix()}entrytmp`') + "
+				. "pg_total_relation_size('`{$this->pdo->prefix()}tag`') + "
+				. "pg_total_relation_size('`{$this->pdo->prefix()}entrytag`')";
```
aledeg: Maybe using the HEREDOC notation here would be useful since you already use the string templates.

app/Models/EntryDAO.php
```diff
@@ -578,14 +578,14 @@ public function cleanOldEntries($id_feed, $options = []) { //Remember to call up
 		if (!empty($options['keep_min']) && $options['keep_min'] > 0) {
 			//Double SELECT for MySQL workaround ERROR 1093 (HY000)
 			$sql .= ' AND `lastSeen` < (SELECT `lastSeen`'
-			      . ' FROM (SELECT e2.`lastSeen` FROM `_entry` e2 WHERE e2.id_feed = :id_feed2'
-			      . ' ORDER BY e2.`lastSeen` DESC LIMIT 1 OFFSET :keep_min) last_seen2)';
+				. ' FROM (SELECT e2.`lastSeen` FROM `_entry` e2 WHERE e2.id_feed = :id_feed2'
+				. ' ORDER BY e2.`lastSeen` DESC LIMIT 1 OFFSET :keep_min) last_seen2)';
```
aledeg, line 582: Here you could use the NOWDOC notation.

app/Models/EntryDAO.php
```diff
@@ -670,7 +670,8 @@ protected function sqlConcat($s1, $s2) {
 		return 'CONCAT(' . $s1 . ',' . $s2 . ')';	//MySQL
 	}
 
-	protected function sqlListEntriesWhere($alias = '', $filters = null, $state = FreshRSS_Entry::STATE_ALL, $order = 'DESC', $firstId = '', $date_min = 0) {
+	protected function sqlListEntriesWhere($alias = '', $filters = null, $state = FreshRSS_Entry::STATE_ALL,
+			$order = 'DESC', $firstId = '', $date_min = 0) {
```
aledeg, line 674: In case like that, don't you think that having one argument per line is more readable?
```php
protected function sqlListEntriesWhere(
	$alias = '',
	$filters = null,
	$state = FreshRSS_Entry::STATE_ALL,
	$order = 'DESC',
	$firstId = '',
	$date_min = 0
) {
```

app/Models/EntryDAOSQLite.php
```diff
@@ -114,13 +114,13 @@ protected function updateCacheUnreads($catId = false, $feedId = false) {
 	public function markRead($ids, $is_read = true) {
 		FreshRSS_UserDAO::touch();
 		if (is_array($ids)) {	//Many IDs at once (used by API)
-			if (true) {	//Speed heuristics	//TODO: Not implemented yet for SQLite (so always call IDs one by one)
+			//if (true) {	//Speed heuristics	//TODO: Not implemented yet for SQLite (so always call IDs one by one)
```
aledeg, line 117: I'd rather remove it completely than having some commented code.

app/Models/FormAuth.php
```diff
@@ -7,9 +7,9 @@ public static function checkCredentials($username, $hash, $nonce, $challenge) {
 				!ctype_graph($challenge) ||
 				!ctype_alnum($nonce)) {
 			Minz_Log::debug('Invalid credential parameters:' .
-			                ' user=' . $username .
-			                ' challenge=' . $challenge .
-			                ' nonce=' . $nonce);
+				' user=' . $username .
+				' challenge=' . $challenge .
+				' nonce=' . $nonce);
```
aledeg: See aforementioned comment

lib/Minz/ExtensionManager.php
```diff
@@ -163,8 +161,7 @@ public static function load($info) {
 
 		// Test if the given extension class exists.
 		if (!class_exists($ext_class_name)) {
-			Minz_Log::warning('`' . $ext_class_name .
-			                  '` cannot be found in `' . $entry_point_filename . '`');
+			Minz_Log::warning('`' . $ext_class_name . '` cannot be found in `' . $entry_point_filename . '`');
```
aledeg: String template?

lib/Minz/ExtensionManager.php
```diff
@@ -180,8 +177,7 @@ public static function load($info) {
 
 		// Test if class is correct.
 		if (!($extension instanceof Minz_Extension)) {
-			Minz_Log::warning('`' . $ext_class_name .
-			                  '` is not an instance of `Minz_Extension`');
+			Minz_Log::warning('`' . $ext_class_name . '` is not an instance of `Minz_Extension`');
```
aledeg: Same here?

aledeg (changes requested):
I am surprised that there is not more impacted files :)

app/Controllers/authController.php
```diff
@@ -160,13 +158,12 @@ public function formLoginAction() {
...
+				Minz_Request::good(_t('feedback.auth.login.success'), [ 'c' => 'index', 'a' => 'index' ]);
 			} else {
-				Minz_Log::warning('Password mismatch for' .
-				                  ' user=' . $username .
-				                  ', nonce=' . $nonce .
-				                  ', c=' . $challenge);
+				Minz_Log::warning('Password mismatch for'
+					. ' user=' . $username
+					. ', nonce=' . $nonce
+					. ', c=' . $challenge);
```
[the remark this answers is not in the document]
Alkarex: I would rather limit this PR to some more simple and direct changes (and I would like to focus on 1.18.0 release)
Indeed I have a few changes to the SQL strings, but I actually regret having done that as part of this PR :-P

By the way, I would be interested in a bit of analysis / benchmarks for the string templates. Last time I looked at it (several years ago), they were slower and more memory intensive than concatenation of simple strings. In any case, I do agree that they have at least other advantages, especially for translation.

app/Models/CategoryDAO.php
Alkarex: Yes - but I actually regret having done that as part of this PR. Small detail...

app/Models/DatabaseDAOPGSQL.php
Alkarex: Indeed, but let's continue addressing them in another PR, not to mix too many things :-)

phpcs.xml
Frenzie: Aren't you just missing a `*` or two?

Frenzie: I.e., either `*.phtml` or `*/*.phtml`.

app/Controllers/authController.php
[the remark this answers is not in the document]
Frenzie: I'd expect concat to be slightly faster on synthetic benchmarks but in actual practice I think the ease of use is quite likely to be worth a slight loss of efficiency on hundreds of thousands of iterations. ;-)

app/Models/EntryDAO.php
Alkarex: Potentially yes for here, although I am not so found of excessive lines in general, as it generates more vertical scrolling, which reduces my ability to get the bigger picture of a piece of code.
In any case, the ambition of this PR (which is already large :-)) was *just* to fix the biggest problems - there is still room for improvement

app/Controllers/authController.php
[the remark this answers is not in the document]
aledeg: I do not have benchmarks for string templates but I went to a conference where the author of that modification explained that the string concatenation generate a lot of overhead. Here is an article https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html

app/Models/EntryDAOSQLite.php
Alkarex: Yes indeed, but in this specific case, it is to match the construct of the parent code and make the TODO easier to understand & address:
https://github.com/FreshRSS/FreshRSS/blob/064288ed6228d3d56b2231aadd507eb62ae8662c/app/Models/EntryDAO.php#L338

app/Controllers/authController.php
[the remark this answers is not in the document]
Frenzie: Thanks for the link. I wasn't aware of this change in performance.

app/Models/FormAuth.php
Alkarex: That is indeed an obvious case :-)

app/Controllers/authController.php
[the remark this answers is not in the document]
Alkarex: I will do a few of the most obvious ones of this PR

app/Models/DatabaseDAOPGSQL.php
Alkarex: Will do here after all :-)

Alkarex:
Here you are @aledeg https://github.com/FreshRSS/FreshRSS/pull/3488/commits/c86067c0de096057fa1e494d5fa947e93e09b81f

app/Controllers/authController.php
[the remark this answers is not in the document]
Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/3488/commits/c86067c0de096057fa1e494d5fa947e93e09b81f

app/Models/DatabaseDAOPGSQL.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3488/commits/c86067c0de096057fa1e494d5fa947e93e09b81f

app/Models/FormAuth.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3488/commits/c86067c0de096057fa1e494d5fa947e93e09b81f

app/Models/EntryDAO.php
Alkarex: Could be, but it is only a part of a longer string construction

aledeg (approved):

phpcs.xml
Alkarex: Ah right. Seems to work though :-O

Alkarex: By the way, can we specify which file extensions to scan inside this file? e.g. https://github.com/squizlabs/PHP_CodeSniffer/issues/2486

Alkarex: Fixed https://github.com/FreshRSS/FreshRSS/pull/3488/commits/6740c5cbbb894d20f6f9422af33ec245c311baac

Frenzie: Might need a `$` at the end for correctness?

Alkarex:
I realised that we can include `.js` files more completely https://github.com/FreshRSS/FreshRSS/pull/3488/commits/02add3419277757a80f6785959e2389c2ab1ddbf

phpcs.xml
Alkarex: Done in https://github.com/FreshRSS/FreshRSS/commit/02add3419277757a80f6785959e2389c2ab1ddbf with a few more changes, in particular for `.js`

Makefile
```diff
@@ -61,11 +61,11 @@ test: bin/phpunit ## Run the test suite
 
 .PHONY: lint
 lint: bin/phpcs ## Run the linter on the PHP files
-	$(PHP) ./bin/phpcs . --standard=phpcs.xml --warning-severity=0 --extensions=php -p
+	$(PHP) ./bin/phpcs . -p -s
```
Alkarex, line 64: @aledeg Does that look good to you for the Makefile?

Alkarex: Now file extensions are defined in `phpcs.xml`
`-s` is to include the name of broken rules in the outputs

PR #3491, opened 2021-02-28
Optimize some themes for mobile view

myuki:
Before:

Before 1

After: 

After 1

---

Before:

Before 2

After: 

After 2


Alkarex:
Thanks, much appreciated 👍🏻

PR #3459, opened 2021-02-18
Update Travis.yaml to support newer PHP versions

BrookeDot:
Closes #3458

Changes proposed in this pull request:
- Updates Travis to add PHP 7.4 and 8.0
- Allow failures on PHP 8 (for now)
- Upgrade PHPUnit to 9.0
- Removes PHP 7.1 and 7.3 from Tests.
-   Updates the Class name for `CategoryTest.php` , `passwordUtilTest.php`, and `MigratorTest.php` to fix Unit test WARNINGs

How to test the feature manually:

I'm fairly new to Travis so I'm not sure, hoping there is a way to test the config without committing this branch. 

Pull request checklist:

- [x] clear commit messages
- [ ] code manually tested
- [ ] unit tests written (optional if too hard)
- [x] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/master/docs/en/developers/04_Pull_requests.md).

PR #3492, opened 2021-02-28
Travis remove superfluous PHP version

Alkarex:
Now that Travis tests with PHP8 are all working fine, and as we have PHP 5.6 with the sames tests for our oldest supported version, let's save resources and remove intermediate versions.
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/3459#issuecomment-781974573

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,18 +18,6 @@ jobs:
         - COMPOSER_BIN=$(composer global config --absolute bin-dir)
         - $COMPOSER_BIN/phpunit --bootstrap ./tests/bootstrap.php --verbose ./tests
         - $COMPOSER_BIN/phpcs . -p -s
-    - name: "PHP 7.4 Syntax, linter and tests"
-      php: "7.4"
-      install:
-        - composer global require squizlabs/php_codesniffer
-        - composer global require phpunit/phpunit ^9
-      script:
-        - phpenv rehash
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

aledeg:
@Alkarex You were mentioning that PHP 5.6 will be dropped soon. Which minimum version do you think we should support? Is it really worth it to drop the 7.4 tests if we need to add it back in a few weeks?
Just asking!

Frenzie:
@aledeg In that case we're just talking about changing the one line `php: "5.6"` to some other version for syntax only. Unless you think the tests themselves should be duplicated.

aledeg:
> @aledeg In that case we're just talking about changing the one line `php: "5.6"` to some other version for syntax only. Unless you think the tests themselves should be duplicated.

I don't think that the tests should be duplicated nor that this PR is irrelevant. I was just wondering.

Alkarex:
I have just checked again different metrics (similar than https://github.com/FreshRSS/FreshRSS/discussions/3321 , dropped by 1%-point since last month) and since there is not much new on WordPress' side (which I would otherwise have liked to follow), I suggest we schedule dropping PHP5 for FreshRSS 1.19.0 (so let's wait a bit more after the release of 1.18.0 in the likely case we will have minor 1.18.x versions), and then move to 7.0 as the oldest supported version. Let's continue in https://github.com/FreshRSS/FreshRSS/discussions/3321 :-)

PR #3494, opened 2021-03-03
Add an option to set a custom user agent to retrieve individual feeds

### LINKED ISSUE #3493: [Feature] Allow setting custom user-agent when retrieving particular feeds
Many websites use Google's FeedBurner to serve their public feeds. This allows Google to track you when you use the feed, and also makes it impossible to fetch full article content with FreshRSS.

Publishers create FeedBurner feeds from normal RSS feeds. However, in many cases the original feed is only accessible with a user-agent of `FeedBurner` (https://support.google.com/feedburner/answer/78464). Adding an option to have FreshRSS request these feeds using a `FeedBurner` user agent would allow users to read the original version of these feeds in FreshRSS.

Jules-Bertholet:
This allows retrieving the original RSS feed when websites use services like FeedBurner.

Closes #3493 

- Add a setting to choose the user agent used when fetching feeds

1. Add a feed that is fronted by FeedBurner to FreshRSS. One such is `https://www.themoneyillusion.com/feed/atom`
2. Set the user agent to `FeedBurner`
3. Ensure the feed location still points to `https://www.themoneyillusion.com/feed/atom`, and was not changed to the FeedBurner feed because of a 3XX redirect before you added the user-agent
4. Re-fetch feed content
5. All feed items should now refer to the original site, not `feedproxy.google.com` or similar

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


app/i18n/en/sub.php
```diff
@@ -90,6 +90,8 @@
 		'title_add' => 'Add an RSS feed',
 		'ttl' => 'Do not automatically refresh more often than',
 		'url' => 'Feed URL',
+		'useragent' => 'Set the user agent for fetching this feed',
+		'useragent_help' => 'Example: <kbd>FeedBurner</kbd>',
```
Alkarex: Please use another example, avoiding specific product names, but a more general browser, e.g.

```
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:86.0)
```

Alkarex:
Please add a line for your new string in `./cli/i18n/ignore/en-us.php` to ignore the fact that it is the same string in en-GB and en-US

app/Controllers/subscriptionController.php
```diff
@@ -123,6 +124,9 @@ public function feedAction() {
 			if ($cookie != '') {
 				$opts[CURLOPT_COOKIE] = $cookie;
 			}
+			if ($useragent != '') {
```
aledeg: Please, use `!==` instead of `!=`.

aledeg (changes requested):
Could you add the translating strings from en_US in the ignore file?
Run this command to do that: `make i18n-ignore-unmodified-keys lang=en-us`

*Edit* This will fix the CI error!

app/Controllers/subscriptionController.php
aledeg: I know that the code around does not use it, but I think it should. We will fix it when working on it.

aledeg (approved):

Alkarex:
Works fine. But maybe this option should be made available from the *Add feed* page (`?c=subscription&a=add`) as well? Otherwise it might not be possible to add a feed, which requires a specific user agent. Around there https://github.com/FreshRSS/FreshRSS/blob/master/app/views/subscription/add.phtml#L62 and there https://github.com/FreshRSS/FreshRSS/blob/master/app/Controllers/subscriptionController.php#L130

Furthermore, we should either only allow those options for admins, or make them use a user-specific cache, as this might otherwise randomly affect other users of the same FreshRSS instance. See https://github.com/FreshRSS/FreshRSS/pull/3367#issuecomment-766250249 This should be in another PR though, but must be done before the release of 1.18.0.

Alkarex:
Patch to avoid SimplePie cache pollution when changing cURL options: https://github.com/FreshRSS/FreshRSS/pull/3502

Alkarex:
@Jules-Bertholet I have added the user-agent (and proxy) option to the subscription page to add a new feed:
https://github.com/FreshRSS/FreshRSS/pull/3494/commits/f9495f5c97f4386c3c72b99b6e375db6122a9775

image

Could you please give it a quick test?

Jules-Bertholet:
@Alkarex Works great! (Proxy not tested)

Frenzie:
Looks nice :-)

Alkarex:
@Jules-Bertholet Could you please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/master/CREDITS.md ?

PR #3500, opened 2021-03-06
fix updateEntryPrepared

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/3461/files#r588928471
Before, there was a risk that the prepared statement would cache a wrong behaviour for an article's read status.


Alkarex:
Tested on PostgreSQL, MySQL, SQLite, with the *mark as unread modified articles* option enabled or disabled, in PHP 5.6 and PHP 8.

PR #3502, opened 2021-03-06
SimplePie prevent cache pollution

Alkarex:
Retrieving a feed with distinct cURL parameters such as user-agent might give different responses, which should not share the same cache:
#fix https://github.com/FreshRSS/FreshRSS/pull/3367#issuecomment-766250249
#fix https://github.com/FreshRSS/FreshRSS/pull/3494#issuecomment-790113663


lib/lib_rss.php
```diff
@@ -173,6 +173,7 @@ function customSimplePie($attributes = array()) {
 	$simplePie = new SimplePie();
 	$simplePie->set_useragent(FRESHRSS_USERAGENT);
 	$simplePie->set_syslog(FreshRSS_Context::$system_conf->simplepie_syslog_enabled);
+	$simplePie->set_cache_name_function('sha1');
```
Alkarex, line 176: Slight upgrade from `md5` (no need to be crypto-safe, but good to reduce risks of collision)

lib/SimplePie/SimplePie.php
```diff
@@ -920,6 +920,22 @@ public function set_cache_location($location = './cache')
 		$this->cache_location = (string) $location;
 	}
 
+	/**
+	 * Return the filename (i.e. hash, without path and without extension) of the file to cache a given URL.
+	 */
+	public function get_cache_filename($url)
+	{
+		// Append custom parameters to the URL to avoid cache polution in case of multiple calls with different parameters.
```
Alkarex:
```suggestion
		// Append custom parameters to the URL to avoid cache pollution in case of multiple calls with different parameters.
```

Alkarex:
Let's merge to get some testing

Alkarex:
Upstream PR: https://github.com/simplepie/simplepie/pull/675

PR #3506, opened 2021-03-07
Reload full content when changed

Alkarex:
If an article is changed, reload also its full content when applicable.
And fix a related bug: the hash to detect changes must be computed before loading the full content.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -377,6 +377,9 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 								continue;
 							}
 
+							// If the entry has changed, there is a good chance for the full content to haved changed as well.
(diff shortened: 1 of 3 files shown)


Alkarex:
Bug discovered when fixing https://github.com/FreshRSS/FreshRSS/pull/3500

app/Controllers/feedController.php
```diff
@@ -377,6 +377,9 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 								continue;
 							}
 
+							// If the entry has changed, there is a good chance for the full content to haved changed as well.
```
```suggestion
							// If the entry has changed, there is a good chance for the full content to have changed as well.
```

Alkarex:
Let's merge to get some testing

PR #3524, opened 2021-03-14
GitHub action latest tag

Alkarex:
Automatically tag the latest release with the `latest` git tag
https://github.com/marketplace/actions/latest-tag

### DIFF
--- a/.github/workflows/latest.yml
+++ b/.github/workflows/latest.yml
@@ -1,0 +1,12 @@
+name: Add latest tag to new release
+on:
+  release:
+    types: [published]
+
+jobs:
+  run:
+    name: Run local action
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run latest-tag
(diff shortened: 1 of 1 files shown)


Alkarex:
P.S. I am not entirely sure of the syntax :-P

.github/workflows/latest.yml
```diff
@@ -0,0 +1,12 @@
...
+  release:
+    types: [published]
+
+jobs:
+  run:
+    name: Run local action
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run latest-tag
+        uses: EndBug/latest-tag@6d22a6738f5c33059e3a8c6ca5dcf8eaf8a14599
```
Frenzie, line 12: They don't have a version?

Alkarex: Yes they have, but this one is more precise (like when using a library with version pinning) - I am not entirely sure of what those marketplace actions can do if they are updated to something evil :-)

Alkarex:
Let's merge it, so we can see whether it works for 1.18.1

PR #3548, opened 2021-03-20
Remove shortcut validation

### LINKED ISSUE #3481: Shortcuts: unicode characters not allowed
Hi,
I’m using the French bépo keyboard layout and it seems keys like "é" or ".", which are easily accessible, cannot be assigned as shortcuts: when I validate the shortcuts after setting them, they revert to the default value.
Is there a way such characters could be considered as valid by FreshRSS?
Thanks!

aledeg:
Closes #3481

- Remove shortcut validation

1. configure a non-standard shortcut key
2. save it to see it stored in the configuration
3. open the feed page and use it

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, only standard shortcuts were supported. When other shortcuts were
configured, they were filtered out.
Now, those shortcuts are stored in the configuration and an alert message
is displayed to alert the user that he is using non-standard shortcuts.

### DIFF
--- a/app/i18n/cz/conf.php
+++ b/app/i18n/cz/conf.php
@@ -191,6 +191,7 @@ return array(
 		'navigation_help' => 'Pomocí přepínače <kbd>⇧ Shift</kbd> fungují navigační zkratky v rámci kanálů.<br/>Pomocí přepínače <kbd>Alt ⎇</kbd> fungují v rámci kategorií.',
 		'navigation_no_mod_help' => 'The following navigation shortcuts do not support modifiers.',	// TODO - Translation
 		'next_article' => 'Skočit na další článek',
+		'non_standard' => 'Some keys (<kbd>%s</kbd>) may not work as shortcuts.',	// TODO - Translation
 		'normal_view' => 'Switch to normal view',	// TODO - Translation
 		'other_action' => 'Ostatní akce',
 		'previous_article' => 'Skočit na předchozí článek',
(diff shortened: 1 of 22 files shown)


Alkarex (approved):

Alkarex:
What could be the unexpected behaviours?
It is fine, but I am wondering whether the list of standard shortcuts is so useful. An alternative would be to show the non-standard shortcuts in use.

aledeg:
> What could be the unexpected behaviours?

If you try to save a quote (") for instance, the shortcut is useless because it is converted to an html entity. We can argue that we can change that behavior but it seems that it could be dangerous since it is injected in the JS.

> It is fine, but I am wondering whether the list of standard shortcuts is so useful. An alternative would be to show the non-standard shortcuts in use.

That's something we can do!

app/i18n/cz/conf.php
```diff
@@ -191,6 +191,7 @@
 		'navigation_help' => 'Pomocí přepínače <kbd>⇧ Shift</kbd> fungují navigační zkratky v rámci kanálů.<br/>Pomocí přepínače <kbd>Alt ⎇</kbd> fungují v rámci kategorií.',
 		'navigation_no_mod_help' => 'The following navigation shortcuts do not support modifiers.',	// TODO - Translation
 		'next_article' => 'Skočit na další článek',
+		'non_standard' => 'You are using non-standard shortcut keys. Be aware that you might experience unexpected behaviors. The standard shortcut keys are the following:',	// TODO - Translation
```
Frenzie: Since the unexpected behavior is just a harmless not working, I'd strongly reword it so as not to sound frightening:
```suggestion
		'non_standard' => 'Some keys may not work as shortcuts. The standard shortcut keys are the following:',	// TODO - Translation
```

Frenzie:
> If you try to save a quote (") for instance, the shortcut is useless because it is converted to an html entity. We can argue that we can change that behavior but it seems that it could be dangerous since it is injected in the JS.

Injected how?

aledeg:
The keys are injected in the context in the `app/views/helpers/javascript_vars.phtml` file. It might break something even though I am not sure how.

aledeg:
I've made a few changes. Here is how it looks now
Screenshot_2021-03-21 Shortcuts · FreshRSS

PR #3555, opened 2021-03-23
tec: Remove data/do-install.txt

marienfressinaud:

- Remove file data/do-install.txt and its references
- Remove deleteInstall() method
- Change install process to start it if data/applied_migrations.txt is missing
- Update documentation to remove the step asking to delete data/do-install.txt

1. git pull new code
2. check data/do-install.txt is still missing
3. check FRSS doesn't offer to be reinstalled, even after `touch data/do-install.txt`
4. remove data/applied_migrations.txt
5. check FRSS offers to be reinstalled
6. run `rm data/do-install.txt` and `touch data/applied_migrations.txt` to be in initial state

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] unit tests written (optional if too hard) N/A
- [x] documentation updated

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -44,7 +44,6 @@ class FreshRSS_update_Controller extends Minz_ActionController {
 			$return = 1;
 		}
 		chdir($cwd);
-		deleteInstall();
 		$line = is_array($output) ? implode('; ', $output) : '' . $output;
 		return $return == 0 ? true : 'Git error: ' . $line;
 	}
--- a/app/i18n/cz/install.php
+++ b/app/i18n/cz/install.php
@@ -111,7 +111,7 @@ return array(
 		'choose' => 'Vyberte jazyk FreshRSS',
 		'defined' => 'Jazyk byl nastaven.',
 	),
-	'not_deleted' => 'Nastala chyba, soubor <em>%s</em> musíte smazat ručně.',
+	'missing_applied_migrations' => 'Something went wrong; you should create an empty file <em>%s</em> manually.',	// TODO - Translation
 	'ok' => 'Instalace byla úspěšná.',
 	'session' => array(
 		'nok' => 'The web server seems to be incorrectly configured for cookies required for PHP sessions!',	// TODO - Translation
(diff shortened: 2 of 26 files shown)


Alkarex (approved):

pthall:
Hi, I am hosting FreshRSS with Heroku, which wipes local filesystem changes on a regular basis.

After upgrading FreshRSS from 1.17 to 1.18, the installation screen appeared. And whenever Heroku did its periodic filesystem wipe, the installation screen returned.

Reading the changes to `do-install.txt` and `applied_migrations.txt`, it's unclear which states of these files triggers the installation screen in 1.18.

To avoid the installation screen from re-appearing repeatedly, I removed the `applied_migrations.txt` line from `data/.gitignore` and `chmod 444 data/applied_migrations.txt`. I didn't dig far enough to figure out whether there is some FreshRSS code intentionally deleting `data/applied_migrations.txt`, or if `applied_migrations.txt` in `data/.gitignore` meant that `applied_migrations.txt` was never created from `git push heroku master`.

Just want to note this in case others have similar web hosts with only short-lived filesystem persistence. The old `do-install.txt` was confusing in Heroku, but `applied_migrations.txt` is potentially more confusing.

marienfressinaud:
@pthall Heroku should not touch `data/*` folder which stores important files. In particular, the users' data are here so I'm wondering why your user isn't deleted :thinking:

pthall:
Everything persisted under `data/` is committed to Git and pushed to Heroku. Heroku is definitely not ideal for FreshRSS hosting.

marienfressinaud:
Indeed, if you have to commit the files under `data/` to Git, you should avoid Heroku to deploy FreshRSS :s A better solution would be to mount an external persistent drive on `data/`, but I don't know if it's possible with Heroku.

Just to clarify how the new `applied_migrations.txt` file works (`do-install.txt` no longer exists):

- installation is triggered if the file is missing (while it was the opposite with `do-install.txt`)
- migrations under `app/migrations/` are executed and added to `applied_migrations.txt` if there are not in it yet. **So you should save the state of this file to avoid executing migrations several times.** For now it's empty because there are no migrations

Issue #3552, opened 2021-03-22
SQL error updateTTL when importing tt-rss opml [BUG]


marrco:
**Describe the bug**
Newly installed 1.80 via cPanel, imported my (less than 100) feeds i prevoiusly exported via an outdated (cPanel) tt-rss instance

feeds don't automatically update. Single feeds when forced update do show new items. Trying to force a complete feed refresh doesn't work

looking at logs there is an error:

2021-03-22 10:23:59
SQL error updateTTL 2: Duplicate column name 'ttl' ALTER TABLE `_feed` ADD COLUMN ttl INT NOT NULL DEFAULT 0
2021-03-22 10:23:59
SQL warning updateTTL 1: Deadlock found when trying to get lock; try restarting transaction UPDATE `_feed` SET ttl=:new_value WHERE ttl=:old_value
2021-03-22 10:23:58
SQL error updateTTL 2: Duplicate column name 'ttl' ALTER TABLE `_feed` ADD COLUMN ttl INT NOT NULL DEFAULT 0
2021-03-22 10:23:58
SQL warning updateTTL 1: Deadlock found when trying to get lock; try restarting transaction UPDATE `_feed` SET ttl=:new_value WHERE ttl=:old_value

**Environment information (please complete the following information):**
cPanel

**Additional context**
I suspect there's an issue with updating imported tt-rss exported feeds.

Alkarex:
I will fix. Would you mind sharing the file (or a subset) you are trying to import?
Either here or [by e-mail](https://alexandre.alapetite.fr/cv/)

marrco:
Thx,
 mail sent. Let me know if you need an account.

clicking on the update button it stalls, each time on a different feed.

[https://i.imgur.com/gB3C50S.png](https://i.imgur.com/gB3C50S.png)

Alkarex:
@marrco What database are you using? Which versions (OS, database, FreshRSS)?
Would you be able to try this branch? https://github.com/FreshRSS/FreshRSS/pull/3553
I recommend making a new FreshRSS user to test the import again, or restart from a fresh install + database

marrco:
Your current version of FreshRSS is 1.18.0.
Your PHP version (7.4.15) is compatible with FreshRSS.
installation check is all green, but it's a cPanel account, so I have no access to the underlying system

cPanel Version | 92.0 (build 12)
-- | --
Apache Version | 2.4.46
PHP Version | 7.4.11
[...]

PR #1470, opened 2017-03-25
Implementation of defered insertion

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/530

PR #3558, opened 2021-03-23
Clean 3+ years old auto-updates

Alkarex:
* Minor auto-update in version 1.10.0 (3 years old) https://github.com/FreshRSS/FreshRSS/pull/1750
    * Causing slight problems, related to https://github.com/FreshRSS/FreshRSS/issues/3552
* Auto-update for temporary table in version 1.7.0 (4 years old) https://github.com/FreshRSS/FreshRSS/pull/1470
    * Just because it was older than the one just above

### DIFF
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -148,7 +148,7 @@ SQL;
 		} else {
 			$info = $stm == null ? $this->pdo->errorInfo() : $stm->errorInfo();
 			if ($this->autoUpdateDb($info)) {
-				return $this->updateCategory($valuesTmp);
+				return $this->updateCategory($id, $valuesTmp);
 			}
 			Minz_Log::error('SQL error updateCategory: ' . json_encode($info));
 			return false;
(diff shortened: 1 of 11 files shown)


app/Models/CategoryDAO.php
```diff
@@ -148,7 +148,7 @@ public function updateCategory($id, $valuesTmp) {
 		} else {
 			$info = $stm == null ? $this->pdo->errorInfo() : $stm->errorInfo();
 			if ($this->autoUpdateDb($info)) {
-				return $this->updateCategory($valuesTmp);
+				return $this->updateCategory($id, $valuesTmp);
```
Alkarex, line 151: And a small auto-update bug fix at the same time

Frenzie (commented):
I forget, can you go 1.7 → 1.8 → 1.9 or does it always try to go straight to the latest

Alkarex:
To the latest @Frenzie . And actually, I am not sure I like this PR after all, especially for a .1 release. Will do something else instead

Alkarex:
Revived in https://github.com/FreshRSS/FreshRSS/pull/5649

Issue #2723, opened 2019-12-21
XAMPP/macOS: Feeds Do Not Populate on Fresh Install?


henrythefifth2:
Running XAMPP 7.3.7 on Mac OSCatalina 10.15.2

Fresh install of FreshRSS 1.15.4-dev.

I've imported 34 RSS feeds.  No error messages.  Feeds to not populate on manual refresh.  

Help with a fix much appreciated.

PHP installation
Your PHP version is 7.3.7, which is compatible with FreshRSS.

You have the Minz framework.

You have the cURL library.

You have PDO and at least one of the supported drivers (pdo_mysql, pdo_sqlite, pdo_pgsql).

You have the required library for regular expressions (PCRE).

You have the required library for character type checking (ctype).

You have the fileinfo library.

You have the required library to browse the DOM.

You have JSON extension.

You have the recommended library mbstring for Unicode.

You have ZIP extension.

File installation
Permissions on data directory are good.

Permissions on cache directory are good.

Permissions on users directory are good.

Permissions on favicons directory are good.

Permissions on tokens directory are good.

Alkarex:
Hello,
What database are you using, and how did you import your feeds (OPML?).
Do you have anything in your browser logs, Apache logs, or FreshRSS logs?

henrythefifth2:
Thanks for replying!

Running MariaDB-10.3.16

Imported OPML feeds directly to application.

=========================

Chrome Console Output on Refresh:

main.js?1576946233:1226 [Violation] Only request notification permission in response to a user gesture.
notifs_html5_ask_permission @ main.js?1576946233:1226
init_notifs_html5 @ main.js?1576946233:1262
init_afterDOM @ main.js?1576946233:1492
(anonymous) @ main.js?1576946233:1505
main.js?1576946233:1226 [Deprecation] The Notification API may no longer be used from insecure origins. You should consider switching your application to a secure origin, such as HTTPS. See https://goo.gl/rStTGz for more details.
notifs_html5_ask_permission @ main.js?1576946233:1226
init_notifs_html5 @ main.js?1576946233:1262
init_afterDOM @ main.js?1576946233:1492
(anonymous) @ main.js?1576946233:1505
main.js?1576946233:1498 FreshRSS main init done.

=========================

Apache Access Log

192.168.0.5 - - [21/Dec/2019:12:22:18 -0500] "GET /ourpage/FreshRSS/p/i/ HTTP/1.1" 304 -
[...]

PR #3577, opened 2021-04-08
Friendly syntax for Intellisense

Alkarex:
Dynamic code declaration is not well detected, especially for constants, so change to more standard syntax, which is correctly detected by Intellisense / Intelephense.

Furthermore, this custom `safe_define()` function has been confusing for some users, e.g. https://github.com/FreshRSS/FreshRSS/issues/2723#issuecomment-753318513

Before the fix:
image


Frenzie (approved):
I find the old one a touch more elegant, but it makes no real difference.

PR #3578, opened 2021-04-08
Fix wrong SimplePie type hint

Alkarex:
`SimplePie_Cache` also exists but is something else.
This created errors in IDE.

image


Alkarex:
Upstream PR: https://github.com/simplepie/simplepie/pull/678

PR #3589, opened 2021-04-17
git update auto change to edge branch

Alkarex:
For existing installations using automatic git update, checkout *edge* branch if it was still using *master* or *dev*.


app/Controllers/updateController.php
```diff
@@ -11,31 +11,54 @@ public static function hasGitUpdate() {
...
 		$return = 1;
 		try {
-			exec('git fetch', $output, $return);
+			exec('git fetch --prune', $output, $return);
 			if ($return == 0) {
+				unset($output);
 				exec('git reset --hard FETCH_HEAD', $output, $return);
 			}
+
+			// Automatic change to the new name of edge branch since FreshRSS 1.18.0
```
marienfressinaud: This can be put in a [migration](https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/developers/Minz/migrations.md) instead to avoid more complicated code here

marienfressinaud: Otherwise, it worked for me :+1:

Alkarex: Ah yes, I am not familiar with this part of the code yet...

marienfressinaud: Something like that should work (not tested):

```php
// File: app/migrations/2021_04_24_GitCheckoutFromMasterToEdge.php
class FreshRSS_Migration_2021_04_24_GitCheckoutFromMasterToEdge {
	public static function migrate() {
		if (!is_dir(FRESHRSS_PATH . '/.git/')) {
			return true; // not a Git installation, nothing to do
		}

		exec('git branch --show-current', $output, $return);
		if ($return != 0) {
			throw new Exception('Can’t checkout to edge branch, please change branch manually.');
		}

		$line = is_array($output) ? implode('', $output) : $output;
		if ($line !== 'master' && $line !== 'dev') {
			return true; // not on master or dev, nothing to do
		}

		unset($output);
		exec('git checkout edge --guess -f', $output, $return);
		if ($return != 0) {
			throw new Exception('Can’t checkout to edge branch, please change branch manually.');
		}

		unset($output);
		exec('git reset --hard FETCH_HEAD', $output, $return);
		if ($return != 0) {
			throw new Exception('Can’t checkout to edge branch, please change branch manually.');
		}

		return true;
	}
}
```

Alkarex: Here you are https://github.com/FreshRSS/FreshRSS/pull/3589/commits/e09a1e9a1558c4cbfd62cd3f7032ef9cb917f7b3

Alkarex: @marienfressinaud It does not look like the migration errors / results / messages are visible in FreshRSS logs though. Only the result of the `git` command is visible in the Web server logs.

Alkarex: I also have a case of white page when using Migration

Alkarex:
Before we would get a white page in case of Migration errors. After https://github.com/FreshRSS/FreshRSS/pull/3589/commits/a0be21d834292bc94cea3359bed90bce13bb9c89 we get:

image

Alkarex:
Ping @marienfressinaud

lib/lib_rss.php
```diff
@@ -548,15 +548,27 @@ function getNonStandardShortcuts($shortcuts) {
...
+	$details = '';
+	// Prevent empty tags by checking if error isn not empty first
+	if ($error) {
 		$error = htmlspecialchars($error, ENT_NOQUOTES, 'UTF-8');
-		$error = "<h2>{$error}</h2>";
+
+		// First line is the main message
+		$message = strtok($error, "\n");
+		// Other lines are the details
+		$details = strtok('');
```
marienfressinaud: I didn't know about `strtok`, it has a very strange behaviour… I think `list($message, $details) = explode("\n", $error, 2)` is the same and a bit more clear (?)

marienfressinaud (approved):

lib/lib_rss.php
Alkarex: I do not mind. I do not think it is more clear, but rather might be a question of habits :-)

Alkarex:
I am not super convinced of using this migration system compared to the first approach inside the update controller. The migration system has a high likelihood of breaking an installation, cannot be skipped / postponed, cannot be re-applied when the situation becomes relevant again, reduces performances at each query.
I will try to add a few more tests for the obvious problems for which the migration system bricks FreshRSS, such as when git does not exist, or when the access rights do not allow the Web server to update automatically, but then I do not know what to do because I do not want to test for those things at each request, and the migration has not passed.

Alkarex:
Arg, another problem: we need to check `Minz_Configuration::get('system')->disable_update` but it is not yet available at the time the Migration system runs.
For instance, it currently breaks Docker deployments mounting the FreshRSS code as a volume.
@marienfressinaud I think I will move the logic back to the update controller, except if you see an easy path with the migration system?

lib/lib_rss.php
Alkarex: Changed https://github.com/FreshRSS/FreshRSS/pull/3589/commits/ee0c8e872134ea81ca46d4f629ae83521de296d7

marienfressinaud:
> I am not super convinced of using this migration system compared to the first approach inside the update controller. The migration system has a high likelihood of breaking an installation, cannot be skipped / postponed, cannot be re-applied when the situation becomes relevant again, reduces performances at each query.

The migration system is here to simplify the complexity of code by moving "update" part in a dedicated part of the application. Some parts are hard to understand and to maintain because they mix concepts.

I don't see why having this code in migrations increase the risk of breaking an installation. It's applied only once, just after getting the new code so it's quite similar to what you wrote first.

But I agree it would be better to don't have the migrations running on each request. Ideally it would be a separated command during update (could be applied automatically with the auto-update system, but should be manual otherwise).

> Arg, another problem: we need to check Minz_Configuration::get('system')->disable_update but it is not yet available at the time the Migration system runs.
> For instance, it currently breaks Docker deployments mounting the FreshRSS code as a volume.

Ah right, I didn't thought about that, but I don't understand why it would break such deployment. Actually I would think checking `disable_update` is not required since it should be applied after getting the new code, but I can miss something!

> @marienfressinaud I think I will move the logic back to the update controller, except if you see an easy path with the migration system?

If it makes your life easier and you feel more confident with this solution, go ahead! I maintain it makes the code harder to understand/maintain, but you spend way more time maintaining it than me so don't worry too much about me ^^

Alkarex:
@marienfressinaud I do like the idea of having a better and more uniform migration system. The key difference here is that if the new migration system fails, the whole FreshRSS installation is bricked until it is fixed (see screenshots above), compared with the approach inside the upgrade controller which will most likely just fail without consequence (and ability to easily retry when wanted) or at worst making the auto-upgrade to fail.
So I indeed do not really dare using the new upgrade mechanism for the next minor release for something as minor as a change of git branch. But this PR also contributes to making the new migration system more mature, e.g. fixing some white screens and logging issues.

PR #3612, opened 2021-05-07
Make SSL options available to non-admins

### LINKED ISSUE #3611: [Feature] To make some options available to non-admin accounts
There are settings, which are hidden for non-admin accounts. For example, "disable SSL verification" in the advanced feed options.
It will be cool to manage such features for all users.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3611
Should be OK thanks to the patch preventing feed cache pollution https://github.com/FreshRSS/FreshRSS/pull/3502


Frenzie (approved):
If you're sure it should work properly. :-)

PR #3615, opened 2021-05-07
Fix several comments syntaxes

Alkarex:
Mainly wrong `@return` types in comments

### DIFF
--- a/app/Models/Share.php
+++ b/app/Models/Share.php
@@ -51,7 +51,7 @@ class FreshRSS_Share {
 
 	/**
 	 * Return the list of sharing options.
-	 * @return an array of FreshRSS_Share objects.
+	 * @return array[FreshRSS_Share]
 	 */
 	public static function enum() {
 		return self::$list_sharing;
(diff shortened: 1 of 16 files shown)


app/Models/StatsDAO.php
```diff
@@ -238,9 +237,7 @@ protected function initStatsArray($min, $max) {
 
 	/**
 	 * Calculates feed count per category.
-	 * Returns the result as a JSON object.
-	 *
-	 * @return JSON object
+	 * @return array
```
Frenzie, line 240: There's no JSON involved at all?

Frenzie (approved):
Thanks!

app/Models/StatsDAO.php
Alkarex: No, it is just an array

app/Models/Share.php
```diff
@@ -51,7 +51,7 @@ public static function load($filename) {
 
 	/**
 	 * Return the list of sharing options.
-	 * @return an array of FreshRSS_Share objects.
+	 * @return array[FreshRSS_Share]
```
aledeg, line 54: I thought the syntax was ```FreshRSS_Share[]```

Alkarex: I think there is more than one syntax, and I am not sure which is most common / best (although I think that `array`  is more usual than `[]`).
Feel free to change.
https://stackoverflow.com/questions/778564/phpdoc-type-hinting-for-array-of-objects/23072048#23072048
https://stackoverflow.com/questions/2713710/comment-associative-array-in-php-documentor/2725073

Issue #3622, opened 2021-05-14
[Feature] Is there a way to restrict adding feeds to a pre-defined list of sites?


tareefdev:
**Is your feature request related to a problem? Please describe.**
I'm wondering if there is a way (like a plugin) to restrict adding feeds to a pre-defined list of sites. Say I have a list of 100 sites and want to make the user able only to follow sites from this list.

Alkarex:
Hello,
No, this is not something we have at the moment. However, it would be possible to make a little FreshRSS extension for that. There is an API hook `check_url_before_add` https://freshrss.github.io/FreshRSS/en/developers/03_Backend/05_Extensions.html

aledeg:
As a side note, this would be a system extension so the user cannot modify that list.

tareefdev:
Thanks, this fits my needs, you could close this issue.

aledeg:
Are you going to build an extension? How is your question answered?
I would like to know how your question has been addressed.
Thank you!

tareefdev:
I mean that having the `check_url_before_add` hook could meet my needs. I'm evaluating several RSS aggregators to see which one I will go with, this process considering several criteria. If I'll go with FreshRSS I could try to write an extension myself first (though I did not use PHP before) or seeking help.

aledeg:
I found your challenge interesting so I started writing an extension. But the core of FreshRss was missing a few things to make it work properly. So I've opened #3629 and #3626.

aledeg:
@tareefdev I have a working extension to check url on a white list. It's not usable until #3626 is merged.
I've checked exact feeds for now but the rule can be enhance to be more flexible.

aledeg:
The white list extension is available [here](https://github.com/aledeg/xExtension-WhiteList). It is not yet visible in the extension page but can still be installed.
It is a bit rough around the edges but it works.

We might just close that issue now.

PR #3626, opened 2021-05-17
Add system configuration for extension

aledeg:

- Add system configuration support in extension

1. In an extension code, create, update or delete system configuration.
2. Check in the configuration file that all actions on step 1 are executed.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

Before, only the user configuration was supported by extensions. But it was
limiting if one has to create a system extension with configuration.
Now, both user and system configuration are supported.


aledeg:
I've figured that something was missing in the extension system when I saw #3622.

PR #3629, opened 2021-05-18
Add a request method to extract arrays

aledeg:

- Add a request method to extract arrays

1. add a request param with new lines and use the new method to parse it.
2. this will produce an array of string where each string is one of the line

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -58,6 +58,19 @@ class Minz_Request {
 		}
 		return $value;
 	}
+	/**
+	 * Extract text lines to array.
+	 *
+	 * It will return an array where each cell contains one line of a text. The new line
+	 * character is used to break the text into lines. This method is well suited to use
+	 * to split textarea content.
+	 */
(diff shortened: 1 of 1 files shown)


Alkarex:
I have not understood the type of expected request. Can you please provide an example?

aledeg:
Sure.
The goal of that method is to convert the content of a textarea into an array where each cell is one line of the textarea. My thoughts were to use that in the whitelist extension mentioned in #3622. It might be useful also for the feed filters.
Let me know if my explanation helps or if you need more information to understand what I have in mind.

Alkarex:
Ok. I was a bit confused because I thought it was an helper for array query params https://php.net/faq.html#faq.html.arrays
Maybe change the name to `paramLines` or something like that?

aledeg:
I understand. I'll find a different name!

aledeg:
@Alkarex I've made some changes and I've added some comments in the code.

PR #3637, opened 2021-05-24
Introduced $noredirect in actualizeAction()

marcohald:

Introduced the variable $noredirect
If set to 1 no redirect will happen after actualizeAction()

1. go to https://Freshrss.example.com/p/i/?c=feed&a=actualize&user=user&token=token&noredirect=1
2. No redirect happen after the fetch of the feeds

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

I will update the Documention if this change make sense to the team

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -547,7 +547,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		$url = Minz_Request::param('url');
 		$force = Minz_Request::param('force');
 		$maxFeeds = (int)Minz_Request::param('maxFeeds');
-		$noredirect = (bool)Minz_Request::param('noredirect') ==1;
+		$noredirect = Minz_Request::paramBoolean('noredirect');
 		$noCommit = Minz_Request::fetchPOST('noCommit', 0) == 1;
 
 		if ($id == -1 && !$noCommit) {	//Special request only to commit & refresh DB cache
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -547,6 +547,7 @@ class FreshRSS_feed_Controller extends Minz_ActionController {
 		$url = Minz_Request::param('url');
 		$force = Minz_Request::param('force');
 		$maxFeeds = (int)Minz_Request::param('maxFeeds');
+		$noredirect = (bool)Minz_Request::param('noredirect') ==1;
 		$noCommit = Minz_Request::fetchPOST('noCommit', 0) == 1;
 
 		if ($id == -1 && !$noCommit) {	//Special request only to commit & refresh DB cache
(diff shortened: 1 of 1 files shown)


app/Controllers/feedController.php
```diff
@@ -547,6 +547,7 @@ public function actualizeAction() {
 		$url = Minz_Request::param('url');
 		$force = Minz_Request::param('force');
 		$maxFeeds = (int)Minz_Request::param('maxFeeds');
+		$noredirect = (bool)Minz_Request::param('noredirect') ==1;
```
```suggestion
		$noredirect = Minz_Request::param('noredirect') == 1;
```

Alkarex:
Hello,
Could you please explain a bit the use-case?

marcohald:
Sure, my favorite webcron service marks every site that responses with an redirect as failed.
To circumvent this, I added the option to only print the result of the refresh and then exit.
It is also much easier to parse the response when no HTML is printed

app/Controllers/feedController.php
aledeg: Maybe using the following is more appropriated:
```php
$noredirect = Minz_Request::paramBoolean('noredirect');
```

app/Controllers/feedController.php
```diff
@@ -547,6 +547,7 @@ public function actualizeAction() {
 		$url = Minz_Request::param('url');
 		$force = Minz_Request::param('force');
 		$maxFeeds = (int)Minz_Request::param('maxFeeds');
+		$noredirect = Minz_Request::paramBoolean('noredirect');
```
aledeg, line 550: There is only one use of that variable. Might be useless to set it.
You could use the parameter content directly where you need it.

app/Controllers/feedController.php
```diff
@@ -571,6 +572,15 @@ public function actualizeAction() {
 			Minz_Request::setGoodNotification(_t('feedback.sub.feed.actualizeds'));
 			// No layout in ajax request.
 			$this->view->_layout(false);
+		} if ($noredirect) {
```
aledeg, line 575: It seems that it should be an `elseif` instead of an `if`. Using an `if` here change the flow and the following `else` is not linked anymore to the original `if`.
If the `if` is intended then the layout of the code is not properly done. There must be a blank line between statements.

aledeg (changes requested):

Alkarex:
We actually already have a parameter that could be used for that. Could you please try the following:

```
https://freshrss.example.net/i/?c=feed&a=actualize&user=alice&token=token123456&ajax=1
```

And if that works for you, an update of our documentation would be welcome

Alkarex:
@marcohald Did you have a chance to try the existing `&ajax=1` parameter?

Alkarex:
The `&ajax=1` parameter is actually already documented https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/users/03_Main_view.md#with-the-actualize_scriptphp-script

PR #3651, opened 2021-05-31
Add HTML5 tags

### LINKED ISSUE #3643: [Feature] use HTML5 tags
I would like to read the feeds with the Browser reader
Firefox Reader View: https://support.mozilla.org/en-US/kb/firefox-reader-view-clutter-free-web-pages

One of the reasons is, that the HTML5 tags <header> <main> <footer> and <nav> are not used.

**Additional context**
It could also help for users with a screen reader to navigate through the page

math-GH:
Closes #3643

added some HTML5 tags: header, main, nav

1. does the frontend looks fine?
2. no broken HTML?

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated

(hopefully it is fine. It is my first pull request)


app/views/index/global.phtml
```diff
@@ -8,49 +8,50 @@
 		$class = ' state_unread';
 	}
 ?>
+<main>
+    <div id="stream" class="global<?= $class ?>">
```
Frenzie: Shouldn't that be simply be main then?

app/views/index/normal.phtml
```diff
@@ -13,111 +13,111 @@ $content_width = FreshRSS_Context::$user_conf->content_width;
 
 $today = @strtotime('today');
 ?>
+<main>
+    <div id="stream" class="normal<?= $hidePosts ? ' hide_posts' : '' ?>">
```
Frenzie: Ditto.

app/views/index/global.phtml
math-GH: I did want to merge this tags, because the extension "ThreePanesView" looks like that it needs id=stream to make some magic (= add a column). I want to make sure that also this extension has a correct <main> wrapper

app/views/index/normal.phtml
math-GH: see comment above

app/views/index/global.phtml
Frenzie: I mean `<main id="stream" etc.>` of course, not `<main>`.

math-GH: I leave `<main>` and `<div id="stream">` splited, because the extension "ThreePanesView" searchs for id="stream"
image
and put its wrapper div around
image
In this way the <main> will wrap around "stream" and "threepanesview"

I would not change it.

Frenzie: Could you please be more specific as to what the problem is? I'll fix the extension if necessary but I don't see it.

math-GH: Which information do you need?

Frenzie: What is it that breaks?

math-GH: If `<main id="stream">` than, after ThreePanesView inserted its `<div id="ThreePanesView">`,  `<div id="ThreePanesView">` would not be IN the `<main>`

Frenzie:
Pinging @aledeg for an opinion on where to effect the required change.

aledeg:
I am not really familiar with ThreePanesView and the DOM for that section. But I think @Frenzie is right.
After looking at the ThreePanesView code, it seems that it will be alright. If it's not, it's possible to add a PR on that repository to make the code better suited to fit FreshRSS architecture.

aledeg:
For me, it makes no sense to have a div in a nav. The idea of nav is to have something in the dom that describes what it does. Having the div inside defeat that purpose. Either we add the nav properly or we don't add the nav. The in-between seems unacceptable.
My 2¢

Frenzie:
I'd prefer to phrase that as unnecessary. I don't think it goes as far as defeating the purpose. :-)

Anyhoo, thanks!

math-GH:
`<div>` inside of a `<nav>` is permitted

see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

aledeg:
It's permitted but in our case, I find that it makes no sense. My 2¢

math-GH:
What is the next todo to bring this PR to a happy end?

Frenzie:
Pinging @Alkarex and @nicofrand for their view, but I think mine's clear. ^_^ Incidentally, ThreePanesView doesn't seem to be working for me.

nicofrand:
> ThreePanesView doesn't seem to be working for me

Hi! After this modification or before? Based on the `id="stream"` the extension should still work. I'll try to find some time to debug it if needed.

Frenzie:
Before, but it'll make no difference either way because of the id. On which version is it supposed to work? I tried 1.16, 1.17 and 1.18.

nicofrand:
It should work for the three of them.
Could you create a ticket on https://framagit.org/nicofrand/xextension-threepanesview/-/issues then? With the skin you use, the version, if you have an error in your browser devtools. A screenshot woudl help too.

Frenzie:
Default everything, view normal, versions listed above, no errors, [window isn't too small](https://framagit.org/nicofrand/xextension-threepanesview/-/blob/229fdf01725147833ad4a47dc1835d125f0151f5/static/threepanesview.js#L14-15), script element is added to the HTML.

Manually running
```js
        var stream = document.getElementById("stream");
        var content = stream.querySelector(".flux.current");
        var html = content ? content.querySelector(".flux_content").innerHTML : "";
        stream.insertAdjacentHTML("beforebegin", `<div id="threepanesviewcontainer"></div>`);
        var wrapper = document.getElementById("threepanesviewcontainer");
        wrapper.appendChild(stream);
        wrapper.insertAdjacentHTML("beforeend", `<div id="threepanesview"><div class="flux">${html}</div></div>`);
```
works to add some elements to the DOM.

Edit: screenshot ;-) https://freshrss.org/images/screenshot.png (actually that one seems to be a touch outdated)

nicofrand:
Is it enabled properly?

image

Let's migrate to https://framagit.org/nicofrand/xextension-threepanesview/-/issues if possible to avoid further flood on this issue.

Frenzie:
> Is it enabled properly?

Yes. As I said, the script element shows up in the HTML when enabled. Someone should submit a tiny PR to fix the extension for screenreaders if this PR is merged in the form desired by @aledeg and I; it doesn't need to be me. But anyway, the reason I pinged you is to ask for your opinion on the `<main>` thing? ;-)

nicofrand:
I don't really have one. This would not make sense regarding the extension because I have to wrap the current node in another one, which should be the `main`. But if that makes more sense for FreshRSS itself, go for it!

Alkarex:
Sorry for the delay. Yes, like @Frenzie and @aledeg , I would also vote for `<main id="stream" ...>` and then fix consequences

app/layout/nav_entries.phtml
```diff
@@ -1,5 +1,7 @@
-<ul id="nav_entries">
-	<li class="item"><a class="previous_entry" href="#"><?= _i('prev') ?></a></li>
-	<li class="item"><a class="up" href="#"><?= _i('up') ?></a></li>
-	<li class="item"><a class="next_entry" href="#"><?= _i('next') ?></a></li>
-</ul>
\ No newline at end of file
+<nav>
+    <ul id="nav_entries">
```
Frenzie: Please don't change the whitespace. Otherwise lgtm.

math-GH: fixed

Frenzie (commented):
I implicitly meant all the whitespace changes, excluding the block that adds extra indentation. ;-)

math-GH:
> 
> 
> I implicitly meant all the whitespace changes, excluding the block that adds extra indentation. ;-)

It was not me, but I fixed it. Hopefully I did not miss anything

math-GH:
How to handle the ThreePanesView extension? The `<main>` is not correct now, because it does not wrap everything.
grafik

app/views/index/global.phtml
```diff
@@ -8,24 +8,23 @@
 		$class = ' state_unread';
 	}
 ?>
+<main id="stream" class="global<?= $class ?>">
+	<?php
```
Frenzie: @mathContao This one's also a whitespace change, but pinging @Alkarex @aledeg as to whether it might be preferred.

app/views/index/normal.phtml
```diff
@@ -110,14 +110,13 @@ $today = @strtotime('today');
...
 		ob_end_clean();	//Discard the articles headers, as we have no articles
 ?>
-<div id="stream" class="prompt alert alert-warn normal">
+<main id="stream" class="prompt alert alert-warn normal">
 	<h2><?= _t('index.feed.empty') ?></h2>
 	<a href="<?= _url('subscription', 'add') ?>"><?= _t('index.feed.add') ?></a><br /><br />
-</div>
+</main>
 <?php endif; ?>
-
```
Frenzie, line 122: Please keep it.

app/views/index/reader.phtml
```diff
@@ -7,8 +7,7 @@ call_user_func($this->callbackBeforeEntries, $this);
 $lazyload = FreshRSS_Context::$user_conf->lazyload;
 $content_width = FreshRSS_Context::$user_conf->content_width;
 ?>
-
```
Frenzie, line 10: Please keep it.

app/views/index/reader.phtml
```diff
@@ -76,12 +75,12 @@ $content_width = FreshRSS_Context::$user_conf->content_width;
...
 		ob_end_clean();	//Discard the articles headers, as we have no articles
 ?>
-<div id="stream" class="prompt alert alert-warn reader">
+<main id="stream" class="prompt alert alert-warn reader">
 	<h2><?= _t('index.feed.empty') ?></h2>
 	<a href="<?= _url('subscription', 'add') ?>"><?= _t('index.feed.add') ?></a><br /><br />
-</div>
-<?php endif; ?>
+</main>
+<?php endif; ?>
```
Frenzie: And let's also not change the end of file. :-)
```suggestion
<?php endif; ?>

```

Frenzie (commented):
final nits

app/views/index/global.phtml
Alkarex: Indeed, let's keep that whitespace as it was for now:

```suggestion
<main id="stream" class="global<?= $class ?>">
<?php
```

math-GH: I do not understand here the issue

app/views/index/normal.phtml
math-GH: fixed,  but I do not understand why it is necessary

app/views/index/reader.phtml
math-GH: fixed,  but I do not understand why it is necessary

app/views/index/global.phtml
Frenzie: Please look at the full diff, without ignoring whitespace. It's just a lot of lines that don't have anything to do with any of this.

math-GH: please help me. I do not understand the issue. I did not changed a lot on this files. Now it feels that I have to clean up a lot.

app/views/index/global.phtml
```diff
@@ -8,24 +8,23 @@
...
-		'params' => $params,
-	);
+		$params = Minz_Request::fetchGET();
+		unset($params['c']);
+		unset($params['a']);
+		$url_base = array(
+			'c' => 'index',
+			'a' => 'normal',
+			'params' => $params,
+		);
```
Frenzie:
```suggestion
	$params = Minz_Request::fetchGET();
	unset($params['c']);
	unset($params['a']);
	$url_base = array(
		'c' => 'index',
		'a' => 'normal',
		'params' => $params,
	);
```

app/views/index/global.phtml
```diff
@@ -20,12 +19,12 @@
 		'params' => $params,
 	);
 
-	foreach ($this->categories as $cat) {
-		$feeds = $cat->feeds();
-		$url_base['params']['get'] = 'c_' . $cat->id();
+		foreach ($this->categories as $cat) {
+			$feeds = $cat->feeds();
+			$url_base['params']['get'] = 'c_' . $cat->id();
```
Frenzie:
```suggestion
	foreach ($this->categories as $cat) {
		$feeds = $cat->feeds();
		$url_base['params']['get'] = 'c_' . $cat->id();
```

Frenzie:
```suggestion
	foreach ($this->categories as $cat) {
		$feeds = $cat->feeds();
		$url_base['params']['get'] = 'c_' . $cat->id();
```

app/views/index/global.phtml
```diff
@@ -24,8 +23,8 @@
 		$feeds = $cat->feeds();
 		$url_base['params']['get'] = 'c_' . $cat->id();
 
-		if (!empty($feeds)) {
-?>
+			if (!empty($feeds)) {
+	?>
```
Frenzie:
```suggestion
		if (!empty($feeds)) {
?>
```

app/views/index/global.phtml
```diff
@@ -8,8 +8,7 @@
 		$class = ' state_unread';
 	}
 ?>
-
-<div id="stream" class="global<?= $class ?>">
+<main id="stream" class="global<?= $class ?>">
```
Frenzie, line 12:
```suggestion

<main id="stream" class="global<?= $class ?>">
```

app/views/index/global.phtml
```diff
@@ -46,11 +46,11 @@
 			<?php } ?>
 		</ul>
 	</div>
-<?php
+	<?php
+			}
```
Frenzie:
```suggestion
<?php
		}
```

app/views/index/global.phtml
```diff
@@ -48,9 +48,9 @@
 	</div>
 <?php
 		}
-	}
-?>
-</div>
+		}
+	?>
```
Frenzie:
```suggestion
	}
?>
```

Frenzie (approved):

app/views/index/global.phtml
Frenzie: Should be good now. But you did change those lines. ;-)

math-GH:
thanks for help&fixing @Frenzie

Frenzie:
No problem.

Alkarex:
Thanks again @mathContao 👍🏻

nicofrand:
> How to handle the ThreePanesView extension? The `<main>` is not correct now, because it does not wrap everything.
> grafik

As long as the extension works I don't think this is a huge issue (please remind that this is a client-side JS extension only).

If I were to remain the current `main` in a `div` so that `<div id="threepanesviewcontainer>` would become `<main id="threepanesviewcontainer">` I fear some other extensions or themes would break.

Alkarex:
@nicofrand All the themes I have checked do not match on `div` but on the ID or class. So I suggest to also use `<main>` on your side if that makes semantics sense. Is there any remaining issue on your side, or all good?

nicofrand:
That would require a bit of work: the current main should then be removed, a div created instead, etc. I am not sure I see the added value on the extension side.

Alkarex:
@nicofrand As long as it works and you are happy, all is good :-)

PR #3655, opened 2021-06-01
Fix navigation JS error

### LINKED ISSUE #3654: [BUG] mobile view: top button with JavaScript error
**Describe the bug**
in mobile view
no article is selected

click on the top button in the navigation
image

an JavaScript error expires
image

 - Browser: Firefox 88
 - FreshRSS version: 1.18.0

aledeg:
Closes #3654

- Fix navigation error when no article is selected

1. Display articles without selecting one
2. Use the up navigation button
3. Check that it does not generate an error in the console

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, the UP navigation button was generating an error when no article was
selected. The process to navigate to the first element was using the position
of the current element to compute the offset, thus generating said error.
Now, if no article is selected, the first article is used as a fall-back.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1093,7 +1093,7 @@ function init_nav_entries() {
 				return false;
 			};
 		nav_entries.querySelector('.up').onclick = function (e) {
-				const active_item = document.querySelector('.flux.current'),
+				const active_item = (document.querySelector('.flux.current') || document.querySelector('.flux')),
 					windowTop = document.scrollingElement.scrollTop,
 					item_top = active_item.offsetParent.offsetTop + active_item.offsetTop;
 


Frenzie (approved):
Not too sure about the variable name with this change since it's not an item with all you might expect from that, but it's limited in scope to the line right below so I suppose it's fine.

aledeg:
@Frenzie I've checked the behavior for the other navigation button when nothing is selected. It works by selecting the first element of the list. I guess it can be safe to assume that the active element is the first one when none are selected.

Frenzie:
@aledeg But the first element is an actual item.

aledeg:
I am not following you. Could you rephrase so I can understand what you have in mind?

Frenzie:
My bad, .flux is a bit misleading. That is an item, not the container of the items.

aledeg:
Oh! I see what you meant now. I understand your worries.
I agree with you that `.flux` could be misleading.

PR #3656, opened 2021-06-01
Fix user creation form redirection

### LINKED ISSUE #3648: [BUG] user create: without a passwort -> redirect to start page
**Describe the bug**
When I try to create a user without a password, than I will redirected to the start page without an error and without creating a new user

**To Reproduce**
go to "create new user",
input a user name (1), leave the password empty (2) and klick "submit" (3)
image

**Expected behavior**
a hint, that the password input field is empty. 
Do not expect a redirect to the start page

 - Browser: Firefox 88
 - FreshRSS version: 1.18.0

aledeg:

- Fix user creation form redirection

1. Try to create a user without a password from the user page.
2. Check there is an error displayed and you're still on the user management page.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, when there was an error during the user creation process, the user was
always redirected to the same page no matter where the creation process was
initiated (registration or creation by an admin). Thus, there were no error
messages displayed when something went bad during the creation process by an
admin.
Now, the origin of the user creation has been added to the form to be able to
redirect the user to the proper page.

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -287,25 +287,29 @@ class FreshRSS_user_Controller extends Minz_ActionController {
 			$new_user_name = Minz_Request::param('new_user_name');
 			$email = Minz_Request::param('new_user_email', '');
 			$passwordPlain = Minz_Request::param('new_user_passwordPlain', '', true);
+			$badRedirectUrl = [
+				'c' => Minz_Request::param('originController', 'auth'),
+				'a' => Minz_Request::param('originAction', 'register'),
+			];
 
 			if (!self::checkUsername($new_user_name)) {
 				Minz_Request::bad(
 					_t('user.username.invalid'),
-					array('c' => 'auth', 'a' => 'register')
+					$badRedirectUrl
 				);
 			}
 
 			if (FreshRSS_UserDAO::exists($new_user_name)) {
 				Minz_Request::bad(
 					_t('user.username.taken', $new_user_name),
-					array('c' => 'auth', 'a' => 'register')
+					$badRedirectUrl
 				);
 			}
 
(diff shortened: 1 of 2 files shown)


app/Controllers/userController.php
```diff
@@ -287,25 +287,29 @@ public function createAction() {
 			$new_user_name = Minz_Request::param('new_user_name');
 			$email = Minz_Request::param('new_user_email', '');
 			$passwordPlain = Minz_Request::param('new_user_passwordPlain', '', true);
+			$badRedirectUrl = [
```
Alkarex, line 290: Did you check whether there is no other call to this endpoint?
Maybe it would be good to have default values `[ 'c' => 'auth', 'a' => 'register' ]` if some of the `origin*` parameters are missing

aledeg: I have not. But I'll add default values.

app/views/auth/register.phtml
```diff
@@ -3,6 +3,8 @@
 
 	<form method="post" action="<?= _url('user', 'create') ?>">
 		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
+		<input type="hidden" name="originController" value="<?= Minz_Request::controllerName(); ?>" />
```
Alkarex:
```suggestion
		<input type="hidden" name="originController" value="<?= Minz_Request::controllerName() ?>" />
```

I do not think it makes a difference for now, but the short syntax for echo does not take a final semicolon https://php.net/echo

aledeg: I'll fix that

aledeg:
@Alkarex I've addressed your remarks.

PR #3659, opened 2021-06-02
Clear cache while adding a new feed

### LINKED ISSUE #3658: [BUG] Subscribe Bookmarklet does not works twice
**Describe the bug**
if I use the bookmarklet twice on a page without saving, the second use is not the same

I mean this bookmarklet
image

**To Reproduce**
step 1
- go to a page with RSS (f.e. https://github.blog/)
- use the bookmarklet as usual
- the "add a RSS feed" page of FreshRSS apears 
- image
- close this page (close the tab without clicking on submit and without cancel button)

step 2
- go back to the page of your interest from step 1
- use the bookmarklet again
- the "ass a RSS feed" page of FreshRSS apears again
- image
- but it did not recognize the RSS feed as the step before

**Expected behavior**
same result in step1 and step2

 - Browser: Firefox 88
 - FreshRSS version: 1.18.0

aledeg:
Closes #3658

- Remove cache feed when adding it

1. Click on the bookmarklet to add a feed
2. Do not save the new feed
3. Click again on the bookmarket

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Before, the feed cache was preventing to add a feed when using twice the
bookmarklet if the first time the user aborted the process.
Now, the cache is removed for the feed before loading information. This
behavior as a limited impact since only the selected feed cache is removed.


aledeg:
@Alkarex This works well for the bookmarklet use-case. I don't know if it has side-effects though. Could you validate that I did not miss something? Thank you!

Frenzie (approved):
Fine by me, definitely seems weird though.

Alkarex:
It also solves another very related bug if you add a feed from one user account, and immediately add it in another user account.
There are several small cache issues like that, in particular when there are errors (error information is not persisted correctly).

Alkarex:
It is not the proper fix, but that is probably better than what we have at the moment.
Maybe we should add a TODO though

app/Controllers/feedController.php
```diff
@@ -218,6 +218,7 @@ public function addAction() {
 			$this->view->feed = new FreshRSS_Feed($url);
 			try {
 				// We try to get more information about the feed.
+				$this->view->feed->_attributes('clear_cache', true);
```
Alkarex, line 340:
```suggestion
				$this->view->feed->_attributes('clear_cache', true);	//TODO: Improve cache reuse logic
```

Alkarex:
Hold on, I think this creates a new bug, when adding a feed using auto-discovery, deleting it, and adding it again (or add the same feed in two different users)

aledeg:
As I mentioned in the original issue, we might just add something in the documentation because it is an edge-case.

Alkarex:
This is not completely trivial. When adding a feed with auto-discovery (i.e. provide https://example.net/ instead of https://example.net/rss.xml ), more than one cache file is created and apparently deleting only one creates other bugs.
As this is not a regression, I suggest we release 1.18.1 very soon (e.g. latest this week-end if there is no other feedback) and come back to a proper cache improvement afterwards

aledeg:
Sounds good!

math-GH:
any news?

Alkarex:
> any news?

I need to take some time to work on it, and it is not completely trivial. I would like to complete a couple of other work-in-progress PRs before

math-GH:
A year later, are here any news?

Alkarex:
> A year later, are here any news?

Time is flying! :-D

aledeg:
I haven't been active in months. We might close that for now.

PR #3668, opened 2021-06-07
[i18n] German: flip alter/neuer typo

Frenzie:
Fixes https://github.com/FreshRSS/FreshRSS/issues/3667

PR #3703, opened 2021-07-19
i18n fr fix old-new tag label

Alkarex:
Old/new were swapped in French...


aledeg (approved):

Alkarex:
Ah, it was the same issue than in German https://github.com/FreshRSS/FreshRSS/pull/3668

PR #3709, opened 2021-07-22
Search labels

### LINKED ISSUE #3704: [Feature] Search for one or more labels in a query
As a user who applies granular labels on articles for easier future lookup, I would like to create user query filters or search for multiple labels at once. An example is the below screenshot where I would create tags for 2 entities, separately:

image

But I want to later lookup articles that contain both entities or tags. I am not aware of how to search for labels in the search bar, however. 

image

image

I am also unable to select more than one label from the feed list, where the labels reside:

image

I look at tags as the closest analogy, where multiple tags in a single search query are supported. 

image

I'd like to be able to create search queries and user queries where multiple labels are selected. This would allow me to create user query filters, or simply perform ad-hoc searches in the future.

I am currently creating tags for combinations of tags. In this example, that would be a tag named `aaa_entity_foo + aaa_entity_bar`, but this is not a scalable solution.

**Additional context**
I am hoping this is simply user error.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3704

Implement searching in labels:

* Search using label IDs: `L:12,13,14`
* Search using label names: `label:something` or `labels:"my first label,my other label,🧪"` (supports Unicode)
* Search any label: `L:*`

~~Quick implementation. The syntax for multiple names could maybe be improved (later :-))~~

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -751,6 +751,44 @@ SQL;
 					$sub_search .= ') ';
 				}
 
+				if ($filter->getLabelIds()) {
+					$sub_search .= 'AND ' . $alias . 'id IN (SELECT et.id_entry FROM `_entrytag` et WHERE et.id_tag IN (';
+					foreach ($filter->getLabelIds() as $label_id) {
+						$sub_search .= '?,';
+						$values[] = $label_id;
+					}
+					$sub_search = rtrim($sub_search, ',');
+					$sub_search .= ')) ';
+				}
+				if ($filter->getNotLabelIds()) {
+					$sub_search .= 'AND ' . $alias . 'id NOT IN (SELECT et.id_entry FROM `_entrytag` et WHERE et.id_tag IN (';
+					foreach ($filter->getNotLabelIds() as $feed_id) {
+						$sub_search .= '?,';
+						$values[] = $feed_id;
+					}
+					$sub_search = rtrim($sub_search, ',');
+					$sub_search .= ')) ';
+				}
+
+				if ($filter->getLabelNames()) {
+					$sub_search .= 'AND ' . $alias . 'id IN (SELECT et.id_entry FROM `_entrytag` et, `_tag` t WHERE et.id_tag = t.id AND t.name IN (';
+					foreach ($filter->getLabelNames() as $label_name) {
+						$sub_search .= '?,';
+						$values[] = $label_name;
+					}
+					$sub_search = rtrim($sub_search, ',');
+					$sub_search .= ')) ';
+				}
+				if ($filter->getNotLabelNames()) {
+					$sub_search .= 'AND ' . $alias . 'id NOT IN (SELECT et.id_entry FROM `_entrytag` et, `_tag` t WHERE et.id_tag = t.id AND t.name IN (';
+					foreach ($filter->getNotLabelNames() as $label_name) {
+						$sub_search .= '?,';
+						$values[] = $label_name;
+					}
+					$sub_search = rtrim($sub_search, ',');
+					$sub_search .= ')) ';
+				}
+
 				if ($filter->getMinDate()) {
 					$sub_search .= 'AND ' . $alias . 'id >= ? ';
 					$values[] = "{$filter->getMinDate()}000000";
@@ -913,14 +951,14 @@ SQL;
 			$where .= 'e.id_feed=? ';
 			$values[] = intval($id);
 			break;
-		case 't':	//Tag
+		case 't':	//Tag (label)
 			$where .= 'et.id_tag=? ';
 			$values[] = intval($id);
 			break;
-		case 'T':	//Any tag
+		case 'T':	//Any tag (label)
 			$where .= '1=1 ';
 			break;
-		case 'ST':	//Starred or tagged
+		case 'ST':	//Starred or tagged (label)
 			$where .= 'e.is_favorite=1 OR EXISTS (SELECT et2.id_tag FROM `_entrytag` et2 WHERE et2.id_entry = e.id) ';
 			break;
 		default:
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -732,63 +732,6 @@ SQL;
 				}
 				$sub_search = '';
 
-				if ($filter->getFeedIds()) {
-					$sub_search .= 'AND ' . $alias . 'id_feed IN (';
-					foreach ($filter->getFeedIds() as $feed_id) {
-						$sub_search .= '?,';
-						$values[] = $feed_id;
-					}
-					$sub_search = rtrim($sub_search, ',');
-					$sub_search .= ') ';
-				}
-				if ($filter->getNotFeedIds()) {
-					$sub_search .= 'AND ' . $alias . 'id_feed NOT IN (';
-					foreach ($filter->getNotFeedIds() as $feed_id) {
-						$sub_search .= '?,';
-						$values[] = $feed_id;
-					}
-					$sub_search = rtrim($sub_search, ',');
-					$sub_search .= ') ';
-				}
-
-				if ($filter->getLabelIds()) {
-					$sub_search .= 'AND ' . $alias . 'id IN (SELECT et.id_entry FROM `_entrytag` et WHERE et.id_tag IN (';
-					foreach ($filter->getLabelIds() as $label_id) {
-						$sub_search .= '?,';
-						$values[] = $label_id;
-					}
-					$sub_search = rtrim($sub_search, ',');
-					$sub_search .= ')) ';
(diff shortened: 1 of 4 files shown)


Alkarex:
@ko Please give it a try. I have not tested much yet. We need in particular to test that marking as read articles that are searched this way is working properly (search, and then mark all as read).

ko:
Hi @Alkarex,

With the helpful default RSS feed, I was able to provide a few test cases. Hopefully these help out.

Steps:

1. Go to `Main stream`
2. Expand 1st item. Create/add labels: `alice` and `bob`. 
3. Expand 3rd item. Add label: `alice`
4. Expand 4th item. Add label: `bob`
5. Click on `My labels` in the left-hand side list
6. See all 3 items ✅ 
image

7. Mark first item as `Unread` using the in-line Mail/Envelope icon
8. See unread counter is 2, as it totals the 1 in each tag. ⚠️ If the total counter is to reflect the number of unique items to read, that should be a 1. Otherwise, this is arguably undefined behavior.
image

9. Click `Mark as read` in the top action bar
image

10. See no change to the Unread item. ⚠️ 
image

11. Go to `Main stream` once more
image

12. Click `Mark as read` in the top action bar
13. See successful marking as read
image

Alkarex:
Thanks @ko , those are bugs indeed, but not related to this PR. Could you please open new issues (best after a quick search to check we do not have them already):
* (your point 8) Some articles with more than one label are counted more than once in the total number of unread labelled articles;
* ~~(your point 10) Marking all labelled articles as read does not work with an SQLite database.~~ https://github.com/FreshRSS/FreshRSS/pull/3711

What needs a bit more testing for this specific PR is actions based on searches such as `labels:"bob,alice"`. I have just made a quick test and it seems to work fine.

Furthermore, does this PR solves your needs expressed in https://github.com/FreshRSS/FreshRSS/issues/3704 ?

ko:
Hi @Alkarex ,

Thanks for the pointer on workflow. In hindsight, I can see that those bugs were not specific to this PR, and could apply to even today's label implementation. l In the future I'll create separate bugs to capture that isolated work, with a few more test cases. Apologies for that. 

Going back to the original intent of this PR / Feature Request -- Is there a way to have the search specify an AND operator instead of an OR operator? Specifically, if I have 3 items with these associated labels:

- Item 1, labels=["alice", "bob"]
- Item 2, labels=["alice"]
- Item 3, labels=["bob"]

Currently the search term `label:"alice,bob"` provides all 3 Items. Rather than only Item 1. 

Using the same dataset from the earlier comment, I would have expected to see only the first item in this list.
image

The use case for myself is that I'd want to see only articles where both `alice` and `bob` labels are both present. Though I do understand there is another use case for those who want to see that union of those 2 sets, rather than the intersection.

**EDIT**:

An example of this AND vs. OR I think can be seen with the tag search implementation found in version 1.18.1. Please let me know if you want me to attempt reproduction on this branch.

In this below screenshot, I am searching for `#unix #debugging` and that results in a single article that has both tags:
image
image

However, searching for `#unix` alone and `#debugging` alone provides more results, as other articles have those individual tags. But only the above article (from those that I've sampled) contains both `#unix` and `#debugging`.

image
image

I'm including it into this PR as I imagine this is a bug related to this PR (given it's about label searching, and not a widespread problem otherwise it would've been seen with the tag search query). Please let me know if you want me to create a new bug on this branch, instead.

Alkarex:
Well spotted, there is a bug for the AND. ~~I will try to fix it during the week-end.~~ Fixed.
* The syntax for the AND: `label:labelA label:labelB`
* The syntax for the OR: `labels:labelA,labelB`
* Other syntax for the OR (less efficient): `label:labelA OR label:labelB`

For completeness, `labels:A,B labels:C,D` is understood as `(A OR B) AND (C OR D)`

Alkarex:
@ko I have fixed the boolean AND in addition of the boolean OR. Please try again. See syntax just above.

app/Models/EntryDAO.php
```diff
@@ -800,6 +781,75 @@ protected function sqlListEntriesWhere($alias = '', $filters = null, $state = Fr
 					$sub_search .= ') ';
 				}
 
+				if ($filter->getFeedIds()) {
```
Alkarex, line 784: Updated to be exact same model as label IDs although it does not make a logical difference for feed IDs because an article cannot be in more than one feed

app/Models/Search.php
```diff
@@ -175,25 +196,127 @@ private static function decodeSpaces($value) {
...
 		if (preg_match_all('/\bf:(?P<search>[0-9,]*)/', $input, $matches)) {
-			$ids_lists = $matches['search'];
 			$input = str_replace($matches[0], '', $input);
-			$ids_lists = self::removeEmptyValues($ids_lists);
-			if (!empty($ids_lists[0])) {
-				$this->feed_ids = explode(',', $ids_lists[0]);
-				array_filter($this->feed_ids, function($v) { $v != ''; });
+			$ids_lists = $matches['search'];
+			$this->feed_ids = [];
+			foreach ($ids_lists as $ids_list) {
```
Alkarex, line 202: Updated to be exact same model as label IDs although it does not make a logical difference for feed IDs because an article cannot be in more than one feed

Alkarex:
Please note that the new search operators also have a negative version:

* `!L:12,13` or `-L:12,13`
* `!labels:A,B,C` or `-labels:A,B,C`

ko:
Hi @Alkarex,

I have re-tested this by updating the build, but without updating the database. Please let me know if you want a clean setup.

Staying consistent with our examples above, the AND appears to work using `label:"alice" label:"bob"` ✅ 
image

Similarly, the OR appears to work using `label` as seen in `label:"alice,bob"` ❓ Is this intentional, to have `label` function as `labels` (as seen immediately below)?
image

Similarly, the OR appears to work using `labels` as seen in `labels:"alice,bob"` ✅ 
image

Full disclosure, I understand this is entering the territory of "contrived example." When I create a compound query (for lack of a better phrase), using `labels:"alice,bob" -label:"bob"`, I'm seeing odd behavior. ❓ 
image

Given the snippet from an earlier comment:

> For completeness, labels:A,B labels:C,D is understood as (A OR B) AND (C OR D)

I would have expected to see articles where `(alice OR bob) AND (NOT bob)`, or more succinctly, `alice`.

In which case, I would have expected that compound query to be equivalent to `label:"alice"` in this environment.
image

Without looking at how this gets mapped into the underlying SQL statement, I can imagine that I simply misunderstand how the `-` search operator would function. If that's the case, and this is expected behavior, then that's not a problem. User error on my part.

Otherwise, this all appears to provide the functionality that I was hoping for! The AND operator in the search for labels provides my own workflow with a lot more flexibility, which I am appreciative for.

Alkarex:
I have just added another option, to search for any label, with `L:*` or no label, with `!L:*`

@ko Yes, I have made `label:` and `labels:` as simple aliases, and one can write either form.

@ko Please try again the `labels:"alice,bob" -label:"bob"` filter, which should work exactly like a `label:"alice" -label:"bob"` (but not like a `label:"alice"`)

ko:
Hi @Alkarex,

> I have just added another option, to search for any label, with L:* or no label, with !L:*

Both features that I wasn't even thinking about, but could be helpful with more complicated search queries!

I've confirmed that `L:*` works as I'd expect ✅
image

and that `!L:*` also works as I'd expect ✅ 
image

> Yes, I have made label: and labels: as simple aliases, and one can write either form.

Very cool!

> Please try again the labels:"alice,bob" -label:"bob" filter, which should work exactly like a label:"alice" -label:"bob" (but not like a label:"alice")

I just tested `label:"alice,bob" -label:"bob"` and this works as _I now_ expect. Candidly, I admit that I may have been confused, myself, about what I initially expected when it comes to that compound query, as I added that case to the previous comment.

Your distinction is a good one. 

The initial query of ("alice" OR "bob"), below would show all 3 items:
image

where the items have these tags:

* `FreshRSS 1.18.1` has the tags `["alice","bob"]`
* `FreshRSS 1.18.0` has the tag `["alice"]`
* `FreshRSS 1.16.2` has the tag `["bob"]`

and appending a `-label:"bob"` to the query results in removing every item from the list that includes `"bob"`, leaving just `FreshRSS 1.18.0`.
image

Such that this is obviously different from the search of `labels:"alice"` which would show both `FreshRSS 1.18.1` and `FreshRSS 1.18.0`. 

---

From my point of view, this passes the basic tests I ran, and looks good to me.

Again, much appreciated @Alkarex!

Alkarex:
Tested with SQLite, PostgreSQL, MySQL, MariaDB

Alkarex:
Thanks for the great feedback and testing @ko 👍🏻

PR #3724, opened 2021-07-26
Input validation: red border

### LINKED ISSUE #3647: [BUG] New user: input is always red, when empty
**Describe the bug**
image
the input is always red, when the user enter this form

**To Reproduce**
go to the add user form

**Expected behavior**
no red border, only then I set my cursor in it and left the input

 - Browser: Firefox 88
 - FreshRSS version: 1.18.0

math-GH:
Closes #3647

- fix #3647 
- the input is not anymore red borderd by default
- red border appears while: focus the input or hover the input

1. go to an form with empty mandatory input (f.e. "Add a feed or category" or "Manage Users"

(tested with Firefox 90 and Microsoft Edge on Windows 10)

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Frenzie:
I see your point, but I'm not sure if only on focus is necessarily the solution. On the flip side, maybe it's good enough when combined with :user-invalid so that a more desirable behavior will automatically appear in the future.
https://drafts.csswg.org/selectors/#user-invalid-pseudo

math-GH:
> 
> 
> I see your point, but I'm not sure if only on focus is necessarily the solution. On the flip side, maybe it's good enough when combined with :user-invalid so that a more desirable behavior will automatically appear in the future.
> https://drafts.csswg.org/selectors/#user-invalid-pseudo

I checked it already: user-invalid works only in Firefox :(
https://caniuse.com/?search=user-invalid

p/themes/Origine/origine.css
```diff
@@ -55,7 +55,8 @@ input:focus, select:focus, textarea:focus {
 	box-shadow: 0 2px 2px #ddf inset;
 }
 
-input:invalid, select:invalid {
+input:focus:invalid, select:focus:invalid,
+input:hover:invalid, select:hover:invalid {
```
Frenzie, line 58: > I checked it already: user-invalid works only in Firefox :(

Hence in the future. :-) But only if you include it.

```suggestion
input:user-invalid, select:user-invalid,
input:focus:invalid, select:focus:invalid,
input:hover:invalid, select:hover:invalid {
```

p/themes/Origine/origine.rtl.css
```diff
@@ -55,7 +55,8 @@ input:focus, select:focus, textarea:focus {
 	box-shadow: 0 2px 2px #ddf inset;
 }
 
-input:invalid, select:invalid {
+input:focus:invalid, select:focus:invalid,
```
Frenzie, line 57: This one too of course! ;-)
```suggestion
input:user-invalid, select:user-invalid,
input:focus:invalid, select:focus:invalid,
```

math-GH: forgot it 🙈

Frenzie (approved):
lgtm, unless we want to involve some script

Right now this only works completely as intended in the latest versions of Firefox, but other browsers will presumably implement it in the near future.

Alkarex:
I suggest we wait a bit to see this syntax implemented in one more browser or moved from draft to Candidate Recommendation (CR) at least
https://developer.mozilla.org/docs/Web/CSS/:user-invalid

Frenzie:
The status is for the whole document, but if you prefer.

math-GH:
closed. I still hate the red borders.

PR #3731, opened 2021-07-28
New article text banner: fix line-height

math-GH:
Theme: Origine

before (big gap between the lines):
grafik

after:
grafik

1. wait till the feeds get an update

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -743,6 +743,7 @@ a.btn {
 	background: #0084cc;
 	text-align: center;
 	font-size: 0.9em;
+	padding: 1em;
 }
 
 #new-article:hover {
@@ -750,7 +751,6 @@ a.btn {
 }
 
 #new-article > a {
-	line-height: 3em;
 	color: #fff;
 	font-weight: bold;
 }
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -740,10 +740,10 @@ a.btn {
 
 /*=== New article notification */
 #new-article {
+	padding: 1em;
 	background: #0084cc;
 	text-align: center;
 	font-size: 0.9em;
-	padding: 1em;
 }
 
 #new-article:hover {


p/themes/Origine/origine.css
```diff
@@ -743,14 +743,14 @@ a.btn {
...
 	font-size: 0.9em;
+	padding: 1em;
 }
 
 #new-article:hover {
 	background: #06c;
 }
 
 #new-article > a {
-	line-height: 3em;
```
Frenzie, line 753: Sidenote but line-height should not normally (if ever) use units.

p/themes/Origine/origine.css
```diff
@@ -743,14 +743,14 @@ a.btn {
 	background: #0084cc;
 	text-align: center;
 	font-size: 0.9em;
+	padding: 1em;
```
Frenzie: Is the order right?

Frenzie (approved):
lgtm in principle

p/themes/Origine/origine.css
math-GH: this line I deleted. It is not added ;)

Frenzie: I know, but it's important not to repeat this mistake.

p/themes/Origine/origine.css
math-GH: is there a rule of ordering the css rules?

Frenzie: Yes, but there seems to be some issue with the CI atm.

Frenzie: https://github.com/FreshRSS/FreshRSS/blob/edge/.stylelintrc

Frenzie: Should go at the top, not bottom.

math-GH: fixed

Alkarex:
Thanks! A check for a similar issue in the other themes would be welcome

PR #3737, opened 2021-07-30
CSS line-height improvement

### LINKED ISSUE #3733: [BUG] Origine line-height with units
@Frenzie does not like the line-height with units (see: https://github.com/FreshRSS/FreshRSS/pull/3731#discussion_r678652289)

grafik

Why it would make sense: https://css-tricks.com/almanac/properties/l/line-height/#unitless-line-heights

**Expected behavior**
no units

 - FreshRSS version: 1.18.1

math-GH:
Closes #3733

- deleted/rewriten line-height

1. look around the UI ;)

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


p/themes/Origine/origine.css
```diff
@@ -175,7 +175,6 @@ form th {
 	border-radius: 3px;
 	min-height: 37px;
```
Frenzie, line 171: No need for this line either.

p/themes/Origine/origine.css
```diff
@@ -184,7 +183,6 @@ form th {
 
 a.btn {
 	min-height: 25px;
```
Frenzie, line 180: Presumably not necessary?

p/themes/Origine/origine.css
```diff
@@ -911,14 +901,14 @@ a.btn {
...
-	padding: 0 0 0 5px;
+	padding: 15px;
 	font-size: 0.9em;
 	border: 1px solid #eeb;
 	border-radius: 3px;
 	box-shadow: 0 0 5px #ddd;
 	text-align: center;
 	font-weight: bold;
-	line-height: 3em;
+	line-height: 1.3;
```
Frenzie: Is that necessary at all? Default is usually between 1.2 and 1.3.

p/themes/Origine/origine.css
math-GH: did not check the min-height. Will do it. Thanks for that hint.

p/themes/Origine/origine.css
math-GH: it is necessary. in base-theme/template.css the line-height is 3em
grafik

Frenzie: That would mean `line-height: **3**`, not 1.3. This line should be removed because 3 is a mistake and 1.3 is useless.

math-GH: ah ok. I got the point now.
honestly: I personally do not like line-height: 3em in a notification. The gap between 2 lines of a notification is huge (especially in mobile view).

I will delete the line...

p/themes/Origine/origine.css
```diff
@@ -465,7 +459,7 @@ a.btn {
 	display: block;
 	color: #333;
 	font-style: italic;
-	line-height: 3em;
+	line-height: 3;
```
Frenzie: While fairly unlikely to be actively harmful, that's still the wrong way to go about this as a matter of principle. Use something like padding-top/bottom .9em instead.

math-GH: I deleted this line because it is a copy from base-theme/base.css.

Frenzie: To be clear, `line-height: 3(em)` seems to be abused as a way of specifying padding-top/bottom: `(3-1.2)/2 = 0.9(r)em`.

math-GH: To clear it it will be a next step in the base.css. With this changes here in these branch I will focus on the Origine.css

Frenzie: Cool, thanks. :-)

math-GH:
I deleted some line with `line-height` because (imho) it does not make sense to have the same lines in the theme (Origine) and default-theme (base-theme).

But: I see that origine.css has all lines copied from base-theme and added more css.

How to procede: 
a) origine.css should be a copy + more theme styling
b) origine.css should not have duplicated lines from bases-theme and should only include theme specific styles.

p/themes/Origine/origine.css
math-GH: it is copied from base-theme/template.css

p/themes/Origine/origine.css
math-GH: it is copied from base-theme/base.css

Frenzie: In that case it's better to either leave the entire definition for another time or to fix both at once.

Alkarex:
In my opinion, no need to repeat any style already defined in base-theme

Alkarex:
It looks like there are a few bugs in the titles:
image

Alkarex (changes requested):
Bug in the title (see screenshot), and please remember the RTL version of the CSS

math-GH:
NOTE: this was/is a DRAFT, not an active PR

math-GH:
I close it. Would restart it from beginning

PR #3761, opened 2021-08-10
Improve the link "back to RSS feeds": set a CSS class

math-GH:
The link "back to RSS feeds" has no CSS class to style it
grafik

This PR wraps the link with an additional `<div>`

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


app/views/auth/index.phtml
```diff
@@ -1,7 +1,9 @@
 <?php $this->partial('aside_configure'); ?>
 
 <div class="post">
-	<a href="<?= _url('index', 'index') ?>"><?= _t('gen.action.back_to_rss_feeds') ?></a>
+	<div class="backToFeeds">
```
Frenzie: You didn't really provide a rationale, but assuming it's desirable that should be something more like `back-link`, `link back` (yes, two separate class names), or something along those general lines.

math-GH: I am not experienced in given the correct class name. What about this one?
```html
<div class="back-link-wrapper">
	<a class="link-back" href="<?= _url('index', 'index') ?>"><?= _t('gen.action.back_to_rss_feeds') ?></a>
</div>
```

Frenzie: It's not about experience, but if you check a few lines down, for example, you see:

```html
		<div class="form-group">
			<div class="group-controls">
```

You'll want to try to blend in with what's already there.

> What about this one?

Without a rationale it's hard to say. The question remains if it should be so specific, but in any case it fits in. ;-)

math-GH: Is this fine?
```html
<div class="post-link">
	<a class="link-back, link back" href="<?= _url('index', 'index') ?>"><?= _t('gen.action.back_to_rss_feeds') ?></a>
</div>
```

Frenzie: No, that would be either `link-back` or `link back`. It's a slightly philosophical difference in how you choose to style things. ;-)

Alkarex:
@mathContao Could you please fix the merge conflict?

math-GH:
> 
> 
> @mathContao Could you please fix the merge conflict?

fixed

Frenzie (approved):
lgtm (didn't test)

PR #3771, opened 2021-08-15
Install routine: Reload button if checks fail

### LINKED ISSUE #964: Reload button when installation checking is red
When on the first step (vérifications), why not add a "reload" or "re-check" button when not all is green?

math-GH:
Closes #964

- added a reload button in the install routine step 2
grafik

- translation made with Google translate + check with my knowledge + recheck with invers translation

1. install the app
2. see step 2, when some checks fail

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] documentation updated


app/install.php
```diff
@@ -438,6 +438,9 @@ function printStep1() {
 	<a class="btn btn-important next-step" href="?step=2" tabindex="1" ><?= _t('install.action.next_step') ?></a>
 	<?php } else { ?>
 	<p class="alert alert-error"><?= _t('install.action.fix_errors_before') ?></p>
+	<a id="actualize" class="btn" href="./index.php?step=1" title="<?= _t('install.check.reload') ?>">
+		<img class="icon" src="../themes/icons/refresh.svg" alt="🔃">
```
```suggestion
		<img class="icon" src="../themes/icons/refresh.svg" alt="🔃" />
```

Personal preference for a stricter syntax

math-GH: you are right. I prefer it too

PR #3785, opened 2021-08-19
Keep <noscript> while fetching the news

### LINKED ISSUE #3688: [BUG] ignore <noscript> when truncated feed content
**Describe the bug**

**To Reproduce**
Feed URL:
https://www.spiegel.de/index.rss

CSS selector `#Inhalt section.relative`
grafik

It comes to doubled pictures
grafik
grafik
(URL of article: https://www.spiegel.de/ausland/haiti-jovenel-moise-ermlordet-staat-der-gewalt-a-2b30564e-ff09-44ae-a4e4-94f1ee78ef44#ref=rss )

The reason:
The image is also in `<noscript>`
grafik

Conclusion:
The function, that extracts the content deletes the `<nosript>` and so the content of `<noscript>` is shown

**Expected behavior**
ignore `<noscript>` content or have an option in to ignore or show it

 - FreshRSS version: [e.g. 1.18.1]

math-GH:
Closes #3688

- deleted noscript from the strip_htmltags array

see #3688

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested: used this configuration since July 12th daily. It works fine
- [ ] documentation updated

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -199,7 +199,7 @@ function customSimplePie($attributes = array()) {
 	$simplePie->strip_htmltags(array(
 		'base', 'blink', 'body', 'doctype', 'embed',
 		'font', 'form', 'frame', 'frameset', 'html',
-		'link', 'input', 'marquee', 'meta', 'noscript',
+		'link', 'input', 'marquee', 'meta',
 		'object', 'param', 'plaintext', 'script', 'style',
 		'svg',	//TODO: Support SVG after sanitizing and URL rewriting of xlink:href
 	));


Alkarex:
On a second thought, I am not convinced. We are removing `<script>...</script>` sections, so we should show the content of `<noscript>...</noscript>` section. I will put another comment in the related issue:

> I believe the feed is wrong because the second image is always displayed and does not depend on supporting `<script>`, so the feed should not have that `<noscript>` section at all.
> As far as I can see, the current behaviour of FreshRSS is correct.
> But to solve your issue more generally, I think the proper fix would be to implement a native filter, allowing to remove elements matching a CSS Path / XPath.

PR #3795, opened 2021-08-23
Form actions buttons improved style (mobile and desktop)

math-GH:
Theme: Origine

before:
buttons are not in the line to the other input elements

desktop:
grafik

mobile:
grafik
(in mobile view: the grey background has a gap to the window border)

In my opinion it does not look good

This PR improves to this better layout:

desktop:
grafik

mobile:
grafik

- CSS improved

see every button row

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -90,13 +90,14 @@ form th {
 /*===============*/
 /*=== Forms */
 .form-group.form-actions {
+	margin-bottom: 40px;
 	padding: 5px 0;
 	background: #f4f4f4;
 	border-top: 1px solid #ddd;
 }
 
 .form-group.form-actions .btn {
-	margin: 0 10px;
+	margin: 0 20px 0 0;
 }
 
 .form-group .group-name {
@@ -1193,6 +1194,13 @@ a.btn {
 
 	.nav_menu .search input:focus {
 		width: 400px;
+	}
+
+	.form-group.form-actions {
+		margin-left: -15px;
+		margin-right: -15px;
+		padding-left: 15px;
+		padding-right: 15px;
 	}
 
 	.day .name {
--- a/p/themes/Origine/origine.rtl.css
+++ b/p/themes/Origine/origine.rtl.css
@@ -90,13 +90,14 @@ form th {
 /*===============*/
 /*=== Forms */
 .form-group.form-actions {
+	margin-bottom: 40px;
 	padding: 5px 0;
 	background: #f4f4f4;
 	border-top: 1px solid #ddd;
 }
 
 .form-group.form-actions .btn {
-	margin: 0 10px;
+	margin: 0 20px 0 0;
 }
 
 .form-group .group-name {
@@ -1193,6 +1194,13 @@ a.btn {
 
 	.nav_menu .search input:focus {
 		width: 400px;
+	}
+
+	.form-group.form-actions {
+		margin-left: -15px;
+		margin-right: -15px;
+		padding-left: 15px;
+		padding-right: 15px;
 	}
 
 	.day .name {


math-GH:
another page with more buttons:

grafik

grafik

p/themes/Origine/origine.css
```diff
@@ -1195,6 +1196,13 @@ a.btn {
 		width: 400px;
 	}
 
+	.form-group.form-actions {
+		margin-left: -15px;
+		margin-right: -15px;
+		padding-left: 15px;
+		padding-right: 15px;
+	}
```
Frenzie, line 1204: What about `.form-group.form-actions .btn:first-child {margin:0}`?

Frenzie: Or something like

```css
.form-group.form-actions .btn {
	margin: 0;
}
.form-group.form-actions .btn+.btn {
	margin: 0 10px;
}
```

math-GH: this "green" code brings the grey background to the window border.

your code does it not.

I will mark you in the code where we should discuss your idea

p/themes/Origine/origine.css
```diff
@@ -90,13 +90,14 @@ form th {
...
 .form-group.form-actions {
+	margin-bottom: 40px;
 	padding: 5px 0;
 	background: #f4f4f4;
 	border-top: 1px solid #ddd;
 }
 
 .form-group.form-actions .btn {
-	margin: 0 10px;
+	margin: 0 20px 0 0;
```
math-GH, line 100: @Frenzie here we should discuss

p/themes/Origine/origine.css
Frenzie: Not quite, I overlooked that you were doing things to more than just buttons. That being said, I would posit something like this as a possible alternative.

```
background: linear-gradient(90deg, rgba(0,0,0,0) 0%, rgb(244, 244, 244) 10%,rgb(244, 244, 244) 90%, rgba(0,0,0,0) 100%);
```

Screenshot_2021-08-23_23-06-54

math-GH: I do not understand why you posted this code line for background

Frenzie: Just thinking aloud. It's an alternative way to approach the issue. A potential advantage would be that you don't need different dedicated workarounds for desktop and mobile, but of course the looks are a matter of taste.

math-GH: I wanted to make the buttons flush to the inputs. That was my main goal. The margin-paddings are just to reach the goal in mobile view.

Frenzie: If it's just to make the buttons flush I'd prefer what I said above with for example :first-child or .btn+.btn.

math-GH: your solution does not make it beautiful in mobile view, because the grey background has on the left no padding to the button with your solution.

math-GH: one more point: yes, it was not the goal to change the background, but the new layout I like more than the old one.

Frenzie: So also do it on larger widths? ^_^

math-GH: I do not understand what you mean.

you see above the results of the CSS changes. What do you want to change?

Frenzie: It would seem that logically you would want to add:

```
margin-left: -50px;
padding-left: 50px;
margin-right: -50px;
padding-right: 50px;
```

for larger screen widths.

math-GH: have you seen that these lines are only for the mobile view?

`@media (max-width: 840px) `

There are 15px between the gray background to the window border. That is why I used -15px/+15px.

Origine theme does not support "larger screen widths" only mobile (<840px) and normal (>840px)

Frenzie: > have you seen that these lines are only for the mobile view?

That's my point… :-)

Frenzie: Screenshot_2021-08-25 Display · FreshRSS

math-GH: Thanks, now I understand your point.

I did not want to change/fix it, because it breaks the layout and is not an upgrade (imho).

Frenzie (approved):
Eh, fine by me.

Issue #3052, opened 2020-06-12
Add information about the Reeder 4 date fix to README


maurice-schleussinger:
Hello 🙂 

I just installed FreshRSS and was wondering why all new feed entries in Reeder 4 are wrongly dated. After skimming trough this lengthy discussion https://github.com/FreshRSS/FreshRSS/issues/2759 it is my understanding that this won't be changed as it is open to interpretation if this is a correct implementation of the Google Reader API or not.

From a more casual user perspective this is an irritating experience overall. I would suggest adding a **very visible** line or footnote about the need to install https://github.com/javerous/freshrss-greader-redate if one wishes to use FreshRSS and Reeder and have correct dates. https://github.com/FreshRSS/FreshRSS/blob/master/README.md#apis--native-apps seems like a good place for that.

Thank you for taking the time to work on this awesome software! 🥳

johnnykang:
Hi, just wonder if the extension still works with Reeder 4.2.5? I installed the extension but the api endpoint still returns the wrong date (as discussed in the #2759). I am sure the extension is loaded.

Thought?

maurice-schleussinger:
@johnnykang I am using it with the iOS Reeder App (v4.2.5) right now. Dates are correctly formatted.

Does you extension configuration show a screen like this?
<img width="1433" alt="image" src="https://user-images.githubusercontent.com/1615089/91634521-57580d80-e9f1-11ea-8516-2fef0b48f97d.png">

Daniel15:
I know this is old, but thanks for posting it!! I was having the same issue in FeedMe and this issue pointed me to the GReader Redate extension. Works perfectly!

PR #3796, opened 2021-08-24
Extend GReader Redate note for FeedMe too

Daniel15:
This extension is also needed for FeedMe (tested that myself), and [its source code](https://github.com/javerous/freshrss-greader-redate/blob/master/xExtension-GReaderRedate/extension.php) implies that it's needed for EasyRSS too.

References #3052


Frenzie:
It's not. ;-)

https://github.com/Alkarex/EasyRSS/pull/49

Daniel15:
@Frenzie  Thanks for confirming. I updated the PR to only add the note for FeedMe.

Frenzie (approved):

Alkarex:
Could you please do the same for that readme too? https://github.com/FreshRSS/FreshRSS/blob/edge/README.fr.md

PR #3804, opened 2021-08-25
Add navigation to about page

math-GH:
The "About" page has no navigation. It is only available via the dropdown menu (and via the login screen)

- add the configuration aside navigation to "about" page. It is now available in the configuration menu (for normal users and for admin users)
- the about page via the login screen does not changed (no navigation aside)

grafik

go to the "about" page

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


aledeg:
By doing such thing, you are breaking the menu. I mean that the about page is separated from the administration section in the menu, not in your modification. It can be confusing.

Alkarex:
> By doing such thing, you are breaking the menu. I mean that the about page is separated from the administration section in the menu, not in your modification. It can be confusing.

I think just adding a line separator could be enough.

Waiting for https://github.com/FreshRSS/FreshRSS/pull/3803 to be merged first

aledeg:
That's what I had in mind!

math-GH:
> I think just adding a line separator could be enough.
> 

What do you think about an empty lightgrey line?
grafik

Alkarex:
> What do you think about an empty lightgrey line?

Fine for me 👍🏻

PR #3805, opened 2021-08-25
UI: Add optional thumbnail and summary on feed items

### LINKED ISSUE #561: UI: Thumbnail mode
In addition to the one-line mode and the fully-expanded mode, it would be nice to have a display mode showing a thumbnail and a bit of text.
See https://diaspora-fr.org/posts/524158

### LINKED ISSUE #2038: Preview mode?
Hi,
I was using my FreshRSS today and was wondering if is it possible to have a "preview mode" for the list.

The idea is to be able to read the first paragraph of the content, instead of only the title.
With so many "click bait titles" or not titles that are really bad, I end up opening each article to decide if it is interesting or not.
It would be fantastic to have this option on the "view modes".
Thanks

ORelio:
Closes #561

**Changes proposed in this pull request**:

Implements Thumbnails as well as Summary on feed items in the normal view.
Example feed: `https://www.futura-sciences.com/rss/definitions.xml`

image

Settings are implemented under _Configuration > Display_ and saved in per-user settings:

image

When enabling Thumbnails only, the image is smaller:

image

I've tested the changes with all built-in themes and made specific adjustments for the Compact theme:

image

The right-to-left variant is also handled:

image

1. Under _Configuration > Display_, check _Thumbnail_, _Summary_, or both
2. Submit changes and go back to the normal view
3. New elements are shown in the feed

**Pull request checklist:**

- [X] clear commit messages
- [X] code manually tested
- [ ] unit tests written (sorry, did not implement tests but let me know if/what tests are needed)
- [ ] documentation updated (should not be needed but let me know if I need to change something)

**Additional notes:**

Regarding translation, items are translated in English and French only at the moment, and left in English for other locales.

I encountered a conflicting `position: absolute` CSS rule from [this commit](https://github.com/FreshRSS/FreshRSS/commit/c5274cfa298f1020bab30bd6026323be9e4be43d#diff-65f6b443253a1121cd18180e99827a32b608e5c3a2fe42efb6f5a6317e3e8d1eR659) that cause glitches when hovering items having a thumbnail or summary, so using an additional CSS class on the `title` item when either element is enabled, the conflicting rule is overridden only when needed and left as is otherwise to avoid reintroducing #367 accidentally.

As this is my first pull request on this project, I'm not very familiar with the codebase and may have missed something. Please let me know if anything is missing so I can improve. 🙂


Frenzie:
If it's a setting for the top line only (which seems a bit questionable? I said similar to content width for a reason ;-) ), then shouldn't the checkbox be disabled for the bottom line? Or is it just that the disabled checkbox looks identical to one that can be clicked on your screenshot.

ORelio:
Hi,

> Or is it just that the disabled checkbox looks identical to one that can be clicked on your screenshot.

They are indeed disabled but it's hard to spot thanks to Xubuntu's default theme 😅

```diff
app/views/configure/display.phtml
@@ -125,6 +133,8 @@
		FreshRSS_Context::$user_conf->bottomline_sharing ? ' checked="checked"' : '' ?>
		data-leave-validation="<?= FreshRSS_Context::$user_conf->bottomline_sharing ?>"/></td>
   	<td><input type="checkbox" disabled="disabled" /></td>
+	<td><input type="checkbox" disabled="disabled" /></td>
+	<td><input type="checkbox" disabled="disabled" /></td>
   	<td><input type="checkbox" name="bottomline_date" value="1"<?=
		FreshRSS_Context::$user_conf->bottomline_date ? ' checked="checked"' : '' ?>
		data-leave-validation="<?= FreshRSS_Context::$user_conf->bottomline_date ?>"/></td>
```

> which seems a bit questionable? I said similar to content width for a reason ;-)

Having a setting similar to content width would give a dropdown menu like this:

image

... Which seems rather cumberstone because the user may want thumbnail without summary and vice-versa. Some relevant article that explains why it's a bad practice: (in French): https://lehollandaisvolant.net/?d=2021/06/20/13/30/37-linterface-deficiente-dans-la-vie-courante

Using checkboxes, which seems more adequate, the settings would look like this:

image

... But then, the checkboxes seems redundant with the table, so I ended up placing the settings directly in the table. After all, the items are shown inside the top line, and the existing layout already has disabled checkboxes 🙂

Frenzie:
> Having a setting similar to content width would give a dropdown menu like this:

Not at all, the salient point is that it's split off from that table. It doesn't imply using a dropdown. :-)

> Using checkboxes, which seems more adequate, the settings would look like this:

I think that's clearer, provided you drop the "in top line" part of that phrase.

It just feels too separate to me, even if technically it might be in the top line.

Let's see what some other people think.

aledeg:
Maybe it's time to rework that section.
I think we could replace "article icons" (which does contain other info that are not icons) by 2 new entries, "top line" and "bottom line" (better naming requested).
Those entries will have all the available options as checkboxes. This way we won't have disabled choices.

Frenzie:
Maybe something like article header & footer?

aledeg:
That naming is way better :+1:

Frenzie (approved):
Anyway, reworking it isn't really part of this PR, unless @ORelio wants it to be of course.

So I'll say I'm on board with this as is… I think. Although I am a bit afraid of a permanent temporary solution. :-)

math-GH:
I like this solution (did not test yet) and after merging I will switch my prod installation to dev :)

ORelio:
I find the table layout pretty clean since most items can be activated for both top and bottom. But with more items, yes maybe we could reorganize this 🤔

I'm okay with discussing this matter in a new issue.

> Although I am a bit afraid of a permanent temporary solution. :-)

#561 was opened in 2014 so if we follow this schedule we've got plenty of time 😆

math-GH:
Could you please add lazy load to the ``?

I would suggest to put the image into an separate `<li>`

And I prefer a square as image (80x80px). It cuts much better the images.

ORelio:
Hi,

> Could you please add lazy load to the ``?

Not sure how to do this, will have a look at the source code and see if it's easy to implement.

> I would suggest to put the image into an separate `<li>`

I'll have a look. Not sure how to handle optional columns but that should also be doable.

> And I prefer a square as image (80x80px). It cuts much better the images.

Most images I encountered in my feeds were cut in 16:9 or at the very least in 4:3, see for yourself:

```
https://www.futura-sciences.com/rss/definitions.xml
https://www.nextinpact.com/rss/news.xml
https://www.numerama.com/feed/
https://www.silicon.fr/feed
```

That's why I used a landscape aspect ratio (1,6: 128x80px) for the images. Feedly also has a similar aspect ratio (1,66: 130x78px):

image

For that reason, I think it would be better to stick to a 16:9 aspect ratio by default, although you would be able to easily change it through a custom theme or, even simpler, a [custom CSS](https://github.com/FreshRSS/Extensions/tree/master/xExtension-CustomCSS) snippet:

```css
.flux .item.title .thumbnail { width: 80px; }
.flux .item.title .thumbnail.small { width: 40px; }
```

aledeg:
or you can show only a square image filling the whole area.
you can do this with `object-fit: cover`.

Frenzie:
You can use loading=lazy.

We also have JS to provide lazy loading since that attribute is pretty new.

Frenzie:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading

As you can see it's only been widely supported for a little over a year, but since the fallback just means a few more images might load I think that shouldn't be an issue.

If we do think it's an issue on older browsers (but current Firefox ESR for example is 78 and it supports it), I think the lazy loading script might take care of it automatically with these attributes? Or it might need to be purposefully initiated, don't recall otoh.
https://github.com/FreshRSS/FreshRSS/blob/97ba6261a8eb7f94963ecc91043e5a2d54ffd248/lib/lib_rss.php#L285-L296

ORelio:
Hi,

> or you can show only a square image filling the whole area.
> you can do this with object-fit: cover.

That's already what I'm doing but since people may want a square image I think I'll just make one more setting after all 😄

```css
.flux .item.title .thumbnail img {
	width: 100%; /* use size set by outer span */
	height: 100%; /* use size set by outer span */
	object-fit: cover; /* fill in all space by auto-cropping */
}
```

> You can use loading=lazy.

I see that under _Configuration > Reading_ there is a setting for Lazy loading:

image

... so it makes sense to respect that setting when enabled. Will do!

ORelio:
The item has now its own `<li>` element, and `loading="lazy"` attribute when the corresponding setting is enabled in config.

image

image

Thumbnail size can be configured, so square thumbnails are now a thing:

image

image

image

image

image

All settings above without summary:

image

image

image

image

All combinations above on Origine-compact:

image

image

image

image

image

image

image

image

All combinations above on right-to-left locale:

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

... Could take a very long time if proceeding to other themes so I'll stop here 🤣

Anyway, all requested changes should be done now.

Frenzie (approved):

Alkarex:
@ORelio Many thanks for this great PR. See my changes https://github.com/FreshRSS/FreshRSS/pull/3805/commits/690515cff4af7e23456e8f4cb99f9c3dc3c54ce2 
* Fallback to images in the body if no valid enclosure is found
* Filter out enclosures, which are not images
* Fix a couple of Unicode encoding issues

Alkarex (approved):

ORelio:
Wow, thanks for improving the internal thumbnail support!
Was thinking of maybe investigating this in a further contribution but you were much faster on this 😄
I'm also OK with the other fixes 👍

>  CSS lint + RTL

Any way I can run the CSS check on my side? E.g. I'm not sure why `max-height` should be specified first.

Frenzie:
`stylelint --syntax scss "**/*.scss"` and `stylelint "**/*.css"`, see

https://github.com/FreshRSS/FreshRSS/blob/812eda1fa05e370c4c1645b5b82f09f9da2c7bf7/.travis.yml#L41-L56

I'll probably migrate it to GitHub Actions since I can't get this new Travis.com thing to work. (So I'm not sure if the configuration's even still the same as Travis.org.)

ORelio:
Okay, thanks. Will use it if I need to change the CSS again.

math-GH:
well done :) Thanks a lot

Hukuma1:
Have been waiting for this for a while. Thank you for this wonderful addition to the project!

cosmo-kray:
Thank you for this feature ! I've noticed there is no thumbnail for youtube channel feeds. Do you think you could manage to enable it for youtube feeds too ?

ORelio:
What method are you using to follow a YouTube feed?
Personally I'm using [RSS-Bridge](https://github.com/RSS-Bridge/rss-bridge) to generate RSS from YouTube so my feeds have thumbnails.

cosmo-kray:
Oups, sorry. Problem was on my side. Looks like the **Youtube extension** was causin this bug in my feeds.

PR #3774, opened 2021-08-16
Modify bright elements in adark.css mobile view

Poorchop:
Elements with bright backgrounds in mobile view now have a dark background.

Changes proposed in this pull request:

- Style certain elements to be more compatible with the dark theme

How to test the feature manually:

Open your browser's developer tools and implement the styles shown here.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #3806, opened 2021-08-26
Darken bright elements in adark.css

Poorchop:
- Table headers and box titles on the subscription management page use bright white backgrounds. This modifies these elements to use a dark background in order to be consistent with the rest of the theme.

- Darken bright elements in adark.css

1. Load the adark.css theme and modify the stylesheet accordingly using the browser's stylesheet editor.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -64,11 +64,11 @@ table {
 
 tr, th, td {
 	padding: 0.5em;
-	border: 1px solid #ddd;
+	border: 1px solid #333;
 }
 
 th {
-	background: #f6f6f6;
+	background: #262626;
 }
 
 form td,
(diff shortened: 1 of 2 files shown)


Frenzie (approved):

Frenzie:
PS Before/after screenshots are helpful to add. ^_^

Alkarex:
Could you please do the same in
https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/Alternative-Dark/adark.rtl.css ?

Poorchop:
Shouldn't I have also done this for #3774 in this case?

Poorchop:
I also think that I should have made the table border dark like so. If you agree, then I'd like to make this change before this gets merged.

screen shot

Frenzie:
Ftr, you shouldn't (or in any case don't have to) do it manually. `make rtl` takes care of that.

Alkarex:
> I also think that I should have made the table border dark like so. If you agree, then I'd like to make this change before this gets merged.

Fine for me :-)

PR #3794, opened 2021-08-23
fix broken sharing services

math-GH:
Closes #3793 

Changes proposed in this pull request:

- Blogotext: deleted
- Lemmy: set link
- Movim: changed link
- mastadon: set link
- reordered array

How to test the feature manually:
1. go to integration/share settings
2. edit/add the services
3. check the help buttons

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

Issue #3793, opened 2021-08-22
[BUG] Sharing/Integration: broken "more information" links


math-GH:
Configuration: Sharing
![grafik](https://user-images.githubusercontent.com/1645099/130353065-5df4ace2-e97b-4fd2-8bea-dccd9df527a6.png)

"more information" button:

- Blogotext: link (https://lehollandaisvolant.net/blogotext/fr/) is broken/useless. Leads to a 404 page. I guess this sharing we could delete. The last release was in 2017 (see https://github.com/BlogoText/blogotext )
- Lemmy: link leads to the sharing paged itself. The link is useless. Is it this Lemmy https://github.com/LemmyNet/lemmy ?
- Movim: link leads to github page. A better link could be https://movim.eu/

**Environment information (please complete the following information):**
 - FreshRSS version: 1.18.1

PR #3812, opened 2021-08-30
 fix broken sharing services (part 2): delete Blogotext

math-GH:
Closes a part of #3793

sub PR of #3794 as @Alkarex suggested

- remove: the blogotext sharing option

check the sharing options

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


math-GH:
What is the plan with this PR?
Set it on hold?
Merge it and wait if blogotext is still needed?

app/shares.php
```diff
@@ -30,13 +30,6 @@
 		'form' => 'advanced',
 		'method' => 'GET',
 	),
-	'blogotext' => array(
-		'url' => '~URL~/admin/links.php?url=~LINK~',
-		'transform' => array(),
-		'help' => 'http://lehollandaisvolant.net/blogotext/fr/',
```
Frenzie: I don't know anything about blogotext, but I will make the semantic point that unmaintained for a little while isn't the same thing as broken. This help link is broken though; presumably that should be something like https://github.com/BlogoText/blogotext instead.

So I'm on the side of keep. It's not like something with fixes for PHP 7.3/8.0 is something hugely outdated that'll only run on 5+ year old systems, even if it might end up there a few years down the line.

math-GH: feel free to refuse this PR.

feel free to fix the link.

As I feel free to delete old stuff that does not make sense to maintain. Blogotext creates 2 PR where blogotext code had to change (1 PR that reordered the array, 1 PR that will fix the link). I do not like technical debts.

If we do not want to delete it I would suggest to mark blogotext as deprecated with the hint that users of FreshRSS and blogotext should sent a message to this issue to mark it as "still in use".

Alkarex: One of the issues is that we have, by design, no telemetry for which features are used, and very little feedback from our user-base, so I tend to be conservative and wait a bit more before deleting things. I can see some user activity from April this year on https://github.com/BlogoText/blogotext/issues
Having a deprecated indication could be interesting.

math-GH:
Blogotext as sharing service in FreshRSS will be marked as deprecated in 1.20.0 and is planned to be removed in 1.22.0 (see #4269)

math-GH:
@Alkarex Is there any reason for postponing this PR from 1.22 to 1.23?

Alkarex:
> @Alkarex Is there any reason for postponing this PR from 1.22 to 1.23?

@math-GH It was mostly to wait a bit more, with the same reasoning as https://github.com/FreshRSS/FreshRSS/pull/3812#discussion_r702465903 . I would rather merge at the start of a new development phase, to get some time to hear whether anyone complains

math-GH:
> > @Alkarex Is there any reason for postponing this PR from 1.22 to 1.23?
> 
> @math-GH It was mostly to wait a bit more, with the same reasoning as [#3812 (comment)](https://github.com/FreshRSS/FreshRSS/pull/3812#discussion_r702465903) . I would rather merge at the start of a new development phase, to get some time to hear whether anyone complains

Is now a good time?

Frenzie (approved):
We're now a few years down the line.

Alkarex:
Sorry I forgot about it. Could be merge right after the release of 1.24?

math-GH:
see #6225

PR #3817, opened 2021-09-01
Improved <kbd> line break

math-GH:
before:
`<kbd>` has no automatic line break. It leads to a horizontal scrolling on smaller screens
grafik

grafik

improved:
line break with text-indent.
grafik

grafik

- CSS changed
-
-

go to the both pages as shown in the screenshots

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -74,12 +74,12 @@ sup {
 
 kbd {
 	background-color: #eee;
-	padding: 2px 4px;
+	padding: 2px 4px 2px 24px;
 	display: inline-block;
 	color: #333;
 	border: 1px solid #b4b4b4;
 	border-radius: 3px;
-	white-space: nowrap;
+	text-indent: -20px;
 }
 
 /*=== Images */
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -80,6 +80,8 @@ kbd {
 	border: 1px solid #b4b4b4;
 	border-radius: 3px;
 	text-indent: -20px;
+	white-space: pre-wrap;
+	overflow-wrap: anywhere;
 }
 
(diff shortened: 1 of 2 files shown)


p/themes/base-theme/template.css
```diff
@@ -74,12 +74,12 @@ sup {
 
 kbd {
 	background-color: #eee;
-	padding: 2px 4px;
+	padding: 2px 4px 2px 24px;
 	display: inline-block;
 	color: #333;
 	border: 1px solid #b4b4b4;
 	border-radius: 3px;
-	white-space: nowrap;
```
Frenzie, line 82: For clarity it might be preferable to replace this with `pre-wrap` instead of what's proposed in this PR, although I do like the look with indentation.

Frenzie: Incidentally, that'd also take care of everything automagically. You'll still have to add for example `overflow-wrap: anywhere` and/or `overflow: auto` to make this PR fully do what it promises. (I.e., try with `abunchofreallylongtextwithoutaslashorsomething`).

math-GH:
improved with the ideas from @Frenzie 

grafik

Frenzie (commented):
lgtm in principle, order still needs a look with stylelint though ^_^

Alkarex:
Quick note, `<kbd>` is also used by articles we retrieve, not only for our own UI

Frenzie:
That's a feature, not a bug. ;-) That means they can currently also overflow with ugly results, in theory anyway. It seems unlikely in practice. Regardless, this PR fixes the issue everywhere.

Alkarex:
https://github.com/FreshRSS/FreshRSS/commit/68dc18a7aa2d9594b0120d1754ab6f6340fc93f1

Issue #448, opened 2014-03-07
Use only one unit in CSS for font-size


aledeg:
To be consistent, it would be better to have only one unit for font-size. REM would be the best choice.
Maybe we can extend that to all units used in CSS files.

PR #3818, opened 2021-09-01
fix: Theme slider on small screens

math-GH:
Theme slider

before:
the right arrow overflows the box on small screens
grafik

after:
grafik

- CSS (width of the arrows improved)

1. on smaller screens

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -1180,7 +1180,7 @@ br {
 .nav label {
 	padding: 0;
 	display: none;
-	width: 10%;
+	width: 20%;
 	height: 100%;
 	color: #fff;
 	font-family: "Varela Round", sans-serif;
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -1187,7 +1187,7 @@ br {
 .nav label {
 	padding: 0;
 	display: none;
-	width: 20%;
+	width: 65px;
 	height: 100%;
 	color: #fff;
 	font-family: "Varela Round", sans-serif;
--- a/p/themes/base-theme/template.rtl.css
+++ b/p/themes/base-theme/template.rtl.css
@@ -1180,7 +1180,7 @@ br {
 .nav label {
 	padding: 0;
 	display: none;
-	width: 10%;
+	width: 20%;
 	height: 100%;
 	color: #fff;
 	font-family: "Varela Round", sans-serif;
--- a/p/themes/base-theme/template.rtl.css
+++ b/p/themes/base-theme/template.rtl.css
@@ -1187,7 +1187,7 @@ br {
 .nav label {
 	padding: 0;
 	display: none;
-	width: 20%;
+	width: 65px;
 	height: 100%;
 	color: #fff;
 	font-family: "Varela Round", sans-serif;


p/themes/base-theme/template.css
```diff
@@ -1180,7 +1180,7 @@ br {
 .nav label {
 	padding: 0;
 	display: none;
-	width: 10%;
+	width: 20%;
```
Frenzie: I don't know if this `.nav label` selector occurs anywhere else, at a quick glance it looks like `70px` for example might make more sense? (Where the `px` as opposed to `%` is the salient point, not the exact value.)

math-GH: I do not understand why it makes more sense to switch from % to px.

Could not find another place where `.nav label` is used

math-GH: The question is, why `<label>` is used for prev/next.
IMHO it is not semantic correct.
(But: It should not be the point for this PR)

Frenzie: > I do not understand why it makes more sense to switch from % to px.

Because 10% is clearly meant as 10% of 640px = 64px. That's when it looks good.

That's what's causing the problem. The reason 20% works is because 20% of 250px (the min-width) is 50px while 10% is only 25px. But then when you get to wider screens suddenly we're talking 128px instead of 64px. It doesn't look horrible but it definitely looks different.

math-GH: I know `git blame`

math-GH: I "just" doubled the previous setting (`10%` => `20%`).

I prefered the % because the font size of the arrows is also in %
https://github.com/FreshRSS/FreshRSS/blob/68dc18a7aa2d9594b0120d1754ab6f6340fc93f1/p/themes/base-theme/template.css#L1194

Frenzie: Those are very different percentages. That's 0.9 * 1000% of what almost everyone leaves at the default font size of 16px, i.e., "always" 144px. This is actually a bit of a case of https://github.com/FreshRSS/FreshRSS/issues/448, albeit not one that's likely to ever become problematic.

The other value is as I said above, a percentage of the width.

To unify it so that they would scale together appropriately, that should be:

```css
font-size: 9rem; /* usually 144 px */
width: 4rem; /* usually 64px */
```

But whatever value rolled out of that (e.g., a user who sets it at 12px would get 108px and 48px), it'd always be in sync. A percentage of a font size and a percentage of a width won't do that. ;-)

math-GH:
did the change from % to px. Thanks to @Frenzie for discussing this "little" thing.

p/themes/base-theme/template.css
```diff
@@ -1187,7 +1187,7 @@ br {
 .nav label {
 	padding: 0;
 	display: none;
-	width: 10%;
+	width: 65px;
```
Frenzie, line 1190: 65 on purpose instead of 64?

Frenzie (approved):
lgtm

p/themes/base-theme/template.css
math-GH: yes. I like numbers that ends on 0 or 5, if there is not a very strong reason to use another one.

Frenzie: The reason is that 64 = 4 * 16. I don't have some weird attachment to that (hence "lgtm") but it is the logical value. ;-)

Issue #3129, opened 2020-07-27
Responsive is broken with a table inside a feed (a global scroll bar appears)


Massedil:
**Reproduce :**

Subscribe to : 
https://btrfs.wiki.kernel.org/index.php?title=Special:RecentChanges&feed=atom

Look a the "24 juillet 2020 à 14:21" article.

|---|---|
|---|---|
![](https://user-images.githubusercontent.com/20357783/88554936-68df7b80-d027-11ea-97d1-45033e9ca40a.png) | ![](https://user-images.githubusercontent.com/20357783/88555368-05098280-d028-11ea-8d1d-6d64616c9680.png)

**Expected :**

The `<table>` is inside a `overflow: auto;` element. Maybe the `<div class="text">` enclosing the `<table>` element can take the `overflow: auto;` CSS property.

![](https://user-images.githubusercontent.com/20357783/88555701-6c273700-d028-11ea-8c48-043cc2e46715.png)

Alkarex:
PR welcome

PR #3819, opened 2021-09-01
Fixed: overflowing tables on small screens

math-GH:
Settings: Display

before:
Article icons table overflows the window on small screens
grafik

after:
table gets scroll bars on small screens
grafik

no scroll bars on bigger screens
grafik

- CSS

1. go to settings: Display.
2. change the window widths

This PR works best with PR #3818

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -172,7 +172,10 @@ button.as-link[disabled] {
 
 /*=== Tables */
 table {
+	display: block;
 	max-width: 100%;
+    overflow-x: auto;
+    white-space: nowrap;
 }
 
 th.numeric,
@@ -206,7 +209,6 @@ td.numeric {
 .form-group .group-controls {
 	min-width: 250px;
 	margin: 0 0 0 220px;
-	overflow-x: auto;
 }
 
 .form-group .group-controls .control {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -206,7 +206,7 @@ td.numeric {
 .form-group .group-controls {
 	min-width: 250px;
 	margin: 0 0 0 220px;
-	overflow: auto;
+	overflow-x: auto;
 }
 
 .form-group .group-controls .control {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -206,6 +206,7 @@ td.numeric {
 .form-group .group-controls {
 	min-width: 250px;
 	margin: 0 0 0 220px;
+	overflow: auto;
 }
 
 .form-group .group-controls .control {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -172,10 +172,7 @@ button.as-link[disabled] {
 
 /*=== Tables */
 table {
-	display: block;
 	max-width: 100%;
-    overflow-x: auto;
-    white-space: nowrap;
 }
 
 th.numeric,
@@ -209,6 +206,7 @@ td.numeric {
 .form-group .group-controls {
 	min-width: 250px;
 	margin: 0 0 0 220px;
+	overflow-x: auto;
 }
 
 .form-group .group-controls .control {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -171,6 +171,10 @@ button.as-link[disabled] {
 }
 
 /*=== Tables */
+.table-wrapper {
+	overflow-x: auto;
+}
+
 table {
 	max-width: 100%;
 }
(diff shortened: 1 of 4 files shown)


Frenzie:
Somewhat related to https://github.com/FreshRSS/FreshRSS/pull/3805. ^_^

Frenzie (approved):
lgtm

p/themes/base-theme/template.css
```diff
@@ -206,6 +206,7 @@ td.numeric {
 .form-group .group-controls {
 	min-width: 250px;
 	margin: 0 0 0 220px;
+	overflow: auto;
```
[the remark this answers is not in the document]
Frenzie: Right, but it's clearer about intent. (Assuming I'm reading it correctly.) :-)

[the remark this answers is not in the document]
math-GH: good point.

changed it.

math-GH:
the 3bc9447 fix all tables, also the manage users user list

grafik

math-GH:
Also (finally) fix overflowing tables in the feeds (f.e. see #3129 )

grafik

p/themes/base-theme/template.css
```diff
@@ -172,7 +172,10 @@ button.as-link[disabled] {
 
 /*=== Tables */
 table {
+	display: block;
```
Frenzie: A table with display: block will not behave correctly because you're telling it not to behave like a table. It doesn't look like you'll be able to do this without a container element.

table-layout: fixed might also help, potentially. Give it a try or remind me another day. ;-)

p/themes/base-theme/template.css
```diff
@@ -172,7 +172,10 @@ button.as-link[disabled] {
...
 	max-width: 100%;
+    overflow-x: auto;
```
Alkarex: Minor inconsistent indent whitespace

p/themes/base-theme/template.css
Frenzie: I did some tests and there doesn't seem to be a way to avoid a container element to see fully correct behavior. If there were a decent method I overlooked I imagine they'd be using it over at https://getbootstrap.com/docs/5.0/content/tables/#always-responsive for example.

We could regard the proposal in this PR as a necessary evil since the internals of the table seem to behave properly, at least at first glance, even with colspan and rowspan. That means the crime's hopefully hidden as long as you ensure background and border on the `<table>` element itself are transparent.

Screenshot_2021-09-02_10-01-19

I'd probably feel better simply using a container element, but then for feed content question becomes whether playing around with `cloneNode()` for example is a bigger evil. It wouldn't be much effort in any case. But the proposal from #3129  also sounds decent enough and very clean, or are there any issues with that?

For reference, here's a simple table.
```html
<table>
	<thead>
		<tr>
			<th colspan="2">Table header</th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<td rowspan="2">Table body</td>
			<td>Second column</td>
		</tr>
		<tr>
			<td>Second column</td>
		</tr>
	</tbody>
</table>
```

math-GH: on which browser `display:block` does not work?

source of my solution: https://stackoverflow.com/questions/5533636/add-horizontal-scrollbar-to-html-table

Frenzie: That page even lists additional problems I wasn't aware of, most notably https://www.tpgi.com/short-note-on-what-css-display-properties-do-to-table-semantics/ 

Here are a few quick examples showing some ways in which display: block does not behave how a table should behave.

[table.zip](https://github.com/FreshRSS/FreshRSS/files/7100531/table.zip)

There are more of course. You can stick a width on the surrounding block all you want, the table contents won't grow or shrink properly.

math-GH: OK, I will revert to 8412836

But it will not fix #3129

For the user list (see https://github.com/FreshRSS/FreshRSS/pull/3819#issuecomment-910741397) I will try to find another solution.

Frenzie: Simply a container around it will do the trick.

math-GH:
fixed (scrollable tables):
- articles icons
- user list
- extension list

Frenzie (approved):
lgtm!

Alkarex:
> Also (finally) fix overflowing tables in the feeds

The final fix does not do that, does it?

Frenzie:
@Alkarex No, as stated above display: block is somewhat problematic with regard to visual table behavior and apparently quite problematic with regard to table accessibility.

But we can easily stick that `<div class="table-wrapper">` around all tables in feeds, whether in JS or PHP.

Basically something like:

```js
// for loop
const table = tables[i];
const tableWrapper = document.createElement('div');
tableWrapper.classList.add('table-wrapper');
tableWrapper.appendChild(table.cloneNode());
table.parentNode.replaceChild(tableWrapper, table);
```

PR #3827, opened 2021-09-05
Settings: User queries - layout upgrade

math-GH:
before:
strange layout
grafik

after:
box layout mostly similar to subscription management:
grafik

- use the box layout
- green alert box layout -> neutral layout (yellow and red alert box did not changed)

1. go to user queries page
2. drag and drop is not changed
3. edit the query is not changed

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


math-GH:
Right now it is "just" a draft. The phtml file consits of any commented out lines of code (TODO).

Please have a look and give feedback if it is worth to go ahead

Frenzie:
The box content is supposed to hide the list aspect? (I.e., no * or borders.) 

Looks fine to me, but I think @Alkarex should have more to say.

Alkarex:
We need at least to see the configuration wheel for each query. Ping @aledeg , who did most of this part

math-GH:
The cog is still there.
Sorry, forget to show it

grafik

aledeg:
I like what you've done. When I've started that, I wanted to reuse what was existing. I am not really good at designing interface. I am ok to change that as long as the features are equivalent.

math-GH:
ready for review

now it looks like:
grafik

Frenzie:
Having to hover doesn't seem like a good idea. This is in the settings after all.

math-GH:
> 
> 
> Having to hover doesn't seem like a good idea. This is in the settings after all.

same as in:
grafik

math-GH:
the yellow alert message box got an little update
grafik

Frenzie:
Hover to show that you can change settings strikes me as questionable in subscription management as well, but fair enough.

math-GH:
> 
> 
> Hover to show that you can change settings strikes me as questionable in subscription management as well, but fair enough.

I do not like the hidden-and-hover setting buttons too.
If you agree I would create another PR that deletes the hover effect (it is used in all/most of the themes) but not here in this PR

Frenzie:
Sounds good to me, but you might want to wait for some other opinions/ideas first. ^_^

Alkarex:
> > Hover to show that you can change settings strikes me as questionable in subscription management as well, but fair enough.
> 
> I do not like the hidden-and-hover setting buttons too.
> If you agree I would create another PR that deletes the hover effect (it is used in all/most of the themes) but not here in this PR

Yes, I think that would improve accessibility

math-GH:
OK. I will do it.

Any thoughts about this PR? ;)

math-GH:
With #3831 the cog wheel icon is always visible
grafik

aledeg:
Looks good.
The only thing I found quite strange is the `next` icon used for the list. I was expecting that it will open some panel or something similar. I think it could be less weird by using the standard signs for list.
My 2¢

Alkarex:
The PR is quite fine as it is, but just like @aledeg , I think a minor improvement could be not to use the "next" icon. I would suggest a native list style.
Also, it looks like the list is draggable, but I do not think we allow so, right?

P.S. We can also merge as it is and leave those minor points for future improvements

math-GH:
I will work on this suggested improvements. :)
yes, the list is draggable. It was also before draggable (and I will think about to make it more visible)

math-GH:
Theme Origin looks now:
grafik

(the others are wip)

math-GH:
All Themes got now the update.

Ready for review

Alkarex (approved):

PR #3842, opened 2021-09-11
Mailer rename .txt to .txt.php

Alkarex:
I do not know why this file was `.txt` instead of `.phtml` but it prevented automatic i18n lint to work properly.
Found during https://github.com/FreshRSS/FreshRSS/pull/3841


marienfressinaud (approved):
It was `.txt` because the mailer sends emails as text :) (see [Mailer.php](https://github.com/FreshRSS/FreshRSS/blob/edge/lib/Minz/Mailer.php#L97)).

I suggest:

- change to `$this->mailer->isHTML(true)` in Mailer class and surround text by `<p>` tags in the view file
- or to find a different solution so the linter handles the `txt` files as well
- or to keep your version, but it can be confusing not to have HTML in a `phtml` file

All the solutions are fine to me, but I find the second solution more robust and better. I don't know what are the impacts though. 

Additional note: it's also possible to set both HTML and text bodies to the mailer, but I think it's out of the scope of this PR:

```php
$mail->isHTML(true);
$mail->CharSet = 'utf-8';
$mail->Subject = $subject;
$mail->Body = $html_body;
$mail->AltBody = $txt_body;
```

Alkarex:
Ah, I see. We already have some `.phtml` files generating other document types though, such as JSON or RSS.
The content of the file is PHP, so the file extension should end by `.php` or `.phtml` to have proper support by default of syntax check, syntax highlighting, linting, etc. So it is more than just our own i18n tools.
In other projects, when generating something else than HTML, I use extensions like `.txt.php`, `.json.php`, `.png.php`, etc.
Here I could then suggest `.txt.php` if you like?

marienfressinaud:
>  In other projects, when generating something else than HTML, I use extensions like .txt.php, .json.php, .png.php, etc. Here I could then suggest .txt.php if you like?

That would be even better :)

Alkarex:
Done https://github.com/FreshRSS/FreshRSS/pull/3842/commits/690be8d5bb33e4e59e231ca46b3008eca61e6824

PR #3846, opened 2021-09-12
i18n: DE - switched from informal designation to formal

math-GH:
In German we have 2 types of "you":
- informal ("Du")
- formal ("Sie")

The German translation uses in common the formal designation, except 2 pieces of labels, that are now fixed too

- i18n/de

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/i18n/de/conf.php
+++ b/app/i18n/de/conf.php
@@ -60,7 +60,7 @@ return array(
 		'api' => 'API-Verwaltung',
 		'delete' => array(
 			'_' => 'Accountlöschung',
-			'warn' => 'Dein Account und alle damit bezogenen Daten werden gelöscht.',
+			'warn' => 'Ihr Account und alle damit bezogenen Daten werden gelöscht.',
 		),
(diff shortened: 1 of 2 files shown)


aledeg (approved):

app/i18n/de/conf.php
```diff
@@ -60,7 +60,7 @@
 		'api' => 'API-Verwaltung',
 		'delete' => array(
 			'_' => 'Accountlöschung',
-			'warn' => 'Dein Account und alle damit bezogenen Daten werden gelöscht.',
+			'warn' => 'Ihr Account und alle damit bezogenen Daten werden gelöscht.',
```
Frenzie: Presumably you could alternatively say `dieses Account`? If that's preferable.

Frenzie (approved):

app/i18n/de/conf.php
math-GH: It would be also fine to make it more clear and less personal to "Dieser Account"

PR #3847, opened 2021-09-12
Delete jQuery

### LINKED ISSUE #3845: [BUG] jQuery still used?
Is jQuery still in usage in FreshRSS, except that it is included in `./p/scripts`?

If the core does not need it anymore, let's create it as an extension.

math-GH:
Closes #3845

- delete jquery.min.js
- delete lib_phpQuery

jQuery is included just in the stats, but it looks like that we does not need it anymore.
So in general jQuery is not needed anymore in the core.

lib_phpQuery looks like unused in the core. 
Also: It looks like that the development and active usage of phpQuery stopped in 2015: https://code.google.com/archive/p/phpquery/issues/246

1. check the stats (because of jQuery)
2. and check the system in general (because of phpQuery)

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

I will publish this changes as draft first! Will see if the changes breaks the system or not.


Frenzie:
Perhaps it's an accidental leftover after https://github.com/FreshRSS/FreshRSS/pull/2234 but let me double check.

Edit: the jQuery thing lgtm in principle.

Frenzie:
> delete lib_phpQuery

I'd rather you didn't try to combine such completely unrelated things in the same PR.

Frenzie (approved):
Definitely used:
https://github.com/FreshRSS/FreshRSS/blob/50ba6bbe07b0bb86eb07e3212ba2e22cb2878cf2/app/Models/Entry.php#L432-L454

Frenzie (changes requested):
Eek, meant the other one.

Alkarex:
phpQuery is still used for full article content retrieval and has been patched by us along the way, e.g. to support PHP8.
We indeed need a replacement (I think there is an issue somewhere but I cannot find it at the moment). I would like something lighter using some more native PHP functions such as XPath.
Let's restrict this PR to jQuery.

math-GH:
Reverted lip_phpQuery

Alkarex:
Let's go for it 🙂

PR #3858, opened 2021-09-25
Stats: replace flotr2 with chart.js

math-GH:
- replace flotr2.js with chart.js

(Flotr2 is not anymore supported. Chart.js looks as a good replacement)

1. go to the stats

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


Frenzie (approved):
Cool, thanks!

I would like to emphasize however that not having received any updates recently is quite irrelevant for something that Just Works™. Old is not broken.

app/Controllers/statsController.php
```diff
@@ -49,22 +49,43 @@ private function convertToPieSerie($data) {
 	 *
 	 * It displays the statistic main page.
 	 * The values computed to display the page are:
-	 *   - repartition of read/unread/favorite/not favorite
-	 *   - number of article per day
-	 *   - number of feed by category
-	 *   - number of article by category
-	 *   - list of most prolific feed
+	 *   - repartition of read/unread/favorite/not favorite (repartition)
```
Frenzie, line 52: Btw, if you're touching all these lines anyway, you might as well replace it all with a correct word like `distribution` instead. Répartition is French.

math-GH: Repartition is also English ;)
https://www.wordreference.com/definition/repartition

math-GH:
One question left:
every JS file starts with this lines:
```
// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-3.0
"use strict";
/* jshint esversion:6, strict:global */
```

what is `@license magnet:`?

app/Controllers/statsController.php
Frenzie: I'm well aware that it is *a* word. It's a false friend. La répartition des articles = distribution of articles (or: article distribution).

Other words that can be used depending a bit on context are breakdown and division. E.g., breakdown by hour instead of distribution by hour.

https://fr.wikipedia.org/wiki/Pyramide_des_%C3%A2ges
> La pyramide des âges représente la répartition par sexe et âge de la population à un instant donné.

https://en.wikipedia.org/wiki/Population_pyramid
> Population distribution by age and sex for Angola in 2005

Or more directly in a translation dictionary:
https://en.pons.com/translate/french-english/r%C3%A9partition#result-sig-61618324-0

> la répartition des revenus en France
    income distribution in France

Whether the word exists is not in question. You won't find this sense in COCA (and yes, I checked to be sure; mainly computers, a bit repartition of Jerusalem/Ireland).

Frenzie:
https://github.com/FreshRSS/FreshRSS/pull/2576

For future reference, git blame will tell you these things.

Alkarex:
Thanks for this nice effort 👍🏻 

There seems to be a few bugs with some graphs, e.g. the (apparently) empty ones below, and some numbers do not seem right:
image

Comparison with before (but I can see that the averages were already wrong, or I do not understand them):
image


`chart.js` should probably be called `chart.min.js` (I believe this also happens to be the original name of the minified version)

Can `flotr2.min.js` be removed?

A few commands to run before commit (also available via `make` and/or integrated in editor ([examples](https://marketplace.visualstudio.com/search?term=phpcs&target=VSCode)) :

```sh
phpcbf -p -s .
phpcs -p -s .
```

app/Controllers/statsController.php
math-GH: let's do this i18n thing in another PR. I have also other things in mind to correct

Frenzie: No problem.

math-GH:
> There seems to be a few bugs with some graphs, e.g. the (apparently) empty ones below, and some numbers do not seem right:

strange. It works fine for me:
grafik

math-GH:
OK, now I have the same issue within another installation.

Will check it and till then I will switch PR to draft mode

math-GH:
that was an long long way to find the simple issue. It is fixed now.
Ready for review.

Alkarex:
> `chart.js` should probably be called `chart.min.js` (I believe this also happens to be the original name of the minified version)
> 
> Can `flotr2.min.js` be removed?
> 
> A few commands to run before commit (also available via `make` and/or integrated in editor ([examples](https://marketplace.visualstudio.com/search?term=phpcs&target=VSCode)) :
> 
> ```shell
> phpcbf -p -s .
> phpcs -p -s .
> ```

I just implemented my comments from before https://github.com/FreshRSS/FreshRSS/pull/3858#issuecomment-927337989

p/scripts/statsWithChartjs.js
```diff
@@ -0,0 +1,194 @@
...
+					borderColor: '#0062BD',
+					data: dataBarChart,
+					barPercentage: 1.0,
+					categoryPercentage: 1.0,
+					order: 2,
+				},
+				{
+					// average line chart
+					type: 'line',
+					label: labelAverage,  // Todo: i18n
```
Inverle, line 136: Where is this label shown?

Frenzie: Looking at <https://www.chartjs.org/chartjs-plugin-annotation/1.2.0/samples/line/average.html> might suggest it used to be displayed in a similar manner?
image

Inverle: Yes there is a similar line, but nothing is displayed on it.

image

Frenzie: Yup, that's what I mean. Maybe it used to be there in an older version and you have to define it slightly differently now if you want it displayed. I can't recall though. (Also it seems fairly obvious without a label.)

Inverle: Still isn't shown in 1.19.0

image

Frenzie: Back in Flot? :sweat_smile:  In any case, I have no attachment  to it. ;-)

Inverle: No, this is chart.js
1.18.0 still used flotr, but there it worked:

image

PR #3861, opened 2021-09-27
Detect enclosure as images using file extension

Alkarex:
Support e.g. https://www.sciencesetavenir.fr/atom.xml , which does not include any type hint:

```html
<link rel="enclosure" href="https://example.net/assets/img/ZZZ.jpg" title="YYY" />
```

Useful for https://github.com/FreshRSS/FreshRSS/pull/3805

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -411,7 +411,7 @@ class FreshRSS_Feed extends Minz_Model {
 						$width = $enclosure->get_width();
 						$length = $enclosure->get_length();
 						if ($medium === 'image' || strpos($mime, 'image') === 0 ||
-							($mime == '' && $length == null && ($width != 0 || $height != 0 || preg_match('/[.](gif|jpe?g|png|svg)$/i', $elink)))) {
+							($mime == '' && $length == null && ($width != 0 || $height != 0 || preg_match('/[.](avif|gif|jpe?g|png|svg|webp)$/i', $elink)))) {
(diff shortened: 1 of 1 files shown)


app/Models/Feed.php
```diff
@@ -410,7 +410,8 @@ public function loadEntries($simplePie) {
 						$height = $enclosure->get_height();
 						$width = $enclosure->get_width();
 						$length = $enclosure->get_length();
-						if ($medium === 'image' || strpos($mime, 'image') === 0 || ($mime == '' && $length == null && ($width != 0 || $height != 0))) {
+						if ($medium === 'image' || strpos($mime, 'image') === 0 ||
+							($mime == '' && $length == null && ($width != 0 || $height != 0 || preg_match('/[.](gif|jpe?g|png|svg)$/i', $elink)))) {
```
Frenzie: Should probably also add WebP and that new one based on AV1.

Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/3861/commits/5fd913d0c8c56638b2998b1359bac4a0e676d620

Frenzie (approved):
I forgot about APNG earlier btw, and fwiw there's also BMP.

Alkarex:
> I forgot about APNG earlier btw, and fwiw there's also BMP.

Indeed, and several more, cf. e.g. https://developer.mozilla.org/docs/Web/Media/Formats/Image_types , but this is just a fallback for cases not providing any other type hint

PR #3871, opened 2021-10-09
fix #3661: not logged in user: button to mark all as read leads to HTTP403 error

### LINKED ISSUE #3661: [BUG]  anonymous Reading: 403 Error while mark all as read
**Describe the bug**
403 error

**To Reproduce**

1. anonymous Reading (not logged in)
2. scroll to the end of feed list:
2. https://demo.freshrss.org/i/?a=normal&get=f_3 
grafik

3. click on "Märk all as read"
4. get the error 403 
grafik

"Sie haben nicht die Berechtigung, diese Seite aufzurufen [CSRF]" => You do not have the permission to open this page [CSRF]

I do not have a clue why it is in German but the rest of texts are in English

**Expected behavior**
no Error message. If I am not allowed to mark all as read than the button should not appear,

 - Browser: Firefox 89
 - FreshRSS version: 1.18.0 demo.freshrss.org

math-GH:
Closes #3661

- if user is not logged in (=default/guest user) than no button to mark all articles as read. Show only a text

1. with default/guest user not logged in
2. scroll to the end of articles list
3. see: "There are no more articles" instead of "mark all as read" button
grafik

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


app/views/helpers/pagination.phtml
```diff
@@ -25,7 +25,8 @@
 		<a id="load_more" href="<?= Minz_Url::display($url_next) ?>">
 			<?= _t('gen.pagination.load_more') ?>
 		</a>
-	<?php } elseif ($url_mark_read) { ?>
+	<?php } elseif ($url_mark_read) {
```
Frenzie: Seems you could just do this one-line change instead?
```suggestion
	<?php } elseif (FreshRSS_Auth::hasAccess() && $url_mark_read) { 
```

math-GH: I thought it too.
But: No, it does not work properly.
Because of the `else`

Frenzie: How do you mean because of the else? My entire point is that it looks like you're unnecessarily duplicating the else.

math-GH: now I see your point.

Thanks!

math-GH: fixed it with the commit

Frenzie (approved):

app/views/helpers/pagination.phtml
```diff
@@ -36,9 +36,9 @@
 			<?= _t('gen.pagination.mark_all_read') ?>
 		</button>
 	<?php } else { ?>
-		<a id="bigMarkAsRead" href=".">
+		<div id="bigMarkAsRead" href=".">
```
Alkarex: `<div href` is probably an error?

Alkarex (changes requested):
Cf. comment

app/views/helpers/pagination.phtml
math-GH: I would delete the `href`, because it is not a real link for the user

PR #3878, opened 2021-10-12
logo with uppercase letters

math-GH:
I suggest to fresh-up the logo.

inspired by "Mapco" theme

Before:

After:

- (same logo, same colors, just the word "FreshRSS" changed)
- uppercase letters
- a little bit smaller font-size
- fresher font-family

Some examples with the themes:
grafik

grafik

grafik


Alkarex:
I do not really mind, but maybe others have style feedback.
@marienfressinaud ?

Frenzie:
I'm less fond of it all caps but I'm not opposed.

marienfressinaud:
I wonder why you want to change the logo: changing a logo is also changing the identity (e.g. the new logo is more assertive, more "robust", but I think that it conveys less the idea of **fresh**ness… maybe?) Also, if we change the logo here, we should change the website as well. So, unless this change is done within a rebrand of the identity with good reasons to change it, I would be rather against this modification (but please consider this as a "conservative" opinion, I would not mind if you all decide to change it anyway ^^).

That being said, I would be happy to have a "fresher" logo, but I would be happier if it's done with a graphic designer who designs it to convey ideas through it :)

math-GH:
I like this discussion.

Why I started with the idea of this small rebranding?
It came from: the UI of FreshRSS 1.18 does not take care of a full stringent logo use.
It sounds strange. But I will show you:

Origine theme looks like this:
grafik
(fine. It looks like usual known)

But: The admin can rename (!) the logo (!)
grafik

So: What is the real logo of FreshRSS? Is it just the blue circle? Or is it the cirle+"FreshRSS"?

Other themes changes also the logo. Not only the color, also capitalize the name and uses a special font-family, like "Mapco" theme:
grafik

The theme "Blue Lagoon" changed also the circle style
grafik

So: till then there is no stringent logo used

That was the reason why I came up with the idea to propose a little fresh-up of the logo. The change to the current logo is minimal, so that there is no must-have-right-now update of the website.

Frenzie:
Fwiw, I think making it all-caps is a much bigger change than what Blue Lagoon does in mildly adjusting the logo, the colors and the font. Furthermore, I laud and love not being stringent about concepts like "brand identity." So if the rationale is to make the themes more similar, I have to be opposed on principle regardless of my opinion on the choices made in one theme or another. ;-)

marienfressinaud:
I think it's different to change the logo in the *default* theme than to change it in the other themes created by the community.

> But: The admin can rename (!) the logo (!)

Good point! But I would ask you why don't you change the name of your instance by "FRESHRSS" then? :D

math-GH:
closed

PR #3880, opened 2021-10-12
Improve conf menu

math-GH:
- reordered the conf menu: for personal profile settings the menu entry is now on the top of navigation list
- in the left navigation in the configuration: a logout link added (so it it similar to the dropdown menu)

Why it makes sense to change:
- As a user I search: where can I change my password and how can I log out
- in most apps it is common to find this 2 point at the top of account/settings drop down list

After:
grafik

1. open the configration drop down list
2. see the navigation on the left side in the configuration

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/i18n/fr/gen.php
+++ b/app/i18n/fr/gen.php
@@ -146,7 +146,7 @@ return array(
 	),
 	'menu' => array(
 		'about' => 'À propos',
-		'account' => 'Account', // Todo: Transation
+		'account' => 'Compte',
 		'admin' => 'Administration',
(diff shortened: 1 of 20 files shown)


app/i18n/fr/gen.php
```diff
@@ -146,6 +146,7 @@
 	),
 	'menu' => array(
 		'about' => 'À propos',
+		'account' => 'Account', // Todo: Transation
```
```suggestion
		'account' => 'Compte',
```

PR #3886, opened 2021-10-14
Use strict comparision

ColonelMoutarde:
Closes #

Use strict comparison

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Controllers/subscriptionController.php
+++ b/app/Controllers/subscriptionController.php
@@ -10,7 +10,7 @@ class FreshRSS_subscription_Controller extends Minz_ActionController {
 	 * underlying framework.
 	 */
 	public function firstAction() {
-		if (!FreshRSS_Auth::hasAccess()) {
+		if (false !== FreshRSS_Auth::hasAccess()) {
 			Minz_Error::error(403);
 		}
 


app/Controllers/subscriptionController.php
```diff
@@ -10,7 +10,7 @@ class FreshRSS_subscription_Controller extends Minz_ActionController {
 	 * underlying framework.
 	 */
 	public function firstAction() {
-		if (!FreshRSS_Auth::hasAccess()) {
+		if (false !== FreshRSS_Auth::hasAccess()) {
```
Frenzie, line 13: Setting aside that this is incorrect (`=== false` or `!== true`), should this be strict at all? It seems to me you want it to fail as much as possible, i.e., when falsy, not when strictly false.

ColonelMoutarde:
"Prior to PHP 8.0.0, if a string is compared to a number or a numeric string then the string was converted to a number before performing the comparison. This can lead to surprising results... "

https://www.php.net/manual/en/language.operators.comparison.php

Frenzie:
Sorry, but I don't understand what you're getting at. The function in question always returns a boolean.

Frenzie:
If we grant for the sake of argument that we should always be using strict comparisons, then we should be using `FreshRSS_Auth::hasAccess() === true` and `FreshRSS_Auth::hasAccess() !== true`. NB I don't think this is something we should be doing.

Using `=== false` or `!== false` would make a (however unlikely) falsy value mean that the unauthorized user is treated as being logged in.

In any case, `!FreshRSS_Auth::hasAccess()` is the same thing as `FreshRSS_Auth::hasAccess() !== true` (or rather it's the same as `FreshRSS_Auth::hasAccess() != true`, but that's not a meaningful distinction in this context).

Alkarex:
There is most likely quite many places in the code to improve, but this specific original line looks fine for now, as Frenzie explained :-)

Issue #1963, opened 2018-07-21
progressive web app


qwertygc:
Hello

An idea it's implant progressive web app in fresh idea. Indeed, with the offline mode we can read without internet and we don't need to use third part app.

Frenzie:
It would be nice, although given that there's a multitude of API-compatible clients out there I'd personally peg it at fairly low priority.

PhieF:
Hi ! You could, at first, just put a manifest for us to be able to create standalone shortcuts, like nextcloud did ! even if we don't have an offline access

Frenzie:
I'm curious, which browser lacks such basic functionality?

madb1lly:
Hi all, I just suggested this is Framateam, thanks to @Frenzie for the link.

The benefit of a PWA would be that it can also store downloaded feeds and articles for offline viewing and would generally be lighter-weight than a "native" app. Also, PWAs are cross-platform, so can be deployed by any browser on any plaform that supports it, e.g. Android, Windows, iOS (I think).

Cheers :)

Frenzie:
Unless you're taking a jab at Java, I think most clients are indeed native, not "native." ;-)

madb1lly:
Hi @Frenzie, fair point, I wasn't clear - I just meant that I think the concept of native vs webapps is a bit old-fashioned when considering PWAs and current javascript engines, especially if WASM can be used in some way.

I was thinking about the possibility of a PWA for this a bit more and since for offline use Freshrss would need to have a way of storing articles - I think this may be with the DBs provided by the web browser but maybe it's possible to install others alongside, for a PWA to use. If it depends on the web browser PWA you'd need to decide which articles to store as I imagine you can't store as many with a PWA as you can on the web server or perhaps in a DB for a traditional app.

Frenzie:
I don't think there are any relevant practical limits on a PWA. It might still be slightly more limited out of the box in which case it would ask for more space after reaching certain thresholds.

yllekz:
Did FreshRSS quietly add this? I noticed on iOS if I add the main page of my FreshRSS hosted instance to the iOS home screen, it opens up as its own dedicated app, so I'm wondering if PWA functionality was added in already.

Frenzie:
I suppose that means one answer to my question from https://github.com/FreshRSS/FreshRSS/issues/1963#issuecomment-564522680 would be old iOS Safari.

proletarius101:
[...]

PR #3890, opened 2021-10-15
Add pwa

### LINKED ISSUE #2033: Add PWA support
I think it would be easyd to add a Manifest file allowing to make a Progressive Web App out of FreshRSS.

For reference : https://developer.mozilla.org/en-US/docs/Web/Apps/Progressive/Installable_PWAs

On android it would be a good alternative to using EasyRSS while not being very time consuming. I think the website works well on mobile but it would be better to have it as a separated app.

proletarius101:
Closes https://github.com/FreshRSS/FreshRSS/issues/2033 https://github.com/FreshRSS/FreshRSS/issues/1963

- Add a pwa manifest as defined in https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Installable_PWAs
-
-

1. Open FreshRSS in a mobile browser
2. Tap the `Install` button in the browser
3. An icon appears in the home screen which can open a standalone FreshRSS app

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Frenzie (approved):
lgtm

proletarius101:
Maybe wait a sec. I'm implementing a service worker

Frenzie:
A second? While I admit I haven't actively investigated (because we already have the likes of EasyRSS at our disposal) I always assumed that'd be a project that'd eat up at least several nights. :rofl:

proletarius101:
Well, you may merge it now. Service worker is more complicated than I think.

p/themes/manifest.json
```diff
@@ -2,15 +2,31 @@
...
   "icons": [
     {
       "src": "icons/favicon-256.png",
       "sizes": "256x256",
       "type": "image/png"
+    },
+    {
+      "src": "/favicon.ico",
+      "type": "image/x-icon",
+      "sizes": "16x16 64x64"
```
Frenzie: It's also 32x32 and 48x48, isn't it?

Frenzie:
We'll have to wait for @Alkarex. As per the repo settings I can only merge PRs once the CI runs successfully (which given that I or at least someone still has to port it over to GitHub Actions is currently never, hint hint if you happen to know GHA and have a half hour you'd like to spend on that :angel:).

p/themes/manifest.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "FreshRSS feed aggregator",
+  "short_name": "FreshRSS",
+  "start_url": "/i/",
```
Frenzie:
```suggestion
  "start_url": "..",
```

Frenzie (changes requested):
(withdrawing approval to prevent confusion)

app/layout/layout.phtml
```diff
@@ -13,6 +13,7 @@ if (_t('gen.dir') === 'rtl') {
 <?php $this->renderHelper('javascript_vars'); ?>
 		</script>
 		<?= self::headScript() ?>
+		<link rel="manifest" href="<?= Minz_Url::display('/themes/manifest.json') ?>">
```
```suggestion
		<link rel="manifest" href="<?= Minz_Url::display('/themes/manifest.json') ?>" />
```

p/themes/manifest.json
```diff
@@ -0,0 +1,28 @@
...
+    }
+  ],
+  "categories": [
+    "news"
+  ],
+  "display": "standalone",
+  "background_color": "#fff",
+  "theme_color": "#0062be",
+  "description": "FreshRSS is a self-hosted RSS feed aggregator."
+}
```
Alkarex:
```suggestion
}

```

p/themes/manifest.json
proletarius101: fixed

p/themes/manifest.json
proletarius101: fixed

p/themes/manifest.json
```diff
@@ -3,7 +3,6 @@
   "short_name": "FreshRSS",
   "start_url": "../i/",
   "scope": "../",
-  "orientation": "any",
```
Frenzie: Is leaving this out equivalent to any? Enforcing portrait would strike me as undesirable.

Frenzie (approved):
lgtm

p/themes/manifest.json
proletarius101: As most apps, the orientation is controlled by the system rather than the app

Frenzie: It's mainly that I don't know what prompted you to include it in the first place. I assumed there was a reason for it. ;-)

Alkarex:
> We'll have to wait for @Alkarex. As per the repo settings I can only merge PRs once the CI runs successfully (which given that I or at least someone still has to port it over to GitHub Actions is currently never, hint hint if you happen to know GHA and have a half hour you'd like to spend on that 👼).

CI fixed https://github.com/FreshRSS/FreshRSS/pull/3920

Alkarex:
I will have to look at bit more about it. I am not (yet) convinced of the advantages. This PR is probably fine as it does not "cost" much, but the related https://github.com/FreshRSS/FreshRSS/pull/3896 is much larger, with dependencies, and for an unclear advantage. Additional input welcome.

proletarius101:
> I will have to look at bit more about it. I am not (yet) convinced of the advantages. This PR is probably fine as it does not "cost" much, but the related #3896 is much larger, with dependencies, and for an unclear advantage. Additional input welcome.

This PR doesn't rely on https://github.com/FreshRSS/FreshRSS/pull/3896. Vise versa. So they can be merged independently.

- This PR is to enable [Installability](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Introduction#installability), making it native to the system
- The service worker is for [Network independence](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Introduction#network_independence), which makes reading pre-cached articles offline possible.
  - It also provides transition between poor connectivity and good connectivity

Frenzie:
A potential advantage of a PWA (not in the current poc unless I missed some more recent changes) would be that syncing read/unread status and the like could be handled by the service worker, without having interrupted or absent connections throwing spanners in the works.

Alkarex:
Ok, let's give this one a try. Yes @Frenzie , regarding the second PR, having a way to work partially offline would be interesting to look into, but that should probably require reworking the UI rendering to be more client side, potentially taking advantage of our API like an app. But the question is then whether a full app is not better for this role and there are a few good clients already.

Frenzie:
Right, I never looked into it myself because we already have EasyRSS and many others. But you can also build up to something progressively. I'm ambivalent towards it. :-) https://gomakethings.com/service-worker-strategies/

math-GH:
I love this PWA feature. Tested it with Mozilla Firefox on Android.

dminca:
tested 1.29.1 on Vanadium (Android) and it's not working as expected, it's just saved as a bookmark on the home.

Frenzie:
It's not that very different from a bookmark anyway, but I see the two separate options.
IMG_20260903_211558

dminca:
This is the difference

IMG_20260903_214258

**Fig. 1** - example with working PWA (in this case a different app called Memos)

----

IMG_20260903_214102

**Fig. 2** - FreshRSS not working PWA

It seems FreshRSS is not available as a PWA

Frenzie:
It works in Chrome and Vivaldi. You'd have to try a few other browsers to see if it's something about that specific installation or something in combination with Vanadium. I assume you mean you have seen the option to "install" in Vanadium before?

dminca:
I just downloaded Vivaldi on Android from the Play Store.

Only have the option to _add page to > Home screen_ and this mimics the same behaviour, still no PWA 🤔 (as seen in Fig. 2 I shared above)

> I assume you mean you have seen the option to "install" in Vanadium before?

yes, it's Fig. 1 from the screenshots I shared above.

Frenzie:
It's probably best to open an issue. I'm out of ideas for the moment. Maybe it's some kind of reverse proxy issue?

dminca:
No worries, will do that. Thanks a lot so far :)

LE: done, carded #9258

PR #3891, opened 2021-10-15
Add shortcut to jump to next unread article (issue #3393)

### LINKED ISSUE #3393: [Feature Request] New Keyboard Shortcut "next unread article"
The keyboard shortcut "next article" is currently not based on the read state, or on the current "view" of the user (scroll position). Instead, it probably scrolls upwards to the article the shortcut has been pressed before. See discussion here: https://github.com/FreshRSS/FreshRSS/issues/3387

As some users may want to have an option to simply jump from last read article to next unread article, I suggest to add an additional keyboard shortcut to jump to the next "unread" article in the current view, so that these users can navigate only by keyboard through their list.

stysebae:
Closes #3393 

- added a new field "Open the next unread article" in configuration page "Shortcuts" (default shortcut suggested: "h")
- features "Open the next unread article" (`next_unread_entry`) and "Open the next unread category" (`next_unread_category`) are implemented in ` p/scripts/main.js`
- added this field "Open the next unread article" in the existing languages (`app/i18n`). However, translation is missing for the following languages: cz, es, he, it, ja, oc, kr, pl, pt-br, ru, sk, tr, zh-cn.

1. Just press `h` to jump to the next unread article (if any)
2. Maintain `Alt+h` to jump to the next category with unread articles

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested: This feature seems to work. However, I noticed that navigation shortcuts with `Shift` did not work. Consequently, `Shift+h` is not implemented (at this moment, it only calls the function `next_feed` that does not work). Opening an issue for this problem (and implement correctly these features) would be better.
- [ ] documentation updated: The documentation does not describe in detail the shortcuts page and I am also not sure if it is necessary to describe it precisely, as the labels of each field are pretty clear.


math-GH:
please check, that tabs are used for indenting the source code instead of spaces

Alkarex:
Thanks and warm welcome, @stysebae 👍🏻 
For whitespace conventions in a given project, https://editorconfig.org is a nice approach.

stysebae:
Hello,
Sorry, I didn't notice that I used spaces instead of tabs; thanks @mathContao for your eagle eye. :-)
Thanks @Alkarex for fixing and for sharing this link about EditorConfig. It is useful, I will install it on my IDE. :-)

Alkarex:
@stysebae Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md :-)

rom-1:
So, I am currently running into a funny problem. I was using "Space" to navigate to the next article before this commit. This worked quite well, my view jumped to the next article.
I was happy to see your commit for a new shortcut to jump to the next unread article, which really was what I wanted to achieve. BUT. The space key is also a Chrome shortcut for scrolling down a page.

So, what happens if I 
- bind space to "jump to next article" shortcut = my view jumps to the next article
- bind space to "jump to next unread article" shortcut = my view jumps to the next article and additionally chrome scrolls down 1 page. 

So somehow, there seem to be 2 different methods right now to jump to the next article, and one of them ignores the internal Chrome behaviour. Any ideas?

Edit: Nevermind. I did a git update to receive this commit, and it looks like the scrolling behaviour with space has been altered by something else. After the git update, every time I am using space with both methods, it scrolls down a bit further. This was not the case before I did the update, but has nothing to do with your patch. Guess I will simply change my shortcut to something else then.

Alkarex:
@rom-1 Are you using the `Space` keyword in the shortcut field, and not a plain space character?

rom-1:
@Alkarex jupp, thats how I do it.
Screenshot 2021-10-27 100941

stysebae:
Hey,
That's strange, because the code is almost identical to the one implemented for the feature "Open the next article". However, I can confirm that there is a behavior change with Google Chrome, at least between v1.18.1 and commit `85b898c623ae09d5caa88041bd8ceb26d4f1defc`: in my case, when I use `Space` shortcut with Google Chrome on my local instance of FreshRSS (version with `85b898c623ae09d5caa88041bd8ceb26d4f1defc` as latest commit), it scrolls down; when I use it on my public "stable" instance of FreshRSS (Docker version, with FreshRSS v1.18.1), `Space` jumps to the next article (if this keyboard shortcut was used to do that).

I can take a look and check if I have introduced this bug. @rom-1, when did you notice this bug? after which updates/commits? We could try to find out when this bug appears.

rom-1:
Hey @stysebae - very sorry but I think I cannot help you to pin it down to a specific commit. I was not very precise above regarding my update procedure. 
As far as I know, I was on the latest stable version 1.18.1 (jumping with space worked perfectly), and used the FreshRSS internal updater to update to "1.19.0-dev" (additional scrolling by Chrome) just to receive your commit. So, there were a lot of commits in between which I did skip.

stysebae:
Hey @rom-1,
This bug seems to be present before my changes were accepted and merged on `edge`: I ran locally [this FreshRSS instance](https://github.com/FreshRSS/FreshRSS/commit/f96563ebb0c1e38a320eda1e7fdd8490785cd047) (commit `f96563ebb0c1e38a320eda1e7fdd8490785cd047`) and the bug is already present in this version.

I think we could open a new issue for solving this problem.

PR #3893, opened 2021-10-15
enhancement subscription management page

### LINKED ISSUE #3257: enhancements: show "show only feeds with errors" at top  and only if errors exist?
can the "show only feeds with errors" be shown at top?
I have too scroll to bottom with is to long because of the quantity of feeds I have.
also it would be nicer to show that only if a feed (or feeds) has errors.

math-GH:
Closes #3257

Before:
grafik

grafik

grafik

After:
grafik

grafik

grafik

see screenshots
(the button "show only feeds with errors" appears only, when there are errors)

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


app/views/helpers/category/update.phtml
```diff
@@ -165,5 +158,25 @@
...
+				<button class="btn btn-attention confirm"
+					data-str-confirm="<?= _t('gen.js.confirm_action_feed_cat') ?>"
+					formaction="<?= _url('category', 'empty', 'id', $this->category->id()) ?>"
+					formmethod="post"><?= _t('gen.action.empty') ?></button>
+				<?php if (!$this->category->isDefault()): ?>
+				<button class="btn btn-attention confirm"
+					data-str-confirm="<?= _t('gen.js.confirm_action_feed_cat') ?>"
+					formaction="<?= _url('category', 'delete', 'id', $this->category->id()) ?>"
+					formmethod="post"><?= _t('gen.action.remove') ?></button>
+					<?php endif;?>
```
Frenzie:
```suggestion
				<?php endif;?>
```

Frenzie (approved):
lgtm, something about the alignment of that button at the top looks potentially a bit weird?

math-GH:
> 
> 
> lgtm, something about the alignment of that button at the top looks potentially a bit weird?

fixed

grafik

Frenzie:
That .box style is only used on that specific page?

math-GH:
> 
> 
> That .box style is only used on that specific page?

also in user queries pages (new layout still on the edge branch)
grafik

Frenzie:
> also in user queries pages (new layout still on the edge branch)

In any case, as long as you've made sure it looks good everywhere it's used. ^_^ :+1:

app/i18n/en/sub.php
```diff
@@ -58,7 +58,6 @@
 			'reload_articles_help' => 'Reload articles and fetch complete content if a selector is defined.',
 			'title' => 'Maintenance',
 		),
-		'moved_category_deleted' => 'When you delete a category, its feeds are automatically classified under <em>%s</em>.',
```
Alkarex: I believe you are still using this translation string, aren't you?

Alkarex (changes requested):

app/i18n/en/sub.php
math-GH: This text is still used. It moved from the "Subscription overview page" to the "category settings slider"

see:
https://user-images.githubusercontent.com/1645099/137562214-245ca7c7-911f-4ec3-9dd0-3e39d034903f.png

Alkarex: Indeed, my point is that it looks like this PR removes this text, which therefore creates an error.

math-GH: ok. you are right.
I do not know why I deleted it.
Fixed it.

Alkarex (approved):

PR #3895, opened 2021-10-16
EditorConfig update

Alkarex:
Small update, adding in particular `trim_trailing_whitespace`
No explicit rule yet for e.g. `*.md`, `Dockerfile*` for which some changes might be needed - left for later.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,18 +1,28 @@
-# Copying and distribution of this file, with or without modification,
-# are permitted in any medium without royalty provided this notice is
-# preserved.  This file is offered as-is, without any warranty.
-# Names of contributors must not be used to endorse or promote products
-# derived from this file without specific prior written permission.
-
-# EditorConfig
 # http://EditorConfig.org
 
+root = true
+
 [*]
+charset = utf-8
 end_of_line = lf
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,18 +1,28 @@
...
+trim_trailing_whitespace = true
 
-[*.php]
+[*.{html,php,phtml}]
+indent_size = 4
 indent_style = tab
+
+[*.js]
 indent_size = 4
-insert_final_newline = true
```
Frenzie, line 18: What's insert_final_newline exactly? Doesn't that sound like something we want?

Alkarex: It tells the editor to make sure that there is a new line character at the end of the document :-)
It avoids things like https://github.com/FreshRSS/FreshRSS/pull/3890/commits/fcf4120abe6f44274264a809c4968710c6950742#diff-db9e0b08abd6544fb688ad33f16634a2c55cf443b7631c41f7a81edadf735d6dL28

Alkarex: Ah, right, now I understand your question: this line was not removed, it is just the git diff that shows it this way. It is still in the `[*]` section

PR #3900, opened 2021-10-18
New shortcut: actualize feeds

math-GH:
- new shortcut that hits the "update" button to actualize the feeds

1. go to feed list (f.e. normal view)
2. click the shortcut (default: `q`)
3. the feeds will be updated

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] PHPCS run


Frenzie (approved):
If you like.

app/i18n/fr/conf.php
```diff
@@ -180,6 +180,7 @@
 	),
 	'shortcut' => array(
 		'_' => 'Raccourcis',
+		'actualize' => 'Actualiser flux',
```
```suggestion
		'actualize' => 'Actualiser les flux',
```

p/scripts/main.js
```diff
@@ -821,6 +821,13 @@ function init_shortcuts() {
 					return false;
 				}
 			}
+			if (k === s.actualize) {
+				var btn = document.getElementById('actualize');
```
```suggestion
				const btn = document.getElementById('actualize');
```

Alkarex:
For the translations, I suggest that we use the same string than `gen.action.actualize` (potentially updating its text to be more precise). Generally, I think we should aim at better reusing translation strings, not to end up with far too many.

math-GH:
> For the translations, I suggest that we use the same string than `gen.action.actualize` (potentially updating its text to be more precise). Generally, I think we should aim at better reusing translation strings, not to end up with far too many.

What is expected to finish this PR?

Alkarex:
@mathContao Instead of adding yet another translation string, my suggestion is to use the existing `gen.action.actualize`.

math-GH:
> 
> 
> @mathContao Instead of adding yet another translation string, my suggestion is to use the existing `gen.action.actualize`.

it would break the status quo of shortcut array in i18n:
https://github.com/FreshRSS/FreshRSS/blob/3623a5421df8198651d8e8a9d676f0661c307cf9/app/i18n/en/conf.php#L181-L190

Alkarex:
> it would break the status quo of shortcut array in i18n:

I think that is less important than continuing introducing (near) duplicates. The general idea I would like to favour being to put the strings where it makes most sense (kind of primary role), and reuse them as much as possible, to reduce duplicates to a minimum.

math-GH:
> 
> 
> > it would break the status quo of shortcut array in i18n:
> 
> I think that is less important than continuing introducing (near) duplicates. The general idea I would like to favour being to put the strings where it makes most sense (kind of primary role), and reuse them as much as possible, to reduce duplicates to a minimum.

It is here the right PR to change it?
Should I create a new structure that solves all shortcuts?

Frenzie:
Solves all shortcuts?

math-GH:
> 
> 
> Solves all shortcuts?

Solves the i18n for all shortcuts.

Alkarex:
> It is here the right PR to change it? Should I create a new structure that solves all shortcuts?

My suggestion is to only address `gen.action.actualize` in this PR, and potentially address the other ones in another PR (that would be welcome :-) )

math-GH:
Would it be fine to leave a comment in the shortcut array to understand, that this shortcut has its i18n on another place/another file?

Frenzie:
> Solves the i18n for all shortcuts.

I don't understand what you mean by solving shortcuts.

Alkarex:
> Would it be fine to leave a comment in the shortcut array to understand, that this shortcut has its i18n on another place/another file?

Unfortunately, our command line i18n manipulation tool would delete comments at the moment.

@Frenzie Some translation strings are more or less duplicates of strings we already have elsewhere - it is not specific to shortcuts.

math-GH:
I have no idea why the Korean language file fails

math-GH:
i18n changed as suggested. Ready to review

app/views/configure/shortcut.phtml
```diff
@@ -185,6 +185,14 @@
 		<fieldset>
 			<legend><?= _t('conf.shortcut.other_action') ?></legend>
 
+			<div class="form-group">
+				<label class="group-name" for="actualize"><?= _t('gen.action.actualize') ?></label>
+				<div class="group-controls">
+					<input type="text" id="actualize" name="shortcuts[actualize]" list="keys" value="<?= $s['actualize'] ?>"
+						data-leave-validation="<?= $s['actualize'] ?>"/>
```
Alkarex:
```suggestion
						data-leave-validation="<?= $s['actualize'] ?>" />
```

math-GH:
something strange happened.... 
I worked in `Main.js ` (see f.e. https://github.com/FreshRSS/FreshRSS/pull/3900/commits/407277b12b26ec80b86f2c42ca0687f432be737d) but at the end it is not changed in `main.js`

math-GH:
It was deleted via https://github.com/FreshRSS/FreshRSS/pull/3900/commits/009ec9281ca152138abd5c203c1c1da4729d1443#diff-86db6bc50f24e839f927bdd2262ce6d58c450fb23b13f8e9e5501b047add9bbaL853-L859 :( (@Alkarex )

Alkarex:
@mathContao The commit / diff you are referring to is not from our main branch.
See https://github.com/FreshRSS/FreshRSS/commits/edge/p/scripts/main.js
My last related change is https://github.com/FreshRSS/FreshRSS/pull/3906/files#diff-86db6bc50f24e839f927bdd2262ce6d58c450fb23b13f8e9e5501b047add9bbaL853
There has probably been a manual merge error somewhere

PR #3901, opened 2021-10-18
close panel with esc shortcut

### LINKED ISSUE #3450: [Feature] Use keyboard to exit a category/feed window in global view
While using global view, it sometimes hard to exit the category/feed since the background area is too narrow.
Would it be possible to use keyboard (e.g, press Esc key) to exit the window?
screenshot

math-GH:
Closes #3450

- `ESC` closes the panel
- little beauty: on larger screens there is more background to close the panel with the mouse

1. go to global view
2. click on category name
3. article list will be opened in a overflow
4. press `ESC` to close the overflow

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


p/scripts/global_view.js
```diff
@@ -73,6 +73,13 @@ function init_close_panel() {
 			document.getElementById('overlay').classList.remove('visible');
 			return false;
 		};
+	document.body.onkeydown = function (ev) {
```
Frenzie: `document.addEventListener('keydown', …`

p/scripts/global_view.js
```diff
@@ -73,6 +73,13 @@ function init_close_panel() {
...
+		let k = (ev.key.trim() || ev.code).toUpperCase();
+		if (k === 'ESC') {
+			k = 'ESCAPE';
+		}
+		document.querySelector('#overlay .close').click();
```
Frenzie: I assume this should be inside the `(k === 'ESC')`? Every keypress will close the popup.

Frenzie (changes requested):

p/scripts/global_view.js
math-GH: 🙈

p/scripts/global_view.js
Frenzie: See https://github.com/FreshRSS/FreshRSS/pull/3902/files for full example.

p/scripts/global_view.js
math-GH: fixed

p/scripts/global_view.js
math-GH: fixed

Frenzie: I think you may have forgotten to push?

math-GH: now the fix is available

p/scripts/global_view.js
```diff
@@ -73,6 +73,16 @@ function init_close_panel() {
...
+	document.addEventListener('keydown', ev => {
+		let k = (ev.key.trim() || ev.code).toUpperCase();
+		if (k === 'ESC') {
+			k = 'ESCAPE';
+		}
+		if (k == 'ESCAPE') {
+			document.querySelector('#overlay .close').click();
+		}
+		return false;
+	}
```
Frenzie:
```suggestion
	});
```
Also there's something weird about the indentation.

p/scripts/global_view.js
```diff
@@ -73,6 +73,16 @@ function init_close_panel() {
...
 			return false;
 		};
+	document.addEventListener('keydown', ev => {
+		let k = (ev.key.trim() || ev.code).toUpperCase();
+		if (k === 'ESC') {
+			k = 'ESCAPE';
+		}
+		if (k == 'ESCAPE') {
+			document.querySelector('#overlay .close').click();
+		}
```
Frenzie: Could you explain why you're doing this?

p/scripts/global_view.js
```diff
@@ -73,6 +73,16 @@ function init_close_panel() {
 			document.getElementById('overlay').classList.remove('visible');
 			return false;
 		};
+	document.addEventListener('keydown', ev => {
+		let k = (ev.key.trim() || ev.code).toUpperCase();
+		if (k === 'ESC') {
```
Frenzie: That is, as opposed to:
```suggestion
		if (k === 'ESC' || k === 'ESCAPE') {
```
(assuming that any of that stuff is necessary *at all* given that IE11 is deprecated and it's probably for IE8 or something)

p/scripts/global_view.js
math-GH: I reuse the click event of the dark overlay around the panel. When the user clicks with the mouse on it, than the panel will be closed.

Frenzie: Yes, but
```
	if (k === 'ESC') {
			k = 'ESCAPE';
		}
		if (k == 'ESCAPE') {
```
Gives me the heebie jeebies. ;-)

p/scripts/global_view.js
```diff
@@ -73,6 +73,13 @@ function init_close_panel() {
 			document.getElementById('overlay').classList.remove('visible');
 			return false;
 		};
+	document.addEventListener('keydown', ev => {
+		let k = (ev.key.trim() || ev.code).toUpperCase();
```
Alkarex:
```suggestion
		const k = (ev.key.trim() || ev.code).toUpperCase();
```

PR #3906, opened 2021-10-19
ESLint upgrade from JSHint 

Alkarex:
* commit corresponding `package.json`
* `npm run fix` for automatic JS and CSS fixes (this is how all the changes have been made in this PR)

Documentation:
* https://eslint.org
* Using most of the rules from https://github.com/standard/standardx , based on https://standardjs.com/rules.html


Frenzie:
Are you sure ESLint checks for all the same things? I'd be inclined to say they're complementary.

Alkarex:
@Frenzie I am not aware of anything in JSHint not available in ESLint. but let me know if you are worried about any specific point

.stylelintrc
```diff
@@ -33,11 +33,9 @@
     "no-eol-whitespace": true,
     "property-no-vendor-prefix": true,
     "rule-empty-line-before": [
-      "always",
-      "except": [
```
Alkarex, line 37: There was a typo in this JSON file, making it invalid (found when updating the NPM commands)

Frenzie:
I seem to recall ESLint not warning about missing break/return statements, variables referenced from a higher scope that could lead to confusing semantics, and various other things. But that was a few years ago.

Frenzie:
Of course JSHint comes with various annoying defaults, like its ridiculous "use dot notation" nonsense. But they've actually recognized since that they should just focus on code quality and get rid of aesthetic preferences (definitely ESLint for that).

Alkarex:
Regarding `break`, if I remove one:
image

If there is any rule that you believe is either lacking or annoying, let's have a look. From what I can see, ESLint is now a large superset of JSHint.

Frenzie:
Maybe just don't delete the config yet so we can easily run them side by side for a while?

Alkarex:
> Maybe just don't delete the config yet so we can easily run them side by side for a while?

@Frenzie Ok, I have restored JSHint for now and made sure it can run fine as well.

Frenzie:
Cool, thanks :+1:

Alkarex:
Any comment from @aledeg ?

aledeg:
Sorry for the lack of activity. I have so many personal project going on at the same time.
I do not have comments on that matter.

Alkarex:
No worries @aledeg :-)

PR #3909, opened 2021-10-21
Fix global context

Alkarex:
Regression from https://github.com/FreshRSS/FreshRSS/pull/3906
Was preventing login, among other consequences.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -35,7 +35,9 @@ function xmlHttpRequestJson(req) {
 // </Utils>
 
 // <Global context>
-let context;
+/* eslint-disable no-var */
+var context;
+/* eslint-enable no-var */
 
 (function parseJsonVars() {
 	const jsonVars = document.getElementById('jsonVars');


p/scripts/main.js
```diff
@@ -35,7 +35,9 @@ function xmlHttpRequestJson(req) {
 // </Utils>
 
 // <Global context>
-let context;
+/* eslint-disable no-var */
+var context;
```
Frenzie, line 39: It should be equally global, but not hoisted. ^_^

Alkarex: I also thought so, but I did not realise that `let` is less global than `var` for additional scripts (e.g., not available in `extra.js`).
We could use another syntax, such as `window.context`, but that was a quick fix. Suggestions welcome.

Frenzie: I'm using that kind of construction a fair bit with no issues. If I remember I'll take a look later, but at the same time I'm hardly bothered by a remaining occasional `var`. ^_^

PR #3917, opened 2021-10-23
Remove references to Ubuntu image in Docker/Readme.md

zoredache:
FreshRSS switched to Debian as the parent For the Image but there
were still a couple references in the README to Ubuntu.

- Minor adjustments to Docker/README.md

Pull request checklist:

- [*] clear commit messages
- [*] documentation updated

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -153,7 +153,7 @@ The tags correspond to FreshRSS branches and versions:
 ### Linux: Debian vs. Alpine
 
 Our default image is based on [Debian](https://www.debian.org/). We offer an alternative based on [Alpine](https://alpinelinux.org/) (with the `*-alpine` tag suffix).
-In [our tests](https://github.com/FreshRSS/FreshRSS/pull/2205), Ubuntu is faster,
+In [our tests](https://github.com/FreshRSS/FreshRSS/pull/2205), Alpine is slower,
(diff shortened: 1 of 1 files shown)


Alkarex:

Docker/README.md
```diff
@@ -153,7 +153,7 @@ The tags correspond to FreshRSS branches and versions:
 ### Linux: Debian vs. Alpine
 
 Our default image is based on [Debian](https://www.debian.org/). We offer an alternative based on [Alpine](https://alpinelinux.org/) (with the `*-alpine` tag suffix).
-In [our tests](https://github.com/FreshRSS/FreshRSS/pull/2205), Ubuntu is faster,
+In [our tests](https://github.com/FreshRSS/FreshRSS/pull/2205), Alpine is slower,
```
aledeg, line 156: I think there is a typo here. Don't you mean Debian?

Frenzie: I agree that the original phrasing "Debian is faster" (so pick that if you value speed) is a lot clearer than "Alpine is slower" (so pick what instead?).

Alkarex: Yes, that would be better indeed. By the way, we should re-run some performance tests to make sure this is still true. Furthermore, when the tests were done, if I remember correctly, we had no PHP opcache in Alpine because that package was not explicitly activated (fixed now).

PR #3918, opened 2021-10-23
Improved markdownlint

Alkarex:
* Relaxed rules slightly
* `npm run markdownlint` for automatic tests
* `npm run markdownlint_fix` for automatic syntax fixing
* Applied the fixes on all our Markdown files


.editorconfig
```diff
@@ -1,7 +1,5 @@
 # http://EditorConfig.org
 
-root = true
```
Alkarex, line 3: Removed to allow users to change their preference of `indent_size` for tabs at a higher level

Docker/README.md
```diff
@@ -226,7 +226,7 @@ Remember not pass the `CRON_MIN` environment variable to your Docker run, to avo
 
 Example on Debian / Ubuntu: Create `/etc/cron.d/FreshRSS` with:
 
-```
+```text
```
Frenzie, line 229: That strikes me as an annoyance compared to default = text, anything syntax highlighted is something special you should specify.

Alkarex: An advantage is that it allowed spotting several instances when the language was just forgotten

Alkarex: This is a rule by default, but it is possible to disable it if too annoying ;-)

Frenzie: > An advantage is that it allowed spotting several instances when the language was just forgotten

That's an interesting point, though I would imagine it's only forgotten when it really doesn't matter at all. (E.g., I spot a line that says `git status` while specifying it's sh.)

Anyway, my two cents are that it's too annoying while authoring to constantly have to specify `text`. ;-)

Alkarex: We have only 5 instances of `text`, and they are all for a line in a cron job, for which there is no syntax highlight, in the otherwise quite comprehensive list of options:
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml
So it is maybe not that often that we hit this case compared to cases when the language was just forgotten? (203 instances where the language should be specified in our codebase)
In this PR, examples where the language was missing include several instances of: sh, ini, apache, nginx.

Regarding console / shell / shell session, you make me realise that there are two different highlights available:
* `Shell` (alias `sh`, `bash`) with pure shell / Bash code
* `ShellSession` (alias `console`) when there is a mix of commands and outputs

```
No language # no styling
```

```sh
echo 'Language sh' # styling
````

```console
$ echo 'Language console' # styling
Language console
```

If we do want to disable this check, the rule is `"fenced-code-language": false,`

Frenzie: Fair enough I suppose. It's true that the need for preformatted plain text should be small.

PR #3924, opened 2021-10-23
Fix getNonStandardShortcuts

### LINKED ISSUE #3922: [BUG] Shortcut: some keys create a warning for preg_match()
**Describe the bug**
some keys for the shortcuts create a warning, f.e. `?` or `/` or `\`

grafik

**To Reproduce**
1. run your application in development mode (data/config.php => L3: `'environment' => 'development',`
2. Go to 'shortcuts config page'
2. edit the shortcuts to `?` or `/` or `\`
3. See warning

- FreshRSS version: 1.18.1

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3922
Hot fix. The `getNonStandardShortcuts()` could maybe be re-written to work with array elements.

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -538,10 +538,8 @@ function getNonStandardShortcuts($shortcuts) {
 	$standard = strtolower(implode(' ', SHORTCUT_KEYS));
 
 	$nonStandard = array_filter($shortcuts, function ($shortcut) use ($standard) {
-		if (false !== strpos($shortcut, ' ')) {
-			return true;
-		}
-		return !preg_match("/${shortcut}/i", $standard);
+		$shortcut = trim($shortcut);
+		return $shortcut !== '' & stripos($standard, $shortcut) === false;
 	});
 
 	return $nonStandard;


math-GH:
works fine

lib/lib_rss.php
```diff
@@ -538,10 +538,8 @@ function getNonStandardShortcuts($shortcuts) {
 	$standard = strtolower(implode(' ', SHORTCUT_KEYS));
 
 	$nonStandard = array_filter($shortcuts, function ($shortcut) use ($standard) {
-		if (false !== strpos($shortcut, ' ')) {
-			return true;
-		}
-		return !preg_match("/${shortcut}/i", $standard);
+		$shortcut = trim($shortcut);
+		return $shortcut !== '' & stripos($standard, $shortcut) === false;
```
Frenzie, line 542: Do we have both a strpos and a stripos or is one of those a typo?

math-GH: `stripos` is case-insensitiv (https://www.php.net/manual/en/function.stripos.php). `strpos` is case-sensitive

Alkarex: No error @Frenzie , but it would be nicer with a rewrite using array elements (but that is less urgent)

Alkarex:
Let's merge this hot fix and potentially come back to a better implementation if anyone is motivated :-)

PR #3927, opened 2021-10-24
Docker: don't add crontab unless needed.

zoredache:
Adding the crontab as part of the image build makes it more complicated
to add custom entries to the crontab. Adjusting the image and entrypoint
to make it so that the crontab is only added when CRON_MIN is set
simplifies having a custom crontab.

I have an example [compose file](https://github.com/zoredache/FreshRSS_Example/blob/main/docker-compose.yml) and [crontab](https://github.com/zoredache/FreshRSS_Example/blob/main/cron/freshrss) so you can see what I am trying to do here.

- Only add the crontab when the environment variable CRON_MIN is set.

I detailed how I tested this manually on [this page](https://github.com/zoredache/FreshRSS_Example/blob/main/Testing_my_PR.md). But basically, you build the image, and then run the container without the CRON_MIN set, and there is no crontab, and when CRON_MIN is set, notice that the crontab does exist, and the minute value has been updated as requested.

Pull request checklist:

- [*] clear commit messages
- [*] code manually tested
- [*] documentation updated

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -45,7 +45,7 @@ RUN sed -r -i "/^\s*(CustomLog|ErrorLog|Listen) /s/^/#/" /etc/apache2/apache2.co
 	touch /var/www/FreshRSS/Docker/env.txt && \
 	echo "7,37 * * * * . /var/www/FreshRSS/Docker/env.txt; \
 		su www-data -s /bin/sh -c 'php /var/www/FreshRSS/app/actualize_script.php' \
-		2>> /proc/1/fd/2 > /tmp/FreshRSS.log" | crontab -
+		2>> /proc/1/fd/2 > /tmp/FreshRSS.log" > /var/www/FreshRSS/Docker/crontab.default
 
 ENV COPY_LOG_TO_SYSLOG On
 ENV COPY_SYSLOG_TO_STDERR On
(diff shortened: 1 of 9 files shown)


Alkarex (approved):
Looks reasonable, thanks 👍🏻

Docker/Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN sed -r -i "/^\s*(CustomLog|ErrorLog|Listen) /s/^/#/" /etc/apache2/apache2.co
 	touch /var/www/FreshRSS/Docker/env.txt && \
 	echo "7,37 * * * * . /var/www/FreshRSS/Docker/env.txt; \
 		su www-data -s /bin/sh -c 'php /var/www/FreshRSS/app/actualize_script.php' \
-		2>> /proc/1/fd/2 > /tmp/FreshRSS.log" | crontab -
+		2>> /proc/1/fd/2 > /tmp/FreshRSS.log" > /var/www/FreshRSS/Docker/crontab.default
```
Alkarex, line 48: I did not realise that this breaks the use-case during development of having a volume for the FreshRSS source code...

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/3933

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/4117

PR #3920, opened 2021-10-23
GitHub Action automatic tests

Alkarex:
Replacement for Travis CI.
So far: PHP lint + phpunit + Manipulate translations + phpcs + eslint + markdownlint + rtlcss + stylelint

PR #3929, opened 2021-10-24
GitHub Actions CI optimisation

Alkarex:
Efforts to reduce the resources used by CI:
* Only one virtual environment and only one git checkout
* Exclusion of irrelevant directories for several commands
* Fix some rtlcss warnings
* Move some commands from `tests.yml` to `composer.json` to make them reusable
* Initial efforts to take advantage of all that from `make` (help welcome)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 *.min.js
 node_modules/
 p/scripts/vendor/
+vendor/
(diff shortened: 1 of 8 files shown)


Alkarex:
With only a single test, seems to shave about ~14 seconds.

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/3920

PR #3933, opened 2021-10-24
Fix cron regression

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/pull/3927/files#r735146297
The path `/var/www/FreshRSS/` might be a Docker volume, breaking files created there during Docker build


Alkarex:
Merging fast, as it is critical

Alkarex:
@zoredache Could you please test again after this patch?

zoredache:
Yes, that updated path tests out fine thanks.

Sorry I didn't realize that some people would have `/var/www/FreshRSS/` set as a volume and wouldn't be a good location for that file.

Issue #3903, opened 2021-10-18
[Feature] Change .onclick handlers to addEventlistener


Frenzie:
**Is your feature request related to a problem? Please describe.**

Using .onclick is generally considered undesirable. I seem to have overlooked them in #2234.

**Describe the solution you'd like**
Replace by addEventListener.

**Describe alternatives you've considered**
Leaving it as is, but see https://github.com/FreshRSS/FreshRSS/pull/3901 for why that's probably not the best idea. People will unsuspectingly copy it.

**Additional context**
There are others like .onerror.

PR #3943, opened 2021-10-27
Change .onclick handlers to addEventlistener 

math-GH:
Reference #3903

- Change .onclick handlers to addEventlistener 

Pull request checklist:

- [x] clear commit messages

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -135,7 +135,7 @@ function init_slider_observers() {
 		return;
 	}
 
-	document.querySelector('.post').onclick = function (ev) {
+	document.querySelector('.post').addEventListener('click', function (ev) {
 		const a = ev.target.closest('.open-slider');
 		if (a) {
 			if (!context.ajax_loading) {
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Not strictly close, but presumably the bulk of it. 

> There are others like .onerror.

Alkarex:
Some clicks do not seem to work. I will try to check.

Alkarex (changes requested):
Many events are not working, e.g. mark article as read/unread/favourite The two syntaxes are not exactly synonymous and do not have the same behaviour when it comes to e.g., propagation and bubbling of events. I cannot check in details now, but I believe additional lines are needed to cancel default events. I am not sure it is worth the extra complexity. Either mark as draft or reject this PR.

math-GH:
will wait for #3950, maybe that could be also a root of problems

math-GH:
The difference: https://www.geeksforgeeks.org/difference-between-addeventlistener-and-onclick-in-javascript/

Alkarex:
Some of the bugs might be fixed with the third parameter `useCapture: true` https://developer.mozilla.org/docs/Web/API/EventTarget/addEventListener

math-GH:
I will close this PR. It is too difficult to replace it and it works.

PR #3945, opened 2021-10-27
fixed: Shortcuts in "Anonym. Mode" broken (#3897)

### LINKED ISSUE #3897: [BUG] Shortcuts in "Anonym. Mode" broken
Anonym. mode (user is not logged in):
grafik

Use the shortcut "2" to view the global view. (the problem next works also with "1", "3" and "4").

A error message "A request has failed, it may have been caused by internet connection problems.".
grafik

Reason for the error message:
grafik

And than the page will reload automatically.

This error appears in the console:
grafik

grafik

It needs this button, that is not available in the Anonym. Mode
grafik

 - FreshRSS version: 1.19.0-dev (edge branch), but I guess in 1.18.1 it should be the same issue

math-GH:
Closes #3897

- do not use delayedClick(). Use the same onclick event as clicking on the view buttons

1. click `1`, or `2` or `3` or `4` (default shortcuts)
2. change to the view without an error

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -201,6 +201,12 @@ const pending_entries = {};
 let mark_read_queue = [];
 
 function send_mark_read_queue(queue, asRead, callback) {
+	if (!queue || queue.length === 0) {
+		if (callback) {
+			callback();
+		}
+		return;
+	}
 	const req = new XMLHttpRequest();
 	req.open('POST', '.?c=entry&a=read' + (asRead ? '' : '&is_read=0'), true);
 	req.responseType = 'json';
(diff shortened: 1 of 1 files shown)


Alkarex:
The role of `delayedClick` is to queue different events, in particular when navigating quickly with shortcuts. It is good to wait for e.g. a mark-as-favourite request before changing view. Removing it will probably break those scenarios. There is probably another way to fix the issue in the anonymous mode. I cannot look at it immediately

Frenzie:
We have `context.anonymous` `true` or `false`.

Screenshot_2021-10-28_09-05-52

math-GH:
I implemented now the context.anonymous into delayedClick()

Alkarex:
@mathContao Please check this alternative approach, which I guess is fixing the root cause https://github.com/FreshRSS/FreshRSS/pull/3945/commits/d064bb34f9110d838d2921a57a76e50986bcc782

Alkarex (approved):

math-GH:
> 
> 
> @mathContao Please check this alternative approach, which I guess is fixing the root cause [d064bb3](https://github.com/FreshRSS/FreshRSS/commit/d064bb34f9110d838d2921a57a76e50986bcc782)

It works fine too

PR #3953, opened 2021-11-01
Improve subscription list drag and drop

### LINKED ISSUE #3892: [BUG] Subscription Management: reordering feeds
Subscription management:

Drag&Drop:
between categories (green marked): no problem
inside of a category (red): 404 error (cannot reorder)

grafik

**To Reproduce**
1. Go to subscription management
2. reorder a feed via drag and drop

**Expected behavior**
can reorder the feed inside of a category

 - FreshRSS version: [e.g. 1.18.1]

math-GH:
Closes #3892

After:
grafik

1. Dragging object marked yellow (cog icon will be hidden)
2. (dragging object)
3. drop zones are marked with yellow stripped background
4. drag-hover marked with yellow plain background
5. dropped items are listed at the top of the list (because they needed to ordered alphabetical. Will be done after reload the page)

grafik
The drag&drop function is now better visible with a changed mouse cursor (now: `move`, before: `grab`)

This PR closes #3892 because you cannot drop a feed within the category where it comes from. So it will not create a 404 error. Technical solution: no dropzone for a `<ul>` where the object comes from.

1. go to subscription manage page
2. drag and drop the feeds around

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested
- [x] phpcs
- [ ] CSS RTL


p/scripts/category.js
```diff
@@ -47,41 +47,53 @@ function init_draggable() {
 	const dropSection = document.querySelector('.drop-section');
 
 	dropSection.ondragstart = function (ev) {
-		const li = ev.target.closest ? ev.target.closest(draggable) : null;
-		if (li) {
+		const li_draggable = ev.target.closest ? ev.target.closest(draggable) : null;
+		if (li_draggable) {
+			const ulClosest = li_draggable.closest('ul');
+			ulClosest.classList.add('disallowDragging');
```
Frenzie: I'd say something more like `drag-disallowed` or simply `no-drag`, for consistency with `drag-active` and `drag-hover`.

math-GH: thanks. Fixed it

p/scripts/category.js
```diff
@@ -90,31 +102,35 @@ function init_draggable() {
...
 			req.onload = function (e) {
 				if (this.status == 200) {
-					li.insertAdjacentHTML('afterend', dragHtml);
-					if (li.classList.contains('disabled')) {
-						li.remove();
+					ul_dropzone.insertAdjacentHTML('afterbegin', dragHtml);
+					ul_dropzone.firstChild.classList.add('moved');
+					ul_dropzone.scrollTop = 0;
+					const disabledElement = ul_dropzone.getElementsByClassName('disabled');
+					if (ul_dropzone.childElementCount <= 3 && disabledElement.length > 0) {
```
Frenzie: What does this mystery number mean exactly? Please either define it above (`const MIN_SOMETHING = 3` for example) or add a comment.

math-GH: it is difficult to explain: the `<ul>` consists of minimum: (1) Add a feed link (2) the message box "Empty category" and (3) the dropped element.
It is used to delete the message box, when the element is dropped to the list.
Any suggestion for a good comment?

BTW, it is not new, just changed the number. See before: 
https://github.com/FreshRSS/FreshRSS/blob/0d31f89dd8519325ef8ed6efbe77eed4c452f62c/p/scripts/category.js#L26-L30

Frenzie: It's not new, I added it myself a couple of weeks ago and you overlooked it, lol :-P https://github.com/FreshRSS/FreshRSS/pull/3893

The original does suffer from a lack of explanation as to why the line's there, but 0 is not a mystery number. It just says "if empty". :-) https://github.com/FreshRSS/FreshRSS/commit/ce43b1761b810c89e5e8660e8ae2638e97ca6339

Anyway, now that you've explained what it's for, what about something more along the lines of `if (!p.querySelector('[draggable=true]'))`? That way, much like the 0 thing, you just say if there are no elements that are draggable, without having a mystery number that would need to be updated every time you change something.

math-GH: I have a much more better idea ;)

Frenzie (approved):

math-GH:
Ready for 1.20 :)

Alkarex:
Minor detail: it looks like feeds remain in italics after a drag & drop. By design?
image

math-GH:
> Minor detail: it looks like feeds remain in italics after a drag & drop. By design? image

yes, it is by design. After dropping the new item is placed at the top and in italics style. 
After reload it will be ordered alphabetically

PR #3956, opened 2021-11-02
Fix #3900 / #3950: main.js incl. update feed shortcut

### LINKED ISSUE #3950: [BUG] main.js broken via PR #3900
**Describe the bug**
main.js is broken via #3900 https://github.com/FreshRSS/FreshRSS/pull/3900#issuecomment-955446934

I found it because the "feed update" shortcut does not work that should be implemented with #3900

math-GH:
Closes #3950 / fixed #3900

- implemented the logic to run the update button made in #3900

1. press the shortcut to update the feeds
2. feeds will be updated

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -857,6 +857,13 @@ function init_shortcuts() {
 				return false;
 			}
 		}
+		if (k === s.actualize) {
+			const btn = document.getElementById('actualize');
+			if (btn) {
+				btn.click();
+			}
+			return false;
+		}
 		if (k === s.next_entry) {
 			if (ev.altKey) {
 				next_category();


p/scripts/main.js
```diff
@@ -857,6 +857,13 @@ function init_shortcuts() {
 				return false;
 			}
 		}
+		if (k === s.actualize) {
+			const btn = document.getElementById('actualize');
+			if (btn) {
+				btn.click();
+			}
```
Frenzie, line 864: Why not just call the action directly?

Frenzie: I see. I suppose that's fine then, but it might mean it's time to refactor it out. I'll defer to @Alkarex.
https://github.com/FreshRSS/FreshRSS/blob/9326e419e3b90b2db4b6a8531503b710605ac6d4/p/scripts/main.js#L1234-L1286

Alkarex: I might have missed a comment. What is your suggestion, @Frenzie ?

Frenzie: Conceptually I'd expect these to be separate actions that trigger the same function, rather than that other actions trigger a click event on the button that triggers the function.

PR #3962, opened 2021-11-06
Improve install routine: database password input with "show password" button

math-GH:
before:
grafik

after:
grafik

1. install routine step 3: database configuration
2. database password: click on button shows the plain password

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] CSS: RTL

### DIFF
--- a/app/install.php
+++ b/app/install.php
@@ -588,7 +588,7 @@ function printStep3() {
 				<div class="stick">
 					<input type="password" id="passwordPlain" name="passwordPlain" pattern=".{7,}"
 						autocomplete="off" <?= $auth_type === 'form' ? ' required="required"' : '' ?> tabindex="5" />
-					<a class="btn toggle-password" data-toggle="passwordPlain"><?= FreshRSS_Themes::icon('key') ?></a>
+					<button type="button" class="btn toggle-password" data-toggle="passwordPlain"><?= FreshRSS_Themes::icon('key') ?></button>
 				</div>
 				<p class="help"><?= _i('help') ?> <?= _t('install.auth.password_format') ?></p>
 				<noscript><b><?= _t('gen.js.should_be_activated') ?></b></noscript>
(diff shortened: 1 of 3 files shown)


app/install.php
```diff
@@ -503,34 +503,37 @@ function printStep2() {
 		<div class="form-group">
 			<label class="group-name" for="pass"><?= _t('install.bdd.password') ?></label>
 			<div class="group-controls">
-				<input type="password" id="pass" name="pass" value="<?=
-					isset($_SESSION['bd_password']) ? $_SESSION['bd_password'] : '' ?>" tabindex="4" autocomplete="off" />
+				<div class="stick">
+					<input type="password" id="pass" name="pass" value="<?=
+						isset($_SESSION['bd_password']) ? $_SESSION['bd_password'] : '' ?>" tabindex="4" autocomplete="off" />
+					<a class="btn toggle-password" data-toggle="pass" tabindex="5"><?= FreshRSS_Themes::icon('key') ?></a>
```
Frenzie, line 509: A `<button>` is normally preferable for, well, buttons.

math-GH: yes, you are right. But: to use `<a>` here is default.
I will change it here in this PR for this button, but it will be a separate task for all other show-password-buttons

p/themes/Origine/origine.css
```diff
@@ -47,10 +47,11 @@ option {
 	padding: 0 .5em;
 }
 
-input:focus, select:focus, textarea:focus {
+input:focus, select:focus, textarea:focus, input[type="password"]:focus + a.toggle-password {
```
Alkarex: This style might have to be updated for `<button>` ?

math-GH: thanks. Fixed it.

app/install.php
math-GH: see PR: #3969

PR #3984, opened 2021-11-21
Ignore Invalid utf-8 characters

a180285:
Closes #

- Some feed's xml has Invalid utf-8 characters. If ignore the invalid chars the feeds is still workable.
-
-

1. E.g. https://coolshell.cn/feed
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


lib/SimplePie/SimplePie.php
```diff
@@ -1559,6 +1596,7 @@ public function init()
 			{
 				// Create new parser
 				$parser = $this->registry->create('Parser');
+				$utf8_data = self::stripInvalidUtf8($utf8_data);
```
Frenzie: Perhaps that should be something more along these lines? Pinging @Alkarex 
```suggestion
				if ($this->force_feed) {
					$utf8_data = self::stripInvalidUtf8($utf8_data);
				}
```

lib/SimplePie/SimplePie.php
```diff
@@ -1382,6 +1382,43 @@ function cleanMd5($rss)
...
+				($current == 0xD) ||
+				(($current >= 0x20) && ($current <= 0xD7FF)) ||
+				(($current >= 0xE000) && ($current <= 0xFFFD)) ||
+				(($current >= 0x10000) && ($current <= 0x10FFFF)))
+			{
+				$ret .= chr($current);
+			}
+			else
+			{
+				$ret .= " ";
```
Frenzie: Or `REPLACEMENT CHARACTER`? I realize it would look uglier, but it's not necessarily clear that hiding the crime would be better.

https://www.fileformat.info/info/unicode/char/fffd/index.htm

Frenzie (commented):
Basically lgtm, left some stray thoughts.

Alkarex:
Could you please try this piece of code instead?

https://github.com/FreshRSS/FreshRSS/blob/392861b7c3f9c6ff4588ad755a62514c6951e86a/lib/lib_rss.php#L100-L106

a180285:
> Could you please try this piece of code instead?
> 
> https://github.com/FreshRSS/FreshRSS/blob/392861b7c3f9c6ff4588ad755a62514c6951e86a/lib/lib_rss.php#L100-L106

a180285:
> Could you please try this piece of code instead?
> 
> https://github.com/FreshRSS/FreshRSS/blob/392861b7c3f9c6ff4588ad755a62514c6951e86a/lib/lib_rss.php#L100-L106

Sorry for the closed.

After testing with both functions: "mb_convert_encoding" and "iconv('UTF-8//IGNORE').
They both are not pass the test. As there are a char "0x08" in the feed. after the above 2 functions. the char '0x08' still there.


I didn't know why as in the doc. iconv("UTF-8//IGNORE") should remove the char "0x08". But it doesn't work.

a180285:
I found a link: https://stackoverflow.com/questions/5742543/an-invalid-xml-character-unicode-0xc-was-found
and: https://stackoverflow.com/questions/730133/what-are-invalid-characters-in-xml

Maybe it's because 0x08 is valid utf-8 but it's not valid for xml? I'm not very sure.

Frenzie:
Maybe iconv has another flag to remove control characters?

a180285:
> Maybe iconv has another flag to remove control characters?

I only find two flag for iconv  "//TRANSLIT" and "//IGNORE" here: https://www.php.net/manual/en/function.iconv.php

And after read more detail on xml 1.0 here: https://www.w3.org/TR/xml/#charsets

<html>
<body>
<h5 style="text-align: left; font: italic 16px sans-serif; color: rgb(0, 0, 0); letter-spacing: normal; orphans: 2; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a name="char32" id="char32">Character Range</a></h5>

[2] | Char | ::= | #x9 \| #xA \| #xD \| [#x20-#xD7FF] \| [#xE000-#xFFFD] \| [#x10000-#x10FFFF]
-- | -- | -- | --



</body>
</html>


It's exactly what the code remove for. So I changed the function name to stripInvalidXmlCharacter instead of stripInvalidUtf8Character

math-GH:
@a180285 Any progress here?

Alkarex:
> Any progress here?

I am not (yet) super keen on it, as it is using a quite slow/expensive loop, will break XML 1.1 and maybe even some editions of XML 1.0 (I need to double-check). I am also unsure whether that works for e.g. UTF-16 or UTF-32. SimplePie already has some error corrections in place, and we need to see what the minimal changes are for the original example to pass without risking breaking any other legitimate use case. We could also have a feed-specific option to perform some fixes / conversions.

a180285:
From my side, I have use my build version for a while. Basically, it works fine. But yes, I'm not fully test other encoding E.g. UTF-16.

Alkarex:
If I am not mistaken, the problematic feed https://coolshell.cn/feed has been fixed, so this PR does not seem necessary anymore, for this feed at least.
Furthermore, our SimplePie fork is now maintained on https://github.com/FreshRSS/simplepie so new PRs should be sent there, or even better upstream https://github.com/simplepie/simplepie (but the upstream repo is very slow).

PR #3997, opened 2021-11-26
Optimize for mobile view: line break in article title

math-GH:
Before:
grafik

After:
grafik

- line breaks instead of truncated title

1. see normal view page

(I tested it with all templates)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
I believe we have had a few discussions about that.
Related to https://github.com/FreshRSS/Extensions/tree/master/xExtension-TitleWrap
Might benefit an option, maybe with a max number of lines in case of very long titles on small windows.
To be experimented combined with the different *thumbnail* view options.

PR #4008, opened 2021-11-30
Improved German translation of "related tags"

math-GH:
grafik

- better translation in German for the article tags. "Verwandte Tags" is not a good translation. The word "Tag"/"Tags" is quite difficult because it looks like "Tag"(=Day). I would suggest to use the word "Hashtag" that makes it more clear

1. see at the end of article.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/i18n/de/conf.php
+++ b/app/i18n/de/conf.php
@@ -27,7 +27,7 @@ return array(
 			'display_authors' => 'Autoren',
 			'entry' => 'Artikel-Symbole',
 			'publication_date' => 'Datum der Veröffentlichung',
-			'related_tags' => 'Verwandte Tags',
+			'related_tags' => 'Hashtags',
 			'sharing' => 'Teilen',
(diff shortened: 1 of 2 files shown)


Frenzie:
What about something like label/zettel/etikett? Hashtag seems weird to me, unless that's how you use it in German these days.

math-GH:
> What about something like label/zettel/etikett? Hashtag seems weird to me, unless that's how you use it in German these days.

`label` is already used by the label feature (see left of the "hashtag").
`Zettel` and `Etikett` are not used. It is the correct translation in the non-online-world, but not used for digital. In digital world we use the English words "Tag" and "Hashtag" similar.

Frenzie:
Right, but a tag isn't a hashtag.

Frenzie:
But if you think it works well enough in German. :)

Frenzie (approved):

PR #4013, opened 2021-12-01
Add preventDefault to keydown event handler

### LINKED ISSUE #3946: Using Space as Keyboard shortcut for "Jump to next Article" suddenly triggers additional scrolling
One of the latest commits seemed to have broken the keyboard shortcut "jump to next article" when using "Space" as the shortcut with Google Chrome.

On the latest stable version 1.18.1, jumping to the next article with "space" worked perfectly. After updating to "1.19.0-dev", FreshRSS jumps to the next article AND Google Chrome does some additional scrolling which is the default behaviour on any website when pressing the space key.

See discussion here: https://github.com/FreshRSS/FreshRSS/issues/3891#issuecomment-953184909_

Frenzie:
Fixes #3946.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -980,7 +980,7 @@ function init_shortcuts() {
 		if (k === s.reading_view) { delayedClick(document.querySelector('#nav_menu_views .view-reader')); ev.preventDefault(); return; }
 		if (k === s.global_view) { delayedClick(document.querySelector('#nav_menu_views .view-global')); ev.preventDefault(); return; }
 		if (k === s.rss_view) { delayedClick(document.querySelector('#nav_menu_views .view-rss')); ev.preventDefault(); return; }
-		if (k === s.toggle_media) { toggle_media(); ev.preventDefault(); return; }
+		if (k === s.toggle_media) { toggle_media(); ev.preventDefault(); }
 	});
 }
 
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -843,6 +843,7 @@ function init_shortcuts() {
 				ev.ctrlKey || ev.metaKey || (ev.altKey && ev.shiftKey)) {
 			return true;
 		}
+		ev.preventDefault();
```
Alkarex: But that would be applied for any key event, right, even shortcuts we do not catch? Probably unwanted side-effects?

p/scripts/main.js
```diff
@@ -937,15 +943,17 @@ function init_shortcuts() {
...
 				}
 			}
-			return false;
+			ev.preventDefault();
+			return;
 		}
 
 		if (ev.altKey || ev.shiftKey) {
-			return true;
+			return;
```
Alkarex, line 951: Using this approach, no difference when returning `true` of `false`?

Frenzie: I can't find it on MDN at the moment, but `return false` in an old style event listener is equivalent to preventDefault(), stopPropagation() and possibly some other things like that as well.

The return value from addEventListener() is ignored.

Issue #4016, opened 2021-12-02
Add PHPstan in project and CI


ColonelMoutarde:
PHPstan is a very powerful tool that does static code analysis. It is highly recommended for this project. 

Attention, it is compatible only from php7. It is therefore an addition that must be made in the roadmap.

Alkarex:
We are already at PHP7+, having dropped PHP5 recently :-)
Additional information and PRs welcome!

math-GH:
@ColonelMoutarde : do you know how to set-up it? I have a look on PHPstan but did not tried it yet

Frenzie:
```
docker run --rm -v $(pwd):/app ghcr.io/phpstan/phpstan analyse /app
```
[phpstan-output.txt](https://github.com/FreshRSS/FreshRSS/files/7644427/phpstan-output.txt)

Alkarex:
We are now passing PHPStan level 0 and 1 https://github.com/FreshRSS/FreshRSS/pull/4021 and it is added to our continuous integration.
Help welcome to pass PHPStan level 2 https://github.com/FreshRSS/FreshRSS/issues/4103

Alkarex:
See https://github.com/FreshRSS/FreshRSS/pull/4110 for PHPStan level 3

Alkarex:
Now up to level 5 https://github.com/FreshRSS/FreshRSS/pull/4110

PR #4019, opened 2021-12-02
Fix some PHPstan errors

Alkarex:
Fix some wrong variables found by https://github.com/FreshRSS/FreshRSS/issues/4016

### DIFF
--- a/app/Models/FeedDAO.php
+++ b/app/Models/FeedDAO.php
@@ -460,7 +460,7 @@ SQL;
 				 . 'SET `cache_nbUnreads`=`cache_nbUnreads`-' . $affected
 				 . ' WHERE id=:id';
 			$stm = $this->pdo->prepare($sql);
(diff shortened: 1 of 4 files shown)


app/Models/FeedDAO.php
```diff
@@ -460,7 +460,7 @@ public function keepMaxUnread($id, $n) {
 				 . 'SET `cache_nbUnreads`=`cache_nbUnreads`-' . $affected
 				 . ' WHERE id=:id';
 			$stm = $this->pdo->prepare($sql);
-			$stm->bindParam(':id', $id_feed, PDO::PARAM_INT);
+			$stm->bindParam(':id', $id, PDO::PARAM_INT);
```
Alkarex, line 463: This is the most significant error: the cache about the number of unread articles after the automatic "keep max unread articles" function was wrong until the next correct update.

PR #4020, opened 2021-12-02
Minz request avoid custom methods

Alkarex:
Take advantage of PHP7+ null-coalescing operator `??` to make code more standard, shorter, and faster instead of custom functions with no extra functionality.

Allows code to be better tested and fix two PHPstan errors:

```
 ------ -----------------------------------------
  Line   app/Controllers/configureController.php
 ------ -----------------------------------------
  410    Cannot unset offset 'rid' on string.
 ------ -----------------------------------------

 ------ ------------------------------------
  Line   lib/Minz/FrontController.php
 ------ ------------------------------------
  70     Cannot unset offset 'c' on string.
  71     Cannot unset offset 'a' on string.
 ------ ------------------------------------
```

https://github.com/FreshRSS/FreshRSS/issues/4016

### DIFF
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -152,7 +152,7 @@ class FreshRSS_configure_Controller extends Minz_ActionController {
 		Minz_View::appendScript(Minz_Url::display('/scripts/draggable.js?' . @filemtime(PUBLIC_PATH . '/scripts/draggable.js')));
 
 		if (Minz_Request::isPost()) {
-			$params = Minz_Request::fetchPOST();
+			$params = $_POST;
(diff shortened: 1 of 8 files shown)


Frenzie (approved):
So the rationale was just the default value option or something else? I feel like I'm missing the point of the current situation since I would've probably just written what's proposed in this PR.

Alkarex:
> So the rationale was just the default value option or something else? I feel like I'm missing the point of the current situation since I would've probably just written what's proposed in this PR.

Yes. I have never been found of that code, but now that `??` is available, it makes the simplification even more obvious.

Issue #3686, opened 2021-07-05
[BUG] A feed could not be found at


ffuentese:
FreshRSS doesn't seem to detect the content type of certain feeds

**To Reproduce**
1. Go to the Add Subscription dialog
2. Add this feed https://daily.bandcamp.com/feed/ or https://daily.bandcamp.com/feed/rss.xml
3. The feed will display an error and won't get added

> A feed could not be found at `https://daily.bandcamp.com/feed/`; the status code is `200` and content-type is `` [https://daily.bandcamp.com/feed/]

**Expected behavior**

Add the blog's feed to my account and retrieve its entries.

**Screenshots**

**Environment information (please complete the following information):**

Versión
    1.19.0-dev
Performed from the web interface. 
FreshRSS is installed on a Raspberry Pi with Debian Buster.

Frenzie:
The bug is with the feeds. They're not served correctly, as indicated by the error message.

```
HTTP/1.1 200 OK
Connection: keep-alive
Server: nginx
Set-Cookie: client_id=B92CE32F9E8CB8612A565EEA9F0DA31723F2050F0C523910EECC061863AE0AD3; domain=.bandcamp.com; path=/; expires=Sat, 05 Jul 2031 19:44:30 -0000; secure; HttpOnly; SameSite=None
Referrer-Policy: no-referrer-when-downgrade
Content-Security-Policy: base-uri 'none'; object-src 'none'; report-uri https://bandcamp.com/api/cspreport/1/violation; script-src http: https: 'nonce-eoIbgbG1RdYa/RTACusqNQ==' 'report-sample' 'strict-dynamic' 'unsafe-eval' 'unsafe-inline'
Set-Cookie: BACKENDID=red-78b9-10; path=/; Secure
Accept-Ranges: bytes
Date: Mon, 05 Jul 2021 19:44:30 GMT
Via: 1.1 varnish
X-Served-By: cache-ams21063-AMS
X-Cache: MISS
X-Cache-Hits: 0
X-Timer: S1625514270.071733,VS0,VE350
transfer-encoding: chunked
```

By comparison, how it should be (`content-type: application/rss+xml`):
```
[…]
content-type: application/rss+xml; charset=UTF-8
```

But as the error message also indicates, you can force a bad feed by adding `#force_feed` to the URL.

ffuentese:
>But as the error message also indicates,

I'm not getting that message

Frenzie:
[...]

PR #4023, opened 2021-12-04
Improve showing notification banner

math-GH:
# Part 1:

While I reproduced #3686 I saw that the time to read the notification is to short (4 seconds).

- if it is a 'bad' (=red) notification, than the time is doubled to disappear (= 8 seconds)
- if the user moves the mouse pointer over the notification, the timer stopps
- if the user's mouse pointer leaves the notification it will disappear after 1 second
- changed the onlick-event of the close icon to addEventListener

grafik

1. go to subscription management: page 'add a feed or category'
2. press the 'add' button of 'add a category' (leave the input empty!)
3. a read notification will appear
4. test the changes

# Part 2
Before:
The text inside of the notification could be hidden behind the close button and the grey button could be invisible over the text. And the line-height is strange.
grafik

After:
No text anymore behind the close button.
grafik

While hovering the grey icon will become white
grafik

I checked all themes and improved it in some little cases (f.e. readable links)

# Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1353,6 +1353,10 @@ function init_notifications() {
 		clearInterval(notification_interval);
 	})
 
+	notification.addEventListener('mouseleave',function(){
+		notification_interval = setTimeout(closeNotification, 1000);
+	})
+
 	if (notification.querySelector('.msg').innerHTML.length > 0) {
 		notification_working = true;
 		if (notification.classList.contains('good')) {
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1349,6 +1349,10 @@ function init_notifications() {
 		return false;
 	};
 
+	notification.addEventListener('mouseenter',function(){
+		clearInterval(notification_interval);
+	})
+
 	if (notification.querySelector('.msg').innerHTML.length > 0) {
 		notification_working = true;
 		if (notification.classList.contains('good')) {
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1327,7 +1327,7 @@ function openNotification(msg, status) {
 	notification.querySelector('.msg').innerHTML = msg;
 	notification.className = 'notification';
 	notification.classList.add(status);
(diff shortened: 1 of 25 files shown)


p/scripts/main.js
```diff
@@ -1340,14 +1344,28 @@ function closeNotification() {
...
-	};
+	});
+
+	notification.addEventListener('mouseenter', function () {
+		clearInterval(notification_interval);
+	});
+
+	notification.addEventListener('mouseleave', function () {
+		notification_interval = setTimeout(closeNotification, 1000);
+	});
```
Frenzie: That seems like it'd be very easy to trigger accidentally and unknowingly. There's already the x if you want to get rid of it sooner.

math-GH: I know. 
Let me explain:
default timer to close the notification 4s or 8s.
When the user moves the mouse pointer over the notification: the timer will be stopped. So the user can read the message how long he/she needs. (Here the user can close the message box via the close icon.)
When the user moves the cursor out of the notification box, than it sets a new timer to 1 second.

Frenzie: I realize that's the reasoning, but I don't think that addresses my remark. Your cursor can be in the general vicinity for any reason, and if you have arthritis your hand just shivers while operating the mouse even when you are completely aware of this behavior.

The long and short of it is that you're then ensuring the notification will disappear within one second after the cursor leaves the target. It should either reset to the full time (maybe minus one or two seconds, but not *to* one second) or track the actual starting time and resume.

math-GH: ah ok. Now I understood.
Will change it to 3 seconds.

math-GH:
shit. this changes should become another PR and not within this PR :/
OK, it can stay here. It is also connected with the notification banners. Will extend the PR text with this changes.

p/scripts/main.js
math-GH: done

PR #4024, opened 2021-12-04
3687 fix error message force feed

### LINKED ISSUE #3687: Update translation(s) for feed error message
I'd look into it right now, but it's almost midnight.

<hr>

I just get a pink error message with a link to go back to my log. Might that be a translation error? 
imagen

_Originally posted by @ffuentese in https://github.com/FreshRSS/FreshRSS/issues/3686#issuecomment-874336822_

math-GH:
Closes #3687

- added `You can try force adding by appending <code>#force_feed</code> to the URL.` to all languages

see #3687

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/i18n/cz/feedback.php
```diff
@@ -84,7 +84,7 @@
 			'cache_cleared' => '<em>%s</em> cache has been cleared',	// TODO - Translation
 			'deleted' => 'Kanál byl smazán',
 			'error' => 'Kanál nelze aktualizovat',
-			'internal_problem' => 'RSS kanál nelze přidat. Pro detaily <a href="%s">zkontrolujte logy FreshRSS</a>.',
+			'internal_problem' => 'RSS kanál nelze přidat. Pro detaily <a href="%s">zkontrolujte logy FreshRSS</a>. You can try force adding by appending <code>#force_feed</code> to the URL.',	// TODO - Translation
```
Frenzie: I don't think this will quite work with the automated tools unfortunately, but the CI will say soon enough.

math-GH: fixed it

Frenzie:
@aledeg Do you have a suggestion as to what would be the best thing to do here?

aledeg:
I am willing to help but I am not sure to understand what you are expecting me to do!

Frenzie:
You know the most about the translation tools. I guess we're missing a dirty like in gettext. So the best thing to do might be to reset the strings completely?

aledeg:
I still don't get it. Do you mean something to notice translator that the string has changed and needs to be translated again?
If so, adding the ''// TODO - Translation'' string at the end of it would do the trick I think. Not sure though.

Frenzie:
> If so, adding the ''// TODO - Translation'' string at the end of it would do the trick I think. Not sure though.

It would, but the way it's set up then the tools will automatically remove the todo because the strings are different.

aledeg:
I see what you mean. Either we keep it as-is for now and we are trying to be a bit more clever with the translation tool or we remove completely the translation.
What could also work is splitting the string in 2. This way, the translated one does not change and the other one needs to be translated.

Frenzie:
That's an interesting alternative solution. I think just removing it is probably no big deal seeing how it's just one string; just a pity we can't easily do dirty atm. ^-^

math-GH:
what about to set this phrase on the ignore list?

Frenzie:
Only deleting it or adding dirty to the tools will have the desired effect.

math-GH:
Please decide what to do here to finish this PR.

Frenzie:
Please delete/reset it in the affected languages.

math-GH:
what should I delete?

Frenzie:
I mean reset it to `'internal_problem' => 'Source English string',	// TODO - Translation`.

math-GH:
What about to set this phrases on the ignore list? Would it work?

math-GH:
Anything left to do?

app/i18n/cz/feedback.php
```diff
@@ -84,7 +84,7 @@
 			'cache_cleared' => '<em>%s</em> cache has been cleared',	// TODO - Translation
 			'deleted' => 'Kanál byl smazán',
 			'error' => 'Kanál nelze aktualizovat',
-			'internal_problem' => 'RSS kanál nelze přidat. Pro detaily <a href="%s">zkontrolujte logy FreshRSS</a>.',
+			'internal_problem' => 'The newsfeed could not be added. <a href=%s>Check FreshRSS logs</a> for details. You can try force adding by appending <code>#force_feed</code> to the URL.',
```
Frenzie, line 87:
```suggestion
			'internal_problem' => 'The newsfeed could not be added. <a href="%s">Check FreshRSS logs</a> for details. You can try force adding by appending <code>#force_feed</code> to the URL.',	// TODO - Translation
```

Frenzie:
I'd noticed the lack of todo but I overlooked that the CI didn't mention it due to a subtly changed string.

math-GH:
> I'd noticed the lack of todo but I overlooked that the CI didn't mention it due to a subtly changed string.

How could I add the TODO via CLI? I thought it will be done automatically via the CLI

Frenzie:
It will do it automatically if you use the correct string. ;-)

math-GH:
better PR: #4145

PR #4025, opened 2021-12-04
Fix: search in mobile view

math-GH:
grafik

- bring back the search input in mobile view. Some themes had not a small search input in the nav menu (Other themes have a big search input)

1. in mobile view

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -1186,7 +1186,7 @@ a.btn {
 		padding: 3px 5px;
 		max-width: 97%;
 		width: 90px;
-		line-height: 25px;
+		line-height: 2;
 	}
 
 	.nav_menu .search input:focus {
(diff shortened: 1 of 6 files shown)


p/themes/Origine/origine.css
```diff
@@ -1186,6 +1186,7 @@ a.btn {
 		padding: 3px 5px;
 		max-width: 97%;
 		width: 90px;
+		line-height: 25px;
```
Frenzie: That should preferably probably be something like 1.5 or 1.6, but depending on intent it might be better to instease the top/bottom padding instead? (`padding: 16px 5px` for example)

math-GH: fixed

math-GH:
@Alkarex : Maybe it is better for 1.19 instead of 1.20, because in 1.18 the search box is visible. This fix will do the same as in 1.18.1

Frenzie (approved):

Issue #2013, opened 2018-09-10
Improving the onboarding


marienfressinaud:
When the user creates its account, he's pleased with a placeholder and a link to the subscriptions management page. That might be confusing for some users that are not aware about what is a "feed" or what FreshRSS can do. We could use the first connection to present some basic information such as what is a feed, how to subscribe to a website's feed (e.g. no need to find the complete URL, FRSS can discover it by itself!), what are the main features.

Also today, two feeds are automatically added (release and old freshrss.org). We could offer the user to add the feeds (and some more?) by himself. I see it by completing the "empty" placeholder with a list of ~5 feeds and buttons to subscribe to them.

Note that my suggestions would be better discussed and tested with "real" users :)

math-GH:
Is it a relevant target user group of FreshRSS who does not know the basics of RSS feeds?

Frenzie:
Not principally perhaps, but yes of course?

PR #4039, opened 2021-12-07
Improve: config option for "Add some feeds" link in "no articles" alert

math-GH:
User Story I: As a power user I do not like the "Add a feed" in the "no articles" alert, because I know how to add feeds.
User Story II: As a new user I want to know how to add (more) feeds, so that I can consume more news.

- add a option in "Reading"
- default: option is enabled, so that new users get the link. 

settings (see last line):
grafik

default: enabled (checked) 
grafik

disabled (unchecked):
grafik

1. switch the config
2. see the normal/reader view with a filter, that has no articles

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] do the i18n after your feedback

### DIFF
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -112,6 +112,7 @@ class FreshRSS_configure_Controller extends Minz_ActionController {
 			FreshRSS_Context::$user_conf->display_categories = Minz_Request::param('display_categories', 'active');
 			FreshRSS_Context::$user_conf->hide_read_feeds = Minz_Request::param('hide_read_feeds', false);
 			FreshRSS_Context::$user_conf->onread_jump_next = Minz_Request::param('onread_jump_next', false);
+			FreshRSS_Context::$user_conf->no_article_add_feed_link = Minz_Request::param('no_article_add_feed_link', false);
 			FreshRSS_Context::$user_conf->lazyload = Minz_Request::param('lazyload', false);
 			FreshRSS_Context::$user_conf->sides_close_article = Minz_Request::param('sides_close_article', false);
 			FreshRSS_Context::$user_conf->sticky_post = Minz_Request::param('sticky_post', false);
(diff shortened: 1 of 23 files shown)


math-GH:
draft: waiting for your feedback of this config. Later I will do some i18n translations

Frenzie:
I don't think this should be an option cluttering the settings. This kind of minor power user customization of minute details is a perfect use case for CustomCSS. [Edit: if it's not easily targeted at the moment, give it a class/id.]

But perhaps your user story 1 is more bothered by the string *Please add some feeds* than by the actual fact that there's a link or button? We could just remove that string and reuse `Add a feed` from the dedicated page.

math-GH:
I am thinking one step further: Should FreshRSS have an "beginner" and "expert" mode? 

See this idea #2013

As I think that most of FreshRSS installed application are used by 1 user each (mostly an IT expert who installed it on own server). This user uses FreshRSS for a long time so he/she is/becomes an "expert" in RSS and FreshRSS.

Frenzie:
There should be a skip button on the introductory wizard described by @marienfressinaud.

The link/button under discussion could be part of that, but I remain of the opinion that it should be either be always there or removed completely. I'm additionally mildly in favor of keeping the link/button, especially on mobile. It's a minor convenience that doesn't get in the way, that provides some filler on what otherwise looks very empty.

math-GH:
Why should it kept especially on mobile?

Frenzie:
Screenshot_2021-12-07-21-47-08-59_d365b52accad0f47adbc08c16219827d

Frenzie:
Note that I'm talking about the concept, not this specific form.

math-GH:
you do not use the edge branch ;) On edge you will find the settings button right top corner.
To add a feed just click on the folder icon (left) to see the subscription management button.

Frenzie:
Top left like on this screenshot or top right with one tap less makes no real difference compared to the much easier to reach middle. But like I said it's the proposed option I'm opposed to, not removal of the concept. Although I do think it's more aesthetically pleasing to put something else there than just the text, regardless if it might be mainly decorative or slightly redundant.

PR #4058, opened 2021-12-12
Search on article IDs

### LINKED ISSUE #4053: [Feature] Want every article has spearate url to make the site looks like a news website.Can we do this or have any plugin can do this ?
[Feature] Want every article has spearate url to make the site looks like a news website
Can we do this or have any plugin can do this ?

Alkarex:
Partial implementation of https://github.com/FreshRSS/FreshRSS/issues/4053

New search keyword:

* by entry (article) ID: `e:1639310674957894` or multiple entry IDs  (*or*): `e:1639310674957894,1639310674957893`

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -732,6 +732,29 @@ SQL;
 				}
 				$sub_search = '';
 
+				if ($filter->getEntryIds()) {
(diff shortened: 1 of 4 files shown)


Frenzie (approved):
Why not. :-)

Alkarex:
Using the reading view and the new entry ID filter, it is now possible to achieve something like that:
image

PR #4061, opened 2021-12-12
Improve: Aside slider: close slider when click on the outside of slider

math-GH:
Before:
When the left hand slider in mobile view is open, than it does not close when you click outside of the slider.

After:
the slider closes.
grafik

- CSS and phtml

1. use mobile view
2. open a left hand side slider (f.e. in config, normal view/reader view, subscription management
3. click on the right to close the slider

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Frenzie (approved):
Looks fine at a glance, didn't test but I trust it works. ^_^ Same method as for non-mobile?

PR #4063, opened 2021-12-13
Add HTTP_REMOTE_USER header for auth

drosoCode:

- Add a new header for the `http_auth` method: `HTTP_REMOTE_USER`
This header is received when using the `linuxserver/freshrss` docker image (which is using nginx and prefixes the headers with `HTTP_`) and Authelia for authentication.

1. Set `auth_type` to `http_auth` in `config.php`
2. Start a FreshRSS container with the `linuxserver/freshrss` image
3.  Make a request to FreshRSS with a `REMOTE_USER` header

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Alkarex:
We need to study a bit the consequences. Anybody can send `HTTP_REMOTE_USER` while `REMOTE_USER` can only be sent by the HTTP server. We could have to introduce a white list of IP addresses allowed to send this header.

drosoCode:
Indeed, this could be dangerous if FreshRSS is directly exposed. I have added an IP whitelist option to use this header. What do you think of this ?

Alkarex:
We probably have to do the same for `HTTP_X_WEBAUTH_USER`.
Furthermore, it would be nice to search if there is maybe any native PHP helper for working with IP addresses.
In particular, the current patch seems to work only with legacy IPv4 and not for IPv6.

drosoCode:
I have updated the code to be compatible with IPv6 and added the IP verification for `HTTP_X_WEBAUTH_USER`.
Unfortunately it seems that there is no built-in PHP function aside from `inet_pton` to work with IP addresses (see the list [here](https://www.php.net/manual/en/ref.network.php)).

math-GH:
@drosoCode Any progress here?

Alkarex:
> Any progress here?

It is mostly me in late for the (security) review

Alkarex (approved):

Alkarex:
@drosoCode Sorry for the delay. I have made a few fixes and added two default trusted sources. Could you please double-check, in particular the IPv6 one?

lib/lib_rss.php
```diff
@@ -554,15 +554,68 @@ function get_user_configuration($username) {
...
 /**
  * @return string
  */
 function httpAuthUser() {
 	if (!empty($_SERVER['REMOTE_USER'])) {
 		return $_SERVER['REMOTE_USER'];
+	} elseif (!empty($_SERVER['HTTP_REMOTE_USER']) && checkTrustedIP()) {
+		return $_SERVER['HTTP_REMOTE_USER'];
 	} elseif (!empty($_SERVER['REDIRECT_REMOTE_USER'])) {
 		return $_SERVER['REDIRECT_REMOTE_USER'];
```
Alkarex, line 617: Not necessarily in this PR, but we also need to double-check the security of that, especially for non-Apache servers

drosoCode:
I have tested the changes on my server, I can confirm that the IPv4 works. The IPv6 should also work, however I haven't tested it as it is not configured on my server.

Alkarex:
We might want to add a Docker environment variable to specify the trusted sources

kristof-mattei:
> We might want to add a Docker environment variable to specify the trusted sources

While this could improve security I think it's enough only trust the headers in the case the user has explicitly configured and enabled proxy auth. 

This implies that the proxy is trusted, and that the proxy is the only way to get to the endpoint (e.g. FreshRSS) 

Using proxy authentication also implies that it's proxy's responsibility to be the gatekeeper of those headers, meaning if there is an outside call coming in with those headers set, it should not pass those through. 

And in the case of non-proxy auth, we're not reading the headers anyway.

Alkarex:
Reworked in https://github.com/FreshRSS/FreshRSS/pull/5549
Tests and feedback welcome

PR #4071, opened 2021-12-18
No-auto-collapse

### LINKED ISSUE #4069: [Feature] [Question] UI issues: strange navigation on article click & forced autoload
I apologize for such a vague title. There are several very annoying UI decisions in an otherwise an excellent system, so I wanted to ask whether this behavior is intended or "to be improved". Here are the two I face most often.

1) When I read an article and want to move to the next one, the system navigates to some counter-intuitive place, in most cases closer to the end or even past the end of the article I am going to read. This can be seen in the gif below:

Animation2

2) Suppose I have a large batch of unread articles (~300). I can quickly scan the titles of the first page of the batch, maybe open and read some of them. Then I want to mark all these articles as "read" to remove them and proceed to the next batch. This is, for example, how Inoreader works. In FreshRSS I have to click on every single title to mark it as "read" as I go. If I approach the end of the page, the system loads the next unread articles, so I am forced to process them all in one sitting. In my view, this is unfortunate. I'd prefer to be able to have a button to mark all the articles of the page as "read" and then display the next page, not to proceed further automatically.

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/4069
Idea to be investigated. Could also be an option, or could be used only when the "sticky" is disabled.


Alkarex:
For the sake of simplicity, I think I would be quite OK merging as it is, and see if anybody reacts one way of another.
Any feedback?

Frenzie:
Does that mean everything will remain open when you press next?

Alkarex:
> Does that mean everything will remain open when you press next?

Indeed @Frenzie , as this PR is, articles are not automatically closed

Frenzie:
I'm not in favor of that. I think the "jumpy" behavior makes more sense.

rg-software:
> I'm not in favor of that. I think the "jumpy" behavior makes more sense.

Just came to check the status update :)

After using the system for a while, I believe that a typical user should enable the "Stick the article to the top when opened" option. One might not like this behavior, but it is consistent and understandable.

However, you can never guess what's going to happen when this option is unchecked (at least, on a desktop). If you have one article uncollapsed, then go a few articles down and open some other article there, you wouldn't be able to predict what is going to be visible on the screen.

Maybe this option makes little sense at all or I am missing something.

rg-software:
Just another quick status update three years later. I still find this feature absolutely essential. Every time I update FreshRSS, I manually edit the JavaScript as suggested here and disable auto-collapse.

I am not against auto-collapsing in principle, but its current implementation for the case of unchecked "Stick the article to the top when opened" option is frustrating and counter-intuitive. You click on the title to uncollapse an article and arrive somewhere in the middle of the text and have to scroll back to the beginning.

No auto-collapsing option is a workaround, but it is vastly superior to the present behavior.

polybjorn:
Hi, @rg-software 

Reading this PR and related issue, trying to understand. You've called this PR a workaround rather than the fix, and said you're not against auto-collapse in principle. The specific complaint is the unchecked Stick-to-top state, where clicking a title lands you mid-body.

One possible direction, if I'm reading that right: when an article above your click collapses, shift scroll up by the height it gave back, so visually the title under your cursor stays put. Auto-collapse keeps working as it does today; the off-state of Stick-to-top becomes predictable. It would only fail near the top of the list, where there isn't enough scroll headroom to absorb the shift.

A few specifics I'd be curious to hear more about:

1. Keyboard nav (n / j / arrow) has no click point. What should the viewport do there?
2. Near the top of the list, the click point moves by whatever can't be absorbed. Acceptable, or dealbreaker?
3. Is your problem with Stick to top the motion itself, or where it lands? If a variant kept the title at your click position instead of at the top, would that change anything?
4. When auto-collapse is off (with your patch), do you make use of having multiple articles expanded at once? Or is the value purely that nothing moves when you click?

rg-software:
Thank you for getting back to me. I have updated the system to make sure I have the latest version (1.29.0).

In Settings, Reading, Articles: Open/Close, there are three options: Show articles unfolded, Stick the article to the top when opened, Clicking outside article closes it.

We are discussing the case when all these options are unchecked. My point is very simple: there may be several ways to handle this option _right_, but the current behavior is obviously _wrong_, since clicking the article to uncollapse it brings the user to an undefined location in the text. There is no way to know where to scroll to get to the beginning of the text, because it depends on the length of the previously uncollapsed article.

To me, uncollapsing an article is like uncollapsing a spoiler tag on a webpage. Personally I think that there is no need to collapse back previously uncollapsed articles, but it is merely a preference, either option is fine. However, it seems natural to me that uncollapsing a spoiler tag should not navigate me elsewhere: I am clicking on a title to read the content, so I presume the title stays right there (under the mouse cursor), and the content is below.

For people who like their article at the top of the screen there is a "Stick to the top" option; we're discussing how the all-unchecked combo should behave.

Keeping this point in mind,

1) Keyboard navigation actually works fine -- say, pressing j brings the next article to the top and opens it. It is possibly not the only way to do it, but perfectly sensible.

2) Not sure what is the scenario here.

3) I don't think that "stick to the top" is a bad option per se, just not my preference. I am trying to make a case here that the combination "three checkboxes unchecked" is legitimate (i.e., can be selected) yet produces inconsistent behavior. It's not about what would make me want to use "stick to the top" option.

4) Personally I don't find uncollapsed articles distracting, so for me keeping them uncollapsed is fine even though I normally don't benefit from this behavior. Again, my mental model is of a "spoiler tag". Others may argue it's like pages in a book so you must close the previous page to open the next page.

Frenzie:
Some follow-up in #8870.

Issue #4070, opened 2021-12-18
[BUG] demo.freshrss.org: theme changed + Hungarian feeeds


math-GH:
Please check demo.freshrss.org

The anonymous user has now not the Origine theme and displays f.e. a Hungarian feed

![grafik](https://user-images.githubusercontent.com/1645099/146643518-49c47f42-6127-48ae-9d58-48ce3efa2a00.png)

Expected:
- Origine theme
- English feeds

Frenzie:
You can log in with demo/demodemo to play around with it. I believe it's automatically reset once an hour or once a day.

math-GH:
Yesterday I saw it too.

math-GH:
let us clarify how often there is a reset..

(I changed the theme back to Origine and deleted the Hungarian feeds right now)

Frenzie:
@marienfressinaud

marienfressinaud:
I reset the demo manually once in a while, mostly when I update the server. I just did it.

Btw, I saw the mentions on the other tickets but I need to take a break from Internet so I'll answer after the holidays.

Frenzie:
No problem. Happy holidays!

marienfressinaud:
I just created a small extension to fill the credentials on the demo instance and to reset the instance every night via a cron task: https://github.com/marienfressinaud/xExtension-Demo

I don't think it should be listed in the https://github.com/FreshRSS/Extensions list, but I will transfer the ownership of the repository to the FreshRSS organization. What do you think?

Frenzie:
Sounds good to me. :-)

(But listing it could perhaps be useful too, for some people? On the flip side, it'd mostly just be noise I suppose.)

math-GH:
I would like to see it in the extension list because every extension could be a good source to develop a new extension.

marienfressinaud:
A better place would be in the developers documentation then, I think the extensions list is more intended to administrators who want to discover extensions (and there are good chances they don't want to install this one :D)

Frenzie:
The "extension" bit seems somewhat… redundant, in any case. It's basically just a sample script that can/should be triggered with cron. The extension around it is just so you've got a reminder in the UI.

math-GH:
@marienfressinaud : Right now it is possible to delete the extension in the demo instance, because the demo user has administrator rights. After the deletion the reset script would not be available.
To mark the extension as system extensions, would not solve this issue. 
[...]

PR #4072, opened 2021-12-18
Improved: Manual: username+pw for demo login

math-GH:
References #4070

- add credentials to login into demo.freshrss.org

Pull request checklist:

- [x] clear commit messages

### DIFF
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -31,7 +31,12 @@ This documentation is split into different sections:
 
 ## Demo
 
-The official demo of FreshRSS is available under [https://demo.freshrss.org/](https://demo.freshrss.org/).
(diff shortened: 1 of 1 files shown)


Frenzie:
I got it from the manual btw, grep for demodemo to see where (and as to whether it should be removed there if you add it here).

math-GH:
> I got it from the manual btw, grep for demodemo to see where (and as to whether it should be removed there if you add it here).

I added the information to the place where everyone, who is interested in FreshRSS, can find the credentials.

Before, the credentials are documented in the upgrade information section (https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/developers/05_Release_new_version.md), that is only interesting for 1 or 2 people of us.

I would leave it as here as well

Frenzie (approved):
Mostly for @marienfressinaud I figure?

Issue #4073, opened 2021-12-19
[BUG] Compressed & Cached JS files


math-GH:
I see, that JS files are not compressed and no a cache time is set

![grafik](https://user-images.githubusercontent.com/1645099/146681112-35a55146-d5cf-47bf-9a16-036ce070d1c7.png)

I found out, that a `.htaccess` file in ./p/scripts with this content could set a cache-control value
```
<IfModule mod_expires.c>
	ExpiresActive	on
	ExpiresByType application/x-javascript "access plus 1 month"
	ExpiresByType text/javascript "access plus 1 month"
	ExpiresByType application/javascript "access plus 1 month"
</IfModule>
```

But I could not find the reason, why CSS files are compressed but not the JS files.

Any idea?

Alkarex:
Please check that your `.htaccess` files have not been moved:
* https://github.com/FreshRSS/FreshRSS/blob/edge/p/.htaccess
* https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/.htaccess

Both JavaScript and CSS files are supposed to be compressed and cached.

Maybe this section could cause problem:

https://github.com/FreshRSS/FreshRSS/blob/1ac3effce53e39d89272dfc95b0c68fcaa122dfb/p/.htaccess#L37-L39

Alkarex:
But this made me find a bug in our Docker configuration https://github.com/FreshRSS/FreshRSS/pull/4074

math-GH:
After the PR:
![grafik](https://user-images.githubusercontent.com/1645099/146835267-f96f91a3-3440-453c-a002-ae655fa15bf9.png)

Alkarex:
@math-GH On which platform did you experience the problem?

math-GH:
> @math-GH On which platform did you experience the problem?

On my web server (Apache 2.4)

Frenzie:
public means it may be cached; does the HTTP spec say that needs a max-age to go along with it?

Alkarex:
Cache duration and cache visibility / sharing are distinct aspects

Alkarex:
@math-GH On which OS?

math-GH:
@Alkarex 
![grafik](https://user-images.githubusercontent.com/1645099/146993447-a7301173-88fd-4182-91d8-644bc80a12e4.png)

PR #4074, opened 2021-12-19
Docker fix missing .htaccess

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/4073
In our Docker configuration, `.htaccess` files are included only once at startup. The one for themes was missing.

### DIFF
--- a/p/.htaccess
+++ b/p/.htaccess
@@ -7,10 +7,7 @@ AddDefaultCharset	UTF-8
 
 <IfModule mod_mime.c>
 	AddType application/json .map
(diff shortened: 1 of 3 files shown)


p/.htaccess
```diff
@@ -7,10 +7,7 @@ AddDefaultCharset	UTF-8
 
 <IfModule mod_mime.c>
 	AddType application/json .map
-	AddType application/font-woff .woff
```
Alkarex, line 10: Duplicates in themes

PR #4101, opened 2021-12-30
Improved: Article header

### LINKED ISSUE #1867: [Feature Request] Make feed tags visible by default
Hello, I'm sorry if this has been brought up before but it didn't come up in any search.

An option for making the feed tags immediately visible in the bottom line or next to the feed author, or perhaps in a new section specifically made for them, would be much appreciated.

Right now it appears they are only displayed in the Related tags drop-down menu, which requires a mouse-click to open.

math-GH:
Closes #1867 + some more improvements

# First draft
(final PR see comments below)

Before:
grafik

After:
grafik

in action;
open-collapse-new-header

1. if the article is open, then the source website, title and date is hidden.
2. Source website is now above the title. Tag list is added (as wished in #1867).
3. close button, that collapse the article

Not visible:
HTML5 tag `article` used

1. open/close articles in normal view

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
Draft PR to collect your feedback

Alkarex:
Personally, I like the tags, but I am not so found of the vanishing top line (title, etc.). I can can better like the first line unchanged when opening / closing.

Frenzie:
I concur. Additionally, it hides the closing touch target.

math-GH:
> Additionally, it hides the closing touch target.

Which closing touch target do you mean? The "header" row is unchanged clickable to collapse the article. Now, the close icon gives a visible hint, how to close the article

Frenzie:
> The "header" row is unchanged clickable to collapse the article.

That's the definition of hidden. ;-)

> Now, the close icon gives a visible hint, how to close the article

Fwiw I use that all the time to open the link. If we want a closing X I think I'd rather have it with the actual content.

math-GH:
The final result looks like this:

grafik

math-GH:
Anything left todo?
Final result see previous comment

p/themes/base-theme/template.css
```diff
@@ -949,6 +949,59 @@ a.website:hover .favicon {
 	max-width: 550px;
 }
 
+.content header {
+	color: #666;
+	font-size: 90%;
```
Frenzie: There could also be a `<header>` element inside the content, so I'd rather see a rem value.

math-GH: thanks. fixed it.

Frenzie: Hm, I'm not convinced. :-) I'd rather see a rem value, because em or percent is almost never what you actually want and then you don't have to "fix" anything.

Frenzie: I'd be inclined to say that for `font-size`, a unit like em or percent simply shouldn't ever be used at all, since it's after January 31, 2020 (IE10 EOL).

math-GH: I agree with you. We need to fix it in every theme. There is a mix of `rem`, `em`, `%`

math-GH: fixed it

app/views/index/normal.phtml
```diff
@@ -73,41 +73,54 @@ $today = @strtotime('today');
...
 								$first = false;
-					?>
-	<em><a href="<?= Minz_Url::display(Minz_Request::modifiedCurrentRequest(
-			['search' => 'author:"' . htmlspecialchars_decode($author, ENT_QUOTES) . '"'])
-		) ?>"><?= $author ?></a></em>
-					<?php	endforeach; ?>
-					</div><?php endif; ?>
-				</div>
+						?>
+		<em><a href="<?= Minz_Url::display(Minz_Request::modifiedCurrentRequest(
```
Frenzie: Indentation feels a bit weird around this general area?

Frenzie (approved):
No real comments, I'll leave the indentation to your and/or @Alkarex' discretion.

app/views/index/normal.phtml
math-GH: Thanks for pointing to this mess. Now it is cleaner.

math-GH:
Does it make sense to keep the tag list at the end of the article (it is optional)?

grafik

Alkarex:
> Does it make sense to keep the tag list at the end of the article (it is optional)?

What do you suggest instead? By the way, that list can be quite long

math-GH:
> > Does it make sense to keep the tag list at the end of the article (it is optional)?
> 
> What do you suggest instead? By the way, that list can be quite long

Do you have an example feed with a lot of article tags?

Alkarex:
Half-random example: https://gizmodo.com/rss

image

math-GH:
If now more than 7 tags are listed the dropdown menu is available

grafik

app/views/index/normal.phtml
```diff
@@ -75,41 +75,77 @@ $today = @strtotime('today');
...
+						<?php
+						$tags = FreshRSS_Context::$user_conf->bottomline_tags ? $this->entry->tags() : null;
+						if (!empty($tags)) {
+							?><?= _i('tag') ?><ul class="list-tags"><?php
+							$i = 0;
+							foreach ($tags as $tag) {
+								if ($i < 7) {
+									$i++;
+									?><li class="item tag"><a class="link-tag" href="<?= _url('index', 'index', 'search', '#' . str_replace(' ', '+', htmlspecialchars_decode($tag, ENT_QUOTES))) ?>" title="<?= _t('gen.action.filter') ?>">#<?= $tag ?></a></li><?php
+								} else { // more than 7 tags: show dropdown menu ?>
```
Alkarex: The if/else should likely be outside the foreach

app/views/index/normal.phtml
```diff
@@ -75,41 +75,77 @@ $today = @strtotime('today');
...
-						<?php endif; ?>
-						<span><?= $this->feed->name() ?></span></a>
+							<img class="favicon" src="<?= $this->feed->favicon() ?>" alt="✇" loading="lazy" /><?php
+						endif; ?><span><?= $this->feed->name() ?></span></a>
 					</div>
-					<div class="date"><?= $this->entry->date() ?></div>
-					<?php
+					<div class="tags">
+						<?php
+						$tags = FreshRSS_Context::$user_conf->bottomline_tags ? $this->entry->tags() : null;
```
Alkarex: Should probably use a new option instead, e.g.:

image

math-GH: great idea.
It is now implemented

app/views/index/normal.phtml
math-GH: why outside of the foreach? Within the foreach the `$i` is increased.

Alkarex: I have coded what I meant: https://github.com/FreshRSS/FreshRSS/pull/4101/commits/bdf754eecdb944dfa3381651b26e4f4d3976f6df
Please have a look.
It is not that is was not working, but more a matter of style: I did not like the fact that two logics were intertwined, requiring another variable to count, a break, and a foreach inside a foreach.

Alkarex:
More or less fine for me now.
One last thing: My personal preference would be to have the feed name on the same line than the author and date, like it was before, to avoid "losing" one line of information density just for that. I do not mind whether it is below the article title (like before) or above.

app/views/index/normal.phtml
math-GH: `array_slice()` was the magic function that I did not know (and unfortunately did not search).
Thanks!

math-GH:
> More or less fine for me now. One last thing: My personal preference would be to have the feed name on the same line than the author and date, like it was before, to avoid "losing" one line of information density just for that. I do not mind whether it is below the article title (like before) or above.

Now (in this PR) the normal view article is similar to the global view (feed name above the title, author and date below).

The more I think about your personal preference and investigate more, the more general questions come up:
- I could bind the different layouts on the `Show tags in article header` configuration: enabled => show the feed name above the tags and title; disabled => show the feed name below the title within the same line of author and date (like before this PR).
- I am not so happy anymore to toggle the `show tags` with the topline configuration. The tags will shown within the article and not in the "topline".
- right now I configured only the normal view but not the `reading view`
- in general the `Article icons` configuration is only valid for the normal view, but not for the `reading view`

Would it make sense do add an additional line `article header` between `top line` and `bottom line`, to configure the article header?

Alkarex:
> Would it make sense do add an additional line `article header` between `top line` and `bottom line`, to configure the article header?

Yes, I think that could indeed be a nice improvement

math-GH:
@Alkarex 
Could you please have a look on the configuration options?
(i18n is still in progress, but it would be nice to have some feedback first)
grafik

Frenzie (approved):
lgtm

app/views/index/normal.phtml
```diff
@@ -72,44 +73,160 @@ $today = @strtotime('today');
...
 					echo $lazyload && $hidePosts ? lazyimg($this->entry->content()) : $this->entry->content();
 				?></div>
-			</div><?php
-
-			$this->renderHelper('index/normal/entry_bottom');
-
-		?></div>
+				<footer>
+					<?php
+					if (FreshRSS_Context::$user_conf->show_author_date == 'h' || FreshRSS_Context::$user_conf->show_author_date == 'b') { ?>
```
Alkarex: All those equalities against a normal string should use triple equal such as `=== 'f'`

Alkarex:
Looks quite fine I guess.

A few comments:

* [x] By default, I got the title displayed also at the bottom. Should probably default to no-change compared to existing settings:

image

* [x] Leaving the reading configuration page shows a change configuration warning, even though there is no change

* [x] *Max number of tags* should probably be *Max number of tags shown* or something like that

app/views/index/normal.phtml
math-GH: fixed

math-GH:
@Alkarex: I fixed all your points

Alkarex:
@math-GH I have fixed a JavaScript error:

image

Please double-check. (Potential side-effect on `?c=configure&a=system`, which uses some of the same logic)

Alkarex (approved):

Alkarex:
P.S. Is it on purpose that a link icon has been added next to the article title?
image

math-GH:
> P.S. Is it on purpose that a link icon has been added next to the article title? image

It is on purpose. The headline is a link to the external article. 
I am not sure if it is a good idea.

Alkarex:
> It is on purpose. The headline is a link to the external article. I am not sure if it is a good idea.

I am not so found of it personally. There is already a clear link, and I think it clutters a bit the interface.

math-GH:
Improved

Alwaysin:
On Nord theme, this PR changes the color of the title from:
chrome_2022-07-23_14-32-20

to:
chrome_2022-07-23_14-34-21

Is this desirable? I find it very hard to read it.

math-GH:
> On Nord theme, this PR changes the color of the title [ ...]
> Is this desirable? I find it very hard to read it.

Thanks. I fixed it with #4459

PR #4102, opened 2021-12-30
SimplePie 1.5.8

math-GH:
SimplePie upgrades to 1.5.8

I guess it could be important because @Alkarex made some PRs.

On some points we have to take a closer look, because there were some lines of code marked with `// FreshRSS`

PR #4011, opened 2021-12-01
Manual update SimplePie

Alkarex:
* https://github.com/simplepie/simplepie/commit/417a1661b2610448bd8e1835d6d4160a0aff9a97
* https://github.com/simplepie/simplepie/commit/ebdd0643ee927edfcaf40bfd27ab23ce9c975120
* https://github.com/simplepie/simplepie/commit/941412027a46979b4f1d29e53edf91829788d781
* https://github.com/simplepie/simplepie/commit/f58a23730f01f26bbcfe02ff9266e337c241be88

Contributes to fixing https://github.com/FreshRSS/FreshRSS/issues/4010

PR #4113, opened 2022-01-01
SimplePie 1.5.8 manual upgrade

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/4011
Instead of https://github.com/FreshRSS/FreshRSS/pull/4102
* https://github.com/simplepie/simplepie/commit/bea5de77afc274dc7c36cac39cdf53de346a5b50
* https://github.com/simplepie/simplepie/commit/5aab77e105d4384c9bb1a747010f380c0506f7a2
* https://github.com/simplepie/simplepie/commit/76be63e72ea6867d9982bf9bd97ba706b196d602

### DIFF
--- a/lib/SimplePie/SimplePie.php
+++ b/lib/SimplePie/SimplePie.php
@@ -33,7 +33,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  *
  * @package SimplePie
- * @version 1.5.6
+ * @version 1.5.8
  * @copyright 2004-2017 Ryan Parman, Sam Sneddon, Ryan McCue
  * @author Ryan Parman
  * @author Sam Sneddon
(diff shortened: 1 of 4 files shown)


lib/SimplePie/SimplePie.php
```diff
@@ -33,7 +33,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  *
  * @package SimplePie
- * @version 1.5.6
+ * @version 1.5.8
  * @copyright 2004-2017 Ryan Parman, Sam Sneddon, Ryan McCue
```
math-GH, line 37: Should we mark the files that are not 100% SimplePie? When I read version 1.5.8, than I would expect, that the source code is 100% SimplePie 1.5.8 without any individual changes.

Alkarex: Ok, but then it is two more lines to update every time :-)

math-GH: Why 2 lines? How would you add this information?

Alkarex: I would rather not, but otherwise the convention (used by e.g. Debian, etc.) is a suffix such as `1.5.8-freshrss`.
The version information is on two different lines

Alkarex:
Let's make another PR for the version thing if wanted

PR #4021, opened 2021-12-02
Add PHPStan

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4016
https://phpstan.org/

```sh
composer run-script phpstan
```

PR #4116, opened 2022-01-02
Fix regression user init

### LINKED ISSUE #4115: [BUG]Error during context user init!
After following the tutorial at https://freshrss.github.io/FreshRSS/en/admins/06_LinuxInstall.html and go to webpage configuring the database successfully, I went to the url/i/ page and when I logged in for the first time, I was only prompted with“ Error during context user init!" Please help me to solve this problem, thank you!

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4115
Bug introduced by https://github.com/FreshRSS/FreshRSS/pull/4021

### DIFF
--- a/app/Models/UserDAO.php
+++ b/app/Models/UserDAO.php
@@ -5,7 +5,7 @@ class FreshRSS_UserDAO extends Minz_ModelPdo {
 		require(APP_PATH . '/SQL/install.sql.' . $this->pdo->dbType() . '.php');
 
 		try {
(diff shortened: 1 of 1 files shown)


Alkarex:
Damn, I did not look carefully enough. The bug was only in `edge`, not in `1.19.0`, so it was not a true regression...
So now I have made a hastily 1.19.1 point-release. My mistake. I hope it will not break something else 🤞🏻

Issue #4112, opened 2022-01-01
PHPStan level 6


Alkarex:
PHPStan is an automated scanner, which reports code that could be improved https://phpstan.org

We already pass PHPStan level 5 since https://github.com/FreshRSS/FreshRSS/pull/4110

Usage:

```sh
# See https://getcomposer.org
composer install --prefer-dist
composer run-script phpstan
```

Now it would be nice to progress to PHPStan level 6 *report missing typehints* https://phpstan.org/user-guide/rule-levels

https://github.com/FreshRSS/FreshRSS/blob/1335a0e3cf11a0d4248e9eaaf748b89e6df741ef/phpstan.neon#L2-L3

```sh
# Manual override of the analyse level
vendor/bin/phpstan analyse . --level 6
```

Help welcome, and it is fine to send multiple PRs, each solving some of the PHPStan warnings (best if covering at least one or more files fully).

When you make more files pass Level 6, you can remove them ([example](https://github.com/FreshRSS/FreshRSS/commit/a5cc7b20f6b05ceae73a90dc418345f7f0dede6f)) from our [list of non-compliant files](https://github.com/FreshRSS/FreshRSS/blob/edge/tests/phpstan-next.txt) (cf. https://github.com/FreshRSS/FreshRSS/pull/5230 )

```sh
composer run-script phpstan-next
```

math-GH:
Are we already on level 6 with 1.20.0-dev?

Alkarex:
> Are we already on level 6 with 1.20.0-dev?

Nope, we are currently at level 5:

https://github.com/FreshRSS/FreshRSS/blob/0ec3874620c03b03a5cb462d99d710bd802d06f9/phpstan.neon#L3

PR #4110, opened 2022-01-01
PHPStan level 5

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4103 (level 2)
and also level 3, level 4, and level 5 https://phpstan.org/user-guide/rule-levels

PR #4119, opened 2022-01-04
Optimise Minz_ModelPdo::class

ColonelMoutarde:
- Fix typo,
- remove unnecessary null in property,
- remove unused property,
- add phpDoc,
- add ext PDO in composer.json,
- use strict comparison,
- indentation


lib/Minz/ModelPdo.php
```diff
@@ -6,22 +6,27 @@
  */
 
 /**
- * La classe Model_sql représente le modèle interragissant avec les bases de données
+ * La classe Model_sql représente le modèle interagissant avec les bases de données
```
Frenzie: If we're changing the line anyway, should we change it to English?

`The Model_sql class represents the database interaction model`, or something along those lines.

ColonelMoutarde: Done

lib/Minz/ModelPdo.php
```diff
@@ -6,22 +6,27 @@
  */
 
 /**
- * La classe Model_sql représente le modèle interragissant avec les bases de données
+ * The Model_sql class represents the model interacting with the databases
```
Frenzie:
```suggestion
 * The Model_sql class represents the model for interacting with databases.
```

lib/Minz/ModelPdo.php
```diff
@@ -68,22 +73,27 @@ private function dbConnect() {
...
 		if ($currentUser === null) {
 			$currentUser = Minz_Session::param('currentUser');
 		}
-		if ($currentPdo != null) {
+		if ($currentPdo !== null) {
 			$this->pdo = $currentPdo;
 			return;
 		}
-		if ($currentUser == '') {
+		if ($currentUser === '') {
```
Frenzie: I'm not sure if stricter matching is necessarily intended here if it was written like that.

Frenzie (commented):
Looks fine to me in principle but @Alkarex should be more familiar with the rationale here. I'm less sure about the strict comparison changes for example.

lib/Minz/ModelPdo.php
ColonelMoutarde: The strict comparison is always more qualitative than the non-strict comparison. The non-strict one has too much risk of side effects.
Look at the table in the PHP documentation
https://www.php.net/manual/en/types.comparisons.php

Alkarex:
Let me merge https://github.com/FreshRSS/FreshRSS/pull/4110 first, to avoid more merge conflicts

lib/Minz/ModelPdo.php
Alkarex: Changing to `===` would mean that you believe that `false`, `0`, `''`, are valid usernames, which is not the case.

ColonelMoutarde: If I set === and $currentUser is an empty string, then an exception is thrown if we don't want 0, false to be accepted, so let's make it explicit.

ColonelMoutarde: changed

Alkarex:
As you can see, the automated tests are failing.
You can run some of them locally, e.g. with `composer test`

Alkarex:
Related https://github.com/FreshRSS/FreshRSS/issues/4112

lib/Minz/ModelPdo.php
Frenzie: Incidentally, my comment was not intended for just this one specific line but just as a representative example. They were all explicitly written to be truthy or falsy, otherwise they would've been strict already.

ColonelMoutarde: ;) https://stackoverflow.com/questions/2401478/why-is-faster-than-in-php

Frenzie: My point isn't that I like `($currentUser == '')` better than `($currentUser === '' || $currentUser === null || $currentUser === false || $currentUser === 0)`)[1] but that it entails all that and produces different results. I.e., that `($currentUser == '')` isn't the same as `($currentUser === '')`. That's the same for probably every instance where you replaced `!=` or `==` by the stricter `!==` or `===`. It's quite unlikely that truthy and falsy were used by mistake (in this particular codebase).

[1] Though for reading I probably do since it says all that so much quicker. :-) But that's neither here nor there.

lib/Minz/ModelPdo.php
```diff
@@ -114,20 +127,43 @@ public function __construct($currentUser = null, $currentPdo = null) {
 			);
 	}
 
-	public function beginTransaction() {
+	/**
+	 * @return void
+	 */
+	public function beginTransaction()
+	{
```
Alkarex: Please keep same curly-brackets alignment than the rest of the document

lib/Minz/ModelPdo.php
```diff
@@ -96,10 +109,10 @@ public function __construct($currentUser = null, $currentPdo = null) {
 			try {
 				$this->dbConnect();
 				return;
-			} catch (PDOException $e) {
+			} catch (Minz_PDOConnectionException $e) {
 				$ex = $e;
-				if (empty($e->errorInfo[0]) || $e->errorInfo[0] !== '08006') {
```
Alkarex, line 101: The filter on error info `08006` seems gone

lib/Minz/ModelPdo.php
```diff
@@ -67,22 +75,27 @@ private function dbConnect() {
...
+	 * @throws Minz_PDOConnectionException
 	 */
-	public function __construct($currentUser = null, $currentPdo = null) {
+	public function __construct($currentUser = null, $currentPdo = null)
+	{
 		if ($currentUser === null) {
 			$currentUser = Minz_Session::param('currentUser');
 		}
-		if ($currentPdo != null) {
+		if ($currentPdo !== null) {
```
Frenzie, line 102: You reverted the one, but as I said I was talking about all of them. This particular one does feel a little suspicious in the original in the sense that using null instead of '' is a bit unidiomatic (in this codebase) but when I read this, I very much do read it as an intentional $currentPdo isn't an empty string (or however unlikely 0 or false) either.

lib/Minz/ModelPdo.php
```diff
@@ -96,10 +109,10 @@ public function __construct($currentUser = null, $currentPdo = null) {
 			try {
 				$this->dbConnect();
 				return;
-			} catch (PDOException $e) {
+			} catch (Minz_PDOConnectionException $e) {
```
Alkarex: We also need to catch other PDO exceptions

Alkarex (changes requested):
See comments

lib/Minz/ModelPdo.php
```diff
@@ -67,22 +75,27 @@ private function dbConnect() {
...
+		if ($currentPdo !== null) {
 			$this->pdo = $currentPdo;
 			return;
 		}
 		if ($currentUser == '') {
 			throw new Minz_PDOConnectionException('Current user must not be empty!', '', Minz_Exception::ERROR);
 		}
-		if (self::$usesSharedPdo && self::$sharedPdo != null &&
-			($currentUser == '' || $currentUser === self::$sharedCurrentUser)) {
+		if (self::$usesSharedPdo && self::$sharedPdo !== null &&
```
Frenzie: Also these.

Frenzie (commented):
(I'm not saying I'm against changing these lines, but in the diff it looks like a seemingly minor aesthetic swap with grave consequences in logic.)

lib/Minz/ModelPdo.php
Alkarex: Although I believe that the original code is safer for now, the new type hints will mitigate the issue by warning us if we have code risking calling with something that is not of type `PDO|null`

lib/Minz/ModelPdo.php
ColonelMoutarde: ok, your not PSR complient ?

lib/Minz/ModelPdo.php
ColonelMoutarde: Go to your mattermost ?

ColonelMoutarde: https://framateam.org

lib/Minz/ModelPdo.php
ColonelMoutarde: This case is tagged to "ok" by phpstan PDOException not

lib/Minz/ModelPdo.php
Frenzie: I see, that's a bit of rationale missing in the commit message that I overlooked. To what extent can these type hints be trusted though?

lib/Minz/ModelPdo.php
Alkarex: See https://github.com/FreshRSS/FreshRSS/blob/edge/phpcs.xml
In general, please keep the convention used in a given file

Alkarex:
Slightly related, I have just updated https://github.com/FreshRSS/FreshRSS/pull/4121 to be able to run tests in PHP 8.1, which also shows that we have a number of PHP 8.1+ warnings to address

lib/Minz/ModelPdo.php
Alkarex: > To what extent can these type hints be trusted though?

Only when we run PHPStan for now, but that is a part of the CI. But Minz (in theory) can also be used outside this context.

lib/Minz/ModelPdo.php
Alkarex: Maybe, but that probably hides another problem, because this is wrong

lib/Minz/ModelPdo.php
Alkarex: @Frenzie If you have a minute, could you please check why this not caught by PHPCS?

Frenzie: At a glance it looks like we only demand a newline after an opening brace and I guess no one's ever tried to use this style before. Probably easy to work around, let me see if I can figure it out in roughly 5 minutes (since I should sleep soon :-) ).

Frenzie: https://github.com/FreshRSS/FreshRSS/pull/4122

Alkarex (approved):

composer.json
```diff
@@ -17,7 +17,9 @@
         "WebSub"
     ],
     "require": {
-        "php": ">=7.0.0"
+        "php": ">=7.0.0",
+        "ext-pdo": "*",
+        "ext-simplexml": "*"
```
Alkarex: This could better be in another PR, but that does not matter, and the list could be extended:
https://github.com/FreshRSS/FreshRSS/blob/8e398d24f132095fffba0db3c7ceea9a43371b4d/Docker/Dockerfile-Alpine#L7-L9

Alkarex:
@ColonelMoutarde I have fixed the Exception issue along with other minor changes. You are welcome to have a look

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/4140

PR #4121, opened 2022-01-04
Docker :newest Alpine PHP 8.1

Alkarex:
Update our `:newest` development image to use PHP 8.1 in `alpine:edge`.

This image is not on Docker Hub so it needs to be built manually:

```sh
docker build --pull --tag freshrss/freshrss:newest -f Docker/Dockerfile-Newest .
```

And then, assuming `composer` is already installed in the local `./bin/` directory:

```sh
cd ./FreshRSS/
docker run --rm -v $PWD:/var/www/FreshRSS freshrss/freshrss:newest bin/composer test
```


Alkarex:
And update PHP's `memory_limit` to be able to run the development tests

PR #4122, opened 2022-01-04
[CI] PHPCS: check for opening brace on same line

Frenzie:

1. Add a brace on a newline

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -1088,9 +1088,11 @@ SQL;
 		}
 	}
 
+	/**
+	 * For API
+	 */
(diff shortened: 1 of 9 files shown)


Frenzie:
@Alkarex As you can see it seems to be working, but I don't have the time to figure out if the auto-fix takes care of it properly today.

Alkarex:
I ran `make fix-all` and a few minor manual adjustments.
The rule does not seem to like inline functions `{ ... }` but we do not had that many so it might not be a problem.

PR #4012, opened 2021-12-01
Fix some PHP 8.1 warnings

Alkarex:
The proper fix will have to wait till we drop PHP7.
#fix https://github.com/FreshRSS/FreshRSS/issues/4010

PR #4123, opened 2022-01-04
Fix several PHP8.1 warnings

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/4012

Taking advantage of https://github.com/FreshRSS/FreshRSS/pull/4121

Allow our default CI to pass without warnings

```sh
# Run all tests with newest PHP 8.1:
docker run --rm -v $PWD:/var/www/FreshRSS freshrss/freshrss:newest bin/composer test

# Run all tests with default PHP 7.4
docker run --rm -v $PWD:/var/www/FreshRSS freshrss/freshrss bin/composer test
# Two inconsequential warnings in PHPStan due to method signatures updated from PHP7 to PHP8:
# ob_implicit_flush and simplexml_load_string

# Run minimal syntax tests with oldest PHP 7.0:
docker run --rm -v $PWD:/var/www/FreshRSS freshrss/freshrss:oldest \
  sh -c "find . -type d -name 'vendor' -prune -o -name '*.php' -print0 | xargs -0 -n1 php -l 1>/dev/null"
```

### DIFF
--- a/lib/lib_phpQuery.php
+++ b/lib/lib_phpQuery.php
@@ -4068,9 +4068,11 @@ class phpQueryObject
 
 	// ITERATOR INTERFACE
 	/**
-   * @access private
+	 * @return void
+	 * @access private
 	 */
-	public function rewind(): void {
+	#[\ReturnTypeWillChange]
+	public function rewind() {
 		$this->debug('iterating foreach');
 //		phpQuery::selectDocument($this->getDocumentID());
 		$this->elementsBackup = $this->elements;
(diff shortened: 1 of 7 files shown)


lib/lib_phpQuery.php
```diff
@@ -145,7 +145,7 @@ class DOMDocumentWrapper {
 	public $id;
 	/**
 	 * @todo Rewrite as method and quess if null.
-	 * @var unknown_type
```
Alkarex, line 148: I only did a minimum in that library but there are many other problems, and it should soon be fully replaced

PR #4125, opened 2022-01-05
Improved: "Pagination" + load more button

math-GH:
Before;
grafik

The "load more" section below the article stream is called internally as "pagination". Also the HTML is a `<ul>` but it is just one `<li>` item. It is not (anymore) a "real" pagination.

In contrast to it: the log. It has a real pagination. 

Both uses the same CSS class "pagination".

While reading the source code it is not clear about which type of "pagination" I read. 

So I thought to clear it and improve the "load more" section.

After:
The HTML is improved. The user sees a real button that invites to interact ("load more articles").
grafik

Internally this section is called "stream-footer" and so it is clear divided from the pagination in the logs. A good point for future CSS work.

I start here with a Draft PR to collect your feedback.

Right now in all themes it looks fine.

Open todos:
[x] rename app/views/helpers/pagination.phtml
[x] rename/new array i18n (gen.php) of `pagination`
[x] clean theme CSS


Alkarex:
Quick note: this system also works with JavaScript disabled (so far)

Frenzie:
Perhaps we should distinguish between `Load more articles at the bottom of the page` enabled and disabled.

This PR seems to be targeting disabled, and it might be an improvement there, but with standard settings all this extra padding would result in more distraction. The current look is close to ideal for that use case, except ideally it should say "Load*ing* new articles*.

math-GH:
> Quick note: this system also works with JavaScript disabled (so far)

Is it a minimum goal, that FreshRSS should (mainly) work disabled JavaScript?

Alkarex:
> Is it a minimum goal, that FreshRSS should (mainly) work disabled JavaScript?

Indeed, for now, the main features are supposed to work with JavaScript disabled. I am not sure how long we will be able to do that, though

Frenzie:
Depends a bit on how you define main features. ;-)

math-GH:
I changed it a little bit. Now it should work without JavaScript too.

math-GH:
> This PR seems to be targeting disabled, and it might be an improvement there, but with standard settings all this extra padding would result in more distraction. The current look is close to ideal for that use case, except ideally it should say "Load_ing_ new articles*.

I do not understand correct what you mean with the extra padding. You will not see the padding and the button, when the next articles are loaded. Please check it.

Here you can see it (see the scroll bar on the right hand side. It jumps after loading)
scroll-auto-load

Frenzie:
Of course you'll never see it if you scroll that slowly. ;-) Scroll faster and/or pretend your network is bad 3G or some such.

Edit: it's just a minor niggle, but I feel the current restrained indicator is slightly nicer. It doesn't really require any extra effort, just nice to have. ^_^

math-GH:
What is left todo?

Frenzie (approved):
lgtm, although like I said I like the current visuals. :-)

PR #4130, opened 2022-01-07
Draft:Add interface and add typehinting

ColonelMoutarde:
Add the DatabaseDAOInterface


app/Models/DatabaseDAO.php
```diff
@@ -3,7 +3,7 @@
 /**
  * This class is used to test database is well-constructed.
  */
-class FreshRSS_DatabaseDAO extends Minz_ModelPdo {
+class FreshRSS_DatabaseDAO extends Minz_ModelPdo implements DatabaseDAOInterface {
```
Alkarex: Why an interface when there is already a base class `FreshRSS_DatabaseDAO`?

app/Models/DatabaseDAOInterface.php
```diff
@@ -0,0 +1,32 @@
...
+	/**
+	 * @param string $table
+	 * @return array
+	 */
+	public function getSchema(string $table): array;
+
+	/**
+	 * @param array $dao
+	 * @return array
+	 */
```
Alkarex: Comments can be removed when there is no additional information than the native type hints

math-GH:
Is it a draft or an open PR?

ColonelMoutarde:
A draft

Alkarex (approved):

Alkarex:
I have simplified and finalised this PR.
For inspiration, more type hints welcome from the list of those compatible with PHP 7.0 https://php.net/language.types.declarations
In particular: `bool, float, int, string`
As contribution to https://github.com/FreshRSS/FreshRSS/issues/4112

PR #4140, opened 2022-01-09
Fix PHP 7.4 signature warnings

Alkarex:
Revert some type hints creating execution warnings in PHP 7.4 (not in PHP 7.0 nor PHP 8.0 nor PHP 8.1)


Alkarex:
Regression due to https://github.com/FreshRSS/FreshRSS/pull/4119

PR #4146, opened 2022-01-11
Fix SCSS differences to the final CSS

math-GH:
While I worked on the themes, I found out that some SCSS files does not compile to the existing CSS file content. In some cases there are some differences.
So I clean it up with this PR

- delete browser prefixes for transitions (I checked the compatibility. Safari, Firefox and Opera does not have been needed the prefix minimum since 2015)
- little corrections (leading is the existing CSS file).

1. check around the themes

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
Solved themes:
- Ansum
- Mapco

other themes are not affected as I see/know

PR #4172, opened 2022-01-24
Fixed: (No) failed login messages

### LINKED ISSUE #4095: [BUG/Improvement] (No) failed login messages
**Describe the bug**
With wrong credentials, the user does not receive clear error messages

**To Reproduce**
When the user tries to login 
1. correct username but wrong password: reload of the login screen
2. not correct username: Error 403 - Forbidden error page
3. correct username + correct password, but the user is disabled: no information. reload of the login screen

**Expected behavior**
1. a error message, that informs that username and password are not correct
2. a error message, that informs that username and password are not correct (not a generic 403 error page)
3. a error message, that informs that user is disabled/blocked

- FreshRSS version: 1.19.0-dev

math-GH:
Closes #4095

# please test it carefully
I tested it for webform login, but I do not know if it affects other login mechanics.

The message "Login is invalid" is now shown
grafik

- When username is not valid
- When username+password is not valid
- When user is disabled

- use Minz_Request::bad instead of Minz_Error::error and Minz_Request::forward
- added 2 lines of comments

1. try to login with a not registered username (f.e. "asdfsadfsadgasg")
2. try to login with a registered username but a wrong password
3. try to login with a registered username and a correct password, but the user account is disabled
4. try to login with a registered username and a correct password (enabled user)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
At a quick glance, it does not look correct, as it removes the HTTP 403 error codes. Such error codes are used to detect such cases, for instance for preventing brute force attacks (fail2ban or similar tools).

app/Controllers/authController.php
```diff
@@ -125,15 +125,15 @@ public function formLoginAction() {
 
 			FreshRSS_Context::initUser($username);
 			if (FreshRSS_Context::$user_conf == null) {
-				// Initialise the default user to be able to display the error page
-				FreshRSS_Context::initUser(FreshRSS_Context::$system_conf->default_user);
-				Minz_Error::error(403, array(_t('feedback.auth.login.invalid')), false);
+				// Username is unknown
+				Minz_Request::bad(_t('feedback.auth.login.invalid'), ['c' => 'auth', 'a' => 'login']);
```
Alkarex, line 130: Missing HTTP 403 status

Alkarex:
A little tip: have a look at the history of a line to get a better understanding of its purpose, e.g. via https://github.com/FreshRSS/FreshRSS/blame/edge/app/Controllers/authController.php to see where it comes from, in this case https://github.com/FreshRSS/FreshRSS/pull/2903

Here is how it looks like in my editor ([VSCode](https://code.visualstudio.com/) with [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens))

image

app/Controllers/authController.php
math-GH: I cannot find any way to inform the user with a useful text + send the HTTP403 status at once.
If 403 sent, than only a generic message is shown.

Alkarex: We might need something similar than lines 166-169 or expand the Minz functions.
I can have a look, but probably first after we have shipped 1.19.2.

math-GH: Line 168 does not have any effect in this case
https://github.com/FreshRSS/FreshRSS/blob/edge/app/Controllers/authController.php#L168

math-GH: HTTP error 401 would be better than 403. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

Alkarex: It is not for the same case. 401 is used for HTTP Basic Auth in particular.

Alkarex (changes requested):
We need to make sure we keep returning HTTP 403

PR #4173, opened 2022-01-24
Update dev dependencies stylelint SASS

Alkarex:
Major update for stylelint with breaking changes
https://stylelint.io/migration-guide/to-14/

Applied automatic SASS migration to current syntax
https://sass-lang.com/documentation/cli/migrator#migrations

And a few manual fixes such as:
https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/dollar-variable-no-missing-interpolation/README.md

Other dev dependencies are minor updates, including fixing a security vulnerability https://github.com/FreshRSS/FreshRSS/pull/4168


Makefile
```diff
@@ -67,6 +67,9 @@ lint: bin/phpcs ## Run the linter on the PHP files
 lint-fix: bin/phpcbf ## Fix the errors detected by the linter
 	$(PHP) ./bin/phpcbf . -p -s
 
+bin/composer:
+
```
Frenzie: Is this missing something?

Alkarex: Yes, sorry, I committee a draft of this file by accident. I can revert or leave if to future improvements

Frenzie: Probably better to revert the file or fix it up? I don't know what you were going for exactly, maybe we're only talking about a few minutes work. ^_^

Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/4173/commits/e9b7ac9f10b53f295ae5a859d231cc150b7ff350

p/themes/Mapco/mapco.css
```diff
@@ -541,7 +541,7 @@ form th {
 }
 .tree .tree-folder .tree-folder-title .title {
 	background: inherit;
-	color: #fff;
+	color: #ffffff;
```
Frenzie, line 544: This is all on purpose btw? I kind of like the shorthand. :-) (But either way.)

Frenzie (approved):
Looks fine, didn't check the SCSS stuff in detail but the fact that the outcome is identical probably speaks for itself.

p/themes/Mapco/mapco.css
Alkarex: Indeed the shorthand would be better, but there is code such as the following, which would break:

https://github.com/FreshRSS/FreshRSS/blob/9012db01554b9ac3b99651be8a3944668757fa64/p/themes/Ansum/_components.scss#L146

Frenzie: Makes sense.

PR #4174, opened 2022-01-25
Fixed: wrong dropdown triangle of labels category

math-GH:
Summary:
shows the triangle to bottom: dropdown is closed
shows the triangle to top: dropdown is open

## Example 1
Before:
grafik

After:
grafik

## Example 2
Before:
grafik

After:
grafik

## Example 3
Before:
grafik

After:
grafik

- improved the logic of the triangle icon

1. filter: show `only `unread articles, labels has some unread articles
2. filter: show unread `and ` read articles, labels has some unread articles
3. filter: show unread `and ` read articles, labels has `no `unread articles

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/layout/aside_feed.phtml
+++ b/app/layout/aside_feed.phtml
@@ -52,7 +52,7 @@
 		?>
 		<li id="tags" class="tree-folder category tags<?= $t_active ? ' active' : '' ?>" data-unread="<?= format_number($this->nbUnreadTags) ?>">
 			<div class="tree-folder-title">
-				<a class="dropdown-toggle" href="#"><?= _i($t_active ? 'up' : 'down') ?></a>
+				<a class="dropdown-toggle" href="#"><?= _i($t_active || (strlen($class) < 1) || (format_number($this->nbUnreadTags) > 0) ? 'up' : 'down') ?></a>
 				<a class="title" data-unread="<?= format_number($this->nbUnreadTags) ?>" href="<?= _url('index', $actual_view, 'get', 'T') . $state_filter_fav ?>"><?= _t('index.menu.tags') ?></a>
 			</div>
 			<ul class="tree-folder-items<?= $t_show ? ' active' : '' ?>">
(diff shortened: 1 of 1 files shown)


app/layout/aside_feed.phtml
```diff
@@ -52,7 +52,7 @@
 		?>
 		<li id="tags" class="tree-folder category tags<?= $t_active ? ' active' : '' ?>" data-unread="<?= format_number($this->nbUnreadTags) ?>">
 			<div class="tree-folder-title">
-				<a class="dropdown-toggle" href="#"><?= _i($t_active ? 'up' : 'down') ?></a>
+				<a class="dropdown-toggle" href="#"><?= _i($t_active || (strlen($class) < 1) || (format_number($this->nbUnreadTags) > 0) ? 'up' : 'down') ?></a>
```
Alkarex:
```suggestion
				<a class="dropdown-toggle" href="#"><?= _i(($t_active || $class === '' || $this->nbUnreadTags > 0) ? 'up' : 'down') ?></a>
```

Alkarex:
The bug was only with the option to unfold all categories by default, right?

app/layout/aside_feed.phtml
Alkarex:
```suggestion
				<a class="dropdown-toggle" href="#"><?= _i($t_show ? 'up' : 'down') ?></a>
```

math-GH:
> The bug was only with the option to unfold all categories by default, right?

it is independent.

Here my "reading" settings from the screenshots above
grafik

app/layout/aside_feed.phtml
math-GH: Solution could be sooooo simple :) Thanks.
I haven't seen it.

Alkarex (approved):

app/layout/aside_feed.phtml
Alkarex: P.S. You can just commit a suggestion from the Web UI here, no need to make a manual edit and commit

math-GH: I know, but I had to test it locally first ;) So I had to takeover it first into my source code before commiting.

PR #4175, opened 2022-01-26
Improved: Fetch articles with selector but do not delete the class attribute. (Simplepie: new method: rename_attribute)

math-GH:
# My User Story:
As a reader I want to configure the content of fetched articles, so that I can hide parts that I do not need.

## User side:
- the `id` and `class` attributes will not deleted anymore. It will renamed to `data-sanitized-id` / `data-sanitized-class`
- with the extension CustomCSS, it is possible to style the `data-sanitized-class` tags (f.e. `div[data-feed="XX"] div[data-sanitized-class*="CLASSNAME"]`

Here an example, that the fetched article footer could be re-styled with CSS (here with `background-color:red`)
grafik

## technical side:
- added the method `rename_attr()`
- default of `$rename_attributes` is an empty array
- for FreshRSS `$simplePie->rename_attributes('class');` is set, so that the `class` attribute is renamed to `data-sanitized-class`

1. fetch articles with a selector
2. see the HTML
3. see the `data-sanitized-class`
4. do some CSS magic to style it

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] if you like it, I would do a PR to the SimplePie repository


lib/SimplePie/SimplePie/Sanitize.php
```diff
@@ -643,6 +671,17 @@ protected function strip_attr($attrib, $xpath)
 		}
 	}
 
+	protected function rename_attr($attrib, $xpath)
+	{
+		$elements = $xpath->query('//*[@' . $attrib . ']');
+
+		foreach ($elements as $element)
+		{
+			$element->setAttribute('data-sanitized-'.$attrib, $element->getAttribute($attrib));
```
Alkarex:
```suggestion
			$element->setAttribute('data-sanitized-' . $attrib, $element->getAttribute($attrib));
```

Alkarex:
Looks good at a first glance 👍🏻 
You should indeed also try to send a PR upstream https://github.com/simplepie/simplepie

Frenzie (commented):
Awesome!

lib/SimplePie/SimplePie/Sanitize.php
```diff
@@ -145,6 +146,25 @@ public function pass_file_data($file_class = 'SimplePie_File', $timeout = 10, $u
...
+			else
+			{
+				$this->rename_attributes = explode(',', $tags);
+			}
+		}
+		else
+		{
+			$this->rename_attributes = false;
+		}
+	}
```
aledeg: You could remove a layer by doing:
```php
if (!$tags) {
    $this->rename_attributes = false;
} elseif (is_array($tags)) {
    $this->rename_attributes = $tags;
} else {
    $this->rename_attributes = explode(',', $tags);
}
```

I find that it makes the code more readable.

aledeg: You could also change the order like:
```php
if (is_array($tags)) {
    $this->rename_attributes = $tags;
} elseif (is_string($tags)) {
    $this->rename_attributes = explode(',', $tags);
} else {
    $this->rename_attributes = false;
}
```

This will reject everything that is not supported. I don't know if this something we want though. In this case if you have a number, it will not be treated as a string.

Frenzie: I don't think it makes sense to reject anything. If it's available as data-whatever you can use it in CSS or JS; if it's not you can't.

aledeg: The first example is not rejecting anything. It's a rewrite without the layers. The second example is just that, an example :)
You're right about not rejecting things at this level.

math-GH: it is copy&paste from `strip_htmltags()` below (lines 168ff).
I would say that it does not make sense to have another structure.

math-GH:
I changed this PR to draft. Have some discussions in the Simplepie repro. Let's wait for the results there. Had already to change something (some copy&paste faults....)

math-GH:
New function is merged in Simplepie upstream.

Ready for review here

lib/SimplePie/SimplePie/Sanitize.php
```diff
@@ -643,6 +671,17 @@ protected function strip_attr($attrib, $xpath)
 		}
 	}
 
+	protected function rename_attr($attrib, $xpath)
+	{
+		$elements = $xpath->query('//*[@' . $attrib . ']');
```
Alkarex, line 676: A bit late for my feedback, but some kind of sanitising of `$attrib` could have been good. Not there, but in the public function.

math-GH: please help me to understand. What kind of sanitizing do you mean?

Alkarex: If `$attrib` happens to be something like  `]` or any other special character, it will make the XPath crash or worse

PR #4181, opened 2022-01-30
Improved: Extension manager: new style

math-GH:
Before:
grafik

It has some issues:

1. on wide screens the text is far away from the buttons (see screenshot on a not very wide screen)
2. on mobile screen the text overlaps the buttons (see next screenshot)
3. the middle button is a status button. The status is less clear (the text disable/enable is not the status. It is the action when the user clicks on it).
4. the config button (cog icon) is in most screens in FreshRSS very clear. It does not need the text "manage"
5. the remove button is not very often used but very prominent placed, but not in the extension manage slider

grafik

After:
grafik

1. fixed. Text is always close to the buttons
2. fixed. No text overlap anymore
3. prominent status (I call it "switch"). Status is very clear.
4. simple manage/config button
5. remove button is removed/moved to the config slider (see screenshot below)

grafik

Other themes:
Alternative dark theme
grafik

Dark theme
grafik

- better HTML
- a lot of CSS

1. go to the extension configuration
2. see also the nice mouse hover on the switches

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] CSS: RTL
- [x] CSS: styling the other themes
- [x] solved the question: Show system extensions to user?


math-GH:
It is a draft PR to collect your first feedback

Alkarex:
Looks very nice so far 👍🏻 
No change in JavaScript requirements?

app/views/helpers/extension/configure.phtml
```diff
@@ -7,6 +7,20 @@
 
 	<p class="alert alert-warn"><?= $this->extension->getDescription() ?> — <?= _t('gen.short.by_author'), ' ', $this->extension->getAuthor() ?></p>
 
+
+	
+	<?php if (FreshRSS_Auth::hasAccess('admin')) { ?>
+		<form id="form-extension" method="post">
+	<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
```
Frenzie: Indentation's a bit weird here.

Frenzie (commented):
Looks good to me in principle. Also note https://codepen.io/Qvcool/pen/bdzVYW

math-GH:
wow, you are very very fast :)
I had a bug, so that the circle was not on the correct position, when it is enabled. Fixed it (also the screenshots are now fixed)

math-GH:
> Looks good to me in principle. Also note https://codepen.io/Qvcool/pen/bdzVYW

I tried also some different styles, but I do not like this kind of switches, because the status there is not 100% clear (for me. I need to think longer. For me it is a strange UX)

app/views/helpers/extension/configure.phtml
math-GH: fixed.

math-GH:
> No change in JavaScript requirements?

Zero JavaScript here. The "disable/enable" button got just a new style. The logic behind the button was not changed.

Frenzie:
> I tried also some different styles, but I do not like this kind of switches, because the status there is not 100% clear (for me. I need to think longer. For me it is a strange UX)

I'm not really sure what you're talking about? I was saying these do the same thing without introducing svg images.

math-GH:
> > I tried also some different styles, but I do not like this kind of switches, because the status there is not 100% clear (for me. I need to think longer. For me it is a strange UX)
> 
> I'm not really sure what you're talking about? I was saying these do the same thing without introducing svg images.

Thanks for the explanation. Now I see what you mean. I meant another topic ;)  
I like this non-svg solution much more. Will implement it. Thanks for this hint.

math-GH:
@Frenzie : It is made with Font Awesome icons. I am not a fan of it.
I will keep the SVG solution.

grafik

Frenzie:
Font Awesome is an irrelevant detail. Pretend it says something like `✓` or `✔` or `🗸` for example instead of Font Awesome. ;-)

Btw, not shown on the screenshots is that these checkbox buttons are currently a bit too jumpy.

math-GH:
> Btw, not shown on the screenshots is that these checkbox buttons are currently a bit too jumpy.

Do you mean this animation?
switch-animatoin
(information: the gif animation has an bug. Issue: The second switch from the bottom does not hide the icon in the animation)

Frenzie:
Sidenote, Netsurf works for me but colorblind people would prefer a regular checkbox. But I'm not sure if or to what extent we should worry about that. ^-^

Screenshot_20220130_160609

Frenzie:
> Do you mean this animation?

Yes, I'd prefer to do without the horizontal sliding movements.

math-GH:
> Sidenote, Netsurf works for me but colorblind people would prefer a regular checkbox. But I'm not sure if or to what extent we should worry about that. ^-^

It is not a checkbox. It is still the same button as before (click on it and it will reload the page).

I would say that Netsurf and color blind people are not a good combination.

Disabled extensions have an italic text style.

app/views/helpers/extension/details.phtml
```diff
@@ -1,22 +1,16 @@
...
+	$ext_enabled = $this->ext_details->isEnabled();
+	if ($ext_enabled) { ?>
+	<button class="switch active" form="form-extension" formaction="<?= _url('extension', 'disable', 'e', $name_encoded) ?>" title="<?= _t('gen.action.disable') ?>"></button>
+	<?php } else { ?>
+	<button class="switch" form="form-extension" formaction="<?= _url('extension', 'enable', 'e', $name_encoded) ?>" title="<?= _t('gen.action.enable') ?>"></button>
+	<?php } ?>
+<?php } else { ?>
+<?= _t('admin.extensions.system.no_rights') ?>
+<?php } ?>
+<a class="open-slider" title="<?= _t('gen.action.manage') ?>" href="<?= _url('extension', 'configure', 'e', $name_encoded) ?>"><?= _i('configure') ?></a>
```
Frenzie: Note that `_i('configure')` does not at this point provide the context for screen readers that the text "Manage" does, only the less helpful text "gear" (as `⚙`).

Frenzie:
> It is not a checkbox. It is still the same button as before (click on it and it will reload the page).

It's a sliding checkbox. :-) It could just as well be represented as a regular checkbox or as a fake checkbox that is compatible with Netsurf. For example, if you add an enabled/disabled SVG icon much like the one proposed here (but represented by Unicode for the sake of putting it together in 10 seconds):

Screenshot_20220130_164055

> I would say that Netsurf and color blind people are not a good combination.

Maybe, maybe not, but the point is that it wasn't an issue before.

app/views/helpers/extension/details.phtml
Frenzie: PS That can be split off into a separate ticket; it's not directly related. But it's important if you intend to remove text from icons because the icons are clear by themselves to us who can see well.

math-GH: the link has now an `title`.

If we want to improve the config icons than we should improve it everywhere. Please check the other cog icons on other places.

Frenzie: Icons, not cog icons.

math-GH: I do not understand what you mean.
`_i('configure')` is a cog icon. The wrapping `<a>` has an `title="manage"`.

Frenzie: It's not important. You said other cog icons, but the problem (if it's a problem) will exist on any type of icon. ^_^

Off the cuff I think it should probably be something closer to `_i('configure', 'Manage')` (where `Manage` would be the alt and title text for the image, no worrying about the surrounding link/button necessary).

math-GH:
@Frenzie : Please check with Netsurf: What is not supported by Netsurf? `::before`/`::after` or `content:`

Frenzie:
Curiously, I see references to :before/:after in the source code but the answer seems to be a simple yes to both.

Anyway, there's no need to worry too much about it, but @Alkarex reminded us recently that we've been forgetting to test in Netsurf and friends. ^_^

There is a standard way to work around the issue with `@supports` btw.

Something like (spitballin', I'd say it's better to use for example an inner span with some text to `display:none` so it also works as expected in Lynx/Elinks/Links2):
```diff
diff --git a/p/themes/base-theme/template.css b/p/themes/base-theme/template.css
index 28756c87..1c41cb8d 100644
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -335,6 +335,18 @@ a.btn {
        background-color: #eee;
 }
 
+.switch {
+       background-image: url('../icons/disabled.svg');
+}
+.switch.active {
+       background-image: url('../icons/enabled.svg');
+}
+@supports selector(.switch::before) {
+       .switch, .switch.active {
+               background-image: none;
+       }
+}
+
 .switch.active::before {
        background-image: url('../icons/enabled.svg');
        background-position: center center;
```

math-GH:
I know `@supports`. That is why I ask, I would try it similar to your example. But do not forget: grey icon on a green background is for red-green blind people grey on grey ;)

Frenzie:
Like I said, just showing the general concept. It's fugly as written in that poc regardless of contrast, lol.

math-GH:
@Frenzie : Please check the fix for netsurf

math-GH:
I installed Netsurf on my Windows machine. It does not display `background-image: url()`. That is the point, that it does not make sense for me to improve it for netwurf.

Frenzie:
> It does not display background-image: url().

Um, yes it does. See what I wrote above.

math-GH:
So the last commit should work, didn't it?

Frenzie:
Sorry, I misunderstood. Anyway, looks great. ^_^
Screenshot_2022-02-05_17-24-47

Frenzie (approved):
I'm not sold on the movements but the code lgtm.

math-GH:
Just to compare: Netsurf 3.10 on Windows looks like this:

grafik

Frenzie:
That's very strange indeed, since I posted a screenshot from 3.10 as well. Oh well. :-)

math-GH:
Does it make sense to display the system extensions to non-admin users?

(screenshot from "before the PR")
grafik

There are 2 possible improvements:
1. do not display system extensions at all, or
2. display the extensions and show the status (enabled/disabled)

math-GH:
> Does it make sense to display the system extensions to non-admin users?
> There are 2 possible improvements:
> 
>     1. do not display system extensions at all, or
> 
>     2. display the extensions and show the status (enabled/disabled)

I chosed 2

p/themes/Alternative-Dark/icons/disabled-light.svg
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="16"
+   width="16"
+   version="1.1"
+   id="svg8"
+   sodipodi:docname="disabled-light.svg"
+   inkscape:version="1.1.1 (3bf5ae0d25, 2021-09-20)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
```
Alkarex: Can you apply the same SVG cleaning process there?

math-GH: done

p/themes/Flat/icons/disabled-light.svg
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="16"
+   width="16"
+   version="1.1"
+   id="svg8"
+   sodipodi:docname="disabled-light.svg"
+   inkscape:version="1.1.1 (3bf5ae0d25, 2021-09-20)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
```
Alkarex: There are still a few SVGs with verbose proprietary attributes ;-)

math-GH: Thanks. Done

app/views/helpers/extension/details.phtml
```diff
@@ -1,22 +1,24 @@
...
+		$name_class = ' disabled';
+		$action = 'enable';
+		$title = _t('gen.action.enable');
+	}
+	if ($this->ext_details->getType() === 'user' || FreshRSS_Auth::hasAccess('admin')) {?>
+	<button class="switch<?= $button_class ?>" form="form-extension" formaction="<?= _url('extension', $action, 'e', $name_encoded) ?>" title="<?= _t('gen.action.enable') ?>"></button>
+	<a class="open-slider" title="<?= _t('gen.action.manage') ?>" href="<?= _url('extension', 'configure', 'e', $name_encoded) ?>"><?= _i('configure') ?></a>
+	<span class="ext_name<?= $name_class ?>"><?= $this->ext_details->getName() ?></span>
+<?php } else { ?>
+	<button class="switch<?= $button_class ?>" title="<?= _t('admin.extensions.system.no_rights') ?>" disabled></button>
```
Alkarex:
```suggestion
	<button class="switch<?= $button_class ?>" title="<?= _t('admin.extensions.system.no_rights') ?>" disabled="disabled"></button>
```

PR #4197, opened 2022-02-05
i18n CLI improve parsing

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/3932#discussion_r800043744

### DIFF
--- a/cli/i18n/I18nFile.php
+++ b/cli/i18n/I18nFile.php
@@ -54,16 +54,24 @@ class I18nFile {
 		$content = str_replace('<?php', '', $content);
 
 		$content = preg_replace([
-			"#',\s*//\s*TODO#i",
-			"#',\s*//\s*DIRTY#i",
-			"#',\s*//\s*IGNORE#i",
+			"#',\s*//\s*TODO.*#i",
+			"#',\s*//\s*DIRTY.*#i",
+			"#',\s*//\s*IGNORE.*#i",
 		], [
 			' -> todo\',',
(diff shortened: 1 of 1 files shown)


cli/i18n/I18nFile.php
```diff
@@ -54,16 +54,24 @@ private function process(string $filename) {
 		$content = str_replace('<?php', '', $content);
 
 		$content = preg_replace([
-			"#',\s*//\s*TODO#i",
-			"#',\s*//\s*DIRTY#i",
-			"#',\s*//\s*IGNORE#i",
+			"#',\s*//\s*TODO.*#i",
+			"#',\s*//\s*DIRTY.*#i",
+			"#',\s*//\s*IGNORE.*#i",
```
Frenzie, line 59: It's not necessarily a given that we want `.*`, though it seems likely. Alternatively:
```suggestion
			"#',\s*//\s*TODO$#i",
			"#',\s*//\s*DIRTY$#i",
			"#',\s*//\s*IGNORE$#i",
```

Alkarex: Ups, that seems to fail

Alkarex: That's good enough for me for now, although I would be interested in an explanation for why the `$` produces a bug here...

Frenzie: Like I said I think .* should be fine, but I just wanted to emphasize I hadn't given it any thought besides the immediate situation.

PR #4201, opened 2022-02-06
Fix Fever 32 bit ID issue + more PHP type hints

### LINKED ISSUE #4200: Problem using Fever API since FreshRSS 1.19.2
I'm using the app Fiery Feeds using the Fever API on iOS and macOS to read RSS based news using FreshRSS. Since upgrading to FreshRSS 1.19.2 the app Fiery Feeds no longer gets any results from FreshRSS. The changelog of FreshRSS does not mention any changes of the Fever API. The logfile of FreshRSS does not show any errors. Is there anything else I can do to analyse a possible problem with the latest version of FreshRSS?

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4200
Follow up and fix regression from https://github.com/FreshRSS/FreshRSS/pull/4110
Entry IDs need to be strings to work on 32-bit systems. There was an error due to enforced Int64 in the Fever API.

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -232,7 +232,7 @@ class FeverAPI
 						break;
 					case 'feed':
 					case 'group':
-						$before = isset($_REQUEST['before']) ? $_REQUEST['before'] : null;
+						$before = $_REQUEST['before'] ?? '';
 						$this->{$method_name}($id, $before);
 						break;
 				}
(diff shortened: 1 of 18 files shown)


p/api/fever.php
```diff
@@ -485,12 +456,18 @@ protected function getItems() {
...
+			}
 		} elseif (isset($_REQUEST['with_ids'])) {
 			$entry_ids = explode(',', $_REQUEST['with_ids']);
 		} elseif (isset($_REQUEST['since_id'])) {
 			// use the since_id argument to request the next $item_limit items
-			$since_id = ctype_digit('' . $_REQUEST['since_id']) ? intval($_REQUEST['since_id']) : null;
+			$since_id = '' . $_REQUEST['since_id'];
+			if (!ctype_digit($since_id)) {
+				$since_id = null;
+			}
```
Alkarex: That is the most relevant par of the bug fix

Issue #4190, opened 2022-02-01
Sprinkle some aria-label around _i()


Frenzie:
**Describe the bug**
There are some icons around that don't have the most helpful alt attribute. For some context see https://github.com/FreshRSS/FreshRSS/pull/4181#discussion_r795205704

https://github.com/FreshRSS/FreshRSS/blob/05ca0517bbcc782a45692dcde0f9258d3165f6bc/lib/lib_rss.php#L585-L587

https://github.com/FreshRSS/FreshRSS/blob/05ca0517bbcc782a45692dcde0f9258d3165f6bc/app/Models/Themes.php#L71-L127

It occurred to me that I was wrong and that the current alt text would be fine if there were an `aria-label` on it. However, the basic principle of what I wrote is true either way: the best solution would be to pass a more descriptive text along in that function, probably optional with a default because many (e.g., `refresh` wouldn't need that.

(This ticket is mainly so it doesn't slip my mind, but feel free to work on it.)

PR #4220, opened 2022-02-13
Implement Web scraping "HTML + XPath"

### LINKED ISSUE #1796: Add <enclosure> in FreshRSS output rss to be compatible with  podcast software
Hi all,
I played a bit with the code this week-end, my idea being the following : 
- I hear a lot of podcasts coming from various sources, mainly heard on my phone (with the excellent gPodder free software), in disconnected mode (downloading them automatically when I have a wifi access)
- Some podcasts I follow only interess me partially, and I don't want to download all episodes on my phone
- after some months of manually crafting a specific rss feed for the episode I want to hear (basically a copy/paste from the original feed, cherry-picking only the interesting episodes, and sometime missing some), I had the idea of the following flow : 
      - subscribe all potential podcasts in freshrss
      - mark (star) the desired episodes in freshrss
      - use the existing fresrss rss streams  of starred podcasts  (e.g https://mydomain.org/rss/i/?a=rss&get=c_10&state=4 )  reinjected in my podcast reader to automatically download the desired episodes

It is almost working out of the box, the only part missing being an additional <enclosure> in the existing rss output stream to be able for the podcast reader to download the episode.
I did a quick and dirty patch to prototype it, and it worked nicely, I  am now able to reinject the starred episodes in gpodder.
My patch only adds one line in app/views/index/rss.phtml, and looks like : 
                    
            <enclosure <?php echo str_replace("src", "url", strstr(strstr($item->content(),"src"),"controls", true));?> type="audio/mpeg" />

(full example here : https://github.com/danc/FreshRSS/blob/podcast/app/views/index/podcast.phtml )

I don't provide a PR though, because the code would need some improvements I was not able to craft with my limited knowledge of FreshRSS base code. It would need to properly retrieve  the download url, the exact mime type, the length of the .mp3, etc ... and verify the compatibility of the rss with the main podcast practises (or manage a specific output apart of the regular rss.phtml).

So this is a submission of the idea for further discussion and improvements, if you find the idea interesting.
I'd be glad to improve my patch to be able to include it in the regular basecode in the longer term.
Any ideas or suggestions ?

Regards,

Alkarex:
This PR adds a (killer? 🤩) functionality, namely the ability to consume any Web site / HTML source, also when an RSS / Atom feed is not available.

It is based on XPath 1.0, which is natively available in PHP:
* https://www.w3.org/TR/xpath-10/
* https://php.net/class.domxpath
* https://developer.mozilla.org/docs/Web/XPath/Axes

This is a light version of what RSS Bridge offers, but natively inside FreshRSS and much easier to use:
* https://github.com/RSS-Bridge/rss-bridge/wiki/The-collectData-function
* https://github.com/RSS-Bridge/rss-bridge/wiki/XPathAbstract

Using a third-party tool did not scale enough anymore for me, when I just need to quickly add another (simple) source.

FreshRSS can then also republish as RSS.

This can be combined with our existing ability to follow article links to retrieve full article content (use with parsimony).

Contributes indirectly to other issues:
* Better handling of enclosures
* Better RSS outputs for https://github.com/FreshRSS/FreshRSS/issues/1796
* Cache HTTP requests when getting full article content
* Better handling of encoding when getting full article content
* Purge old feed caches regularly
* Keep the original order of articles when they all have the same date (before, we got a random order)
* More PHP type hints
* Typos in French

What remains to be done (other PRs):
* [ ] Preview to help writing the XPath expressions just like for our full article content retrieval (help welcome as I will not be able to do much in the coming two weeks)
* [ ] Consume "JSON +XPath" sources (only very little needs to be changed to support that)
* [x] A few new TODOs added in the code (in particular an easy performance improvement related to iterating over SimplePie entries) https://github.com/FreshRSS/FreshRSS/pull/4263
* [ ] Documentation (help welcome)
  * Technical note: all XPath evaluations (i.e. XPath supposed to return some text and not the queries returning a collection) are automatically wrapped into a [`normalize-space()`](https://developer.mozilla.org/docs/Web/XPath/Functions/normalize-space) function to cast the result to string if not already done by the user and also remove superfluous white spaces.
* [x] Include information in OPML export / import? https://github.com/FreshRSS/FreshRSS/pull/4342
* [ ] Custom date format, e.g. with https://php.net/datetime.createfromformat
* [x] Make corresponding upstream PR, reverting https://github.com/simplepie/simplepie/commit/e49c578817aa504d8d05cd7f33857aeda9d41908#commitcomment-67585150 : https://github.com/simplepie/simplepie/pull/723

Summary of the tasks from comments below:
* [x] Implement support of `<base>`


Alkarex:
Example for `https://www.france.tv/france-2/les-petits-meurtres-d-agatha-christie/toutes-les-videos/`

image

image

Result:
image

app/Models/Feed.php
```diff
@@ -489,6 +542,87 @@ public function loadEntries(SimplePie $simplePie) {
 		}
 	}
 
+	/**
+	 * @param array<string,mixed> $attributes
+	 * @return SimplePie|null
+	 */
+	public function loadHtmlXpath(bool $loadDetails = false, bool $noCache = false, array $attributes = []) {
```
Alkarex, line 549: This function is the only true new functional code for this feature, the rest of the PR being mainly related improvements and UI code

app/views/index/rss.phtml
```diff
@@ -27,12 +36,23 @@ foreach ($this->entries as $item) {
...
 				}
+				$enclosures = $item->enclosures(false);
+				if (is_array($enclosures)) {
+					foreach ($enclosures as $enclosure) {
+						// https://www.rssboard.org/media-rss
+						echo "\t\t\t" , '<media:content url="' . $enclosure['url']
+							. (empty($enclosure['medium']) ? '' : '" medium="' . $enclosure['medium'])
+							. (empty($enclosure['type']) ? '' : '" type="' . $enclosure['type'])
+							. (empty($enclosure['length']) ? '' : '" length="' . $enclosure['length'])
+							. '"></media:content>', "\n";
```
Alkarex, line 49: https://github.com/FreshRSS/FreshRSS/issues/1796

lib/lib_rss.php
```diff
@@ -218,6 +218,7 @@ function customSimplePie($attributes = array()): SimplePie {
 	$simplePie->set_cache_name_function('sha1');
 	$simplePie->set_cache_location(CACHE_PATH);
 	$simplePie->set_cache_duration($limits['cache_duration']);
+	$simplePie->enable_order_by_date(false);
```
Alkarex, line 221: We sort ourselves. This created a bug randomising the order of entries when several have the same date.
Also improves performances slightly by avoiding a sort.

Alkarex:
Another example. Test page: https://alexandre.alapetite.fr/blog/actu.fr.html

Reading from a basic HTML structure
```html
<dl>
  <dt id="entry1">Entry 1 title</dt>
  <dd>Entry 1 content</dd>
  <dt id="entry2">Entry 2 title</dt>
  <dd>Entry 2 content</dd>
</dl>
```

image

app/Models/Feed.php
```diff
@@ -428,15 +478,18 @@ public function loadEntries(SimplePie $simplePie) {
...
 								. ($length == null ? '' : '" data-length="' . intval($length))
-								. '" data-type="' . htmlspecialchars($mime, ENT_COMPAT, 'UTF-8')
+								. ($mime == '' ? '' : '" data-type="' . htmlspecialchars($mime, ENT_COMPAT, 'UTF-8'))
 								. '" controls="controls"></video> <a download="" href="' . $elink . '">💾</a></p>';
 						} else {	//e.g. application, text, unknown
-							$enclosureContent .= '<p class="enclosure-content"><a download="" href="' . $elink . '">💾</a></p>';
+							$enclosureContent .= '<p class="enclosure-content"><a download="" href="' . $elink
+								. ($mime == '' ? '' : '" data-type="' . htmlspecialchars($mime, ENT_COMPAT, 'UTF-8'))
+								. ($medium == '' ? '' : '" data-medium="' . htmlspecialchars($medium, ENT_COMPAT, 'UTF-8'))
+								. '">💾</a></p>';
```
Frenzie, line 492: Not related to the PR but this caught my eye due to being in the diff. It'd probably make sense to stick in an aria-label="Save" or some such. Cf. https://github.com/FreshRSS/FreshRSS/issues/4190

Alkarex: There is already the `download` attribute, which can be used by assistive tools to know the intent. Maybe a `title` though, to help understanding the floppy disk icon for both screen and screen-reader users...

Alkarex:
Now with automatic form validation of XPath expressions
image

CE1212:
This is something I have been looking for...Great

Alkarex:
> This is something I have been looking for...Great

@CE1212 Tests welcome ;-)

CE1212:
So would it be possible to have it get the RSS then go to the site and scrape further information, say we only get the headline and a link, so the Freshrss once ready could go to the page and get say the author link also once its on the site.

If it could do that and rotate proxies that would be amazing.

Can't wait for it to be ready...awesome stuff

Alkarex:
@CE1212 This feature is an alternative to an RSS / Atom feed, not a supplement. To supplement an existing RSS / Atom feed, FreshRSS already has a full-content retrieval mechanism. I am not sure what you mean with proxy rotation, but it is likely not related.

CE1212:
@Alkarex 
Ah I just totally misunderstood, the proxy rotation I meant having the script cycle through a series of ip's in case you are accessing the same site really often.

I don't seem to have the type of feed source option on my install

I'm running
Your current version of FreshRSS is 1.20.0-dev.

unless I am just not seeing it probably me doing something wrong but any help would be appreciated 

I am happy to test as I know enough just to break things lol, thanks for your replies

Edit-
I had the wrong version installed, I do only know enough to break things.....thanks for your hard work

math-GH:
I play a little bit around with this PR.

Here some findings:
found this warning in my log
grafik

Git does not ignore the cached HTML file in data/cache
grafik


What does the orange mean?
grafik
grafik

math-GH:
Could you please try, whether you are able to generate a feed from this page?

https://contao.org/en/news.html?year=2021

my settings:
grafik

I am not sure if XPath has issues with HTML5 and its not closed tags like f.e. `<meta` or `<link>`.
I tried XPath on https://www.freeformatter.com/xpath-tester.html and it gives error messages because of missed `</meta>` and `</link>`

math-GH:
the XPath for feed title is used only once: when the feed is added new to the system.

When the feed config is changed it does not have any affect on the title anymore
grafik

Frenzie:
For me that'd very much be the desired behavior; I wouldn't want it randomly changing the name into something I didn't put in there.

math-GH:
> For me that'd very much be the desired behavior; I wouldn't want it randomly changing the name into something I didn't put in there.

So it would make sense to hide the config textarea if it has no affect anymore.

Frenzie:
> I am not sure if XPath has issues with HTML5 and its not closed tags like f.e. <meta or <link>.

The data should be fed through https://www.php.net/manual/en/domdocument.loadhtml.php

For a tool like you linked you'd have to do that yourself, i.e., change `<meta>` into a self-closing `<meta/>` for example.

Frenzie:
Anyway, here's why `querySelectorAll()`'s sometimes nicer than XPath. I figure something like this is what we want.
```
//div[contains(concat(' ', @class, ' '), 'odd') or contains(concat(' ', @class, ' '), 'even')]
```
[Edit: sorry, overlooked that I'd need to stick in even more spaces, something like shown in <https://stackoverflow.com/a/1390680>]

Screenshot_2022-02-23_22-41-24

Screenshot_2022-02-23_22-40-57

math-GH:
The link to strtotime() could be more useful with a `target="_blank"` so that the input is not lost while editing and loading a new page in the same tab.
grafik

math-GH:
strtotime() is only useful for English pages. Maybe there could be a way to have customized dates

Maybe https://www.php.net/manual/en/datetime.createfromformat.php could be an idea

math-GH:
grafik
It "eats" `&amp;cHash`

HTML on the page:
grafik

but the article link is (see right hand side of `5831`:
`....../?tx_ttnews%5Btt_news%5D=5831=863ee361abe89b13ebcef16ee0b3024a`

math-GH:
The creation of links works not best.

"Feed Url": `https://domain.tld/sub/`
but: `<base href="https://domain.tld" />`
links on the page: `/sub/newsarticle-1234.html`

The generated link: `https://domain.tld/sub/sub/newsarticle-1234.html`.
Expected link: `https://domain.tld/sub/newsarticle-1234.html`

This functionality needs to search and follow the `<base>`

Frenzie:
Is that from the source or from the browser? That is, the XPath output would have performed an `&amp;` → `&` transformation, but logically then you'd still have `&cHash` remaining.

Edit: of course there would heed to be an htmlspecialchars() applied to prevent anything weird from happening in the browser, not talking about that part. ^_^

app/Models/Feed.php
```diff
@@ -489,6 +542,95 @@ public function loadEntries(SimplePie $simplePie) {
...
+	 * @return SimplePie|null
+	 */
+	public function loadHtmlXpath(bool $loadDetails = false, bool $noCache = false, array $attributes = []) {
+		if ($this->url == '') {
+			return null;
+		}
+		$feedSourceUrl = htmlspecialchars_decode($this->url, ENT_QUOTES);
+		if ($this->httpAuth != '') {
+			$feedSourceUrl = preg_replace('#((.+)://)(.+)#', '${1}' . $this->httpAuth . '@${3}', $feedSourceUrl);
+		}
```
Frenzie, line 556: Cf. discussion in https://github.com/FreshRSS/FreshRSS/pull/4220#issuecomment-1050289052

I don't have time to check right now but there might be something double going on? I think the XPath output is already something akin to `htmlspecialchars_decode()` out of the box, unless I'm mistaken.

Frenzie: @math-GH Btw, what's a sample page? I don't really want to create a testcase myself. ;-)

math-GH:
From source: `wasserball/aktuelles-wasserball/lesen/?tx_ttnews%5Btt_news%5D=5831&cHash=863ee361abe89b13ebcef16ee0b3024a`

math-GH:
I do not want to publish and archive concrete links, so I would use an etherpad to share the link and configs:
https://yopad.eu/p/xpath-365days

app/Models/Feed.php
math-GH: see my next comment

math-GH:
If someone needs an XPath sandbox to find the correct XPath rules: http://xpather.com/ (maybe we could add it later into the documentation)

Frenzie:
Cheers, that would've been nice to have 20 years ago. (It seems to date back to 2018; I actually haven't used XPath/XSLT since 2015 iirc.)

Alkarex:
I am not fully back yet, so just a few quick things:

> found this warning in my log

Normal. A new SQL column is automatically added on demand.

> Git does not ignore the cached HTML file in data/cache

Done: https://github.com/FreshRSS/FreshRSS/pull/4220/commits/839270835b2457fc1f705d7b957f0062dcf2b283

> What does the orange mean?

Orange is not related to this PR but is probably when your feed is empty.

> The link to strtotime() could be more useful with a `target="_blank"` so that the input is not lost while editing and loading a new page in the same tab.

Done: https://github.com/FreshRSS/FreshRSS/pull/4220/commits/aa7901f4e119389b705b428607811249f3119774

Alkarex:
> strtotime() is only useful for English pages. Maybe there could be a way to have customized dates
> 
> Maybe https://www.php.net/manual/en/datetime.createfromformat.php could be an idea

Good idea with custom format, but let's keep that for another PR. `strtotime()` is not only for English dates, but also for e.g. ISO and other formats using only numbers

Frenzie:
As an aside, many pages use a proper `<time>` attribute, so you should use the `datetime` attribute while ignoring the text content. E.g.,

```
<time datetime="2022-02-19T11:34:23+01:00">February 19, 2022 at 11:34</time>
```

Alkarex:
> Could you please try, whether you are able to generate a feed from this page? https://contao.org/en/news.html?year=2021

A simpler version based on @Frenzie 's version:

image

Alkarex:
> Anyway, here's why `querySelectorAll()`'s sometimes nicer than XPath.

@Frenzie , yes, CSS selectors are easier (but less expressive) than XPath. I believe we should have both. I started with XPath because there is native PHP support (of XPath 1.0) and because I believe our future "HTML + CSS Selector" scraping method could behind the scenes perform a CSS Selector to XPath translation (I believe there are some light libraries doing that - findings welcome). That could also be used to remove our `lib_phpQuery.php`

Alkarex:
> It "eats" `&amp;cHash`

@math-GH Indeed. Should be fixed by https://github.com/FreshRSS/FreshRSS/pull/4220/commits/998babbe94755c832a226c79527abbd5b466bbe3

app/Models/Feed.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/4220/commits/998babbe94755c832a226c79527abbd5b466bbe3
The rest looks good

Alkarex:
> This functionality needs to search and follow the `<base>`

@math-GH Should be fixed by https://github.com/FreshRSS/FreshRSS/pull/4220/commits/f3f92aa1889950d543e4c0ac1ecd5e4c59aa37df . Tests welcome

Related upstream PR: https://github.com/simplepie/simplepie/pull/723

math-GH:
I love it. It works very fine now.

Alkarex:
As all seems fine, let's merge and test a bit more

lib/lib_phpQuery.php
```diff
@@ -436,7 +436,8 @@ protected function isXHTML($markup = null) {
 	}
 	protected function isXML($markup) {
 //		return strpos($markup, '<?xml') !== false && stripos($markup, 'xhtml') === false;
-		return strpos(substr($markup, 0, 100), '<'.'?xml') !== false;
+		$head = substr($markup, 0, 100);
+		return strpos($head, '<'.'?xml') !== false && stripos($head, '<html ') === false;
```
Alkarex, line 440: https://github.com/FreshRSS/FreshRSS/pull/4243

Alkarex:
> @Frenzie , yes, CSS selectors are easier (but less expressive) than XPath. I believe we should have both. I started with XPath because there is native PHP support (of XPath 1.0) and because I believe our future "HTML + CSS Selector" scraping method could behind the scenes perform a CSS Selector to XPath translation (I believe there are some light libraries doing that - findings welcome). That could also be used to remove our `lib_phpQuery.php`

https://github.com/FreshRSS/FreshRSS/pull/4261

InterferencePattern:
When will this feature be included in a release?

Alkarex:
> When will this feature be included in a release?

@jimbudarz You are welcome to already use our edge branch, which is a relatively stable rolling release, available by git, Docker, ZIP. Feedback welcome, and we need more testers. Next stable version 1.20.0 is coming in I guess ~2-3 weeks.

math-GH:
I can confirm that the edge (V1.20.0-dev) works very well. I use it on my prod instance. The Web scraping works great there.

Alkarex:
Nice article on the subject: https://danq.me/2022/09/27/freshrss-xpath/

_Originally posted by @marienfressinaud in https://github.com/FreshRSS/FreshRSS/discussions/4647#discussioncomment-3777171_

math-GH:
More amazing feedback: https://forum.cloudron.io/topic/7651/freshrss-1-2-0-released-with-killer-feature-track-any-website

app/Models/Feed.php
```diff
@@ -489,6 +542,97 @@ public function loadEntries(SimplePie $simplePie) {
...
+		$xPathItemUri = $xPathSettings['itemUri'] ?? '';
+		$xPathItemAuthor = $xPathSettings['itemAuthor'] ?? '';
+		$xPathItemTimestamp = $xPathSettings['itemTimestamp'] ?? '';
+		$xPathItemThumbnail = $xPathSettings['itemThumbnail'] ?? '';
+		$xPathItemCategories = $xPathSettings['itemCategories'] ?? '';
+		if ($xPathItem == '') {
+			return null;
+		}
+
+		$html = getHtml($feedSourceUrl, $attributes);
```
Alkarex, line 575: Bug fix `$attributes` fixed in https://github.com/FreshRSS/FreshRSS/pull/4759

lib/lib_rss.php
```diff
@@ -311,6 +315,127 @@ function sanitizeHTML($data, $base = '', $maxLength = false) {
...
+		Minz_Log::warning('Error fetching content: HTTP code ' . $c_status . ': ' . $c_error . ' ' . $url);
+	}
+	if ($html == false) {
+		$html = '';
+	} else {
+		$html = enforceHttpEncoding($html, $c_content_type);
+	}
+
+	if (file_put_contents($cachePath, $html) === false) {
+		Minz_Log::warning("Error saving cache $cachePath for $url");
```
Inverle, line 433: https://github.com/FreshRSS/FreshRSS/pull/9206

PR #4238, opened 2022-02-26
Improved: Show URL button

math-GH:
The show the website button is improved now:
grafik

- before: the button has a static URL. When user changed the URL in the text input the button opened the "old" URL. It needs a page reload to have the new URL available with the button
- after: the button opens the shown URL from the text input
- new: mouse title "show URL" to make it more understandable what is behind the button

2. manage a feed
3. change the URL of feed or website
4. click the button
5. the changed URL opens in the new tab

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


p/scripts/extra.js
```diff
@@ -123,6 +123,29 @@ function init_password_observers() {
...
+		this.href = url;
+		return true;
+	} else {
+		urlField.focus();
+		this.removeAttribute('href');
+		return false;
+	}
+}
+
+// set event listener on "show url" buttons
```
Frenzie, line 141: This seems a bit complex, although presumably it would be possible to replace all of them with `focus` or `click`. (If not, I suspect you need focus in addition to the rest.)

It seems more intuitive to use the `input` event on the text/url input.

app/i18n/en/gen.php
```diff
@@ -32,6 +32,7 @@
 		'purge' => 'Purge',
 		'remove' => 'Remove',
 		'rename' => 'Rename',
+		'see_url' => 'See URL',	// TODO
```
Frenzie: `View URL` would be the correct form of this, but I think `Open URL` would be better.

```suggestion
		'see_url' => 'Open URL',
```

math-GH: should I rename the key too?

p/scripts/extra.js
```diff
@@ -123,6 +123,29 @@ function init_password_observers() {
...
 
+// overwrites the href attribute from the url input
+function updateHref() {
+	const urlField = document.getElementById(this.getAttribute('data-input'));
+	const url = urlField.value;
+	if (url.length > 0) {
+		this.href = url;
+		return true;
+	} else {
+		urlField.focus();
```
Frenzie, line 134: Btw, this line would yield rather unexpected results in combination with mouseover (and even mousedown/keydown). It only goes with click. :-)

Incidentally, I suspect with the return false here you mean `preventDefault()`?
```suggestion
		e.preventDefault();
		urlField.focus();
```

p/scripts/extra.js
```diff
@@ -123,6 +123,29 @@ function init_password_observers() {
 	});
 }
 
+// overwrites the href attribute from the url input
+function updateHref() {
```
Frenzie:
```suggestion
function updateHref(e) {
```

app/i18n/en/gen.php
Frenzie: At this stage definitely. :+1:

p/scripts/extra.js
math-GH: mousedown and keydown would be enough in general. The mouseover is used to update the link preview.

(see browser corner)
grafik

I do not understand why it would be better to use the input event

p/scripts/extra.js
math-GH: I tried it too, but e.preventDefault() did not work. That is why the `href `will be deleted.

p/scripts/extra.js
Frenzie: focus is made for the purpose, mousedown and keydown is not. It's a more complex way of doing more or less the same thing hopefully identically and potentially slightly worse. mousedown and keydown should not be used in this way.

> I do not understand why it would be better to use the input event

More correctly stated, the blur, change or input event.

You have three different behaviors:

* mouseover to update preview, but this introduces a lot of insecurities, like what does tap & hold, copy link on a mobile browser do in terms of events, how buggy are they, could we get a flash of outdated URL
* focus to update right before the actual click action
* click to check if the URL is empty and focus the input field

But why bother with the hours of time spent testing assistive technologies and mobile browsers when you can just update the DOM on input?

Additionally if you don't do it on input it won't work as intended with the links panel or with link selectors, or bookmarklets such as [these](https://www.squarefree.com/bookmarklets/pagelinks.html).

p/scripts/extra.js
Frenzie: That's because this should trigger on `click`.

p/scripts/extra.js
```diff
@@ -123,6 +123,29 @@ function init_password_observers() {
...
+		urlField.focus();
+		this.removeAttribute('href');
+		ev.preventDefault();
+		return false;
+	}
+}
+
+// set event listener on "show url" buttons
+function init_url_observers() {
+	document.querySelectorAll('.open-url').forEach(function (btn) {
```
Alkarex, line 143: Not super important because it is not our main page, but a loop to add event listeners is a bad sign. It is lighter to add a single event listener and use "event delegation" https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#event_delegation

Several examples in e.g. this function:
https://github.com/FreshRSS/FreshRSS/blob/da2adaec8a4ea69bf837c7166398568e34b6d066/p/scripts/main.js#L777-L779

Frenzie: It's also not uncommon to just set it all on document (with a similar event.target/event.target.matches/event.target.closest pattern). The overhead from a bit of if/else or switch or whatever is significantly less than adding an eventlistener on a gazillion elements,particularly in memory. But yeah, I doubt it matters here.

math-GH: it is the same like the show password buttons.

app/i18n/en/gen.php
math-GH: done

p/scripts/extra.js
math-GH: I changed to click event.

app/i18n/fr/gen.php
```diff
@@ -28,6 +28,7 @@
 		'load_default_shortcuts' => 'Utiliser les raccourcis par défaut',
 		'manage' => 'Gérer',
 		'mark_read' => 'Marquer comme lu',
+		'open_url' => 'Open URL',	// TODO
```
```suggestion
		'open_url' => 'Ouvrir l’URL',
```

PR #4258, opened 2022-03-04
Improved: Feed config: title, website url, feed url

math-GH:
## empty website URL

If the website URL input is empty:
grafik

Before: 
grafik

After:
grafik

## title + feed URL required
After:
grafik

Before: the fields were not required

- Javascript: do not show "See website"
- feed update template: required fileds + input type=url

1. Go to subscription management and edit a feed
2. delete content of title, website URL, Feed URL

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/views/helpers/feed/update.phtml
```diff
@@ -32,7 +32,7 @@
 		<div class="form-group">
 			<label class="group-name" for="name"><?= _t('sub.feed.title') ?></label>
 			<div class="group-controls">
-				<input type="text" name="name" id="name" class="extend" value="<?= $this->feed->name(true) ?>" />
+				<input type="text" name="name" id="name" class="extend" value="<?= $this->feed->name(true) ?>" required />
```
Alkarex:
```suggestion
				<input type="text" name="name" id="name" class="extend" value="<?= $this->feed->name(true) ?>" required="required" />
```

app/views/helpers/feed/update.phtml
```diff
@@ -54,7 +54,7 @@
 			<label class="group-name" for="url"><?= _t('sub.feed.url') ?></label>
 			<div class="group-controls">
 				<div class="stick">
-					<input type="text" name="url" id="url" class="extend" value="<?= $this->feed->url() ?>" />
+					<input type="url" name="url" id="url" class="extend" value="<?= $this->feed->url() ?>" required />
```
Alkarex:
```suggestion
					<input type="url" name="url" id="url" class="extend" value="<?= $this->feed->url() ?>" required="required" />
```

math-GH: Not needed for HTML5. Do we follow XHTML? I would like it too

Alkarex: We have been using the strict syntax as much as possible, kind of XHTML, yes

app/layout/aside_feed.phtml
```diff
@@ -138,7 +138,7 @@
 		<?php if (FreshRSS_Auth::hasAccess()) { ?>
 		<li class="item"><a href="<?= _url('stats', 'repartition', 'id', '------') ?>"><?= _t('index.menu.stats') ?></a></li>
 		<?php } ?>
-		<li class="item"><a target="_blank" rel="noreferrer" href="http://example.net/"><?= _t('gen.action.see_website') ?></a></li>
+		<li class="item" id="websiteURL"><a target="_blank" rel="noreferrer" href="http://example.net/"><?= _t('gen.action.see_website') ?></a></li>
```
Alkarex: Not good, as we end up with multiple elements with the same ID in the DOM. Could be changed to a class instead, but see my alternative proposition below.

Alkarex:
Instead of removing the Website link, we could consider falling back on feed URL when the Website is empty. Easier logic and I would find it more consistent:

https://github.com/FreshRSS/FreshRSS/blob/926ce8ff5f2595bdf29f6a8fb61e8e40adb1251a/app/layout/aside_feed.phtml#L100

math-GH:
> Instead of removing the Website link, we could consider falling back on feed URL when the Website is empty. Easier logic and I would find it more consistent:
> 

I am not happy with this idea (or I do not understand it correctly).

On my device (Firefox on Windows), an RSS link would open an download prompt. I would be irritated, when I clicked on "Website" and would expect an opened website of the feed.

math-GH:
Improved. Now it is changed from ID to class

p/scripts/main.js
```diff
@@ -846,6 +846,9 @@ function init_column_categories() {
 				a.href = '#dropdown-' + id;
 				div.querySelector('.dropdown-target').id = 'dropdown-' + id;
 				div.insertAdjacentHTML('beforeend', template);
+				if (feed_web.length < 1) {
```
Alkarex, line 849: What was the reason for having this logic client side in JavaScript?
Could be put server side in PHP in `app/layout/aside_feed.phtml`, don't you think?

Alkarex, line 849: Sorry, wrong comment. It is in a template, so all good

PR #4261, opened 2022-03-05
Replace lib_phpQuery by PhpGt/CssXPath

Alkarex:
https://github.com/PhpGt/CssXPath
Much lighter and cleaner


app/Models/Entry.php
```diff
@@ -480,40 +480,37 @@ public function isDay(int $day, int $today): bool {
...
+			$doc->loadHTML($html, LIBXML_NONET | LIBXML_NOERROR | LIBXML_NOWARNING);
+			$xpath = new DOMXPath($doc);
 
 			if ($maxRedirs > 0) {
 				//Follow any HTML redirection
-				/**
-				 * @var phpQueryObject @metas
-				 */
-				$metas = $doc->find('meta[http-equiv][content]');
+				$metas = $xpath->query('//meta[@content]');
```
Frenzie, line 489: Does this deal with random whitespace that doesn't affect HTML?

Alkarex: Yes

Alkarex:
Ping @marienfressinaud and @aledeg : if you use full content retrieval for some of your feeds, it would be nice if you could give this a try

Alkarex:
@javerous You did a a great work regarding full content retrieval https://github.com/FreshRSS/FreshRSS/pull/2778 ; so if you are around, it would be nice if you could give this PR a try, as there are risks of regressions :-)
And if you have not seen it, you might be interested in this slightly related feature, which freshly landed in the edge branch: https://github.com/FreshRSS/FreshRSS/pull/4220

math-GH:
First glance it looks fine. Did not checked the code.

phpstan.neon
```diff
@@ -7,7 +7,6 @@ parameters:
 	paths:
 		- .
 	excludePaths:
-		- lib/lib_phpQuery.php
```
math-GH, line 10: should CssXPath be excluded from phpstan scan?

Frenzie: Exclusion doesn't seem necessary if it's not causing issues, but potentially fair point from a handful of wasted cpu cycles.

Alkarex: For now, I would rather keep it (phpstan uses some caching) to be informed in case of regression

marienfressinaud:
> Ping @marienfressinaud and @aledeg : if you use full content retrieval for some of your feeds, it would be nice if you could give this a try

For the few I had, it seems to work well :+1: (I prefer your approach rather than previous lib_phpQuery btw :))

Alkarex:
> For the few I had, it seems to work well 👍 (I prefer your approach rather than previous lib_phpQuery btw :))

Cool :-) Let's merge then

javerous:
> @javerous You did a a great work regarding full content retrieval #2778 ; so if you are around, it would be nice if you could give this PR a try, as there are risks of regressions :-)
> And if you have not seen it, you might be interested in this slightly related feature, which freshly landed in the edge branch: #4220

Hi @Alkarex - sorry for the late reply.

Thank you for your comment. I tested it a bit (even if the feature has been merged since a long time now…), I didn't notice any issues with the selectors I'm using !

The "HTML+XPath (Web scraping)" feature seems really cool ! I tested it a bit, it's very nice 👏 (and well, it may also deserve some way to validate / preview what was entered, in the future - it took me some time to have what I wanted 😅)

Alkarex:
Thanks for the feedback @javerous :-)

PR #4266, opened 2022-03-07
Documentation: added information about code checks

math-GH:
some more information added to the manual about code testing/styling


Alkarex:
Maybe a note somewhere that tests can be run locally, e.g. by running `make test-all` (require npm and node.js installed), and several problems can be automatically fixed by running `make fix-all`

PR #4271, opened 2022-03-14
New sharing service: Web sharing API

math-GH:

- added: https://www.w3.org/TR/web-share/ and https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share use the default sharing option of the operating system
- added: `HTMLtag` option for the sharing services (optional) to allow `button` as list item type that is not a link

1. add `Default Sharing Service` in Sharing option
2. use it in the normal view within the share menu
3. share the link to your favorite app

Important information: The menu item in share menu is not available, if you browser does not support it (f.e. Firefox on Windows)

grafik
grafik

Example on Chromium on Windows 10: (1) the title of the shared article + (2) the shared link
grafik

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


app/i18n/en/gen.php
```diff
@@ -203,17 +203,18 @@
...
 		'raindrop' => 'Raindrop.io',
-		'reddit' => 'Reddit',	// TODO
+		'reddit' => 'Reddit',
 		'shaarli' => 'Shaarli',
 		'twitter' => 'Twitter',
 		'wallabag' => 'wallabag v1',
 		'wallabagv2' => 'wallabag v2',
-		'whatsapp' => 'Whatsapp',	// TODO
-		'xing' => 'Xing',	// TODO
+		'web-sharing-api' => 'Default System Sharing',
```
Frenzie:
```suggestion
		'web-sharing-api' => 'Default system sharing',
```


But maybe something like?
```suggestion
		'web-sharing-api' => 'Share through system',
```

p/scripts/main.js
```diff
@@ -1068,6 +1078,14 @@ function init_stream(stream) {
 		}
 	};
 
+	if (!navigator.share) {	// https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
+		document.querySelectorAll('.item.share > button[data-type="web-sharing-api"]').forEach(
+			function (item) {
+				item.style.display = 'none';
```
Frenzie: I'd reverse the direction.

math-GH: I do not understand what you mean

Frenzie: if navigator.share etc.

math-GH: most common browsers support navigator.share, except Firefox
grafik

Frenzie: I'm not sure what you think I said but I don't see how that's relevant to my remark about code architecture and maintainability. :-)

Frenzie: As an aside, my Firefox definitely seems to support it.

math-GH: Still I do not understand your `I'd reverse the direction.`. What would you do? Could you support with an example?

> As an aside, my Firefox definitely seems to support it.
Which Firefox Version on which operating system?

Frenzie: 98 on Linux.

> Still I do not understand your I'd reverse the direction.. What would you do? Could you support with an example?

Identical in effect, but if navigator.share etc. instead of if !navigator.share etc. That way you have the standard feature detection pattern in line with expectations of future readers.

app/i18n/en/gen.php
math-GH: `Share through system` does not make me more happier.

p/scripts/main.js
math-GH: I don't get the point what is wrong to hide the menu item for a minor set of browsers?
Why should the `if` check for the majority (navigator.share)? The menu items are generated in `integration.phtml` It is much more easier to list all sharing item and then hide the 1 item, that does not work, than do not list the item and add it via javascript.

Frenzie: I'm still not sure what you think I said; reversing the logic shouldn't make anything any harder or easier.

But now that you mention it, you could simplify it along these lines, for performance reasons presumably still preferably with a class name, but in this sample with an attribute selector:

```css
button[data-type="web-sharing-api"] {display:none}
```

```js
if (navigator.share) {
 document.styleSheets[0].insertRule('button[data-type="web-sharing-api"] {display:initial}', document.styleSheets[0].cssRules.length)
}
```

Or the other way around, but as I said I think that's slightly worse for future ease of reading.
```js
if (!navigator.share) {
 document.styleSheets[0].insertRule('button[data-type="web-sharing-api"] {display:none}', document.styleSheets[0].cssRules.length)
}
```

Frenzie: PS Also for the ease of future reading, I actually mean that as something more along these lines in the final result, not as my quick one-liner from the console. ;-)
```js
if (navigator.share) {
  document.styleSheets[0].insertRule(`
    button[data-type="web-sharing-api"] {
      display: initial;
    }`,
    document.styleSheets[0].cssRules.length)
}

Comment (math-GH) [reply p/scripts/main.js]: Thanks. It helped me a lot.

Comment (Frenzie) [reply p/scripts/main.js]: Fun fact, dom.webshare.enabled is set to **false** in my Firefox, yet it works. Weird.

**Review by Frenzie — APPROVED** [review]
Agree to disagree, I suppose. ;-)

File: app/i18n/en/gen.php (hunk 1 above)
Comment (math-GH) [reply app/i18n/en/gen.php]: Maybe @Alkarex has a nice idea

File: app/i18n/en/gen.php [hunk 3]
```diff
@@ -198,17 +198,18 @@
...
 		'raindrop' => 'Raindrop.io',
-		'reddit' => 'Reddit',	// TODO
+		'reddit' => 'Reddit',
 		'shaarli' => 'Shaarli',
 		'twitter' => 'Twitter',
 		'wallabag' => 'wallabag v1',
 		'wallabagv2' => 'wallabag v2',
-		'whatsapp' => 'Whatsapp',	// TODO
-		'xing' => 'Xing',	// TODO
+		'web-sharing-api' => 'Default system sharing',
```
Frenzie: Another possibility could be something like:
```suggestion
		'web-sharing-api' => 'Built-in sharing',
```

app/i18n/fr/gen.php
```diff
@@ -198,17 +198,18 @@
...
 		'raindrop' => 'Raindrop.io',	// IGNORE
-		'reddit' => 'Reddit',	// TODO
+		'reddit' => 'Reddit',	// IGNORE
 		'shaarli' => 'Shaarli',	// IGNORE
 		'twitter' => 'Twitter',	// IGNORE
 		'wallabag' => 'wallabag v1',	// IGNORE
 		'wallabagv2' => 'wallabag v2',	// IGNORE
-		'whatsapp' => 'Whatsapp',	// TODO
-		'xing' => 'Xing',	// TODO
+		'web-sharing-api' => 'Default system sharing',	// TODO
```
```suggestion
		'web-sharing-api' => 'Partage standard',
```

app/i18n/en/gen.php
Alkarex: I cannot find anything much better. I have proposed "Standard share" for the French version

p/scripts/main.js
```diff
@@ -1068,6 +1078,15 @@ function init_stream(stream) {
 		}
 	};
 
+	if (!navigator.share) {
+		// https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
+		// do not show the menu entry if browser does not support navigator.share
+		document.styleSheets[0].insertRule(
```
Alkarex, line 1084: Ah, I overlooked that this might not work until enough of the DOM is loaded, otherwise errors sometimes happen depending on the exact orders
image
I believe this code should be moved to a function called by `init_afterDOM()`.

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/4321

PR #4275, opened 2022-03-18
Set feed error state when XPath does not match

Alkarex:
Gives the ability to filter HTML+XPath feeds that are not working.
Simple implementation to start with. Could be made smarter with various error codes in the future.


Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/pull/4220

PR #4277, opened 2022-03-18
Fix some search expressions

Alkarex:
* Fix decoding bug when using quotes to search free text such as `"ab cd"`
* Fix use of `-` wrongly triggering a negative search in e.g. `ab-cd`


Alkarex:
Merging those fixes as I need them for a demo :-)

PR #4286, opened 2022-03-22
WebSub explicit hub.lease_seconds

Alkarex:
https://www.w3.org/TR/websub/#subscriber-sends-subscription-request
We ask for a lease of 10 days just like the default for
https://github.com/flusio/Webubbub/blob/6007039eb7bb29a1a6af7f6546a8c8995c2475a5/src/models/Subscription.php#L48

Before, WebSub hubs such as https://github.com/flusio/Webubbub would fall back to 1 day if we do not specify anything, which is a bit too short for FreshRSS logic.

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -32,6 +32,9 @@ class FreshRSS_Feed extends Minz_Model {
 	const ARCHIVING_RETENTION_COUNT_LIMIT = 10000;
 	const ARCHIVING_RETENTION_PERIOD = 'P3M';
 
+	/** Number of seconds for which we would like to have the WebSub subscription active */
+	const WEBSUB_LEASE_SECONDS = 864000;
+
 	/** @var int */
 	private $id = 0;
 	/** @var string */
@@ -948,6 +951,7 @@ class FreshRSS_Feed extends Minz_Model {
 						'hub.mode' => $state ? 'subscribe' : 'unsubscribe',
 						'hub.topic' => $url,
 						'hub.callback' => $callbackUrl,
+						'hub.lease_seconds' => self::WEBSUB_LEASE_SECONDS,
 						)),
 					CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 					CURLOPT_MAXREDIRS => 10,


Alkarex:
@marienfressinaud With this patch, I receive the following error from your WebSub:

```
400 `pending_lease_seconds` property must be an integer
```

It seems to be an internal bug in the WebSub server, as I am only providing `hub.lease_seconds`, not `pending_lease_seconds`.

Example of request POST content:

```php
http_build_query([
        'hub.verify' => 'sync',
        'hub.mode' => 'subscribe',
        'hub.topic' => 'https://marienfressinaud.fr/feeds/all.atom.xml',
        'hub.callback' => 'https://freshrss.example.net/api/pshb.php?k=xxx',
        'hub.lease_seconds' => 864000,
]);
```

which gives:
```
hub.verify=sync&hub.mode=subscribe&hub.topic=https%3A%2F%2Fmarienfressinaud.fr%2Ffeeds%2Fall.atom.xml&hub.callback=https%3A%2F%2Ffreshrss.example.net%2Fapi%2Fpshb.php%3Fk%3Dxxx&hub.lease_seconds=864000
```

Alkarex:
I actually think it might be better to leave the choice of the lease to the WebSub server (so close this PR).
For the record, here are some examples of values:
* Lease seconds: 315360000: https://blog.ouseful.info/feed/
* Lease seconds: 432000: http://feeds.exploringbinary.com/ExploringBinary
* Lease seconds: 432000: http://feeds.feedburner.com/Techcrunch
* Lease seconds: 432000: https://fransdejonge.com/feed/
* Lease seconds: 432000: https://sspai.typlog.io/feed/audio.xml
* Lease seconds: 432000: https://www.blogger.com/feeds/21402852/posts/default?alt=atom
* Lease seconds: 86400: https://marienfressinaud.fr/feeds/all.atom.xml

marienfressinaud:
Thanks for the feedback @Alkarex :) It should be fixed, and the default value for lease seconds as well (cf. https://github.com/flusio/Webubbub/commit/d81f2940115a3701b1b9787264984728690f639d)

PR #4292, opened 2022-03-26
Improved: Slider: Close area = grey background

math-GH:
When a slider is open, the left area is clickable and closes the slider.

After:
grey transparent to give the focus to the open slider and mark the visible items as "disabled" while the slider is open.
(similar to #4061)
grafik

- CSS

1. open config -> extensions
2. edit an extension
3. see the greyed area left of the slider (that is the new "feature")
4. click on it
5. it closes the slider

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


p/themes/base-theme/template.rtl.css
```diff
@@ -1418,13 +1418,28 @@ br {
...
+	top: 0.5rem;
+	right: 0.5rem;
+	filter: grayscale(100%) brightness(2.5);
+}
+
+#close-slider.active:hover img {
+	background: rgba(0, 0, 0, 0.2);
+	border-radius: 50%;
+}
+
```
Frenzie: This and the whole X strikes me as odd, making it both disconnected and as if you have to click on that X.

Frenzie (commented):
Btw, `#close-slider` and `#slider` would need a `z-index: 1` in this concept or they come up against `#nav_entries`.
Screenshot 2022-03-27 at 09-44-51 (37) Main stream · FreshRSS

p/themes/base-theme/template.rtl.css
math-GH: I changed it to the default behavior like the other close areas

math-GH: And fixed the CSS.

p/themes/base-theme/template.rtl.css
```diff
@@ -1418,13 +1419,25 @@ br {
 }
 
 #close-slider.active {
+	background: rgba(0, 0, 0, 0.2);
+	font-size: 0;
```
Frenzie, line 1464: I'm not sure what the intent here is. Presumably one of display: none, visibility: hidden or visibility: collapse would communicate the intent properly. ^_^

math-GH: The reason for `font-size: 0`:

`<a href="#" id="close-slider"<?= $class ?>>❌</a>`

Similar to: 
https://github.com/FreshRSS/FreshRSS/blob/edge/app/layout/aside_feed.phtml#L118

https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/base-theme/template.css#L1763-L1774

Frenzie (approved):

app/views/configure/queries.phtml
```diff
@@ -61,9 +61,7 @@
 	</form>
 
 	<?php $class = $this->displaySlider ? ' class="active"' : ''; ?>
-	<a href="#" id="close-slider"<?= $class ?>>
-		<?= _i('close') ?>
-	</a>
+	<a href="#" id="close-slider"<?= $class ?>>❌</a>
```
Alkarex: Was there a problem with the `_i('close')` approach?

math-GH: @Frenzie wrote: "This and the whole X strikes me as odd"

And now it is similar to https://github.com/FreshRSS/FreshRSS/blob/edge/app/layout/aside_feed.phtml#L118

PR #4293, opened 2022-03-27
Improved: dropdown close area

math-GH:

- some HTML improvements: every dropdown menu has the same "close button" structure like the mobile aside bar (f.e. in the normal view). It means: `.dropdown-close` is not anymore a part of the `.dropdown-menu`, but it is the next element.
- CSS is cleaned up
- visible new result: grey clickable area left of the mobile config menu
grafik

1. use FreshRSS on a small display ("mobile view")
2. open the config dropdown menu
3. see the grey area left of the menu
4. click on it
5. the menu closes

Plus: Check all dropdown menus:
normal view:
- user queries dropdown (top menu)
- mark as read dropdown (top menu)
- share dropdown (below the article)
- tags dropdown (below the article)
- labels dropdown (below the article)
- feed dropdown (left navigation)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -945,7 +945,7 @@ input[type="search"] {
 /*=== Feed article header and footer */
 .flux_header {
 	background: inherit;
-	position: relative;
+	position: static;
 }
 
 .flux .item {
--- a/p/themes/base-theme/template.css
+++ b/p/themes/base-theme/template.css
@@ -527,32 +527,18 @@ input[type="checkbox"]:focus-visible {
 	z-index: 1000;
 }
 
-.dropdown-close {
-	display: inline;
+.dropdown-menu + .dropdown-close {
+	display: none;
 }
 
-.dropdown-close a {
+.dropdown-target:target ~ .dropdown-close {
 	display: block;
 	font-size: 0;
 	position: fixed;
 	top: 0; bottom: 0;
 	left: 0; right: 0;
-	z-index: -11;
-	cursor: default;
-}
-
-.dropdown div.dropdown-close {
-	display: none;
-}
-
-.dropdown-target:target ~ div.dropdown-close {
-	display: block;
 	z-index: 999;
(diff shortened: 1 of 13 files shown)


math-GH:
It is the same greyed background like in #4292 and #4061

p/themes/base-theme/template.css
```diff
@@ -1769,7 +1756,7 @@ input:checked + .slide-container .properties {
 		bottom: 0;
 		left: 0;
 		right: 0;
-		cursor: pointer;
+		cursor: default;
```
Frenzie, line 1759: Is there somewhere else in the cascade that changes it? As written it looks like the line should simply be removed.

p/themes/base-theme/template.css
```diff
@@ -527,32 +527,18 @@ input[type="checkbox"]:focus-visible {
...
-
-.dropdown-target:target ~ div.dropdown-close {
-	display: block;
 	z-index: 999;
-	position: relative;
-}
-
-.dropdown-menu-scrollable .dropdown-close {
-	display: none;
+	cursor: default;
```
Frenzie, line 541: Same here.

p/themes/base-theme/template.css
math-GH: `.dropdown-close` is a `<a>` and the default cursor would be the "hand" icon.
`cursor: default` = maus pointer (arrow).
Maybe you mix it with `cursor: auto`?

grafik

Frenzie: No, I was just commenting on the diff.

math-GH: I checked. It is correct as it is

Alkarex: I did not follow the whole discussion. What is the status? :-)

Frenzie: All good probably.

PR #4297, opened 2022-03-29
Improved: slider

math-GH:
After:
grafik

1. Fix slider width (750px). (There is no benefit to have a wider slider.
2. Inputs/dropdowns/textareas could have a width of 50% and 100%. It improves the general layout and give the space for a better text input. The "extend" inputs are replaced with 100% widths inputs. No expanding & jumping inputs anymore.
3. slim scrollbar for slider content
4. deleted: the feed description. It is editable some centimeters below. It does not make sense to display redundant information.

In mobile view it looks also much better
grafik

- templates (.phtml)
- CSS

2. edit a feed
3. see the slider content

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/helpers/feed/update.phtml
+++ b/app/views/helpers/feed/update.phtml
@@ -8,8 +8,6 @@
 		<a href="<?= _url('stats', 'repartition', 'id', $this->feed->id()) ?>"><?= _i('stats') ?> <?= _t('sub.feed.stats') ?></a>
 	</div>
 
-	<p><?= sanitizeHTML($this->feed->description()) ?></p>
-
 	<?php $nbEntries = $this->feed->nbEntries(); ?>
 
 	<?php if ($this->feed->inError()) { ?>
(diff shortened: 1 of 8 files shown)


app/views/helpers/feed/update.phtml
```diff
@@ -8,8 +8,6 @@
 		<a href="<?= _url('stats', 'repartition', 'id', $this->feed->id()) ?>"><?= _i('stats') ?> <?= _t('sub.feed.stats') ?></a>
 	</div>
 
-	<p><?= sanitizeHTML($this->feed->description()) ?></p>
```
Alkarex, line 11: This was not supposed to be removed. https://github.com/FreshRSS/FreshRSS/issues/5320

math-GH: It was supposed to. See number 4:

> deleted: the feed description. It is editable some centimeters below. It does not make sense to display redundant information.

Frenzie: It's not redundant fwiw; the one is plain text input and the removed one is HTML.

math-GH: Here a screen shot of FreshRSS 1.19.0.
What is the benefit to see the description in (1), while it is written in (2) too?

grafik

Alkarex: Just for the record here as well: the (1) is in HTML, meaning that it can contain hyperlinks, images, etc.

PR #4310, opened 2022-04-11
Do not render irrelevant HTML for Ajax calls

### LINKED ISSUE #4309: Useless data transfer for loading articles
When ajax requests a new article, the article data is contained in a complete html file. For example: `https://domain.ltd/i/?a=normal&rid=xxx&next=xxx&state=3&ajax=1`

But this html will contain a lot of useless parts(html nodes) other than the actual data of the article. Especially when I subscribe to a lot of feeds, The `<ul id="sidebar" class="tree">` tree can get particularly large. While Gzip can compress this part to a few tens of KiB, users frequently swipe the screen to load new article, and these useless tens of KiB are repeatedly transferred.

And to generate these useless parts adds a little bit of performance consumption, I guess.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4309

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -47,8 +47,10 @@ if (_t('gen.dir') === 'rtl') {
 	</head>
 	<body class="<?= Minz_Request::actionName() ?>">
 <?php
-	flush();
-	$this->partial('header');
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
lgtm!

PR #4326, opened 2022-04-23
Update 08_sharing_services.md

math-GH:
Link #4150 in the docs


Frenzie (changes requested):
Nope, nope, nope! Definitely not. ;-)

I did forget about the minor problem mentioned there, but the proper solution for that would look something like this (emphasis on something like, I didn't investigate if this is the correct place to do it, just a 10-second search as to where it might be done):

```diff
diff --git a/app/Models/Share.php b/app/Models/Share.php
index 375b17f0..d88d18c7 100644
--- a/app/Models/Share.php
+++ b/app/Models/Share.php
@@ -228,7 +228,7 @@ class FreshRSS_Share {
                );
                $replaces = array(
                        $this->id(),
-                       $this->base_url,
+                       rtrim($this->base_url, '/'),
                        $this->title(),
                        $this->link(),
                );
```

But we should not link to a long highly irrelevant discussion. If felt to be necessary, just say something like "copy/paste the URL you see after logging in".

PR #4327, opened 2022-04-23
Fix: show password with touch

math-GH:
grafik

The issue was before:
The show password button did not work with touch events

- show password buttons works now with touch event
- had to change the behavior of the show password button. Before: show password while the buttons keep pressed. After: Click on button = show password. Click on the button again = hide the password. After 5 seconds the password will be hidden automatically

1. go to login page
2. type some password
3. click on the `show password` button
4. the password is shown
5. after 5 seconds the password will be hidden or click again on the `show password` button
6. the password text is hidden

Check it on "desktop" device (without touch, f.e. with a mouse) and on mobile device (with touch)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


p/scripts/extra.js
```diff
@@ -98,28 +98,40 @@ function init_crypto_form() {
...
-		passwordField.setAttribute('type', 'text');
-		this.classList.add('active');
+let timeoutHide;
+
+function showPW_this(ev) {
+	const id_passwordField = this.getAttribute('data-toggle');
+	if (this.classList.contains('active')) {
+		hidePW(id_passwordField);
+	} else {
+		if (ev.type == 'click' || ev.buttons || ev.key == ' ' || ev.key.toUpperCase() == 'ENTER') {
```
Alkarex:
```suggestion
		if (ev.type === 'click' || ev.buttons || ev.key === ' ' || ev.key.toUpperCase() === 'ENTER') {
```

(I know this line has just been moved, but let's address this tiny aspect at the same time)

PR #4330, opened 2022-04-25
Fix lazyimg single quotes

Alkarex:
Little bug for cases with single quote in URL such as

```html
<img src="123?format='jpg'" />
```

Could probably be replaced by `loading="lazy"` in the future, and with a better DOM-aware method, for instance during SimplePie `add_attributes()`


Alkarex:
Merging because we have a brand new user who is about to try FreshRSS and has a source, which requires this fix :-)

PR #4331, opened 2022-04-25
fix depending results after changing value of "entry->isRead"

cn-tools:

- if anybody use the hook 'entry_before_insert' and change the value of the property 'entry->isRead', the depending results are correct now.
- In my case my config let all new entrys as "isRead=FALSE" as i will. A new extension which is in progress, can/should/must change the value of this property, but the depending results of this property are wrong before this fix.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -484,8 +484,8 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 
 						if (!$entry->isRead()) {
 							$feed->incPendingUnread();
-							$nb_new_articles++;
 						}
+						$nb_new_articles++;
 					}
 				}
 				$entryDAO->updateLastSeen($feed->id(), $oldGuids, $mtime);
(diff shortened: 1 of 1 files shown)


app/Controllers/feedController.php
```diff
@@ -480,8 +481,11 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
 							$entryDAO->beginTransaction();
 						}
 						$entryDAO->addEntry($entry->toArray());
-						$feed->incPendingUnread();
-						$nb_new_articles++;
+
+						if (!$entry->isRead()) {
+							$feed->incPendingUnread();
+							$nb_new_articles++;
+						}
```
```suggestion
						}
						$nb_new_articles++;
```

I believe `$nb_new_articles++` needs to be out of the `if` condition, as we want to increase it for all new articles, even when new articles are automatically marked as read for one reason or another

Alkarex:
Thanks, looks reasonable (besides my little suggestion)

cn-tools:
Thanks! And i think you are right with your suggestion, so i merged it in.

PR #4347, opened 2022-05-05
Allow many (20k+) feeds

Alkarex:
Work in progress to allow scaling to 20k+ (testing with 50k+) feeds (one of my own use-cases for a little project at work).
Fix errors such as `SQL error listWhereRaw: ERROR: value "42926" is out of range for type smallint`
Start of UI and back-end performance improvements accordingly.

PR #4353, opened 2022-05-07
Option icons-as-emojis for performance

Alkarex:
New mode to use only emojis instead of icons.
Considerably improves performances.
Needed for the 20k+ feeds scenario https://github.com/FreshRSS/FreshRSS/pull/4347
The main problem was the ⚙️ configuration image in the list of feeds on the left (together with the existing option to hide favicons https://github.com/FreshRSS/FreshRSS/pull/2821 )

### DIFF
--- a/app/Models/Themes.php
+++ b/app/Models/Themes.php
@@ -70,52 +70,52 @@ class FreshRSS_Themes extends Minz_Model {
 
 	public static function alt($name) {
 		static $alts = array(
-			'add' => '✚',
+			'add' => '➕',	//✚
 			'all' => '☰',
-			'bookmark' => '★',
-			'bookmark-add' => '✚',
-			'bookmark-tag' => '🔖',
-			'category' => '☷',
+			'bookmark' => '✨',	//★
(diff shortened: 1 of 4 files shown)


Alkarex:
Here is how this mode looks like:
image

Frenzie (approved):
Cool. :-)

ParoTheParrot:
How can I use this option?
I didn't find this in the configuration menu.
And my attempt with a config-user.custo.phpon./data does not work.

I need advice, TIA!

Alkarex:
> How can I use this option? I didn't find this in the configuration menu. And my attempt with a config-user.custo.phpon./data does not work.

@ParoTheParrot Hello,
There is no UI for it at the moment, so you need to change the option in `./freshrss/data/users/alice/config.php`

`./data/config-user.custom.php` is only used as a template when creating a new user and will not modify the settings for existing users

ParoTheParrot:
good enough. Thanks for the quick reply!
Kepp up the good work with FreshRSS!

PR #4356, opened 2022-05-10
Refactor OPML export categories

Alkarex:
Simplify code to comply with types hints.
And renamed a property to plural.

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -6,7 +6,7 @@ class FreshRSS_Category extends Minz_Model {
 	 */
 	private $id = 0;
 	private $name;
-	private $nbFeed = -1;
+	private $nbFeeds = -1;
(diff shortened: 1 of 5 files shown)


app/Models/Category.php
```diff
@@ -56,18 +56,18 @@ public function feeds(): array {
...
 			$this->nbNotRead = 0;
 			foreach ($this->feeds as $feed) {
-				$this->nbFeed++;
+				$this->nbFeeds++;
 				$this->nbNotRead += $feed->nbNotRead();
 				$this->hasFeedsWithError |= $feed->inError();
 			}
-		}
 
-		usort($this->feeds, function ($a, $b) {
```
Alkarex, line 68: Moved in the cached section one line higher up instead of being recomputed every time

PR #3830, opened 2021-09-07
Headline structure improvement

math-GH:
Closes #3673

Changes proposed in this pull request:
all templates:
- `<h1>` around the FreshRSS logo: removed (because it is not the main headline)
- each page has a `<h1>` as a real headline
- lot of fake `<legend>` changed to mostly `<h2>` headlines
- now we have a clear headline structure on each page

How to test the feature manually:
see on each page the headlines

Pull request checklist:
- [x] clear commit messages
- [x] code manually tested

PR #4369, opened 2022-05-16
Custom logo HTML

Alkarex:
Add option for custom HTML logo/title in the main Web UI view.
Can potentially be different per user.
#fix https://github.com/FreshRSS/FreshRSS/pull/3830/files#r850472247


Alkarex:
@marienfressinaud Could that work for your own needs as well?

math-GH:
Is it a "hidden" feature that cannot configured via the config menu?

Alkarex:
> Is it a "hidden" feature that cannot configured via the config menu?

For now yes, but could be exposed in the HTML UI if there is a demand

math-GH:
I am not a big fan that it is tied to the user's settings and not to the system's settings. I would say, that in very rare cases the user has the access to the config files, mostly the user is admin.
I would not expect, that I - as a user - could edit my config file. As a admin I would expect to edit the system's config files.

Alkarex:
@math-GH My thought was to define it in `./data/config-user.custom.php` prior to creating a user, but fair enough

marienfressinaud:
> @marienfressinaud Could that work for your own needs as well?

Yes, it should. Could you add "FreshRSS" to the `alt` attribute as well?

> I am not a big fan that it is tied to the user's settings and not to the system's settings.

I agree: if I want to change the logo, I don't want to change it for each user one by one.

Alkarex:
Ok, new approach with `logo_html` in main config.
With new `./data/config.custom.php` to provide custom values before install

math-GH:
(comment deleted. It did it wrong. Forgot to pull. Sorry. My fault)

math-GH:
It works fine for me (did not test it with a fresh new instance).

2 side notes:
- a documentation in the docs is needed
- any ideas how to style it? Inline CSS is not permitted. Theme CSS does not cover this feature. CustomCSS extension does not cover every user. Maybe "logo_html" should active an additional CSS file?

Alkarex:
Regarding styling, the CSS extension can be used to style a class, or the logo_html can point to a CSS file:

```php
'logo_html' => '<link rel="stylesheet" href="../themes/xCustom/logo.css" /><img class="logo" src="https://example.net/logo.png" alt="Logo" />',
```

app/Controllers/statsController.php
```diff
@@ -17,6 +17,7 @@ public function firstAction() {
 
 		$this->_csp([
 			'default-src' => "'self'",
+			'img-src' => '* data:',
```
marienfressinaud, line 20: I don't understand how this change is related to the PR

app/FreshRSS.php
```diff
@@ -32,6 +32,14 @@ public function init() {
 			die($message);
 		}
 
+		if (FreshRSS_Context::$system_conf->logo_html != '') {
+			// Relax Content Security Policy to allow external images if a custom logo HTML is used
+			Minz_ActionController::_defaultCsp([
```
marienfressinaud, line 37: Instead of changing the default CSP, I would rather set only the `img-src` policy. The `_csp()` method could be adapted to allow to change a specific policy (something like `$this->csp_policies = array_merge($this->csp_policies, $policies);`), but you would need to remove the default CSP policy warning as well (which is fine to me).

marienfressinaud (approved):

app/Controllers/statsController.php
Alkarex: In case the logo points to an image from an HTTP address

app/FreshRSS.php
Alkarex: I agree that the CSP methods could be made better, but editing (e.g. adding lines / merging) will likely open a number of cases, which I would rather keep to another PR. For instance, it could be nice to allow something for one domain, and then allow the same for another domain in two different calls (e.g. for extensions).
Having a way to set the global default CSP is still nice in my opinion, to avoid having to set it for each controller.

Alkarex:
@math-GH Some documentation added

app/Controllers/statsController.php
marienfressinaud: But if you change the default CSP in `app/FreshRSS.php`, is this line still needed?

Alkarex: Yes, because at the moment we do not have the ability to edit the CSP in a finer manner, so we are replacing the full CSP declaration

app/FreshRSS.php
marienfressinaud: I'm always a bit reluctant to change default values (I feel like we should avoid that as much as possible), but it's good enough :+1:

app/Controllers/statsController.php
marienfressinaud: Ah right, I thought default CSP and declared CSP were merged in `declareCspHeader()`!

Alkarex: That would be smart but not yet. Adding rules to make the policies *stricter* is trivial (natively supported by CSP by just adding lines), but merging / relaxing is more difficult

PR #4368, opened 2022-05-16
Allow CssXPath updates with composer

Art4:
Refs #4329

Changes proposed in this pull request:

- Restore the original folder structure in CssXPath.
- This allows easier updates of CssXPath (optionally with Composer).
- Add test for autoloading the `Gt\CssXPath\Translator` class

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] unit tests written (optional if too hard)
- [x] documentation updated
- [x] add irrelevant files to `.gitignore`

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #4374, opened 2022-05-17
Allow SimplePie updates with composer

Art4:
Refs #4329

- Restore original code of SimplePie to version 1.6.0
- This allows easier updates of SimplePie (optionally with Composer).
- Add tests for autoloading the SimplePie library
- Move all customized code to new lib `CustomSimplePie`
- classes in `CustomSimplePie` extends the namespaces SimplePie classes
- Add tests for autoloading `CustomSimplePie` classes

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated
- [ ] Take use of `CustomSimplePie` in app code

What do you think about this PR?

And I have more questions:

1. I was not sure where to locate the customized code from SimplePie. Therefore I put it to `lib/CustomSimplePie/`. Is this ok or could you tell me where I should move this code?
2. Should I remove/change the header in the `CustomSimplePie` classes?


Alkarex:
Hello,
I am not sure to understand the extension idea? It looks like we end up with two copies of SimplePie? At a first glance, it does not look easier :-)

P.S. A good part of our local patches are fixes inside the SimplePie code and not only additional functions (for which I agree that extending classes is an elegant approach)

Art4:
> Hello, I am not sure to understand the extension idea? It looks like we end up with two copies of SimplePie? At a first glance, it does not look easier :-)
> 
> P.S. A good part of our local patches are fixes inside the SimplePie code and not only additional functions (for which I agree that extending classes is an elegant approach)

The idea behind the extension is that the SimplePie code remains unchanged because major changes are due here in the future. At the same time, it should be possible to make bugfixes or changes as before (because merging in SimplePie unfortunately sometimes takes a bit longer).

CustomSimplePie contains only classes whose code is different from the original SimplePie. This way you don't have "two copies", but you can better see how CustomSimplePie differs from SimplePie. All of your local patches, that are not (yet) merged in SimplePie 1.6.0 are still in CustomSimplePie.

If changes in SimplePie have been merged, the SimplePie version can be updated and the changes in CustomSimplePie can be removed again. 

I have to check and rework this further, so that everything works as before. And at the moment the app code does not use CustomSimplePie but Original SimplePie. This is on the todo list.

Art4:
To continue here I would need answers to my questions.
Or does anyone have a better suggestion on how to solve the problem described?

Alkarex:
I still fail to understand the proposal, or if I understand it correctly, then I do not find it convenient.
Let's say one line of code needs to be changed in one file, then one would need to duplicate that file? But when that files gets updated upstream, we need to update two files? Twice the work? And lost ability to diff the whole code at once. It all sounds to be more complicated and error prone, and I fail to see any advantage

Art4:
I will try to explain it in another way. So far it looks like this:

**Problem 1**: You find a bug in SimplePie.
**Solution**: You create a PR in the SimplePie repository.

**Problem 2**: SimplePie takes a long time to merge, but you want to use the bugfix already in FreshRSS.
**Solution**: You have created a fork of SimplePie within FreshRSS where the bug is fixed. You then maintain changes in SimplePie back into the fork.

**Problem 3**: A feature is missing in SimplePie, e.g. `SimplePie::cleanMd5()`.
**Solution**: The feature is only implemented in the fork. The fork is therefore always slightly different from the original.

Have I understood this correctly so far?

Then comes the next problem, which I am trying to solve:

**Problem 4**: SimplePie does a big refactoring. To merge that into the fork, all differences have to be found, the refactored code has to be merged into the fork and the changes have to be applied back into the fork.

In https://github.com/FreshRSS/FreshRSS/pull/4329#issuecomment-1114677213 you stated:

>  Regarding SimplePie, I believe we will want to keep the ability to ship customised versions anyway, especially when waiting for upstream PRs to be merged, or when upstream PRs are not accepted

My suggested solution with this PR:

1. extract all changes of the fork into a custom library (here called `CustomSimplePie` for lack of creativity).
2. set the fork back to the original state
3. let `CustomSimplePie` build on the original SimplePie (first extend it, later possibly decorate it)
4. use in FreshRSS only `CustomSimplePie`.
5. Bonus: Future updates of SimplePie can be done with Composer, like CssXPath or PHPMailer.

If problem 1,2 or 3 occurs again in the future:

1. create the feature/bugfix in `CustomSimplePie`.
2. create a PR in SimplePie
3. a) If the PR was merged, update SimplePie in `lib` and remove the code from `CustomSimplePie
3. b) If the PR was not merged, leave the feature/bugfix in `CustomSimplePie`.

All these measures are a preparation for solving another problem that will come in the future.

**Problem 5**: Future versions of SimplePie (v2) might get dependencies to PSR interfaces and might require PSR implementations (e.g. cache or HTTP client).
**Solution**: PSR interfaces can be installed in `lib` folders, optionally also with Composer. Provide a PSR implementation based on the interfaces.

I would like to say again what my motivation is behind this PR. I am not (yet) a user of FreshRSS.

However, I have seen that you are a regular contributor to SimplePie. With the major changes being made (in part because of me) to SimplePie and more to come, I suspect that the way you contribute to SimplePie will become more difficult. Therefore, it feels to me like I'm making things more difficult for you by contributing to SimplePie. I don't like that.

In https://github.com/FreshRSS/FreshRSS/pull/4329#issuecomment-1114724600 you stated:

> Personally, I have a git clone of the libraries we use, and update them via a diff. In the case of SimplePie, which contains more changes, I cherry pick the latest commits.

I suspect that with the major changes in SimplePie, this will be more difficult, or it will require more work. Work that I would like to help with.

Whether this PR is accepted or not is fundamentally not important to me. I may also be wrong in my guess that maintaining your fork will be more difficult. Then maybe this PR can at least help you to find the differences between your fork and SimplePie.

Either way, I'm just trying to make sure my work is useful.

**Update** to answer your questions:

> Let's say one line of code needs to be changed in one file, then one would need to duplicate that file?

Not necessarily. It might also be possible to place the desired feature/bugfix differently in `CustomSimplePie` than in the PR at SimplePie. But to have a comparability to the PR, this would probably be recommended.

> But when that files gets updated upstream, we need to update two files? Twice the work?

You already have this double work, firstly by adding your bug fixes to your fork and secondly by adding the cherry-picking from upstream. After this PR you dont have the cherry-picking anymore, but you have to revert your own changes in `CustomSimplePie`.

And I would like to say that we only have this problem ("Problem 2") because SimplePie takes a lot of time with merging and releasing. If there were faster updates there, you wouldn't need to maintain a fork, you would just create PRs in SimplePie and then update your copy in `lib`. (And if a PR is not accepted, the change is applied to `CustomSimplePie`).

> And lost ability to diff the whole code at once.

I may be wrong, but you're losing that ability right now too, because SimplePie's code has changed a lot.

Frenzie:
I fundamentally don't understand how this approach is supposed to help with problem 4. If you copy entire classes and methods to make small changes in them, it seems to me that at best you've just kept the problem exactly the same, and frankly it looks an awful lot like double the work or worse.

> I suspect that the way you contribute to SimplePie will become more difficult.

Is it possible that you're coming at this from the wrong angle? It's not necessary to apply these patches at the level of the "compiled" library, which seems to be the underlying assumption behind everything you wrote. Take that assumption away and I don't see how the rest would follow.

Alkarex:
> **Problem 1**: You find a bug in SimplePie. **Solution**: You create a PR in the SimplePie repository.

Almost: first fixed in the FreshRSS version of SimplePie and if it seems to work as intended, a PR is then proposed upstream.

> **Problem 2**: SimplePie takes a long time to merge, but you want to use the bugfix already in FreshRSS. **Solution**: You have created a fork of SimplePie within FreshRSS where the bug is fixed. You then maintain changes in SimplePie back into the fork.
> 
> **Problem 3**: A feature is missing in SimplePie, e.g. `SimplePie::cleanMd5()`. **Solution**: The feature is only implemented in the fork. The fork is therefore always slightly different from the original.
> 
> Have I understood this correctly so far?

For now and at least the near future, problems 1-3 are near identical and result in a local SimplePie soft-"fork".

> **Problem 4**: SimplePie does a big refactoring. To merge that into the fork, all differences have to be found, the refactored code has to be merged into the fork and the changes have to be applied back into the fork.

This upstream refactoring will surely introduce some lag (i.e. sticking with 1.6.x for some time), especially if that big refactoring has little intrinsic value, because the motivation to produce this work will be low.

Yes, the most likely approach will indeed be to re-apply manually, once, the local diffs/patches on top of the refactored version.
I have not assessed how much work that would be.

Meanwhile, if there is any desirable patch landing in SimplePie, it might be cherry-picked and back-ported to 1.6.x.

Note that I still hope to continue reducing the local differences by proposing (or re-proposing, as some had been refused) some changes upstream.

> **Problem 5**: Future versions of SimplePie (v2) might get dependencies to PSR interfaces and might require PSR implementations (e.g. cache or HTTP client). **Solution**: PSR interfaces can be installed in `lib` folders, optionally also with Composer. Provide a PSR implementation based on the interfaces.

One problem at a time :-) I do not expect that to be a significant challenge.

> I would like to say again what my motivation is behind this PR. I am not (yet) a user of FreshRSS.
> 
> However, I have seen that you are a regular contributor to SimplePie. With the major changes being made (in part because of me) to SimplePie and more to come, I suspect that the way you contribute to SimplePie will become more difficult. Therefore, it feels to me like I'm making things more difficult for you by contributing to SimplePie. I don't like that.

I very much appreciate your efforts, attempts, and explanations 👍🏻 

> I suspect that with the major changes in SimplePie, this will be more difficult, or it will require more work. Work that I would like to help with.

I think the most likely, as written above, is that it will introduce some lag, but not necessarily that it will remain more difficult down the line.

> Whether this PR is accepted or not is fundamentally not important to me. I may also be wrong in my guess that maintaining your fork will be more difficult. Then maybe this PR can at least help you to find the differences between your fork and SimplePie.
> 
> Either way, I'm just trying to make sure my work is useful.

Thanks :-) Yes, I indeed do not believe that it will make upstream PR significantly more difficult once the (non-rewarding) work of merging the big refactoring has been done.

Art4:
@Frenzie 

> I fundamentally don't understand how this approach is supposed to help with problem 4. If you copy entire classes and methods to make small changes in them, it seems to me that at best you've just kept the problem exactly the same, and frankly it looks an awful lot like double the work or worse.

Please keep in mind that this PR is only a draft for now. It is possible to refactor the minimal changes in the many classes into a few classes. For now, I haven't done that to show the differences between CustomSimplePie and SimplePie.

The goal could be that CustomSimplePie does not need the fork of individual classes, but acts as a decorator that overrides functionality of SimplePie in case of bugs or missing features. So you could also put the function `customSimplePie()` into the `CustomSimplePie` library. https://github.com/FreshRSS/FreshRSS/blob/1.19.2/lib/lib_rss.php#L215

> > I suspect that the way you contribute to SimplePie will become more difficult.
> 
> Is it possible that you're coming at this from the wrong angle? It's not necessary to apply these patches at the level of the "compiled" library, which seems to be the underlying assumption behind everything you wrote. Take that assumption away and I don't see how the rest would follow.

Sure, it is also possible that I am completely overestimating problem 4 (and 5), or that there are better solutions for it. In that case, this PR can simply be closed and please forgive me for wasting your time. If not, I'd be happy to get hints on what direction the PR needs to change to be helpful. Then I would also be willing to invest more time in this PR.

Frenzie:
> Please keep in mind that this PR is only a draft for now. It is possible to refactor the minimal changes in the many classes into a few classes. For now, I haven't done that to show the differences between CustomSimplePie and SimplePie.

It's the concept I'm talking about, not anything done in this PR.

Without a big refactoring, applying a few patches is generally quite straightforward.

After a big refactoring, naturally it takes more effort to adapt.

In my view, this is a proposal to always make it hard because you always have to put in effort that would normally be performed mostly by tools.

Take the cleanMd5() example from above. You can define a CustomSimplePie class with a cleanMd5() method. Simple, that seems to make a lot of sense. But it looks like now you have to also override all of init() along with it. So now I have to somehow extract init() and diff it against my own custom init() in order to do the exact same thing as what was previously an almost zero-effort rebase against the latest upstream tag. Every rebase is now a problematic rebase because the usual tools can't do their job.

> If not, I'd be happy to get hints on what direction the PR needs to change to be helpful.

Assuming the Composer-based flow is helpful, make it use https://github.com/FreshRSS/simplepie instead of upstream. I.e., completely identical to #4329 and #4368.

Art4:
I have time to devote to this PR again.

> Assuming the Composer-based flow is helpful, make it use https://github.com/FreshRSS/simplepie instead of upstream. I.e., completely identical to #4329 and #4368.

If I see this correctly, this repo is not actively maintained. @Alkarex  Would it be helpful if I change this PR to use this repo?

Alkarex:
> I have time to devote to this PR again.
> 
> > Assuming the Composer-based flow is helpful, make it use https://github.com/FreshRSS/simplepie instead of upstream. I.e., completely identical to #4329 and #4368.
> 
> If I see this correctly, this repo is not actively maintained. @Alkarex Would it be helpful if I change this PR to use this repo?

Yes, that could actually be a good idea to experiment. https://github.com/FreshRSS/simplepie is not in sync with the code shipped in FreshRSS yet though. It has so far mostly be used to send some pull requests upstream

lib/composer.json
```diff
@@ -13,7 +17,8 @@
     "require": {
         "php": ">=7.0.0",
         "phpgt/cssxpath": "dev-master#4fbe420aba3d9e729940107ded4236a835a1a132",
-        "phpmailer/phpmailer": "6.6.0"
+        "phpmailer/phpmailer": "6.6.0",
+        "simplepie/simplepie": "dev-fix-base-url"
```
Art4: I've used the newest brach for test purposes. This must be changed to `dev-master` or git tags.

Art4:
OK, I changed the PR so that the code is now loaded from the https://github.com/FreshRSS/simplepie/ repository. I used the `fix-base-url` branch for testing purposes. But this branch does not contain all local modifications like the `$simplepie->set_syslog()` method.

I also used the namespace SimplePie classes directly, so I could delete the deprecated `simplepie/simplepie/library` folder.

@Alkarex Should I open a PR in https://github.com/FreshRSS/simplepie with all diffs between `lib/SimplePie` folder and the `fix-base-url` branch?

Frenzie:
Thanks, that significantly improves this PR.

> Should I open a PR in https://github.com/FreshRSS/simplepie with all diffs between lib/SimplePie folder and the fix-base-url branch?

If you mean just as a poc not to be merged, sure. ^_^

lib/composer.json
Frenzie: I would propose something like a freshrss-1.6.0 branch (and then freshrss-1.6.1 etc. in the future), where our few modifications are rebased or cherry-picked on top of upstream if necessary. Or something similar with tags if that can work in an acceptable manner.

Something like `dev-master` doesn't sound good at all if you want to check out FreshRSS 1.20.0 a few years down the line. ^_^

Art4:
> Thanks, that significantly improves this PR.
> 
> > Should I open a PR in https://github.com/FreshRSS/simplepie with all diffs between lib/SimplePie folder and the fix-base-url branch?
> 
> If you mean just as a poc not to be merged, sure. ^_^

Yes, of course. I just need an up-to-date branch where all diffs from FreshRSS are included. And at this point it just occurs to me that I could also just use my repository at https://github.com/Art4/simplepie for this poc. 😁

Art4:
I've created a new branch `freshrss` on my repository at https://github.com/Art4/simplepie and created 2 commits and git tags there.

- Tag [1.5.8-freshrss](https://github.com/Art4/simplepie/releases/tag/1.5.8-freshrss) represents SimplePie 1.5.8 + latest changes from FreshRSS
- Tag [1.6.0-freshrss](https://github.com/Art4/simplepie/releases/tag/1.6.0-freshrss) represents SimplePie 1.6.0 + latest changes from FreshRSS

Actually, now I wanted to require these git tags in composer.json, but composer can't handle these git tags. 😓 
Changing the repo url and setting the version to `"simplepie/simplepie": "1.5.8-freshrss"` results in error:

```
In VersionParser.php line 521:

  Could not parse version constraint 1.5.8-freshrss: Invalid version string "1.5.8-freshrss"
```

Is this a bug in Composer or am I doing something wrong?

Frenzie:
Shouldn't Composer version strings be prefixed by `@` or something like that?

Alkarex:
See example:

https://github.com/FreshRSS/FreshRSS/blob/fea12a0957c5156b8d4ce3f329aa20a819a1e316/lib/composer.json#L15

Art4:
Using the commit hash works fine, but I tried explicitly requiring the named git tags. Composer seems to have a limitation or a bug in the SemVer VersionParser though. When I request the (non-existent) tag `100.0.0-alpha`, Composer reports that the tag does not exist. If I request the (also non-existent) tag `100.0.0-foobar`, Composer aborts with the above mentioned error. 😅 

I will use the commit hash then.

Frenzie:
@Alkarex Right, I mean wasn't there also some kind of syntax like `"phpgt/cssxpath": "@some-tag", `?

Alkarex:
For git custom repositories, I only know the `dev-` prefix and the hash `#`

Frenzie:
I'm probably confused with some similar system then. :sweat_smile:

Art4:
I have now implemented the Poc so far. The main changes are now no longer in this PR, but in the `freshrss` branch in my SimplePie repo. The FreshRSS customizations can be compared with the official SimplePie repo: https://github.com/simplepie/simplepie/compare/master...Art4:freshrss?expand=1

**What are the next steps?**

If this poc is accepted so far, I can get to work on fixing the failing checks. Also, in the FreshRSS/SimplePie repository a branch would need to be created to match the changes from hte `freshrss` branch in my repo. Then in the `composer.json` the repository has to be changed and the commit hash have to be adjusted to the `freshrss` branch or the commit hash.

Future customizations to the SimplePie fork would then work like this: A change is tested in the fork (or before in FreshRSS `lib/simplepie/simplepie`. Then an upstream PR to SimplePie can be created via the fork. Changes in SimplePie can now be simply merged into the fork repository. And the code of SimplePie in FreshRSS can be updated by (eventually changing the commit hash and) running `composer update`.

Frenzie:
Ignoring the namespace stuff, it's basically been reduced to adding `"simplepie/simplepie": "dev-freshrss#94db37e06b20bab99d7474f58346b3e98ee90c0e"` to composer.json? That works for me as far as this PR goes.

For the special branches ideally the changes would be a touch more meaningful than one giant lump of a commit both for clarity and rebasing but that's not of the greatest importance. @Alkarex will be able to judge that better.

Art4:
> Ignoring the namespace stuff, it's basically been reduced to adding `"simplepie/simplepie": "dev-freshrss#94db37e06b20bab99d7474f58346b3e98ee90c0e"` to composer.json? That works for me as far as this PR goes.

Yes, that would describe it best for the FreshRSS repository. As I said the main changes are now in the `freshrss` branch in my SimplePie repo.

> For the special branches ideally the changes would be a touch more meaningful than one giant lump of a commit both for clarity and rebasing but that's not of the greatest importance. @Alkarex will be able to judge that better.

I have split the changes into 3 large commits in the hope that the changes can be better reviewed. So "the (non-rewarding) work of merging the big refactoring" has now been done, to quote @Alkarex. 😄

Even though I took a lot of effort to work carefully, these changes now need to be reviewed again carefully. If there is a up-to-date branch in the FreshRSS/SimplePie repo, I could create a PR against it and help with the review and other questions.

Alkarex:
Thanks for the efforts so far @Art4 👍🏻 
I will look at it after we release 1.20.1

math-GH:
> I will look at it after we release 1.20.1

It is a good time now to remind that 1.20.1 is already released ;)

Alkarex:
Resuming the work here.
@Art4 's branch published to https://github.com/FreshRSS/simplepie/tree/freshrss
Need to bit more cleaning and update, and will be used as reference for our composer

tests/lib/SimplePie/SimplePieTest.php
```diff
@@ -0,0 +1,21 @@
...
+		self::assertTrue(class_exists(\SimplePie\Misc::class));
+	}
+
+	public function testPsr0SimplePieClassExists(): void {
+		self::assertTrue(class_exists('SimplePie'));
+	}
+
+	public function testPsr0SimplePieMiscClassExists(): void {
+		self::assertTrue(class_exists('SimplePie_Misc'));
+	}
```
Alkarex: I do not believe those two last PSR0 tests are needed, are they?

Art4: They where added to confirm BC. If FreshRSS uses the PSR-4 classes already, the BC tests could be removed.

Alkarex:
Upstream PRs:
* https://github.com/simplepie/simplepie/pull/869
* https://github.com/simplepie/simplepie/pull/868

Alkarex:
@Art4 If you are around, your input would be welcome.

What I lack to finalise this work is to re-implement the cache-related features needed for the performance of FreshRSS:
https://github.com/FreshRSS/simplepie/pull/11

Requirements:
* Skip processing (parsing, sanitizing, etc.) when there is a cache match (e.g. server returning 304) to save resources;
* Do the same when the content is unchanged (based on a hash), even when the server returns a 200 instead of 304;
* Clean the content from noise (e.g. build date, statistics, comments...) before computing the cache hash;
* A way to know whether the SimplePie cache was used or not;
* Modification time of the SimplePie cache.

But it must be the third time I look into and I am yet again giving up tonight.

The old SimplePie code was admittedly a bit challenging to patch but doable; now it feels like a maze 😵
The new PSR stuff gets in the way without being helpful - but maybe I am missing something?

The new SimplePie caching tests are full of hardcoded things, there are new cache classes (e.g. `Psr16`) which seem to be intended to replace other ones, but only partially (not finished? - if so, why are there used?), and seems to lack even basic methods such as getting modification time.

For instance, the `set_cache()` method does not seem coherent with the `get_cache()` one, which does not use the `new Psr16()` stuff for the defaults:

https://github.com/simplepie/simplepie/blob/9385c54382edec0a768fdbe310c3e580c1b5dc81/src/SimplePie.php#L958-L961

https://github.com/simplepie/simplepie/blob/9385c54382edec0a768fdbe310c3e580c1b5dc81/src/SimplePie.php#L3417-L3428

So if you have some insights about what approach to use to best implement the requirements stated above, please let me know 💡

Art4:
@Alkarex The new internally cache interface is called [`DataCache`](https://github.com/simplepie/simplepie/blob/1.8.0/src/Cache/DataCache.php) and is a subset of the current [`Base`](https://github.com/simplepie/simplepie/blob/1.8.0/src/Cache/Base.php) cache interface and the [PSR-16](https://www.php-fig.org/psr/psr-16/) cache interface. It was created for BC reasons.

The DataCache only supports `get_data()`, `set_data()` and `delete_data()`. There is no longer a concept of a modification time in the cache or refreshing a value.

If the (deprecated) `Base` cache is removed in v2, the (internal) `DataCache` will be removed as well and SimplePie will only use PSR-16 internally and could use more features from PSR-16 like `setMultiple()` to speed things up.

Best way for future compatibility would be to create a new or use an existing [PSR-16 implementation](https://packagist.org/providers/psr/simple-cache-implementation) and provide this to SimplePie via `$simplepie->set_cache($psr16Cache)`.

So to adress your points:

> - Skip processing (parsing, sanitizing, etc.) when there is a cache match (e.g. server returning 304) to save resources;
> - Do the same when the content is unchanged (based on a hash), even when the server returns a 200 instead of 304;
> - Clean the content from noise (e.g. build date, statistics, comments...) before computing the cache hash;

These points has to be done in the fork of the SimplePie class. This is no longer part of the cache implementation.

> - A way to know whether the SimplePie cache was used or not;

Provide your own PSR-16 implementation cache, that will be called for getting, setting or deleting a value. There you have full control over the value, ttl, etc.

> - Modification time of the SimplePie cache.

This concept of `mtime()` and `touch()` was removed because it only works efficient for the DB or file cache but not for Redis or Memcache. But it might be possible to implement this concept in your PSR-16 implementation.

Alkarex:
@Art4 Thanks for the feedback! If you can, please check that draft https://github.com/FreshRSS/simplepie/pull/11 and let me know if you can think of a better approach.
My problem with `$simplepie->set_cache($psr16Cache)` is that it does not seem to allow adding any of my above requirements, as you seem to confirm for the first three points, and even for the last two points, adding them to the psr16Cache would not expose their features in any usable way that I can see.
To avoid a fork/patch of the SimplePie class, I believe a few customisation features should be added, such as callbacks at critical points in the logic.

Alkarex (approved):

Frenzie:
I trust you @Alkarex but let us know if you'd like anything specific to be tested. :-)

Alkarex:
> I trust you @Alkarex but let us know if you'd like anything specific to be tested. :-)

Hehe 😅 I am doing it on my side already, but basically running with this branch and keeping an eye on the logs would be nice (in development mode to get the debug messages as well), including the logs of the servers you own, which are serving RSS feeds (expecting the proper HTTP 304 responses). So far so good for me.

Alkarex:
Additional minor upstream PRs
* https://github.com/simplepie/simplepie/pull/878
* https://github.com/simplepie/simplepie/pull/879

And this one still pending:
* https://github.com/simplepie/simplepie/pull/861

Art4:
@Alkarex Any thoughts about the https://github.com/Art4/FreshRSS/pull/1 PSR-16 cache implementation? Should I port the PR to the FreshRSS/FreshRSS repository or will it not be needed at all?

Alkarex:
@Art4 Thanks for following up. Yes please, see:

https://github.com/FreshRSS/simplepie/pull/11#discussion_r1749304406
> @Art4 Could you please change the target of your PR to https://github.com/FreshRSS/simplepie/tree/freshrss I am not sure I will be able to use it as it does not seem to allow offloading much of my needed patches, while adding 500 lines of boilerplate, but I will try to give it another look later. In particular, I still fail to see so far how to achieve your claim of _Makes most changes in the cache class of the SimplePie fork not necessary_

And any tips welcome on how to reduce the patches
https://github.com/simplepie/simplepie/compare/master...FreshRSS:simplepie:freshrss

Art4:
> @Art4 Thanks for following up. Yes please, see:
> 
> [FreshRSS/simplepie#11 (comment)](https://github.com/FreshRSS/simplepie/pull/11#discussion_r1749304406)
> 
> > @Art4 Could you please change the target of your PR to https://github.com/FreshRSS/simplepie/tree/freshrss I am not sure I will be able to use it as it does not seem to allow offloading much of my needed patches, while adding 500 lines of boilerplate, but I will try to give it another look later. In particular, I still fail to see so far how to achieve your claim of _Makes most changes in the cache class of the SimplePie fork not necessary_
> 

Oh, thank you. I missed this comment in the noisy Github notifications.

> And any tips welcome on how to reduce the patches [simplepie/simplepie@master...FreshRSS:simplepie:freshrss](https://github.com/simplepie/simplepie/compare/master...FreshRSS:simplepie:freshrss)

Ok, will take a look.

Alkarex:
I have just added some better / finer logging.
Indeed, after the cache stuff, I am tracking down a double HTTP request being made to the server in case of redirection. So we get for instance: `302, 200, 200`, instead of just `302, 200`.
Edit: Seems to be a cache not working

Alkarex:
I have fixed a cache-related case but I believe there is still a problem with the way SimplePie caches HTTP redirections, or more exactly does not seem to do so. But that might wait till another PR, as I do not think this is a new bug.

Alkarex:
I need to clean a bit more the cache metadata

Alkarex:
Let's give it a spin 🚀

Alkarex:
Follow-up to start adding some HTTP caching compliance https://github.com/FreshRSS/simplepie/pull/26
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

PR #1828, opened 2018-03-10
Light Boolean search implementation

Alkarex:
`Hello intitle:World OR date:P1D example OR author:Else intitle:"This or that"`
https://github.com/FreshRSS/FreshRSS/issues/879

PR #4378, opened 2022-05-18
New search engine

Alkarex:
Support (nested) parentheses:

* `(author:Alice OR intitle:hello) (author:Bob OR intitle:world)`
* `(author:Alice intitle:hello) OR (author:Bob intitle:world)`

Support calling user queries from the search field: `search:"My query"` or `search:QueryA`.

A use-case is to expose an RSS feed with a stable address or an external API call with the ability to update the underlying user query. Can be combined with other filters such as `search:"My query" date:P3d`

* `search:QA search:QB`
* `search:QA OR search:QB`


Alkarex:
Will implement a parenthesis mechanism after all to make that more useful.

Alkarex:
Added support for (nested) parentheses:

* `(author:Alice OR intitle:hello) (author:Bob OR intitle:world)`
* `(author:Alice intitle:hello) OR (author:Bob intitle:world)`

app/Models/EntryDAO.php
```diff
@@ -696,281 +700,301 @@ public function searchIdByGuid($id_feed, $guid) {
...
-	}
+	/** @param FreshRSS_BooleanSearch $filters */
+	public static function sqlBooleanSearch(string $alias, $filters, int $level = 0) {
+		$search = '';
+		$values = [];
 
-	/**
-	 * @param FreshRSS_BooleanSearch|null $filters
-	 */
-	protected function sqlListEntriesWhere(string $alias = '', $filters = null, int $state = FreshRSS_Entry::STATE_ALL,
```
Alkarex, line 706: This looks like a big diff in this view, but most lines changed under `sqlListEntriesWhere()` have just been moved/extracted to a new dedicated function `sqlBooleanSearch()` responsible for generating the SQL search expression.

app/Models/EntryDAO.php
```diff
@@ -2,23 +2,27 @@
 
 class FreshRSS_EntryDAO extends Minz_ModelPdo implements FreshRSS_Searchable {
 
-	public function isCompressed(): bool {
+	public static function isCompressed(): bool {
```
Alkarex, line 5: I have made a few functions static so that the new `sqlBooleanSearch()` can also be static and easier to test.

math-GH:
How does `S:Q1` work?

Why `S?` for user query?
Is `Q1` for the first user query? How do I know the correct user query number?

math-GH:
I tried something and it looks fine for me :)

Alkarex:
> Is `Q1` for the first user query? How do I know the correct user query number?

I have replaced the examples here to QA to underline the fact that it is based on the user query *name*. 

> How does `S:Q1` work?

It works with a simple search&replace

> Why `S?` for user query? 

I did not like Q too much because it typically refers to the full query (like in a URL with `q=xxx`), so S is for *search*, but there was not much more thoughts to it :-P

> I tried something and it looks fine for me :)

Great :-)

math-GH:
maybe `UQ` (user query) would be better than `S`.

`UQ:id` f.e. `UQ:1` for the user query with the ID 1 (the IDs starts with a 0).
`uq:"query name"` f.e. `uq:"Query n°2"` for the user query with the name "Query n°2" (I am not a big fan of using editable names, but I understand the need)

math-GH:
And the docs need the new feature: https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/users/03_Main_view.md#user-content-with-the-search-field ;)

Frenzie:
I'm not overly fond of the name "user query" regardless. It's very technical/jargon. Perhaps something like custom search/filter? But agreed that UQ seems more a propos if user query is the term we use for the functionality.

Alkarex:
> And the docs need the new feature: https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/users/03_Main_view.md#user-content-with-the-search-field ;)

@math-GH I have already added a few lines in the doc in this PR

> I'm not overly fond of the name "user query" regardless. It's very technical/jargon. Perhaps something like custom search/filter? But agreed that UQ seems more a propos if user query is the term we use for the functionality.

@Frenzie Indeed, I am not so found of "user query" either, so I would rather not cement it further. *Saved search* maybe (avoiding the unfortunate *SS* abbreviation...)?

* https://support.mozilla.org/en-US/kb/using-saved-searches
* https://docs.microsoft.com/en-us/windows/win32/search/-search-savedsearchfileformat

> maybe `UQ` (user query) would be better than `S`.
> 
> `UQ:id` f.e. `UQ:1` for the user query with the ID 1 (the IDs starts with a 0). `uq:"query name"` f.e. `uq:"Query n°2"` for the user query with the name "Query n°2" (I am not a big fan of using editable names, but I understand the need)

At the moment, IDs are very unstable. For instance, if the user deletes a query, all the ones after it will change ID, which is a bit unfortunate and without good mitigation option. We do not even show the ID in the configuration view.

But indeed, to keep a similar convention than for labels (short `L:12` for IDs and long `label:MyLabel` for names) we could go for something else such as `uquery:something` (and potentially `uq:12` for the ID version):

1. uquery / uq
2. ssearch / sh
3. squery / sq
4. ...?

What do you prefer?

Frenzie:
> Saved search maybe (avoiding the unfortunate SS abbreviation...)?

I think I'd simply call them *Filters* without adding words like custom/user/saved. The same applies to *Searches*, but perhaps that would indeed sound better in combination with an adjective. For any combination with search I think it's fine to simply use S to refer to a single search, regardless if the lot of them is called saved searches.

> ssearch / sh

Btw, what's an sh?

math-GH:
The User Queries are bookmarked filters (or favorite filters).

Frenzie:
_Favorite filters_ has some alliterative qualities, but I wouldn't add a pleonastic adjective like bookmarked or favorite since that's just what they are.

In any case, the main point is to avoid the word query, which just means question in regular people language. Almost anything with _Filter_ or _Search_ is better. :-)

Frenzie:
PS The manual also calls them filters already (which I like ;-) )

image

math-GH:
As a reminder: "filter" is already used often:

grafik

grafik

Frenzie:
That's an argument in favor, but it sounds like you mean it negatively? ;-P

Alkarex:
"Saved searches" really seems to be the usual expression; another example https://guides.lib.unc.edu/search-pubmed/save-search
So in another PR, I think we should change our terminology accordingly, if someone is motivated :-)

In this PR, based on your feedback, I have updated the name-based search to `search:MySearch` or `search:"My Search"` and introduced a new possibility to address saved searches by ID like `S:3` as desired by @math-GH 

Note @math-GH that our saved search IDs start at 1, not 0:

image

Alkarex:
Improvements in follow-up: https://github.com/FreshRSS/FreshRSS/pull/4503
Was itself a follow-up of: https://github.com/FreshRSS/FreshRSS/pull/1828

PR #4407, opened 2022-06-07
Dynamic OPML

### LINKED ISSUE #4191: [Feature] Subscribe to a dynamic OPML Feed
**Description**

I am currently an active user of Inoreader and an exciting feature that they provide is dynamic OPML Subscriptions.

Rather than importing an OPML feed, you would subscribe to it.

In their words:

> Importing an OPML
> If you import an OPML file in your RSS reader, you are creating a static copy of the reading list. If the owner of the original list updates the list, this change will not be shown in your feed collection. 

> Subscribing to an OPML
> An OPML subscription, however, creates a live connection between the original source and your feed folder. Whenever a feed is added or deleted from the original list, the change will be reflected in your subscriptions. The notification system in Inoreader will notify you that new feeds were imported or existing feeds were removed. 

I currently subscribe to the Engineering Blogs List OPML from this Github Repo (https://github.com/kilimchoi/engineering-blogs). The OPML feed that I subscribe to is here [engineering_blogs.opml](https://raw.githubusercontent.com/kilimchoi/engineering-blogs/master/engineering_blogs.opml)

Since this feed is dynamically updated with new feeds and deletion of stale feeds, the OPML is ever-changing. It's nice to see that those feed changes are reflected in my FreshRSS instance.

Provide the ability to subscribe to an OPML feed such as: `https://raw.githubusercontent.com/kilimchoi/engineering-blogs/master/engineering_blogs.opml`.

Configuration options could be:

* URL
* Description of this OPML subscription
* Folder (all feeds are categorized in this folder)
* Synchronization Options (Append Only, Full-Sync)

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4191

* Add the ability for a category to be dynamically populated with a list of feeds provided by a remote OPML, that is to say to **consume** a dynamic OPML / reading list.
  * This includes automatically adding new feeds, and muting (disabling) feeds that have been removed from the list (with a button to manually delete all the muted / disabled feeds).
  * Does not support - by design - sub-categories, but will import feeds from sub-categories.
* Add the ability to dynamically export a FreshRSS view (all, feed, category) into a dynamic OPML / reading list, that is to say to **produce** a dynamic OPML / reading list.
  * E.g. `https://freshrss.example.net/?a=opml&get=c3&token=...`

Background reading:
* http://web.archive.org/web/20060101040915/http://www.reallysimplesyndication.com/2005/10/13
* https://techcrunch.com/2005/10/16/my-thoughts-on-reading-lists/
* http://scripting.com/2017/06/26.html#a080636

* [ ] (Future PR) Better security tokens will be added to make the FreshRSS publication of dynamic OPML / reading lists more useful.
* [ ] (Future PR) Some more documentation for this dynamic OPML feature. Help welcome.
* [ ] (Future PR) The dynamic OPML export is not visible at the moment (only in the HTML code with a standard as shown below), so that should be improved somehow. Help welcome.

```html
<meta rel="outline" type="text/x-opml" title="OPML" href="./?a=opml" />
```

* [ ] (Future PR) Make a custom TTL (refresh frequency) per category just like we have for feeds, in addition of the global parameter, which is by default to refresh dynamic OPMLs every 12 hours

There are a few other remaining edge cases, such as when feeds with a URL that has been (permanently) redirected, and feeds referenced in more than one dynamic OPML.


Alkarex:
@reidcooper if you are around, this PR starts to work, so tests welcome already:

image


image

What mainly lack is the finalisation of the automatic refreshing of the dynamic OPMLs (the manual button works already, including muting/disabling feeds that are not in the OPML anymore) and a number of edge cases (such as the tricky case of feeds with HTTP redirections).

reidcooper:
@Alkarex Hey hey! That's great news. Thanks for the update and powering through the update. I can give the beta a shot and see how it goes. I may be able to get to it this week.

app/views/subscription/add.phtml
```diff
@@ -218,4 +218,35 @@
...
+	<h2>
+		<?= _t('sub.title.add_dynamic_opml') ?>
+		<span title="<?= _t('sub.category.dynamic_opml') ?>">🗲</span>
+	</h2>
+	<form action="<?= _url('category', 'create') ?>" method="post">
+		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
+		<div class="form-group">
+			<label class="group-name" for="new-category"><?= _t('sub.category') ?></label>
+			<div class="group-controls">
+				<input id="new-category" name="new-category" type="text" autocomplete="off"/>
```
math-GH: `required="required"` missed (fetched by the backend)

app/views/subscription/add.phtml
```diff
@@ -218,4 +218,35 @@
 			</div>
 		</div>
 	</form>
+
+	<h2>
+		<?= _t('sub.title.add_dynamic_opml') ?>
+		<span title="<?= _t('sub.category.dynamic_opml') ?>">🗲</span>
```
math-GH: I am not so happy to have here an icon. It would be the first time to have an icon in the headline

Alkarex: Actually, my intention is to have that icon not only in the title, but after the name of a dynamic category everywhere it is used, since it is a behaviour that is so peculiar that it can lead to confusion when forgotten:

image

image

app/views/subscription/add.phtml
Alkarex: Fixed

app/views/subscription/add.phtml
math-GH: I do not see the benefit right now, because the update of feed list needs an manual update.

Maybe it would make (more) sense to mark the feeds that are imported with dynamic OPML? These feeds could disappear after a feed list update.

math-GH:
Does it make sense to add feeds manually to the dyn. OPML category? What will happen after an update of the feed list?

Should we block to add more feeds into the dyn. OPML category?

app/views/subscription/add.phtml
Alkarex: The feeds under a dynamic category will automatically change over time. It is the category that makes it special, not the feeds, which can be for instance moved to another *normal* category

Alkarex:
> Does it make sense to add feeds manually to the dyn. OPML category?
> 
> Should we block to add more feeds into the dyn. OPML category?

Good point, I have disabled adding *new* feeds to a dynamic category. However, it is still possible to drag & drop an existing feed to such a category.

Indeed, I would like to keep the option to move feeds out to another *normal* category. But then, what if you change your mind and want to move it back? Suddenly we would have to keep track of which feeds would be allowed or not, which would require a bunch of custom code and would not necessarily result in a better experience.

> What will happen after an update of the feed list?

Feeds, which URL is not in the dynamic OPML will be muted (disabled)

app/views/subscription/add.phtml
Alkarex: (I am not sure I have fully understood your comment)

math-GH:
I would suggest to use a "official" OPML logo (see also: https://github.com/scripting/opml.org/issues/5):
opml

Right now there is no icon for dynamic OPML, so I created a new one:
opml-dyn

math-GH:
Should we add also an dyn. OPML import on the import/export screen?

grafik

Alkarex:
> Right now there is no icon for dynamic OPML, so I created a new one: opml-dyn

Looks great 👍🏻

Alkarex:
I have experimented a big with your nice OPML logos and I will clearly use them for links to our OPML views 👍🏻 

But for inline use, although it works ok, it is not as readable and I think I still have a little preference to a Unicode or emoji such as 🗲
What do people think? :-)

image

image

image

image

image

image

image

image

Frenzie:
I agree with the gist about being able to tell what it looks like better, but I'm not quite sure how lightning/electricity relates to OPML. :-)

Alkarex:
> I agree with the gist about being able to tell what it looks like better, but I'm not quite sure how lightning/electricity relates to OPML. :-)

The ideas I wanted to convey are: dynamic / magic but also dangerous

Alkarex:
> Should we add also an dyn. OPML import on the import/export screen?

@math-GH Ok, something like that?

image

Alkarex:
Let's merge that to get some more tests and potentially come back to the cosmetic details in another PR.
Furthermore, I have another feature regarding sharing / tokens, which builds upon this PR, and which I would ideally like to implement before releasing 1.20.0.

reidcooper:
@Alkarex I was finally able to try out the Docker Edge release here, added an OPML feed, and it worked! I was able to see incoming feeds sync and mark them as read. Let me try this out for a few days and see how it performs. Sorry for the late reply and thank you very much for your hard work on developing this feature <3

app/Models/FeedDAO.php
```diff
@@ -121,11 +100,36 @@ public function addFeedObject(FreshRSS_Feed $feed): int {
...
-		}
+		} else {
+			// The feed already exists so make sure it is not muted
+			$feed->_ttl($feed_search->ttl());
+			$feed->_mute(false);
+
+			// Merge existing and import attributes
+			$existingAttributes = $feed_search->attributes();
+			$importAttributes = $feed->attributes();
+			$feed->_attributes('', array_merge_recursive($existingAttributes, $importAttributes));
```
Alkarex, line 111: Merge bug fixed in https://github.com/FreshRSS/FreshRSS/pull/4708

app/Models/Feed.php
```diff
@@ -584,7 +594,8 @@ public function loadHtmlXpath(bool $loadDetails = false, bool $noCache = false,
 			return null;
 		}
 
-		$html = getHtml($feedSourceUrl, $attributes);
+		$cachePath = FreshRSS_Feed::cacheFilename($feedSourceUrl, $attributes, FreshRSS_Feed::KIND_HTML_XPATH);
+		$html = httpGet($feedSourceUrl, $cachePath, 'html', $attributes);
```
Alkarex, line 598: Bug with `$attributes` fixed in https://github.com/FreshRSS/FreshRSS/pull/4759

Alkarex:
A little article on the Fediverse https://framapiaf.org/@freshrss/115395177812589475

PR #4311, opened 2022-04-12
Created new theme: Dark pink

miicat:
Created Dark pink theme based on Alternative Dark theme

PR #4409, opened 2022-06-11
Added themeing to missing "category" -icon

miicat:
"Category" -icon was only shown on mobile and it's now themed correctly, too


Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/4311

PR #4394, opened 2022-05-28
Slider: close via shortcut

math-GH:
Before:
It made me odd, that the slider cannot be closed via a shortcut (f.e. `ESC`).

After:
The sliders can be closed via the shortcut that closes the menu.

Additional: Via shortcut (f.e. `F1`), you can now open the documentation everywhere.

Changes proposed in this pull request:

- sliders open/close via anchor (the same as the menus) (before: sliders worked only with JavaScript)
- shortcuts are available everywhere (before: only in the feed streams)
- HTML: slider closer is now after the slider (same as the menus)

How to test the feature manually:

1. open the sliders (f.e. subscription management -> config a feed or category; or config an extension manager ->config an extension)
2. the slider opens
3. close the slider via "close menus" shortcut  (f.e. `ESC`)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #4416, opened 2022-06-16
Fix/Improved: Slider mobile

math-GH:
Part 1: Regression from #4394 (The mobile view was broken (not 100% width, close button not visible in the slider))
Part 2: slider loads only the content, not the whole layout (introduced with #4310)

1. open a slider (f.e. config a feed)
2. see this slider in desktop mode
3. see this slider in mobile mode

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/views/extension/index.phtml
```diff
@@ -80,13 +80,17 @@
 </main>
 
 <?php $class = isset($this->extension) ? ' active' : ''; ?>
-<aside id="slider" class="scrollbar-thin<?= $class ?>">
-	<?php
-		if (isset($this->extension)) {
-			$this->renderHelper('extension/configure');
-		}
-	?>
+<?php $closelink = isset($this->extension) ? _url('extension', 'index') : ''; ?>
```
Alkarex, line 83: I know that it is not from this PR, so feel free to leave it as it, but `isset($this->extension)` is strange or at least not optimal, and I would rather want a `!empty($this->extension)` if there is a risk that the property is not defined at all, or `$this->extension != null` or similar otherwise.

Alkarex:
By the way, I cannot remember exactly when that changed, but when clicking on a "submit" button in a feed configuration in a slider, the slider closes. This is a bit annoying when more than one parameter needs to be changed. Indeed, some parameters needs to be submitted before applying the next one, for instance updating the feed URL and then refresh the feed.

app/layout/aside_feed.phtml
```diff
@@ -156,7 +156,7 @@
 			$url = _url('subscription', 'feed', 'id', '------', 'get', $get, 'from', $actual_view);
 		}
 		?>
-		<li class="item"><a class="configure open-slider" href="<?= $url ?>"><?= _t('gen.action.manage') ?></a></li>
+		<li class="item"><a class="configure open-slider" href="<?= $url ?>&ajax=1#slider"><?= _t('gen.action.manage') ?></a></li>
```
Alkarex: Isn't it already done in extra.js?

https://github.com/FreshRSS/FreshRSS/blob/880f500bb322522437ae44c94d8bc6b84abedd23/p/scripts/extra.js#L147-L155

In any case, the `&ajax=1` parameter should best be added dynamically in the (existing) JavaScript listener so this link keeps working without JavaScript enabled.

math-GH: fixed

app/views/extension/index.phtml
math-GH: I cannot find an option to add a `#slider` to the URL via `Minz_Request::forward()` and I do not have any idea how to add it.

The `#slider` would keep the slider open after submitting.

Alkarex: Not tested at all, but here is a suggestion https://github.com/FreshRSS/FreshRSS/pull/4416/commits/823392dd1dce83a824a4bfda4ba5a203fefe4f32
I doubt that the anchor is sufficient though, as the ID of the thing being configured in the slider is probably lost

P.S.: You can pass `['#' => 'slider']`  in the parameters of the redirection

Alkarex:
Merging as it is already an improvement.
For a future improvements, opening a feed configuration slider in desktop mode, and validating any option closes the slider

PR #4423, opened 2022-06-20
Better trim base_url

Alkarex:
Avoid usual errors for instance with quotes, especially when provided through Docker / CLI

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -133,10 +133,8 @@ class Minz_Request {
 
 	/**
 	 * Return true if the request is over HTTPS, false otherwise (HTTP)
-	 *
-	 * @return boolean
 	 */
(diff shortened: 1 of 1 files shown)


lib/Minz/Request.php
```diff
@@ -235,13 +224,11 @@ private static function extractPath(): string {
...
-	 * Return the base_url from configuration and add a suffix if given.
-	 *
-	 * @return string base_url with a suffix.
+	 * Return the base_url from configuration
 	 */
-	public static function getBaseUrl() {
+	public static function getBaseUrl(): string {
 		$conf = Minz_Configuration::get('system');
-		$url = rtrim($conf->base_url, '/\\');
+		$url = trim($conf->base_url, ' /\\"');
```
Alkarex, line 231: This is the relevant change

PR #4426, opened 2022-06-21
New option to automatically mark as read gone articles

Alkarex:
Option (off / disabled by default) to automatically and immediately mark as read entries / articles that are no longer provided in their upstream RSS / ATOM / XPath feed

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -655,12 +655,24 @@ class FreshRSS_Feed extends Minz_Model {
 
 	public function keepMaxUnread() {
 		$keepMaxUnread = $this->attributes('keep_max_n_unread');
-		if ($keepMaxUnread == false) {
+		if ($keepMaxUnread === null) {
 			$keepMaxUnread = FreshRSS_Context::$user_conf->mark_when['max_n_unread'];
 		}
 		if ($keepMaxUnread > 0 && $this->nbNotRead(false) + $this->nbPendingNotRead > $keepMaxUnread) {
 			$feedDAO = FreshRSS_Factory::createFeedDao();
 			$feedDAO->keepMaxUnread($this->id(), max(0, $keepMaxUnread - $this->nbPendingNotRead));
+		}
+	}
+
+	/** Applies the *mark as read upon gone* policy, if enabled */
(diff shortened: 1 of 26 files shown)


Alkarex:
image

image

I have slightly reordered a couple of options to have them in the same order on both the global and feed-local pages.

app/Models/Feed.php
```diff
@@ -653,15 +653,37 @@ public function incPendingUnread(int $n = 1) {
...
 	}
 
+	/**
+	 * Remember to call updateCachedValue($id_feed) or updateCachedValues() just after.
+	 * @return int|false the number of lines affected, or false if not applicable
+	 */
 	public function keepMaxUnread() {
 		$keepMaxUnread = $this->attributes('keep_max_n_unread');
-		if ($keepMaxUnread == false) {
+		if ($keepMaxUnread === null) {
```
Alkarex, line 662: Minor bug fixed regarding the *keep max unread* policy

math-GH:
What could be a user story of this feature?

Alkarex:
> What could be a user story of this feature?

* My main own usecase (for work) is to track some job position openings, by RSS or by XPath. When the jobs positions are no longer open, they disappear from the RSS feed and/or the HTML page (through XPath). With this option enabled, only the current job openings are shown in FreshRSS.

* Another usecase I have (private) is to track a Web page with a list of episodes from a TV series. When an episode is no longer available, it disappears from the HTML page (through XPath). With this option enabled, only the still-available episodes are shown in FreshRSS.

* It also allows using FreshRSS in a way similar to RSS-Bridge by re-sharing by RSS the XPath extraction of an HTML page, which stays in sync thanks to this option (I need to implement a new security token mechanism to make that truly relevant).

* It might be used to reduce information overload - one user in my family is only interested in reading the top / most recent news and gets scared away when there are too many unread articles :-P

* Finally - although this is not something I use myself - it is also relevant for an anonymous FreshRSS instance where nobody actively mark articles as read, ensuring that only current news are shown.

I hope that makes sense?

math-GH:
Great. That is very cool and useful.

PR #4434, opened 2022-07-03
Update config.default.php

Tealk:
Closes #4430

- description of default_user adapted
-
-

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/config.default.php
+++ b/config.default.php
@@ -36,7 +36,7 @@ return array(
 	# Example: '<img class="logo" src="https://example.net/Hello.png" alt="Logo Example" /> Hello'
 	'logo_html' => '',
 
-	# Name of the user that has administration rights.
+	# Name of the public user important for anonymous reading..
 	'default_user' => '_',
 
 	# Force users to validate their email address. If `true`, an email with a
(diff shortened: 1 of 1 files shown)


config.default.php
```diff
@@ -36,7 +36,7 @@
 	# Example: '<img class="logo" src="https://example.net/Hello.png" alt="Logo Example" /> Hello'
 	'logo_html' => '',
 
-	# Name of the user that has administration rights.
+	# Name of the public user important for anonymous reading..
```
```suggestion
	# Name of the default user. Also used as the public user for anonymous reading.
```

Tealk: perhaps the term "anonymous" would be better? The variable is also called "allow_anonymous".
`# Name of the anonymous user. Used as the public user for reading.`

or is the public user still used somewhere else?

Alkarex: The default user is used for several things: it cannot be deleted, it is the default for showing e.g. errors when not logged in, etc. One of the features is that this is the one used for anonymous reading

PR #4436, opened 2022-07-06
Use keyword "static"

ColonelMoutarde:
static closure
While methods may be static or not, function has no such alternative : they just can’t be static. Yet, there is one final type of method that may be static : static closures.
Closure are functions that may be stored in a variable : functions may have their name stored in a variable, though. Closure also have the ability to aggregate variables from the context of their creation, for future use. As such, $this is available in a closure that is created inside an object.
See more ... https://www.exakat.io/en/5-usages-of-static-keyword-in-php/

### DIFF
--- a/app/Services/ImportService.php
+++ b/app/Services/ImportService.php
@@ -70,7 +70,7 @@ class FreshRSS_Import_Service {
 		$limits = FreshRSS_Context::$system_conf->limits;
 
 		//Sort with categories first
-		usort($opml_elements, function ($a, $b) {
+		usort($opml_elements, static function ($a, $b) {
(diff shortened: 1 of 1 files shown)


Alkarex:
Fine, and although I am not sure it makes much of a difference in practice, there are several other examples in our codebase with the same case, e.g.:

https://github.com/FreshRSS/FreshRSS/blob/4f32939ec3500d08a2b68cf02cb7a7b49f2b7dee/app/Models/CategoryDAO.php#L265

ColonelMoutarde:
> 

That's true, but every little improvement is a step forward

PR #4437, opened 2022-07-06
Use PDO::PARAM_STR

ColonelMoutarde:
add bindParams


Alkarex:
`PDO::PARAM_STR` is already the default value https://php.net/pdostatement.bindvalue and there are many more places than those three lines where we use this construct.
In general, I like to keep a clean and usable git log / blame whenever possible and those small changes are reducing this aspect.
When relevant, I would rather change such minor details at the same time as a feature change or bug fix.

ColonelMoutarde:
Ok, I understand

PR #4438, opened 2022-07-08
Strict comparaision for boolean and null

ColonelMoutarde:
use strict


app/Controllers/categoryController.php
```diff
@@ -47,7 +47,7 @@ public function createAction() {
 
 			$cat = new FreshRSS_Category($cat_name);
 
-			if ($catDAO->searchByName($cat->name()) != null) {
+			if ($catDAO->searchByName($cat->name()) !== null) {
```
Alkarex, line 50: See https://github.com/FreshRSS/FreshRSS/pull/4439/files#r919511685
Also, it should probably be in the same PR

PR #4439, opened 2022-07-08
Use strict for real control values

ColonelMoutarde:
https://stackoverflow.com/questions/137487/null-vs-false-vs-0-in-php

### DIFF
--- a/app/Controllers/categoryController.php
+++ b/app/Controllers/categoryController.php
@@ -12,7 +12,7 @@ class FreshRSS_category_Controller extends FreshRSS_ActionController {
 	 *
 	 */
 	public function firstAction() {
-		if (!FreshRSS_Auth::hasAccess()) {
+		if (FreshRSS_Auth::hasAccess() === false) {
 			Minz_Error::error(403);
 		}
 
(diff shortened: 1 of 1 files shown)


app/Controllers/categoryController.php
```diff
@@ -41,18 +41,18 @@ public function createAction() {
...
 			$cat_name = trim(Minz_Request::param('new-category', ''));
-			if ($cat_name == '') {
+			if ($cat_name === '') {
 				Minz_Request::bad(_t('feedback.sub.category.no_name'), $url_redirect);
 			}
 
 			$cat = new FreshRSS_Category($cat_name);
 
-			if ($catDAO->searchByName($cat->name()) != null) {
+			if ($catDAO->searchByName($cat->name()) !== null) {
```
Alkarex, line 50: Many of our database functions return false during errors (though not done for `searchByName`, but it should have been done).

app/Controllers/categoryController.php
```diff
@@ -91,7 +91,7 @@ public function updateAction() {
 				Minz_Request::bad(_t('feedback.sub.category.no_name'), $url_redirect);
 			}
 
-			if ($catDAO->searchById($id) == null) {
+			if ($catDAO->searchById($id) === null) {
```
Alkarex, line 94: Same comment than for `searchByName`

app/Controllers/categoryController.php
```diff
@@ -41,18 +41,18 @@ public function createAction() {
...
 			$cat = new FreshRSS_Category($cat_name);
 
-			if ($catDAO->searchByName($cat->name()) != null) {
+			if ($catDAO->searchByName($cat->name()) !== null) {
 				Minz_Request::bad(_t('feedback.sub.category.name_exists'), $url_redirect);
 			}
 
 			$opml_url = checkUrl(Minz_Request::param('opml_url', ''));
-			if ($opml_url != '') {
+			if ($opml_url !== '') {
```
Alkarex, line 55: No, `checkUrl` might return `false`, so this change would introduce a bug.
Please don't blindly change operators. Most have been chosen for a reason ;-)

app/Controllers/categoryController.php
```diff
@@ -41,18 +41,18 @@ public function createAction() {
 			invalidateHttpCache();
 
 			$cat_name = trim(Minz_Request::param('new-category', ''));
-			if ($cat_name == '') {
+			if ($cat_name === '') {
```
Alkarex, line 44: That is the only line that is OK, and I would rather not change such a minor thing except when rewriting this piece of code for another reason (e.g. feature, bug fix) or if really required for a task such as https://github.com/FreshRSS/FreshRSS/issues/4112

PR #4400, opened 2022-06-01
Added Nord based theme

joelchrono12:
Closes #

Changes proposed in this pull request:

- A new theme!-

How to test the feature manually:

1. Pick it from the settings in Display

I am still kinda noob at this so if anything else is needed please let me know

PR #4451, opened 2022-07-17
Fix #4447: Nord theme

### LINKED ISSUE #4447: Nord theme: loads every article
I think there's definetly a problem with this theme. On latest Chrome version on Windows, my tab of FreshRSS uses less than 200Mo of memory. When I switch to this theme, it uses circa 1Go. That's probably why it sometimes don't show the banner.
It is even more visible on a mobile phone, everything is so slow.

EDIT: Got it, the theme is loading EVERY article. I have 14000+ unread, so yeah.

_Originally posted by @Alwaysin in https://github.com/FreshRSS/FreshRSS/issues/4400#issuecomment-1186154075_

math-GH:
Closes #4447 
ref. #4400

- `html` need `height: 100%`
- added it to `template.css`
- removed `html {height: 100%}` from the themes, because it is included in `template.css` now

1. use `Nord theme`
2. scroll in normal view down
3. check, that it does not loead to much next articles

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -4,7 +4,6 @@
 /*============*/
 html, body {
 	background: #171717;
-	height: 100%;
 	color: #999;
 	font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "PingFang SC", "Microsoft YaHei", sans-serif;
 }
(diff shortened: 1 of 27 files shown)


Frenzie:
Which themes have it "added" this way? Any besides Nord?

math-GH:
> Which themes have it "added" this way? Any besides Nord?

Before:
every theme had included `html {height: 100%}`, except Nord theme.

Now:
every theme has included `html {height: 100%}` via `template.css`

Frenzie (approved):
Ok, cool, very safe then. :-)

PR #4452, opened 2022-07-18
Fix: header height

math-GH:
Before:
grafik

After:
grafik

- Set a fixed height to the header
- calculate the height of the `#global`
- default header height is now 85px (some themes have another height)

1. test it with every theme
2. go to normal view and mark all articles as read
3. check the left navigation footer

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -500,7 +500,6 @@ a.btn {
 /*=== Header */
 .header {
 	background: #171717;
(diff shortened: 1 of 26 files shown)


math-GH:
Kudos to @Frenzie who gave an idea (see: https://github.com/FreshRSS/FreshRSS/issues/4447#issuecomment-1186473977)

Alkarex (approved):

PR #4465, opened 2022-07-27
Improved: error page

math-GH:
Before:
grafik

After:
grafik

- the error message is now a real HTML document (add: `<!DOCTYPE><html><header><title><body>`.... `</body></html>`
- error message is now a HTTP 500 error
- error message has now a CSP header
- copied the technical information to our documentation, cut out the text from the error message and added a link to the documentation
- the log file name is now defined as constant

1. go to the logs page
2. manipulate the page attribute (f.e. `page=abc`)
3. check the error message
4. see the 500 error in the browser's development tool

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


lib/lib_rss.php
```diff
@@ -823,25 +823,17 @@ function errorMessageInfo($errorTitle, $error = '') {
 		$details = "<pre>{$details}</pre>";
 	}
 
+	header($_SERVER['SERVER_PROTOCOL'] . ' 500 Internal Server Error', true, 500);
```
```suggestion
	header('HTTP 1.1 500 Internal Server Error', true, 500);
```

Our response is still HTTP 1.1 compliant, even if client is talking HTTP 2.0 or 3.0 and that gets translated if necessary by the Web server and/or proxy anyway

lib/lib_rss.php
```diff
@@ -823,25 +823,17 @@ function errorMessageInfo($errorTitle, $error = '') {
...
-	<li>If using Docker: <code>docker logs -f freshrss</code></li>
-	<li>To check Web server logs on a Linux system using systemd: <code>journalctl -xeu apache2</code>
-	and if you are using php-fpm: <code>journalctl -xeu php-fpm</code></li>
-	<li>Otherwise, Web server logs are typically located in <code>/var/log/apache2/</code> or similar</li>
-	<li>System logs may also contain relevant information in <code>/var/log/syslog</code>, or if using systemd: <code>sudo journalctl -xe</code></li>
-	</ul>
-	<p>More logs can be generated by enabling <code>'environment' => 'development',</code> in <code>./FreshRSS/data/config.php</code></p>
-	<p>Running the feed update script (with the same user and PHP version as your Web server) might provide other hints, e.g.:
-		<code>sudo -u www-data /usr/bin/php ./FreshRSS/app/actualize_script.php</code></p>
+	<hr>
```
```suggestion
	<hr />
```

lib/lib_rss.php
Alkarex: Maybe that line should be moved to the `killApp()` function

math-GH: `killApp()` is not enough.

errorMessageInfo() is used in `kippApp()` and:

https://github.com/FreshRSS/FreshRSS/blob/71ff54378368e41824010cb2766b984e08e0c815/p/i/index.php#L65-L69

So I would keep it as it is

lib/lib_rss.php
```diff
@@ -823,25 +823,17 @@ function errorMessageInfo($errorTitle, $error = '') {
 		$details = "<pre>{$details}</pre>";
 	}
 
+	header('HTTP 1.1 500 Internal Server Error', true, 500);
```
Alkarex: The function `errorMessageInfo()` returns some text so it should avoid emitting headers (side effect). I suggest to move this line in the `killApp()` function.

lib/lib_rss.php
```diff
@@ -823,25 +823,17 @@ function errorMessageInfo($errorTitle, $error = '') {
...
+	header("Content-Security-Policy: default-src 'self'");
```
Alkarex, line 826: Similar to comment just above, except that I am not sure of where to best put that line, so maybe just leave it there for now

lib/lib_rss.php
math-GH: see my comment above

Alkarex: We could `throw new Minz_Exception($error);` in https://github.com/FreshRSS/FreshRSS/blob/edge/p/i/index.php#L65-L69 instead of the two last lines

math-GH: Could you please support here with a change/push?

Alkarex: Something like that https://github.com/FreshRSS/FreshRSS/pull/4465/commits/53ee1d1b6801e72230403ec354b11cbe859e0e2d (not tested)

PR #4490, opened 2022-08-05
Refactor entry-to-GReader API format

Alkarex:
Some code was duplicated and not completely uniform.
Cleaning of related variables and functions (e.g. better types for entries and categories as objects vs. as IDs).
Use-case: I need to call the same GReader-compatible serialization from an extension


app/views/helpers/export/articles.phtml
```diff
@@ -18,51 +18,20 @@ if (empty($this->entryIdsTagNames)) {
...
-		$article['categories'][] = 'user/-/state/com.google/starred';
-	}
-	$tagNames = isset($this->entryIdsTagNames['e_' . $entry->id()]) ? $this->entryIdsTagNames['e_' . $entry->id()] : array();
-	foreach ($tagNames as $tagName) {
-		$article['categories'][] = 'user/-/label/' . $tagName;
+	$feed = $this->feed ?? FreshRSS_CategoryDAO::findFeed($this->categories, $entry->feedId());
+	$entry->_feed($feed);
+
+	if (isset($this->entryIdsTagNames['e_' . $entry->id()])) {
+		$entry->_tags($this->entryIdsTagNames['e_' . $entry->id()]);
```
Alkarex, line 30: This wrongly mixed tags and labels. Regression fixed in https://github.com/FreshRSS/FreshRSS/pull/5626

PR #4493, opened 2022-08-07
Swage theme cleanup

math-GH:
### before
grafik

grafik

grafik

### After
much more beautiful. The login button is styled. Every page has the blue header banner. The form is clear styled now. 

grafik

grafik

grafik

### before
very grey. Grey inputs and selects looks like disabled elements.
grafik

select has not the same height like the input
grafik

### after
highlighted `<legend>`, fresh inputs and selects with a grey border and same height.
(I personally would not highlight the `<legend>` but I wanted to keep the idea of the theme author
grafik

grafik

### before
not fixed header. The width of the search input was not good. The user query button was on the left side, above the read button. The "mark as read" button was not visible
grafik

### after
The header is sticky to the top of the page. The content scrolls below
grafik

### before
grafik

grafik

grafik

### after
The config icon is now on the right side. In the user query menu the config icon is now placed by `position: absolute` instead `float: right`. When hovering over the icon, the background gives feedback with a lighter blue background color. 

grafik

grafik

grafik

### before
Grey text on grey background. The config button of the feeds is weird (it is a white cog icon on grey background). 
grafik
 

White category names of dark background. Dark feed config icon. The empty category "Add category" has now a better visible grey box border. Hover over the cog icons changes the color of the cog to orange.
grafik

### before
the header of the overlay is behind the menu banner. And the width is wider than necessary.
grafik

### after
grafik

### before
grafik

### after
a little bit is now a structure visible
grafik

### before
grafik

### after
grafik

- mostly (S)CSS files
- a little bit JS and phtml

1. see the feature list above

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Have you tried `make fix-all` ?

math-GH:
Ready for review.
I guess I documented the important changes.

To fix/improve this theme was much more time consuming as I planned before.

Alkarex:
Ping @pattems if you are around. Any comment welcome :-)

Alkarex:
This looks like a great work, well done @math-GH 👍🏻

p/scripts/main.js
```diff
@@ -426,10 +426,17 @@ function toggleContent(new_active, old_active, skipping) {
 
 	if (context.sticky_post) {	// Stick the article to the top when opened
 		const prev_article = new_active.previousElementSibling;
-		let new_pos = new_active.offsetParent.offsetTop + new_active.offsetTop;
+		const nav_menu = document.querySelector('.nav_menu');
+		let nav_menu_height = 0;
+
+		if (getComputedStyle(nav_menu).position === 'fixed' || getComputedStyle(nav_menu).position === 'sticky') {
```
Alkarex, line 432: I have not tested yet. Any impact on the other themes?

math-GH: I tested it with Origine. So it should not have an negative impact on other themes.

Let me explain this lines:
The changes in the JS file are needed when the user navigates with the left, top, right arrow navigation through the feed list. An additional factor was needed to calculate the correct position of the opened article.

Alkarex:
Ups, a few conflicts now

pattems:
Thanks for the ping! I really like 99% of these changes; they definitely round out the rough edges. The one thing I will say I _really_ dislike is moving the new article banner; being able to stick it in the sidebar where it made more sense to me was one of the main reasons I made this theme in the first place.

What I'd prefer to see happen would be using @media rules to dump the new article banner out of the sidebar and into the stream **only** if the sidebar's collapsed, but I can always tweak my local copy of the theme if that doesn't sound reasonable. I'd offer to make that change myself but I'm drowning in other projects at the minute and not in a position to set up a dev environment to do proper testing.

math-GH:
Thanks @pattems for your quick feedback and the insight that it is very important for you to have the banner on that place.

I reverted it. Now the banner is back on the place.

Alkarex:
Let's merge and come back to it in another PR if @pattems has more comments

Alwaysin:
I always come late but when you have a low-resolution screen, you can't access the whole admin menu:
chrome_2022-08-18_11-27-33

With the default menu which I think every other theme uses, when you scroll down, you can access the lower elements, but here it stays stuck. For example here I can't access what is below authentication and I have to zoom out (CTRL- minus sign) to see everything:
chrome_2022-08-18_11-30-56

math-GH:
@Alwaysin I cannot reproduce it. I can scroll. When you move with your mouse cursor over the dropdown menu a white scroll bar will appear

grafik

Which browser do you use?

math-GH:
ah ok. now I got it. It does not happen in the mobile view, it happens in the "desktop view". I checked the demo and it happens there too, so this issue is not new with this PR.

To be honest: If you have this small screen (bigger than mobile view, smaller than a normal screen) than we could have issues also on other places.

Please let me know your standard resolution.

Alwaysin:
My screen is 1920x1080, but zoomed at 150%. I'm on a 14" laptop.

ApplicationFrameHost_2022-08-18_21-32-59

From what I can say this is the standard Windows settings.

math-GH:
thanks for this information. I can confirm that 150% is the recommended zoom in Windows10 for a 14" notebook with this resolution.

PR #4497, opened 2022-08-09
Composer require php extensions

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/discussions/4494

### DIFF
--- a/p/ext.php
+++ b/p/ext.php
@@ -60,6 +60,7 @@ function is_valid_path_extension($path, $extensionPath, $isStatic = true) {
 
 	// Static files to serve must be under a `ext_dir/static/` directory.
 	$path_relative_to_ext = substr($path, strlen($real_ext_path) + 1);
+	// @phpstan-ignore-next-line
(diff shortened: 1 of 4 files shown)


Alkarex:
Ping @Aevrin

p/ext.php
```diff
@@ -60,6 +60,7 @@ function is_valid_path_extension($path, $extensionPath, $isStatic = true) {
 
 	// Static files to serve must be under a `ext_dir/static/` directory.
 	$path_relative_to_ext = substr($path, strlen($real_ext_path) + 1);
+	// @phpstan-ignore-next-line
```
Alkarex, line 63: https://github.com/phpstan/phpstan/issues/7764

Alkarex: https://github.com/phpstan/phpstan-src/pull/1576

Alkarex:
Let's merge to make it easier to test. It should not impact the normal way to use FreshRSS

Issue #4526, opened 2022-08-22
[Improvement] i18n: upgrade our translations. 18 languages available. Ready for 1.20.0


math-GH:
Dear community,

[Version 1.20 knocks on the door](https://github.com/FreshRSS/FreshRSS/discussions/4517).

FreshRSS user interface is translated into 18 languages. Not all languages are covered by active developers.

Available languages:
- [ ] CZ: Czech (some phrases need a translation EN -> CZ; 91.8% complete) 🇨🇿
- [x] DE: German (active supported. 100% is translated) 🇩🇪🇦🇹🇨🇭
- [x] EN: British English (active supported. 100% is translated) 🇬🇧
- [x] EN_US: USA English (active supported. 100% is translated) 🇺🇸
- [ ] ES: Spanish (some phrases need a translation EN -> ES; 91.6% is translated) 🇪🇸🇲🇽
- [x] FR: French (active supported. 100% is translated) 🇫🇷
- [ ] HE: Hebrew (some phrases need a translation EN -> HE; 55.3% is translated) 🇮🇱
- [ ] IT: Italian (some phrases need a translation EN -> IT; 69.2% is translated) 🇮🇹
- [ ] JA: Japanese (Some less phrases need a translation EN -> JA; 91.8% is translated)
- [x] KO: Korean (some less phrases need a translation EN -> KO; 91.8% is translated) 🇰🇷🇰🇵
- [x] NL: Dutch (active supported. 91.8% is translated)🇳🇱
- [x] OC: Occitan (some phrases need a translation EN -> OC;  91.8% is translated)🇫🇷🇮🇹
- [x] PL: Polish (some phrases need a translation EN -> PL; 97.4% is translated)🇵🇱
- [ ] PT_BR: Brazil Portuguese  (some phrases need a translation EN -> PT_BR;  91.8% is translated) 🇧🇷
- [ ] RU: Russian (some phrases need a translation EN -> RU;  96.4% is translated) 🇷🇺
- [x] SK: Slovakian (100%) 🇸🇰
- [ ] TR: Turkish (some phrases need a translation EN -> TR;  90.0% is translated) 🇹🇷
- [ ] ZH_CN: Chinese (some last phrases need a translation EN -> ZH_CN;  99.6% is translated)🇨🇳

The language files are PHP files in this folder: https://github.com/FreshRSS/FreshRSS/tree/edge/app/i18n

Do you miss your favorite language? Feel free to add it to our supported languages.

[...]

PR #4528, opened 2022-08-23
[i18n] Add Dutch translation for new strings

Frenzie:
References #4526.


Frenzie:
Whoops, ignore this one. It's something old I still had lying around.

PR #4491, opened 2022-08-05
Fix: Mapco theme cleaning

math-GH:
The Mapco theme is one of the most favorite themes (see #4295), but it has some broke UI

Before (login page)
grafik

After (login page):
The form has now the same grey background as in the configuration. Now it can uses the default form components. The width of the form is a bit wider.
The search button is not blue, when the search is not focused (the search button becomes blue, when the mouse hovers or the focus is inside the search input).
grafik

Before (registration page):
grafik

After (registration page):
same layout like the login page.
The select box has now the same size like the inputs
grafik

Before (configs)
grafik

After (configs)
The select box has now the same size like the inputs
grafik

Before (menu buttons)
grafik

After (menu buttons)
Hover background is set to a darker grey. It is now the same like the other buttons within `.stick`
grafik

Changes proposed in this pull request:

- clean up the (S)CSS
- delete commented out lines
- translated the French comments into English comments
- delete unused icons
- compress icons

How to test the feature manually:
see before/after screenshots

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #4538, opened 2022-08-25
Fix: Ansum theme cleaning

math-GH:
Almost similar to #4491

Most visible change: The login/register form.

Before:
grafik

After:
grafik

- (S)CSS
- icons

use Ansum theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
@Alkarex Wow, that was the fasted PR "review" ;)

Alkarex:
:-D

Alwaysin:
Thanks @math-GH, you make me rediscovering all the themes :+1:

Not really specific to this one but it feels even "odder" with it, is the label page when you don't have any:
chrome_2022-08-25_22-08-42
Maybe there should be some logic to not display the "rename" and "delete" parts when there is no label. Also, in the URL it is called "tag" but then everywhere else it's "labels" but that's another story 😅 

Something else that strikes me as odd is that h1 and h2 titles look very similar.

math-GH:
@Alwaysin let us discuss this in #4545

PR #4548, opened 2022-08-27
lang: update Occitan

Mejans:
I hope it's okay :)

### DIFF
--- a/app/i18n/oc/sub.php
+++ b/app/i18n/oc/sub.php
@@ -75,7 +75,7 @@ return array(
 		'kind' => array(
 			'_' => 'Tipe de font de flux',
 			'html_xpath' => array(
-				'_' => 'HTML + XPath (Web scraping)', // IGNORE
+				'_' => 'HTML + XPath (Web scraping)',	// IGNORE
 				'feed_title' => array(
 					'_' => 'títol del flux',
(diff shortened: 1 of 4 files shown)


app/i18n/oc/sub.php
```diff
@@ -73,51 +73,51 @@
 		'information' => 'Informacions',
 		'keep_min' => 'Nombre minimum d’articles de servar',
 		'kind' => array(
-			'_' => 'Type of feed source',	// TODO
+			'_' => 'Tipe de font de flux',
 			'html_xpath' => array(
-				'_' => 'HTML + XPath (Web scraping)',	// TODO
+				'_' => 'HTML + XPath (Web scraping)',
```
math-GH: Is this already translated?
If yes, please add ` // IGNORE` at the end, so that our pipeline understand, that it is not a translation issue

Mejans: Oh yes, it's best to use the English term here. On my way to change this.

math-GH:
It looks fine now. @Alkarex @Frenzie are you able to check translation? I do not know how similar Occitan is to French

Mejans:
> 

ah ah with any other romance language, you can understand Occitan, even Romanian ^^

Frenzie:
I'm sure it's fine. But yeah, it's comprehensible.

Issue #4545, opened 2022-08-27
Ansum theme: disabled input/select + headline sizes / Label management: Rename+Delete section when 0 labels


math-GH:
(about: Ansum theme)

Not really specific to this one but it feels even "odder" with it, is the label page when you don't have any:
<img width="282" alt="chrome_2022-08-25_22-08-42" src="https://user-images.githubusercontent.com/7189936/186758921-70c036ed-35f2-4219-84cf-10ddfb44ccb9.png">
Maybe there should be some logic to not display the "rename" and "delete" parts when there is no label. Also, in the URL it is called "tag" but then everywhere else it's "labels" but that's another story 😅 

Something else that strikes me as odd is that h1 and h2 titles look very similar.

_Originally posted by @Alwaysin in https://github.com/FreshRSS/FreshRSS/issues/4538#issuecomment-1227718921_

math-GH:
Thanks for your thoughts. Let us divide it in some topics:

1) Ansum theme: disabled input/select are less/not visible 
2) Hide "rename" and "delete" section, when there is no label
3) `h1` and `h2` have the same size
4) "tag" vs "labels"

Have I overseen an idea/comment?

math-GH:
My comments:

to 1): oh, nice finding. That needs an fix. Good catch!

to 2): yes, good idea, I think it is worth to improve it.

to 3): I did not wanted to change the style made by the origin author of Ansum theme. But you are right, here we need an improvement.

PRs will come.

math-GH:
to 4): I guess that it has some historic reasons within FreshRSS. Right now we have the "labels" feature (individually user's tagging feature) and the "Article tags" feature (tags given by the article, cannot be changed by the user). Both are "tags". I guess that the "label" feature was implemented first and got the internal "tags" name, that you still can see in the URL

Alwaysin:
> Have I overseen an idea/comment?

You got everything! Thanks for taking care of this!

PR #4549, opened 2022-08-27
Improved: Mapco/Ansum headline sizes

math-GH:
Ref. https://github.com/FreshRSS/FreshRSS/issues/4545 (improves (3))

- (S)CSS of themes Mapco and Ansum

Before:
H1 and H2 were very big (3rem) and had the same size
grafik

After:
H1's size is 2rem, H2's size is 1.5rem
grafik

1. go to various config pages, f.e. label management or import/export
2. see the headlines

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Ansum/_configuration.scss
+++ b/p/themes/Ansum/_configuration.scss
@@ -20,10 +20,14 @@
 
 	h1, h2 {
 		color: variables.$main-font-color;
-		font-size: 3rem;
+		font-size: 2rem;
 		margin-top: 1.75rem;
 		font-weight: 300;
 		line-height: 1.2em;
+	}
+
+	h2 {
+		font-size: 1.5rem;
 	}
 
 	a[href="./"] {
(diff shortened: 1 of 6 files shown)


Alwaysin:
It feels better now, it was way too big before.

I've just found something else that also doesn't feel right: `<legend>` is capitalized. E.g. in Configuration -> Reading or Sharing.

math-GH:
> I've just found something else that also doesn't feel right: `<legend>` is capitalized. E.g. in Configuration -> Reading or Sharing.

I would keep it as it is right now. The issue is in other themes too. The root cause is that right now we do not have a clear guide line where to use `<legend>` and where to use `<h2>`.

If you want to discuss it more deep and/or want to have a fix in future FreshRSS version please open a new issue.

Alwaysin:
It seems to only  be capitalized on Ansum and Mapco.

PR #4550, opened 2022-08-27
Update about page

Alwaysin:
The link for Gnome icons wasn't working anymore.

Also, request for comments: I'd like to add a target="_blank" to all links on this about page in this same PR, would that be OK? I feel like all external links should open in a new tab.


math-GH:
General: I am not very happy with the icons information on that page. The usage of icons depends on the theme. A not so small amount of icons are already "self-made".

To this PR: Thanks for updating the link. A would prefer a `target="_blank"`

Alkarex:
Let's merge this fix already, and come back to @math-GH 's suggestions in another PR (indeed, I believe we could potentially have some credits in a readme but not there)

Alwaysin:
> Let's merge this fix already, and come back to @math-GH 's suggestions in another PR (indeed, I believe we could potentially have some credits in a readme but not there)

What about adding a section prior to "FreshRSS Contributors" in CREDITS.md ?

Alkarex:
> > Let's merge this fix already, and come back to @math-GH 's suggestions in another PR (indeed, I believe we could potentially have some credits in a readme but not there)
> 
> What about adding a section prior to "FreshRSS Contributors" in CREDITS.md ?

Yes, or we already have a related section in https://github.com/FreshRSS/FreshRSS/blob/edge/README.md#included-libraries

PR #4563, opened 2022-08-30
Fix: Global view - No articles banner

math-GH:
Regression #4042

Before:
No unread feeds + button "show read feeds" active: the yellow "no article" banner is shown
grafik

After:
grafik

- PHTML

1. go global view
2. mark all feeds as read
3. activate "show read articles"
4. see the yellow banner (it will not shown anymore)

Check also: the yellow banner is shown, when no feed is shown

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/index/global.phtml
+++ b/app/views/index/global.phtml
@@ -57,7 +57,7 @@
 		}
 	}
 
-	if ($unreadArticles < 1 && $state_unread == true) {
+	if ($unreadArticles < 1 && $state_unread) {
 		?>
 	<div id="noArticlesToShow" class="prompt alert alert-warn">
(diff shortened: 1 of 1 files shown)


app/views/index/global.phtml
```diff
@@ -54,7 +57,7 @@
 		}
 	}
 
-	if ($unreadArticles < 1) {
+	if ($unreadArticles < 1 && $state_unread == true) {
```
Alkarex:
```suggestion
	if ($unreadArticles < 1 && $state_unread) {
```

Let's keep the boolean syntax simple

math-GH: yes, you are right. It was kept after "playing around"

Issue #3752, opened 2021-08-08
[BUG] SSL error even after disabling verification of SSL security


Poorchop:
**Describe the bug**
Some of my feeds use self-signed certificates. When I set "Verify SSL security" to "No" for these feeds individually, I am still not able to access the feeds via FreshRSS. I get the following error:

```
cURL error 59: failed setting cipher list: DEFAULT@SECLEVEL=1 [https://www.example.com]
```

**To Reproduce**
Some of the feeds are on my own server using a self-certificate that I generated myself using the following:

```
openssl req -x509 -nodes -days 365 -newkey rsa:2048
```

The feed is password-protected so I set the correct username and password in FreshRSS, change "Verify SSL security" to "No," and I try to refresh the feed but it fails with the above error.

**Expected behavior**
FreshRSS should disregard any SSL errors and fetch the feed.

**Environment information (please complete the following information):**
 - OS: OpenBSD 6.9
 - Browser: Firefox 90.0.2
 - FreshRSS version: 1.18.1
 - Database version: PostgreSQL 13.2
 - PHP version: PHP 8.0.9
 - Installation type: Self-hosted server

**Additional context**
 - PHP-FPM runs in a chrooted environment by default in OpenBSD. This led to some php-curl errors for me until I figured out that I need to copy `/etc/resolv.conf` and `/etc/ssl/` to the chroot, but I doubt that this is still related to the issue.
 - These feeds updated fine in another RSS reader that I was using previously when I enabled the option to ignore self-signed certificates.
 - I saw the issues that seem somewhat related like #3029 but I would like to avoid making these security changes global as a workaround. I just want to disable SSL error checking for specific feeds so that the validity of other feeds is still checked.

Frenzie:
Did you check if this has to do with self-signed as opposed to options along the general lines of `-sha256 -newkey rsa:4096`? (Or at the very least  `-sha256 -newkey rsa:2048`.) The error message doesn't seem to suggest self-signed is the issue, imho.

The above suggestion is based on what I quoted in the issue you linked:

[...]

PR #4573, opened 2022-08-31
Expose CURLOPT_IPRESOLVE to UI

### LINKED ISSUE #4560: [Feature] Force ipv4 for individual feeds
Some feedes could not be added, because the following error will appear:
cURL error 35: OpenSSL SSL_connect: Connection reset by peer in connection

Example: Webserver ist hosted by strato.de
Feed: https://www.heise.de/rss/heise-atom.xml will not work.

With the curl-command you can set the option "--ipv4" to download the feed without errors from there.

It would be very helpful if this option could be set in the subscription management in FreshRSS for individual feeds.

Alwaysin:
Closes #4560

- Exposing CURLOPT_RESOLVE to UI to force IPv4 or IPv6 of a feed

Do note that:
- I did not apply this change on ```app/views/subscription/add.phtml``` as I believe you usually don't know in advance that you need to specifically use IPV4 or IPV6 - what do you think? EDIT: I still have included support in ```subscriptionController.php```, damn! 🤦 
- I did not add translation for "IPv4" and "IPv6" as I believe it will be translated the same in all languages
- last but not least I am no developer, it is just a kind of "challenge", I will not be offended if this is rejected. In particular, I have high doubts about the true/false I have used, I think this is not the right way in this case but I took example out of ssl_verify, so it was easier for me. But it really seems to be working. I may have a second look if asked for and if I have time/motivation for it!

1. Select IPv4 for a feed, get the IP of the server and check traffic with ```tcpdump ip && host <ip>```
2. Select IPv6 for a feed, get the IP of the server and check traffic with ```tcpdump ip6 && host <ip>```

Option is located in "Advanced" part, at the end, just after "Verify SSL security"
ApplicationFrameHost_2022-08-31_23-51-51

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -198,6 +198,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				'curl_params' => empty($opts) ? null : $opts,
 			);
 			$attributes['ssl_verify'] = Minz_Request::paramTernary('ssl_verify');
+			$attributes['ipresolve'] = Minz_Request::paramTernary('ipresolve');
 			$timeout = intval(Minz_Request::param('timeout', 0));
 			$attributes['timeout'] = $timeout > 0 ? $timeout : null;
 
--- a/app/views/helpers/feed/update.phtml
+++ b/app/views/helpers/feed/update.phtml
@@ -611,6 +611,17 @@
 		</div>
 
 		<div class="form-group">
+			<label class="group-name" for="ipresolve"><?= _t('sub.feed.ipresolve') ?></label>
+			<div class="group-controls">
+				<select name="ipresolve" id="ipresolve" class="w50">
+					<option value=""<?= $this->feed->attributes('ipresolve') === null ? ' selected="selected"' : '' ?>><?= _t('gen.short.by_default') ?></option>
+					<option value="0"<?= $this->feed->attributes('ipresolve') === false ? ' selected="selected"' : '' ?>>IPv4</option>
+					<option value="1"<?= $this->feed->attributes('ipresolve') === true ? ' selected="selected"' : '' ?>>IPv6</option>
+				</select>
+			</div>
+		</div>
+
+		<div class="form-group">
 			<div class="group-controls">
 				<label class="checkbox" for="clear_cache">
 					<input type="checkbox" name="clear_cache" id="clear_cache" value="1"<?= $this->feed->attributes('clear_cache') ? ' checked="checked"' : '' ?> />
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -248,7 +248,7 @@ function customSimplePie($attributes = array()): SimplePie {
 		}
 	}
 	if (isset($attributes['ipresolve'])) {
-		$curl_options[CURLOPT_IPRESOLVE] = $attributes['ipresolve'] ? CURL_IPRESOLVE_V6 : CURL_IPRESOLVE_V4;
+		$curl_options[CURLOPT_IPRESOLVE] = $attributes['ipresolve'] ? "CURL_IPRESOLVE_V6" : "CURL_IPRESOLVE_V4";
 	}
 	if (!empty($attributes['curl_params']) && is_array($attributes['curl_params'])) {
 		foreach ($attributes['curl_params'] as $co => $v) {
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -247,6 +247,9 @@ function customSimplePie($attributes = array()): SimplePie {
 			$curl_options[CURLOPT_SSL_CIPHER_LIST] = 'DEFAULT@SECLEVEL=1';
 		}
 	}
+	if (isset($attributes['ipresolve'])) {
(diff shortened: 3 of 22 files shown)


app/Controllers/feedController.php
```diff
@@ -198,6 +198,7 @@ public function addAction() {
 				'curl_params' => empty($opts) ? null : $opts,
 			);
 			$attributes['ssl_verify'] = Minz_Request::paramTernary('ssl_verify');
+			$attributes['ipresolve'] = Minz_Request::paramTernary('ipresolve');
```
Alkarex, line 201: Instead of a boolean (true / false), I would prefer a string such as `IPv4` or `IPv6`.

app/views/helpers/feed/update.phtml
```diff
@@ -610,6 +610,17 @@
 			</div>
 		</div>
 
+		<div class="form-group">
+			<label class="group-name" for="ipresolve"><?= _t('sub.feed.ipresolve') ?></label>
+			<div class="group-controls">
+				<select name="ipresolve" id="ipresolve" class="w50">
+					<option value=""<?= $this->feed->attributes('ipresolve') === null ? ' selected="selected"' : '' ?>><?= _t('gen.short.by_default') ?></option>
```
Alkarex, line 617: Please put the value inside `curl_params`. See example a bit higher up with e.g. `CURLOPT_PROXY`

lib/lib_rss.php
```diff
@@ -247,6 +247,9 @@ function customSimplePie($attributes = array()): SimplePie {
 			$curl_options[CURLOPT_SSL_CIPHER_LIST] = 'DEFAULT@SECLEVEL=1';
 		}
 	}
+	if (isset($attributes['ipresolve'])) {
+		$curl_options[CURLOPT_IPRESOLVE] = $attributes['ipresolve'] ? "CURL_IPRESOLVE_V6" : "CURL_IPRESOLVE_V4";
```
Alkarex, line 251: No double quotes around your constants

Alkarex:
It would be good to also add the option when adding a new feed, as it might otherwise fail before the feed is added

Frenzie:
I think "automatic" would be more descriptive than "by default", which in context seems to imply the default is already one of those two just set somewhere else.

Alkarex:
> I think "automatic" would be more descriptive than "by default", which in context seems to imply the default is already one of those two just set somewhere else.

Quick note: We use "By default" at local feed level when the same option can also be set at global level

Alwaysin:
Well, I'm sorry, it's been hours I'm trying to figure things out but I can't seem to be able to make it to work.
That was too much of a task for my skills 😅 
Sorry for the noise!

Frenzie:
@Alkarex 
> Quick note: We use "By default" at local feed level when the same option can also be set at global level

In that case I misunderstood and that particular phrasing makes sense.

On the screenshot I thought it meant:

* By default: whatever cURL chooses, probably IPv6 if available, otherwise IPv4
* IPv4: force IPv4 even if IPv6 is available
* IPv6: half the internet won't work

But whatever is actually intended, I don't think it's clear.

Alkarex:
> Well, I'm sorry, it's been hours I'm trying to figure things out but I can't seem to be able to make it to work. That was too much of a task for my skills 😅 Sorry for the noise!

Let's keep it open and I will give it a shot in the coming days

Alkarex:
When we restart this PR, we should also add an option for HTTP/1.1, HTTP/2, HTTP/3
https://github.com/FreshRSS/FreshRSS/issues/5438

Alkarex:
Related need of cURL options https://github.com/FreshRSS/FreshRSS/issues/3752

benjuade:
Feature needed for reading feeds from websites that may have common misconfigurations like missing intermediate certificates.

Alkarex:
> missing intermediate certificates

While waiting for this PR (help welcome), some misconfigured Web sites might work by using the existing option to disable the SSL check.

benjuade:
> > missing intermediate certificates
> 
> While waiting for this PR (help welcome), some misconfigured Web sites might work by using the existing option to disable the SSL check.
> 

Thanks for pointing that. If your believe this PR may be useful, can you share a list of things missing to approve it? 3years passed by its opening, so things may have changed.

Furthermore, it seems a pure frontend change.

Alkarex:
> Thanks for pointing that. If your believe this PR may be useful, can you share a list of things missing to approve it? 3years passed by its opening, so things may have changed.
> 
> Furthermore, it seems a pure frontend change.

I think this PR needs to be re-done as it branch has been deleted. There are a few review comments higher up to address, and it needs to be tested. Help welcome.

benjuade:
> > Thanks for pointing that. If your believe this PR may be useful, can you share a list of things missing to approve it? 3years passed by its opening, so things may have changed.
> > Furthermore, it seems a pure frontend change.
> 
> I think this PR needs to be re-done as it branch has been deleted. There are a few review comments higher up to address, and it needs to be tested. Help welcome.

Upon better looking at the webapplication, I found that validate SSL is already a curl flag configurable for each feed:
immagine

PR #4583, opened 2022-09-04
API protection limit content size

Alkarex:
We had no limit to the length of an article sent through our GReader API to clients, which was a problem:

The article https://devblogs.microsoft.com/dotnet/performance_improvements_in_net_7/
from the feed https://devblogs.microsoft.com/dotnet/feed/
is about 700KB long, which makes some clients (tested with News+) fail.

We should try other clients to verify that 500KB is a safe enough limit.


app/Models/Entry.php
```diff
@@ -663,6 +663,9 @@ public function toGReader(string $mode = ''): array {
 		$feed = $this->feed();
 		$category = $feed == null ? null : $feed->category();
 
+		// Some clients (tested with News+) would fail if sending too long item content
+		$maxCompatContentLength = 500000;
```
aledeg: I would rather use a class constant. It will work the same but I think it's cleaner since it cannot be modified by the code.

aledeg (changes requested):

app/Models/Entry.php
Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/4583/commits/8b99285c067cd5a64ea3012d0062a05f2ace158f

aledeg (approved):
Usually, it's a good practice to declare constants at the top of the class. But no need to change that.

Alkarex:
> Usually, it's a good practice to declare constants at the top of the class. But no need to change that.

Indeed, but I favoured the local context this time, otherwise it becomes quite far apart.

PR #4586, opened 2022-09-06
New option: Wrap titles

### LINKED ISSUE #3740: [Feature] Display all title on mobile view
Is it possilble to display all title on mobile view ?

-

-

**Additional context**
20210802_125717941_iOS

Thank you

### LINKED ISSUE #4428: [Feature] wrap title text (to see full title) when on mobile web version
I'm looking to switch to FreshRSS from Bazqux, but I've been using Bazqux for 8 years now and I've gotten quite comfortable with how it works on mobile. FreshRSS has a lot of similarities but there's one thing that's bugging me in particular that I can't seem to fix. It's the ability to see the full title text on the web version. With Bazqux's web version, the title of articles will wrap and increase the height of the item's row to capture the entire title. With FreshRSS it is cut off with a ...

Each Item Title should be displayed in full, without "..." cutting off the title text, by wrapping the text for the item, increasing the item's height.

math-GH:
Closes #4428
Closes #3740
Ref. #3997

- add an option to wrap long titles in normal view and global view
- options for big screens ("Desktop") and small screens ("mobile devices")

grafik

1. go to `configuration` -> `reading`
2. enable/disable the checkboxes
3. check the titles in normal view or global view

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] i18n


math-GH:
Right now the text is still hard coded.

I am waiting for your feedback first.

p/themes/base-theme/template.css
```diff
@@ -1066,18 +1071,19 @@ input[type="search"] {
...
-.flux .item.manage,
-.flux .item.link {
+.flux .flux_header .item.manage,
+.flux .flux_header .item.link {
 	width: 40px;
 	text-align: center;
 }
 
-.flux .item.website {
+.flux .flux_header .item.website {
```
Frenzie, line 1161: As an aside, the PR currently has this visual regression:

Frenzie:
I'm not in favor of making this an option, or at least not in this form. Something similar can be done with the number of lines fairly easily along these lines, defaulting to 3 or 4.

```css
.flux .item.title a {
 display: -webkit-box;
 -webkit-line-clamp: 3;
 -webkit-box-orient: vertical;
}
```

Then you can override the `-webkit-line-clamp` attribute, whether with a CSS variable or more traditionally.

It'll then simply grow up to the desired number of lines before being cut off by ellipses. These properties are also supported by Firefox.

p/themes/base-theme/template.css
math-GH: thanks! fixed

math-GH:
Now it is possible to chose how many lines maximal should be displayed (1 till 4)
grafik

Frenzie:
Nice! :+1: 

But I'd like to set it to something a lot higher, preferably just a regular `input type=number`. :-) 4 is a good default maximum but for example a max of about 12 would be good for preventing titles extending much past half a phone screen just in case. That's de facto no ellipsis rather than truly no ellipsis. That being said, I'm not sure I've ever actually seen more than 4.

app/views/configure/reading.phtml
```diff
@@ -68,6 +68,27 @@
 				</div>
 			</div>
 
+			<div class="form-group">
+				<label class="group-name" for="wrapTitle">Wrap long titles in Normal View and Global view</label>
```
Frenzie, line 72: Something like clip or cut off, not wrap.

Frenzie: Hm, although on second thought it'll wrap too… but wrap is a bit technical.

Use multiple lines for… ?

math-GH:
> But I'd like to set it to something a lot higher, preferably just a regular `input type=number`. :-) 4 is a good default maximum but for example a max of about 12 would be good for preventing titles extending much past half a phone screen just in case. That's de facto no ellipsis rather than truly no ellipsis. That being said, I'm not sure I've ever actually seen more than 4.

Let's wait here for a feature request in future. It is still a title, not the whole article ;)

app/views/configure/reading.phtml
math-GH: Would it be better just to have `titles in Normal View and Global view`?

Then we need a good word for "no wrap" as first select list item

Frenzie:
Yes, that's my feature request so I can stop using my custom CSS rather than having to keep it or replace it with another. ;-)

Here's an extremely rare example of more than 4 lines of title. I don't find it bothersome. Perhaps I would if it were more common. I used to run a max-height: 40vh on it just in case, but I don't think it's ever been reached by anything real.

### Laptop


### Mobile


### Artificial gigantic title cut off at 40vh (in this case ~12 lines)

app/views/configure/reading.phtml
Frenzie: I'd only looked at the CSS in detail; I misinterpreted "no wrapping" to mean "no clipping". One line doesn't make a lick of sense to me, hence the misunderstanding.

math-GH:
I understand. That is weird. Anything that has a user input is impossible (?) to handle it with CSS. For each number I have to prepare a CSS class, because of  `-webkit-line-clamp: 3;`

I could add:
- 5 lines
- unlimited (show the full title). Here you can do you individual height cut off with Custom CSS, but you do not need to manage all the other stuff in Custom CSS anymore (hover with the mouse over the line, vertical alignment of the icons etc.). And you can configure it individually for mobile and desktop view.

app/views/configure/reading.phtml
math-GH: I am not sure if I understand you correct.
Could you please give your idea of the left text (`wrap long titles in Normal View and Global view`) and of the first select option (`no wrap`)?

Frenzie:
D'oh, my bad, I skimmed too quickly. I assumed things and when you assume…

Anyway, I figured you'd be using `Minz_View::appendStyle` to override it on the page. But you can actually also do what you were doing here.

```css
.flux .item.title a {
 --clamp: 3;
 -webkit-line-clamp: var(--clamp);
}
```

And then on the element to override:
```php
style="--clamp: <?php FreshRSS_Context::$user_conf->clamp_title_mobile ?>"
```

math-GH:
> And then on the element to override:
> 
> ```
> style="--clamp: <?php FreshRSS_Context::$user_conf->clamp_title_mobile ?>"
> ```

inline style is not allowed because of the Content Security Policy

Frenzie:
You expressed a preference for doing it that particular way. As I said I think it's much more logical to me to simply use PHP to set something like:

```css
.flux  .item.title a {
	-webkit-line-clamp: $clamp_value;
}
```

math-GH:
> You expressed a preference for doing it that particular way. As I said I think it's much more logical to me to simply use PHP to set something like:
> 
> ```css
> .flux  .item.title a {
> 	-webkit-line-clamp: $clamp_value;
> }
> ```

I have no idea how to transfer the value from PHP to CSS, when I cannot use inline CSS

Frenzie:
I can probably look into it next week if you like, but I think at least one potential answer is self-evident from the question. Put another way, see CustomCSS.

math-GH:
I lost control over the merging with main branch (about 100 commits and a lot of conflicts).

I created a new PR: #4857

PR #4532, opened 2022-08-24
Shortcuts: auto-load at least a few articles

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3974

Issue #3974, opened 2021-11-12
[BUG] Jumping to next article with shortcuts does not load more articles at the bottom


rom-1:
**Describe the bug**
When jumping to the next post/article using a keyboard shortcut works only until "the end of the page" is reached, new articles are not loaded even when there are ones in queue. Pressing the keyboard shortcut in this state, FreshRSS jumps to the next feed, leaving unread articles in the previous one.

In the setttings I have enabled the option: "Load more articles at the bottom of the page"

**To Reproduce**
1. Have a feed with about 80 posts // with a lot of content. (I am using CSS selectors to load content of the article)
2. Press keyboard shortcut for "Open the next (unread) article"
3. Press keyboard shortcut several times, watch the scoll bar, it will scroll down until the bottom of the page.
4. When at the end of the scroll bar pressing the keyboard shortcut again will switch to the next feed. There are still about 30 unread articles in the feed menu, but no content is loaded at the bottom of the page.

If there are no additional feeds, the view will simply stop, and won't switch anywhere though there are additonal articles available. Simply by using the mouse wheel to scroll up a little bit, then scroll down a bit, suddenly loads additional articles at the bottom of the page.

**Expected behavior**
There should be content loaded at the bottom of the page when using the keyboard shortcut to scroll down.

**Screenshots**
Not applicable

**Environment information (please complete the following information):**
- Device: PC
- OS: Windows 10
- Browser: Chrome 95
- FreshRSS version: 1.19.0-dev
- Database version: not relevant
- PHP version: not relevant]
- Installation type: selfhosted? shared hoster?

**Additional context**
No context

math-GH:
@rom-1 is it still an issue? 
You found this issue on 1.19.0-dev. Now we are on 1.20.0-dev and I cannot reproduce it.

rom-1:
@math-GH Hey! Yes, this issue does still persist.

[See demonstration video here](https://user-images.githubusercontent.com/37835829/185823424-83c8c257-5ad8-4bee-8948-d76db28b205a.mp4). I have this feed from "heise online" which contains 128 articles. If I use the "ENTER" Button to jump to the next unread article, you can see how the scroll bar on the right reaches its bottom very soon but there should still be some additional content below. Instead, FreshRSS jumps to the next feed/category when pressing "Enter". This does not happen if I use my mouse wheel to scroll down. Using the mouse, addidiontal content is loaded on the fly.

[...]

PR #4593, opened 2022-09-09
Fix auto-loading of more articles

Alkarex:
Fix regression from https://github.com/FreshRSS/FreshRSS/pull/4532
https://github.com/FreshRSS/FreshRSS/issues/3974


Frenzie (approved):
Btw, what happens here if no more can be loaded?

aledeg (approved):

Alkarex:
> Btw, what happens here if no more can be loaded?

Either more are loaded thanks to the position / pixels rule, or we have reached the end of the list

PR #4596, opened 2022-09-10
improved: CSS: clean .horizontal-list

math-GH:
Ref. #4192
Ref. #4181

Both PR reduced the usage of `.horizontal-list`

Now only `app\views\helpers\index\normal\entry_bottom.phtml` and `app\views\helpers\index\normal\entry_header.phtml` use `.horizontal-list`.

So it makes sense to clean the CSS.

- (S)CSS

just Code

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Ansum/ansum.rtl.css
+++ b/p/themes/Ansum/ansum.rtl.css
@@ -221,11 +221,7 @@ form th {
 /*=== Forms */
 /*=== Horizontal-list */
 .horizontal-list {
-	margin: 0;
 	padding: 0.1rem 0;
-}
-.horizontal-list .item {
-	vertical-align: middle;
 }
 .horizontal-list .item:first-child {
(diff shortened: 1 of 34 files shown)


p/themes/Ansum/ansum.rtl.css
```diff
@@ -1137,9 +1133,6 @@ main.prompt {
 .post form {
 	margin: 1rem 0;
 }
-.post form .horizontal-list {
-	margin-bottom: 0.5rem;
-}
```
Frenzie, line 1142: I would be somewhat careful with a few details like this one as a rule of thumb. It's not inconceivable that an extension expects to be able to use it.

But in this case it shouldn't be very harmful. ^_^

Frenzie (approved):

PR #4598, opened 2022-09-10
Add documentation to work with feed snapshots

aledeg:

- Add documentation for feed testing and debugging

1. N/A

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/docs/en/developers/03_Running_tests.md
+++ b/docs/en/developers/03_Running_tests.md
@@ -34,3 +34,43 @@ They are performed with [GitHub Actions](https://github.com/FreshRSS/FreshRSS/ac
 This ensures your code will not introduce some kind of regression. We will not merge a PR if tests fail so we will ask you to fix any bugs before reviewing your code.
 
 If you are interested, you can take a look at [the configuration file](https://github.com/FreshRSS/FreshRSS/blob/edge/.github/workflows/tests.yml).
+
+## Using feed snapshots
+
+As feed data is volatile, it’s better to work with snapshots when debugging some issues.
+You can find the description to retrieve a snapshot [here](06_Reporting_Bugs.md#how-to-provide-feed-data).
+
+To serve those snapshots, you can use a mock server.
+Here we will demonstrate how to work with [WireMock](https://wiremock.org/) but other solutions exist.
+Here are the steps to start using the WireMock mock server:
+
+1. Go to the mock server home folder.
+If you do not have one, you need to create one.
+1. Inside the mock server home folder, create the ___file_ and _mappings_ folders.
+1. Copy or move your snapshots in the ___file_ folder.
+1. Create the _feed.json_ file in the _mappings_ folder with the following content:
+	```js
+	{
+		"request": {
+			"method": "GET",
+			"urlPathPattern": "/.*"
+		},
+		"response": {
+			"status": 200,
+			"bodyFileName": "{{request.pathSegments.[0]}}",
+			"transformers": ["response-template"],
+			"headers": {
+				"Content-Type": "application/rss+xml"
+			}
+		}
+	}
+	```
+1. Launch the containerized server with the following command:
+	```bash
+	# <PORT> is the port used on the host to communicate with the server
+	# <NETWORK> is the name of the docker network used (by default, it’s freshrss-network)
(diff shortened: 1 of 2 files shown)


docs/en/developers/06_Reporting_Bugs.md
```diff
@@ -64,3 +66,17 @@ Remember to give the following information if you know it:
...
+
+## How to provide feed data
+
+If you need us to investigate a feed problem, it will be easier if you provide a snapshot of the feed data.
+To do that, you can launch the following command:
+
+```bash
+wget <feed url> -O output.rss
+```
+As github does not allow to attach files to an issue, it’s better to create a public gist containing the file.
```
Alkarex: But it is possible to attach files to an issue ;-)

Frenzie: In fact it's preferable because the chance of it working a few months or sometimes even a few days from now is significantly higher if it's attached to GH.

But for a file like output.rss you'll either need to rename it to output.rss.txt or zip it. (Or rename it to output.rss.zip but that would be lacking in clarity. ^_^)

Frenzie: Pardon, I'd overlooked the gist part which is also part of GH but what I said is still true. Gists disappear if users do some cleanup or if their account is deleted, and I'm pretty sure there are at least two distinct issues in this very repo where a gist has disappeared, while an attached (zipped if necessary) file would still be there.

docs/en/developers/03_Running_tests.md
```diff
@@ -34,3 +34,43 @@ They are performed with [GitHub Actions](https://github.com/FreshRSS/FreshRSS/ac
...
 
 If you are interested, you can take a look at [the configuration file](https://github.com/FreshRSS/FreshRSS/blob/edge/.github/workflows/tests.yml).
+
+## Using feed snapshots
+
+As feed data is volatile, it’s better to work with snapshots when debugging some issues.
+You can find the description to retrieve a snapshot [here](06_Reporting_Bugs.md#how-to-provide-feed-data).
+
+To serve those snapshots, you can use a mock server.
+Here we will demonstrate how to work with [WireMock](https://wiremock.org/) but other solutions exist.
```
Frenzie, line 44: Interesting. For the record, I'd intuitively probably be inclined to do something like:

```
docker run -p 80:80 -v $(pwd):/usr/share/nginx/html nginx
```

Which is basically the same thing but without the WireMock part. ;-)

marien-probesys: Or `python3 -m http.server 8080` :)

marienfressinaud: Or `python3 -m http.server 8080` :)

docs/en/developers/06_Reporting_Bugs.md
aledeg: I couldn't find a way to do that but I agree with you. It's better to keep it in the issue. Could you tell me where I can find info on that so I can change that part?

docs/en/developers/03_Running_tests.md
aledeg: I haven't though on that but those are valid suggestions. Do you think I should add them or completely disregard the wiremock part?

docs/en/developers/06_Reporting_Bugs.md
Frenzie: You can either click on the bottom where it says "attach files" or just drag & drop into the textbox. ^_^

aledeg: That's weird. I cannot find the attach files section when using github. But I tried the drag and drop method and it works. I will update this PR and get rid of the gist section of the documentation.

aledeg: Thank you!

aledeg: I've made the modification.

docs/en/developers/03_Running_tests.md
Frenzie: I do believe both suggestions above are an order of magnitude simpler to use compared to messing with JSON for the vast majority of use cases.

* Name your file lala.rss → content-type: application/rss+xml
* Name it asdf.atom → content-type: application/atom+xml

So you can also easily test incorrect mimetypes with e.g. .txt and .html.

aledeg: In that case, I will remove that section and only mention few ways one can achieve that.

PR #4607, opened 2022-09-12
npm scripts for windows

math-GH:
Ref https://github.com/FreshRSS/FreshRSS/discussions/4602

The npm scripts (package.json) does not work on a Windows 10 machine.

So I started to investigate how to run it on Windows.

Here is my solution draft. Not complete yet. See it as proof of concept.

### DIFF
--- a/package-windows.json
+++ b/package-windows.json
@@ -1,0 +1,46 @@
+{
+  "name": "freshrss",
+  "description": "A free, self-hostable aggregator",
+  "homepage": "https://freshrss.org/",
+  "readmeFilename": "README.md",
+  "bugs": {
+    "url": "https://github.com/FreshRSS/FreshRSS/issues"
+  },
+  "keywords": [
+    "news",
+    "aggregator",
+    "RSS",
+    "Atom",
+    "WebSub"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/FreshRSS/FreshRSS.git"
+  },
+  "license": "AGPL-3.0",
+  "scripts": {
+    "eslint": "eslint --ext .js .",
+    "eslint_fix": "eslint --fix --ext .js .",
+    "markdownlint": "markdownlint '**/*.md'",
+    "markdownlint_fix": "markdownlint --fix '**/*.md'",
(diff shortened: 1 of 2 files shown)


math-GH:
A little explanation: 

the package.json has some unescaped `'` character that CMD/Powershell does not like.

So I created a copy of `package.json`  -> `package-windows.json` that improves it for the usage on Windows.

`runscripts.bat` renames the `package.json` to another temp name and `package-windows.json` -> `package.json` and runs then the npm scripts.
At the end it renames it back.

It works fine:
grafik

package-windows.json
```diff
@@ -0,0 +1,46 @@
...
+    "RSS",
+    "Atom",
+    "WebSub"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/FreshRSS/FreshRSS.git"
+  },
+  "license": "AGPL-3.0",
+  "scripts": {
```
Alkarex, line 21: instead of a new file, you could have just the scripts that require a Windows-specific command, e.g. `"stylelint_win"`

Alkarex: And maybe we can find a syntax that works on both platforms

Frenzie: I suspect the syntax as written is already a syntax that works on both platforms. I probably just used native shell instead because it's more intuitive. Cf. https://www.npmjs.com/package/fast-glob which is used by Stylelint. (I don't have npm set up on this laptop.)

Frenzie: Put another way, just change that single line in the original file, introduce an error and see if the CI catches it.

Alkarex: No, the syntax without the quotes does not work in Bash, but another syntax might be possible

package-windows.json
```diff
@@ -0,0 +1,46 @@
...
+    "eslint": "eslint --ext .js .",
+    "eslint_fix": "eslint --fix --ext .js .",
+    "markdownlint": "markdownlint '**/*.md'",
+    "markdownlint_fix": "markdownlint --fix '**/*.md'",
+    "rtlcss": "rtlcss -d p/themes/ && find p/themes/ -type f -name '*.rtl.rtl.css' -delete",
```
Alkarex, line 26: I am afraid other things would not work on DOS-style command-line anyway, such as this command using `find`. But all that works fine on Windows with WSL.

Issue #3057, opened 2020-06-14
Display default theme on login page


aledeg:
At the moment, the theme displayed on the login page is the one from the default user. This is fine until there are many users using an instance and the default user decide to change his theme.
It can worry some users because they don't recognize the page they are used to.

aledeg:
Apparently, this is now working with the latest version. I don't know if something changed recently.
Do you have the same behavior on your end? If so, we could close that issue.

PR #4620, opened 2022-09-18
Force default user before rendering login page

aledeg:
Closes #3706, #3057

Changes proposed in this pull request:

- Force the use of the default user before rendering the login page

How to test the feature manually:

1. Add a user with a theme different from the installation theme.
2. Try to login with a wrong password.
3. Validate that the displayed theme is the installation theme and not the user theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #4621, opened 2022-09-18
Fix login page when user does not exist

### LINKED ISSUE #3706: [BUG] login screen: UI changed after finding out an account
I am not sure if it could be a security issue (but is not urgent).

the default user uses the default template.
grafik

If the username is correct but the password is incorrect than the theme of this user is used for the login screen:
grafik

So it is a clear sign that you find out a valid user name. Now you need "just" to try out the password.

**To Reproduce**
1. Go to login screen
2. try out an random username and a random password: the UI will not change after submitting
3. use a valid username (this user does not use the default thema) and a random password: after submitting the UI will change to the user's theme

**Expected behavior**
keep the default UI

 - Browser: Firefox 90
 - FreshRSS version: 1.18.1

aledeg:
Closes #3706, #3057

- Force the use of the default user before rendering the login page

1. Add a user with a theme different from the installation theme.
2. Try to login with a wrong password.
3. Validate that the displayed theme is the installation theme and not the user theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -72,7 +72,7 @@ class FreshRSS_auth_Controller extends FreshRSS_ActionController {
 		}
 
 		$auth_type = FreshRSS_Context::$system_conf->auth_type;
-		FreshRSS_Context::initUser('_');
+		FreshRSS_Context::initUser('_', false);
 		switch ($auth_type) {
 			case 'form':
 				Minz_Request::forward(array('c' => 'auth', 'a' => 'formLogin'));


aledeg:
@math-GH Could you give this a try?

aledeg:
@marienfressinaud @Alkarex Do you remember the idea behind the use of the user `_`?

math-GH:
> @math-GH Could you give this a try?

No error anymore. I can login.

But, why is the login page as Origine theme and not the theme of the default user?
Right now: My default user has enabled the alternative dark theme. When a guest (not logged in) opens FreshRSS, then he/she sees everything in alternative black theme, but the login page is in Origine theme.

aledeg:
You can see my concern when I opened #3057

math-GH:
> You can see my concern when I opened #3057

I cannot agree with this. The `default user` is a very special user: this user has admin rights plus the theme's configuration for the guests.

Right now it is maximum confusing:
default user use swage theme.
User "user123" uses Ansum theme.

User123 opens FreshRSS and sees swage theme. Click on login link. Sees Origine theme. After login: sees Ansum theme.

It means, that no theme has the chance to configure the login page.

Please use the default user's theme for login page too.

aledeg:
Even if we have a default user, I don't think that we should use its configuration to be the default configuration. We have a file to handle such thing and we should use that instead.
My solution is naive and we should leverage that default configuration instead.

I made a quick test to see how the custom file is handled and it does not work at the moment.
I'll make it work before submitting more code here.

Alkarex:
Let's merge to fix the regression from https://github.com/FreshRSS/FreshRSS/pull/4620 , which risks being quite common, and continue in another PR, please @aledeg

marienfressinaud:
> Do you remember the idea behind the use of the user `_`?

I don't :grimacing: 

> Even if we have a default user, I don't think that we should use its configuration to be the default configuration.

I agree with this. As admin of a public instance with opened registrations and a custom home page (https://flus.io), I cannot change my own theme or it would change the theme of the home and login pages. When I want to verify something (e.g. if a theme is broken), I have to login with a different user.

My opinion is that a default theme should be used for the users not connected (maybe also in "anonymous" mode). It can either be in the configuration of the `_` user, or in the global configuration (something like `default_theme`; my preference but I'm fine with the other solution).

PR #4636, opened 2022-09-22
Improved: Origine compact inherit from origine

math-GH:
Goal: Improve the maintenance of `Origine compact` theme, that is just a copy of "Origine" theme plus some little changes.

It reduce the amount of work, when a CSS changing is neccessary.

Changes proposed in this pull request:

- manifest.json: `Origine compact` theme inherits from `Origine`
- deleted: all duplicated CSS lines

How to test the feature manually:

1. use the `Origine compact` theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #4641, opened 2022-09-25
Improved: template.css (colors)

math-GH:
Similar to #4636.

All colors are now collected within CSS variables in `template.css`
It gives a better overview of used colors and adjusting in.
All variables starts with `--template` to mark them as template.css variables. They could be used by the themes to easily adjust them instead of copy&paste each CSS line.

- CSS
- little improvements for the default configs.

1. check the other themes
2. hopefully nothing / not much changed there

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Ansum/ansum.css
+++ b/p/themes/Ansum/ansum.css
@@ -1350,6 +1350,7 @@ main.prompt {
 /*============*/
 html, body {
 	background: #f5f0ec;
+	color: #161a38;
 	font-family: "lato", "Helvetica", "Arial", sans-serif;
 	font-size: 0.875rem;
 }
--- a/p/themes/Ansum/ansum.rtl.css
+++ b/p/themes/Ansum/ansum.rtl.css
@@ -1350,6 +1350,7 @@ main.prompt {
 /*============*/
 html, body {
 	background: #f5f0ec;
+	color: #161a38;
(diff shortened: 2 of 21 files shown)


math-GH:
when this PR is merged I will work further on template.css. I will improve the font-size and line-height too (switch from `px` to `rem`, from `em` to `rem`)

Alkarex:
Very minor and I do not really mind, but maybe consider using `frss-` as prefix (shorter and more specific) - that is the same one used for our XML Namespace

math-GH:
> Very minor and I do not really mind, but maybe consider using `frss-` as prefix (shorter and more specific) - that is the same one used for our XML Namespace

I thought the same before: `--frss-variable-name` is not very specific to `template.css`. It could be also a variable of `adark.css`.

I am not happy to use `template` because it is not a template. It is just the base-theme. But the name of the CSS file is `template`.

If we would rename `template.css` to `frss.css` then I would be very happy and `--frss-variable-name` would fit perfect.

Any idea?

Alkarex:
> If we would rename `template.css` to `frss.css` then I would be very happy and `--frss-variable-name` would fit perfect.

Fine with me

math-GH:
Rename PR: #4644

Issue #4585, opened 2022-09-06
Alternative Dark theme: not best buttons color, not fully optimized for mobile


math-GH:
> Question: `Which one do you use?` and `Why do you like it?`
> 
Alternative Dark
best dark version but sadly not best buttons color and also not fully optimized for mobile.

_Originally posted by @beerisgood in https://github.com/FreshRSS/FreshRSS/issues/4295#issuecomment-1237418148_

math-GH:
Hey @beerisgood,
thank you for your additional feedback. Let's use this ticket here to gain more insights.

What do you mean with "best buttons"?
What do you mean with "not fully optimized for mobile"?

Which FreshRSS version do you use?

beerisgood:
layout for mobile isn't that nice:
![telegram-cloud-photo-size-4-5816864034098625245-y](https://user-images.githubusercontent.com/9625159/188702502-3545a7b2-a7f4-4387-a4e6-eebe80a84b2e.jpg)

clicking on folder icon sometimes results in "mark all as read" button below. the other icons are also very close together.
i would also move the icons themselves.

color of the button does not read well with black theme:
![image](https://user-images.githubusercontent.com/9625159/188703051-1aef2787-343d-4e17-a513-f977b2db0328.png)
(this is on desktop)

FreshRSS 1.19.2

math-GH:
Thanks for this insights.

I cannot reproduce the little stairs in the mobile view.

There is some work left to improve the mobile layout in general.

The "buttons" are select lists. Your operating system or browser theme change it to a very "special" design. 

Which operating system and browser do you use?

beerisgood:
> Which operating system and browser do you use?
>

latest stable Safari in MacOS & iOS

Frenzie:
I'll submit a quick fix.

math-GH:
I will keep it open for the other thing.

math-GH:
I checked it now with Safari and can confirm the "stairs" in the menu. Will try to fix it

Poorchop:
I liked the alternative dark theme the best when I first started using FreshRSS but I had some issues with it. There were some bugs, some of which I think I tried to address, but there are also some cosmetic changes that I personally preferred. I use the custom CSS extension with my own additions:

```css
@media (min-width: 841px) {
	.header {
		height: 40px;
	}

	.header > .item {
		padding: 0;
	}

	.header > .item.title h1 {
		font-size: 1em;
		margin: 0;
	}

[...]

PR #4643, opened 2022-09-26
fixed: Safari issue with .stick dropdowns

math-GH:
ref. #4585 

before: see bug issue
after: 
(no stairs anymore)
grafik

- css: do not use anymore the `font-size: 0` trick. Instead: use `display: inline-flex`
- all themes: cleaned the technical code from themes and put it into the template.css

1. use an Apple device with Safari browser
2. see the navigation menu bar

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


beerisgood:
👍

PR #3859, opened 2021-09-26
Update README.md

martinrotter:
Sync in RSS Guard is now much faster, better (optional) syncing algorithm was implemented

Closes #

Changes proposed in this pull request:

-
-
-

How to test the feature manually:

1.
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

Issue #3417, opened 2021-02-04
Sync with Vienna is not working


fabiosirna:
Hi,
I've configured FreshRSS on a docker container and via web works really well.
However, I'd like to use a desktop reader like Vienna but I'm not able to sync it via the greader API.

A test I did via terminal:

```
#curl 'http://192.168.1.2:60040/api/greader.php/accounts/ClientLogin?Email=fabio&Passwd=cijvax2pExrefotcah'

SID=fabio/975db1b731c02f74270740db330222c77790e022
LSID=null
Auth=fabio/975db1b731c02f74270740db330222c77790e022


#curl -s -H "Authorization:GoogleLogin auth=fabio/975db1b731c02f74270740db330222c77790e022" \
  'http://192.168.1.2:60040/api/greader.php/reader/api/0/stream/contents/reading-list' | jq .
```

After the second command I'm able to get a JSON with a list.

I'm using the same parameters on Vienna preferences panel:

<img width="512" alt="Screenshot 2021-02-04 at 00 52 07" src="https://user-images.githubusercontent.com/2249234/106826551-6f5e4800-6687-11eb-82c2-8657156ca26f.png">

For the password I tried everything:
- user pass
- mobile token
- some token you can see in the terminal command

I always get this:

<img width="272" alt="Screenshot 2021-02-04 at 01 24 29" src="https://user-images.githubusercontent.com/2249234/106826779-e398eb80-6687-11eb-851e-6939ce224792.png">

and if I remove the http I get an SSL error like this:

<img width="271" alt="Screenshot 2021-02-04 at 01 26 24" src="https://user-images.githubusercontent.com/2249234/106826848-04f9d780-6688-11eb-9f64-a93dec0725f1.png">

Is there anything I can do to get it working?

I'm looking also for suggestions about a IOS mobile app that can work with FreshRSS (the web ui via mobile doesn't have a good ux).

Thanks!

aledeg:
Have you checked #2573 and #2091?

Alkarex:
In addition to @aledeg 's suggestions, please check your Web server logs, and `./FreshRSS/data/users/_/log_api.txt`
Please give also some details about your setup (OS, Web server, versions, proxy, type of FreshRSS installation...)

math-GH:
@fabiosirna 
Is it still an issue?
Maybe you could help with your experiences and help to update our apps list on https://github.com/FreshRSS/FreshRSS/blob/edge/README.md

[...]

PR #4650, opened 2022-09-27
Update NewsFlash app entry

jangernert:

- updated NewsFlash app feature matrix
- GReader API is now implemented
- Labels (Tags) are supported

I wasn't sure about two features:

- there is an offline mode, but it not remember read articles and push status updates once back online
- you can add/remove/edit feeds etc., but moving them around via drag & drop is still missing from the Gtk 4 port

So I left them unchecked. Also whats the metric for the "Fast Sync" stars?

Pull request checklist:

- [x] clear commit messages
- [ ] code manually tested
- [x] documentation updated


Alkarex:
Thanks!

> * GReader API is now implemented

Let's put it before Fever then, as it is a more powerful and efficient API

> * Labels (Tags) are supported

Great! That means adding and syncing custom user-defined labels, right? (and not the tags provided by the upstream feed)

> Also whats the metric for the "Fast Sync" stars?

I have written a vague attempt at an explanation on
https://github.com/FreshRSS/FreshRSS/issues/3417#issuecomment-1252556926
Does that help?

I will try to give a try to the latest NewsFlash. I cannot get Flatpak to work on my machine but I will try on another machine another day. Would you happen to have an Ubuntu-compatible binary?

README.md
```diff
@@ -122,7 +122,7 @@ and [Fever API](https://freshrss.github.io/FreshRSS/en/users/06_Fever_API.html)
...
 | [FeedReader](https://jangernert.github.io/FeedReader/)                                | GNU/Linux   | [✔️](https://jangernert.github.io/FeedReader/)                | 2020                   | GReader          | ✔️            | ⭐⭐      | ➖                             | ✔️                  | ✔️         | ➖     | ✔️       | ✔️           |
-| [NewsFlash](https://gitlab.com/news-flash/news_flash_gtk)                             | GNU/Linux   | [✔️](https://gitlab.com/news-flash/news_flash_gtk)            | ✔️✔️                   | Fever, (GReader) | ➖            | ⭐⭐      | ✔️                           | ✔️                | ✔️       | ➖    | ➖      | ➖          |
+| [NewsFlash](https://gitlab.com/news-flash/news_flash_gtk)                             | GNU/Linux   | [✔️](https://gitlab.com/news-flash/news_flash_gtk)            | ✔️✔️                   | Fever, GReader | ➖            | ⭐⭐      | ✔️                           | ✔️                | ✔️       | ✔️    | ➖      | ➖          |
```
Alkarex: And when it is ready, please put the same line in the French readme https://github.com/FreshRSS/FreshRSS/blob/edge/README.fr.md#apis-et-applications-natives

jangernert:
> Great! That means adding and syncing custom user-defined labels, right? (and not the tags provided by the upstream feed)

Yes, I think so. I'm not aware of any other labels.

Screenshot from 2022-09-28 17-03-19

> 5k+ unread articles and 1k+ favourites, using < 10 HTTP requests.

I don't have anywhere near that number of articles and favorites in my demo account (provided by a user of NewsFlash). But for a normal sync with the above mentioned numbers we're doing 9 API calls:
- subscription_list
- tag_list
- all unread articles in batches of 1k (so 5 calls for 5k)
- all marked/starred articles in batches of 1k (so 1 call for 1k)
- 100 latest read articles (1 call)

https://gitlab.com/news-flash/news_flash/-/blob/master/src/feed_api_implementations/freshrss/mod.rs#L156

> I will try to give a try to the latest NewsFlash. I cannot get Flatpak to work on my machine but I will try on another machine another day. Would you happen to have an Ubuntu-compatible binary?

No, I myself use fedora. But flatpak should be easy to set up according to the flathub guide: https://flatpak.org/setup/Ubuntu

There are a few fixes that are only in master and not released yet:
- auto adding "api/greader.php" to the server url if it is missing
- properly escaping some special characters in passwords
- and an issue renaming tags i just noticed right now testing it for the screenshot above

I'll release these changes soon.

jangernert:
New version with [FreshRSS fixes](https://gitlab.com/news-flash/news_flash_gtk/-/blob/master/data/com.gitlab.newsflash.appdata.xml.in.in#L61) is out. I think this is now ready.

While I was at it I also removed FeedReader from the list of clients. I abandoned it quite a while ago by now.

Alkarex:
I have found the time at last to give it a try on another machine. It looks pretty good already.

But the synchronisation does not work on my test, because it does not scale to cases with more than 1000 articles, e.g. stared.

```
PHP Warning:  Unknown: Input variables exceeded 1000. To increase the limit change max_input_vars in php.ini. in Unknown on line 0
```

Same bug than https://github.com/FreshRSS/FreshRSS/pull/3859#issuecomment-1242709955

The sync shows me 47 unread articles in total, while I have 8112.

I can provide a test account and/or test database if you like.

Edit: Ref https://gitlab.com/news-flash/news_flash/-/issues/84

Alkarex:
P.S. It does not like the app supports *Fetch more in individual views* at the moment, so I have removed the tick there.
An example would be to go to a given label or feed, and refresh to get more (older) articles from the server of that specific label or feed.

PR #4651, opened 2022-09-28
Improved: Minz Framework. Byebye Internet Explorer and its Conditional comments

math-GH:
The Internet Explorer is past.
Its [conditional comments](https://en.wikipedia.org/wiki/Conditional_comment) are gone, too.

- Minz-Framework View.php

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


lib/Minz/View.php
```diff
@@ -202,36 +202,25 @@ public static function headStyle () {
...
-				$styles .= '<![endif]-->';
-			}
-
 			$styles .= "\n";
 		}
 
 		return $styles;
 	}
-	public static function prependStyle ($url, $media = 'all', $cond = false) {
+	public static function prependStyle ($url, $media = 'all') {
```
Alkarex:
```suggestion
	public static function prependStyle($url, $media = 'all') {
```

Let's take the opportunity to use our current whitespace convention (no space before function parenthesis). Same below

lib/Minz/View.php
```diff
@@ -258,29 +242,22 @@ public static function headScript () {
...
-				$scripts .= '<![endif]-->';
-			}
-
 			$scripts .= "\n";
 		}
 
 		return $scripts;
 	}
-	public static function prependScript ($url, $cond = false, $defer = true, $async = true, $id = '') {
+	public static function prependScript ($url, $defer = true, $async = true, $id = '') {
```
Alkarex: It is a bit dangerous to change the signature definition of a public method of this type. It does not look like we use the additional parameters in the core of FreshRSS, but it could be good to double-check some of our known extensions

Alkarex:
```suggestion
	public static function prependScript($url, $defer = true, $async = true, $id = '') {
```

math-GH:
How to proceed here?

Alkarex:
> How to proceed here?

It could be good to double-check some of our known extensions, to make sure none is using this method

Frenzie:
prepend/append style and script? I'd say those are used by most extensions, although it's possible the other arguments such as defer and async aren't used much.

math-GH:
I checked the extension repository:

- `prependStyle()` /`prependScript()` is not used
- `appendStyle()` is used, but only the first argument is used
- `appendScript()` is used:

grafik

I understand that an API change is not a good idea.

Should I revert these changes? All changes? Only the appendScript() change?

Frenzie:
Definitely not only appendScript; that'd be really confusing.

I would close this PR, but you could log a depreciation notice instead.

math-GH:
> I would close this PR, but you could log a depreciation notice instead.

This PR is not about the deletion of a parameter at a first place. It was just a follow-up of the other things.

lib/Minz/View.php
```diff
@@ -202,36 +202,25 @@ public static function headStyle () {
...
-				$styles .= '<![endif]-->';
-			}
-
 			$styles .= "\n";
 		}
 
 		return $styles;
 	}
-	public static function prependStyle ($url, $media = 'all', $cond = false) {
+	public static function prependStyle($url, $media = 'all') {
```
Frenzie: Something like this perhaps, @Alkarex ?
```suggestion
	/**
	 * Prepends a <link> element referencing stylesheet.
	 *
	 * @param string $url
	 * @param string $media
	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
	 */
	public static function prependStyle($url, $media = 'all', $cond = false) {
```

Frenzie:
Yup, my statement follows the same logic in reverse. But I also said I'm fine with a deprecation notice. ^_^

lib/Minz/View.php
Alkarex: Would `@deprecated` work for a parameter?

Frenzie: Not that I know of or can find at a glance. Unless you mean `@param deprecated $cond`; that'll probably work as a custom type.

lib/Minz/View.php
```diff
@@ -202,36 +202,32 @@ public static function headStyle () {
...
+	public static function prependStyle($url, $media = 'all', $cond = false) {
 		array_unshift (self::$styles, array (
 			'url' => $url,
-			'media' => $media,
-			'cond' => $cond
+			'media' => $media
 		));
 	}
-	public static function appendStyle ($url, $media = 'all', $cond = false) {
+	public static function appendStyle ($url, $media = 'all') {
```
Alkarex:
```suggestion
	/**
	 * Append a `<link>` element referencing stylesheet.
	 * @param string $url
	 * @param string $media
	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
	 */
	public static function appendStyle($url, $media = 'all', $cond = false) {
```

lib/Minz/View.php
```diff
@@ -202,36 +202,32 @@ public static function headStyle () {
...
+	 * @param string $url
+	 * @param string $media
+	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
+	 */
+	public static function prependStyle($url, $media = 'all', $cond = false) {
 		array_unshift (self::$styles, array (
 			'url' => $url,
-			'media' => $media,
-			'cond' => $cond
+			'media' => $media
```
Alkarex:
```suggestion
			'media' => $media,
```

lib/Minz/View.php
```diff
@@ -202,36 +202,32 @@ public static function headStyle () {
...
-			'cond' => $cond
 		));
 	}
-	public static function appendStyle ($url, $media = 'all', $cond = false) {
+	public static function appendStyle ($url, $media = 'all') {
 		self::$styles[] = array (
 			'url' => $url,
-			'media' => $media,
-			'cond' => $cond
+			'media' => $media
```
Alkarex:
```suggestion
			'media' => $media,
```

lib/Minz/View.php
```diff
@@ -258,29 +255,22 @@ public static function headScript () {
...
-				$scripts .= '<![endif]-->';
-			}
-
 			$scripts .= "\n";
 		}
 
 		return $scripts;
 	}
-	public static function prependScript ($url, $cond = false, $defer = true, $async = true, $id = '') {
+	public static function prependScript($url, $defer = true, $async = true, $id = '') {
```
Alkarex:
```suggestion
	/**
	 * Prepend a `<script>` element.
	 * @param string $url
	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
	 * @param bool $defer Use `defer` flag
	 * @param bool $async Use `async` flag
	 * @param string $id Add a script `id` attribute
	 */
	public static function prependScript($url, $cond = false, $defer = true, $async = true, $id = '') {
```

lib/Minz/View.php
```diff
@@ -258,29 +255,30 @@ public static function headScript () {
...
 		array_unshift(self::$scripts, array (
 			'url' => $url,
-			'cond' => $cond,
 			'defer' => $defer,
 			'async' => $async,
 			'id' => $id,
 		));
 	}
-	public static function appendScript ($url, $cond = false, $defer = true, $async = true, $id = '') {
+	public static function appendScript ($url, $defer = true, $async = true, $id = '') {
```
Alkarex:
```suggestion
/**
	 * Append a `<script>` element.
	 * @param string $url
	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
	 * @param bool $defer Use `defer` flag
	 * @param bool $async Use `async` flag
	 * @param string $id Add a script `id` attribute
	 */
	public static function appendScript ($url, $defer = true, $async = true, $id = '') {
```

Alkarex (approved):

lib/Minz/View.php
```diff
@@ -258,29 +255,38 @@ public static function headScript () {
...
-	public static function appendScript ($url, $cond = false, $defer = true, $async = true, $id = '') {
+/**
+	 * Append a `<script>` element.
+	 * @param string $url
+	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
+	 * @param bool $defer Use `defer` flag
+	 * @param bool $async Use `async` flag
+	 * @param string $id Add a script `id` attribute
+	 */
+	public static function appendScript ($url, $defer = true, $async = true, $id = '') {
```
Alkarex:
```suggestion
	public static function appendScript ($url, $cond = false, $defer = true, $async = true, $id = '') {
```

lib/Minz/View.php
```diff
@@ -258,29 +255,38 @@ public static function headScript () {
...
 	}
+/**
+	 * Append a `<script>` element.
+	 * @param string $url
+	 * @param bool $cond Conditional comment for IE, now deprecated and ignored
+	 * @param bool $defer Use `defer` flag
+	 * @param bool $async Use `async` flag
+	 * @param string $id Add a script `id` attribute
+	 */
 	public static function appendScript ($url, $cond = false, $defer = true, $async = true, $id = '') {
```
Alkarex:
```suggestion
	public static function appendScript($url, $cond = false, $defer = true, $async = true, $id = '') {
```

PR #4652, opened 2022-09-28
Improved: Reader view: Mark article as read while scrolling

math-GH:
## Before

Reader view:

when the user clicks left or right hand side of the article then the article gets marked as read
grafik

grafik

It is not a good UX, because it is unexpected to mark the article as read while clicking on an empty space.

When the user scrolls down and will expect, that all articles are marked as read while scrolling, the "clicked" article keeps in the stream. But the "clicked" article keeps. Also when it is again marked as unread.

## After:
Click left/right fro the article does not change the reading status
When scrolling, all unread articles disappear after scrolling.

- main.js (add an exception for the click events, that aborts the click event when reader view is selected)

Reading settings:
grafik

1. Filter: show only unread articles
2. View: Reader
3. click left/right of the article
4. Scroll down
5. scroll up
6. no unread article anymore above

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1109,7 +1109,7 @@ function init_stream(stream) {
 
 		el = ev.target.closest('.flux_header, .flux_content');
 		if (el) {	// flux_toggle
-			if (ev.target.closest('.content, .item.website, .item.link, .dropdown')) {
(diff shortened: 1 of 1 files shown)


beerisgood:
Awesome!

PR #4654, opened 2022-09-29
Allow redirection after login

### LINKED ISSUE #3663: [Feature] Add URL query parameter to login page to get redirected to a specific page after login process

A consequent number FreshRSS users don't understand the correct setup process to get the HTTP API working. A lot of [FreshRSS Android](https://git.feneas.org/christophehenry/freshrss-android) users come to me unable to login to their instance believing in a problem with the app when actually their instance is misconfigured.

I'm currently developping a new login panel to detect such misconfigurations and I need to be able to point users to the relevant pages to fix them. But these pages just return 404 unless you're logged in which can be disconcerting. A query parameter on the login page to directly redirect to a specifc route after the login process would be a nice UX improvement.

Some query param like `redirect=/i/?c=auth` to get redirected to auth configurations page, for instance, would be nice.

Also, don't 404 on valid routes when not logged in, just redirect to login page would be a viable alternative.

aledeg:
Before, if you've tried to reach a page without being logged, you'll be automatically redirected to the index page after login. Now, the original page is used after login.

Closes #3663

- allows redirection after login

1. try to reach a specific page without being logged-in.
2. log-in.
3. validate that the displayed page is the targeted page.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -160,7 +160,7 @@ class FreshRSS_auth_Controller extends FreshRSS_ActionController {
 				Minz_Translate::init(FreshRSS_Context::$user_conf->language);
 
 				// All is good, go back to the index.
-				Minz_Request::good(_t('feedback.auth.login.success'), [ 'c' => 'index', 'a' => 'index' ]);
+				Minz_Request::good(_t('feedback.auth.login.success'), Minz_Url::unserialize(Minz_Request::param('original_request')));
 			} else {
 				Minz_Log::warning("Password mismatch for user={$username}, nonce={$nonce}, c={$challenge}");
 
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -15,6 +15,8 @@ class Minz_Request {
 	private static $default_controller_name = 'index';
 	private static $default_action_name = 'index';
 
+	private static $originalRequest;
+
 	/**
 	 * Getteurs
 	 */
@@ -91,6 +93,9 @@ class Minz_Request {
 			'a' => self::$action_name,
 			'params' => self::$params,
 		);
+	}
+	public static function originalRequest() {
+		return self::$originalRequest;
 	}
 	public static function modifiedCurrentRequest(array $extraParams = null) {
 		$currentRequest = self::currentRequest();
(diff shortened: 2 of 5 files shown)


aledeg:
I have no idea if the PR makes sense since it only address the form login. Do we need to make more things? Do we need things differently?

aledeg:
I just realize that I missed the point since we should add that on the API. doh!
Maybe we can use that anyway. Let me know.

app/Controllers/authController.php
```diff
@@ -160,7 +160,7 @@ public function formLoginAction() {
 				Minz_Translate::init(FreshRSS_Context::$user_conf->language);
 
 				// All is good, go back to the index.
-				Minz_Request::good(_t('feedback.auth.login.success'), [ 'c' => 'index', 'a' => 'index' ]);
+				Minz_Request::good(_t('feedback.auth.login.success'), Minz_Url::unserialize(Minz_Request::param('original_request')));
```
Alkarex, line 163: What if we land on the login form to start with? Risk of infinite loop?

aledeg: You're right. I've removed the original request if it's an auth request.

lib/Minz/Request.php
```diff
@@ -327,6 +332,10 @@ public static function getNotification() {
 	 *                > sinon, le dispatcher recharge en interne
 	 */
 	public static function forward($url = array(), $redirect = false) {
+		if (Minz_Request::originalRequest() === null && strpos('auth', json_encode($url)) !== false) {
```
Alkarex, line 335: Cant' the `'auth'` be tested on the right $url array element without `json_encode()` ?

aledeg: It could if we were sure its an array but as per the line 339, we are not.

app/Controllers/authController.php
Alkarex, line 163: Probable regression: https://github.com/FreshRSS/FreshRSS/issues/4766

Alwaysin: I had a problem of a white page after first logging in on a brand new install, I had to refresh the page to get it working. Could it be linked? I thought it was my specific setup but I can try to reproduce it.

Alkarex: Yes, that's it @Alwaysin

aledeg: @Alwaysin could you check #4778 and tell me if this fixes your issue?

Alwaysin: @aledeg Yes, it does!

lib/Minz/Request.php
Alkarex: The `strpos()` parameters were reversed. I believe this worked just by chance :-P
https://github.com/FreshRSS/FreshRSS/pull/5271

PR #4666, opened 2022-09-30
Improved: Article summary. Cut the text with better CSS

math-GH:
Ref #3805

After:
(before it looks similar)
grafik

- CSS: instead of using a max-height: using `-webkit-line-clamp: 2;`. 

1. enable "summary" in config -> display
2. see normal view

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1206,14 +1206,15 @@ a.website:hover .favicon {
 }
 
 .flux .item.title .summary {
-	max-height: 3em;
+	display: -webkit-box;
 	color: var(--frss-font-color-grey-dark);
 	font-size: 0.9em;
 	line-height: 1.5em;
 	font-weight: normal;
-	white-space: initial;
-	overflow: hidden;
 	text-overflow: ellipsis;
+	-webkit-line-clamp: 2;
+	-webkit-box-orient: vertical;
+	white-space: break-spaces;
 }
 
 .flux .item.title .author {
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -62,9 +62,7 @@
 			?></span><?php
 		endif;
 		if ($topline_summary):
-			?><div class="summary">
-				<?= mb_substr(strip_tags($this->entry->content()), 0, 500, 'UTF-8') ?>
-			</div><?php
+			?><div class="summary"><?= mb_substr(strip_tags($this->entry->content()), 0, 500, 'UTF-8') ?></div><?php
 		endif;
 	?></a></li>
 	<?php if ($topline_date) { ?><li class="item date"><time datetime="<?= $this->entry->machineReadableDate() ?>"><?= $this->entry->date() ?></time>&nbsp;</li><?php } ?>
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -62,7 +62,7 @@
 			?></span><?php
 		endif;
 		if ($topline_summary):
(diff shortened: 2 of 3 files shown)


aledeg:
Can't we remove the modification done on the php side and use `text-overflow: ellipsis` on css side?

p/themes/base-theme/frss.css
```diff
@@ -1206,14 +1206,15 @@ a.website:hover .favicon {
 }
 
 .flux .item.title .summary {
-	max-height: 3em;
+	display: -webkit-box;
```
Alkarex, line 1209: I am not so found of introducing proprietary attributes, if that could be avoided. What about non-Webkit browsers?

math-GH: Unfortunately not. But it is implemented in the major browsers:
https://caniuse.com/?search=-webkit-box

math-GH:
> Can't we remove the modification done on the php side and use `text-overflow: ellipsis` on css side?

Unfortunately not. For the summary text, it is used the content text. If you scrap the content, then you could have a 10.000+words + images content. We need to cut it before. :/

aledeg:
> > Can't we remove the modification done on the php side and use `text-overflow: ellipsis` on css side?
> 
> Unfortunately not. For the summary text, it is used the content text. If you scrap the content, then you could have a 10.000+words + images content. We need to cut it before. :/

I understand that. And I think that if you set a size for the box, the content will overflow it. Thus the use of `text-overflow:ellipsis`.

See https://codepen.io/joelsaupe/pen/ojmLWB for an example.

math-GH:
> > > Can't we remove the modification done on the php side and use `text-overflow: ellipsis` on css side?
> > 
> > 
> > Unfortunately not. For the summary text, it is used the content text. If you scrap the content, then you could have a 10.000+words + images content. We need to cut it before. :/
> 
> I understand that. And I think that if you set a size for the box, the content will overflow it. Thus the use of `text-overflow:ellipsis`.
> 
> See https://codepen.io/joelsaupe/pen/ojmLWB for an example.

I am not sure that we understand eachother -.-

I would keep the text cut with PHP `mb_substr(strip_tags($this->entry->content()), 0, 500, 'UTF-8')` for the summary. It does not make sense to blow up the HTML with double content.

Frenzie:
On your screenshot 500 looks a bit short in combination with letters that kern well together. Raising the limit to 800 or so in PHP and using CSS to make it more uniform could be an idea. (But I don't know if it's worth the effort.)

Frenzie (approved):

math-GH:
> On your screenshot 500 looks a bit short in combination with letters that kern well together. Raising the limit to 800 or so in PHP and using CSS to make it more uniform could be an idea. (But I don't know if it's worth the effort.)

I did not touch the text cut limit. It was set by #3805

This PR just wanted to change the CSS code from 
```
overflow: hidden;
max-height: 3em;
```
to
```
display: -webkit-box;
-webkit-line-clamp: 2;
```

I had to add the PHP `trim()` because of the `strip_tags()` that leave a lot of white spaces, if the content is scrapped via CSS selectors.

Frenzie:
> I did not touch the text cut limit.

Right, but I'm saying it'd probably look better if you did, whether in this PR or another. But I also expressed that I think this PR is good in its current form. ;-)

Alwaysin:
@math-GH It looks like from this commit, origine-compact is no compact anymore (looks like same problem as described in https://github.com/FreshRSS/FreshRSS/pull/4636#issuecomment-1255902903 by Alkarex (or even worse!), fixed there, but broke again here :()

math-GH:
I will check, but it cannot come from this PR.

Alwaysin:
Sorry I mixed things up.

Real problematic commit is https://github.com/FreshRSS/FreshRSS/commit/01a8c37b8308c283b2c71d629260bd30f8cf85cc

PR #4671, opened 2022-10-01
Improved: Frss.css - font size + line height with 'rem' values

### LINKED ISSUE #3733: [BUG] Origine line-height with units

math-GH:
Step 3 of #4642 is done for frss.css (formerly known as template.css)
(Close #3733)

- CSS: `font-size` lines have now `rem` units
- CSS: `line-height` lines are now unit less
- PHTML: to remove some CSS lines, the close "icon" (it was a UTF-8 icon) was replaced by the common close icon

Information: I checked (hopefully everywhere that the 'rem' values are (more or less) equal to the previous values.

1. general everywhere ;)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Quick test with Pafat theme:
In normal view, hovering over article titles make the lines jump a bit.
The line height is slightly different in Pafat (but not necessarily a problem - looks actually a tad better to me)

math-GH:
> Quick test with Pafat theme: In normal view, hovering over article titles make the lines jump a bit. The line height is slightly different in Pafat (but not necessarily a problem - looks actually a tad better to me)

Fixed

p/themes/base-theme/frss.css
```diff
@@ -37,6 +37,9 @@
 	--frss-scrollbar-track-hover: #0001;
 
 	--frss-loading-image: url("loader.gif");
+
+	font-size: 16px; /* do we really need it or should we trust the browser config? */
```
Frenzie: I would definitely prefer to *respect* (not trust!) the browser config. If someone changes it they're doing so for a reason.

p/themes/base-theme/frss.rtl.css
```diff
@@ -1346,7 +1344,7 @@ a.website:hover .favicon {
 }
 
 br {
-	line-height: 1em;
+	line-height: 1;
```
Frenzie: Hm, I'm not convinced we should change this at all.

Frenzie (commented):
lgtm in principle, left a couple quick remarks

p/themes/base-theme/frss.rtl.css
math-GH: I was not sure to delete this lines.

That is the commit of this line: https://github.com/math-GH/FreshRSS/commit/0ef1fd6e4dd55f3626bae585b04b82ad194b8c63

Frenzie: I'd definitely delete both of those. (Sorry @Alkarex ;-) )

I may have noticed the negative effects in the past but I didn't have time to investigate and wrote it off as a broken feed. (I.e., line breaks that should've been there weren't.)

p/themes/base-theme/frss.css
math-GH: Any suggestion how to change? delete this line? `font-size: 100%`?

Frenzie: Just delete, yes.

math-GH: Would it makes sense to have an option to set/change the default size within FreshRSS?

p/themes/base-theme/frss.rtl.css
math-GH: Any veto from @Alkarex ?

Alkarex: All good for me :-) I cannot remember having said otherwise, have I?

p/themes/base-theme/frss.css
Frenzie: If you think it wouldn't make things too crowded under appearance I suppose. Is it really necessary though? It's really easy to adjust the zoom level and browsers remember it.

p/themes/base-theme/frss.rtl.css
Frenzie: What about the `br + br + br` one?

p/themes/base-theme/frss.css
math-GH: Font size and zoom levels are 2 different things ;)

You could change the font-size to 20 and leave the zoom level on 100%.

I propose to have it right now without a config option. Will see if someone will open a feature request/bug issue report

Frenzie: I'm aware that they're different; I'm questioning whether they're different enough to justify the cost. I'd either add it or not. It doesn't make much sense to add maintenance complexity for what's a CustomCSS one-liner imho.

p/themes/base-theme/frss.rtl.css
math-GH: where is it?

Frenzie: I don't know, are you saying it's already gone?

I mean from the commit you posted: https://github.com/FreshRSS/FreshRSS/commit/0ef1fd6e4dd55f3626bae585b04b82ad194b8c63

math-GH: It is already gone

Alkarex:
See regression probably coming from this PR: https://github.com/FreshRSS/FreshRSS/issues/4712

nicofrand:
Hi,

FYI this also fixes an issue on FreshRSS 1.20.0 where `.flux_header` on hover did not have `position: relative` and the title just kinda disappears, on origine compact at least.

Issue #2459, opened 2019-07-24
Feature Request: Subscription Management


ddeimeke:
I subscribed to around 600 Feeds.

The pages showing all categories and only the first seven feeds gives no overview.

It would be great to have a list, which can be collapsed by category (seeing no feeds, only category titles) or expanded to see all categories and content.

Having separate lists with failed feeds and a last updated date would be a plus as well.

The gear is great for changing properties, having a separate unsubscribe or remove icon would be perfect.

Alkarex:
While waiting for the rest, there is already an option to filter failed feeds at the bottom of the subscription page:
![image](https://user-images.githubusercontent.com/1008324/77257044-27eaeb00-6c72-11ea-9d31-d56689ead857.png)

ddeimeke:
That helps a lot, thanks @Alkarex

math-GH:
@ddeimeke have you tried https://github.com/FreshRSS/Extensions/tree/master/xExtension-CustomCSS? There you could change:

```
.box .box-content {
max-height: none;
}
```

It will expand the lists to have a full overview.

With https://github.com/FreshRSS/Extensions/tree/master/xExtension-CustomJS it should be easy to develop a expand toggle

PR #4672, opened 2022-10-01
Improved: Subscription management with long lists

math-GH:
Rref #2459

# (Still a draft, feedback welcome)

The issue right now: the category boxes are very short. It is difficult to have a proper overview about all feeds, if the user has a lot feeds.

My idea:
- switch the view via button: from box-view to a column view with full extended lists

Before
grafik

After
1) horizontal scrolling
2) fully expanded feed list
grafik

- tbd
-
-

1. go to "subscription management"
2. click on the button ↔️
3. see the "Trello like" column view

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


aledeg:
That's a good idea. It could be done also with a collapsible tree view.

Alkarex:
Great, I was in need of something of this type. For the record, I have one use-case with a category containing 2k+ feeds

aledeg:
> Great, I was in need of something of this type. For the record, I have one use-case with a category containing 2k+ feeds

Wow! That's intense/insane!

Alkarex:
> Wow! That's intense/insane!

2860 feeds to be precise :-) It is for a usecase at work, based on Dynamic OPML, with one feed per company of that category.

Related to:

* Allow many (50k+) feeds [#4347](https://github.com/FreshRSS/FreshRSS/pull/4347)
  * See also [#4357](https://github.com/FreshRSS/FreshRSS/pull/4357), [#4353](https://github.com/FreshRSS/FreshRSS/pull/4353), [#4417](https://github.com/FreshRSS/FreshRSS/pull/4417), [#4502](https://github.com/FreshRSS/FreshRSS/pull/4502)

aledeg:
Maybe it could be easier to have a less "fun" view and have a basic table with filters to display the list of feeds and categories? My 2¢

math-GH:
> Maybe it could be easier to have a less "fun" view and have a basic table with filters to display the list of feeds and categories? My 2¢

Why not both?
I see here different interests: The heavy user with hundreds/thousands feeds and the casual user with dozens feeds. Both requires different UI/UX.
The current UI/UX focuses on the casual user with a drag&drop sorting.

aledeg:
I have nothing against both but we have to keep in mind that it doubles the amount of work.

Alkarex:
An easy approach would be a toggle to just turn on/off the CSS `max-height` of the boxes.

PR #4679, opened 2022-10-02
Improved: Subscription Management: Show the position number

math-GH:
The issue: it is a bit tricky to order the categories. The position number is more or less invisible.

The position is visible only in the settings:
grafik

Now the number is shown in the box
grafik

- PHTML and CSS

2. config the position number
3. see the number on the box

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/subscription/index.phtml
+++ b/app/views/subscription/index.phtml
@@ -33,13 +33,8 @@
 				$feeds = $cat->feeds();
 		?>
 		<div class="box">
-			<?php
-				$categoryDAO = FreshRSS_Factory::createCategoryDao();
-				$category = $categoryDAO->searchById($cat->id());
-				$pos = $category->attributes('position');
-				?>
 			<div class="box-title">
-				<a class="configure open-slider" href="<?= _url('subscription', 'category', 'id', $cat->id()) ?>" data-cat-position="<?= $pos ?>"><?= _i('configure') ?></a>
+				<a class="configure open-slider" href="<?= _url('subscription', 'category', 'id', $cat->id()) ?>" data-cat-position="<?= $cat->attributes('position') ?>"><?= _i('configure') ?></a>
 				<?= $cat->name() ?>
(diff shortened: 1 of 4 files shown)


math-GH:
Maybe someone could support here:

These lines are not the best in the template:
```
$categoryDAO = FreshRSS_Factory::createCategoryDao();
$category = $categoryDAO->searchById($cat->id());
$pos = $category->attributes('position');
```

aledeg:
> Maybe someone could support here:
> 
> These lines are not the best in the template:
> 
> ```
> $categoryDAO = FreshRSS_Factory::createCategoryDao();
> $category = $categoryDAO->searchById($cat->id());
> $pos = $category->attributes('position');
> ```

I think (not tested) that you could drop that and instead of `$pos` use `$cat->attributes('position')` directly in the template. You are already in a loop where you are using categories.

app/views/subscription/index.phtml
```diff
@@ -33,8 +33,13 @@
 				$feeds = $cat->feeds();
 		?>
 		<div class="box">
+			<?php
+				$categoryDAO = FreshRSS_Factory::createCategoryDao();
+				$category = $categoryDAO->searchById($cat->id());
```
Alkarex: We probably already have all the categories in our context. See also `$this->categories` and its definition in:

https://github.com/FreshRSS/FreshRSS/blob/3b54f8cd8196728df14590bfb1560dd4f6aaf685/app/Controllers/subscriptionController.php#L22

Ups, by the way, I can see that it looks like we have one `true` too many

math-GH: I fixed the doubled parameter

math-GH:
Thanks for your help. It works :)

aledeg:
@math-GH There is an issue when the category name is really long. The number and the name are entangled. It's quite impossible to see it. It might be a good idea to add something to handle that case.

Issue #2016, opened 2018-09-10
Improving the "mark all as read" big button


marienfressinaud:
This button happened to be a bit confusing for users. First, it says there is no more articles, so why should we mark them as read?

Also, they generally don't understand why it is so big.

Then, once we click on it, why do we still have articles that appear?

lapineige:
In addition, even with the confirmation pop-up, it's still a bit too "dangerous" (especially on mobile devices) to have such a big button, it's easy to miss click on it, and remove weeks/months(/years ?) of filtering between read and to-read articles.

marienfressinaud:
@lapineige I opened a new ticket considering the issue of loosing the list of unread articles https://github.com/FreshRSS/FreshRSS/issues/2018. I don't think the button is the real problem in this case, but more probably FreshRSS don't support the feature that would fit with your needs :)

Edit: btw, it's a problem that is raised very often to me, I would love to find the correct answer!

lapineige:
> Edit: btw, it's a problem that is raised very often to me, I would love to find the correct answer!

An "undo" button ? :D

lapineige:
> There is already an option to confirm before marking as read

Hopefully, but I still activated it by error a few times ^^
(I love my backups :)

> Unread articles are purged when they get too old, unlike favourites (which are more appropriate to keep articles anyway)

Except if we enable the option to keep them, isn't it ?
(anyway I'm purging anything older than 2 years, which is already very long)

GLLM:
My 2 cents : I never marked as read articles by mistake using the Mark all as read buttons, thanks to the warning pop-up !
Have I ever corn so, I would have gone mad because I have plenty of unread articles meant to be read later.

sirideain:
I too find the large Mark All As Read button annoying on mobile. I would love an option to remove the button. If I want to mark all as read, I can use the button in the toolbar which is far harder to accidentally hit.

math-GH:
Please check the current 1.21.0-dev. #4681 is implemented and was inspired by this discussion. 

Could we close this ticket?

PR #4681, opened 2022-10-03
Improved: show bigmarkasread button only, when there are unread articles

math-GH:
It will improve the UX of the "Mark all as read" button

New:

If there is no unread article, then the "Mark all as read" button is hidden.
grafik

If min. 1 article is marked as unread, then the button is shown.
grafik

The button's visibility toggles when the user toggle the read status of the articles.

- some little JavaScript
- PHTML: the "There are no more articles" text is not anymore part of the button

1. load the article list
2. toggle the read/unread status of the articles

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Maybe an unwanted side-effect: some people might use this button anyway to move to the next view (next feed / next category...)

math-GH:
> Maybe an unwanted side-effect: some people might use this button anyway to move to the next view (next feed / next category...)

Loud thoughts:
It depends on this config
grafik

So it would mean, that a button "go to next unread feed/category" would be better in that case.

math-GH:
Added:

If there is no unread article, but the option `jump to next unread sibling (feed or category)` is enabled:
grafik

if there is unread article and the option `jump to next unread sibling (feed or category)` is enabled:
grafik

p/scripts/main.js
```diff
@@ -1661,6 +1662,23 @@ function refreshUnreads() {
 	req.send();
 }
 
+function toggle_bigMarkAsRead_button() {
+	var bigMarkAsRead_button = document.getElementById('bigMarkAsRead');
```
```suggestion
	const bigMarkAsRead_button = document.getElementById('bigMarkAsRead');
```

math-GH:
ref #2016

app/views/helpers/stream-footer.phtml
```diff
@@ -30,19 +30,20 @@ if ($hasAccess) { ?>
...
 			class="as-link <?= FreshRSS_Context::$user_conf->reading_confirm ? 'confirm" disabled="disabled' : '' ?>"
 			form="stream-footer"
 			formaction="<?= Minz_Url::display($url_mark_read) ?>"
 			type="submit">
-			<?= _t('gen.stream.nothing_to_load') ?><br />
 			<span class="bigTick">✓</span><br />
-			<?= _t('gen.stream.mark_all_read') ?>
+			<span class="markAllRead"><?= _t('gen.stream.mark_all_read') ?></span>
+			<?php if (FreshRSS_Context::$user_conf->onread_jump_next) { ?>
+				<div class="jumpNext"><?= _t('conf.reading.jump_next') ?></div>
```
Alkarex, line 42: Element `div` is not allowed as child of element `button`.
Should be replaced by a `span` after checking that styles and scripts are OK

PR #4709, opened 2022-10-10
Main font color from blue to regular dark

### LINKED ISSUE #4695: [BUG] Can't distinguish links on mapco theme
Example:

chrome_2022-10-06_14-25-50

chrome_2022-10-06_14-26-04

You just know there's a link there if you hover over it. Maybe a slightly darker blue would do the trick.

But the fact that everything is overrided by blue is a bit disturbing, for example:

chrome_2022-10-06_17-24-14

You lose information compared to the original with another theme:

chrome_2022-10-06_17-23-36

batchennings:
Closes #4695

### DIFF
--- a/p/themes/Mapco/mapco.css
+++ b/p/themes/Mapco/mapco.css
@@ -1370,7 +1370,7 @@ main.prompt {
 /*============*/
 html, body {
 	background: #eff0f2;
-	color: #36c;
+	color: #303136;
 	font-family: "lato", "Helvetica", "Arial", sans-serif;
 	font-size: 0.875rem;
 }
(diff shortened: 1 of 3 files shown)


Alkarex (approved):

p/themes/Mapco/mapco.css
```diff
@@ -1384,4 +1384,4 @@ body.register {
 a, button.as-link {
 	outline: none;
 	color: #36c;
-}
\ No newline at end of file
+}
```
math-GH: I think it changed to another line break.

math-GH: strange. It is marked as "outdated"

math-GH (approved):
It is fine for me. The end of file think I do not understand what is happened there

PR #4710, opened 2022-10-10
Update README.md for ReadKit

gingerbeardman:
I use ReadKit (free version) on iOS and macOS. 

Pro purchase is required to manage feeds, which I don't agree with given that we're using FreshRSS. Managing local feeds I can understand that could be value add.

- correct ReadKit info

Pull request checklist:

- [X] clear commit messages
- [X] documentation updated


gingerbeardman:
> I use ReadKit (free version) on iOS and macOS.
> 
> Pro purchase is required to manage feeds, which I don't agree with given that we're using FreshRSS. Managing local feeds I can understand that could be value add.
> 
> Changes proposed in this pull request:
> 
> * correct ReadKit info
> 
> Pull request checklist:
> 
> * [x]  clear commit messages
> * [x]  documentation updated

aledeg:
@gingerbeardman I would add a footnote on the manage feed to mentioned that you need to pay for that in-app feature.

Alkarex:
I have finally found an iOS device and the time to test and update the remaining details.

gingerbeardman:
Many thanks

PR #4711, opened 2022-10-11
Mapco text display improvements

### LINKED ISSUE #4695: [BUG] Can't distinguish links on mapco theme

batchennings:
Closes #4695


batchennings:
ouch, automated testing is demanding, i got errors on minor things (from my point of view). should i fix everything for this PR to be accepted ?

Alkarex:
The easiest is to run `make fix-all` (which I have just done) or the specific automated methods.

p/themes/Mapco/mapco.scss
```diff
@@ -51,3 +51,7 @@ a, button.as-link {
 	outline: none;
 	color: variables.$main-first;
 }
+
+.icon {
+	height: inherit;
+}
```
math-GH: What is the reason for this line?

Frenzie:
> i got errors on minor things (from my point of view). should i fix everything for this PR to be accepted ?

Yes, it whines about it so I don't have to whine about it. ;-P

p/themes/Mapco/mapco.scss
batchennings: It overwrites a rule in `p/themes/base-theme/frss.css` that vertically scales the image icon, and distort it.

```css
.icon {
	display: inline-block;
	max-width: none;
	width: 1rem;
	height: 1rem; /* this value distorts icon vertically*/
	vertical-align: middle;
	line-height: 1;
}
```

https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/base-theme/frss.css#L726

patjennings: https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/base-theme/frss.css#L726

math-GH: ah I see. I would prefer to improve frss.css and delete the `width` line and change `height: 1rem` to `height: inherit`. `height` and `width` were `16px` before I changed it to `1rem`. It should not be the issue of the themes to handle it by default.

batchennings:
i'm struggling with the `make fix-all` command. My poor skills barely give me a clue on how to solve it. I let as is, sorry for that.

p/themes/Mapco/mapco.rtl.css
```diff
@@ -1385,3 +1373,7 @@ a, button.as-link {
 	outline: none;
 	color: #36c;
 }
+
+.icon {
+	height: inherit;
+}
```
math-GH: I guess you forgot to commit mapco.scss, that inserts this lines into the CSS file.

I would suggest to do another PR for changing the width/height behavior of `.icon` and its `height`. Let's do not do it with this PR. I can support you with a PR.

p/themes/base-theme/frss.css
```diff
@@ -723,7 +723,7 @@ input[type="checkbox"]:focus-visible {
 	display: inline-block;
 	max-width: none;
 	width: 1rem;
-	height: 1rem;
+	height: inherit;
```
math-GH, line 730: see my comment above

phpcs.xml
```diff
@@ -14,6 +14,10 @@
 	<exclude-pattern>./p/scripts/vendor/</exclude-pattern>
 	<exclude-pattern>./vendor/</exclude-pattern>
 	<exclude-pattern>*.min.js$</exclude-pattern>
+	<exclude-pattern>./p/themes/Ansum/*.css$</exclude-pattern>
+	<exclude-pattern>./p/themes/Mapco/*.css$</exclude-pattern>
+	<exclude-pattern>./p/themes/Swage/*.css$</exclude-pattern>
+	<exclude-pattern>*.rtl.css$</exclude-pattern>
```
math-GH: I am not so happy to exclude this files.
The CSS files were changed without the SCSS files in the past, so we should keep the phpcs check for this CSS files.

Frenzie: All files excluded here are normally auto-generated, so in a way it does make sense. That being said, what specifically is the reason to exclude these randomly in this PR?

Alkarex: That is my fault, because there was several times in a raw fixes to be done in this PR to pass the automated tests, with changes that did not really make sense

Alkarex: @math-GH The excluded files are all auto-generated. But I do not mind reverting that change

Frenzie: Looking in closer detail, if you're talking about https://github.com/FreshRSS/FreshRSS/pull/4711/commits/1bab465af9715346afd4ce56a049a9edf563dcae then that definitely *does* make sense because frss.rtl.css was out of sync with frss.css. ^_^

It's not quite the correct test or fix but I'd rather keep it in place as an indicator unless someone writes one.

Perhaps it'd be as simple as adding `rtl` to this line:
https://github.com/FreshRSS/FreshRSS/blob/cb1c2da8e6bd923044df3ab1e5fd56cf95e24ef2/Makefile#L225-L226

Alkarex: `rtlcss` is already called via `make fix-all` via `npm run fix`.
But it is not a part of the tests indeed

Alkarex: Maybe we could/should automatically call `make fix-all` in PRs:
https://peterevans.dev/posts/github-actions-how-to-automate-code-formatting-in-pull-requests/

Frenzie: That might not be a bad idea, although I do worry it could be confusing.

math-GH:
It is fine for me how it is now

phpcs.xml
Alkarex: I have reverted those changes for now

Issue #4684, opened 2022-10-04
Ability to add a prefix on a title news


Lordingard:
Hello,

I search an option for adding a prefix to a news title like " 'feed name' : 'news title' ".
This would be useful for example on the relase.atom feeds of GitHub which often only displays the version number of a release without indicating the repository. Information that would however be welcome in a global view.
Does it already exist and if yes, how to enable it?
If not, can you please add the feature in the future? ^^
Thank you!

aledeg:
This is a duplicate of #4335.

math-GH:
Workaround with build-in features:
- name the feed as you want
- display the author

So you have a very good overview of the release feed of GitHub:
![grafik](https://user-images.githubusercontent.com/1645099/193859058-e860d536-d7ee-4ab0-be28-d42efbcf9b20.png)

Lordingard:
Thank you, but I already found it complicated to read on a smartphone with a vertical screen. If I add its columns, it will be even worse. My request is really about modifying particular feeds and not all.

aledeg:
@Lordingard please check attached PR. Thank you!

Alkarex:
@Lordingard Any feedback on https://github.com/FreshRSS/FreshRSS/pull/4728 ?

Lordingard:
> @Lordingard Any feedback on #4728 ?

@Alkarex It's perfect ! Thank you.

PR #4728, opened 2022-10-14
Add title rewriting

### LINKED ISSUE #4335: [Feature] Manipulation of Article Titles
A few of my feeds are provided by a service via my local library and add a [SERVICENAME] prefix to all articles provided by their service. This isn't a major issue but does become tiring over time and hinders readability to see the same block of text reproduced several dozen times each morning.

The ability to add some basic string manipulation (find/replace, trim, ect.) on an on per subscription level would be greatly apricated. 

So far "ignoring", "dealing with", and "putting up with" the issue has provided workable results but may not be viable over long periods of time. I've also reached out to the provider regarding potentially altering their feeds, but since they are repackaging major newspapers I doubt they would be interested in voluntarily removing their stamp on the content. 

**Additional context**
This isn't load bearing or particularly critical by any means, but it would be a nice QoL feature for anyone having to deal with subscriptions that may not be formatted in a user friendly manner. My particular use case is just sourced in minor annoyance, but I could see some potential with this functionality in terms of accessibility.

aledeg:
Closes #4335, #4684

- Rewrite article title upon creation.

1. Add a new title rewriting rule in the feed configuration page.
2. Load new articles.
3. Check that the titles are following the rewrite rule.
4. Enjoy!

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

The title rewriting feature allows title rewriting upon article creation in the database. Once the rewriting is done, it cannot be undone.

At the moment, there is 2 variables and 3 filters available in the simple templating system.

### DIFF
--- a/app/Handlers/TitleRewritingHandler.php
+++ b/app/Handlers/TitleRewritingHandler.php
@@ -1,0 +1,290 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @phpstan-type FilterNames 'ireplace'|'replace'|'trim'
+ * @phpstan-type VariableNames 'feed'|'title'
+ * @phpstan-type SupportedFilter array{name: FilterNames, parameters: int[]}
+ * @phpstan-type Filter array{name: FilterNames, parameters?: array<array-key, string>}
+ * @phpstan-type Rule array{variable: VariableNames, filters?: list<Filter>}
+ */
+class FreshRSS_TitleRewriting_Handler {
+	private const TOKEN_VARIABLE_START = '{';
+	private const TOKEN_VARIABLE_END = '}';
+	private const TOKEN_VARIABLE_FILTER = '|';
+	private const TOKEN_VARIABLE_FILTER_PARAM_START = '(';
+	private const TOKEN_VARIABLE_FILTER_PARAM_END = ')';
+	private const TOKEN_VARIABLE_FILTER_PARAM_DELIMITER = ',';
+	private const TOKEN_SPACE = ' ';
+	private const TOKEN_PARAM_STRING_DELIMITER = '"';
+
+	/** @var list<VariableNames> */
+	private const SUPPORTED_VARIABLES = [
+		'title',
+		'feed',
+	];
+	/** @var list<SupportedFilter> */
+	private const SUPPORTED_FILTERS = [
+		[
+			'name' => 'trim',
+			'parameters' => [
+				0,
+				1,
+			],
+		],
+		[
+			'name' => 'replace',
+			'parameters' => [
+				2
+			],
+		],
+		[
+			'name' => 'ireplace',
+			'parameters' => [
+				2
+			],
+		],
+	];
+
+	/** @var list<Rule> */
+	private array $rules = [];
+
+	public function __construct(string $rules) {
+		$this->parseRules($rules);
+		$this->cleanRules();
+	}
+
+	private function parseRules(string $rules): void {
+		if ($rules === '') {
+			return;
+		}
+
+		$isParsingVariable = false;
+		$isParsingFilter = false;
+		$isParsingString = true;
+		$isParsingParameter = false;
+		$isParsingParameterString = false;
+		$currentSegment = '';
+		$tokens = mb_str_split($rules);
+		$length = count($tokens);
+		$chunks = [];
+		/** @var VariableNames|null */
+		$variable = null;
+		/** @var list<Filter> */
+		$filters = [];
+		/** @var FilterNames|null */
+		$filter = null;
+		/** @var array<array-key, string> */
+		$parameters = [];
+
+		for ($i = 0; $i < $length; $i++) {
+			$current = $tokens[$i];
+			$next = null;
+
+			if (($i + 1) < $length) {
+				$next = $tokens[$i + 1];
+			}
+
+			// Error detection
+			if ($current === self::TOKEN_VARIABLE_FILTER) {
+				if ($isParsingParameter && !$isParsingParameterString) { // @phpstan-ignore-line
+					throw new FreshRSS_Parsing_Exception('Missing filter end delimiter');
+				}
+			} elseif ($current === self::TOKEN_VARIABLE_END) {
+				if ($next === self::TOKEN_VARIABLE_END) {
+					if ($isParsingParameter) {
+						throw new FreshRSS_Parsing_Exception('Missing filter end delimiter');
+					}
+				}
+			}
+
+			// Parsing
+			if ($current === self::TOKEN_PARAM_STRING_DELIMITER) {
+				if ($isParsingParameter) {
+					if ($isParsingParameterString) { // @phpstan-ignore-line
+						$isParsingParameterString = false;
+						continue;
+					}
+
+					$isParsingParameterString = true;
+					$currentSegment = '';
+					continue;
+				}
+			} elseif ($current === self::TOKEN_VARIABLE_START) {
+				if ($next === self::TOKEN_VARIABLE_START) {
+					if ($isParsingString && $currentSegment !== '') {
+						$chunks[] = $currentSegment;
+					}
+					$i++;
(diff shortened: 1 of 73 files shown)


aledeg:
There is known limitations but I can address that later if it's blocking.
There is only 2 variables and 3 filters supported but it's easy to expand the list if needed.

nbgibson:
Ok, that took a bit more doing than I care to admit (entirely on my end), but I got things up and running. I do see the title rewrite field as expected, but I'm a bit stumped on how the rules work at first glance. Can you point me at some documentation on how they work or a quick example? Apologies if this is a known quantity, I've got baby brain in a bad way right now.

nbgibson:
Ok, that took a bit more doing than I care to admit (entirely on my end), but I got things up and running. I do see the title rewrite field as expected, but I'm a bit stumped on how the rules work at first glance. Can you point me at some documentation on how they work or a quick example? Apologies if this is a known quantity, I've got baby brain in a bad way right now.

aledeg:
You can check the documentation in that PR.
If you have hard time to find it. Here is a link on my repo https://github.com/aledeg/FreshRSS/blob/feed/title-rewriting/docs/en/users/04_Subscriptions.md#title-rewriting

nbgibson:
Perfect, thank you. I've applied some rules and am awaiting to see how it goes. We should have some results by the morning.

nbgibson:
First glance through things appear to be working as intended. I did encounter what appeared to be a bug where entering a filter would cause titles to appear as just the article URLs, but I'm working on replicating it. It could be the LAMP stack I have running this being weird/dumb.

nbgibson (approved):
I've spent some time playing around with this feature set and everything appears to be working as expected. I did see a one off bug where titles would be rendered as their source URLs, but I've been unable to replicate it and I suspect it was on the side of my specific setup and not the fault of the code.

aledeg:
If you happen to find out why, let me know. I know there are some issues with some special characters. But we might address that later.

app/Handlers/TitleRewritingHandler.php
```diff
@@ -0,0 +1,272 @@
...
+				2
+			],
+		],
+		[
+			'name' => 'ireplace',
+			'parameters' => [
+				2
+			],
+		],
+	];
```
ColonelMoutarde, line 48: Pleas add the visibility of the constant like "public" in this case

app/Handlers/TitleRewritingHandler.php
```diff
@@ -0,0 +1,272 @@
...
+
+	private $rule = [];
+
+	public function __construct(string $rule) {
+		$this->parseRule($rule);
+		$this->cleanRule();
+	}
+
+	private function parseRule(string $rule) {
```
ColonelMoutarde: Please, split this method for better readability and maintainability

app/Models/Feed.php
```diff
@@ -200,6 +200,13 @@ public function attributes($key = '') {
 		}
 	}
 
+	public function titleRewriteHandler() {
```
ColonelMoutarde, line 203:
```suggestion
	public function titleRewriteHandler(): FreshRSS_TitleRewriting_Handler {
```

app/Handlers/TitleRewritingHandler.php
```diff
@@ -0,0 +1,272 @@
+<?php
+
```
ColonelMoutarde, line 2: suggestion: 
when you add a new class, it is wise to add this at the beginning of the file so that php does not transform the types in an uncontrollable way.

```suggestion

declare(strict_types=1);

```

app/Handlers/TitleRewritingHandler.php
aledeg: Done

app/Handlers/TitleRewritingHandler.php
aledeg: Done

app/Models/Feed.php
aledeg, line 385: I can not do that since this method could return void. With PHP7.2, I can not type-hint the 2 values.

app/Handlers/TitleRewritingHandler.php
aledeg: I am not sure if it will make this more readable. I'll try something.

app/Models/Feed.php
ColonelMoutarde: you can add a php like that:

```suggestion
        /**
        * @return toto|bool
        */
	public function titleRewriteHandler() {
```

tests/app/Handlers/TitleRewritingHandlerTest.php
```diff
@@ -0,0 +1,470 @@
+<?php
+
```
ColonelMoutarde, line 2:
```suggestion

declare(strict_types=1);

```

tests/app/Handlers/TitleRewritingHandlerTest.php
```diff
@@ -0,0 +1,470 @@
...
+class TitleRewritingHandlerTest extends PHPUnit\Framework\TestCase {
+	/**
+	 * @dataProvider provideValidTemplate
+	 */
+	public function test__construct_whenValidTemplate_storeRule($template, $rule) {
```
ColonelMoutarde:
```suggestion
	public function test__construct_whenValidTemplate_storeRule($template, $rule): void {
```

Alkarex:
I had a bit forgotten about that PR, sorry @aledeg :-/
Back then, I postponed it because although the functionality is nice, it looks like the implementation is really huge (relatively to the task).
I have to admit I have not given it much thoughts, but I am wondering whether a few lines of code just allowing a regex-like replace wouldn't do the job, while also being simpler, more powerful, and more standard.

nbgibson:
It would likely be all those things, but it would also be a fairly significant ask on the average end user to learn and implement a Regex filter to achieve the desired result. It would also be a deviation from the existing article filter rule implementation methods.

Unless you mean on the back end and not end user facing. If so then just ignore me.

Alkarex:
> Unless you mean on the back end and not end user facing.

My main concern is indeed the (backend) implementation. But on the frontend, it is quite also usual to have a toggle to allow the more expressive regex syntax. But again, I have to admit that I have not yet spent much time thinking about the current case.

aledeg:
> I had a bit forgotten about that PR, sorry @aledeg :-/ Back then, I postponed it because although the functionality is nice, it looks like the implementation is really huge (relatively to the task). I have to admit I have not given it much thoughts, but I am wondering whether a few lines of code just allowing a regex-like replace wouldn't do the job, while also being simpler, more powerful, and more standard.

@Alkarex don't worry about the delay.

I know the implementation of a parser is quite a lot of code. But I figured that it might be extended in an easier way compared to regex.
I figured that if we make it powerful enough, we could reuse it somewhere else for various purposes. It easy to add modifiers (bold, emphasis, underline, ...) and to improve the parsing.
It could be use as a templating engine to modify article content, or even the application itself. We can also make it extendable to be used in extensions.

If you think it would be better to go some other directions, feel free to close that PR.

As @nbgibson mentioned, I think that offering regex can be tricky for a lot of users, even simple one.

chenyucheng0503:
Can we add the author name to rewrite title?

aledeg:
Probably not as-is. But with some tweaking, it might be possible. I have to make time to work on that.

app/Models/Feed.php
aledeg: Done

tests/app/Handlers/TitleRewritingHandlerTest.php
aledeg: Done

tests/app/Handlers/TitleRewritingHandlerTest.php
aledeg: Done

PR #4736, opened 2022-10-15
Change default values to have predictable default state

### LINKED ISSUE #4713: [BUG] Clicking next to article makes it collapse
Issue:
When clicking next to an open article (and also when selecting text and releasing the mouse button in this area), the article collapses. I'm reporting this as a bug since there is no visual feedback ("click"-cursor) for this, unlike on the article header.

1. Open article
2. Click on the margin of the content < div> (see screenshot)
3. Article collapses as if its header was clicked

Expected behavior:
Nothing should happen.

image

Environment information (please complete the following information):
- Device: PC with a large enough screen
- OS: any (tested on Windows 10 and EndeavourOS)
- Browser: any (tested on Chrome 106 and Firefox 105)
- FreshRSS version: 1.20.0
- PHP version: 7.4.30

Suggested solution:
Make only the < li> "item title" clickable for the collapsing behaviour and not the whole < div> "flux active current" as I guess it is the case now.

PS: thanks for your great work, FreshRSS is an awesome piece of software!

aledeg:
Closes #4713

- Change default values to have predictable default state

1. Create a new user and check its configuration

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Alkarex (approved):

config-user.default.php
```diff
@@ -41,7 +41,7 @@
 	'hide_read_feeds' => true,
 	'onread_jump_next' => true,
 	'lazyload' => true,
-	'sides_close_article' => true,
+	'sides_close_article' => false,
```
math-GH, line 44: "Clicking outside of article text area closes the article"

config-user.default.php
```diff
@@ -53,13 +53,13 @@
 	'sort_order' => 'DESC',
 	'anon_access' => false,
 	'mark_when' => array (
-		'article' => true,
+		'article' => false,
```
math-GH: "when the article is viewed"

config-user.default.php
```diff
@@ -53,13 +53,13 @@
...
 		'gone' => false,
 		'max_n_unread' => false,
 		'reception' => false,
 		'same_title_in_feed' => false,
-		'scroll' => true,
-		'site' => true,
+		'scroll' => false,
+		'site' => false,
```
math-GH: " while scrolling "
"when the article is opened on its original website "

config-user.default.php
math-GH: I like it

config-user.default.php
math-GH: I do not like this change

config-user.default.php
math-GH: first: I like
second: I do not like this change

config-user.default.php
Frenzie: I think true is probably a more expected default for this one, yeah.

config-user.default.php
Frenzie: Having the second as true may be worse than the outside of text area setting. You know you're clicking on it, but you may not even realize you're accidentally marking things as read in the background.

But you could make an argument that it's similar to opening the article within the FreshRSS UI.

config-user.default.php
Alwaysin: I agree

config-user.default.php
Alwaysin: No strong opinion on this one but I join math, I think it makes more sense to mark it as read when you open the article on the original website.

config-user.default.php
aledeg: Fixed

config-user.default.php
aledeg: fixed

Frenzie (approved):

PR #4741, opened 2022-10-18
Added bottom nav padding for iOS #4740

### LINKED ISSUE #4740: [BUG] Feed navigation overlaps safe area in iOS
**Describe the bug**
When FreshRSS is used as an app within iOS (using 'Add to Homescreen' from Safari), the bottom navigation bar overlaps with the swipe up for changing applications. Because of this the navigation buttons can be hard to use unless you use a theme with large buttons.

**To Reproduce**
1. In Safari, use 'Add to Home Screen' from the 'Share' menu
2. Open FreshRSS using the new item on the home screen
3. Try to use the navigation buttons at the bottom. They only work if you manage to press them above the safe area

**Expected behavior**
The bottom navigation should have sufficient padding on iOS to push it above the safe area

**Screenshots**
WIN_20221018_13_32_22_Pro

- Device: iPhone 12
- OS: iOS 16.0.2
- Browser: N/A
- FreshRSS version: 1.20.0
- Installation type: Docker

**Additional context**
This should be able to be fixed using ```padding-bottom: env(safe-area-inset-bottom);```

That should add enough padding for iOS without affecting other devices.

acbgbca:
Closes #4740 

- Added additional bottom padding to the navigation when required. This means the navigation no longer overlaps the safe area in iOS. The viewport meta tag changes were required for safe-area-inset-bottom to have a value.

1. Add FreshRSS to the iOS Home screen
2. Open the application
3. Note that the menu no longer overlaps the safe area

I have observed the following based on my testing:
* Desktop browser - no change
* Safari in iOS, portrait - no change (iOS has the URL bar at the bottom which covers the safe area)
* Safari in iOS, landscape - The bottom navigation has the padding added pushing it above the safe area
* iOS as home screen app - The navigation has the padding addedd
* Android - Unfortunately I have no android device on which to test. However unless the device specifically sets 'safe-area-inset-bottom' it should have no change

WIN_20221018_13_48_43_Pro

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -10,7 +10,7 @@ if (_t('gen.dir') === 'rtl') {
 ?>>
 	<head>
 		<meta charset="UTF-8" />
-		<meta name="viewport" content="initial-scale=1.0" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
 		<?= FreshRSS_View::headStyle() ?>
 		<script id="jsonVars" type="application/json">
 <?php $this->renderHelper('javascript_vars'); ?>
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1454,6 +1454,7 @@ a.website:hover .favicon {
 	bottom: 0; left: 0;
 	width: 300px;
 	table-layout: fixed;
+	padding-bottom: env(safe-area-inset-bottom);
 }
 
 #nav_entries .item {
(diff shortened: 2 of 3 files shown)


math-GH:
Should be fine. Cannot test it because I do not have a iPhone device.

Alkarex:
> Should be fine. Cannot test it because I do not have a iPhone device.

The main thing is to make sure there is no regression on other platforms

math-GH:
> > Should be fine. Cannot test it because I do not have a iPhone device.
> 
> The main thing is to make sure there is no regression on other platforms

I checked it in a very quick try out on my mobile phone (Android, with a standard rectangle screen). It is fine. But I cannot check if it really works ;)

Alkarex:
Ref https://developer.mozilla.org/docs/Web/CSS/env

PR #4750, opened 2022-10-19
Fix 4712

### LINKED ISSUE #4712: UI regression icons click areas
The click area for the icons such as mark-as-read / favourite have become much smaller, generating many wrong clicks.

This seems to be due to a mislocated padding rule. Regression probably due to https://github.com/FreshRSS/FreshRSS/pull/4671

In FreshRSS 1.20.0:

image

In edge:

image

math-GH:
Closes #4712

- padding

see #4712

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -16,7 +16,7 @@
 				if ($this->entry->isRead()) {
 					$arUrl['params']['is_read'] = 0;
 				}
-				?><a class="read" href="<?= Minz_Url::display($arUrl) ?>" title="<?= _t('conf.shortcut.mark_read') ?>"><?php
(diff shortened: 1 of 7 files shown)


Alkarex:
Better, but not quite the full area as it was before. There is still some margin/padding at the top and bottom, which would be nice to remove:

image

math-GH:
Is it just the Pafat theme?

Alkarex:
There is also a bit of padding/margin in e.g. Origine, although a bit less than in Pafat

image

math-GH:
Pafat is now fixed.

To make it clear: This is just a fix to make it work again. The root cause of all this issues are much deeper and it needs a refactoring of all themes.

Alkarex:
> To make it clear: This is just a fix to make it work again. The root cause of all this issues are much deeper and it needs a refactoring of all themes.

Thanks, and that is fine for now. I would just like to fix the regressions before releasing 1.20.1

PR #4770, opened 2022-10-24
Fixed: "tag" -> "label"

math-GH:
Before:
When a label is created the notification calls it a "tag".
grafik

Edit user queries:
grafik

* English i18n: "label" -> "tag"
* German: improved
* some other languages: commented the line as `// DIRTY`

1. check how it is called in the language:
2. go to "Label management" (`./p/i/?c=tag`)

1. Create a new label
3. see the notification

1. Create/edit a user query
2. see the filter list

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/i18n/nl/conf.php
+++ b/app/i18n/nl/conf.php
@@ -102,7 +102,7 @@ return array(
 			'order' => 'Sorteren op datum',
 			'search' => 'Expressie',
 			'state' => 'Status',
-			'tags' => 'Weergeven op tag',	// DIRTY
+			'tags' => 'Weergeven op label',
 			'type' => 'Type',	// IGNORE
 		),
 		'get_all' => 'Toon alle artikelen',
--- a/app/i18n/nl/conf.php
+++ b/app/i18n/nl/conf.php
@@ -102,7 +102,7 @@ return array(
 			'order' => 'Sorteren op datum',
 			'search' => 'Expressie',
 			'state' => 'Status',
-			'tags' => 'Weergeven op tag',
+			'tags' => 'Weergeven op tag',	// DIRTY
 			'type' => 'Type',	// IGNORE
 		),
 		'get_all' => 'Toon alle artikelen',
--- a/app/i18n/nl/feedback.php
+++ b/app/i18n/nl/feedback.php
@@ -115,9 +115,9 @@ return array(
 		'purge_completed' => 'Opschonen klaar (%d artikelen verwijderd)',
 	),
 	'tag' => array(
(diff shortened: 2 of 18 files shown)


math-GH:
Ping @Frenzie : Please check the Dutch. I labeled some lines as DIRTY

aledeg:
Shouldn't we also modify the action to reflect that?

aledeg:
I think, this modification should be added also on the translation key to avoid confusion.

math-GH:
> Shouldn't we also modify the action to reflect that?

It would be a bigger refactoring. Also the i18n keys need a renaming.

Here I just wanted to fix the text for 1.20.1

aledeg:
Ok then. But could you open an issue to reflect the changes I was mentioning earlier?

aledeg (approved):

app/i18n/nl/conf.php
```diff
@@ -102,7 +102,7 @@
 			'order' => 'Sorteren op datum',
 			'search' => 'Expressie',
 			'state' => 'Status',
-			'tags' => 'Weergeven op tag',
+			'tags' => 'Weergeven op tag',	// DIRTY
```
Alkarex: Maybe @Frenzie could have a quick fix there?

Alkarex (approved):

app/i18n/nl/conf.php
Frenzie: You can just replace tag with label. (Outside of software _etiket_ might be better.)

Frenzie:
```suggestion
			'tags' => 'Weergeven op label',
```

app/i18n/nl/feedback.php
```diff
@@ -115,9 +115,9 @@
 		'purge_completed' => 'Opschonen klaar (%d artikelen verwijderd)',
 	),
 	'tag' => array(
-		'created' => 'Tag „%s” is aangemaakt.',
-		'name_exists' => 'Tagnaam bestaat al.',
-		'renamed' => 'Tag „%s” hernoemd naar „%s”.',
+		'created' => 'Tag „%s” is aangemaakt.',	// DIRTY
+		'name_exists' => 'Tagnaam bestaat al.',	// DIRTY
+		'renamed' => 'Tag „%s” hernoemd naar „%s”.',	// DIRTY
```
Frenzie:
```suggestion
		'created' => 'Label „%s” aangemaakt.',
		'name_exists' => 'Label bestaat al.',
		'renamed' => 'Label „%s” hernoemd naar „%s”.',
```

PR #4775, opened 2022-10-25
fix #4772

### LINKED ISSUE #4772: Bottom line UI regression
The bottom line of text and icons (tested in Origine and Pafat) has gotten too low (not enough margin with the line just below)

image

math-GH:
Closes #4772

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Better, although I believe the root cause might have been something else. The Pafat theme has become more compact (probably less margin at the bottom of each line, maybe also related to the fact that the number of articles for each category are not vertically centred anymore)

1.20.0
image

Edge
image

As this remaining regression seems limited to a single theme, I will open a new issue

PR #4780, opened 2022-10-26
NFS-friendly is_writable() checks

### LINKED ISSUE #4779: [BUG] check if (data) directory is writable fails on NFS
After installing FreshRSS in my environment I had several problems with features blocked due to "missing write permissions" of my __data__ directory. This happened whenever my data directory was located on a mounted __NFS__ share. The problem occured in several places of the application:

- during the initial setup using the web interface
- in the _"Installation check"_ view
- when calling _"app/actualize_script.php"_ 

It is possible, that other parts of the application and other checked directories are also affected.

**To Reproduce**
1. install FreshRSS
2. use a NFS mount for the data directory
3. try to open  _"Installation check"_ or to run  _"app/actualize_script.php"_ 

**Expected behavior**
If the permissions are set correctly and the application is able to write in the data directory the check should not fail.

- Host-OS: CentOS Stream 8
- FreshRSS version: freshrss/freshrss:edge (0c13c07b9581)
- NFS-Server: Synology

**Root cause analysis and possible solution**
I was able to track the problem down to the PHP function _is_writable_, which is known not to work on NFS shares. Nextcloud had a similar problem a few years ago [ Php v7 filesystem check bug with nfs v3 mount from synology #7124 ](https://github.com/nextcloud/server/issues/7124).

They solved the problem by trying to write an actial file, if _is_writable_ fails. [Commit with the workaround in Nextcloud](https://github.com/nextcloud/server/commit/ff4dc3b66b03c7153c26b5313a639bbd114dfb46)

Based on that I already tested to solve the problem using this function:

```php
// workaround, as is_writable does not work on NFS shares
function checkWritable($path) {
	if (!is_writable($path)){
		$testFile = sprintf('%s/%s.tmp', $path, uniqid('data_dir_writability_test_'));
		$handle = fopen($testFile, 'w');
		if (!$handle || fwrite($handle, 'Test write operation') === FALSE) {
			return false;
		} else {
			fclose($handle);
			unlink($testFile);
		}
	}
	return true;
}
```
and then replaced _is_writable()_ with _checkWritable()_ in various places, what worked fine for me.

As I'm not a good PHP developer, I have not created a pull request yet. 

If you say that bug should be solved, give me an advice in which file the wrapped function should live (I used _lib/lib_install.php_ during my tests) and if replacing all occurrences of _is_writable()_, even if I had no problems with the part of the code so far, seems to be a good idea, I'm happy to provide the pull request soon.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4779

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -14,7 +14,7 @@ class FreshRSS_update_Controller extends FreshRSS_ActionController {
 	public static function migrateToGitEdge() {
 		$errorMessage = 'Error during git checkout to edge branch. Please change branch manually!';
 
-		if (!is_writable(FRESHRSS_PATH . '/.git/')) {
+		if (!is_writable(FRESHRSS_PATH . '/.git/config')) {
 			throw new Exception($errorMessage);
 		}
 
@@ -118,7 +118,7 @@ class FreshRSS_update_Controller extends FreshRSS_ActionController {
 			if ($version == '') {
 				$version = 'unknown';
 			}
-			if (is_writable(FRESHRSS_PATH)) {
+			if (touch(FRESHRSS_PATH . '/index.html')) {
 				$this->view->update_to_apply = true;
 				$this->view->message = array(
 					'status' => 'good',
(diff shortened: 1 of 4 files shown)


Alkarex:
P.S. *touching* the `index.html` files is actually a minor positive feature, as it ensures they exist, which is good for avoiding listing directories on badly configured servers.

PR #4792, opened 2022-10-30
improved: Pafat theme aside bar

math-GH:
Before:
grafik

After:
1 + 2: same space between the blocks
3: a bit more space around the button's text
grafik

Before:
grafik

After:
1: if active category then the unread counter has the same darker blue like the category
2: if category is not active than it has the same (lighter) blue like the category
3: text cut improved
grafik

- CSS

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Thanks for all this great work

PR #4796, opened 2022-10-31
fix: Pafat theme: line-height

math-GH:
Follow up of https://github.com/FreshRSS/FreshRSS/issues/3733. This PR is for Pafat theme

It improves also a bit the input styles. Now it looks a bit more beautiful:

Before:
grafik

After:
grafik

- CSS: `line-height` without units

1. use Pafat theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Pafat/pafat.css
+++ b/p/themes/Pafat/pafat.css
@@ -455,7 +455,6 @@ a.btn {
 .tree-folder.active .tree-folder-title {
 	background: #39b3d7;
 	font-weight: bold;
-	font-size: 1rem;
 	border-top: 1px solid #666;
 	border-bottom: 1px solid #666;
 }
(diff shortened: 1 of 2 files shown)


math-GH:
This PR will make @Alkarex (who like Pafat theme) and @Frenzie (who does not like units for `line-height`) happy ;)

Frenzie:
Seldom, not never. ;-) (But yes, these don't look like places where it makes much sense to use px.)

Alkarex:
Nice improvements, thanks 👍🏻

Alkarex:
It looks like the number badges are not centred anymore.

Before:
image

After:
image

math-GH:
fixed

Alkarex:
Looks good 👍🏻 
One final detail: the font size of the main view has gotten a bit too small in my opinion (should be a bit larger than the other categories):

Before:
image


After:
image

math-GH:
> Looks good 👍🏻 One final detail: the font size of the main view has gotten a bit too small in my opinion (should be a bit larger than the other categories):

I removed the bigger font-size because:
- The active "category" had a bigger font size (not only the main stream)
- the bigger font size leads to a higher category box

Should I revert it?

Alkarex:
Ah, you are right, it is indeed the active category, which is larger. Let's say it is good enough for now (I would like a bit more emphasis on the current view, e.g. larger font or other visual hints)

PR #4693, opened 2022-10-05
Improved: Origine theme (colors in vars)

math-GH:
Ref. #4642

grafik

Changes proposed in this pull request:

- CSS: colors in variables
- CSS: some little improvements

How to test the feature manually:

1. use Origine theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

Issue #4515, opened 2022-08-21
Feature Request: autom. dark mode


math-GH:
let's collect all feature requests around a autom. dark mode here.

Alkarex:
* https://github.com/FreshRSS/Extensions/pull/140
* https://github.com/FreshRSS/FreshRSS/discussions/4508

fireglow:
I'm very interested in this, for anybody looking into this, [here](https://addons.mozilla.org/en-US/firefox/addon/dark-mode-website-switcher/) is a firefox addon that makes switching prefers-color-scheme very easy, just one click.

Alkarex:
@fireglow This is already implemented in our edge branch. Give it a try :-)

Alkarex:
P.S.: Firefox supports automatic dark/light mode switch natively without extension, if you just pick the "auto" theme in Firefox, which will obey OS-level preferences.
E.g. for Windows, see https://github.com/AutoDarkMode/Windows-Auto-Night-Mode

Alkarex:
* https://github.com/FreshRSS/FreshRSS/pull/5582

Issue #2793, opened 2020-02-06
[Suggestion] option for automatic time-based dark theme


Draky50110:
Hello.

Would it be possible to add an option to switch auto on dark theme or light theme based on time of the day ?
Both theme (light and dark) should be a choice also.

Thanks.

Jucgshu:
Some websites can now follow the systems settings. I find it elegant & handy.

You can then delegate the auto switch to the program of your choosing (for Gnome, see https://extensions.gnome.org/extension/2236/night-theme-switcher/ for example).

math-GH:
#4515

Alkarex:
https://github.com/FreshRSS/Extensions/pull/140

aledeg:
As there is an extension to handle that, could we close this issue?

Jucgshu:
Even better: it's a bundle option in FreshRSS for quite some time now. So yes, this issue could be closed now.

PR #4843, opened 2022-11-13
Added: Dark mode for Origine +Origine compact themes

math-GH:
Ref #4515
Ref #2793
Follow up of #4693

This PR shows how easy it is to use the variables to create cool new stuff.

This themes have now a dark mode:
- Origine
- Origine compact

grafik

- media query that ask for the browsers preferences (CSS)

1. if you have a dark mode in your browser enabled: you see the dark theme
2. if you have NOT a dark mode in your browser enabled: you see the (default) light theme

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
This PR is a draft PR to collect your feedback, that is very welcome

Frenzie:
Sure, why not. Are the colors based on anything in particular?

math-GH:
>  Are the colors based on anything in particular?

I tried to invert the white/grey colors to black/dark grey.
Used the same blue color like in the light theme.
Then I tried to modify the dark colors to keep the character of the Origine theme.

I am open for ideas of another color values.

math-GH:
I had to change a bit the blue color for a better reading on the dark background.

New added:
The new option in the theme configuration:
grafik

app/i18n/fr/conf.php
```diff
@@ -32,6 +32,7 @@
 	),
 	'display' => array(
 		'_' => 'Affichage',
+		'darkMode' => 'Automatically dark mode (Beta)',	// TODO
```
```suggestion
		'darkMode' => 'Mode sombre automatique (bêta)',
```

Frenzie: The original should be `Automatic dark mode (beta)` btw.

Automatic for grammar, beta for style.

config-user.default.php
```diff
@@ -62,6 +62,7 @@
 		'site' => true,
 	),
 	'theme' => 'Origine',
+	'darkMode' => false,
```
Alkarex: Do we maybe want three states: { yes, no, automatic } ?

Frenzie: Agreed, probably with automatic as the default?

math-GH: Maybe later.
`yes` would mean to find another solution than the media query `@media (prefers-color-scheme: dark)`.

Alkarex: There is a simple approach on https://stackoverflow.com/questions/56300132/how-to-override-css-prefers-color-scheme-setting/68824350#68824350
But it requires having two copies of the variables, which is not super elegant.

Alkarex: I hope we can find a satisfying approach for { auto, yes, no, }, but if we do not implement *yes* for now, I think in any case we should have { auto, no, }, default being auto.

Frenzie: Definitely not a hold-up from my end; I was just agreeing it'd be nice to have. The SO link there looks reasonably elegant; I can't think of anything better.

math-GH: yes, this is the only solution, as far as I know.
With SCSS we could keep the maintenance effort small.

math-GH: `no` and `auto` is now implemented

math-GH:
`No` and `auto` is now implemented as select list instead a checkbox.
Please give feedback

Todo: i18n String for `No` and `Auto`.

Out of scope of this PR: `yes` mode

Frenzie:
Code part lgtm

math-GH:
What could be the best i18n String?
label: `Automatic dark mode (beta)` => `Dark mode (beta)`
`no` = `No automatic dark mode`
`auto` = `Use the browser preference`? `Automatic dark mode`?

Frenzie:
Why beta at this point? Because it only applies to one theme?

Anyway, what about:
`Appearance`/`Color scheme`

* `Browser theme`/`Browser color scheme` ("auto")
* `Light` ("no")
* `Dark` (potential future "yes")

math-GH:
> Why beta at this point? Because it only applies to one theme?

yes. Only Origine/Origine compact will have this dark color scheme at the moment. Two themes of 13.
Each dark theme needs some testing period for fine tuning.

And we need to think about our 4 dark themes (`Alternative dark`, `dark`, `dark pink` and `Nord`). How to handle them?

That is why I would like to label it as "beta".

> 
> Anyway, what about: `Appearance`/`Color scheme`
> 
>     * `Browser theme`/`Browser color scheme` ("auto")
> 
>     * `Light` ("no")
> 
>     * `Dark` (potential future "yes")

I like this idea

Frenzie:
> And we need to think about our 4 dark themes (Alternative dark, dark, dark pink and Nord). How to handle them?

I'd propose simply changing the screenshots, which are probably outdated and ripe for an update anyway, showing both variants with a split in the middle. Possibly with one of those fancy sliders like [this](https://levelup.gitconnected.com/how-to-create-a-before-after-image-slider-with-css-and-js-a609d9ba77bf). I don't think having a theme without variants is problematic, but if you want to give them all a light variant that's your prerogative. ;-)

math-GH:
Let us merge this PR and collect experiences with this first dark theme.

Alkarex:
Sorry for the delay

PR #4561, opened 2022-08-30
Typing function checkToken

ColonelMoutarde:
Type hinting in function

PR #4855, opened 2022-11-15
Fix type hints regressions

### LINKED ISSUE #4854: [BUG] greader API latest Docker Dev Build
**Describe the bug**
After installing the latest ARM Docker Dev build 1.21-0-Dev an the previous build, my third-party Reeder application on MacOS and iOS no longer works/syncs to the application.  I do not see anything when I look at the logs for FreshRSS, and the Reeder app does not have any logs I can see.  When I try to sync, it immediately completes the sync as if there is not a connection.   Is there a way test or validate the API that previously worked on the earlier docker builds. With the latest I was able to sync once but viewing in the Reeder Application did not update the application with any changes.  

UPDATE: I changed the API password in the FreshRSS web UI and updated it in the reeder application.  Once I did that, I was able to start a sync as normal.  However, the mark as read and any sync attempts afterwards fail.  I can get the same results if I change the API password again, but the connection seems lost after the first connection.  

**To Reproduce**
Attempting to sync with the gReader API no longer works with Reeder App after latest upgrade to the Docker ARM Dev package

**Expected behavior**
Two way sync and marking as read as previously functioned in earlier builds

**Screenshots**

-Reeder App on both iOS, PadOS and MacOS

**Additional context**
I can try to rebuild the API connection as if new to see if this corrects the problem.  Happy for any other recommendations.

Alkarex:
Fix regressions from https://github.com/FreshRSS/FreshRSS/pull/4561 , which added some hard type hints but without ensuring that the code paths would provide strictly identical class types.

Example:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to checkToken() must be an instance of FreshRSS_UserConfiguration, instance of Minz_Configuration given, called in /var/www/FreshRSS/p/api/greader.php on line 1091 and defined in /var/www/FreshRSS/p/api/greader.php:223
Stack trace:
#0 /var/www/FreshRSS/p/api/greader.php(1091): checkToken()
#1 {main}
  thrown in /var/www/FreshRSS/p/api/greader.php on line 223
```

I refactored the way we access our configuration to have stricter types, improvement of https://github.com/FreshRSS/FreshRSS/pull/4110
More work could be done to simplify the way we work with configurations.

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -217,7 +217,7 @@ class FreshRSS_update_Controller extends FreshRSS_ActionController {
 	}
 
 	public function applyAction() {
-		if (Minz_Configuration::get('system')->disable_update || !file_exists(UPDATE_FILENAME) || !touch(FRESHRSS_PATH . '/index.html')) {
+		if (FreshRSS_Context::$system_conf->disable_update || !file_exists(UPDATE_FILENAME) || !touch(FRESHRSS_PATH . '/index.html')) {
 			Minz_Request::forward(array('c' => 'update'), true);
 		}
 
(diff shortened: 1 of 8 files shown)


Alkarex:
Quick merge to ease testing in https://github.com/FreshRSS/FreshRSS/issues/4854

PR #4859, opened 2022-11-16
Experiment Development Containers / GitHub Codespaces

Alkarex:
https://containers.dev/
https://github.com/features/codespaces
Allows having a full development environment, fully configured and ready for coding in one click, either in the Web browser or locally (e.g., in Visual Studio Code)

PR #4860, opened 2022-11-16
Fix Development Containers / GitHub Codespaces

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/4859


Alkarex:
Needs merging to test. Discussion in https://github.com/FreshRSS/FreshRSS/discussions/4861

PR #4865, opened 2022-11-17
Fix regression on header size due to padding

nicofrand:

- Fix a regression on the origine-compact theme after https://github.com/FreshRSS/FreshRSS/commit/9621ab17479425845e66c2533817e6dcfe20feec#diff-892599b0ad323507d983e1465de5959f0008f4b38fc5f0b2afd196f487f2fd93 : the window now has a vertical scrollbar because the header is bigger than the expected 40px due to an unexpected padding (`.header > .item.title` having precedence on `.header > .item`.

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


math-GH:
I do not understand the issue. Could you please support with a screenshot?

nicofrand:
On origin-compact there is a vertical scrollbars because the header now 
takes 40px + the padding on .item-title instead of just 40px.

math-GH:
I optimized it a bit. Could you please check the result?

Alwaysin:
Once again a discrepancy in zoom level on Windows I guess beacause I don't see the issue. This commit only makes the header a bit less compact to me (it does not bother me at all).

nicofrand:
I am not zooming nor using Windows…

See: 

image

nicofrand:
> I optimized it a bit. Could you please check the result?

That's not good, this needs to be applied on .item too (not only configure but search).

math-GH:
let's step one behind.
I see an issue, that a clear bug description is missed.

Let me do it:
----------------

Affected theme: Origine compact theme

FreshRSS version: 1.20.1

Current behavior:
The header is higher than 40px
grafik

That leads to a always visible scrollbar, because the `#global` calculates with 40px from header.
grafik


Expected behavior:
Header height = 40px
No scrollbar, when it is not needed

nicofrand:
OK I'll take that one but I am not the one who amended the merge request. 
The fix I provided works fine.

math-GH:
Now it should work fine. I forget yesterday to use advantage of the optimization.

math-GH:
> OK I'll take that one but I am not the one who amended the merge request. The fix I provided works fine.

I am happy that you found this issue and provided a fix that worked in general. Thanks for your contribution.

While I checked your PR I saw some basic issues that came from the Origine theme and made the fix of Origine compact theme a bit less clear. And I saw, that header height = 40px was missed.

Now the issue should be solved on the root

PR #4794, opened 2022-10-30
Improved: implement CSS variable --frss-padding-top-bottom

math-GH:
Similar to #4793 

Now it is possible to use the CSS variable `--frss-padding-top-bottom` to tweak the y-axis.
Default: `0.5rem`
For the compact view (f.e. Origine Compact theme): `0.125rem`

It is the next step forward to implement a compact option for all themes.

What is affected:
- the header
- normal/reader view aside navigation
- navigation buttons bar
- new article banner

Changes proposed in this pull request:

- CSS

How to test the feature manually:

1. check the themes
2. especially: Origine theme: non-compact
3. especially: Origine compact theme: compact

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #4900, opened 2022-11-27
Make header logo take the whole height

christophehenry:

- CSS to make header logo take the whole height
- Having big buttons is always nice, it prevents misclicks
- Screenshot_20221127_182002

### DIFF
--- a/p/themes/base-theme/base.css
+++ b/p/themes/base-theme/base.css
@@ -394,13 +394,18 @@ a.btn {
 }
 
 .header > .item {
-	padding: 10px;
 	vertical-align: middle;
 	text-align: center;
 }
 
 .header > .item.title {
 	width: 230px;
+}
+
+.header .item.title a {
+	padding: 1rem;
+	display: block;
+	height: 100%;
 }
 
 .header > .item.title h1 {


p/themes/base-theme/base.css
```diff
@@ -394,11 +394,16 @@ a.btn {
 }
 
 .header > .item {
-	padding: 10px;
 	vertical-align: middle;
 	text-align: center;
 }
 
+.header a[href="./"] {
```
Frenzie: I'd rather make that something like `.header > .item.title > a`. Incidentally, is that `.header > .item.title h1` stuff still relevant?

christophehenry: It's not an `h1` it's a `div`. If it's certain that only the logo link will be enclosed in a `.title`, then yes, `.header > .item.title > a`  is suitable. I saw `a[href="./"]` used everywhere else in the CSS.

Frenzie: > It's not an h1 it's a div.

Exactly, hence the question. ;-) I'm asking if those are dead styles that should be removed.

> I saw a[href="./"] used everywhere else in the CSS.

Everywhere is a strong word. It looks a tad hackish to me, used in a theme or two (not coincidentally by the same author :-) ) when the HTML should probably just be improved instead, generally with `class=…`, possibly more structurally.

math-GH:
It does not make sense to change it in the `base.css` that has no influence on the themes.

The refactored themes use for this padding a variable.

In general I am not fan of changing it as purposed in this PR. The logo is big enough and it is not a good UX to extend the clickable area around very much.

Frenzie:
I agree 10px is a bit much, though I do think extending it a couple of pixels would map better to where it feels like it's clickable compared to where it actually is. But then again, as you said it's a pretty large area already and I certainly never noticed. ^_^

christophehenry:
> It does not make sense to change it in the base.css that has no influence on the themes.

Well, I'm using Pafat, and it works for that theme. Besides, I've tested for every theme available to me. I didn't break anything. So I would argue otherwise.

> But then again, as you said it's a pretty large area already and I certainly never noticed. ^_^

Have the link logo mapping the whole header is kind of accessibility good practice.

math-GH:
> > It does not make sense to change it in the base.css that has no influence on the themes.
> 
> Well, I'm using Pafat, and it works for that theme. Besides, I've tested for every theme available to me. I didn't break anything. So I would argue otherwise.
> 

Believe me, `p/themes/base-theme/base.css` (that is the file you change within this PR) has no effect on the themes. `base.css` is just a template to copy&paste as starting point for an own theme.

Frenzie:
> Have the link logo mapping the whole header is kind of accessibility good practice.

I was just trying to find a middle ground and it doesn't describe your 10px, and I'll leave it at that. ^_^

christophehenry:
> I was just trying to find a middle ground and it doesn't describe your 10px, and I'll leave it at that. ^_^

Oh, the 10px padding is just the padding the header had before. That doesn't change the size of anything. Only the padding will be ported by the link now. I don't really care about the size of the padding. The logo link just has to take 100% the height of the header and be big enough to prevent misclicks, that it :wink:

math-GH:
> be big enough to prevent misclicks, that it 

which "misclicks" do you mean?

math-GH:
This would solve a lot of discussion here (for Pafat theme)

(replace the margin with padding:)
```
.header > .item.title .logo {
  padding: 0.5em 0;
}
```

If you want to have it in all themes: It will not be easy, because Origine compact and Swage theme will not have space around the logo. Please keep #4794 in mind.

Frenzie:
@christophehenry I understand what you're saying, but I don't think you're looking at the whole picture or you're not expressing what you mean accurately. You don't want tapping a bit too high while aiming for subscriptions or user queries to result in a tap on the logo. That is why your 10px clickable area expansion is basically fine even though it's definitely on the larger side, precisely because it *doesn't* take up the whole header.

christophehenry:
Ah yes indeed. Then I propose to increase the `.aside.aside_feed` padding too. Something like `20px` could be suitable? Or better: `1rem`, which makes more sense when talking about padding and margin?

Frenzie:
That would definitely not be my preference. If anything I'd severely shrink the header if it's felt that the whole header should be clickable.

And yes, some .2rem to .5rem is what I'd use as the expansion of the clickable area. It's intended to capture minor missed, not to create them.

christophehenry:
> That would definitely not be my preference. If anything I'd severely shrink the header if it's felt that the whole header should be clickable.

I would argue against this. Some themes use a compact layout with a small header and it's ok but in general, this increases the chances of misclicks. The logo has become the go-to place to come back to the home these last few years. It's important it is easily reachable.

A compromise could be found using CSS variables and a "compact layout" option common to all themes but I think this is out of scope.

Frenzie:
Like I said, even though I think it should be less, 10px is fine in practice and doesn't cause issues. It's specifically your description that doesn't actually describe the PR that I would be opposed to.

christophehenry:
Changes done.

PR #4906, opened 2022-11-28
User-defined time zone

### LINKED ISSUE #2754: Add Time zone options for each user
Currently, only time zone settings in php.ini can be used, which is very bad, and you should be able to set the global time zone at config.php, as well as each user setting the time zone separately.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/2754

*Default* is server / PHP default, e.g. as defined by `TZ` in our Docker images, or in `php.ini` `date.timezone`

image

### DIFF
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -99,6 +99,9 @@ class FreshRSS extends Minz_FrontController {
 
 		Minz_Session::_param('language', $language);
 		Minz_Translate::init($language);
+
+		$timezone = isset(FreshRSS_Context::$user_conf) ? FreshRSS_Context::$user_conf->timezone : '';
+		date_default_timezone_set($timezone != '' ? $timezone : ini_get('date.timezone'));
 	}
 
 	private static function getThemeFileUrl($theme_id, $filename) {
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -101,7 +101,7 @@ class FreshRSS extends Minz_FrontController {
 		Minz_Translate::init($language);
 
 		$timezone = isset(FreshRSS_Context::$user_conf) ? FreshRSS_Context::$user_conf->timezone : '';
-		date_default_timezone_set($timezone != '' ? $timezone : ini_get('date.timezone'));
+		date_default_timezone_set($timezone != '' ? $timezone : '' . ini_get('date.timezone'));
 	}
 
 	private static function getThemeFileUrl($theme_id, $filename) {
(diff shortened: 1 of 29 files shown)


Alkarex:
image

app/i18n/nl/conf.php
```diff
@@ -57,6 +57,7 @@
 			'portrait' => 'Staand',
 			'square' => 'Vierkant',
 		),
+		'timezone' => 'Time zone',	// TODO
```
```suggestion
		'timezone' => 'Tijdzone',
```

Frenzie (approved):
Looks straightforward, can't think of any negatives.

aledeg:
Correct me if I'm wrong, but for what I can gather from the code, there is no use of the timezone elsewhere than in the configuration. Do you plan on adding some code to utilize it in the interface or am I missing something?

app/FreshRSS.php
```diff
@@ -99,6 +99,9 @@ private static function initI18n() {
 
 		Minz_Session::_param('language', $language);
 		Minz_Translate::init($language);
+
+		$timezone = isset(FreshRSS_Context::$user_conf) ? FreshRSS_Context::$user_conf->timezone : '';
+		date_default_timezone_set($timezone != '' ? $timezone : ini_get('date.timezone'));
```
Alkarex: @aledeg See here for the actual use

Alkarex:
> Correct me if I'm wrong, but for what I can gather from the code, there is no use of the timezone elsewhere than in the configuration. Do you plan on adding some code to utilize it in the interface or am I missing something?

See the `date_default_timezone_set()` line for the actual use. It changes the time representations in the UI

app/FreshRSS.php
aledeg: Thank you. I missed that part.

aledeg (approved):

Alkarex:
I have just added the possibility to pick the timezone when creating a new user:

image

Alwaysin:
@Alkarex I think this PR led to a notice in PHP at each page load:
```
AH01071: Got error 'PHP message: PHP Notice:  date_default_timezone_set(): Timezone ID '' is invalid in /var/www/html/freshrss/app/FreshRSS.php on line 104'

```
Is this normal? I guess I just have to set a timezone in my PHP conf which has none by default, or select one in FreshRSS conf?

Alkarex:
@Alwaysin Please try https://github.com/FreshRSS/FreshRSS/pull/5021

PR #4943, opened 2022-12-13
Help needed: $test2 does not work

math-GH:
I need your help:

My general goal: I want to improve the enclosures, that HTML is set in `Feed->loadEntries()` but it would make sense to have the HTML in `normal.phtml`.

I added `FreshRSS_Entry->content_enclosure` and it setter `_content_enclosures()` and getter `content_enclosures()`.

The issue: with $test (in the constructor) it works, but $test2 (in loadEntries()) it does not work,

How to test:
1. open an article with an enclosure.
2. see the `print_r()` output

It is now:
grafik

Expected behaviour:
`Array ( [0] => lo [1] => ad [2] => Ent [3] => ry ) 1 `

Any help?

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -20,6 +20,7 @@ class FreshRSS_Entry extends Minz_Model {
 	private $title;
 	private $authors;
 	private $content;
+	private $content_enclosures = [];
 	private $link;
 	private $date;
 	private $date_added = 0; //In microseconds
@@ -52,6 +53,8 @@ class FreshRSS_Entry extends Minz_Model {
 		$this->_title($title);
 		$this->_authors($authors);
 		$this->_content($content);
+		$test = ["cons","truct"];
+		$this->_content_enclosures($test);
 		$this->_link($link);
 		$this->_date($pubdate);
 		$this->_isRead($is_read);
@@ -118,6 +121,9 @@ class FreshRSS_Entry extends Minz_Model {
 	}
 	public function content(): string {
 		return $this->content;
+	}
+	public function content_enclosures(): array {
+		return $this->content_enclosures;
 	}
 
 	/** @return array<array<string,string>> */
@@ -317,6 +323,11 @@ class FreshRSS_Entry extends Minz_Model {
 	public function _content(string $value) {
 		$this->hash = '';
 		$this->content = $value;
+	}
+
+	public function _content_enclosures($value) {
+		$this->hash = '';
+		$this->content_enclosures = $value;
 	}
 	public function _link(string $value) {
 		$this->hash = '';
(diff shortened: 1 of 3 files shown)


Frenzie:
Isn't that basically local in scope? That is, shouldn't you need something like `public $test` or `$this->test`?

Apologies if that's a silly remark; I should head off to bed so I only skimmed it. But in case that's all it is it might suffice to resolve the problem. ^_^

app/Models/Entry.php
```diff
@@ -52,6 +53,8 @@ public function __construct(int $feedId = 0, string $guid = '', string $title =
 		$this->_title($title);
 		$this->_authors($authors);
 		$this->_content($content);
+		$test = ["cons","truct"];
+		$this->_content_enclosures($test);
```
math-GH, line 57: in `__construct()` the setter `_content_enclosures()` works

app/Models/Feed.php
```diff
@@ -586,6 +586,8 @@ public function loadEntries(SimplePie $simplePie) {
 			);
 			$entry->_tags($tags);
 			$entry->_feed($this);
+			$test2 = ["lo","ad","Ent","ry"];
+			$entry->_content_enclosures($test2);
 			$entry->hash();	//Must be computed before loading full content
```
math-GH, line 591: this setter `_content_enclosures()` does not work

math-GH:
> Isn't that basically local in scope? That is, shouldn't you need something like `public $test` or `$this->test`?
> 
> Apologies if that's a silly remark; I should head off to bed so I only skimmed it. But in case that's all it is it might suffice to resolve the problem. ^_^

please have a second look on the source code :)

app/Models/Feed.php
Alkarex: I have not looked into it. but first test would be to get the content again there.

Either via debugging, writing to a test file, or writing to the logs, which I often do, e.g.:

```php
syslog(LOG_DEBUG, __METHOD__ . ' ' . json_encode($entry->content_enclosures()));
```

Alkarex:
By the way, I have a draft of code on my side to save the enclosures when we receive them, and be better able to output them again

math-GH:
> By the way, I have a draft of code on my side to save the enclosures when we receive them, and be better able to output them again

How is there the current status?
To be honest: I am weak in backend development to improve the enclosure system well. I would be able to refactor the HTML code from Feed.php to the phtml templates. I would prefer to use your code ;)

Alkarex:
@math-GH Here is my draft: https://github.com/FreshRSS/FreshRSS/pull/4944
It is not completely finished but almost, and not tested. Tests welcome already.

math-GH:
I close this draft. I go forward with #4944

PR #4979, opened 2023-01-01
Fix: Nord Theme: "Mark all" menu item

### LINKED ISSUE #4978: Feed menu: "mark as read" is to big
If you click manage in this menu:
image

is "Mark as read" supposed to be so large? That's with the latest edge, Nord theme.

_Originally posted by @Frenzie in https://github.com/FreshRSS/FreshRSS/issues/4977#issuecomment-1368481030_

math-GH:
Closes #4978

- CSS for Nord theme
- (S)CSS for other themes. font-size: inherit makes the code a bit clearer

1. see #4978

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -348,7 +348,7 @@ a.btn {
 .dropdown-menu .item > .as-link {
 	padding: 0 22px;
 	line-height: 2.5;
-	font-size: 0.8rem;
+	font-size: inherit;
 }
 
 .dropdown-menu .dropdown-section .item > a,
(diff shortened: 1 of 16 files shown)


p/themes/Alternative-Dark/adark.css
```diff
@@ -348,7 +348,7 @@ a.btn {
 .dropdown-menu .item > .as-link {
 	padding: 0 22px;
 	line-height: 2.5;
-	font-size: 0.8rem;
+	font-size: inherit;
```
Frenzie, line 351: Wouldn't removing the line have the same effect?

math-GH: I thought so, too. But it does not have the same effect.

`font-size: inherit` = 12,8pixel
deleted line = 13.33 pixel

To be honest: I did not understand why it has not the same effect.

Alkarex: > I did not understand why it has not the same effect

I have not checked, but this is typically due to the fact that without this line, another rule is applied from another selector.

Frenzie: The inspector will say which rule. Anyway, just double checking.

PR #4980, opened 2023-01-01
Improve: open "manage feed" with Ctrl in a new tab

### LINKED ISSUE #4977: [Feature]ctrl+left click on manage button on feed management should open in new tab.
when user opens feed management and I try to open a feed management in new tab, I either have to right click on it and select to open in new tab, or middle-click.
but everywhere else (even in other part of freshrss) ctrl+clicking on a link (or button) opens it in new tab.
can this be fixed?
right now it opens it in a sliding page not a new page which happens with just clicking also.

math-GH:
Closes #4977

Fix provided by @Frenzie 

- JS

1. Press CTRL key and click on "manage"
2. it will open the feed settings in a new tab

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -145,6 +145,9 @@ function init_archiving(parent) {
 const freshrssSliderLoadEvent = new Event('freshrss:slider-load');
 
 function open_slider_listener(ev) {
+	if (ev.ctrlKey || ev.shiftKey) {
+		return;
+	}
 	const a = ev.target.closest('.open-slider');
 	if (a) {
 		if (!context.ajax_loading) {


p/scripts/extra.js
```diff
@@ -145,6 +145,9 @@ function init_archiving(parent) {
 const freshrssSliderLoadEvent = new Event('freshrss:slider-load');
 
 function open_slider_listener(ev) {
+	if (ev.ctrlKey || ev.shiftKey) {
+		return;
+	}
```
math-GH, line 150: @Frenzie Why is is important to respect the `Shift` key?

Frenzie: That's new window.

Frenzie (approved):

PR #4989, opened 2023-01-07
Better handling of special chars in searches

### LINKED ISSUE #4987: [BUG] Unable to use parentheses in search/queries
**Describe the bug**
When composing a search, feed rule, or user query — it is no longer possible to use parentheses `()` to match title strings (instead, they are favoured for rule grouping).

**To Reproduce**
1. Find an article that contains a string in parentheses.
2. Attempt to build a query using `intitle:({{ string }})`
3. Observe more articles are returned than expected.

**Expected behavior**
Only the articles matching the exact query string should be returned.

**Screenshots**
Attempting to search articles with `(sponsor)` in the title, but other, incorrect matches are returned.
image

- Device: MacBook Pro (16-inch, 2021)
- OS: macOS 12.6
- Browser: Safari 16.1
- FreshRSS version: 1.20.2
- Database version: MariaDB 10.6
- PHP version: PHP 8.1.2
- Installation type: Self-host, Ubuntu 22.04 VM under ESXi 7.0

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4987

Allows searching for parentheses with `\(` or `\) ` ~~and a better handing of `_` and `#` which means something special in SQL `LIKE`~~

### DIFF
--- a/app/Models/BooleanSearch.php
+++ b/app/Models/BooleanSearch.php
@@ -165,7 +165,7 @@ class FreshRSS_BooleanSearch {
 				$i++;
 				while ($i < $length) {
 					$c = $input[$i];
-					$backslashed = $i >= 1 ? $input[$i - 1] === '\\' : false;
+					$backslashed = $input[$i - 1] === '\\';
 					if ($c === '(' && !$backslashed) {
 						// One nested level deeper
 						$parentheses++;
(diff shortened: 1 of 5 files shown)


math-GH:
Should the [documentation ](https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html#with-the-search-field) also have an update?

Alkarex:
> Should the [documentation ](https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html#with-the-search-field) also have an update?

@math-GH Yes. I am a bit hesitating about `#` and `_` (special or not special by default) and maybe I should remove the changes in this PR to address only `( )` for now

Alkarex:
Documentation added

PR #4994, opened 2023-01-08
Search Improved: dropdown

math-GH:
## Before:
wide screen:
grafik

mobile view (focus NOT in the search input)
grafik

mobile view (focus in the search input)
grafik

## After:
wide view:
grafik

mobile view:
grafik

- wide view:
  - deleted: search input in the header
- mobile view:
  - deleted: search input in the navigation (it expands when it is focused. It breaks the layout)
- new: search button in the filter menu bar, that opens a dropdown with the search input. The seach button icon is blue, when a search key is given (Origine theme)

Why I think it is an improvement:
- I do not like the expanding feature in the mobile view, that breaks the layout
- the button needs less space than the input in the mobile view
- I do not like the search input on every page (f.e. config pages)
- the search influences the filter (read/unread, favorite/non-favorite) and can be save in a user query

1. use the seach

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
This PR is marked as draft for collecting feedback

Frenzie:
I don't really like the idea of having to perform an extra action. On mobile that might potentially be a necessary evil but on a larger screen not so much. But perhaps it would suffice to automatically focus the input. (That still leaves an additional action when modifying searches though.)

Also if you merge those together I think I'd rather move everything into the header and get rid of the navigation. :-P

Alwaysin:
I also prefer having the search box readily available

math-GH:
The search bar in the desktop view is back.
The search button will be shown only in the mobile view.

A link to the search documentation is added
image

p/themes/base-theme/frss.css
```diff
@@ -115,6 +115,10 @@ h3 {
 	display: none;
 }
 
+.no-desktop {
```
Frenzie: Is this a phrasing that's already used elsewhere? If this is brand new something like d-l-none/d-large-none or d-wide-none would be more neutral. But looking at the diff I'd call this d-none/display-none or something along those lines and the other d-mobile/small/whatever-unset.

math-GH: `.no-mobile` is already used. That is why I introduced `.no-desktop`

Frenzie: While I don't necessarily like that one either then, it does work better since in practical terms it doesn't implicitly seem to suggest that one might be trying to exclude tablets.

In short, I want it to be as clear as possible while reading without having to look up whether it actually just means large/wide or if it actually means desktop. That's not always possible, but in this case it seems fairly easily attainable using something like the `.d-none` combined with `.d-sm-unset` pair I suggested above.

@Alkarex what do you think?

math-GH: Right now we have just 2 screen sizes in the CSS: Screens smaller than `840px` (`@media (max-width: 840px)`) are the "mobile" screens.
`.no-mobile` is used to hide elements, that should not be shown on small screens.
The code has already an implicite `.no-desktop`, but it has not that CSS class.

Frenzie: Right, but it's a neutral large/wide, not "desktop". If you write "desktop" it makes you think of things like `@media (hover: hover)` vs `@media (hover: none)`.

math-GH: What would you suggest of a better naming?

Frenzie: Like I said, something involving for example `.d-none`, `.d-sm-unset` and/or `.d-lg-none`.

math-GH: what does `d` and `sm` and `lg` mean?

Frenzie: In the context of utility classes d is a common abbreviation for display.

So something like includes for example `.display-none` and `display-small-unset`.

math-GH: it would be a complete new coding style for the CSS in FreshRSS

Frenzie: It's not quite clear to me what you're referring to. My very argument is that something along the lines of `small` and `large` constitutes neutral names mostly free from the potentially misleading assumptions that come along with `mobile` and particularly the proposed `desktop`. So yes, it'd be new to the 4+2 lines of code where `.no-mobile` is currently used.

To perhaps illustrate the problem better, consider calling it `.no-tv` instead.

math-GH: I changed it to `.only-mobile`

Frenzie: Well… okay, lol.

Frenzie (approved):

math-GH:
Ready for final review

app/i18n/fr/gen.php
```diff
@@ -174,6 +174,7 @@
 		'queries' => 'Filtres utilisateurs',
 		'reading' => 'Lecture',
 		'search' => 'Rechercher des mots ou des #tags',
+		'search_help' => 'See documentation for advanced <a href="https://freshrss.github.io/FreshRSS/en/users/10_filter.html#with-the-search-field" target="_blank">search parameters</a>',	// TODO
```
```suggestion
		'search_help' => 'Voir <a href="https://freshrss.github.io/FreshRSS/fr/users/03_Main_view.html#gr%C3%A2ce-au-champ-de-recherche" target="_blank">la documentation pour la syntaxe des recherches avancées</a>',
```

app/layout/nav_menu.phtml
```diff
@@ -29,10 +29,53 @@
...
+				<li class="item">
+					<span>
+						<form action="<?= _url('index', 'index') ?>" method="get">
+							<?php $param_a = Minz_Request::actionName(); ?>
+							<?php if (in_array($param_a, ['normal', 'global', 'reader'])) { ?>
+							<input type="hidden" name="a" value="<?= $param_a ?>" />
+							<?php } ?>
+
+							<?php $get = Minz_Request::param('get', ''); ?>
+							<?php if($get != '') { ?>
```
Alkarex:
```suggestion
							<?php if ($get != '') { ?>
```

app/layout/nav_menu.phtml
```diff
@@ -29,10 +29,53 @@
...
+							<input type="hidden" name="get" value="<?= $get ?>" />
+							<?php } ?>
+
+							<?php $order = Minz_Request::param('order', ''); ?>
+							<?php if($order != '') { ?>
```
Alkarex:
```suggestion
							<?php if ($order != '') { ?>
```

app/layout/nav_menu.phtml
```diff
@@ -29,10 +29,53 @@
...
+							<input type="hidden" name="order" value="<?= $order ?>" />
+							<?php } ?>
+
+							<?php $state = Minz_Request::param('state', ''); ?>
+							<?php if($state != '') { ?>
```
Alkarex:
```suggestion
							<?php if ($state != '') { ?>
```

Alkarex:
Quick test:

image

1) When the search field is focused, clicking on the search button makes the search field lose focus, shrink, and therefore the button moves and the click does nothing (tested in Firefox, with Pafat and Origine themes)

2) Minor alignment issue of the button (in Pafat at least)

Alkarex (changes requested):
See comments

math-GH:
> Quick test:
> 
> image
> 
>     1. When the search field is focused, clicking on the search button makes the search field lose focus, shrink, and therefore the button moves and the click does nothing (tested in Firefox, with Pafat and Origine themes)
> 
>     2. Minor alignment issue of the button (in Pafat at least)

1. It was not changed by this PR. I do not like the expanding thing (CSS class `extend`), because it is a bad UX (I could kick out this expanding thing everywhere in another PR).
2. I do not see any alignment issue in Pafat. Could you please support here with an screenshot?

Frenzie:
> It was not changed by this PR. I do not like the expanding thing (CSS class extend), because it is a bad UX (I could kick out this expanding thing everywhere in another PR).

You'd already done all the work, but my first reaction was simply to get rid of the expansion on mobile (obviously) and possibly on desktop too 'cause I'm not super fond of it anyway. ;-)

Alkarex:
> 2\. I do not see any alignment issue in Pafat. Could you please support here with an screenshot?

image

math-GH:
> > 2. I do not see any alignment issue in Pafat. Could you please support here with an screenshot?
> 
> image

It is by design of Pafat theme. Focused inputs have a blue outline, that is outside of the border. The light grey border does not change.

math-GH:
> > It was not changed by this PR. I do not like the expanding thing (CSS class extend), because it is a bad UX (I could kick out this expanding thing everywhere in another PR).
> 
> You'd already done all the work, but my first reaction was simply to get rid of the expansion on mobile (obviously) and possibly on desktop too 'cause I'm not super fond of it anyway. ;-)

I opened another PR for it: #5040

math-GH:
Oh, I forgot Ansum and Mapco :/

math-GH:
Ansum + Mapco themes are done too

PR #4998, opened 2023-01-09
Lazy load logo on install, logo on header and theme thumbs in config slide

### LINKED ISSUE #2371: Consider using native loading="lazy"

sad270:
closes **Partialy** #2371

- Lazy load logo on install
- Fill the `alt` attribute of logo on install
- Lazy load logo on header
- Lazy load theme thumbs in config slide 

1. After reseting all project cache, open project url, you will be on installation page.
2. Inspect element, and verify the  FreshRSS logo `alt` attribute and `loading` attribute
3. After installation, inspect element and verify the  FreshRSS logo (on header) `loading` attribute
4. Go to Configuration / Display
5. Inspect element and verify theme thumbs image `loading` attribute

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/p/index.html
+++ b/p/index.html
@@ -14,6 +14,6 @@
 
 <body>
 <h1><a href="i/">FreshRSS</a></h1>
-<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" loading="lazy"/></a></p>
+<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" loading="lazy" /></a></p>
 </body>
 </html>
--- a/p/index.html
+++ b/p/index.html
@@ -14,6 +14,6 @@
 
 <body>
 <h1><a href="i/">FreshRSS</a></h1>
-<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" /></a></p>
+<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" loading="lazy"/></a></p>
 </body>
 </html>
--- a/app/install.php
+++ b/app/install.php
@@ -445,7 +445,7 @@ function printStep1() {
 	<?php } else { ?>
 	<p class="alert alert-error"><?= _t('install.action.fix_errors_before') ?></p>
 	<a id="actualize" class="btn" href="./index.php?step=1" title="<?= _t('install.check.reload') ?>">
-		<img class="icon" src="../themes/icons/refresh.svg" alt="🔃" />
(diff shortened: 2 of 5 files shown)


Frenzie:
I suppose it (hopefully) shouldn't hurt but I'm not sure how relevant it is to #2371. The current behavior is that if you open an extremely long article with dozens or even hundreds of images, they'll all be loaded right at that moment. If they had a `loading=lazy` then most of that would be delayed until you scrolled closer. It's an edge case that doesn't happen much but it could save some data. Anyway, that's where I think it'd make a difference. :-)

p/index.html
```diff
@@ -14,6 +14,6 @@
 
 <body>
 <h1><a href="i/">FreshRSS</a></h1>
-<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" /></a></p>
+<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" loading="lazy"/></a></p>
```
Alkarex:
```suggestion
<p><a href="i/"><img class="logo" width="25%" src="themes/icons/icon.svg" alt="⊚" loading="lazy" /></a></p>
```

Minor style consistency

sad270:
@Frenzie yes I know, i tag the issue, because his name is "Consider using native loading="lazy"". It is a "global" issue about lazy loading and this issue include the lazy loading of image in content of article.

I start to watch how to do it, i create first PR with all little lazy loading midding in the app. And if i have time, and I will create another PR for the lazy load of the image in content.

I see that there is "enclosure" and some image is in enclosure, what is this ? Should they be in lazy load too ? I didn't change them it seems to be the image returning in GReaderApi endpoint.

Frenzie:
@sad270 Images are non-blocking and this particular image will be loaded regardless because it'll be in the viewport.

My point is that there's really only a single category of images where it does make sense. Other images needn't or even shouldn't be changed, imho.

Frenzie:
In Chromia you can also use the Lighthouse tool for another illustration of that principle. You shouldn't take everything it says at face value, but the login page and install page receive a performance score of 100% (before and after this PR). And it's rather overzealous about recommending lazyload. ;-)

sad270:
Ok I understand

Frenzie:
Anyway, like I said I don't think it really harms anything. One optimization Lighthouse pointed out that makes sense btw is to provide a width and height attribute on some of the images, so the size they take up can already be painted before they're loaded. :-)

sad270:
I didn't know all this optimization. I'm not expert in front dev :D thanks for all this informations.

Alkarex:
See https://github.com/FreshRSS/FreshRSS/issues/2371

PR #5001, opened 2023-01-09
API avoid logging passwords

Alkarex:
* Strip passwords and tokens from API logs
* Only log failed requests information when in debug mode

### DIFF
--- a/cli/sensitive-log.sh
+++ b/cli/sensitive-log.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+# Strips sensitive passwords from (Apache) logs
+
+# For e.g. GNU systems such as Debian
+# N.B.: `sed -u` is not available in BusyBox and without it there are buffering delays (even with stdbuf)
+sed -Eu 's/([?&]Passwd)=[^& \t\n]+/\1=█/ig' 2>/dev/null ||
+
+# For systems with gawk (not available by default in Docker of Debian or Alpine) or with BuzyBox such as Alpine
+$(which gawk || which awk) -v IGNORECASE=1 '{ print gensub(/([?&]Passwd)=[^& \t\n]+/, "\\1=█", "g") }'
--- a/cli/sensitive-log.sh
+++ b/cli/sensitive-log.sh
@@ -5,5 +5,5 @@
 # N.B.: `sed -u` is not available in BusyBox and without it there are buffering delays (even with stdbuf)
 sed -Eu 's/([?&])(Passwd|token)=[^& \t\n]+/\1\2=█/ig' 2>/dev/null ||
 
-# For systems with gawk (not available by default in Docker of Debian or Alpine) or with BuzyBox such as Alpine
-$(which gawk || which awk) -v IGNORECASE=1 '{ print gensub(/([?&])(Passwd|token)=[^& \t\n]+/, "\\1\\2=█", "g") }'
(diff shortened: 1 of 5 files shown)


Alkarex:
Any comment, e.g. @Frenzie ?

cli/sensitive-log.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+# Strips sensitive passwords from (Apache) logs
+
+# For e.g. GNU systems such as Debian
+# N.B.: `sed -u` is not available in BusyBox and without it there are buffering delays (even with stdbuf)
+sed -Eu 's/([?&])(Passwd|token)=[^& \t\n]+/\1\2=█/ig' 2>/dev/null ||
```
Frenzie: It's not entirely clear to me why you're avoiding all of `[& \t]`? Particularly ampersand and space could simply be part of the password even if they should normally be escaped. Simply replacing everything until `\n` feels safer. (Also it'd be helpful to provide some sample logs to judge better. ;-)

As for the replacement perhaps just something simpler like `xxx` would be nicer to look at?

Alkarex: `\n` was unneeded.

`&` and ` ` cannot be a part of the password as everything is URL-encoded in a parameter. `&` is solely for separating fields by definition.

Deleting everything until the end of the line would break the log format (in case of downstream use).

I like the Unicode █ as it seems to be (becoming) the de-facto standard for redacted texts, e.g. https://wet-boew.github.io/GCWeb/components/redacted-en.html
However, it makes the URL not strictly valid, which might potentially break something. So if you do not like it, I could propose `[REDACTED]` instead?

Example of log:

```console
$ echo '2023-01-09T13:43:32.567928288Z 127.0.0.1 - - [09/Jan/2023:14:43:32 +0100] "GET /api/greader.php/accounts/ClientLogin?Email=alice&passwd=secret123&misc HTTP/1.1" 401 13 "-" "curl/7.81.0"' | \
sed -Eu 's/([?&])(Passwd|token)=[^& \t]+/\1\2=█/ig'

2023-01-09T13:43:32.567928288Z 127.0.0.1 - - [09/Jan/2023:14:43:32 +0100] "GET /api/greader.php/accounts/ClientLogin?Email=alice&passwd=█&misc HTTP/1.1" 401 13 "-" "curl/7.81.0"
```

Frenzie: The missing sample log line provides most of the explanation needed. ;-) Though I'll note that a badly written HTTP client could still accidentally break the `&` expectations, at least for a very short while.

> However, it makes the URL not strictly valid, which might potentially break something. So if you do not like it, I could propose [REDACTED] instead?

[REDACTED] still suffers from most of the same problem imo, namely that it draws all the attention of the eye when the value of Passwd/token is much closer to irrelevant than to redacted. For that matter even for "real" redacted I'm not sure how much sense such an eyesore makes when you could use something less visually displeasing like ▂▂▂▂▂▂▂ (U+2582, lower one quarter block). ████ looks like what you'd do on paper by necessity, not because it's such a fantastic look.

Alkarex: I have changed for `redacted` . Ok?

Frenzie: Haha, you can use `█` if you like it. I just think it doesn't look great but it's not a big deal. But yes, I definitely prefer `redacted`. It looks similar enough to a real value not to draw unnecessary attention to itself.

PR #5021, opened 2023-01-18
Safer timezone set

Alkarex:
* Add missing tzdata in Docker :newest
* Fallback to UTC if no timezone is defined at all
#fix https://github.com/FreshRSS/FreshRSS/pull/4906#issuecomment-1386747169

### DIFF
--- a/Docker/Dockerfile-Newest
+++ b/Docker/Dockerfile-Newest
@@ -4,6 +4,7 @@ ENV TZ UTC
 SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories && \
 	apk add --no-cache \
+	tzdata \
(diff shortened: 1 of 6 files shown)


Alkarex:
@Alwaysin Please try. This should fix your warning. And then define your `date.timezone` in `php.ini` and/or define the new timezone option in user options.

Alwaysin:
I can confirm it suppressed the warning, but the UTC default doesn't seem to have been picked up by the UI:
chrome_2023-01-19_10-06-57

Isn't there a way to detect the timezone from the browser instead of defaulting to UTC?

Anyway, I've added my timezone in php.ini and all is good for me now!

Alkarex:
Could you please try again @Alwaysin ?
I have fixed the visualisation of the default timezone.

For the things that are server-side rendered, it is not so practical to detect and use browser settings. But not impossible. Anyway, not something for now.

Alwaysin:
All good now!

PR #4807, opened 2022-11-01
Added: Mark theme as deprecated

math-GH:
While I update all themes and look into the little survey (#4295) I see 3 themes, that are not used often:
- BlueLagoon
- Flat
- Screwdriver

It would make sense to reduce our offered themes by this 3 themes. Some new themes were added in 1.20.

(IMHO: this 3 theme's designs are out of date)

Changes proposed in this pull request:

- option to mark a theme as deprecated via `manifest.json`
- marked the 3 themes as deprecated
- show the deprecation status on the reading configuration
- documentation improved (added the new themes, added the deprecation information)

How to test the feature manually:

1. see the theme slider

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #5022, opened 2023-01-18
Minz i18n safer vsprintf

Alkarex:
Fix bug introduced by https://github.com/FreshRSS/FreshRSS/pull/4807 though only for the French i18n, which has some `%` signs in the URL, making `vsprintf` to crash.

Only use `vsprintf` when we call the translation with some parameters, otherwise skip it (probably faster as well).

```
Warning: vsprintf(): Too few arguments in /var/www/FreshRSS/lib/Minz/Translate.php on line 235
```

But with PHP 8.2+, fatal error:

```
PHP Fatal error:  Uncaught ValueError: The arguments array must contain 2 items, 0 given
```

### DIFF
--- a/lib/Minz/Translate.php
+++ b/lib/Minz/Translate.php
@@ -232,7 +232,7 @@ class Minz_Translate {
 		}
 
 		// Get the facultative arguments to replace i18n variables.
-		return vsprintf($translation_value, $args);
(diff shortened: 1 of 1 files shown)


Alkarex:
Quick merge as the current French version is broken, but comments welcome anyway

PR #5038, opened 2023-01-24
Increase max feed URL length and drop unicity

### LINKED ISSUE #1677: “Feed URL” field character limit
I can paste the complete URL in the “Add a RSS feed” field under “Subscriptions management”. When I click the + sign I get an error message saying “[Feed title] could not be added” if the URL is over 511 characters. Under the “Information” header (settings for the feed) “Feed URL” shows the URL with the first 511 characters.

I'm trying to subscribe to a multireddit feed, which is now impossible (100 chars short). If technically possible, an increased limit would be much welcome. [Read this Stack Overflow reply](https://stackoverflow.com/questions/29458445/what-is-a-safe-maximum-length-a-segment-in-a-url-path-should-be/33733386#33733386) for more info on URL limitations.

### LINKED ISSUE #4229: [BUG] SQL error updateFeed: UNIQUE constraint failed: feed.url for feed
**To Reproduce**
go to subscription management page
grafik

Copy the feed URL from 1 feed (= feed URL 1)

Open another feed (= feed URL 2)
grafik

Paste the feed into the feed URL (=overwrite feed URL 2 with feed URL 1)
grafik

save it. Error message "Feed cannot be updated"
grafik

Go to log and see the error message `SQL error updateFeed: UNIQUE constraint failed: feed.url for feed`
grafik

**Expected behavior**
no error message in log.
Maybe another error text like "Duplicated feed URL"

- FreshRSS version: 1.20.0-dev

### LINKED ISSUE #4338: [BUG] Long Feed URLs get truncated
**Describe the bug**
I am using RSSBridge to generate some feeds. RSSBridge can produce lengthy URLs. I create one feed url composed of 482 characters but when used in FreshRSS it gets truncated to 475 characters. I noticed the mysql database field allows 511 as length. Somewhere else in the code, the feed url is being truncated. I have manually pasted the url in mysql and the feed works.

**To Reproduce**
1. Produce a url with more than 475 characters
2. create a new feed
3. After creation, check feed url to see that it gets truncated.

**Expected behavior**
Feed URL will get truncated to 475 characters

- Device: x86 pc
- OS: Ubuntu 20.04
- Browser: Firefox
- FreshRSS version: 1.19.2
- Database version: 10.3.34-MariaDB
- PHP version: 7.4

**Suggestion**
Solve truncation problem and possibly increase length of Feed URL for compatibility with the useful RSSBridge

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4338

* Drop the unicity constraint on our feeds' URL. In practice, this did not add much value as identical feeds could have different URLs. In addition of being a technical annoyance, it generated several problems, for instance during renaming or automatic redirections such as from HTTP to HTTPS, with collisions for which we do not have any elegant handling.
* I have kept a high limit of 32768 because cURL does not seem to have any limit except memory, which might lead to memory problems. This is the highest server-side value reported by https://www.sistrix.com/ask-sistrix/technical-seo/site-structure/url-length-how-long-can-a-url-be

fix https://github.com/FreshRSS/FreshRSS/issues/4229


Frenzie (approved):
lgtm, the old limit was surprisingly low really :-)

Alkarex:
> lgtm, the old limit was surprisingly low really :-)

Yes, the old limit was due to unicity constraints in MySQL...

Frenzie:
Right, I meant I find it surprising that it can't do it with a slightly higher value than varchar(255), but more important as you said the added value seems rather limited regardless.

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/4229

PR #5040, opened 2023-01-24
Improved: expanding inputs

math-GH:
Ref from discussion https://github.com/FreshRSS/FreshRSS/pull/4994#issuecomment-1401710283

- deleted the `extend` CSS classes (except one case. it will be deleted with #4994

1. check the text inputs: set the mouse focus in it (for example to write text in it)
2. they will not change the widths anymore

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/layout/header.phtml
+++ b/app/layout/header.phtml
@@ -15,7 +15,7 @@
 		<?php if (FreshRSS_Auth::hasAccess() || FreshRSS_Context::$system_conf->allow_anonymous) { ?>
 		<form action="<?= _url('index', 'index') ?>" method="get">
 			<div class="stick">
-				<input type="search" name="search" id="search" class="extend"
+				<input type="search" name="search" id="search"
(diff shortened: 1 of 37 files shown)


app/layout/header.phtml
```diff
@@ -15,7 +15,7 @@
 		<?php if (FreshRSS_Auth::hasAccess() || FreshRSS_Context::$system_conf->allow_anonymous) { ?>
 		<form action="<?= _url('index', 'index') ?>" method="get">
 			<div class="stick">
-				<input type="search" name="search" id="search" class="extend"
+				<input type="search" name="search" id="search"
```
Frenzie, line 18: I'd make it class=long.

Btw, isn't there also `.header .item.search input:focus` in some of the themes or was that changed in the meantime?

math-GH: fixed

math-GH:
the text inputs have now the width of expanded status.

Frenzie (approved):

PR #5049, opened 2023-01-28
PHPStan level 9 for ./p/ and lib_rss.php

Alkarex:
And `app/FreshRSS.php`.
Contributes to https://github.com/FreshRSS/FreshRSS/issues/4112 (addresses 146 level-6 errors)
Refactored `greader.php` as a class with private functions.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -17,10 +17,14 @@ class FreshRSS_Entry extends Minz_Model {
 	 */
 	private $guid;
 
+	/** @var string */
 	private $title;
 	private $authors;
+	/** @var string */
 	private $content;
+	/** @var string */
 	private $link;
+	/** @var int */
 	private $date;
 	private $date_added = 0; //In microseconds
 	/**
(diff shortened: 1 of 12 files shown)


app/Models/Entry.php
```diff
@@ -17,10 +17,14 @@ class FreshRSS_Entry extends Minz_Model {
 	 */
 	private $guid;
 
+	/** @var string */
```
aledeg, line 20: Can't you add the type directly in the property definition?

Alkarex: That would require PHP 7.4+: https://php.net/language.types.declarations

With PHP 7.2 (which we support at the moment https://github.com/FreshRSS/FreshRSS/pull/4848 ): `PHP Parse error:  syntax error, unexpected 'string' (T_STRING), expecting function (T_FUNCTION) or const (T_CONST) in ./app/Models/Entry.php on line 18`

Alkarex: Let's come back to it in a few months https://github.com/FreshRSS/FreshRSS/discussions/3321#discussioncomment-835704

Alkarex: Interesting reading; there are several details I did not know (e.g. about uninitialized properties) https://wiki.php.net/rfc/typed_properties_v2

Issue #5048, opened 2023-01-27
[BUG] Feed entry thumbnail always shows black/white stripes, no matter the selected format


k5123:
When showing thumbnails, the images always show with black and white stripes, no matter their actual format and the configured display option.

**To Reproduce**
1. Go to Display
2. Select any format except 'none' for Thumbnail
3. Go back to feed
4. Thumbnails now have black and white stripes visible, irregardless of the format selected in 2

**Expected behavior**
No black and white stripes

**Screenshots**
![image](https://user-images.githubusercontent.com/9386722/215217192-d35ea7dc-054b-46b0-ba17-82544e31bb0e.png)

**Environment information (please complete the following information):**
Browser: Firefox 108.2
FreshRSS: 1.20.2

Frenzie:
It seems to come from the standard padding on .item-element.

k5123:
I was digging a little bit too, and if you remove the background: line from the `.flux .item.thumbnail img` block, it seems to be fine - layout and otherwise.

Alkarex:
> I was digging a little bit too, and if you remove the background: line from the `.flux .item.thumbnail img` block, it seems to be fine - layout and otherwise.

@k5123 PR welcome

math-GH:
@k5123 Please check the fix #5058

ltschk:
First of all: I pretty much enjoy using FreshRSS! Using it probably since 2 years now. Great tool, great work!! Thanks for that!!

I switched from the Origine theme to **Origine-compact** today and got those black/white stripes from the background as mentioned [above](https://github.com/FreshRSS/FreshRSS/issues/5048#issue-1560505594).
![grafik](https://github.com/FreshRSS/FreshRSS/assets/150657587/469457eb-6a5d-4da6-a164-5da882f830f3)

I could fix this by adjusting line [128](https://github.com/FreshRSS/FreshRSS/blob/354a5e83888cebe89843b754d0936f3633323cef/p/themes/Origine-compact/origine-compact.css#L128) of p/themes/Origine-compact/origine-compact.css from `padding: 0.25rem 0;` to `padding: 0 0;` and in line [128](https://github.com/FreshRSS/FreshRSS/blob/354a5e83888cebe89843b754d0936f3633323cef/p/themes/Origine-compact/origine-compact.rtl.css#L128) of p/themes/Origine-compact/origine-compact.rtl.css respectively.

math-GH:
Thanks for that finding.
[...]

PR #5052, opened 2023-01-29
Issue #5048 - removed striped background from thumbnails

k5123:
Closes #

- frss.css changed to remove the striped background: attribute from the thumbnail style

1. Enable thumbnails
2. Images - if exist - should display without any black and white stripes around them

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested


math-GH:
I would not delete this background.

Let me explain shortly because I implemented it in the past.

The striped background should not be visible in general. It is a bug, that the stripes are visible above and below the image.

The stripes are behind the image. If the image is not (yet) loaded than the stripes are a placeholder. Sometimes the images are very huge so that loading needs some seconds.
Without the striped background there would be a big gap between the feed name and the article title.

We need to find another solution to fix this issue. The issue is with the `padding`...

Frenzie:
Like I said in https://github.com/FreshRSS/FreshRSS/issues/5048#issuecomment-1407431867, you could just remove the vertical padding in that specific scenario.

k5123:
> We need to find another solution to fix this issue. The issue is with the `padding`...

Ok thanks for the feedback. Not into CSS enough to implement a proper fix. Will wait for the official one :)

PR #5062, opened 2023-02-01
More robust application of access permissions

Alkarex:
We were in particular missing directory search `+X` in our current recommendations.
Extracted to own shell script `cli/access-permissions.sh` so it can easily be invoked.
Update access permissions in Docker to account to be more robust.
#fix https://github.com/FreshRSS/FreshRSS/discussions/5037

### DIFF
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -7,8 +7,6 @@ find /etc/php*/ -type f -name php.ini -exec sed -r -i "\\#^;?date.timezone#s#^.*
 find /etc/php*/ -type f -name php.ini -exec sed -r -i "\\#^;?post_max_size#s#^.*#post_max_size = 32M#" {} \;
 find /etc/php*/ -type f -name php.ini -exec sed -r -i "\\#^;?upload_max_filesize#s#^.*#upload_max_filesize = 32M#" {} \;
 
-php -f ./cli/prepare.php >/dev/null
-
 if [ -n "$LISTEN" ]; then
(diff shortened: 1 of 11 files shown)


Docker/entrypoint.sh
```diff
@@ -24,6 +22,10 @@ if [ -n "$CRON_MIN" ]; then
 		-r "s#^[^ ]+ #$CRON_MIN #" | crontab -
 fi
 
+./cli/access-permissions.sh
```
Alkarex, line 25: Access rights need to be ensured also before our CLI commands, in the case of external volumes with incompatible permissions.

cli/access-permissions.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [ "$(id -u)" -ne 0 ]; then
+	echo >&2 '⛔ Applying access permissions require running as root or sudo!'
+	exit 3
+fi
+
+# Based on group access
+chown -R :www-data .
+# Read files, and directory traversal
+chmod -R g+rX .
```
Alkarex, line 17: `+X` (directory search) was what was missing in particular for some of the reported user issues

Frenzie:
+X is only the owner?

Alkarex:
> +X is only the owner?

`g+X` gives directory traversal permissions to the group (www-data)

Frenzie:
I don't really understand what that means. How does that differ from recursive?

Alkarex:
> I don't really understand what that means. How does that differ from recursive?

* `-R` (for recursive) applies the permissions to the current folder and everything below. We are doing that.
* `+X` (uppercase) is a type of permission, just like `+r` (read) and `+w` (write). It is the `+x` (lowercase) but only for directories and allows listing the content of directories (also called *search* in man page).

ColonelMoutarde (approved):

PR #5030, opened 2023-01-21
Code enchancement and typhinting

ColonelMoutarde:
Changes proposed in this pull request:

- Code enchancement and typhinting

Remarque:
- This class can be split in many classes with one responsibility by class (S.O.L.I.D).

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #5063, opened 2023-02-01
Type hinting and doc

ColonelMoutarde:
Closes #

- type hinting

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Utils/feverUtil.php
+++ b/app/Utils/feverUtil.php
@@ -1,46 +1,41 @@
 <?php
 
+declare(strict_types=1);
+
 class FreshRSS_fever_Util {
-	const FEVER_PATH = DATA_PATH . '/fever';
+	private const FEVER_PATH = DATA_PATH . '/fever';
 
 	/**
 	 * Make sure the fever path exists and is writable.
 	 *
-	 * @return boolean true if the path is writable, else false.
+	 * @return bool true if the path is writable, else false.
 	 */
-	public static function checkFeverPath() {
+	public static function checkFeverPath(): bool {
 		if (!file_exists(self::FEVER_PATH)) {
 			@mkdir(self::FEVER_PATH, 0770, true);
 		}
 
-		$ok = touch(self::FEVER_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
-		if (!$ok) {
+		$isOk = touch(self::FEVER_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
+		if (false === $isOk) {
 			Minz_Log::error("Could not save Fever API credentials. The directory does not have write access.");
 		}
-		return $ok;
+		return $isOk;
 	}
 
 	/**
 	 * Return the corresponding path for a fever key.
-	 *
-	 * @param string $feverKey
-	 * @return string
 	 */
-	public static function getKeyPath($feverKey) {
+	public static function getKeyPath(string $feverKey): string {
 		$salt = sha1(FreshRSS_Context::$system_conf->salt);
 		return self::FEVER_PATH . '/.key-' . $salt . '-' . $feverKey . '.txt';
 	}
 
 	/**
 	 * Update the fever key of a user.
-	 *
-	 * @param string $username
-	 * @param string $passwordPlain
 	 * @return string|false the Fever key, or false if the update failed
 	 */
-	public static function updateKey($username, $passwordPlain) {
-		$ok = self::checkFeverPath();
-		if (!$ok) {
+	public static function updateKey(string $username, string $passwordPlain) {
+		if (self::checkFeverPath() === false) {
 			return false;
 		}
 
@@ -48,22 +43,20 @@ class FreshRSS_fever_Util {
 
 		$feverKey = strtolower(md5("{$username}:{$passwordPlain}"));
 		$feverKeyPath = self::getKeyPath($feverKey);
-		$res = file_put_contents($feverKeyPath, $username);
-		if ($res !== false) {
+		$result = file_put_contents($feverKeyPath, $username);
+		if (is_int($result) === true) {
 			return $feverKey;
-		} else {
-			Minz_Log::warning('Could not save Fever API credentials. Unknown error.', ADMIN_LOG);
-			return false;
 		}
+		Minz_Log::warning('Could not save Fever API credentials. Unknown error.', ADMIN_LOG);
+		return false;
 	}
 
 	/**
 	 * Delete the Fever key of a user.
 	 *
-	 * @param string $username
-	 * @return boolean true if the deletion succeeded, else false.
+	 * @return bool true if the deletion succeeded, else false.
 	 */
-	public static function deleteKey($username) {
+	public static function deleteKey(string $username) {
 		$userConfig = get_user_configuration($username);
 		if ($userConfig === null) {
 			return false;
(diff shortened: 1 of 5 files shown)


app/Utils/passwordUtil.php
```diff
@@ -1,40 +1,36 @@
...
 			PASSWORD_BCRYPT,
 			array('cost' => self::BCRYPT_COST)
 		);
-		$passwordPlain = '';
 
 		// Compatibility with bcrypt.js
 		$passwordHash = preg_replace('/^\$2[xy]\$/', '\$2a\$', $passwordHash);
 
-		return $passwordHash == '' ? '' : $passwordHash;
+		return $passwordHash === '' ? '' : $passwordHash;
```
Alkarex: `preg_replace()` may return null. `==` was better

app/Utils/feverUtil.php
```diff
@@ -1,69 +1,62 @@
 <?php
 
+declare(strict_types=1);
```
Alkarex: This will just make the application crash in production if any type is not strictly as declared, which is most likely a bad idea at the moment.
Let's come back to that when we have a higher score on PHPStan https://github.com/FreshRSS/FreshRSS/issues/4112

app/Utils/feverUtil.php
```diff
@@ -1,69 +1,62 @@
...
-	public static function checkFeverPath() {
+	public static function checkFeverPath(): bool {
 		if (!file_exists(self::FEVER_PATH)) {
 			@mkdir(self::FEVER_PATH, 0770, true);
 		}
 
-		$ok = touch(self::FEVER_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
-		if (!$ok) {
+		$isOk = touch(self::FEVER_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
+		if (false === $isOk) {
```
Alkarex: Please keep the boolean syntax. We are not going to change to such a verbose syntax everywhere

Alkarex:
Please indicate which PhpStan level your changes allow to pass and for which files. It is almost good for level 9:

```console
$ vendor/bin/phpstan analyse --level 9 app/Models/Days.php  app/Utils/feverUtil.php app/Utils/passwordUtil.php 
Note: Using configuration file /home/alex/GitHub/FreshRSS/phpstan.neon.
 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ -------------------------------------------------------------------- 
  Line   Utils/feverUtil.php                                                 
 ------ -------------------------------------------------------------------- 
  :29    Cannot access property $salt on FreshRSS_SystemConfiguration|null.  
 ------ -------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------- 
  Line   Utils/passwordUtil.php                                                               
 ------ ------------------------------------------------------------------------------------- 
  :23    Method FreshRSS_password_Util::hash() should return string but returns string|null.  
 ------ ------------------------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 2 errors
```

app/Utils/passwordUtil.php
ColonelMoutarde: Every time we use the non-strict comparison, a little fluffy kitten cries out in heartbreaking distress.... ...joking aside, there are huge differences between the soft comparison and the hard comparison. Also the results are not the same between php7 and PHP8 https://www.php.net/manual/en/types.comparisons.php

it is humanly impossible to know all the subtleties of the non-strict comparison

it is possible to simplify like this:

```suggestion
		if ($passwordHash === '' || $passwordHash === null) {
		    return '';
		}
		return $passwordHash;
```

it is more explicit

;)

app/Utils/feverUtil.php
ColonelMoutarde: Be careful when you just do "!" it's not a strict comparison 

see next comment please

app/Utils/feverUtil.php
ColonelMoutarde: https://dev.to/rocksheep/the-way-stricttypes-works-in-php-eb7
In this class, there is no problem

ColonelMoutarde:
> Please indicate which PhpStan level your changes allow to pass and for which files. It is almost good for level 9:
> 
> ```
> $ vendor/bin/phpstan analyse --level 9 app/Models/Days.php  app/Utils/feverUtil.php app/Utils/passwordUtil.php 
> Note: Using configuration file /home/alex/GitHub/FreshRSS/phpstan.neon.
>  3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
> 
>  ------ -------------------------------------------------------------------- 
>   Line   Utils/feverUtil.php                                                 
>  ------ -------------------------------------------------------------------- 
>   :29    Cannot access property $salt on FreshRSS_SystemConfiguration|null.  
>  ------ -------------------------------------------------------------------- 
> 
>  ------ ------------------------------------------------------------------------------------- 
>   Line   Utils/passwordUtil.php                                                               
>  ------ ------------------------------------------------------------------------------------- 
>   :23    Method FreshRSS_password_Util::hash() should return string but returns string|null.  
>  ------ ------------------------------------------------------------------------------------- 
> 
>                                                                                                                         
>  [ERROR] Found 2 errors
> ```

Pass on phpstan level 5

ColonelMoutarde:
Pass on phpstan level 5

Alkarex:
> Pass on phpstan level 5

Yes of course, our whole codebase is already passing PHPStan level 5 :-P

This PR currently upgrades to level 6 the files `app/Models/Days.php app/Utils/feverUtil.php app/Utils/passwordUtil.php` and would even pass level 9 if the two remaining fixes are done.

Cf. my comment on https://github.com/FreshRSS/FreshRSS/pull/5030#issuecomment-1399317663

app/Utils/passwordUtil.php
Alkarex: String comparisons are for strict types, and it is dangerous to replace existing comparisons blindly and without covering all the cases. Most comparisons we have in the current code base have been considered and reviewed.

app/Utils/feverUtil.php
Alkarex: The negation of a strict boolean is also a strict boolean, so the goal for now is to be more strict with types, not to change boolean syntax (so no `=== true` or similar).

Alkarex:
I have just made a few changes and those files are now passing PHPStan level 9, while still be compatible with our oldest supported PHP version (PHP 7.2 at the moment);

```console
$ vendor/bin/phpstan analyse --level 9 app/Models/Days.php  app/Utils/feverUtil.php app/Utils/passwordUtil.php 
Note: Using configuration file /home/alex/GitHub/FreshRSS/phpstan.neon.
 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        

$ docker run --rm -v $PWD:/var/www/FreshRSS freshrss/freshrss:oldest sh -c "find . -type d -name 'vendor' -prune -o -name '*.php' -print0 | xargs -0 -n1 php -l 1>/dev/null"
$
```

Alkarex (approved):

app/Utils/feverUtil.php
ColonelMoutarde: A good practice is to always test what you want, not what you don't want. 
Adding an exclamation mark slows down the reading and increases the cognitive load.
If you test against what you want, there is less brain gymnastics to do, and you can more easily understand the code

PR #5064, opened 2023-02-02
Cleaning code and typehinting

ColonelMoutarde:

- Cleaning code and typehinting
- add extentions in composer.json
-
Pull request checklist:
- [x] clear commit messages
- [x] code manually tested


app/Models/Context.php
```diff
@@ -9,12 +9,12 @@ class FreshRSS_Context {
...
 	 * @var FreshRSS_UserConfiguration|null
 	 */
-	public static $user_conf = null;
+	public static $user_conf;
 
 	/**
 	 * @var FreshRSS_SystemConfiguration|null
 	 */
-	public static $system_conf = null;
+	public static $system_conf;
```
ColonelMoutarde: In this case, the properties are always null.

app/Models/Context.php
```diff
@@ -55,22 +55,23 @@ class FreshRSS_Context {
 	/**
 	 * Initialize the context for the global system.
 	 */
-	public static function initSystem($reload = false) {
-		if ($reload || FreshRSS_Context::$system_conf == null) {
+	public static function initSystem($reload = false): ?\FreshRSS_SystemConfiguration {
+		if ($reload || self::$system_conf == null) {
 			//TODO: Keep in session what we need instead of always reloading from disk
-			FreshRSS_Context::$system_conf = FreshRSS_SystemConfiguration::init(DATA_PATH . '/config.php', FRESHRSS_PATH . '/config.default.php');
+			self::$system_conf = FreshRSS_SystemConfiguration::init(DATA_PATH . '/config.php', FRESHRSS_PATH . '/config.default.php');
```
Alkarex: Any advantage of using self instead of the class name here?
If find it more clear to use the same construct (for public members) everywhere.

composer.json
```diff
@@ -35,7 +35,9 @@
         "ext-xmlreader": "*",
         "ext-zend-opcache": "*",
         "ext-zip": "*",
-        "ext-zlib": "*"
+        "ext-zlib": "*",
+        "ext-pdo": "*",
```
Alkarex: Please keep alphabetic order

Alkarex:
Please remember to indicate PHPStan status before/after the changes

app/Models/Context.php
ColonelMoutarde: There are several advantages:

- reduced cognitive load because we know unambiguously that we are calling the current class and not an external class
- performance, the php parser doesn't have to go through the autoload again because it is explicitly told that the reference is to itself.
- If you change the name of the class, you don't have to go through all the calls again, at the risk of forgetting some and introducing a bug

composer.json
ColonelMoutarde: ok, i do

app/Models/Context.php
Alkarex: If you are able to find a reference for the performance difference, I would be interested.
For the other two points, I find the opposite (lower cognitive load because it is the same everywhere, and easier to check in the whole codebase in case of changes), so that seems to be more a matter of personal preference (and please refrain from changing things that are in a category of pure personal preference, as we can otherwise spend all our time on that, it risks introducing bugs, and it pollutes the git history).

ColonelMoutarde: By using self instead of the current class name, you can ensure that you're calling a method or static property of the class you're in, rather than an inherited or redefined class. This avoids ambiguity and ensures that the expected behavior is executed. In addition, it makes the code clearer and more maintainable by explicitly stating what a static method belongs to.

Is not personal preference, that a standard in PHP industrie.
I had read that the performance was improved, I can't find the article anymore but as soon as I do I'll share it with you

ColonelMoutarde:
My code pass phpstan level 6

app/Models/Context.php
Alkarex: `self::` can also be used to call inherited methods. The rest of what you say can be seen both ways. So while awaiting more evidence on the subject, I would still leave it to personal preference.

ColonelMoutarde: last little question, why use self:: for properties but not for methods?
cf lines : 325 to 337

Alkarex: Written by different persons and different points in time (some lines are 8 years old, I can see), so a bit of inconsistency in personal preferences, which is inconsequential. I would rather not update such lines except if there is a compelling reason to do so, as I give a higher value to a lean git history and to approaches limiting edit wars (in other words, in terms of personal preferences, the original author of those lines should mostly win).

One case see that one of the lines is mine, and I have kept the style of the surrounding lines:
image

ColonelMoutarde: rollbacked

app/Models/Context.php
```diff
@@ -162,7 +164,7 @@ public static function updateUsingRequest() {
 		self::$state = Minz_Request::param(
 			'state', self::$user_conf->default_state
 		);
-		$state_forced_by_user = Minz_Request::param('state', false) !== false;
```
Alkarex, line 166: This is also an example of minor personal preference that I would rather avoid changing, but it does not matter for this time

app/Models/Context.php
```diff
@@ -55,7 +55,7 @@ class FreshRSS_Context {
 	/**
 	 * Initialize the context for the global system.
 	 */
-	public static function initSystem($reload = false) {
+	public static function initSystem($reload = false): ?\FreshRSS_SystemConfiguration {
```
Alkarex: We are not using the backslash syntax anywhere, except in the libraries

app/Models/Context.php
```diff
@@ -216,27 +217,29 @@ public static function getRevertState($state) {
 	public static function currentGet($array = false) {
 		if (self::$current_get['all']) {
 			return 'a';
-		} elseif (self::$current_get['starred']) {
+		}
+		if (self::$current_get['starred']) {
```
Alkarex: Fine for this time, but such examples of minor restructuring are in my opinion more negative than positive, as they pollute changelogs without clear added value (making it more instance more difficult to find regressions, etc.), make reviews longer, incur a risk of introducing bugs, etc.
And by the way, on a personal note, I find the previous syntax both easier to read and more robust: for instance if one of the cases does not return anymore, you might end up entering another case without warning, while with the previous syntax the error would be caught by PHPStan. So the new syntax has removed a security barrier.

Alkarex:
> My code pass phpstan level 6

It does not look like it is passing PHPStan level 6 (cf. https://github.com/FreshRSS/FreshRSS/issues/4112 ):

```console
$ vendor/bin/phpstan analyse --level 6 app/Models/Context.php 
Note: Using configuration file /home/alex/GitHub/FreshRSS/phpstan.neon.
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------------------------------------------ 
  Line   Context.php                                                                               
 ------ ------------------------------------------------------------------------------------------ 
  :19    Property FreshRSS_Context::$categories has no type specified.                             
  :20    Property FreshRSS_Context::$tags has no type specified.                                   
  :22    Property FreshRSS_Context::$name has no type specified.                                   
  :23    Property FreshRSS_Context::$description has no type specified.                            
  :25    Property FreshRSS_Context::$total_unread has no type specified.                           
  :26    Property FreshRSS_Context::$total_starred has no type specified.                          
  :32    Property FreshRSS_Context::$get_unread has no type specified.                             
  :33    Property FreshRSS_Context::$current_get has no type specified.                            
  :41    Property FreshRSS_Context::$next_get has no type specified.                               
  :43    Property FreshRSS_Context::$state has no type specified.                                  
  :44    Property FreshRSS_Context::$order has no type specified.                                  
  :45    Property FreshRSS_Context::$number has no type specified.                                 
  :48    Property FreshRSS_Context::$first_id has no type specified.                               
  :49    Property FreshRSS_Context::$next_id has no type specified.                                
  :50    Property FreshRSS_Context::$id_max has no type specified.                                 
  :51    Property FreshRSS_Context::$sinceHours has no type specified.                             
  :53    Property FreshRSS_Context::$isCli has no type specified.                                  
  :58    Method FreshRSS_Context::initSystem() has parameter $reload with no type specified.       
  :73    Method FreshRSS_Context::initUser() has no return type specified.                         
  :73    Method FreshRSS_Context::initUser() has parameter $userMustExist with no type specified.  
  :73    Method FreshRSS_Context::initUser() has parameter $username with no type specified.       
  :196   Method FreshRSS_Context::isStateEnabled() has no return type specified.                   
  :204   Method FreshRSS_Context::getRevertState() has no return type specified.                   
  :217   Method FreshRSS_Context::currentGet() has no return type specified.                       
  :217   Method FreshRSS_Context::currentGet() has parameter $array with no type specified.        
  :278   Method FreshRSS_Context::isCurrentGet() has parameter $get with no type specified.        
  :314   Method FreshRSS_Context::_get() has parameter $get with no type specified.                
 ------ ------------------------------------------------------------------------------------------ 


                                                                                                                        
 [ERROR] Found 27 errors
```

Alkarex (changes requested):
Please pass at least PHPStan level 6, and more if possible

app/Models/Context.php
```diff
@@ -55,7 +55,7 @@ class FreshRSS_Context {
 	/**
 	 * Initialize the context for the global system.
 	 */
-	public static function initSystem($reload = false) {
+	public static function initSystem($reload = false): ?FreshRSS_SystemConfiguration {
```
aledeg: Please check if this syntax works with PHP 7.2. I am not sure it does

app/Models/Context.php
```diff
@@ -67,9 +67,13 @@ public static function initSystem($reload = false) {
...
 	/**
-	 * Initialize the context for the current user.
+	 * Initialize the context for the current user
+	 *
+	 * @return false|FreshRSS_UserConfiguration|null
+	 *
+	 * @throws Minz_ConfigurationParamException
 	 */
-	public static function initUser($username = '', $userMustExist = true) {
+	public static function initUser(?string $username = '', $userMustExist = true) {
```
aledeg: Same here

app/Models/Context.php
```diff
@@ -177,43 +182,41 @@ public static function updateUsingRequest() {
...
 
 	/**
 	 * Return the current get as a string or an array.
 	 *
 	 * If $array is true, the first item of the returned value is 'f' or 'c' and
 	 * the second is the id.
+	 * @return array|string|void
 	 */
-	public static function currentGet($array = false) {
+	public static function currentGet(?bool $array = false) {
```
aledeg: Same here

app/Models/Context.php
ColonelMoutarde: this syntax work for PHP7.2 since PHP7.1 https://kinsta.com/blog/php-7-1-0/#nullable-types

ColonelMoutarde: https://www.php.net/manual/en/migration71.new-features.php
https://www.php.net/manual/en/migration72.new-features.php

Alkarex:
Still not passing PHPStan level 6, right?

ColonelMoutarde:
> Still not passing PHPStan level 6, right?

Pass phpstan6, but not 9

Alkarex:
Nope, still not level 6. Did you try the command from my comment above?

```console
$ vendor/bin/phpstan analyse --level 6 app/Models/Context.php
Note: Using configuration file /home/alexandre/GitHub/FreshRSS/phpstan.neon.
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------------------------------------------------------- 
  Line   Context.php                                                                                            
 ------ ------------------------------------------------------------------------------------------------------- 
  :19    Property FreshRSS_Context::$categories has no type specified.                                          
  :20    Property FreshRSS_Context::$tags has no type specified.                                                
  :22    Property FreshRSS_Context::$name has no type specified.                                                
  :23    Property FreshRSS_Context::$description has no type specified.                                         
  :25    Property FreshRSS_Context::$total_unread has no type specified.                                        
  :26    Property FreshRSS_Context::$total_starred has no type specified.                                       
  :32    Property FreshRSS_Context::$get_unread has no type specified.                                          
  :33    Property FreshRSS_Context::$current_get has no type specified.                                         
  :41    Property FreshRSS_Context::$next_get has no type specified.                                            
  :43    Property FreshRSS_Context::$state has no type specified.                                               
  :44    Property FreshRSS_Context::$order has no type specified.                                               
  :45    Property FreshRSS_Context::$number has no type specified.                                              
  :48    Property FreshRSS_Context::$first_id has no type specified.                                            
  :49    Property FreshRSS_Context::$next_id has no type specified.                                             
  :50    Property FreshRSS_Context::$id_max has no type specified.                                              
  :51    Property FreshRSS_Context::$sinceHours has no type specified.                                          
  :53    Property FreshRSS_Context::$isCli has no type specified.                                               
  :58    Method FreshRSS_Context::initSystem() has parameter $reload with no type specified.                    
  :76    Method FreshRSS_Context::initUser() has parameter $userMustExist with no type specified.               
  :219   Method FreshRSS_Context::currentGet() return type has no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type              
  :278   Method FreshRSS_Context::isCurrentGet() has parameter $get with no type specified.                     
  :314   Method FreshRSS_Context::_get() has parameter $get with no type specified.                             
 ------ ------------------------------------------------------------------------------------------------------- 


                                                                                                                        
 [ERROR] Found 22 errors
```

ColonelMoutarde:
> Nope, still not level 6. Did you try the command from my comment above?
> 
> ```
> $ vendor/bin/phpstan analyse --level 6 app/Models/Context.php
> Note: Using configuration file /home/alexandre/GitHub/FreshRSS/phpstan.neon.
>  1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
> 
>  ------ ------------------------------------------------------------------------------------------------------- 
>   Line   Context.php                                                                                            
>  ------ ------------------------------------------------------------------------------------------------------- 
>   :19    Property FreshRSS_Context::$categories has no type specified.                                          
>   :20    Property FreshRSS_Context::$tags has no type specified.                                                
>   :22    Property FreshRSS_Context::$name has no type specified.                                                
>   :23    Property FreshRSS_Context::$description has no type specified.                                         
>   :25    Property FreshRSS_Context::$total_unread has no type specified.                                        
>   :26    Property FreshRSS_Context::$total_starred has no type specified.                                       
>   :32    Property FreshRSS_Context::$get_unread has no type specified.                                          
>   :33    Property FreshRSS_Context::$current_get has no type specified.                                         
>   :41    Property FreshRSS_Context::$next_get has no type specified.                                            
>   :43    Property FreshRSS_Context::$state has no type specified.                                               
>   :44    Property FreshRSS_Context::$order has no type specified.                                               
>   :45    Property FreshRSS_Context::$number has no type specified.                                              
>   :48    Property FreshRSS_Context::$first_id has no type specified.                                            
>   :49    Property FreshRSS_Context::$next_id has no type specified.                                             
>   :50    Property FreshRSS_Context::$id_max has no type specified.                                              
>   :51    Property FreshRSS_Context::$sinceHours has no type specified.                                          
>   :53    Property FreshRSS_Context::$isCli has no type specified.                                               
>   :58    Method FreshRSS_Context::initSystem() has parameter $reload with no type specified.                    
>   :76    Method FreshRSS_Context::initUser() has parameter $userMustExist with no type specified.               
>   :219   Method FreshRSS_Context::currentGet() return type has no value type specified in iterable type array.  
>          💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type              
>   :278   Method FreshRSS_Context::isCurrentGet() has parameter $get with no type specified.                     
>   :314   Method FreshRSS_Context::_get() has parameter $get with no type specified.                             
>  ------ ------------------------------------------------------------------------------------------------------- 
> 
> 
>                                                                                                                         
>  [ERROR] Found 22 errors
> ```

Fixed for level 6

app/Models/Context.php
```diff
@@ -177,61 +229,59 @@ public static function updateUsingRequest() {
 		self::$order = Minz_Request::param(
 			'order', self::$user_conf->sort_order
 		);
-		self::$number = intval(Minz_Request::param('nb', self::$user_conf->posts_per_page));
+		self::$number = (int)Minz_Request::param('nb', self::$user_conf->posts_per_page);
```
Alkarex, line 230: In PHPStan level 9, such changes introduce a new error: `Cannot cast mixed to int.`
Cf. https://github.com/phpstan/phpstan/issues/4490#issuecomment-773119128

Alkarex: Although `intval()` and `(int)` might be doing exactly the same thing (I have not checked), there are semantically not exactly equivalent

app/Controllers/indexController.php
```diff
@@ -218,7 +218,7 @@ public function opmlAction() {
 				break;
 			case 'f':
 				// We most likely already have the feed object in cache
-				$feed = FreshRSS_CategoryDAO::findFeed($categories, $id);
+				$feed = FreshRSS_CategoryDAO::findFeed($categories, (int)$id);
```
Alkarex: I have changed to cast higher up: https://github.com/FreshRSS/FreshRSS/pull/5064/commits/727ba1331c014cd6e1a69dbba9594cc7779e2aaa

app/Models/Context.php
```diff
@@ -242,37 +292,37 @@ public static function currentGet($array = false) {
 	}
 
 	/**
-	 * @return bool true if the current request targets all feeds (main view), false otherwise.
+	 * true if the current request targets all feeds (main view), false otherwise.
```
Alkarex: Removing the `@return` instruction breaks the understanding of the PHPDocs.
Same comment than https://github.com/FreshRSS/FreshRSS/pull/5106#discussion_r1130735488

composer.json
```diff
@@ -25,9 +25,11 @@
         "ext-gmp": "*",
         "ext-intl": "*",
         "ext-json": "*",
+        "ext-libxml": "*",
```
Alkarex, line 28: Although this is fine, I would have preferred another PR, as it is not related to the topic of the current PR.
No need to revert, though, but for another time.

app/Models/Context.php
```diff
@@ -215,25 +261,25 @@ public static function getRevertState($state) {
 	 * the second is the id.
 	 * @return string|array{string,bool|int}
 	 */
-	public static function currentGet($array = false) {
+	public static function currentGet(?bool $bool = false) {
```
Alkarex: Why renaming this variable? It was more informative before

app/Models/Context.php
```diff
@@ -9,27 +9,47 @@ final class FreshRSS_Context {
...
+	 * @var string
+	 */
 	public static $description = '';
-
+	/**
+	 * @var int
+	 */
 	public static $total_unread = 0;
+	/**
+	 * @var array<string, int>
```
Alkarex: Please use tuples syntax instead. Fix coming

app/Models/Context.php
```diff
@@ -9,27 +9,47 @@ final class FreshRSS_Context {
...
 		'read' => 0,
 		'unread' => 0,
 	);
-
+	/**
+	 * @var int
+	 */
 	public static $get_unread = 0;
+	/**
+	 * @var array<string, bool>
```
Alkarex: Wrong type, the values can also be integers. Fix coming

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/5064/commits/194911b263098d91764f92bfc9c7a9a4843a67ef

app/Models/Context.php
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/5064/commits/194911b263098d91764f92bfc9c7a9a4843a67ef

PR #4944, opened 2022-12-14
Better enclosures

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/4702
Improvement of https://github.com/FreshRSS/FreshRSS/pull/2898

PR #5072, opened 2023-02-05
Improved: Enclosures: HTML from Entry.php to enclosures.phtml

math-GH:
Next step after #4944 this PR cuts out the HTML from `Entry.php` and put it into `enclosure.phtml`.

(Ref #4943 where I tried it first time)

Important: This PR creates not the best `.phtml`, because of many `echo` for printing the HTML.

I would prefer to have a PR review here, while it is clear how I cut out the HTML from `Entry.php` and did not change the HTML structure. Improvement PR will follow.

General goal: I want to improve the displaying of enclosures. One step is to bring the output into the `.phtml`. Later I will improve this `.phtml`

- HTML from `Entry.php` to `enclosure.phtml` as pure extract.

1. open the normal view or reader view
2. take a look onto articles with enclosures (I support some feeds with enclosures in my repo: https://github.com/math-GH/testfeed)
3. the enclosures will be displayed (and nothing should changed in comparison to before)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/views/helpers/index/enclosures.phtml
```diff
@@ -0,0 +1,53 @@
...
+		}
+	}
+
+	switch ($enclosures[$i]['type']) {
+		case 'image':
+			echo '<p class="enclosure-content"><img src="' . $enclosures[$i]['link'] . '" alt="" title="' . $enclosures[$i]['title'] . '" /></p>';
+			break;
+		case 'audio':
+			echo '<p class="enclosure-content"><audio preload="none" src="' . $enclosures[$i]['link']
+			. ($enclosures[$i]['data-length'] != '' ? '" data-length="' . $enclosures[$i]['data-length'] . '"' : '')
```
Alkarex: I would prefer an `empty($enclosures[$i]['data-length'])` as the test, to avoid relying on the fact that those array entries are always created before.

math-GH: Done

app/Models/Entry.php
```diff
@@ -173,57 +168,48 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
...
+			$this->content_enclosures[] = $temp;
 		}
 
 		return $content;
 	}
 
+	public function content_enclosures() {
+		return $this->content_enclosures;
+	}
+
```
ColonelMoutarde: Add return type

ColonelMoutarde: Suggestion: I don't think that mixing the view and the logic in the same file is a good idea. there is surely some refactoring to do

math-GH: > Add return type

Done. Is it correct now? (I am not so experienced with return types/typehints yet)

math-GH: > Suggestion: I don't think that mixing the view and the logic in the same file is a good idea. there is surely some refactoring to do

I invite you to help here. My goal for this PR is "just" to cut out the HTML. Later we can do more cleaning and help with the backend part is very welcomed.

app/Models/Entry.php
```diff
@@ -173,57 +168,51 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
...
+			$this->content_enclosures[] = $temp;
 		}
 
 		return $content;
 	}
 
+	/**
+	 * @return array<int,string>|null
+	 */
+	public function content_enclosures(): array {
```
math-GH: @Alkarex @ColonelMoutarde I need your help. I cannot find a correct PHPstand comment for this Array. Could you please support here?

ColonelMoutarde: go to https://framateam.org/freshrss

app/Models/Entry.php
```diff
@@ -22,6 +22,7 @@ class FreshRSS_Entry extends Minz_Model {
 	private $authors;
 	/** @var string */
 	private $content;
+	private $content_enclosures = [];
```
ColonelMoutarde:
```suggestion
        /** @var string[]|null */
	private $content_enclosures;
```

app/Models/Entry.php
ColonelMoutarde: Your return is always an array, not null.

if you want twice try this:

... ): ?array {

ColonelMoutarde (approved):

app/Models/Entry.php
math-GH: I fixed it

app/views/helpers/index/enclosures.phtml
```diff
@@ -0,0 +1,53 @@
...
+if (empty($enclosures[0]['main-thumbnail']['src'])) { ?>
+	<figure class="enclosure">
+		<p class="enclosure-content">
+			<img class="enclosure-thumbnail" src="<?= $enclosures[0]['main-thumbnail']['src'] ?>" alt="" />
+		</p>
+	</figure>
+	<?php
+	$i++;
+}
+for ($i; $i < count($enclosures); $i++) {
```
ColonelMoutarde:
```suggestion
$countedEnclosures = count($enclosures)
for ($i; $i < $countedEnclosures; $i++) {
```
Best performances

app/Models/Entry.php
```diff
@@ -173,57 +169,51 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
...
 			if (self::enclosureIsImage($enclosure)) {
-				$content .= '<p class="enclosure-content"><img src="' . $elink . '" alt="" title="' . $etitle . '" /></p>';
+				$temp['type'] = 'image';
 			} elseif ($medium === 'audio' || strpos($mime, 'audio') === 0) {
-				$content .= '<p class="enclosure-content"><audio preload="none" src="' . $elink
-					. ($length == null ? '' : '" data-length="' . intval($length))
-					. ($mime == '' ? '' : '" data-type="' . htmlspecialchars($mime, ENT_COMPAT, 'UTF-8'))
-					. '" controls="controls" title="' . $etitle . '"></audio> <a download="" href="' . $elink . '">💾</a></p>';
+				$temp['type'] = 'audio';
+				$temp['data-length'] = $length == null ? '' : intval($length);
```
ColonelMoutarde, line 209:
```suggestion
				$temp['data-length'] = $length === null ? '' : (int)$length;
```

ColonelMoutarde: Modernize your code

app/Models/Entry.php
```diff
@@ -173,57 +169,51 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
...
+				$temp['data-type'] = htmlspecialchars($mime, ENT_COMPAT, 'UTF-8');
 			} elseif ($medium === 'video' || strpos($mime, 'video') === 0) {
-				$content .= '<p class="enclosure-content"><video preload="none" src="' . $elink
-					. ($length == null ? '' : '" data-length="' . intval($length))
-					. ($mime == '' ? '' : '" data-type="' . htmlspecialchars($mime, ENT_COMPAT, 'UTF-8'))
-					. '" controls="controls" title="' . $etitle . '"></video> <a download="" href="' . $elink . '">💾</a></p>';
+				$temp['type'] = 'video';
+				$temp['data-length'] = $length == null ? '' : intval($length);
```
ColonelMoutarde, line 213: same...

ColonelMoutarde:
Please use the strict comparison in new code

app/Models/Entry.php
Alkarex: `==` is safer than `===` here and is thus perfectly fine.
We do want to exclude all falsy cases such as `''` or `0` and  `===` fails to do that.
But `== ''` would be semantically marginally better, as I do not believe it can ever be `null`.

Alkarex:
@math-GH A little merge conflict to fix here

math-GH:
Could someone please help here to fix the PHPstan issues? I am a bit lost here

Alkarex:
Here you are @math-GH https://github.com/FreshRSS/FreshRSS/pull/5072/commits/a8264ead3d5a635ebd758ea5da46ef174cb784e1

math-GH:
Thanks. I checked the result again. Would say ready to merge

Alkarex:
It looks like there are two functions now doing something similar, namely, `->enclosures()` and `->content_enclosures()`.
I have not looked at all the details yet, but couldn't we do with the existing `->enclosures()`, and skip the `->content_enclosures` and `->content_enclosures()`?

math-GH:
I do not understand  `->enclosures()` and cannot say if it similar to `->content_enclosures()`.

Would it make sense to merge this result and optimize the source code with a next PR?

Alkarex:
I have not yet checked all details, but I believe it breaks:
* the output of enclosures in the API outputs
* the lazy loading of enclosure images

I will try to do something in the coming days

math-GH:
> I have not yet checked all details, but I believe it breaks:
> 
>     * (...)
>     * the lazy loading of enclosure images
> 


The lazy loading does work fine for me

app/Models/Entry.php
```diff
@@ -17,6 +17,8 @@ class FreshRSS_Entry extends Minz_Model {
 	private $authors;
 	/** @var string */
 	private $content;
+	/** @var array<array{'url':string,'type'?:string,'medium'?:string,'length'?:int,'title':string,'description'?:string,'credit'?:string,'height'?:int,'width'?:int,'thumbnails'?:array<string>}>|null */
+	private $content_enclosures;
```
ColonelMoutarde, line 21:
```suggestion
	private ?array $content_enclosures;
```

math-GH:
How to proceed here? Does it make sense to invest time to solve the conflicts?

Inverle:
It would be a nice change

PR #5074, opened 2023-02-06
Increase max HTTP timeout

Alkarex:
* Increase custom maximum HTTP request timeout from 2 minutes to 15 minutes;
    * Reason: I have some feeds generated by RSS Bridge, which can take several minutes.
* Increase default HTTP request timeout from 15 to 20s.
    * Reason: I regularly observe feeds, which are slow to answer.


Frenzie (approved):
2 mins is a tad short as a maximum for a custom setting; hard to imagine needing more than 5 but it definitely can't hurt :-)

PR #5076, opened 2023-02-06
XML+XPath

### LINKED ISSUE #5075: XML+XPath
A simplified XML case: https://gist.githubusercontent.com/mgnsk/39ffd7f8e0a24038373a53f76c317d93/raw/0c772312121ce089bc4c14082ae54e8882298d74/freshrss_test.xml

* `//array/object` item and `./string[@name="title"]` title gives only a single entry with the title `Meta title`
* `/array/object` item and `./string[@name="title"]` title gives an empty feed

---

Now the same XML but the inner array has a name attribute which is similar to the actual XML I dealt with: https://gist.githubusercontent.com/mgnsk/0aab124cfd3516cb8ccddb8d39d5c5a7/raw/e9fe30a5cdedcc736027353e08b4be994f91beb2/freshrss_test.xml

* `//array/object` item and `./string[@name="title"]` title gives me 4 entries
* `/array/object` item and `./string[@name="title"]` title gives empty feed

I was finally able to solve it with
* `//array[count(@*)=0]/object` item and `./string[@name="title"]` title gives correct 2 entries `Item 1` and `Item 2`
but still wondering why just `/array/object` couldn't work.

_Originally posted by @mgnsk in https://github.com/FreshRSS/FreshRSS/discussions/5071#discussioncomment-4886226_

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/5075
Implementation allowing to take an XML document as input using an XML parser (instead of an HTML parser for HTML+XPath)

image

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -387,12 +387,12 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 					$simplePie = $simplePiePush;	//Used by WebSub
 				} elseif ($feed->kind() === FreshRSS_Feed::KIND_HTML_XPATH) {
 					$simplePie = $feed->loadHtmlXpath();
-					if ($simplePie == null) {
+					if ($simplePie === null) {
 						throw new FreshRSS_Feed_Exception('HTML+XPath Web scraping failed for [' . $feed->url(false) . ']');
 					}
 				} elseif ($feed->kind() === FreshRSS_Feed::KIND_XML_XPATH) {
 					$simplePie = $feed->loadHtmlXpath();
-					if ($simplePie == null) {
+					if ($simplePie === null) {
 						throw new FreshRSS_Feed_Exception('XML+XPath parsing failed for [' . $feed->url(false) . ']');
 					}
 				} else {
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -81,6 +81,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				$feed->load(true);	//Throws FreshRSS_Feed_Exception, Minz_FileNotExistException
 				break;
 			case FreshRSS_Feed::KIND_HTML_XPATH:
+			case FreshRSS_Feed::KIND_XML_XPATH:
 				$feed->_website($url);
 				break;
 		}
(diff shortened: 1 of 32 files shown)


Alkarex:
@mgnsk Would you be able to test this patch?

See the new XML+XPath mode:
image

mgnsk:
> @mgnsk Would you be able to test this patch?
> 
> 
> 
> See the new XML+XPath mode:
> 
> image
> 
> 

Thanks, I'll check it out tomorrow evening.

Alkarex:
Related experimental JSON+XPath functionality: https://github.com/FreshRSS/FreshRSS/pull/5079

app/Controllers/feedController.php
```diff
@@ -385,10 +386,15 @@ public static function actualizeFeed($feed_id, $feed_url, $force, $simplePiePush
...
 					$simplePie = $simplePiePush;	//Used by WebSub
 				} elseif ($feed->kind() === FreshRSS_Feed::KIND_HTML_XPATH) {
-					$simplePie = $feed->loadHtmlXpath(false, $isNewFeed);
+					$simplePie = $feed->loadHtmlXpath();
 					if ($simplePie == null) {
 						throw new FreshRSS_Feed_Exception('HTML+XPath Web scraping failed for [' . $feed->url(false) . ']');
 					}
+				} elseif ($feed->kind() === FreshRSS_Feed::KIND_XML_XPATH) {
+					$simplePie = $feed->loadHtmlXpath();
+					if ($simplePie == null) {
```
ColonelMoutarde: suggestion :  use strict comparaison
```suggestion
					if ($simplePie === null) {
```

app/Models/Feed.php
```diff
@@ -17,6 +17,11 @@ class FreshRSS_Feed extends Minz_Model {
 	 * @var int
 	 */
 	const KIND_HTML_XPATH = 10;
+	/**
+	 * Normal XML with XPath scraping
+	 * @var int
+	 */
+	const KIND_XML_XPATH = 15;
```
ColonelMoutarde, line 24: recommandation:  use visibilty to constant

Alkarex: I agree when there is a mix of constants with different visibilities, or potentially the first time we introduce a constant in a class. Otherwise it does not matter as they are public by default, just like classes, and I would rather keep the same syntax than the other surrounding constants.

app/Controllers/feedController.php
Alkarex: Done

mgnsk:
> @mgnsk Would you be able to test this patch?

Yes this works as expected. If the XML contains HTML, it must use a CDATA section just like RSS.

Alkarex:
Thanks for the feedback!

Issue #1551, opened 2017-05-26
[Feature request] JSON Feed


Epy:
Hello
I drop this here only for later, if someone wants to add this or I may try to do it by myself

A couple of devs is pushing a JSON Feed format to replace RSS/Atom ones based on XML:
https://jsonfeed.org/version/1

This feels like already a new format although we didn't already push XML to its limits but it's not so stupid: JSON is lighter and they thought of compatbility upwards

aledeg:
I think it's doable when a lib to convert JSON feed to Atom is released. Or maybe included in SimplePie library. See https://github.com/simplepie/simplepie/issues/519 for more info on that.

Frenzie:
I hope that won't lead to an increase of websites like this one: https://frerejones.com/blog

Without JS there's nothing there. With JS enabled it loads the blog content from https://frerejones.com/api/articles

Oh well, adding a converter to SimplePie and/or rss-bridge should be simple enough I guess.

> For most developers, JSON is *far* easier to read and write than XML. 

Nonsense. JSON's inability to quickly comment out a line makes it much harder to write than XML in spite of it otherwise being mostly easier. YAML or some such is much more straightforward imho.

Alkarex:
I am in favour of supporting this format when it starts being used, but I otherwise do not think it helps to have yet another format. The current problems with RSS/Atom are not due to XML, quite the contrary. If the spec is very precise and very strict, then it might be OK in the long run; if not, then it will be yet another problematic format to support.

Frenzie:
Atom's good; RSS 2 isn't problematic either. The wild days were in the early 2000s with the conflicting RSS standards, half a lifetime ago (for me). 

Incidentally, glancing at that page again I came across this beauty.

> `version` (required, string) is the URL of the version of the format the feed uses. This should appear at the very top, though we recognize that not all JSON generators allow for ordering.

There's a format where order is significant. It's not JSON. I think it starts with an X…

Frenzie:
> I think you forgot about it stuck_out_tongue

Haha, indeed, except I was less diplomatic in the words I chose a few years ago than those I chose yesterday. I definitely don't see the point since at best it's just Atom in JSON, and most of their arguments read like a parody of trendy hackers.

Like this one:

> Developers may groan at picking up an XML parser, but decoding JSON is often just a single line of code.

That doesn't make a lick of sense. Reading junk like that is exactly why my opinion went from /shrug to negative three years ago, and the same applies today.

[...]

PR #5079, opened 2023-02-07
JSON+XPath (experimental)

Alkarex:
Based on https://github.com/FreshRSS/FreshRSS/pull/5076
Transforms a JSON document to XML before using the XML+XPath method.


Alkarex:
Test: https://gist.githubusercontent.com/Alkarex/bddeaf3b4034ad3e877a165e121efb51/raw/75e579c7fc7ec4b00ae4d7bd7e711e815c979d8b/hello.json

* Items: `/array/object`
* Titles: `descendant::value[@key="title"]`
* Contents: `descendant::value[@key="body"]`

image

Original JSON:

```json
[
	{
		"title": "Item1",
		"body": "<b>Hello</b>"
	},
	{
		"title": "Item2",
		"body": "World"
	}
]
```

Intermediate XML:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<array>
  <object>
    <value key="title">
      <string><![CDATA[Item1]]></string>
    </value>
    <value key="body">
      <string><![CDATA[<b>Hello</b>]]></string>
    </value>
  </object>
  <object>
    <value key="title">
      <string><![CDATA[Item2]]></string>
    </value>
    <value key="body">
      <string><![CDATA[World]]></string>
    </value>
  </object>
</array>
```

See more examples of JSON to XML conversions in
https://github.com/FreshRSS/FreshRSS/blob/27130a474c45444e72462e550a3f3e925962c8c6/tests/app/Services/JsonServiceTest.php

app/Services/JsonService.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+if (!function_exists('array_is_list')) {
```
ColonelMoutarde, line 3: suggestion: Add declare(strict_types=1); in new class and files

tests/app/Services/JsonServiceTest.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
```
ColonelMoutarde, line 3: [declare(strict_types=1);](suggestion: Add declare(strict_types=1); in new class and files)

app/Services/JsonService.php
Alkarex: At the moment, I do not believe there is much value in doing so. We can catch errors during development and test phase with e.g. PHPStan instead of risking some crashes in production. I suggest we move to that when we are higher up with our PHPStan levels.

mgnsk:
That's great. In https://github.com/FreshRSS/FreshRSS/pull/5076 I'm actually using it to scrape [Mastodon API](https://docs.joinmastodon.org/client/public/#timelines) which I'm running through a json2xml service. Will check out later how both these PRs work, hopefully I can drop my json2xml hack after this gets implemented.

There's also a thing called JSON Feed which has a consistent structure but I haven't seen it used much.

app/Services/JsonService.php
ColonelMoutarde: when you add a new class, it is wise to add this at the beginning of the file so that php does not transform the types in an uncontrollable way.

tests/app/Services/JsonServiceTest.php
Alkarex: But good idea for test files

mgnsk:
This kind of works but if it could treat ~~every~~ content string as raw HTML, it would work better.

For example, consider the "content" string is used as both title and content:
```json
{"content": "<b>Test</b>"}
```
The feed item's title becomes something like `bTestb` and the content is `<b>Test</b>`. If it could run something like `strip_tags()` for the title and treat content as raw HTML then it would work for the Mastodon API output. The previous XML patch seemed to work this way, i.e. strip tags from title and with CDATA, keep content HTML.

Also it doesn't feel neat to specify XPath for JSON since there is a conversion step. It would be nicer to specify JSONPath but this adds more complexity. That would mean scraping JSON APIs would become simple but then probably would need more authentication methods if not only public APIs are scraped (HTTP headers, OAuth, etc).

Alkarex:
Thanks for the feedback. Could you please try again?

mgnsk:
I don't think it's intuitive to use XPath mixed with JSON, it wouldn't be a feature I'd like to see in FreshRSS. For pure XML, XPath is fine though. The recently proposed XML patch would solve the XPath specificity problem. Regarding JSON I think if anything it should support either the pure JSON Feed with `content_html` or JSON API with multiple authentication options and JSONPath selection.

I'm gonna link an issue here https://github.com/mastodon/mastodon/issues/17269. I don't think FreshRSS needs JSON if Mastodon had proper RSS output.

Alkarex:
Mastodon used to have RSS / ATOM and even WebSub, which used to work great with clients such as FreshRSS. I even had a PR for some related fixes https://github.com/mastodon/mastodon/pull/9302

There are other software in the Fediverse than Mastodon, which have a greater interest in larger compatibility. For instance Friendica is both compatible with ActivityPub and several other protocols, including of course RSS / ATOM. I believe there are also some ActivityPub to RSS bridges.

Back to our JSON feature here. It would be for many more cases than for consuming Mastodon timelines.

I agree that XPath is far from ideal for JSON, but I am not aware of many good options in the JSON world (especially not for a light PHP project). I am using JSONata in other projects, which is quite nice but large, language-dependant, and not quite standard. [JsonPath](https://goessner.net/articles/JsonPath/) seems forked into various versions, not standard at all, and seems to lack basic expressions such as string manipulations (e.g. concatenation).

JSON Feed is something we might add in the future https://github.com/FreshRSS/FreshRSS/issues/1551 , but I have yet to see it in use in places where there is not already a proper RSS / ATOM feed. And supporting JSON Feed would not allow consuming other random JSON documents.

Consuming various (proprietary) APIs would be out of scope for the core of FreshRSS, but could be considered for FreshRSS extensions. More realistically, that seems like a job for an RSS bridge service.

Regarding this PR, let's see whether we get more feedback. We could try to re-work the JSON to XML conversion to make it more pleasant to work with. It is a relatively short piece of code to enable this JSON+XPath feature, so it is not so costly.

Frenzie:
Since when has Mastodon removed RSS?

Anyway, I concur. This is a really good way to introduce JSON scraping.

app/i18n/nl/sub.php
```diff
@@ -121,7 +121,9 @@
 				'relative' => 'XPath (relatief naar bericht) voor:',
 				'xpath' => 'XPath voor:',
 			),
+			'json_xpath' => 'JSON + XPath (experimental)',	// TODO
```
Frenzie:
```suggestion
			'json_xpath' => 'JSON + XPath (experimenteel)',
```

app/i18n/nl/sub.php
```diff
@@ -121,7 +121,9 @@
...
 			'rss' => 'RSS / Atom (standaard)',
+			'xml_xpath' => 'XML + XPath',	// TODO
```
Frenzie: Can be ignored.

Alkarex:
> Since when has Mastodon removed RSS?

I had not tested recently, but there still seems to be some RSS features in Mastodon according to the source code at least
https://github.com/mastodon/mastodon/search?q=rss

Quick test seems to work as well:

```sh
curl -v -H 'Accept: application/rss+xml,application/atom+xml,application/xml;q=0.9' 'https://framapiaf.org/@davidrevoy'
```

Does not seem to work with FreshRSS though. I am assuming a picky HTTP Accept negotiation. Will require investigation.

Edit: yes, there seems to be a bug in Mastodon in respecting HTTP Accept priorities. This one (used by SimplePie) returns HTML:

```sh
curl -v -H 'Accept: application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1' 'https://framapiaf.org/@davidrevoy'
```

Alkarex:
@mgnsk Please try https://github.com/FreshRSS/FreshRSS/pull/5083 , which allows consuming an RSS feed directly from Mastodon, e.g. https://framapiaf.org/@davidrevoy

Alkarex:
Ah, actually, a more precise URL will circumvent Mastodon's (broken?) HTTP Accept negotiation. E.g. https://framapiaf.org/@davidrevoy.rss

mgnsk:
> @mgnsk Please try #5083 , which allows consuming an RSS feed directly from Mastodon, e.g. https://framapiaf.org/@davidrevoy

This seems to allow subscribing directly to `https://framapiaf.org/@davidrevoy` whereas I never had problems by using the explicit URL `https://framapiaf.org/@davidrevoy.rss`. It's just that Mastodon's user page HTML is broken:
```html
<link href='' rel='alternate' type='application/rss+xml'>
```

Alkarex:
I got confused: @mgnsk would like to retrieve the timeline of a full Mastodon instance, while the bug I discussed is for accessing the RSS feed of a single Mastodon user, and the bug seems to vary depending on Mastodon version.

mgnsk:
> I got confused: @mgnsk would like to retrieve the timeline of a full Mastodon instance, while the bug I discussed is for accessing the RSS feed of a single Mastodon user, and the bug seems to vary depending on Mastodon version.

Yeah, https://github.com/FreshRSS/FreshRSS/pull/5083 is not related to this.

mgnsk:
> Thanks for the feedback. Could you please try again?

I tried the CDATA fix, it works for content and HTML gets rendered. But if a title contains HTML, it's still rendered as `<b>Title</b>` -> `bTitleb`. Some tag stripping should be needed there.

app/i18n/nl/sub.php
Alkarex: Done

Alkarex:
> I tried the CDATA fix, it works for content and HTML gets rendered. But if a title contains HTML, it's still rendered as `<b>Title</b>` -> `bTitleb`. Some tag stripping should be needed there.

@mgnsk It seems to work fine for me with this test https://gist.githubusercontent.com/Alkarex/bddeaf3b4034ad3e877a165e121efb51/raw/75e579c7fc7ec4b00ae4d7bd7e711e815c979d8b/hello.json

Could you please share an example to reproduce the problem?

mgnsk:
> Could you please share an example to reproduce the problem?

Sorry, it works now that I've tried again. Must've looked at some old data.

mrnoname1000:
Another option would be [this jq binding](https://github.com/kjdev/php-ext-jq) which would provide a more standard, ergonomic interface but also add a dependency on the jq extension being installed. Maybe it could be an optional feature?

Alkarex:
> Another option would be [this jq binding](https://github.com/kjdev/php-ext-jq) which would provide a more standard, ergonomic interface but also add a dependency on the jq extension being installed. Maybe it could be an optional feature?

It looks like it would add some binary dependencies, meaning that we would have to compile for multiple platforms, and also require additional exec permissions

mrnoname1000:
> It looks like it would add some binary dependencies, meaning that we would have to compile for multiple platforms, and also require additional exec permissions

I don't think it would be a problem for bare-metal installs as the onus would be on the administrator to install the extension, but for docker images you're right, unless it's been packaged for multiple platforms. Sadly I can't find a single distro that has packaged it so maybe it's not an option after all. 🙁

mrnoname1000:
Another option I found recently is [JMESPath](https://jmespath.org). It has a spec and compliant PHP library. Not sure what basic expressions are necessary to supplant XPath, but it does support string concatenation with the `join` function.

danburd:
Hello - this pull request recently came to my attention as a possible resolution to an issue I've noticed with one of my feeds. As of June 7, the topic pages of the Associated Press website - for example, https://apnews.com/hub/world-news - changed to rendering the articles with javascript rather than with HTML, making them not able to be scraped by the tools currently present in FreshRSS (as far as I know). But interestingly, all of the content that had been scraped is still present in a single line of JSON within the page's code. It's possible that the page's javascript uses the JSON to render HTML. Would a JSON scraper like the one proposed by this pull request be able to capture JSON embedded within HTML, or is that beyond the intended featureset? Thank you.

Alkarex:
> Would a JSON scraper like the one proposed by this pull request be able to capture JSON embedded within HTML, or is that beyond the intended featureset?

@danburd Yes, that is the idea of what this feature should allow. The current PR is only for pure JSON though. Extracting a JSON from HTML would be another step.

Chris3773:
> Another option I found recently is [JMESPath](https://jmespath.org). It has a spec and compliant PHP library. Not sure what basic expressions are necessary to supplant XPath, but it does support string concatenation with the `join` function.

Another spec I found was JsonPath, with a php library.  https://github.com/Galbar/JsonPath-PHP or https://github.com/SoftCreatR/JSONPath

Alkarex:
> Another spec I found was JsonPath, with a php library

Interesting, thanks! It does not look like there are any operators, though, in particular string split or concatenation, which I find quite important for such a job

mrnoname1000:
Now that you mention string split, I looked further into JMESPath and found that they still have yet to implement it ([see here](https://github.com/jmespath/jmespath.jep/issues/3)). Furthermore, jmespath.php has yet to implement [custom functions](https://github.com/jmespath/jmespath.php/pull/28). However, there is a [workaround](https://github.com/jmespath/jmespath.php/issues/20#issuecomment-78392503). I played around with it a bit and here's a simple sort function with no validation:
```php
require 'vendor/autoload.php';

$defaultDispatcher = new JmesPath\FnDispatcher();
$customDispatcher = function ($fn, array $args) use ($defaultDispatcher) {
    switch ($fn) {
        case 'split':
            return explode($args[0], $args[1]);
    }
    return $defaultDispatcher($fn, $args);
};
$runtime = new JmesPath\AstRuntime(null, $customDispatcher);

// Output: [0 => 'foo', 1 => 'bar', 2 => 'baz']
$array = $runtime('split(\' \', @)', 'foo bar baz');
```

ColonelMoutarde:
> Now that you mention string split, I looked further into JMESPath and found that they still have yet to implement it ([see here](https://github.com/jmespath/jmespath.jep/issues/3)). Furthermore, jmespath.php has yet to implement [custom functions](https://github.com/jmespath/jmespath.php/pull/28). However, there is a [workaround](https://github.com/jmespath/jmespath.php/issues/20#issuecomment-78392503). I played around with it a bit and here's a simple sort function with no validation:
> 
> ```
> require 'vendor/autoload.php';
> 
> $defaultDispatcher = new JmesPath\FnDispatcher();
> $customDispatcher = function ($fn, array $args) use ($defaultDispatcher) {
>     switch ($fn) {
>         case 'split':
>             return explode($args[0], $args[1]);
>     }
>     return $defaultDispatcher($fn, $args);
> };
> $runtime = new JmesPath\AstRuntime(null, $customDispatcher);
> 
> // Output: [0 => 'foo', 1 => 'bar', 2 => 'baz']
> $array = $runtime('split(\' \', @)', 'foo bar baz');
> ```

I hope one Time the project us all fonctions of composer. 🥲

mrnoname1000:
> I hope one Time the project us all fonctions of composer. 🥲

I don't quite understand but it should be easy to import it to `lib` and `require LIB_PATH . '/jmespath/jmespath.php/src/JmesPath.php';` for example. The only dependency is a polyfill for mbstring, which FreshRSS recommends anyway (and could be required by this feature).

ColonelMoutarde:
> > I hope one Time the project us all fonctions of composer. 🥲
> 
> I don't quite understand but it should be easy to import it to `lib` and `require LIB_PATH . 'jmespath/jmespath.php/src/JmesPath.php';` for example. The only dependency is a polyfill for mbstring, which FreshRSS recommends anyway (and could be required by this feature).

As it stands, it is possible to add this polyfill to the project by setting it to global function, checking beforehand that the function is not already present.

mrnoname1000:
> As it stands, it is possible to add this polyfill to the project by setting it to global function, checking beforehand that the function is not already present.

True, there is flexibility. I was thinking if we want to avoid another dependency, we could require the user to enable mbstring to use the JSON scraping feature at all. It all depends on what the maintainers want 🤷 

BTW jmespath.php hasn't been updated since 2021 so it should be easy to keep up to date 😉

Alkarex:
> Now that you mention string split, I looked further into JMESPath and found that they still have yet to implement it ([see here](https://github.com/jmespath/jmespath.jep/issues/3)). Furthermore, jmespath.php has yet to implement [custom functions](https://github.com/jmespath/jmespath.php/pull/28). However, there is a [workaround](https://github.com/jmespath/jmespath.php/issues/20#issuecomment-78392503). I played around with it a bit and here's a simple sort function with no validation:

Interesting. It would however require adding some of our own syntax, which I would be reluctant to do. But that is an option to keep in mind.

Alkarex:
* https://github.com/FreshRSS/FreshRSS/pull/5662

Alkarex:
Replaced by https://github.com/FreshRSS/FreshRSS/pull/5662

PR #5083, opened 2023-02-08
SimplePie use single constant for default HTTP Accept header

Alkarex:
Follow-up of https://github.com/simplepie/simplepie/commit/5d966b9f64a6034f89b019152d7cad9059f8819f
* Use single constant for default SimplePie HTTP Accept
* Missing headers in `SimplePie_Locator::body()`

~~Fix https://github.com/FreshRSS/FreshRSS/pull/5079#issuecomment-1421619176~~
~~The `*/*` breaks Mastodon content negotiation~~

### DIFF
--- a/lib/SimplePie/SimplePie/File.php
+++ b/lib/SimplePie/SimplePie/File.php
@@ -54,9 +54,6 @@
  */
 class SimplePie_File
 {
-
-	const DEFAULT_HTTP_ACCEPT = 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1';
-
 	var $url;
 	var $useragent;
 	var $success = true;
--- a/lib/SimplePie/SimplePie/File.php
+++ b/lib/SimplePie/SimplePie/File.php
@@ -55,7 +55,7 @@
 class SimplePie_File
 {
 
-	const DEFAULT_HTTP_ACCEPT = 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.7, text/html;q=0.6, text/plain;q=0.5, application/octet-stream;q=0.1';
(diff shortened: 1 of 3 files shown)


lib/SimplePie/SimplePie/File.php
```diff
@@ -54,6 +54,9 @@
  */
 class SimplePie_File
 {
+
+	const DEFAULT_HTTP_ACCEPT = 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.7, text/html;q=0.6, text/plain;q=0.5, application/octet-stream;q=0.1';
```
Frenzie: Hold on, are you absolutely sure you want to remove `*/*`? If Mastodon doesn't work that's Mastodon's problem. That doesn't mean we should break everything.

Frenzie: Also I'd like to emphasize there's nothing "fix" about "work around Mastodon's problem".

Alkarex: The thing is that we are not able to read other things, so we are actually lying when saying `*/*`.
But I do believe there is a bug in Mastodon, and I might try to have a look (but not in the coming days).
We can postpone to 1.22 to give it a longer try in edge.

Frenzie: How I've interpreted this header for the past decade is that the idea is to tell the server to send over its stuff and we'll give it a go, because it might be XML we can handle with some random incorrect mimetype.

Imho it's far from obvious that we should give broken Mastodon preference, because broken server configurations are usually just a minor user mistake of some sort in e.g. Apache or Nginx which are otherwise not broken.

Alkarex:
Partial upstream PR: https://github.com/simplepie/simplepie/pull/784

ColonelMoutarde (approved):

lib/SimplePie/SimplePie/File.php
Alkarex: I have reverted the change of HTTP Accept for now, especially since there seems to be an easy fix for Mastodon (only some versions seem broken), by being explicit about the RSS feed address. Example: https://framapiaf.org/@davidrevoy.rss

PR #5085, opened 2023-02-08
Add var dumper in dev environment only

ColonelMoutarde:

- add possibility to use symfony/var-dumper in dev environment **only**

1. use like this `dump($someVariable)` or `dd($someVariable)` to dump and die();

more informations https://symfony.com/doc/current/components/var_dumper.html

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
I dislike adding dependencies to third-party libraries, especially for trivial things like that.
You can always use it during your own local development without committing it.

PR #5087, opened 2023-02-09
Add test and type hinting

ColonelMoutarde:

- Add test and type hinting
- pass PhpStan Level 9

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] unit tested


tests/app/Models/LogDAOTest.php
```diff
@@ -0,0 +1,23 @@
...
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+class LogDAOTest extends TestCase {
+	private $logDAO;
+	protected function setUp(): void {
+		$this->logDAO = new FreshRSS_LogDAO();
+		file_put_contents(ADMIN_LOG, '[Wed, 08 Feb 2023 15:35:05 +0000] [notice] --- Migration 2019_12_22_FooBar: OK');
```
Alkarex: I am not sure I like this whole test. In particular, it overrides the local admin log, if I am not mistaken.

ColonelMoutarde: This is the principle of a unit test. I'm not hijacking the original functionality. I'm testing it from start to finish

Alkarex: Using the same principles, you could also drop the whole database, which I do not believe would be desirable.
This has too much side effects, as nothing says that the system is not in production.

ColonelMoutarde: We never run the tests in production. 
I was inspired by what was done for MigratorTest.php which also creates log files.
if we want to test the interactions with the database, then we create a database dedicated to the tests

Alkarex: You might not run tests in production, but people / users might. Yes, the Migrator stuff (which is not used anyway at the moment) also needs some rework, as it pollutes the disk with plenty of files which are not properly cleaned.

ColonelMoutarde: I added a function to clean up after the test

Alkarex: I suggest you use a temporary file instead of `ADMIN_LOG`, which might contain some information

ColonelMoutarde: I did what you asked me to do

aledeg: Maybe you could define envvars in the phpunit config file. This way, it works both in production environment and test environment

aledeg: Going this way will force us to use envvars for configuration. But I think it will be interesting. We can still use current values as default values when using envvars but this will ease configuration for powerusers.

app/Models/LogDAO.php
```diff
@@ -19,12 +22,20 @@ public static function lines() {
...
 		if (FreshRSS_Auth::hasAccess('admin')) {
 			file_put_contents(ADMIN_LOG, '');
 			file_put_contents(API_LOG, '');
 			file_put_contents(PSHB_LOG, '');
 		}
 	}
+
+	private static function currentUserLogPath(?string $logFileName = null): string {
+
+		if (is_null($logFileName)) {
```
Alkarex: No need for a function call there

```suggestion
		if ($logFileName !== null) {
```

Alkarex: But as other falsy values such as `''` should also be rejected, I would better like a more general test:

```suggestion
		if ($logFileName != '') {
```

ColonelMoutarde: Rememeber,  behavior between PHP7 and PHP8 changes for non-strict comparisons. These changes are not general but they may introduce bugs that are complicated to fix. Strict comparison will never change

Alkarex: Testing for only null is exactly what is introducing a bug ;-)
Acceptations should be strict, rejections should be lenient.

PR #5090, opened 2023-02-10
Type hinting for Factory.php up to PHPStan 9

ColonelMoutarde:
Closes #

- Type hinting for Factory.php up to PHPStan 9

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/Models/EntryDAO.php
```diff
@@ -1197,10 +1197,15 @@ public function listByIds(array $ids, string $order = 'DESC') {
 
 	/**
 	 * For API
-	 * @return array<string>
+	 * @return array<string>|false
 	 */
-	public function listIdsWhere($type = 'a', $id = '', $state = FreshRSS_Entry::STATE_ALL,
-			$order = 'DESC', $limit = 1, $firstId = '', $filters = null): array {
+	public function listIdsWhere($type = 'a',
```
Alkarex: Why no types here?

app/Models/Factory.php
```diff
@@ -46,7 +70,12 @@ public static function createTagDao($username = null) {
 		}
 	}
 
-	public static function createStatsDAO($username = null) {
+	/**
+	 * @return FreshRSS_StatsDAOSQLite|FreshRSS_StatsDAOPGSQL|FreshRSS_StatsDAO
```
Alkarex: `FreshRSS_TagDAO` should be enough

app/Models/EntryDAO.php
ColonelMoutarde: your right ! I have forgotten

ColonelMoutarde: i remember... I focused on Factory.php this file does not need to be fully modified for phpstan to pass. My target is factory.php

Alkarex: Yes fine, but then please avoid reformatting

ColonelMoutarde: done

app/Models/Factory.php
ColonelMoutarde: good remark, done

app/Controllers/entryController.php
```diff
@@ -78,10 +78,10 @@ public function readAction() {
 				$get = substr($get, 2);
```
Alkarex:
```suggestion
				$get = (int)substr($get, 2);
```

And remove the casts further down

app/Models/Factory.php
```diff
@@ -15,7 +24,12 @@ public static function createCategoryDao($username = null) {
 		}
 	}
 
-	public static function createFeedDao($username = null) {
+	/**
+	 * @return FreshRSS_FeedDAOSQLite|FreshRSS_FeedDAO
```
Alkarex:
```suggestion
	 * @return FreshRSS_FeedDAO
```

Returning the top parent class should be sufficient. Same for other lines below

PR #5105, opened 2023-02-13
Add: <meta name="theme-color">

### LINKED ISSUE #5088: [Feature] header keep white
Hi,

when I open the mobile site of FreshRSS  then the header keep white and don’t turn in dark:

AAE41887-C309-443F-B6D0-1FD0C4B86DB5

For example in google everything turns in dark:
FA3D272C-B310-4410-8043-33CDD26C4672

math-GH:
Closes #5088

- via theme's `metadata.json` is it now possible to give a `theme-color`
- each theme has now the color defined in its `metadata.json`

metadata.json
```
{
	"name": "Dark",
	.....
	"theme-color": "#111111"
}
```
`<meta name="theme-color" content="#111111">`
(`"theme-color": {"default": "#111111"}` would do the same)

Origine theme has a CSS color scheme that respects the operating system/browser preference for light/dark themes
```
{
	"name": "Origine",
	.....
	"theme-color": {"dark": "#1f1f1f", "light": "#f0f0f0"}
}
```
`<meta name="theme-color" media="(prefers-color-scheme: light)" content="#f0f0f0">`
`<meta name="theme-color" media="(prefers-color-scheme: dark)" content="#1f1f1f">`

grafik
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

1. use Safari Browser on iPhone (Chrome on Android does not respect the `dark`, but works fine with `light`
2. see the color of address bar and operating system bar

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/FreshRSS.php
```diff
@@ -141,6 +141,10 @@ public static function loadStylesAndScripts(): void {
 						FreshRSS_View::prependStyle(Minz_Url::display(FreshRSS::getThemeFileUrl($theme_id, $filename)));
 				}
 			}
+
+			if (!empty($theme['theme-color'])) {
```
ColonelMoutarde, line 145: opinion: using empty() is an bad pratice 

```
"" (an empty string)
0 (0 as an integer)
0.0 (0 as a float)
"0" (0 as a string)
NULL
FALSE
array() (an empty array)
```
Return false.
they not a real empty. i suggest use `array_keys_exist()` instead of.


more informations: https://dev.to/aleksikauppila/using-isset-and-empty-hurts-your-code-aaa

lib/Minz/View.php
```diff
@@ -237,6 +238,38 @@ public static function appendStyle($url, $media = 'all', $cond = false) {
...
+	 * @return void
+	 */
+	public static function appendThemeColors($themeColors): void {
+		self::$themeColors = $themeColors;
+	}
+
+	/**
+	 * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color
+	 *
+	 * @return String
```
ColonelMoutarde:
```suggestion
```

ColonelMoutarde: you can remove this line, the return value is explicite by typehinting

app/FreshRSS.php
Alkarex: I personally prefer `empty()` here, which is exactly for what we need: check that the value exists and that it is not falsy.
Not tested recently, but `array_key_exists()` is also slower.

ColonelMoutarde: Although the empty() function is useful for checking if a variable is empty, it can have some drawbacks. Here are some arguments against using empty():

Unexpected behavior with certain data types: 

1. The empty() function can produce unexpected results with certain data types, including numerical values, strings that contain digits, strings containing spaces or zeros, etc. For example, empty('0') returns true, which can be confusing.
2. Ambiguity with boolean values: When used with boolean values, the empty() function can produce unexpected results. For example, empty(false) returns true, which can be confusing.
3. Poor code readability: Using empty() can make code less readable, as it requires a double negation to get the opposite result. For example,` if (!empty($var)) { ... }` may be more difficult to understand than` if (isset($var)) { ... }`
4. Limitations for checking array keys: The empty() function cannot be used to check if a specific key exists in an array. For this task, you should use array_key_exists().
5. Risk of modifying the original variable: The empty() function modifies the original variable by converting it to a boolean type, which can cause unintended side effects in your code.

In summary, although the empty() function is useful in some situations, it is important to understand its limitations and consider other options when writing more complex code.

Alkarex: None of that applies there, as far as I can see

lib/Minz/View.php
```diff
@@ -237,6 +238,37 @@ public static function appendStyle($url, $media = 'all', $cond = false) {
 		);
 	}
 
+	/**
+	 * @param array $themeColors
+	 * @return void
```
Alkarex:
```suggestion
```

lib/Minz/View.php
```diff
@@ -237,6 +238,37 @@ public static function appendStyle($url, $media = 'all', $cond = false) {
...
+	 */
+	public static function appendThemeColors($themeColors): void {
+		self::$themeColors = $themeColors;
+	}
+
+	/**
+	 * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color
+	 *
+	 */
+	public static function metaThemeColor(): String {
```
Alkarex:
```suggestion
	 */
	public static function metaThemeColor(): string {
```

lib/Minz/View.php
```diff
@@ -237,6 +238,37 @@ public static function appendStyle($url, $media = 'all', $cond = false) {
...
+		if (!empty(self::$themeColors['light'])) {
+			$meta .= '<meta name="theme-color" media="(prefers-color-scheme: light)" content="' . self::$themeColors['light'] . '" />';
+		}
+		if (!empty(self::$themeColors['dark'])) {
+			$meta .= '<meta name="theme-color" media="(prefers-color-scheme: dark)" content="' . self::$themeColors['dark'] . '" />';
+		}
+		if (!empty(self::$themeColors['default'])) {
+			$meta .= '<meta name="theme-color" content="' . self::$themeColors['default'] . '" />';
+		}
+		if (empty(self::$themeColors['default']) && !empty(self::$themeColors && empty(self::$themeColors['light']) && empty(self::$themeColors['dark']))) {
```
Alkarex:
```suggestion
		if (empty(self::$themeColors['default']) && !empty(self::$themeColors) && empty(self::$themeColors['light']) && empty(self::$themeColors['dark'])) {
```

lib/Minz/View.php
```diff
@@ -237,6 +238,37 @@ public static function appendStyle($url, $media = 'all', $cond = false) {
...
+			$meta .= '<meta name="theme-color" content="' . self::$themeColors . '" />';
```
Alkarex:
```suggestion
			$meta .= '<meta name="theme-color" content="' . htmlspecialchars(self::$themeColors) . '" />';
```

Same for the other lines. Although the input is not supposed to be anything else than a colour string, as themes can be provided by third-parties, we should have some minimum security.

lib/Minz/View.php
Alkarex: You can just click on *Commit suggestion* from the Web UI (or from e.g. the GitHub extension in VSCode) in order to avoid the *Outdated* status and the need to manually resolve conversation. Another benefit is to preserve credits to the suggestion's author

math-GH: I know it and I did it :)

lib/Minz/View.php
math-GH: thanks for the review. Done.

lib/Minz/View.php
Alkarex: Ah, sorry :-)

PR #5106, opened 2023-02-13
Typehint some Controllers

ColonelMoutarde:

- typehinting
- phpstan level 8

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/Controllers/authController.php
```diff
@@ -103,7 +103,9 @@ public function loginAction() {
 	 *   - challenge (default: '')
 	 *   - keep_logged_in (default: false)
 	 *
+	 * @throws Exception
 	 * @todo move unsafe autologin in an extension.
+	 * @throws Exception
```
Alkarex, line 107: Why twice?

app/Controllers/authController.php
```diff
@@ -122,7 +124,7 @@ public function formLoginAction() {
 			$username = Minz_Request::param('username', '');
 			$challenge = Minz_Request::param('challenge', '');
 
-			usleep(rand(100, 10000));	//Primitive mitigation of timing attacks, in μs
+			usleep(random_int(100, 10000));	//Primitive mitigation of timing attacks, in μs
```
Alkarex, line 126: I think the much slower `random_int` is overkill here. We do not need a cryptographically secure value here

app/Controllers/importExportController.php
```diff
@@ -28,11 +28,16 @@ public function firstAction() {
...
 	 */
-	public function indexAction() {
+	public function indexAction(): void {
 		$this->view->feeds = $this->feedDAO->listFeeds();
 		FreshRSS_View::prependTitle(_t('sub.import_export.title') . ' · ');
 	}
 
+	/**
+	 * @param $size_str
+	 * @return float|int
```
Alkarex: We should probably always return float

app/Controllers/importExportController.php
```diff
@@ -256,11 +264,11 @@ private function ttrssXmlToJson(string $xml) {
 	/**
 	 * This method import a JSON-based file (Google Reader format).
 	 *
-	 * @param string $article_file the JSON file content.
-	 * @param boolean $starred true if articles from the file must be starred.
+	 * $article_file the JSON file content.
```
Alkarex, line 268: Why did you remove `@param` ?

app/Controllers/authController.php
ColonelMoutarde: my apologies

app/Controllers/authController.php
ColonelMoutarde: after tests

```php
<?php
$start = microtime(true);
$sum = 0.0;
for ($i = 0; $i < 1000000; $i++) {
    $sum += rand(0, 32767);
}
printf('[rand] Time: %.3f s%s', microtime(true) - $start, PHP_EOL);

$start = microtime(true);
$sum = 0.0;
for ($i = 0; $i < 1000000; $i++) {
    $sum += mt_rand(0, 32767);
}
printf('[mt_rand] Time: %.3f s%s', microtime(true) - $start, PHP_EOL);

$start = microtime(true);
$sum = 0.0;
for ($i = 0; $i < 1000000; $i++) {
    $sum += random_int(0, 32767);
}
printf('[random_int] Time: %.3f s%s', microtime(true) - $start, PHP_EOL);

```
## result in PHP 8.1
```
[rand] Time: 0.044 s
[mt_rand] Time: 0.074 s
[random_int] Time: 0.732 s
```

## result in PHP 7.2
```
[rand] Time: 0.088 s
[mt_rand] Time: 0.071 s
[random_int] Time: 0.641 s
```
All these results are for 1 million iterations which is not our case we use it **only once**. The performances are therefore acceptable and of better quality

app/Controllers/importExportController.php
ColonelMoutarde: after retesting, I confirm that we can have integers and floats as results

app/Controllers/importExportController.php
ColonelMoutarde: As the function is typed and not to mislead the developer in case the documentation is not up-to-date. I could also have strange behaviors with phpstan

app/Controllers/importExportController.php
```diff
@@ -256,11 +262,11 @@ private function ttrssXmlToJson(string $xml) {
...
+	 * true if articles from the file must be starred.
 	 * @return boolean false if an error occurred, true otherwise.
 	 */
-	private function importJson($article_file, $starred = false) {
+	private function importJson(string $article_file, ?bool $starred = false): bool {
```
Alkarex:
```suggestion
	private function importJson(string $article_file, bool $starred = false): bool {
```

I would need to check in more details, but I believe optional parameters are not supposed to have a `?` nullable type. When a `null` is passed, the default boolean value is assigned. So in our function, we always expect a boolean, not a nullable boolean.

app/Controllers/importExportController.php
```diff
@@ -503,9 +509,9 @@ private function importJson($article_file, $starred = false) {
 	 * This method import a JSON-based feed (Google Reader format).
 	 *
 	 * @param array<string,string> $origin represents a feed.
-	 * @return FreshRSS_Feed|null if feed is in database at the end of the process, else null.
+	 * return if feed is in database at the end of the process, else null.
```
Alkarex: I think the `@return` is still needed to assign the proper explanation to the return definition

app/Controllers/tagController.php
```diff
@@ -105,6 +105,9 @@ public function addAction() {
 		Minz_Request::bad(_t('feedback.tag.name_exists', $name), ['c' => 'tag', 'a' => 'index']);
 	}
 
+	/**
+	 * @return void|null
```
Alkarex, line 116: Where does the `null` come from?

app/Controllers/importExportController.php
ColonelMoutarde: the return is clear, i use "?" to inform the result is nullable.

Alkarex: What I mean is that there is an error in the PHPDocs syntax, preventing attaching **the description** of the result correctly.

Before your changes (correct):
image

After your changes (wrong):
image

app/Controllers/importExportController.php
ColonelMoutarde: done

app/Controllers/tagController.php
ColonelMoutarde: done

app/Controllers/importExportController.php
ColonelMoutarde: done

Alkarex:
Good start, but this not level 9 (517 errors), and not even level 6 (87 errors).

Example:

```console
$ vendor/bin/phpstan analyse --level 6 app/Controllers/
...
 ------ ------------------------------------------------------------------------------ 
  Line   tagController.php                                                             
 ------ ------------------------------------------------------------------------------ 
  :108   Method FreshRSS_tag_Controller::renameAction() has no return type specified.  
 ------ ------------------------------------------------------------------------------ 
...
[ERROR] Found 87 errors
```

Could you try to pass at least level 6 for the files you are addressing?

I am wondering how did you test that this passed level 9? Maybe a wrong command?

ColonelMoutarde:
> Good start, but this not level 9 (517 errors), and not even level 6 (87 errors).
> 
> Example:
> 
> ```
> $ vendor/bin/phpstan analyse --level 6 app/Controllers/
> ...
>  ------ ------------------------------------------------------------------------------ 
>   Line   tagController.php                                                             
>  ------ ------------------------------------------------------------------------------ 
>   :108   Method FreshRSS_tag_Controller::renameAction() has no return type specified.  
>  ------ ------------------------------------------------------------------------------ 
> ...
> [ERROR] Found 87 errors
> ```
> 
> Could you try to pass at least level 6 for the files you are addressing?
> 
> I am wondering how did you test that this passed level 9? Maybe a wrong command?

Hi, 
the files that you mention as being in error, are files that I have not touched. I have not done all the files in the controller folder

Alkarex:
Check again: the example I gave, `tagController.php`, **is** one you have touched, and it does not pass level 6 ;-)

image

Fair enough if you have not addressed all the controllers, but in this case, the PR description is inaccurate and needs to be corrected...

ColonelMoutarde:
Up to level 8

Alkarex:
Ok, looks good now. The command to check is something like:

```console
$ vendor/bin/phpstan analyse --level 6 app/Controllers/apiController.php app/Controllers/authController.php app/Controllers/categoryController.php app/Controllers/configureController.php app/Controllers/entryController.php app/Controllers/errorController.php app/Controllers/importExportController.php app/Controllers/javascriptController.php app/Controllers/statsController.php app/Controllers/subscriptionController.php app/Controllers/tagController.php 
Note: Using configuration file /home/alexandre/GitHub/FreshRSS/phpstan.neon.
 11/11 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] No errors
```

PR #5108, opened 2023-02-14
Typehint to greader.php file

ColonelMoutarde:

- type hint the code and cleaning

- [X] clear commit messages
- [X] code manually tested

Pass PhpStan level 7


p/api/greader.php
```diff
@@ -104,17 +108,15 @@ function debugInfo(): string {
 
 final class GReaderAPI {
 
-	/** @return never */
-	private static function badRequest() {
+	private static function badRequest(): void {
```
aledeg: I think this is not the intend. We should use `never` as a type-hint but it is not available until PHP 8.1. The same applies for other methods ending with a `die`. Maybe this applies also for methods ending with an `exit`.

ColonelMoutarde: I understand. I can't wait for the release of the new version of debian, we will be able to type for php 7.4 at least

ColonelMoutarde: done

aledeg (approved):

p/api/greader.php
```diff
@@ -30,17 +30,21 @@
 
 if (PHP_INT_SIZE < 8) {	//32-bit
 	function hex2dec(string $hex): string {
-		if (!ctype_xdigit($hex)) return '0';
+		if (!ctype_xdigit($hex)) {
```
Alkarex: Although I do not mind the proposed syntax, please avoid minor stylistic code changes, especially when not related to the PR.

p/api/greader.php
```diff
@@ -425,7 +429,7 @@ private static function subscriptionEdit(array $streamNames, array $titles, stri
 					$feed = $feedDAO->searchByUrl($streamUrl);
 					$feedId = $feed == null ? -1 : $feed->id();
 				}
-				$title = isset($titles[$i]) ? $titles[$i] : '';
+				$title = $titles[$i] ?? '';
```
Alkarex, line 430: Same here: I like the new syntax (which is [exactly equivalent to the former one](https://php.net/language.operators.comparison#language.operators.comparison.coalesce)), but I would rather not change it in random places, especially when not related to the PR

p/api/greader.php
```diff
@@ -467,7 +471,7 @@ private static function subscriptionEdit(array $streamNames, array $titles, stri
 	private static function quickadd(string $url) {
 		try {
 			$url = htmlspecialchars($url, ENT_COMPAT, 'UTF-8');
-			if (substr($url, 0, 5) === 'feed/') {
+			if (0 === strpos($url, 'feed/')) {
```
Alkarex: That is also not related to this PR, but if anything, the new [`str_starts_with()`](https://php.net/function.str-starts-with) should be used, with a polyfill in `lib_rss.php`.

p/api/greader.php
```diff
@@ -668,9 +672,6 @@ private static function streamContents(string $path, string $include_target, int
 		header('Content-Type: application/json; charset=UTF-8');
 
 		switch ($path) {
-			case 'reading-list':
```
Alkarex, line 671: Why did you remove this section? If anything, it can be moved together with default

p/api/greader.php
Alkarex: No need to change it back though

p/api/greader.php
Alkarex: No need to change it back though

p/api/greader.php
ColonelMoutarde: By deleting it, it is automatically directed to the default

p/api/greader.php
ColonelMoutarde: done

p/api/greader.php
Alkarex: Yes I know, but by removing it, you lose the information about what you are supposed to support

Frenzie: Right now it's clear that it's supposed to do that. Losing that clarity is not a good thing. :-)

p/api/greader.php
ColonelMoutarde: Done

p/api/greader.php
Alkarex: > If anything, it can be moved together with default

Addressed in https://github.com/FreshRSS/FreshRSS/pull/5108/commits/6e330aec62107297b192e0fe65a86276a50cce88

PR #5140, opened 2023-02-23
Corrected `frss:xPathItemTitle` definition

### LINKED ISSUE #5133: Is the documentation for `frss:xPathItemTitle` correct?
In [the FreshRSS OPML extension documentation](https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/developers/OPML.md), the `frss:xPathItemTitle` attribute is specified as being the "XPath expression for extracting the feed title from the source page." Was "source page" meant to be "item context" like most of the other XPath attributes defined there?  
The example given for `frss:xPathItemTitle`'s value is `descendant::h2`, which, if it really were relative to the source page, would return *every* `<h2>` element on the source page, which does not seem like the most helpful example.

mincerafter42:
Closes #5133.

### DIFF
--- a/docs/en/developers/OPML.md
+++ b/docs/en/developers/OPML.md
@@ -29,7 +29,7 @@ The following attributes are using similar naming conventions than [RSS-Bridge](
 
 * `frss:xPathItem`: XPath expression for extracting the feed items from the source page.
 	* Example: `//div[@class="news-item"]`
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

Alkarex:
Thanks! You are welcome to put a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #5141, opened 2023-02-23
Call to an undefined method SimplePie_Cache_MySQL::execute().

ColonelMoutarde:

- Call to an undefined method SimplePie_Cache_MySQL::execute().

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Undefined how? I do not believe we are using this code. If anything, please check instead whether anything should be fixed in the upstream SimplePie library.

ColonelMoutarde:
> Undefined how? I do not believe we are using this code. If anything, please check instead whether anything should be fixed in the upstream SimplePie library.

Indeed, the code in simplePie has been corrected in the same way as what I proposed in my presentation. I think it is necessary to update simplePie

Alkarex:
Yes. Regarding SimplePie, there has been a big code restyling upstream leading to a big mess so some update work is needed. It has been started https://github.com/FreshRSS/FreshRSS/pull/4374#issuecomment-1239278052 but with low priority

PR #5144, opened 2023-02-23
defined('CLEANCACHE_HOURS')

### LINKED ISSUE #4627: [BUG] ./data/cache with 6700+ files 2+GB
While I backed up my server I see that the /data/cache folder has more than 6700 files that need more than 2GB storage.

The files are created between August 20nd 2022 and September 19th 2022 (today).

Biggest files:
grafik

Most files are `html` files, that are a copy from websites for CSS selector that retrieves truncated RSS feeds.
And `spc` files that are JSON files.

Does it makes sense to keep this cache files for 1 month?
Would it be possible to use less storage?

- FreshRSS version: 1.20.0-dev (before updating to 1.20.0)

math-GH:
Closes #4627

now it is possible to config the cache via `./constants.local.php`
```
<?php
	define('CLEANCACHE_HOURS', 240);
```

default: `720` hours

- introduce the constant `CLEANCACHE_HOURS`

1. check the content of the folder `./data/cache`
2. create `./constants.local.php` and set the hours
3. go to the frontend and reload the page (it would help to comment out the condition `if (mt_rand(0, 30) === 1)` in `lib/lib_rss.php`)
4. old cache files are deleted in `./data/cache`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/constants.php
+++ b/constants.php
@@ -57,3 +57,6 @@ defined('EXTENSIONS_PATH') or define('EXTENSIONS_PATH', FRESHRSS_PATH . '/extens
 
 //Directory used for feed mutex with *.freshrss.lock files. Must be writable.
 defined('TMP_PATH') or define('TMP_PATH', sys_get_temp_dir());
+
+//clean the chacke after x hours (720 hours = 30 days)
+defined('CLEANCACHE_HOURS') or define('CLEANCACHE_HOURS', 720);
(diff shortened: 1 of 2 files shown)


ColonelMoutarde:
hello, Why a constant and not make it customizable?

Alkarex:
> hello, Why a constant and not make it customizable?

While user options are good, having too many gives some clutter and increases complexity, so that should be balanced with the benefits. In this specific case, I think a constant would be a good start, while we think about better default values and behaviours.

ColonelMoutarde (approved):

PR #5147, opened 2023-02-26
Improve UI/UX install process

### LINKED ISSUE #3961: [Improvement] install process: less buttons on first step to chose the language
The install routine starts with a language selection:
grafik

When already a language is chosen (f.e. when you skip back from step 3 to step 1) than a third button appears (marked with (1)). 

This Button (1) is not intuitive and imho useless. It is not a real form button. It is (technically) just a link to the next step without (!) saving the form above. But it looks like a "save and go forward" button.
I would suggest to remove this button. I prefer to use the "submit" button to confirm the selected language.

Also the "cancel" button is imho useless. It is just a reset button and resets the select to the previous language. If the user did not changed the language than nothing will be changed. It does not cancel the installation routine.

I would prefer the smart one-button start of the install routine.

math-GH:
# General
## Show password button
grafik

Before: keep the button pressed to see the password
After: same behaviour like everywhere: press the button and the password will be shown for 5 seconds.

## make the menu for mobile accessable
Before: No button. No access to the hidden menu

After: 
grafik

## Favicon + header icon
Before: 
No favicon, header icon in the middle
grafik

After:
Favicon. And header icon is on the left side as default
grafik

# Navigation
## Before
deactivated steps has not padding to the left border:
grafik

## After
grafik

# Step 1

## Before:
First step does not need so much buttons for a better UX (Closes #3961)
grafik

## After:
grafik

## Before:
If user steps back in process:
grafik

## After:
instead of a button: a text link (UX: a button saves the form, a link does not save the form)
grafik

# Step 2
## Before
no grey background behind the buttons
grafik

## After
grafik

## Before
in mobile view: no gap between the buttons
grafik

## After
grafik

## Before
No sub-headlines
grafik

## After
similar to "installation check" in admin config
grafik

# Step 4
## Before
grafik

## After
grafik

# Step 5
## Before
grafik

## After
grafik

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


p/scripts/install.js
```diff
@@ -1,26 +1,45 @@
...
-	const button = ev.currentTarget;
-	const passwordField = document.getElementById(button.getAttribute('data-toggle'));
+let timeoutHide;
+
+function showPW_this(ev) {
+	const id_passwordField = this.getAttribute('data-toggle');
+	if (this.classList.contains('active')) {
+		hidePW(id_passwordField);
+	} else {
+		if (ev.type === 'click' || ev.buttons || ev.key === ' ' || ev.key.toUpperCase() === 'ENTER') {
```
Frenzie: The values of `ev.key` here are standardized btw, so it's always `Enter` on all browsers.

https://www.w3.org/TR/uievents-key/#keys-whitespace

Frenzie (approved):
lgtm on the screenshots, hardly tested

p/scripts/install.js
math-GH: It is nice to know. I did just copy&paste from `extra.js`. I would keep it as it is right now.
With a refactoring, we could find a better solution as duplicated source code

Frenzie: Looking at it a bit closer, what's the purpose of this check?

It'll only ever be called from a click event:
https://github.com/FreshRSS/FreshRSS/blob/c170c390b7d51f809b3e0064bf883221037397a3/p/scripts/extra.js#L118

So this would show the same behavior:
```diff
diff --git a/p/scripts/extra.js b/p/scripts/extra.js
index c068a998..be26242c 100644
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -84,14 +84,12 @@ function init_crypto_form() {
 // <show password>
 let timeoutHide;
 
-function showPW_this(ev) {
+function showPW_this() {
        const id_passwordField = this.getAttribute('data-toggle');
        if (this.classList.contains('active')) {
                hidePW(id_passwordField);
        } else {
-               if (ev.type === 'click' || ev.buttons || ev.key === ' ' || ev.key.toUpperCase() === 'ENTER') {
-                       showPW(id_passwordField);
-               }
+               showPW(id_passwordField);
        }
        return false;
 }
```

math-GH: the user could tab from the password input to the show-pw-button and then press `enter` or `space`

Frenzie: If that didn't generate a click event, this code would never be reached. ;-) That would only be relevant in combination with a `btn.addEventListener('keydown', showPW_this);` (or key-anything), which to be clear is not necessary because pressing space or enter on a button or a link generates a click event.

ColonelMoutarde (approved):

p/scripts/install.js
math-GH: you are right. Now it is improved. Thanks!

Issue #1077, opened 2016-02-16
Bug with srcset images


Alkarex:
Images with `srcset` are not displayed. E.g. from Numerama:

``` html
<img src="../themes/icons/grey.gif" data-original="http://www.numerama.com/content/uploads/2016/02/amitie-1024x576.jpg" alt="Amitié" srcset="http://www.numerama.com/content/uploads/2016/02/amitie-680x383.jpg 680w, http://www.numerama.com/content/uploads/2016/02/amitie-1024x576.jpg 1024w, http://www.numerama.com/content/uploads/2016/02/amitie-270x152.jpg 270w" sizes="(max-width: 1024px) 100vw, 1024px" lazyload="" postpone="" height="576" width="1024">
```

Alkarex:
While waiting for a better approach https://github.com/FreshRSS/FreshRSS/issues/1080

Alkarex:
As an improvement, consider implementing a `data-originalset` to keep the original `srcset` instead of removing it as done in https://github.com/FreshRSS/FreshRSS/issues/1080

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/1081#issuecomment-186192835

Alkarex:
This is good enough for the time being.

PR #5154, opened 2023-03-02
fix: scrset in <picture><source>

### LINKED ISSUE #5016: [BUG] Fetching article by CSS-selectors: <picture> element is wrong stripped
**Describe the bug**
I subscribed to https://www.spiegel.de/schlagzeilen/index.rss

CSS selectors: `#Inhalt section.relative`
grafik

The website: have a look into the `<picture><source>` params with focus on the mandatory `srcset`
grafik

After the SimplePie HTML stripping: the `srcset` param in `<source>` is lost
grafik

Why is it important:
The `<picture>` give the opportunity to load better compressed pictures given in `<source>` (here: webp file). The `` is just the fallback with default files (here: jpg file)

**Expected behavior**
`<source>` has the `srcset` param

math-GH:
Closes #5016

reverts #1080 (Ref #1077)

`srcset` is required in `<source>` as sub tag of `<picture>`, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-srcset

- delete `srcset `from `strip_attributes`

see #5016

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Frenzie:
https://github.com/FreshRSS/FreshRSS/issues/1077#issuecomment-185662754 suggests to use a `data-originalset`. Or perhaps more to the point, are we sure this doesn't reintroduce the original issue? :-)

math-GH:
> [#1077 (comment)](https://github.com/FreshRSS/FreshRSS/issues/1077#issuecomment-185662754) suggests to use a `data-originalset`. Or perhaps more to the point, are we sure this doesn't reintroduce the original issue? :-)

The corresponding PR is #1080. It added the srcset to the strip_attributes. This PR here will revert #1080.
IMHO I do not see any benefit of #1080 (`srcset` is required in `<source>` as sub tag of `<picture>`, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-srcset)

(will add this information to the PR main info)

Frenzie:
I agree in principle, but I imagine so does @Alkarex. The exact contours of the problem it's supposed to work around aren't completely clear to me at the moment, is all.

Frenzie (approved):

ColonelMoutarde (approved):

Alkarex:
This PR would re-introduce the bug described in https://github.com/FreshRSS/FreshRSS/issues/1077
But in the meantime, the problematic feed changed structure https://www.numerama.com/feed/
Note that the problem was the use of `srcset` in an `img` tag, not in a `source` tag.
Ideally, we should:

* extend the SimplePie `strip_attributes()` method on the model of `add_attributes()` (which I authored :-)) to be able to remove attributes only for some specific tags.
* Better address the case of `srcset` attributes of `img` elements (or strip them) (test welcome)

Frenzie:
But what precisely *is* the bug described there? `srcset` on `img` is desirable tout court.

Alkarex:
> But what precisely _is_ the bug described there? `srcset` on `img` is desirable tout court.

7 years ago, so I cannot remember in details, but some images were not displayed at all, probably due to our custom lazy-load logic. We should at least test with some examples of `` to see how it behaves.

Frenzie:
Speaking of lazy loading, since ca. 2019/2020 `loading="lazy"` has been available natively, so now also in ESR and the like. It may be possible to replace most or all custom JS logic with that simple attribute.
<https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading>

Alkarex:
Indeed Frenzie, although I would first be interested in a few tests to check whether there is any significant difference

math-GH:
What I did not understand: If `` and "srcset" has the value for lazy loading to "src" and FreshRSS deletes "srcset" what will happen? "src" is still unchanged and has the "before ladzy loading image".
So I do not see the benefit to remove the `srcset` yet

Alkarex:
We need to fix lazyload of srcset, ideally before merging this PR, as this PR otherwise risks loading many unwanted pictures, which may be problematic on slow and/or metered connections.

math-GH:
Current state (before this PR)

`` => `` shows `url1.jpg`

after this PR:

`` => `` shows `url1.jpg`

Do I oversee something?

Alkarex:
> Do I oversee something?

Yes, the relevant test is when lazy loading is used, meaning that **no image at all** should be loaded:

```
<img src="../themes/icons/grey.gif" data-original="https://example.net/large.jpg" />
```

P.S.: This can be seen in the DOM of articles that are still collapsed

Frenzie:
Presumably it'd be a fairly minor effort to add an additional data-srcset or some such.
https://github.com/FreshRSS/FreshRSS/blob/1c1e63c6ad27d0ebd34324b09f176b6767adc0cf/p/scripts/main.js#L402-L406

data-original is a bit vague btw, data-original-src or just data-src seems like it'd be more descriptive? :-)

Alkarex:
@Frenzie Yes, but before embarking on more custom coding, I think it would be worth making a few tests with the standard `lazyload` to check whether it would be sufficient (if yes, we could completely remove our own code)

Frenzie:
Well yeah, that's exactly what I said before. :-)

PR #5180, opened 2023-03-08
[Docker] Listening for IPv6 connections

### LINKED ISSUE #5178: [Feature] [Docker] Listening for IPv6 connections
I noticed that the FreshRSS web server in Docker does not accept incoming connections via IPv6 by default.
This is because the web server currently only binds to `0.0.0.0:80`:
https://github.com/FreshRSS/FreshRSS/blob/1c502aaac2d53ca652e21c4e3ccd47362812626b/Docker/FreshRSS.Apache.conf#L2
I suggest adding the `Listen [::]:80` directive by default, so that the web server accepts both IPv4 and IPv6 connections.
(sorry if there is already this issue here, but I couldn't find it)

Rufubi:
Closes #5178

- Added `Listen [::]:80` directive in Apache configuration
- Added `LISTEN6` environment variable

1. Enable IPv6 in Docker
2. Create container (with or without `LISTEN` and `LISTEN6`)
3. Try to connect via IPv6 (with `curl`, for example)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


Alkarex:
Another approach to listen on all interfaces is to use the default `Listen 80`.
But the original use of `0.0.0.0:80` was to address IPv6 problems https://github.com/FreshRSS/FreshRSS/pull/1813
So I am worried of the default `Listen [::]:80`: we need to test whether that works fine in multiple situations, including Docker configurations without IPv6 enabled, and hosts without IPv6 enabled.
Furthermore, due to the current test of `LISTEN6`, it does not seem possible to disable it, right?

Docker/FreshRSS.Apache.conf
```diff
@@ -1,5 +1,6 @@
 ServerName freshrss.localhost
 Listen 0.0.0.0:80
+Listen [::]:80
```
Alkarex: Any advantage compared to a single `Listen 80` (which is the default Apache value)?
I am a bit worried about the risks of regressions by providing this IPv6-specific line by default, e.g. for platforms without IPv6. At least we would need some tests to see how it behaves

Frenzie: > which is the default Apache value

I thought Apache without a Listen directive won't start at all?

Alkarex: > I thought Apache without a Listen directive won't start at all?

Apache does have a default Listen directive by default: `Listen 80`

Rufubi: The single `Listen 80` only accepts IPv4 connections. I couldn't find any other way except to make a docker volume for the config and edit it manually to make Apache accept connections via IPv6.
I don't know how Apache will behave on machines without IPv6 support. I think I could either make a separate variable like `ENABLE_IPV6` or add `Listen` directive only when `LISTEN6` is set

Alkarex: > The single `Listen 80` only accepts IPv4 connections

Could you please double-check that?

 `Listen 80` works for me both for our Debian and our Alpine images, that is to say with IPv4 + IPv6. This is also what the documentation says https://httpd.apache.org/docs/current/mod/mpm_common.html#listen

```sh
docker run --rm -e 'LISTEN=8080' --network host -v freshrss_data:/var/www/FreshRSS/data freshrss/freshrss:edge
# or
docker run --rm -e 'LISTEN=8080' --network host -v freshrss_data:/var/www/FreshRSS/data freshrss/freshrss:edge-alpine

curl -v 'http://127.0.0.1:8080/'
curl -v 'http://[::1]:8080/'
```

Rufubi: > Could you please double-check that?

Oh, my bad. The configuration was `Listen 0.0.0.0:80` and I didn't even try to do `Listen 80`. Yes, `Listen 0.0.0.0:80` only accepts IPv4, but `Listen 80` accepts both IPv4 and IPv6.
It's much simpler this way. I will make changes to leave the ability to bind to specific IPv6 addresses

Frenzie: On an ancient Debian I can confirm that Listen 80 doesn't do that. On some version you should actually be using it works though. :-P
```
apache2/oldoldstable,now 2.4.25-3+deb9u13 amd64 [installed,automatic]
  Apache HTTP Server
```

```
/etc/apache2# grep -R Listen
ports.conf:Listen 80
ports.conf:     Listen 443
ports.conf:     Listen 443
```

```
curl -v 'http://127.0.0.1:80/'
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.52.1
> Accept: */*
> 
< HTTP/1.1 200 OK
```

```
curl -v 'http://[::1]:80/'
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 80 failed: Connection refused
* Failed to connect to ::1 port 80: Connection refused
* Closing connection 0
curl: (7) Failed to connect to ::1 port 80: Connection refused
```

vs

```
ping ::1
PING ::1(::1) 56 data bytes
64 bytes from ::1: icmp_seq=1 ttl=64 time=0.042 ms
64 bytes from ::1: icmp_seq=2 ttl=64 time=0.054 ms
64 bytes from ::1: icmp_seq=3 ttl=64 time=0.050 ms
```

Rufubi: Hmm. What do you think about making the LISTEN variable a comma-separated list of addresses/ports? This would allow you not to bind to just one address and allow you to bind to multiple addresses at once. But I think that would be a bit harder to implement

Alkarex: As this is solely for our Docker images, we control Apache versions.
One thing at a time: could you please try again with the `Listen 80` approach? (Check my example above)

Alkarex: Ah sorry, I did not see your answer...

Alkarex: Given that `Listen 80` works, and with our existing environment variable, we already cover the ability to listen on IPv4, IPv6, or both, which I believe is sufficient for the context of being used in a container.
But we should at least update the documentation to explain how to do that.
We could also test some more to consider using `Listen 80` by default

Rufubi:
Finally I realized my idea. I made the `LISTEN` variable a comma-separated list of values. It should work with unset `LISTEN` as well as with one or more values. I used awk instead of sed and it worked, I checked it with Debian and Alpine images.
For busybox-awk there was no parameter for in place changes, so I had to do a loop of several commands. Inspired by examples from here: https://www.shellcheck.net/wiki/SC2044

Alkarex:
Thanks for the efforts @Rufubi , but do we have any significant use case for this feature (remembering that this is for a container, supposed to be behind a reverse proxy when in production)? I would rather avoid introducing unneeded complexity, which incurs more maintenance, documentation, testing efforts, potential regressions in our code that needs to detect FreshRSS address, etc.

Also, the YAML approach to that would be something like:

```yml
LISTEN
  - 0.0.0.0:80
  - [::]:80
```

So I would rather keep your code for the record here in the case an actual significant use case appears, and instead, as I was writing above, just update the documentation for now to explain how to enable IPv4, IPv6, or both.

It would also be good to investigate enabling both IPv4 and IPv6 by default with `Listen 80` (a few tests would be needed)

Alkarex:
For the record, [`--network host`](https://docs.docker.com/network/host/) is needed to test the IPv4-only or IPv6-only, because in normal Docker bridge network mode, when publishing ports, both stacks seem to be always mapped (Docker receives the IPv6 query and uses IPv4 routing to the container):

```
PORTS
0.0.0.0:8080->80/tcp, :::8080->80/tcp
```

I have updated my test examples a bit higher up.

Rufubi:
1. There is no need to map ports in IPv6 because port mapping is literally NAT. Does Docker do port mapping (DNAT) for IPv6? I just disabled "iptables" and do port mapping manually and don't know how Docker behaves.
2. Is there any reason to keep the container behind a reverse proxy (in IPv6)? I use reverse proxy for IPv4 connections because I only have 1 IPv4 address and I have other web services. But as for IPv6, I have a /56 prefix and I can assign each Docker bridge a subnet. I don't have the ability to do load balancing, and as for access restrictions, I have a firewall.
Anyway, my changes don't change the standard FreshRSS/Apache behavior (except that IPv6 connections are now possible). As for IPv6 disabled and `Listen 80`, I think Apache can handle that situation.
3. I tried to make a compose variable like in your YAML example, but it didn't work. I don't think Docker supports this.
4. This is not so much a question about IPv6 as it is about listening to connections in general. I don't know if anyone needs to listen on more than one port (in general, not just IPv6). But why not?

Anyway, if you accept at least changing `Listen 0.0.0.0:80` to `Listen 80` in Apache configuration, that's enough for me personally.

math-GH:
@Rufubi 
Please have a look into this warning from the CICD pipeline

```
Warning: Docker/entrypoint.sh does not abide by coding style, diff for expected style:
11,14c11,13
< 	find /etc/apache2/ -type f -name FreshRSS.Apache.conf > files.tmp
< 	while IFS= read -r file
< 	do
< 		awk '{ split(ENVIRON["LISTEN"], list, ","); if ($1 ~ "Listen") for (i in list) print "Listen " list[i]; else print $0 }' "$file" > "$file".tmp
---
> 	find /etc/apache2/ -type f -name FreshRSS.Apache.conf >files.tmp
> 	while IFS= read -r file; do
> 		awk '{ split(ENVIRON["LISTEN"], list, ","); if ($1 ~ "Listen") for (i in list) print "Listen " list[i]; else print $0 }' "$file" >"$file".tmp
16c15
< 	done < files.tmp
---
> 	done <files.tmp
```

Alkarex:
> Please have a look into this warning from the CICD pipeline

I do not think we will merge this change for now. There does not seem to be any use-case justifying it.
We can already address the need to listen in *[host networking](https://docs.docker.com/network/host/)* mode to whatever interface, including IPv4+IPv6, thanks to the existing `LISTEN` environment variable.

The remaining question is whether we should change our default to Listen 80, which might be OK. But we should check whether it does not introduce any other problem. Note that this only changes anything for the *[host networking](https://docs.docker.com/network/host/)*, which is not the most usual way to run Docker, as it in particular removes all isolation and does not scale to multiple instances.

Alkarex:
> Anyway, if you accept at least changing `Listen 0.0.0.0:80` to `Listen 80` in Apache configuration, that's enough for me personally.

@Rufubi Let's go for that

Alkarex:
Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md :-)

PR #5185, opened 2023-03-09
Fever API fix mark-all-as-read

### LINKED ISSUE #5183: [BUG] fever does not work as intended
**Describe the bug**
I use Unread my feeds.
Since the last majorupdate the function mark all as read does not work anymore
On the webserver side the command is acknowledged with a HTTP code 200 and does not throw an error.
As soon as I do a refresh the unread items are present again.

**To Reproduce**
1. Go to a Folder with any feeds and unread articles in it
2. Mark the all as read
3. Refresh feeds again on the client
4. the unread items/acticles are here again

**Expected behavior**
Arcticles should be marked as read and do not come back as unread arcticles after refresh.

- Device: iPhone OS 16.3.1 / iPad OS 16.3.1
- OS: Debian 11 Bullseye
- Browser: Not needed
- FreshRSS version: 1.21.0
- Database version: SQLite
- PHP version: PHP 8.0.28
- Installation type: manualy

Bildschirm­foto 2023-03-09 um 15 12 03

**Additional context**
After copying the old fever.php from the last stable Release into /var/www/html/freshrss/p/api it works as intended again.

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/5183

And set protected methods as private since there is no subclass, and in order to get warnings for unused methods, which would have spotted the bug.

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -65,7 +65,7 @@ final class FeverDAO extends Minz_ModelPdo
 	 * @param array<string|int> $values
 	 * @param array<string,string|int> $bindArray
 	 */
-	protected function bindParamArray(string $prefix, array $values, array &$bindArray): string {
+	private function bindParamArray(string $prefix, array $values, array &$bindArray): string {
 		$str = '';
 		for ($i = 0; $i < count($values); $i++) {
 			$str .= ':' . $prefix . $i . ',';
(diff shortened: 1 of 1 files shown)


p/api/fever.php
```diff
@@ -256,7 +256,7 @@ public function process(): array {
 				case 'group':
 					switch ($_REQUEST['as']) {
 						case 'read':
-							$this->setFeedAsRead($id, $before);
+							$this->setGroupAsRead($id, $before);
```
Alkarex, line 259: This is the fix of the actual bug

PR #5212, opened 2023-03-18
Remove FreshRSS_Searchable for better types

Alkarex:
The interface was not used, and it was preventing more precise types for the different `searchById()` methods, as they each have different input and output types.


ColonelMoutarde:
Suggestion : The interface allows making sure that the method ::searchById($id) is well implemented, it has a real utility. 
It allows among other things to make test like SomeClass instanceof FreshRSS_Searchable.
 I think that this Pr risks to bring more verbose code later. 
It is better to make evolve the interface

Alkarex:
I would like to point you to the  [YAGNI (*You aren't gonna need it*)](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle

> it has a real utility

It is not used at all at the moment, so it does not make sure of anything, and therefore has no utility. Plus, it is not even the same types for all implementations, so it is not a good case for an interface.

> more verbose code later

It is having this interface that makes the code more verbose *right now*.

Here is a bit of background https://betterprogramming.pub/avoiding-premature-software-abstractions-8ba2e990930a
(I had another article in mind, more focussed on this case, but cannot find it at the moment...)

Alkarex:
Let's move on. It is easy to add again, should it ever become relevant, and in the meantime, it stands in the way

ColonelMoutarde:
> Let's move on. It is easy to add again, should it ever become relevant, and in the meantime, it stands in the way

You’r right ☺️

PR #5213, opened 2023-03-19
Consistent entry ID type (32-bit compatibility)

Alkarex:
Entry IDs (which are 64-bit integers) must be processed as string to be compatible with 32-bit platforms.
In some previous *type hints* PR, this was a bit messed with.
(NB: Category / Feed / Tag IDs are 32-bit integers, so they can be passed as `int` without problem)

https://github.com/FreshRSS/FreshRSS/pull/5212 must be merged first.

P.S. The last commits add [PHPStan level 6](https://github.com/FreshRSS/FreshRSS/issues/4112) for affected files

```sh
vendor/bin/phpstan analyse --level 6 app/Models/EntryDAO.php app/Models/EntryDAOPGSQL.php app/Models/EntryDAOSQLite.php
```


p/api/fever.php
```diff
@@ -420,40 +420,40 @@ private function entriesToIdList(array $ids = array()): string {
...
-		$entries = $this->entryDAO->listIdsWhere('a', '', FreshRSS_Entry::STATE_FAVORITE, 'ASC', 0) ?: [];
+		$entries = $this->entryDAO->listIdsWhere('a', 0, FreshRSS_Entry::STATE_FAVORITE, 'ASC', 0) ?: [];
 		return $this->entriesToIdList($entries);
 	}
 
 	/**
 	 * @return integer|false
 	 */
-	private function setItemAsRead(int $id) {
+	private function setItemAsRead(string $id) {
```
Alkarex, line 435: Example, which was breaking 32-bit compatibility

app/Models/EntryDAO.php
```diff
@@ -303,7 +307,7 @@ public function updateEntry(array $valuesTmp) {
 	 * @todo simplify the query by removing the str_repeat. I am pretty sure
 	 * there is an other way to do that.
 	 *
-	 * @param integer|array $ids
+	 * @param string|array<string> $ids
```
Alkarex, line 310: Example of wrong type hint, breaking 32-bit compatibility

ColonelMoutarde (approved):

PR #5214, opened 2023-03-19
FeverAPI fix enclosures

### LINKED ISSUE #5209: [BUG] Images in enclosure in the Fever API
**Describe the bug**
The `html` field of articles in the Fever API had image enclosure attached to the article content, for feeds that had images as enclosures, but not in the content.

This is no longer the case for articles fetched after upgrading to 1.21.0, when syncing with the Fever API, but still works when syncing with the Google Reader API.

**To Reproduce**
- Refresh the feed on 1.21.0
- Sync with Fever API

**Expected behavior**
The article content should be the same in the FreshRSS and Google Reader APIs.

- Example feed: https://www.digi24.ro/rss
- FreshRSS version: 1.21.0

**Additional context**
Fiery Feeds dev here - I'm going to switch to the Google Reader API soonish, if that is relevant for prioritisation :)

Alkarex:
Fix regression: provide enclosures as part of HTML content body in FeverAPI
#fix https://github.com/FreshRSS/FreshRSS/issues/5209

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -187,6 +187,7 @@ HTML;
 			$thumbnails = $enclosure['thumbnails'] ?? [];
 			$etitle = $enclosure['title'] ?? '';
 
+			$content .= "\n";
 			$content .= '<figure class="enclosure">';
 
 			foreach ($thumbnails as $thumbnail) {
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -86,7 +86,7 @@ final class FeverDAO extends Minz_ModelPdo
 
 		$sql = 'SELECT id, guid, title, author, '
 			. ($entryDAO::isCompressed() ? 'UNCOMPRESS(content_bin) AS content' : 'content')
-			. ', link, date, is_read, is_favorite, id_feed '
+			. ', link, date, is_read, is_favorite, id_feed, attributes '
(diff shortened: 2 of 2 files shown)


Frenzie (approved):
Looks simple, but I don't have Fever things at hand to check.

Alkarex:
Can be tested with cURL:

```sh
api_key=`echo -n "alice:secretPassword" | md5sum | cut -d' ' -f1`
curl -s -F "api_key=$api_key" 'https://freshrss.example.net/api/fever.php?api&items&with_ids=1679231293688643' | jq .
```

app/Models/Entry.php
```diff
@@ -187,6 +187,7 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
 			$thumbnails = $enclosure['thumbnails'] ?? [];
 			$etitle = $enclosure['title'] ?? '';
 
+			$content .= "\n";
```
math-GH, line 190: @Alkarex  How is it important for the Fever API?

Alkarex: Not much, but we are concatenating a `<figure ...>...` HTML code right after, and that made the output code difficult to read. Furthermore, I cannot remember for sure, but I believe that having no white-space at all can sometimes lead to a different rendering for some styling rules.

PR #5215, opened 2023-03-19
Improved: show Terms of Service in config menu

math-GH:
Ref #5189

Changes proposed in this pull request:

Before:
logged in users cannot read the Terms of Service, if enabled

After:
if TOS (Terms of Service) is set, it is readable in config menu
grafik

Before:
No information about TOS for admins

After:
Information about TOS given in System Configuration.
TOS not given:
grafik

TOS given:
grafik

Technical changes: 
- introduced `TOS_FILENAME` as constant, that points to the `tos.html`
- CSS: moved some standard CSS lines from themes into `frss.css`
- `tos.html`/`tos.example.html`: `<h1>` headline is not needed anymore in the html file. It is now included by default on the page, that displays the TOS

How to test the feature manually:

1. enable/disable the TOS
2. look into the config menu
3. read the TOS

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

Issue #5189, opened 2023-03-10
[Feature] Having a footer with a link to gdpr since it is obligatory


realex64:
Hi folks,

i would like to have a footer in freshrss webgui, where i can place a link to gdpr text.

In the EU this is mandatory, even for private instances. Having no privacy statement every lawyer can send you a warning notice with costs... No fun...

Thank you!

Kind regards

Alex

math-GH:
It is possible to enable the TOC (Terms of Service), that will shown on the registration form:

https://freshrss.github.io/FreshRSS/en/admins/12_User_management.html

Right now the GUI does not have any footer.

math-GH:
@realex64 Does #5215 solve your needs?

realex64:
Hi math-GH,

thank you very much for having a look at my suggestion.

I am afraid, that using TOS is not enough, since the gdpr has to be reachable from everywhere within a website, even before user registration. This is, because logging the ip address for example is already the handling of private data.

Often this is achieved by a permanent link in the header or footer of a website. In german speaking nations it is called "Datenschutzerklärung", i guess it is called "privacy policy" in english.

It is the same at github.com. Even on the start page without being logged in there is a "privacy" link in the footer. After logging in, it stays there...

Having a second line kind of "about FreshRSS" at the start page with a link to gdpr.html - similar to tos.html could be a way to solve the problem. This link should be present in every site of FreshRSS.

Kind regards

Alex

aledeg:
GDPR or not, I think it would be a nice addition to have a footer support. It could be use for more things than just privacy concerns. I understand that not everyone want such thing so it could be either an option or an extension.

If it's an extension, we will need to add a hook in the template to do such thing.

math-GH:
While I implemented #5215 I had similar thought about the needs of having a data privacy section. Next thought was: German websites need an imprint page (Impressum) too. So there are some needs to improve the legal things.

I would prefer to have extensions for this cases, that cannot be covered by the TOS. FreshRSS is used worldwide. The application cannot cover all legislation worldwide, but it could support the chance to add needed information.

We do not have an overview how many installation are public and how many are hosted f.e. just on a RaspberryPi or other Docker environments. I would say that just a small amount of installations need the legal links. Right now some instances are public available and did not ask for such a page: https://freshrss.org/cloud-providers.html (it does not mean, that they do not need).

math-GH:
[...]

PR #5222, opened 2023-03-21
TOS in footer

math-GH:
Ref #5189
Follow up of #5215

Add the link to the Terms of Service to:

1) Login screen
grafik

2) Registration form
grafik

3) add the "about FreshRSS" link to the TOS page to keep the sites consistent and having a `<footer>`
grafik

1. create data/tos.html
2. see the screens

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/index/about.phtml
+++ b/app/views/index/about.phtml
@@ -45,3 +45,11 @@
 	<h2><?= _t('index.about.credits') ?></h2>
 	<p><?= _t('index.about.credits_content') ?></p>
 </main>
+
+<?php if (!FreshRSS_Auth::hasAccess()) { ?>
+<footer class="main-footer">
+	<?php if (file_exists(TOS_FILENAME)) { ?>
+		<a href="<?= _url('index', 'tos') ?>"><?= _t('index.tos.title')?></a>
+	<?php } ?>
+</footer>
+<?php } ?>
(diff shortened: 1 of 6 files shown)


app/views/index/tos.phtml
```diff
@@ -27,3 +27,9 @@
 	<h1><?= _t('index.tos.title')?></h1>
 	<?= $this->terms_of_service ?>
 </main>
+
+<?php if (!FreshRSS_Auth::hasAccess()) { ?>
```
Alkarex, line 31: Probably fine, but why the special case here?

Alkarex (approved):

app/views/index/tos.phtml
math-GH: If the user is not logged in then he/she has no navigation to switch to the about page from the TOS.

If the user is logged in, then he/she will find the TOS in the configs section, that has a navigation menu on the left hand side, where the About page is linked to. In the logged-in state the page does not have a footer.

app/views/index/about.phtml
```diff
@@ -45,3 +45,11 @@
 	<h2><?= _t('index.about.credits') ?></h2>
 	<p><?= _t('index.about.credits_content') ?></p>
 </main>
+
+<?php if (!FreshRSS_Auth::hasAccess()) { ?>
```
Alkarex, line 49: Ah, it is maybe more this special case that I was wondering about. Isn't it fine to always show the TOS there?

math-GH: I would like to show the footer only in the case if the user is not logged in, because in this case there is no left side menu.

If the user is logged in, then the TOS is shown in the menu one line bellow. A footer does not fit into the design yet.
grafik

PR #5230, opened 2023-03-26
Automated test towards PHPStan level 6

Alkarex:
With this automated test, files that are passing Level 6 are tested to avoid regressions towards Level 5 (which is the current baseline for the whole code-base)

It is included when calling `make test-all` or can be called specifically by `composer run-script phpstan-next`


tests/phpstan-next.txt
Alkarex, line 1: The idea is to remove lines from that list when pull requests are claiming an improved PHPStan level for some specific files

composer.json
```diff
@@ -57,14 +57,16 @@
         "phpcs": "phpcs . -s",
         "phpcbf": "phpcbf . -p -s",
         "phpstan": "phpstan analyse --memory-limit 512M .",
+        "phpstan-next": "phpstan analyse --level 6 --memory-limit 512M $(find . -type d -name 'vendor' -prune -o -name '*.php' -o -name '*.phtml' | grep -Fxvf ./tests/phpstan-next.txt | sort | paste -s)",
```
Alkarex, line 60: This command finds all relevant `*.php` and `*.phtml` files and then remove those that are in `phpstan-next.txt` before calling PHPStan on the remaining ones.

aledeg (approved):

composer.json
Alkarex: Example: https://github.com/FreshRSS/FreshRSS/pull/5230/commits/a5cc7b20f6b05ceae73a90dc418345f7f0dede6f

tests/phpstan-next.txt
Alkarex: Example: https://github.com/FreshRSS/FreshRSS/pull/5230/commits/a5cc7b20f6b05ceae73a90dc418345f7f0dede6f

tests/phpstan-next.txt
```diff
@@ -0,0 +1,93 @@
+# List of files, which are not yet passing PHPStan level 6 https://phpstan.org/user-guide/rule-levels
```
Alkarex, line 1: The idea is to remove lines from that list when pull requests are claiming an improved PHPStan level for some specific files.
Example: https://github.com/FreshRSS/FreshRSS/pull/5230/commits/a5cc7b20f6b05ceae73a90dc418345f7f0dede6f

PR #5234, opened 2023-03-28
Simplify extension method

Alkarex:
One parameter was not used

Furthermore, that unused parameter was preventing simpler calls (i.e. without a dedicated extension class). Now it is possible to register a call back with a lambda from anywhere:

```php
Minz_ExtensionManager::addHook('feed_before_actualize', function () {
	// Hello
});
```

### DIFF
--- a/lib/Minz/ExtensionManager.php
+++ b/lib/Minz/ExtensionManager.php
@@ -76,8 +76,6 @@ class Minz_ExtensionManager {
 			'signature' => 'PassArguments',
 		),
 	);
-	// @phpstan-ignore-next-line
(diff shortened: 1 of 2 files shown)


lib/Minz/ExtensionManager.php
```diff
@@ -76,8 +76,6 @@ class Minz_ExtensionManager {
 			'signature' => 'PassArguments',
 		),
 	);
-	// @phpstan-ignore-next-line
-	private static $ext_to_hooks = array();
```
Alkarex, line 80: PHPStan had previously detected that this property was not used.

PR #5239, opened 2023-03-29
PHPStan level 6 for all PDO and Exception classes

Alkarex:


app/Controllers/feedController.php
```diff
@@ -47,10 +47,11 @@ public static function addFeed($url, $title = '', $cat_id = 0, $new_cat_name = '
 		$url = trim($url);
 
 		/** @var string|null $url */
-		$url = Minz_ExtensionManager::callHook('check_url_before_add', $url);
-		if (null === $url) {
+		$urlHooked = Minz_ExtensionManager::callHook('check_url_before_add', $url);
+		if ($urlHooked === $url) {
 			throw new FreshRSS_FeedNotAdded_Exception($url);
```
Alkarex, line 52: There was a minor error here, caught by type check

ColonelMoutarde (approved):

app/Controllers/feedController.php
Alkarex: Fix regression https://github.com/FreshRSS/FreshRSS/pull/5253

PR #5243, opened 2023-03-30
Fix extensions in actualize_script

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/3440

Fix multiple bugs:

* The extension hook was called before registering all the extensions for the current user
* During a cron job to refresh feeds, extensions enabled by a previous user risked being enabled for another user.
* During a cron job to refresh feeds, system extensions were enabled multiple times
* And more details (see comments)


Alkarex:
@aledeg If you are around, do you happen to have a test case for that?

app/actualize_script.php
```diff
@@ -75,11 +75,10 @@ function notice(string $message): void {
 
 	FreshRSS_Auth::giveAccess();
 
-	Minz_ExtensionManager::callHook('freshrss_user_maintenance');
-
 	$app->init();
```
Alkarex, line 101: User extensions are first registered in `$app->init();`

aledeg:
> @aledeg If you are around, do you happen to have a test case for that?

Not at the moment. I was meaning to create one to do backend stuff but never took the time.

app/Controllers/extensionController.php
```diff
@@ -124,26 +126,36 @@ public function enableAction() {
...
 
 			$res = $ext->install();
 
 			if ($res === true) {
 				$ext_list = $conf->extensions_enabled;
+				$ext_list = array_filter($ext_list, function($key) use($type) {
+					// Remove from list the extensions that have disappeared or changed type
+					$extension = Minz_ExtensionManager::findExtension($key);
+					return $extension !== null && $extension->getType() === $type;
+				}, ARRAY_FILTER_USE_KEY);
```
Alkarex, line 150: We had a case of when extensions were deleted, they would stay enabled for ever in the configuration.

lib/Minz/ExtensionManager.php
```diff
@@ -88,7 +111,9 @@ class Minz_ExtensionManager {
 	 * <name> must match with the entry point in metadata.json. This class must
 	 * inherit from Minz_Extension class.
 	 */
-	public static function init() {
+	public static function init(): void {
+		self::reset();
```
Alkarex, line 115: During a cron job to refresh feeds, extensions enabled by a previous user risked being enabled for another user.
Furthermore, system extensions were enabled multiple times.

lib/Minz/ExtensionManager.php
```diff
@@ -208,10 +233,17 @@ public static function register($ext) {
...
+	 * @param 'system'|'user'|null $onlyOfType only enable if the extension matches that type. Set to null to load all.
 	 */
-	public static function enable($ext_name) {
+	private static function enable(string $ext_name, ?string $onlyOfType = null): void {
 		if (isset(self::$ext_list[$ext_name])) {
 			$ext = self::$ext_list[$ext_name];
+
+			if ($onlyOfType !== null && $ext->getType() !== $onlyOfType) {
+				// Do not enable an extension of the wrong type
+				return;
```
Alkarex, line 246: In the case of extensions changing type (from `user` to `system` or vice versa), we risked enabling wrongly or two times an extension

app/Controllers/extensionController.php
```diff
@@ -175,30 +187,36 @@ public function disableAction() {
...
-			} else {
-				Minz_Request::bad(_t('feedback.extensions.no_access', $ext_name), $url_redirect);
 			}
 
 			$res = $ext->uninstall();
 
 			if ($res === true) {
 				$ext_list = $conf->extensions_enabled;
-				$legacyKey = array_search($ext_name, $ext_list, true);
-				if ($legacyKey !== false) {	//Legacy format FreshRSS < 1.11.1
```
Alkarex, line 192: Removed support for 5 year-old configuration syntax for FreshRSS < 1.11.1.
Automatically updated by just saving the configuration

lib/Minz/Configuration.php
```diff
@@ -91,24 +91,6 @@ public static function get($namespace) {
...
-		$legacyKey = array_search($ext_name, $this->extensions_enabled, true);
-		if ($legacyKey !== false) {	//Legacy format FreshRSS < 1.11.1
-			unset($this->extensions_enabled[$legacyKey]);
-		}
-	}
-	public function addExtension($ext_name) {
-		if (!isset($this->extensions_enabled[$ext_name])) {
-			$this->extensions_enabled[$ext_name] = true;
-		}
-	}
```
Alkarex, line 110: Apparently dead code, unused

Alkarex:
Digging into the extensions code, there were multiple bugs, some a bit severe.

@aledeg and @Frenzie It would be great if you could give a try to this PR with some of your extensions, at least your own use cases.

lib/Minz/Extension.php
```diff
@@ -70,22 +83,23 @@ public function uninstall() {
 	/**
 	 * Call at the initialization of the extension (i.e. when the extension is
 	 * enabled by the extension manager).
+	 * @return void
 	 */
 	abstract public function init();
```
Alkarex, line 88: Note: We should not change the signature of methods supposed to be overridden

app/Controllers/extensionController.php
Alkarex: This would also do strange things with extensions that have changed type (e.g. enable an extension, which should not)

Alkarex:
Ping @marienfressinaud also, if you could give a quick try at this PR with your extensions

Alkarex:
Let's merge to start testing a bit more widely (Marien is not available in the coming days but will test a bit later)

marienfressinaud:
It took me some time, but it seems to work fine for my extensions :+1:

Alkarex:
Excellent :-) I will try to prepare a new release in the coming days

PR #5251, opened 2023-04-02
Remove ConfigurationSetter

Alkarex:
This class has not been maintained for a while. Only a subset of our configuration properties are there, and some code is not relevant anymore. Furthermore, it is relying exclusively on dynamically invoked functions, making it challenging to maintain, in particular to find out what is used and what is not, what is handled and what is not. It is not well suited for changes in data formats, which have been handled in the Context class instead. It is also not able to handle configuration properties that are missing. It is the class with most errors for PHPStan level 6 (179 errors). It is also making intense use of is_callable and call_user_func_array, which are performance killers.

Should the need arise again to perform validation of our internal configuration files, I suggest an implementation with the opposite approach, namely driven by our code instead of driven by the data.

In summary, at the moment, this class is costly, while not offering many guarantees.


lib/Minz/Configuration.php
```diff
@@ -73,42 +74,28 @@ public static function get($namespace) {
...
-		if ($legacyKey !== false) {	//Legacy format FreshRSS < 1.11.1
-			unset($this->extensions_enabled[$legacyKey]);
-		}
-	}
-	public function addExtension($ext_name) {
-		if (!isset($this->extensions_enabled[$ext_name])) {
-			$this->extensions_enabled[$ext_name] = true;
-		}
-	}
-
```
Alkarex: See https://github.com/FreshRSS/FreshRSS/pull/5243

ColonelMoutarde (approved):

app/Models/ConfigurationSetter.php
```diff
@@ -1,429 +0,0 @@
...
-	private function _posts_per_page(&$data, $value) {
-		$value = intval($value);
-		$data['posts_per_page'] = $value > 0 ? $value : 10;
-	}
-
-	private function _queries(&$data, $values) {
-		$data['queries'] = array();
-		foreach ($values as $value) {
-			if ($value instanceof FreshRSS_UserQuery) {
-				$data['queries'][] = $value->toArray();
```
Alkarex, line 105: This magic call was forgotten. Fixed in https://github.com/FreshRSS/FreshRSS/pull/5318

app/Models/ConfigurationSetter.php
```diff
@@ -1,429 +0,0 @@
...
-		$data['html5_notif_timeout'] = $value >= 0 ? $value : 0;
-	}
-
-	// It works for system config too!
-	private function _language(&$data, $value) {
-		$value = strtolower($value);
-		$languages = Minz_Translate::availableLanguages();
-		if (!in_array($value, $languages)) {
-			$value = 'en';
-		}
```
Inverle, line 88: https://github.com/FreshRSS/FreshRSS/pull/7878

PR #5254, opened 2023-04-03
Latvian localizations

Exerra:
This pull request adds i18n localizations for Latvian. Most of it is done, but I do have some questions before marking this PR as ready.

1. In the `gen.php` file, where are the `last_*` (starting from line 94) locales used? I need more info on this to be able to figure out the best declension.
2. In the `gen.php` file, what is the `dir` locale (line 122) used for?
3. In the `gen.php` file, where is the `category_empty` (line 128) locale used? I already translated, and it should be fine as is, but I want to double-check.
4. In the `conf.php` file, in line 133, what is `n°` meant to be?
5. I noticed that in `gen.php` the `lang` locale (line 140) contains names for each language. Do I need to go ahead and paste the LV lang locale in every single languages locale files?

PR #5256, opened 2023-04-03
Added myself to the credits

Exerra:
I have nothing to say, except that it feels weird adding myself to the credits


Alkarex:
Well deserved!
Ref https://github.com/FreshRSS/FreshRSS/pull/5254

PR #5261, opened 2023-04-04
Share in anonymous mode

### LINKED ISSUE #5248: [Feature] bouton partage pour toustes /  Share button for all
A clear and concise description of what the problem is. Ex. I’m always frustrated when […]

FR : Le bouton partage n'est pas visible en non connecté
En utilisation Planet le fait que les lecteurs et lectrices puissent aussi  partager les articles serait un plus

The share button is not visible when not connected
In use Planet the fact that readers can also share articles would be a plus

FR : Rendre les partages via les réseaux asociaux possible par tous et toutes.

Make sharing via asocial networks possible by all.

I don't now
**Additional context**

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/5248


math-GH:
What could be the security risks here?

I see a little risk, that the user (as default user) will enter some sharing services with credentials/API keys, that should not be published with unregistered users.

What else?

Alkarex:
> I see a little risk, that the user (as default user) will enter some sharing services with credentials/API keys, that should not be published with unregistered users.

Indeed, that is mostly it.

math-GH:
Ok, in general I am already thinking about to have a visible information, that the current user is the default user and the default user is used for anonym. reading.

For this PR, a quick win could be to have a simple warning banner on the sharing service configuration page. Something like: "Attention. This sharing services are used for public. Do not have private credentials (username, password, API keys) with the share URL"

Alkarex:
Let's review the types of shares and which of them are problematic, if any.
If there is any, maybe a safer approach would be to have a new property to enable some types of shares to be visible in anonymous mode.

filougeeknick:
Tests completed and passed
Info from my system: 
FreshRss 1.21.0 
PHP 7.3.16
Great thanks for the job

PR #5265, opened 2023-04-05
add missing php extension in composer.json

ColonelMoutarde:

- add missing php extension in composer.json

Pull request checklist:

- [x ] clear commit messages
- [ x] code manually tested


Alkarex:
We already have those dependencies, and some more, in our main https://github.com/FreshRSS/FreshRSS/blob/edge/composer.json
This `./lib/composer.json` is only for semi-automatic update of our third-party libraries.
So I am not sure of the need to add those dependencies. Have you experienced any bug / error?

If we add some there, it should be for a clear goal, while also making sure there are all those that are needed.
Maybe listing all the dependencies needed by the third-party libraries (and only those dependencies)?

PR #5283, opened 2023-04-10
PHPstan level 6 for Migrator.php

ColonelMoutarde:

- PHPstan level 6 for Migrator.php

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/lib/Minz/Migrator.php
+++ b/lib/Minz/Migrator.php
@@ -55,7 +55,7 @@ class Minz_Migrator
 		}
 
 		$lock_path = $applied_migrations_path . '.lock';
-		if (!mkdir($lock_path, 0770, true) && !is_dir($lock_path)) {
+		if (!@mkdir($lock_path, 0770, true)) {
 			// Someone is probably already executing the migrations (the folder
 			// already exists).
 			// We should probably return something else, but we don't want the
--- a/lib/Minz/Migrator.php
+++ b/lib/Minz/Migrator.php
@@ -159,13 +159,13 @@ class Minz_Migrator
 	 *
 	 * @param string $version The version of the migration (be careful, migrations
 	 *                        are sorted with the `strnatcmp` function)
-	 * @param callable $callback The migration function to execute, it should
+	 * @param ?callable $callback The migration function to execute, it should
 	 *                           return true on success and must return false
 	 *                           on error
 	 *
 	 * @throws BadFunctionCallException if the callback isn't callable.
 	 */
-	public function addMigration(string $version, callable $callback): void {
+	public function addMigration(string $version, ?callable $callback): void {
 		if (!is_callable($callback)) {
 			throw new BadFunctionCallException("{$version} migration cannot be called.");
 		}
--- a/lib/Minz/Migrator.php
+++ b/lib/Minz/Migrator.php
@@ -245,7 +245,7 @@ class Minz_Migrator
 	 * considered as successful. It is considered as good practice to return
 	 * true on success though.
 	 *
-	 * @return array<string|true> Return the results of each executed migration. If an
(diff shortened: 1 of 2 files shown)


Alkarex:
Don't you mean Level 6?

lib/Minz/Migrator.php
```diff
@@ -21,16 +21,16 @@ class Minz_Migrator
 	 * @param string $migrations_path
 	 * @param string $applied_migrations_path
 	 *
-	 * @throws BadFunctionCallException if a callback isn't callable.
-	 * @throws DomainException if there is no migrations corresponding to the
+	 * @return boolean|string Returns true if execute succeeds to apply
```
Alkarex:
```suggestion
	 * @return true|string Returns true if execute succeeds to apply
```

lib/Minz/Migrator.php
```diff
@@ -21,16 +21,16 @@ class Minz_Migrator
...
+	 *                        migrations, or a string if it fails.
+	 *@throws DomainException if there is no migrations corresponding to the
```
Alkarex:
```suggestion
	 * @throws DomainException if there is no migrations corresponding to the
```

lib/Minz/Migrator.php
```diff
@@ -55,7 +55,7 @@ public static function execute($migrations_path, $applied_migrations_path) {
 		}
 
 		$lock_path = $applied_migrations_path . '.lock';
-		if (!@mkdir($lock_path, 0770, true)) {
+		if (!mkdir($lock_path, 0770, true) && !is_dir($lock_path)) {
```
Alkarex: Why did you remove the `@`? It risks breaking some outputs

lib/Minz/Migrator.php
```diff
@@ -180,9 +178,9 @@ public function addMigration($version, $callback) {
 	 *
 	 * @see https://www.php.net/manual/en/function.strnatcmp.php
 	 *
-	 * @return array
+	 * @return array<string, callable>
```
Alkarex:
```suggestion
	 * @return array<string,callable>
```

My syntax highlight does not work well when there are spaces in types :-)

lib/Minz/Migrator.php
Alkarex: Why the `!is_dir($lock_path)` ? This looks wrong

Alkarex:

lib/Minz/Migrator.php
ColonelMoutarde: that's for remove the @

ColonelMoutarde:
> Don't you mean Level 6?

you'r right sorry

lib/Minz/Migrator.php
Alkarex: But I do not believe your new version is equivalent at all. The first mkdir is supposed to fail if the path exists. It is a mutex

lib/Minz/Migrator.php
```diff
@@ -247,11 +245,11 @@ public function upToDate() {
 	 * considered as successful. It is considered as good practice to return
 	 * true on success though.
 	 *
-	 * @return array Return the results of each executed migration. If an
+	 * @return array<string|bool> Return the results of each executed migration. If an
```
Alkarex, line 248: I believe this might be the following (to be double-checked)

```suggestion
	 * @return array<string|true> Return the results of each executed migration. If an
```

lib/Minz/Migrator.php
Alkarex: I am reverting this line, as i believe the new one is wrong

```suggestion
		if (!@mkdir($lock_path, 0770, true)) {
```

PR #5298, opened 2023-04-14
PHPstan level 6 for Category.php

ColonelMoutarde:
- PHPstan level 6 for Category.php

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -70,7 +70,7 @@ class FreshRSS_Category extends Minz_Model {
 		return $this->error;
 	}
 
-	/** @param bool|int|string $value */
+	/** @param bool|int $value */
 	public function _error($value): void {
 		$this->error = (bool)$value;
 	}
@@ -134,9 +134,7 @@ class FreshRSS_Category extends Minz_Model {
 		}
 	}
 
-	/** @param int|string $id */
-	public function _id($id): void {
-		$this->id = (int)$id;
+	public function _id(int $id): void {
 		if ($id == FreshRSS_CategoryDAO::DEFAULTCATEGORYID) {
 			$this->_name(_t('gen.short.default_category'));
 		}
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -12,7 +12,7 @@ class FreshRSS_Category extends Minz_Model {
 	 */
 	public const KIND_DYNAMIC_OPML = 2;
 
-	/**  @var int */
+	/** @var int */
 	private $id = 0;
 	/** @var int */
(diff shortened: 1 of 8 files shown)


app/Models/Category.php
```diff
@@ -4,46 +4,48 @@ class FreshRSS_Category extends Minz_Model {
...
+	/** @var int */
 	private $nbFeeds = -1;
+	/** @var int */
 	private $nbNotRead = -1;
 	/** @var array<FreshRSS_Feed>|null */
-	private $feeds = null;
+	private $feeds;
+	/** @var bool */
 	private $hasFeedsWithError = false;
+	/** @var array<string> */
```
Alkarex: From mobile, but I believe this is more like `array<string,mixed>`

app/Models/Category.php
```diff
@@ -4,46 +4,48 @@ class FreshRSS_Category extends Minz_Model {
...
-		if (isset($feeds)) {
+		if ($feeds !== null) {
 			$this->_feeds($feeds);
 			$this->nbFeeds = 0;
 			$this->nbNotRead = 0;
 			foreach ($feeds as $feed) {
 				$this->nbFeeds++;
 				$this->nbNotRead += $feed->nbNotRead();
-				$this->hasFeedsWithError |= $feed->inError();
+				$this->hasFeedsWithError = $feed->inError();
```
Alkarex: This change looks wrong

app/Models/Category.php
```diff
@@ -100,7 +111,7 @@ public function feeds(): array {
 			foreach ($this->feeds as $feed) {
 				$this->nbFeeds++;
 				$this->nbNotRead += $feed->nbNotRead();
-				$this->hasFeedsWithError |= $feed->inError();
+				$this->hasFeedsWithError = $feed->inError();
```
Alkarex: This change also looks wrong

Alkarex: You can make `hasFeedsWithError` to a `bool|int` to avoid having to cast all the time and also keep performance

app/Models/Category.php
```diff
@@ -109,34 +120,38 @@ public function feeds(): array {
...
-	public function hasFeedsWithError() {
+	public function hasFeedsWithError(): bool {
 		return $this->hasFeedsWithError;
 	}
 
-	public function attributes($key = '') {
-		if ($key == '') {
+	/**
+	 * @param string $key
+	 * @return string|array<string>|null
```
Alkarex: Cf. comment about `array<string,mixed` for attributes

PR #5303, opened 2023-04-15
Improved: CSS classes for theme preview slider changed

math-GH:
No functionality change.

This PR changes the CSS class names of the theme preview slider:
grafik

Before:
The class names used the term `slide` that could be misleading to the ID `#slider` (the slider is the right hand side slider, that appears for editing data).

After:
Introduced the term `theme-preview` and used it instead of `slide`. Now the term `slider` is only connected to the right hand slider.

It helps me a lot to keep the CSS code understandable. I will not have any effect on the functionality of the theme slider.

- CSS, PHTML: renamed classes

Use the theme slider

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/configure/display.phtml
+++ b/app/views/configure/display.phtml
@@ -47,7 +47,7 @@
 		<div class="form-group">
 			<label class="group-name" for="theme"><?= _t('conf.display.theme') ?></label>
 			<div class="group-controls">
-				<ul class="slides">
+				<ul class="theme-preview-list">
 					<?php $slides = count($this->themes); $i = 1; $themeAvailable = false; ?>
 					<?php  /** @var array{'id':string, 'deprecated':bool, 'author':string, 'name':string, 'description':string} $theme */
 						foreach($this->themes as $theme) { ?>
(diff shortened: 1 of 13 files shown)


Frenzie (commented):
What about something in the direction of .slideshow? The old name is a generic picture slider, the proposed name is specific to themes.

math-GH:
I thought about it too.
Right now this slide show is used only for the theme preview.

From the theme designing perspective it is better to know, that it is used only for the theme preview slide show.

And the `slideshow` has again the `slide` in its name and the `slides` itself should have also another name instead of `slide`.

Frenzie:
> From the theme designing perspective it is better to know, that it is used only for the theme preview slide show.

Hm, but that's a choice. The more you make it specific for that purpose, the less easily you can quickly drop it in on a feed that always sends out a post with multiple image attachments or something. Though personally I'd just prefer to scroll in that case.

There are some potential implications for the future, but perhaps that's alright. :-)

Frenzie (approved):

PR #5308, opened 2023-04-17
More consistent use of iterable type

Alkarex:
For `yield`
https://php.net/language.types.iterable


ColonelMoutarde (approved):

Alkarex:
Reverted in https://github.com/FreshRSS/FreshRSS/pull/5327

PR #5324, opened 2023-04-21
Incorrect url for articles in feed

### LINKED ISSUE #5323: [BUG] Incorrect URLs used in place of <link> tag
**Describe the bug**
RSS feed items can have multiple URLs associated with them, including their images, alongside the main article link. FreshRSS is sometimes confusing which link deserves primacy, and is not always using the `<link>` tag as the primary link for an article.

**To Reproduce**
Example feed: <https://www.theatlantic.com/feed/all/>
1. Add this feed to FreshRSS
2. Click on an article heading from this feed within FreshRSS or an API client. The user will be taken to a cover image for the article in their web browser.

**Expected behavior**
When clicking on the article heading, the user should be taken to the article in the feed item's `<link>` tag.

**Environment information**
- FreshRSS version: 1.21.0
- Installation type: Docker

**Additional context**
I have verified that the expected behaviour works in other RSS readers.

gjimish:
Closes #5323

- Get article URI from `attribs`

1. Subscribe [theatlantic.com](https://www.theatlantic.com/feed/all/) feed.
2. Instead of an image, each article's title should redirect you to the appropriate url.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -211,3 +211,4 @@ People are sorted by name so please keep this order.
 * [Yamakuni](https://github.com/Yamakuni): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:Yamakuni), [Web](https://ofanch.me/)
 * [yzqzss|一座桥在水上](https://github.com/yzqzss): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:yzqzss), [Web](https://blog.othing.xyz/)
 * [Zhiyuan Zheng](https://github.com/zhzy0077): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:zhzy0077)
+* [Jimish Gamit](https://github.com/gjimish): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:gjimish)
--- a/lib/SimplePie/SimplePie/Item.php
+++ b/lib/SimplePie/SimplePie/Item.php
@@ -1022,7 +1022,18 @@ class SimplePie_Item
 			}
 			if ($links = $this->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_20, 'link'))
 			{
-				$this->data['links']['alternate'][] = $this->sanitize($links[0]['data'], SIMPLEPIE_CONSTRUCT_IRI, $this->get_base($links[0]));
+				if (!empty($links[0]['data']))
+				{
+					$this->data['links']['alternate'][] = $this->sanitize($links[0]['data'], SIMPLEPIE_CONSTRUCT_IRI, $this->get_base($links[0]));
+				}
+				else
+				{
+					if (isset($links[0]['attribs']['']['href']))
+					{
+						$link_rel = (isset($links[0]['attribs']['']['rel'])) ? $links[0]['attribs']['']['rel'] : 'alternate';
+						$this->data['links'][$link_rel][] = $this->sanitize($links[0]['attribs']['']['href'], SIMPLEPIE_CONSTRUCT_IRI, $this->get_base($links[0]));
+					}
+				}
 			}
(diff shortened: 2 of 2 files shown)


Alkarex:
Copy of my response:

> I check the problem:
> 
>     * First, the feed respond with various formats (Atom, RSS 2.0) depending on HTTP Accept header.
> 
> 
> To get the version received by FreshRSS:
> 
> ```shell
> curl -v -A 'FreshRSS/1.21.1-dev (Linux; https://freshrss.org)' -H 'Accept: application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1' 'https://www.theatlantic.com/feed/all/'
> ```
> 
>     * Second, while the Atom version is fine, the RSS 2.0 version is broken, using a wrong link construct (I have replaced all the content with generic test strings):
> 
> 
> ```
> <?xml version="1.0" encoding="utf-8"?>
> <rss version="2.0"
>  xmlns:media="http://search.yahoo.com/mrss/"
>  xmlns:atom="http://www.w3.org/2005/Atom">
> 	<channel>
> 		<title>A feed title</title>
> 		<link>http://example.net/</link>
> 		<atom:link rel="self" href="http://example.net/a-feed.xml"/>
> 		<item>
> 			<guid isPermaLink="false">guid:123</guid>
> 			
> 			<link href="http://example.net/a-link-alternate" rel="alternate"/>
> 			<title type="html">An HTML item title</title>
> 			<media:content url="http://example.net/a-media-content.jpg"/>
> 		</item>
> 	</channel>
> </rss>
> ```
> 
> image
> 
> So the upstream feed is broken, and I am not sure we should support this bug. Maybe you could try to propose your patch to https://github.com/simplepie/simplepie to see what other people think of it. (We do have a local version of SimplePie, so we are able to ship custom patches though)
> 
> What we should do on FreshRSS side in any case is to allow custom HTTP headers in the advanced sections of feeds, for instance to override the default HTTP Accept header or other headers.

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/issues/5323#issuecomment-1517815117_

ColonelMoutarde (approved):

jtojnar:
I am not sure we want SimplePie to work around buggy feeds in single specific site. Unless there are other sites that also exhibit this error, maybe try letting TheAtlantic know first?

mikeage:
This patch fixes the link for me, but it still only displays the image associated with the article, rather than the full feed. This is still a functional regression from about 2-3 weeks ago, where the entire article was displayed in FreshRSS (very useful to those who do not have a paid subscription)

brevilo:
Looking at the (validating) atom version of said feed I think the issue might be that FreshRSS confuses the `<media:content>` element with the `<content>` and `<link>` elements. IOW, it seems as if only the former's `url` child is used as the image (source) and as the article link, while `<content>` and `<link>`  seem to be ignored (despite their existence). Just an idea...

But I agree, it's a feed-related regression since I didn't touch FreshRSS for quite a while. But they could have added the [Media RSS](https://www.rssboard.org/media-rss) stuff a few weeks ago and that might have triggered a pre-existing FreshRSS bug. Who knows... 🤷‍♂️

math-GH:
Any news here?

Alkarex:
If I am not mistaken, the problematic feed https://www.theatlantic.com/feed/all/ has been fixed, so this PR does not seem necessary anymore, for this feed at least.
Furthermore, after a huge delay to refactor the use of SimplePie (big mess) https://github.com/FreshRSS/FreshRSS/pull/4374, our SimplePie fork is now maintained on https://github.com/FreshRSS/simplepie so new PRs should be sent there, or even better upstream https://github.com/simplepie/simplepie (but the upstream repo is very slow).

PR #5328, opened 2023-04-23
PHPStan Level 7 for more DAO PDO

Alkarex:
With new PDO functions to address common type and check problems


lib/Minz/ModelPdo.php
```diff
@@ -179,4 +179,64 @@ public static function clean() {
...
+		$info = $stm == null ? $this->pdo->errorInfo() : $stm->errorInfo();
+		Minz_Log::error('SQL error ' . $callingFunction . ' ' . json_encode($info));
+		return null;
+	}
+
+	/**
+	 * @param array<string,int|string|null> $values
+	 * @return array<array<string,int|string|null>>|null
+	 */
+	public function fetchAssoc(string $sql, array $values = []): ?array {
```
Alkarex, line 231: New function(s) that can be used in many situations to handle the several null checks and logs needed when fetching data

app/Models/CategoryDAO.php
```diff
@@ -250,19 +250,12 @@ public function searchById(int $id): ?FreshRSS_Category {
 	/** @return FreshRSS_Category|null|false */
 	public function searchByName(string $name) {
 		$sql = 'SELECT * FROM `_category` WHERE name=:name';
-		$stm = $this->pdo->prepare($sql);
-		if ($stm == false) {
+		$res = $this->fetchAssoc($sql, ['name' => $name]);
```
Alkarex, line 253: Example of use of the new function to shorten code a lot while being stricter

app/Controllers/categoryController.php
```diff
@@ -97,14 +97,16 @@ public function updateAction(): void {
...
 				Minz_Request::bad(_t('feedback.sub.category.not_exist'), $url_redirect);
 			}
 
-			$cat = new FreshRSS_Category($name);
-			$values = array(
+			$values = [
 				'name' => $cat->name(),
-			);
+				'kind' => $cat->kind(),
+				'attributes' => $cat->attributes(),
```
Alkarex, line 110: We had a bug there, which was erasing the values of other columns when updating the name of a category.
I am not sure this function is used at all though (the one in `FreshRSS_subscription_Controller::categoryAction()` is used) - to be checked

p/api/greader.php
```diff
@@ -898,13 +898,15 @@ private static function renameTag(string $s, string $dest) {
 			$categoryDAO = FreshRSS_Factory::createCategoryDao();
 			$cat = $categoryDAO->searchByName($s);
 			if ($cat != null) {
-				$categoryDAO->updateCategory($cat->id(), array('name' => $dest));
+				$categoryDAO->updateCategory($cat->id(), [
+					'name' => $dest, 'kind' => $cat->kind(), 'attributes' => $cat->attributes()
```
Alkarex, line 902: We had a bug here: updating the category name would clear its other columns

app/Models/CategoryDAO.php
```diff
@@ -143,7 +143,7 @@ public function addCategoryObject(FreshRSS_Category $category) {
 	}
 
 	/**
-	 * @param array<string,mixed> $valuesTmp
+	 * @param array{'name':string,'kind':int,'attributes':array<string,mixed>} $valuesTmp
```
Alkarex, line 146: Adding this stricter type discovered two bugs (see below)

app/Controllers/tagController.php
```diff
@@ -132,7 +132,7 @@ public function renameAction(): void {
 		$targetTag = $tagDAO->searchByName($targetName);
 		if ($targetTag === null) {
 			// There is no existing tag with the same target name
-			$tagDAO->updateTag($sourceId, ['name' => $targetName]);
+			$tagDAO->updateTagName($sourceId, $targetName);
```
Alkarex, line 135: To avoid the same bugs than for categories (see below)

ColonelMoutarde (approved):

ColonelMoutarde (approved):

ColonelMoutarde (approved):

Issue #5134, opened 2023-02-23
GUI Configuration Complains About Not Having Write Permissions


jsphnecclia:
Unable to reach step 2 of the configuration process, the GUI says the several files and directories do not have group write permissions for group www-data. See image here:

![Screenshot 2023-02-23 at 1 04 01 AM](https://user-images.githubusercontent.com/51343659/220840990-88dc37e9-e466-48e2-8a00-0501774bb9f7.png)

 
However, the TMP directory is not among the directories whose group write permissions are ignored. It appears the difference might be in the use of touch instead of is_writable? I don't know. These are the permissions:

![Screenshot 2023-02-23 at 1 15 33 AM](https://user-images.githubusercontent.com/51343659/220841568-94703b00-cb88-48d3-a61b-bffe26f1ee8e.png)

math-GH:
Not always is the 'vistor' user the same like the admin/FTP/SSH user

jsphnecclia:
Im not sure what you mean. What would the visitor user be if not www-data? Should I change the user (or group)? or should I chmod o=rw ?

Alkarex:
Please tell more about your setup: OS, type of storage, how you deployed and installed FreshRSS, versions...

jsphnecclia:
Arch Linux, postgres, I followed the general installation (https://freshrss.github.io/FreshRSS/en/admins/03_Installation.html) with some help from "Installation on Debian 9/Ubuntu 16.04", running nginx. Let me know if you need to see any specific configs or logs!

I think Math-GH's solution seems to be what would fix my problem, but I dont fully understood what was meant. Do I need to have a dedicated freshrss user? And how would I go about setting that up?

Alkarex:
Could you try to run `./cli/access-permissions.sh` (first available in our current edge version of FreshRSS)

jsphnecclia:
maybe this might help: I dont think I have the edge version, however I do see the access-permissions.sh script. However, my FreshRSS folder is in /usr/share/FreshRSS, while in /srv/http/ I only have a link to p.

I ran the ./cli/access-permissions.sh, it did not complain, however, it also didn't fix the issue. It  appears the same as in my previous screenshots.

jsphnecclia:
> Not always is the 'vistor' user the same like the admin/FTP/SSH user

What do you mean by this?

jsphnecclia:
[...]

PR #5341, opened 2023-04-25
Improved: Show file permission information in install process, if the access fails

math-GH:
Inspired by #5134

This PR improves the install process step 2 (PHP + folder check).

Before:
just a information came up, that there are not enough permissions on a `./data` directory.

After:
If the permission check fails it shows more information:
- Owner of the directory
- Group of the directory
- file permissions of the directory
- process user (it was displayed in some languages, but not in all)

grafik
1) directory
2) Owner of the directory
3) Group of the directory
4) file permissions of the directory
5) Process owner
6) this directory is not available (because I deleted it ;) )

1. run a fresh installation
2. have "wrong" access configurations to the folders
3. see the messages

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
more notices:
- the source code in `install.php` is not in a good state. A refactoring is needed
- a common code base together with `checkInstall.php` would be great

lib/lib_install.php
```diff
@@ -75,6 +75,101 @@ function checkRequirements(string $dbType = ''): array {
...
+	$owner = [];
+	if(!$stat_file) {
+		//Couldnt stat file
+		$owner['fileowner'] = 'unknown';
+		$owner['filegroup'] = 'unknown';
+		return $owner;
+	}
+	if (function_exists('posix_getpwuid') && function_exists('posix_getgrgid')) {
+		$owner['fileowner'] = @posix_getpwuid($stat_file['uid'])['name'];
+		$owner['filegroup'] = @posix_getgrgid($stat_file['gid'])['name'];
```
math-GH: I need help here. I cannot solve this PHPstan level 7 error:
```
 Line   lib/lib_install.php                                                 
 ------ -------------------------------------------------------------------- 
  109    Cannot access offset 'name' on array|false.                         
  110    Cannot access offset 'name' on array{name: string, passwd: string,  
         gid: int, members: array<int, string>}|false.      
```

Alkarex: It is because `posix_getpwuid()` may return false.
If possible, try also on Windows as those functions are platform-dependent.

math-GH: fixed

math-GH: Windows has no user permissions things as far as I know. I tried the install process with XAMP on Win10:
grafik

Alkarex: I meant that the such system functions might not exist / work on other platforms such as Windows (which does have user and group permissions - cf. NTFS rights), so we should make sure that it does not fail / crash on non-Linux platforms.

lib/lib_install.php
```diff
@@ -75,6 +75,103 @@ function checkRequirements(string $dbType = ''): array {
...
+		$owner['filegroup'] = _t('install.check.unknown_process_username');
+		return $owner;
+	}
+	if (function_exists('posix_getpwuid') && function_exists('posix_getgrgid')) {
+		$uid_name = @posix_getpwuid($stat_file['uid']);
+		$gid_name = @posix_getgrgid($stat_file['gid']);
+		$owner['fileowner'] = is_array($uid_name) ? $uid_name['name'] : _t('install.check.unknown_process_username');
+		$owner['filegroup'] = is_array($gid_name) ? $gid_name['name'] : _t('install.check.unknown_process_username');
+		return $owner;
+	};
```
math-GH, line 109: @Alkarex 
Line 107: Is it not enough?

Alkarex: It probably is - I have not read the PR yet. We just need to be extra careful for things that are not natively cross-platform, to test them in multiple configurations

lib/lib_install.php
```diff
@@ -75,6 +75,103 @@ function checkRequirements(string $dbType = ''): array {
...
+	$owner['fileowner'] = 'UID: ' . $stat_file['uid'];
+	$owner['filegroup'] = 'GID: ' . $stat_file['gid'];
+	return $owner;
+}
+
+/* File permissions (read, write, executable)
+   Source: https://www.php.net/manual/en/function.fileperms.php
+*/
+function file_permissions(String $path): String {
```
ColonelMoutarde:
```suggestion
function file_permissions(string $path): string {
```

attention to the capitalization

lib/lib_install.php
```diff
@@ -75,6 +75,103 @@ function checkRequirements(string $dbType = ''): array {
...
+		if (!empty($output[0])) {
+			return $output[0];
+		}
+	}
+
+	return _t('install.check.unknown_process_username');
+}
+
+/** @return array<string> */
+function getOwnerOfFile(String $file): array {
```
ColonelMoutarde:
```suggestion
function getOwnerOfFile(string $file): array {
```

attention to the capitalization

lib/lib_install.php
```diff
@@ -41,14 +41,14 @@ function checkRequirements(string $dbType = ''): array {
 	$json = function_exists('json_encode');
 	$mbstring = extension_loaded('mbstring');
 	// @phpstan-ignore-next-line
-	$data = DATA_PATH && touch(DATA_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
+	$data = DATA_PATH && @touch(DATA_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
```
ColonelMoutarde: dont use '@' manage false return

lib/lib_install.php
```diff
@@ -41,14 +41,14 @@ function checkRequirements(string $dbType = ''): array {
...
 	// @phpstan-ignore-next-line
-	$cache = CACHE_PATH && touch(CACHE_PATH . '/index.html');
+	$cache = CACHE_PATH && @touch(CACHE_PATH . '/index.html');
```
ColonelMoutarde: same...

lib/lib_install.php
```diff
@@ -41,14 +41,14 @@ function checkRequirements(string $dbType = ''): array {
...
 	// @phpstan-ignore-next-line
 	$tmp = TMP_PATH && is_writable(TMP_PATH);
 	// @phpstan-ignore-next-line
-	$users = USERS_PATH && touch(USERS_PATH . '/index.html');
-	$favicons = touch(DATA_PATH . '/favicons/index.html');
+	$users = USERS_PATH && @touch(USERS_PATH . '/index.html');
+	$favicons = @touch(DATA_PATH . '/favicons/index.html');
```
ColonelMoutarde: same...

lib/lib_install.php
math-GH: I added it on purpose....
While I checked this and had the file not available, I got a fatal error

ColonelMoutarde: catch the fatal...

ColonelMoutarde: https://www.php.net/manual/en/function.touch.php

touch() return a boolean 
```
if (touch($filename)) {
    echo $filename . ' the modification time has been changed to the current time or the file has been created';
} else {
    echo 'an error during the creation of ' . $filename;
}
```

lib/lib_install.php
math-GH: Strange. I cannot reproduce the fatal error

lib/lib_install.php
math-GH: Strange. I cannot reproduce the fatal error

lib/lib_install.php
```diff
@@ -75,6 +75,103 @@ function checkRequirements(string $dbType = ''): array {
...
+		}
+	}
+
+	return _t('install.check.unknown_process_username');
+}
+
+/** @return array<string> */
+function getOwnerOfFile(string $file): array {
+	clearstatcache();
+	$stat_file = @stat($file);
```
ColonelMoutarde, line 95:
```suggestion
	$stat_file = stat($file);
```

lib/lib_install.php
Alkarex: I disagree about not using `@` as a general rule. Did you try your own suggestion @ColonelMoutarde ?

Here is an example:

```php
<?php
try {
	$ok = touch('/blabla/truc');
	echo json_encode($ok), "\n";
} catch (Throwable $ex) {
	echo 'Handled exception ' . $ex, "\n";
}
```

It outputs:

```
PHP Warning:  touch(): Unable to create file /blabla/truc because No such file or directory in /home/alexandre/GitHub/FreshRSS/tests/test-touch.php on line 3
false
```

lib/lib_install.php
Alkarex: Cf. my answer above. I disagree

lib/lib_install.php
ColonelMoutarde: https://stackoverflow.com/questions/1241728/can-i-try-catch-a-warning

Alkarex: Exactly, one cannot *catch the fatal* or *manage false return*. Approaches other than `@` that I have seen so far are verbose, heavy, and error prone.

Frenzie: Agreed, it's less readable at best, though a PHP newbie might not be familiar with `@` so in that sense a try/catch is potentially mildly preferable, though I doubt it. :-)

ColonelMoutarde: transforming warnings into exceptions is the method used by modern frameworks (laravel, symfony, yii...).
```php
<?php
set_error_handler(function ($severity, $message, $file, $line) {
    throw new \ErrorException($message, $severity, $severity, $file, $line);
});

$result = file_get_contents($url);
```
It works very well

Alkarex: Not for now, sorry. And we are not interested in raising any exception in this case, just suppressing the warning, as we are properly handling the different return codes.

math-GH: more about the `touch()`: `If the file does not exist, it will be created.` It will create an empty file. https://www.php.net/manual/en/function.touch.php
But: If the process has no rights to create the file, then it will throw the fatal error. 
the information `file does not exists` only would be enough here.

I would go forward with the `@` to suppress the error.

ColonelMoutarde: To check the permissions of a file via a PHP script, i.e. dynamically, we can use the PHP function fileperms() which returns the permissions for a file.

This function will return the permissions of a file in numerical form. We can then convert the result into octal form with the decoct() function to obtain the representation of the permissions according to our three numbers.

Note that fileperms() can return additional information depending on the system used. The last three digits returned are the permissions.

We can also use the functions is_readable() and is_writable() which will respectively determine if the file can be read and written to. These functions will return true if it is the case or false if it is not, which makes them very practical to use within a condition.

https://www.pierre-giraud.com/php-mysql-apprendre-coder-cours/existence-permission-effacer-fichier/

math-GH: see `lib_install.php` line 122:

`function file_permissions(string $path): string {`

math-GH: I added now a `file_exists()`

New code line:
`$data = DATA_PATH && file_exists(DATA_PATH . '/index.html') && touch(DATA_PATH . '/index.html');`

lib/lib_install.php
math-GH: I added now a file_exists()

New code line:
$data = DATA_PATH && file_exists(DATA_PATH . '/index.html') && touch(DATA_PATH . '/index.html');

lib/lib_install.php
math-GH: I added now a file_exists()

New code line:
$data = DATA_PATH && file_exists(DATA_PATH . '/index.html') && touch(DATA_PATH . '/index.html');

lib/lib_install.php
```diff
@@ -41,14 +41,14 @@ function checkRequirements(string $dbType = ''): array {
 	$json = function_exists('json_encode');
 	$mbstring = extension_loaded('mbstring');
 	// @phpstan-ignore-next-line
-	$data = DATA_PATH && touch(DATA_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
+	$data = DATA_PATH && file_exists(DATA_PATH . '/index.html') && touch(DATA_PATH . '/index.html');	// is_writable() is not reliable for a folder on NFS
 	// @phpstan-ignore-next-line
-	$cache = CACHE_PATH && touch(CACHE_PATH . '/index.html');
+	$cache = CACHE_PATH && file_exists(CACHE_PATH . '/index.html') && touch(CACHE_PATH . '/index.html');
```
Alkarex: Please remove the `file_exist` as it might not exist

math-GH: I do not understand it

constants.php
```diff
@@ -44,6 +44,7 @@
 
 defined('UPDATE_FILENAME') or define('UPDATE_FILENAME', DATA_PATH . '/update.php');
 defined('USERS_PATH') or define('USERS_PATH', DATA_PATH . '/users');
+defined('FAVICONS_PATH') or define('FAVICONS_PATH', DATA_PATH . '/favicons');
```
Alkarex, line 52: Let's not add one global constant for all the folders we have ;-)
If you make that configurable, as is the case now, then you need to handle the case when it might be redefined to be elsewhere than `./data/favicons` for instance in our script to apply rights.
The rule so far is that only `./data/` and sub-folders should have write access (and `/tmp` or the system equivalent)

lib/lib_install.php
Alkarex: `./data/` can for instance be mounted as an empty Docker volume. We do have a script initialising some minimal structure, but we should have as little requirements about existing files there as possible.
Plus, it does not provide any additional information.

lib/lib_install.php
Alkarex: Nope, `is_writable()` is what we used before (please check git line log, and it is even in comment there) and it does not work for instance on a folder on NFS

lib/lib_install.php
math-GH: Please edit this PR as you wish. It would support me and this PR a lot. The discussion is very long around the `touch()`. I am not so familiar with this topics in details.

Alkarex: No need to change anything to those lines. The whole `touch()` part was exactly as it should. `@` is fine, `is_writable()` does not work for instance on a folder on NFS, `fileperms()` is a bad idea as it is poorly multi-platform.

constants.php
math-GH: I understand.
But ` DATA_PATH . '/favicons/'` is used 6 times in the code. I know the good practice that it is useful to reduce the string values that are often used and replace them with a constant.

Alkarex: Yes, but then if possible, make that a class constant or local constant instead of a global constant

app/i18n/en/install.php
```diff
@@ -55,15 +55,15 @@
 			'ok' => 'You have the cURL library.',
 		),
 		'data' => array(
-			'nok' => 'Check permissions on the <em>%1$s</em> directory for <em>%2$s</em> user. The HTTP server must have write permission.',
+			'nok' => 'Check permissions on the <em>%s</em> directory (Owner: %s, Group: %s, Rights: %s).<br />The HTTP server (User name: %s) must have write permissions.',
```
Alkarex, line 58: I am not so found of having so many duplicate strings. Let's make a little refactoring for having a single copy (both for `nok` and `ok`) with the folder name as parameter. We can use the `data` key.

lib/lib_install.php
```diff
@@ -71,6 +71,104 @@ function checkRequirements(string $dbType = ''): array {
...
+	$owner['fileowner'] = 'UID: ' . $stat_file['uid'];
+	$owner['filegroup'] = 'GID: ' . $stat_file['gid'];
+	return $owner;
+}
+
+/**
+ * File permissions (read, write, executable)
+ * Source: https://php.net/function.fileperms
+ */
+function file_permissions(string $path): string {
```
Alkarex, line 119: I am a bit ambivalent there. While slightly interesting, it is a lot of platform-specific low-level code, which I would rather avoid. It might also be misleading, because we are not checking for all types of errors, especially not on sub-folders/files.

math-GH: `fileperms()` return value will be `false` in any case if it has an error. file_permissions() will be `u` on Windows.

math-GH: How to proceed here? I would like to give information about the `current ` file permission as it is read by the system.
F.e. in https://github.com/FreshRSS/FreshRSS/issues/5134 we can see a lot of unclear information what is the current file permission situation.

Alkarex: It is mostly this function, which I do not like, as explained. I will try to find time to think a bit more about it, but I believe this would be better addressed with either:
* a bit more documentation
* some help scripts
* suggest Docker, YunoHost...
* (maybe) try to use system calls such as `exec('ls...')` instead of shipping low-level platform-specific code

PR #5353, opened 2023-04-28
phpstan level 9 for check.translation.php

ColonelMoutarde:

- phpstan level 9 for check.translation.php

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/cli/check.translation.php
+++ b/cli/check.translation.php
@@ -23,17 +23,17 @@ $displayResults = array_key_exists('d', $options);
 $displayReport = array_key_exists('r', $options);
 
 $isValidated = true;
-$result = array();
-$report = array();
+$result = [];
+$report = [];
 
+/** @var string $language */
 foreach ($languages as $language) {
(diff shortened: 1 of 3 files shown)


cli/check.translation.php
```diff
@@ -23,17 +23,17 @@
 $displayReport = array_key_exists('r', $options);
 
 $isValidated = true;
-$result = array();
-$report = array();
+$result = [];
+$report = [];
 
+/** @var string $language */
```
Alkarex: Please try to make sure that `$languages` has a known type before arriving there, instead of manually giving the type.
I think just making sure that `$options['l']` is a string should be enough

Alkarex:
I have fixed the types, and also updated `phpstan-next.txt`
https://github.com/FreshRSS/FreshRSS/pull/5353/commits/06815bb34c54216e8ffbb3ddf4dbad653b425e85

ColonelMoutarde:
> 

Thank you ☺️

PR #5354, opened 2023-04-28
PHPStan Level 7 complete DAOs

Alkarex:


app/Controllers/indexController.php
```diff
@@ -194,7 +194,7 @@ public function opmlAction(): void {
 		$get = FreshRSS_Context::currentGet(true);
 		if (is_array($get)) {
 			$type = $get[0];
-			$id = $get[1];
+			$id = (int)($get[1]);
```
ColonelMoutarde:
```suggestion
			$id = (int)$get[1];
```

app/Models/CategoryDAO.php
```diff
@@ -381,27 +377,20 @@ public function checkDefault() {
 
 	public function count(): int {
 		$sql = 'SELECT COUNT(*) AS count FROM `_category`';
-		$stm = $this->pdo->query($sql);
-		$res = $stm->fetchAll(PDO::FETCH_ASSOC);
-		return $res[0]['count'];
+		$res = $this->fetchColumn($sql, 0);
+		return isset($res[0]) ? (int)($res[0]) : -1;
```
ColonelMoutarde:
```suggestion
		return isset($res[0]) ? (int)$res[0] : -1;
```

app/Models/CategoryDAO.php
```diff
@@ -381,27 +377,20 @@ public function checkDefault() {
...
 
 	public function countFeed(int $id): int {
 		$sql = 'SELECT COUNT(*) AS count FROM `_feed` WHERE category=:id';
-		$stm = $this->pdo->prepare($sql);
-		$stm->bindParam(':id', $id, PDO::PARAM_INT);
-		$stm->execute();
-		$res = $stm->fetchAll(PDO::FETCH_ASSOC);
-		return $res[0]['count'];
+		$res = $this->fetchColumn($sql, 0, [':id' => $id]);
+		return isset($res[0]) ? (int)($res[0]) : -1;
```
ColonelMoutarde:
```suggestion
		return isset($res[0]) ? (int)$res[0] : -1;
```

app/Models/CategoryDAO.php
```diff
@@ -381,27 +377,20 @@ public function checkDefault() {
...
 
 	public function countNotRead(int $id): int {
 		$sql = 'SELECT COUNT(*) AS count FROM `_entry` e INNER JOIN `_feed` f ON e.id_feed=f.id WHERE category=:id AND e.is_read=0';
-		$stm = $this->pdo->prepare($sql);
-		$stm->bindParam(':id', $id, PDO::PARAM_INT);
-		$stm->execute();
-		$res = $stm->fetchAll(PDO::FETCH_ASSOC);
-		return $res[0]['count'];
+		$res = $this->fetchColumn($sql, 0, [':id' => $id]);
+		return isset($res[0]) ? (int)($res[0]) : -1;
```
ColonelMoutarde:
```suggestion
		return isset($res[0]) ? (int)$res[0] : -1;
```

app/Models/FeedDAO.php
```diff
@@ -589,7 +594,7 @@ public static function daoToFeed(array $listDAO, ?int $catID = null): array {
 				continue;
 			}
 			if (isset($dao['id'])) {
-				$key = $dao['id'];
+				$key = (int)($dao['id']);
```
ColonelMoutarde:
```suggestion
				$key = (int)$dao['id'];
```

app/Controllers/indexController.php
Alkarex: Yes, it works without the parentheses, but I find the syntax dangerous as it requires to be sure of the priorities and I can better like to be explicit as soon as there is a series of connections (here, a variable and then an array index, but it could be more complex).

ColonelMoutarde: I reassure you, there is no risk. The cast is done on everything on the right. In the same way if you use (object) on an array it is the whole array that is transformed into StdClass:: . If you use (object)$myArray['foo'] it is only the key foo that is cast.

Alkarex: My point is that it requires not making mistakes with operator priorities, because it is not entirely true that it is everything on the right. See examples, which outputs two different things (quite trivial with the `+` operator, but still):

```php
<?php
$x = 1;
echo json_encode((string)$x + 1), "\n";
```

```php
<?php
$x = 1;
echo json_encode((string)($x + 1)), "\n";
```

But you are right that the current cases are very simple

ColonelMoutarde (approved):

PR #5358, opened 2023-04-29
Automatic trusted_sources during install

### LINKED ISSUE #5357: [BUG] Installation process step 3 not possible when already integrated in traefik and authelia
**Describe the bug**
If you use the Docker image from linux servers (lscr.io/linuxserver/freshrss), there will be a fatal error thrown when calling httpAuthUser() in step 3 of the installation.

```html
<br />
--
  | <b>Fatal error</b>:  Uncaught FreshRSS_Context_Exception: System configuration not initialised! in /app/www/lib/lib_rss.php:670
  | Stack trace:
  | #0 /app/www/lib/lib_rss.php(685): checkTrustedIP()
  | #1 /app/www/app/install.php(571): httpAuthUser()
  | #2 /app/www/app/install.php(738): printStep3()
  | #3 /app/www/p/i/index.php(30): require('...')
  | #4 {main}
  | thrown in <b>/app/www/lib/lib_rss.php</b> on line <b>670</b><br />
```

httpAuthUser will call checkTrustedIP but the configuration about the trusted IP is not done at this point.
This means $_SERVER['HTTP_REMOTE_USER'] is present, checkTrustedIP is executed but not configured now and there is no check if you are at the installation process.

**To Reproduce**
* Docker installation with traefik reverse proxy and authelia SSO
* Should also be reproduceble when http header "Remote-User" is present and step 3 is shown in browser.

**Expected behavior**
At least a installation process which can be completed.
To have a full working authetication with trusted http headers will need the trusted ip configuration too which is not shown in the installation.

My personal expectation will be to have at least the checkTrustedIP function written in a way that it detects a FreshRSS installation process and accept any IP at this point.

**Screenshots**
n/a

- Device: any
- OS: any
- Browser: any
- FreshRSS version: 1.21.0
- Database version: sqlite
- PHP version: 8.1
- Installation type: Docker

**Additional context**
* Docker configured
* Traefik working as reverse proxy
* Authelia configured as SSO container
* New FreshRSS docker container already configured with needed docker labels for traefik and authelia

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/5357
Fix also the fact that http_auth was broken during install.
And make the `checkCIDR()` robust when using IP addresses without a range suffix (was previously generating warnings).

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -634,13 +634,22 @@ function ipToBits(string $ip): string {
  */
 function checkCIDR(string $ip, string $range): bool {
 	$binary_ip = ipToBits($ip);
-	list($subnet, $mask_bits) = explode('/', $range);
-	$mask_bits = intval($mask_bits);
+	$split = explode('/', $range);
+
+	$subnet = $split[0] ?? '';
+	if ($subnet == '') {
+		return false;
+	}
 	$binary_subnet = ipToBits($subnet);
+
+	$mask_bits = $split[1] ?? '';
+	$mask_bits = intval($mask_bits);
+	if ($mask_bits == 0) {
+		$mask_bits = null;
+	}
 
 	$ip_net_bits = substr($binary_ip, 0, $mask_bits);
 	$subnet_bits = substr($binary_subnet, 0, $mask_bits);
-
 	return $ip_net_bits === $subnet_bits;
 }
 
@@ -665,14 +674,14 @@ function checkTrustedIP(): bool {
 	return false;
 }
 
-function httpAuthUser(bool $onlyTrustedIp = true): string {
+function httpAuthUser(bool $onlyTrusted = true): string {
 	if (!empty($_SERVER['REMOTE_USER'])) {
 		return $_SERVER['REMOTE_USER'];
 	}
 	if (!empty($_SERVER['REDIRECT_REMOTE_USER'])) {
 		return $_SERVER['REDIRECT_REMOTE_USER'];
 	}
-	if (!$onlyTrustedIp || checkTrustedIP()) {
+	if (!$onlyTrusted || checkTrustedIP()) {
 		if (!empty($_SERVER['HTTP_REMOTE_USER'])) {
 			return $_SERVER['HTTP_REMOTE_USER'];
 		}
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -644,7 +644,7 @@ function checkCIDR(string $ip, string $range): bool {
 
 	$mask_bits = $split[1] ?? '';
 	$mask_bits = intval($mask_bits);
-	if ($mask_bits == 0) {
+	if ($mask_bits === 0) {
 		$mask_bits = null;
 	}
 
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -643,7 +643,7 @@ function checkCIDR(string $ip, string $range): bool {
 	$binary_subnet = ipToBits($subnet);
 
 	$mask_bits = $split[1] ?? '';
-	$mask_bits = intval($mask_bits);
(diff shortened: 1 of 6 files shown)


CDaxi:
**OK:**
The installation form issue is fixed.

**Still an issue:**
At the config.php:
* auth_type is still set to "form"
* trusted_sources will not contain the ip

There should be at least an information that these data has to manually changed by the admin.
The better way would be to add the input for trusted_sources at the installation process and display it only if the http_auth is selected.

Thanks for the quick response to my issue.

Alkarex:
Could you please try again @CDaxi ?

lib/lib_rss.php
```diff
@@ -634,13 +634,22 @@ function ipToBits(string $ip): string {
...
+	$split = explode('/', $range);
+
+	$subnet = $split[0] ?? '';
+	if ($subnet == '') {
+		return false;
+	}
 	$binary_subnet = ipToBits($subnet);
 
+	$mask_bits = $split[1] ?? '';
+	$mask_bits = intval($mask_bits);
```
ColonelMoutarde:
```suggestion
	$mask_bits = (int)$mask_bits;
```

lib/lib_rss.php
```diff
@@ -634,13 +634,22 @@ function ipToBits(string $ip): string {
...
+	if ($mask_bits == 0) {
```
ColonelMoutarde:
```suggestion
	if ($mask_bits === 0) {
```

CDaxi:
Installation is working fine now.
Trusted proxy with subnet /32 is automatically set - great.
But after logging out from authelia and logging in with an non existing user, I will get a 500 error.

PHP Fatal error:  Uncaught TypeError: FreshRSS_user_Controller::createUser(): Argument #2 ($email) must be of type string, null given, called in /var/www/FreshRSS/app/Models/Auth.php on line 83 and defined in /var/www/FreshRSS/app/Controllers/userController.php:215\nStack trace:\n#0 /var/www/FreshRSS/app/Models/Auth.php(83): FreshRSS_user_Controller::createUser()\n#1 /var/www/FreshRSS/app/Models/Auth.php(36): FreshRSS_Auth::accessControl()\n#2 /var/www/FreshRSS/app/FreshRSS.php(75): FreshRSS_Auth::init()\n#3 /var/www/FreshRSS/app/FreshRSS.php(48): FreshRSS::initAuth()\n#4 /var/www/FreshRSS/p/i/index.php(55): FreshRSS->init()\n#5 {main}\n  thrown in /var/www/FreshRSS/app/Controllers/userController.php on line 215

The user can not be created because there is null handed over at createUser at argument 2.
Authelia will provide the users E-Mail at the Header "Remote-Email" (as described here: https://www.authelia.com/integration/trusted-header-sso/introduction/).

After adding "HTTP_REMOTE_EMAIL" in config.php at "http_auth_auto_register_email_field", the user can be created.
This configuration can also not be set at the installation process.

Perhaps there can also be a check if $_SERVER['HTTP_REMOTE_EMAIL'] is present and be automatically set in the config?
The perfect solution will be a prefilled installation input for the trusted_sources and http_auth_auto_register_email_field only visible when selection http_auth at installation.

Alkarex:
Thanks for the tests and feedback. The null exception should be fixed. The e-mail is not actually used to anything for this type of login. Could you please try again?

I would like in the current PR to only fix the (more urgent) regressions, while new features and enhancements can be done in a future one.

CDaxi:
Auth works now like a charm.
I agree, to get the additional configurations available at frontend at a different PR in future.

Thanks a lot for the quick response and fixes. Great job. 👏

Alkarex:
Thanks for the feedback! Feel free to collect your ideas of improvements into a new issue, if sufficiently relevant

PR #5359, opened 2023-04-29
Update .gitignore: ignore the directory data.back

math-GH:
While an update the `./data` directory is backed-up into `./data.back`.
It is not needed for git, while `./data.backup` could store gigabytes of data (because of the `cache `directory)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /bin/
 /node_modules/
 /vendor/
-/data.backup/
+/data.back/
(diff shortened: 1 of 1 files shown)


Alkarex:
Where does this `./data.backup/` come from?

math-GH:
> Where does this `./data.backup/` come from?

oh, my fault. It is `./data.back/`, created by `update.php` via https://update.freshrss.org/?v=1.20.0

PR #5373, opened 2023-05-05
phpstan level 7 for feedController.php

ColonelMoutarde:

- phpstan level 7 for feedController.php

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -364,7 +364,6 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				continue;	//When PubSubHubbub is used, do not pull refresh so often
 			}
 
-			$mtime = 0;
 			if ($feed->mute()) {
 				continue;	//Feed refresh is disabled
 			}
@@ -373,7 +372,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				($feed->lastUpdate() + 10 >= time() - (
 					$ttl == FreshRSS_Feed::TTL_DEFAULT ? FreshRSS_Context::$user_conf->ttl_default : $ttl))) {
 				//Too early to refresh from source, but check whether the feed was updated by another user
-				$mtime = (int) $feed->cacheModifiedTime();
+				$mtime = (int)$feed->cacheModifiedTime();
 				if ($feed->lastUpdate() + 10 >= $mtime) {
 					continue;	//Nothing newer from other users
 				}
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -364,7 +364,6 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				continue;	//When PubSubHubbub is used, do not pull refresh so often
 			}
 
-			$mtime = 0;
 			if ($feed->mute()) {
 				continue;	//Feed refresh is disabled
 			}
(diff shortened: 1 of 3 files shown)


app/Controllers/feedController.php
```diff
@@ -530,10 +541,10 @@ public static function actualizeFeed(int $feed_id, string $feed_url, bool $force
 			}
 
 			$feedDAO->updateLastUpdate($feed->id(), false, $mtime);
-			$needFeedCacheRefresh |= ($feed->keepMaxUnread() != false);
+			$needFeedCacheRefresh |= ($feed->keepMaxUnread() !== false);
```
Alkarex: Watch out, `!==` is indeed more strict, yes, but also less sure (more risk of errors because the tests are not lenient enough). This line is a good example.

```
<?php
$a = true;
$b = false;
$b |= $a;
echo json_encode($b), "\n";
// Output: 1
```

This is a PHP shortcoming, which changes type under the hood from `bool` to `int` (I hope this will be fixed in a future version), but it is anyway safer / more robust to test with *truthy* / *falsy*.

I am sending a patch with various fixes shortly

app/Controllers/feedController.php
```diff
@@ -564,21 +575,21 @@ public static function actualizeFeed(int $feed_id, string $feed_url, bool $force
 			}
 
 			if ($simplePie != null) {
-				if ($feed->name(true) == '') {
+				if ($feed->name(true) === '') {
```
Alkarex, line 576: It does not matter much here, but this is another example where `===` is less safe.
Imagine we change `->name()` to return `null` if undefined or `false` in case of error (which is quite common), and the new code will lead to a bug, while the previous code was safe.

app/Controllers/feedController.php
```diff
@@ -795,8 +806,11 @@ public function deleteAction(): void {
 				break;
 			default:
 				$redirect_url = Minz_Request::paramString('r', true);
-				if ($redirect_url == '') {
-					$redirect_url = array('c' => 'subscription', 'a' => 'index');
+				if ($redirect_url === '') {
+					$redirect_url = ['c' => 'subscription', 'a' => 'index'];
+				}
+				if (!is_array($redirect_url)) {
+					$redirect_url = ['c' => 'subscription', 'a' => 'index'];
```
Alkarex: That looks wrong: You never use the provided `$redirect_url`

Alkarex: Ah, it looks like it was a TODO (cf. comment in code higher up)

Alkarex:
A few fixes. Double-check welcome https://github.com/FreshRSS/FreshRSS/pull/5373/commits/fef793926dba5b0675a860d436f374d92a7b534d

app/Controllers/feedController.php
ColonelMoutarde: It doesn't matter much here, but this is another example where === is less safe.
Let's say we change ->name() so that it returns null if undefined or false on error (which is quite common), and the new code will lead to a bug, whereas the previous code was safe.

Here is the function that is called when we do ->name();

```php
public function name(bool $raw = false): string {
		return $raw || $this->name != '' ? $this->name : preg_replace('%^https?://(www[.])?%i', '', $this->url);
	}
```

In any case we have a string in output if the function, what you do not want is that your string is of type: ""

ColonelMoutarde:
allright for me

app/Controllers/feedController.php
Alkarex: I know what we have now. My point was that it could change, and null / false are quite common

PR #5406, opened 2023-05-16
PHPStan Level 7 complete

Alkarex:
And initiate PHPStan Level 8 https://phpstan.org/user-guide/rule-levels


ColonelMoutarde (approved):

ColonelMoutarde (approved):

Alkarex:
For PHPStan Level 8, as we have many errors due to the same root cause, I need to think of a couple of global approaches

PR #5434, opened 2023-05-30
phpstan-8 for category class

ColonelMoutarde:

- phpstan-8 for category class

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -22,7 +22,7 @@ class FreshRSS_Category extends Minz_Model {
 	private $nbFeeds = -1;
 	/** @var int */
 	private $nbNotRead = -1;
-	/** @var array<FreshRSS_Feed>|null */
+	/** @var array<FreshRSS_Feed> */
 	private $feeds;
 	/** @var bool|int */
 	private $hasFeedsWithError = false;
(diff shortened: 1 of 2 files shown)


app/Models/Category.php
```diff
@@ -22,7 +22,7 @@ class FreshRSS_Category extends Minz_Model {
 	private $nbFeeds = -1;
 	/** @var int */
 	private $nbNotRead = -1;
-	/** @var array<FreshRSS_Feed>|null */
+	/** @var array<FreshRSS_Feed> */
```
Alkarex: I am not sure this is a good solution, as we need to test for nullability below: `if ($this->feeds === null)`

ColonelMoutarde: got to matermost ?

ColonelMoutarde: $feedDAO->listByCategory($this->id()); return always a array() of FreshRSS_Feed

app/Models/Category.php
```diff
@@ -99,7 +99,12 @@ public function nbNotRead(): int {
...
 
-	/** @return array<FreshRSS_Feed> */
+	/**
+	 * @return array<FreshRSS_Feed>
+	 * @throws Minz_ConfigurationNamespaceException
+	 * @throws Minz_PDOConnectionException
+	 */
+
 	public function feeds(): array {
 		if ($this->feeds === null) {
```
Alkarex, line 108: This is the line testing for null

app/Models/Category.php
Alkarex: See below, line 109: `if ($this->feeds === null)`

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/5434/commits/2a0cbaaf03a4f56031ded209f8920c3322dc513c

Alkarex (approved):

PR #5439, opened 2023-06-04
Update dark theme css to lower brightness to all icons

Alwaysin:
Some elements had lower brightness already like sites favicons, but it felt off to have the checkbox, nav icons, etc to have full brightness and was hurting my eyes ^^'

### DIFF
--- a/p/themes/Dark/dark.css
+++ b/p/themes/Dark/dark.css
@@ -110,6 +110,7 @@ img.favicon {
 
 body img,
 body video,
+p.help .icon,
 .icon {
 	filter: brightness(.6) contrast(1.2);
 }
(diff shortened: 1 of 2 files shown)


Alwaysin:
P.S.: this makes sense only when having ```'icons_as_emojis' => true``` in the user config.

math-GH:
I do not understand what is your issue and what is your goal.

Your PR right now has this result:
Before: The help icon is darker
grafik

AfteR: The help icon is lighter
grafik

Because of the rules in frss.css

Alwaysin:
As I've said, you need the special parameter in your user config file to see the thing.

Before:

image

After:

image

The emojis are hurting the eyes when you're in a completely dark room.

I've removed p.help cause I couldn't find where it was and thought it was unused, but we could put it back there, just adding the .icon alone.

Alkarex:
All good @math-GH ?

math-GH (approved):

PR #5449, opened 2023-06-09
Allow deep link to extension configuration

Alkarex:
Allow a full screen view of an extension configuration, with a deep-link URL.
Example:
image

Background: I have a use-case for which I need to provide a direct link to an extension configuration page.


Alkarex:
For those working with extensions, could you please double-check that it does not break your use-cases?

math-GH:
There is a bug:
1. go to extension config page
2. open the settings of a extension, f.e. CustomCSS (the slider will be opened)
3. save the settings

expected: the slider will be shown as slider or closed.
What is visible: The settings will be opened in full screen

Alkarex:
What about now @math-GH ?

ColonelMoutarde (approved):

math-GH (approved):

math-GH:
Additional question:
If you open the config in a new tab, than the new tab has no navigation on the left hand side. Is it on purpose?
Compared to the feed configuration, where the left navigation is given in a new tab

Alkarex:
> Additional question:
> If you open the config in a new tab, than the new tab has no navigation on the left hand side. Is it on purpose?
> Compared to the feed configuration, where the left navigation is given in a new tab

No, good point, it was not on purpose

Alkarex:
@math-GH Done https://github.com/FreshRSS/FreshRSS/pull/5449/commits/d546aff5bc52a62ff3cc8e92f80e49af5b7bd476

math-GH:
Please double check: Does it make sense in the "full screen mode" (extension configs in new tab opened), that the config page will no be closed, when clicking the submit button?

Alkarex:
> Does it make sense in the "full screen mode" (extension configs in new tab opened), that the config page will no be closed, when clicking the submit button?

Do you mean automatically closing the window/tab (if so, then no: in my case, it is the main window, not a new tab; and in other cases, I only think this should be done for modal windows, not tabs), or redirecting to another page after submission (if so, I do not really mind, but I find it fine as it is now)

math-GH:
fine for me too

math-GH:
@Alkarex I found one thing after it was merged: The `<title>` is just the default `FreshRSS`. It would be better to have the title of the extension in the tab.

Alkarex:
> @Alkarex I found one thing after it was merged: The `<title>` is just the default `FreshRSS`. It would be better to have the title of the extension in the tab.

Done in https://github.com/FreshRSS/FreshRSS/pull/5469

Issue #3374, opened 2021-01-15
request for documentation of "archiving" section


rezad1393:
can someone explain the archiving section?
I have 274 feeds. because I removed and added many feeds the number that is shows for adding new feeds ( for example like this  freshrssdomain//i/?c=subscription&id=1283 ) is now in 1300.
I have set 

Purge policy(By default)Maximum number of articles to keep 8
Maximum age of articles to keep: 1 month
Never delete favourites:  yes
Never delete labels: yes 
Never delete unread articles: yes
Minimum number of articles to keep per feed: 5

I have read that the first one doesn't count if it is too low.
I also have zero unread articles.
and I don't know what label articles are so I ignore that.

but even setting all this and having 300 favorite items and using purge , I still have this:
Maintenance Current user
4 644 articles (23.11 MiB)

how do I still have 4644 articles?
(Minimum number of articles to keep per feed=5) * 247 feeds + 300 favs = at most 1600

can someone explain this section in simple terms and show when they overlap what happens?
maybe with an example with numbers?

Alkarex:
Hello,
FreshRSS will at least keep the articles that are still on the original servers, which is often 10, 20, 50 articles per feed. The idea is more to mark articles as read (that can soon be done automatically https://github.com/FreshRSS/FreshRSS/pull/3303 ) rather than deleting them so aggressively. I do not recommend values below 100 for maximum number of articles per feed.

rezad1393:
> Hello,
> FreshRSS will at least keep the articles that are still on the original servers, which is often 10, 20, 50 articles per feed. The idea is more to mark articles as read (that can soon be done automatically #3303 ) rather than deleting them so aggressively. I do not recommend values below 100 for maximum number of articles per feed.

1-why?
freshrss uses sqlite correct?
then having more articles decreases the performance right?
maybe I am missiong something.

2-so all my questions about articles not deleted are related to the "FreshRSS will at least keep the articles that are still on the original servers," part?
that is necessary for feed update check to diff the data and cant be changed, correct?

3-100 is too much. but if I put a number like 20 for global minimum per feed then feeds that I higher that that on update, they just ignore this limit right?


UPDATE: never mind. I tested and this not global sum but per feed.
maybe this should be put in text of option that this is per feed.
4- is the "Maximum number of articles to keep" is a global limit for sum of  all articles for all feeds that if for example I put 100000 on it and remove the time limit then it will not remove articles until sum of all them execs 100000?:

[...]

PR #5458, opened 2023-06-11
Clarify that maximum number to keep is per feed

yubiuser:
Closes #https://github.com/FreshRSS/FreshRSS/issues/3374

Within the archiving settings there is an option to set the maximum number of articles to keep. So far, this sounded like a **global** option, but it is **per feed**. I just amended the text (in the languages I feel confident) to clarify the option.


app/i18n/fr/conf.php
```diff
@@ -133,7 +133,7 @@
 		'number' => 'Filtre n°%d',
 		'order_asc' => 'Afficher les articles les plus anciens en premier',
 		'order_desc' => 'Afficher les articles les plus récents en premier',
-		'search' => 'Recherche de « %s »',
+		'search' => 'Recherche de « %s »',
```
Alkarex: Did you change the whitespace there? In any case, does not seem related to this PR

Alkarex:
Thanks and good idea.
Please add a `// DIRTY` comment at then end of the lines for other languages

app/i18n/fr/conf.php
yubiuser: Not on purpose. Must have happend automatically by `vscode`. I'll take a look to see if I can undo that.

yubiuser: I have no idea what happend here or how to reverse it. Looking at the raw file and checking the ["hidden" characters ](https://www.soscisurvey.de/tools/view-chars.php) both versions are exactly the same.
Current `edge`: https://raw.githubusercontent.com/FreshRSS/FreshRSS/edge/app/i18n/fr/conf.php
My change: https://raw.githubusercontent.com/FreshRSS/FreshRSS/68c132bdd005d42ae92f2ee5c4f5d4e892dc7822/app/i18n/fr/conf.php

Alkarex:
You had a copy-paste error regarding `// DIRTY`. Fixed with `make fix-all`

app/i18n/fr/conf.php
Alkarex: No, the whitespace is not the same. We were using a non-breaking space. In any case, you can just avoid committing unrelated lines, but I would be curious to know what has changed it.
image

Alkarex: Fixed in https://github.com/FreshRSS/FreshRSS/pull/5458/commits/926774fdce386d156373c802d950b00593a490ff

yubiuser: Thanks.

Alkarex (approved):

Alkarex:

app/i18n/fr/conf.php
yubiuser: > but I would be curious to know what has changed it.

I tried to reproduce it again, but was not able to do so.

PR #5461, opened 2023-06-13
Docker: Debian 12

Alkarex:
https://www.debian.org/releases/bookworm/
With PHP 8.2.5 and Apache 2.4.57


Alkarex:
Let's merge that to get some larger feedback

PR #5480, opened 2023-06-18
Remove unneeded quotes in warning

### LINKED ISSUE #5479: Error log: "cURL error 22: The requested URL returned error: 503" [https://domain.tld/.rss]
cURL errors appears with additional `"`

grafik

This could be the root, but it is not the reason for this additional character in the log
https://github.com/FreshRSS/FreshRSS/blob/723f7577d0a388a90779930754c5aacb9f66b168/lib/SimplePie/SimplePie/File.php#L127-L131

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5479

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -400,11 +400,16 @@ class FreshRSS_Feed extends Minz_Model {
 
 				if ((!$mtime) || $simplePie->error()) {
 					$errorMessage = $simplePie->error();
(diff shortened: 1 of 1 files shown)


ColonelMoutarde (approved):

math-GH (approved):
it works fine: 
grafik

PR #5351, opened 2023-04-28
Add OpenID Connect

aaronschif:
Hello, word of warning, this is maybe the first PHP I have written, so I hope that I haven't done anything too crazy. 

I realize that env vars are not so common a configuration practice in PHP land, but it did seem a logical way of handling things in this case since all the config is actually in apache. I will add docs and more testing to make sure that this doesn't break anything. Tested with keycloak.

Changes proposed in this pull request:

- Fix install issue that stopped auth-type from being set.
- Add oidc support.

Todo

- [ ] Tests
- [ ] Docs

How to test the feature manually:

1.
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #5481, opened 2023-06-18
Allow choosing the OIDC remote user claim and scopes to request from the identity provider

otaconix:

- Allow setting the `OIDC_SCOPES` environment variable to control what scopes to request from the OIDC identity provider. This defaults to `openid`.
- Allow setting the `OIDC_REMOTE_USER_CLAIM` environment to control what claim to use as the username. This defaults to `preferred_username`.
- Document these environment variables.

1. I personally needed this to integrate with Authentik, so configure that, create an OIDC provider and an application.
2. Configure a FreshRSS Debian docker container (I used `Docker/Dockerfile`) to use Authentik as the OIDC identity provider. Use the following environment variables besides the ones needed to get OIDC in FreshRSS:
   - `OIDC_SCOPES=openid profile` (`profile` is required to get the `preferred_username` claim)
4. Open the FreshRSS instance. You should be redirected to Authentik to login, then go back to FreshRSS and be asked to setup the new instance. If it doesn't work, you'll be greeted with some error page. If it works, you can setup the instance.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/Docker/FreshRSS.Apache.conf
+++ b/Docker/FreshRSS.Apache.conf
@@ -10,6 +10,8 @@ AllowEncodedSlashes On
 ServerTokens OS
 TraceEnable Off
 
+# Workaround to be able to check whether an environment variable is set
+# See: https://serverfault.com/questions/1022233/using-ifdefine-with-environment-variables/1022234#1022234
 Define VStart "${"
 Define VEnd "}"
 
--- a/Docker/FreshRSS.Apache.conf
+++ b/Docker/FreshRSS.Apache.conf
@@ -10,6 +10,9 @@ AllowEncodedSlashes On
 ServerTokens OS
 TraceEnable Off
 
+Define VStart "${"
+Define VEnd "}"
+
 <IfDefine OIDC_ENABLED>
 	<IfModule !auth_openidc_module>
 		Error "The auth_openidc_module is not available. Install it or unset environment variable OIDC_ENABLED."
@@ -22,8 +25,20 @@ TraceEnable Off
 	OIDCRedirectURI /i/oidc/
 	OIDCCryptoPassphrase ${OIDC_CLIENT_CRYPTO_KEY}
 
-	OIDCRemoteUserClaim preferred_username
-	OIDCScope "openid"
+	Define "Test_${OIDC_REMOTE_USER_CLAIM}"
+	<IfDefine Test_${VStart}OIDC_REMOTE_USER_CLAIM${VEnd}>
+		OIDCRemoteUserClaim preferred_username
+	</IfDefine>
+	<IfDefine !Test_${VStart}OIDC_REMOTE_USER_CLAIM${VEnd}>
+		OIDCRemoteUserClaim "${OIDC_REMOTE_USER_CLAIM}"
+	</IfDefine>
+	Define "Test_${OIDC_SCOPES}"
+	<IfDefine Test_${VStart}OIDC_SCOPES${VEnd}>
+		OIDCScope openid
+	</IfDefine>
+	<IfDefine !Test_${VStart}OIDC_SCOPES${VEnd}>
+		OIDCScope "${OIDC_SCOPES}"
+	</IfDefine>
 
 	OIDCRefreshAccessTokenBeforeExpiry 30
 </IfDefine>
--- a/Docker/FreshRSS.Apache.conf
+++ b/Docker/FreshRSS.Apache.conf
@@ -10,15 +10,15 @@ AllowEncodedSlashes On
 ServerTokens OS
 TraceEnable Off
 
-# Workaround to be able to check whether an environment variable is set
-# See: https://serverfault.com/questions/1022233/using-ifdefine-with-environment-variables/1022234#1022234
-Define VStart "${"
-Define VEnd "}"
-
 <IfDefine OIDC_ENABLED>
 	<IfModule !auth_openidc_module>
 		Error "The auth_openidc_module is not available. Install it or unset environment variable OIDC_ENABLED."
 	</IfModule>
+
+	# Workaround to be able to check whether an environment variable is set
+	# See: https://serverfault.com/questions/1022233/using-ifdefine-with-environment-variables/1022234#1022234
+	Define VStart "${"
+	Define VEnd "}"
 
 	OIDCProviderMetadataURL ${OIDC_PROVIDER_METADATA_URL}
 	OIDCClientID ${OIDC_CLIENT_ID}
--- a/docs/en/admins/16_OpenID-Connect.md
+++ b/docs/en/admins/16_OpenID-Connect.md
@@ -20,6 +20,8 @@ OIDC support in Docker is activated by the presence of a non-empty non-zero `OID
 * `OIDC_CLIENT_ID`: The OIDC client id from your issuer.
 * `OIDC_CLIENT_SECRET`: The OIDC client secret issuer.
 * `OIDC_CLIENT_CRYPTO_KEY`: An opaque key used for internal encryption.
+* `OIDC_REMOTE_USER_CLAIM`: The claim to use as the username within FreshRSS. Defaults to `preferred_username`. Depending on what you choose here, and your identity provider, you'll need to adjust the scopes you request so that this claim will be accessible. Refer to your identity provider's documentation.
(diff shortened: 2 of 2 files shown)


Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/5351
Ping @aaronschif

Docker/Dockerfile
```diff
@@ -58,6 +58,8 @@ ENV DATA_PATH ''
 ENV FRESHRSS_ENV ''
 ENV LISTEN ''
 ENV OIDC_ENABLED ''
+ENV OIDC_REMOTE_USER_CLAIM 'preferred_username'
+ENV OIDC_SCOPES 'openid'
```
Alkarex: If possible, I would rather put the default values in `FreshRSS.Apache.conf` instead of in each of our Dockerfiles

Docker/FreshRSS.Apache.conf
```diff
@@ -14,6 +14,14 @@ TraceEnable Off
 	<IfModule !auth_openidc_module>
 		Error "The auth_openidc_module is not available. Install it or unset environment variable OIDC_ENABLED."
 	</IfModule>
+	Define "Test_OIDC_REMOTE_USER_CLAIM${OIDC_REMOTE_USER_CLAIM}"
+	<IfDefine Test_OIDC_REMOTE_USER_CLAIM>
```
Alkarex: Ah, interesting. I had to look it up to understand the approach :-)

```suggestion
	# https://serverfault.com/questions/1022233/using-ifdefine-with-environment-variables/1022234#1022234
	Define "Test_OIDC_REMOTE_USER_CLAIM${OIDC_REMOTE_USER_CLAIM}"
	<IfDefine Test_OIDC_REMOTE_USER_CLAIM>
```

otaconix: Yeah, that's where I found it! I'll add a comment. It doesn't yet work though, I'm currently debugging to figure out why :slightly_smiling_face:

docs/en/admins/16_OpenID-Connect.md
```diff
@@ -20,6 +20,8 @@ OIDC support in Docker is activated by the presence of a non-empty non-zero `OID
 * `OIDC_CLIENT_ID`: The OIDC client id from your issuer.
 * `OIDC_CLIENT_SECRET`: The OIDC client secret issuer.
 * `OIDC_CLIENT_CRYPTO_KEY`: An opaque key used for internal encryption.
+* `OIDC_REMOTE_USER_CLAIM`: The claim to use as the username within FreshRSS. Defaults to `preferred_username`. Depending on what you choose here, and your identity provider, you'll need to adjust the scopes you request so that this claim will be accessible. Refer to your identity provider's documentation.
+* `OIDC_SCOPES`: The OIDC scopes to request. Defaults to `openid`. As mentioned previously, make sure the scopes you pick contain whatever `OIDC_REMOTE_USER_CLAIM` you chose.
```
Alkarex: Minor typographic quotes

```suggestion
* `OIDC_REMOTE_USER_CLAIM`: The claim to use as the username within FreshRSS. Defaults to `preferred_username`. Depending on what you choose here, and your identity provider, you’ll need to adjust the scopes you request so that this claim will be accessible. Refer to your identity provider’s documentation.
* `OIDC_SCOPES`: The OIDC scopes to request. Defaults to `openid`. As mentioned previously, make sure the scopes you pick contain whatever `OIDC_REMOTE_USER_CLAIM` you chose.
```

otaconix:
Alright, I cleaned up my commits a little, and finally got something to work for using default values for unset environment variables. The end-result may not be very pretty, if anyone more knowledgeable than me has suggestions to make this better, I'm all ears!

Docker/FreshRSS.Apache.conf
```diff
@@ -10,6 +10,11 @@ AllowEncodedSlashes On
 ServerTokens OS
 TraceEnable Off
 
+# Workaround to be able to check whether an environment variable is set
+# See: https://serverfault.com/questions/1022233/using-ifdefine-with-environment-variables/1022234#1022234
+Define VStart "${"
+Define VEnd "}"
+
```
Alkarex: Can that be put inside the `<IfDefine OIDC_ENABLED>` block?

otaconix: It certainly can. I put those there because I considered using the same trick to get rid of the `-D OIDC_ENABLED` in the `Dockerfile`s, but decided not to so it wouldn't further "pollute" the PR. I can still do either (move these `Define` into the `OIDC_ENABLED` conditional, or use the trick to get rid of the conditional in the `Dockerfile`s). I'll leave that to you.

Alkarex: I have not fully understood your response :-)

If possible, I would suggest the following:

```apache
<IfDefine OIDC_ENABLED>
	<IfModule !auth_openidc_module>
		Error "The auth_openidc_module is not available. Install it or unset environment variable OIDC_ENABLED."
	</IfModule>

	# Workaround to be able to check whether an OS environment variable is set
	# https://serverfault.com/questions/1022233/using-ifdefine-with-environment-variables/1022234#1022234
	Define VStart "${"
	Define VEnd "}"

	OIDCProviderMetadataURL ${OIDC_PROVIDER_METADATA_URL}
	OIDCClientID ${OIDC_CLIENT_ID}
	OIDCClientSecret ${OIDC_CLIENT_SECRET}

	OIDCRedirectURI /i/oidc/
	OIDCCryptoPassphrase ${OIDC_CLIENT_CRYPTO_KEY}

	Define "Test_${OIDC_REMOTE_USER_CLAIM}"
	<IfDefine Test_${VStart}OIDC_REMOTE_USER_CLAIM${VEnd}>
		OIDCRemoteUserClaim preferred_username
	</IfDefine>
	<IfDefine !Test_${VStart}OIDC_REMOTE_USER_CLAIM${VEnd}>
		OIDCRemoteUserClaim "${OIDC_REMOTE_USER_CLAIM}"
	</IfDefine>
	Define "Test_${OIDC_SCOPES}"
	<IfDefine Test_${VStart}OIDC_SCOPES${VEnd}>
		OIDCScope openid
	</IfDefine>
	<IfDefine !Test_${VStart}OIDC_SCOPES${VEnd}>
		OIDCScope "${OIDC_SCOPES}"
	</IfDefine>

	OIDCRefreshAccessTokenBeforeExpiry 30
</IfDefine>
```

otaconix: > I have not fully understood your response :-)

Apologies, I don't know how best to express this without actually showing what I mean in a diff. But I was referring to this: https://github.com/FreshRSS/FreshRSS/blob/666e951fa3e30c3a1b7f5ae68a2be4e06577b75d/Docker/Dockerfile#L68

The `OIDC_ENABLED`  variable is defined on the commandline inside each of the `Dockerfile`s, and it doesn't need to be if we use the same pattern for checking whether an environment variable is set or not.

I applied your suggestion anyway. Let me know if there's anything else :+1:

Alkarex: This `OIDC_ENABLED` variable is special in the sense that, as far as I can see, it is the only one that changes the Apache command line (we want to add a `-D OIDC_ENABLED`), so I do not believe it is the same situation than the other environment variables

Alkarex (approved):

Alkarex:
Looks all good to me now 👍🏻

Alkarex:
@otaconix Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

Docker/FreshRSS.Apache.conf
otaconix: But why do you *want* to add a `-D OIDC_ENABLED`? Instead (and I'm not saying this is winning any beauty contests, but still), it could've been done like so (without the `-D OIDC_ENABLED`):

```
Define VStart "${"
Define VEnd "}"
Define "Test_${OIDC_ENABLED}"
<IfDefine !Test_${VStart}OIDC_ENABLED${VEnd}>
   # [...]
</IfDefine>
```

IMO, Apache's config doesn't lend itself well to this kind of runtime logic. Maybe some kind of templating solution would be better, but then you'd have one more moving piece in your docker image. I'm not sure what the best solution would be, honestly. What you have now works fine, so just keep that if you want :slightly_smiling_face:

PR #5489, opened 2023-06-23
improved: add timeout of "load more articles"

math-GH:
I had very slow internet yesterday. So it was very difficult to load more articles and I saw very often for a long time the "loading circle" at the bottom of the articles list. I did not knew whether it is still loading or already failed.

- add a timeout of 30 seconds to load the next articles
- if timeout is reached: show the "Load more articles" button again (instead of the loading circle gif)

1. block your internet connection or slow it down (f.e. with the Firefox browser tool that tests the screen resolution)
2. scroll down the articles list till the next articles should load
3. wait for timeout
4. the loading circle will disappear and the loading button will be shown

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
The question is: is a config option needed here to change the 30 seconds to another value?

Side task: we should check all Ajax call for having a timeout

Frenzie:
Do you think there's much to be gained from a value higher than about 10?

p/scripts/main.js
```diff
@@ -1697,6 +1697,11 @@ function load_more_posts() {
 	const req = new XMLHttpRequest();
 	req.open('GET', url_load_more, true);
 	req.responseType = 'document';
+	req.timeout = 30000;
+	req.ontimeout = function (e) {
+		document.getElementById('load_more').classList.remove('loading');
+		load_more = false;
+	}
```
```suggestion
	};
```

Frenzie (approved):

Frenzie:
I take that back, I did some quick experiments and 10 seconds might be cutting it rather close or even be premature in some (presumably) edge cases. Though upping what I said to 15 or 20 seconds wouldn't change much about the thought behind it. :-)

math-GH:
I did the same research with the same result in this moment.
With throttled internet connection down to GPRS/G1. It fetched 10 articles in about 8-13 seconds.
Let's got with 20 seconds and will see if it is fine.
It should be "only" a fall back in cases where it does not make sense anymore to wait for an response.

Alkarex:
Hum, I would like a bit more testing for that.
Browsers are supposed to have their own detection when a request fail.
Furthermore, a FreshRSS instance might itself be slow to answer, in particular for a complicated search query.

Frenzie:
Good point, it looks like there's no error handler as I had assumed? So logically it'd make more sense to forget about the timeout (at least initially) and change that `ontimeout` to `onerror`.

math-GH:
thanks for the discussion.
I added `onerror` but kept `ontimeout` to have it as fallback.
For user queries, I added another timout (45seconds. Default: 20seconds).

What do you think?

p/scripts/main.js
```diff
@@ -1692,11 +1692,31 @@ function load_more_posts() {
...
 	const req = new XMLHttpRequest();
 	req.open('GET', url_load_more, true);
 	req.responseType = 'document';
+	if (url_load_more.includes('&search=')) {
+		// user queries could run a bit longer than usual
+		req.timeout = 45000;
+	} else {
+		// default: wait max 20 seconds
+		req.timeout = 20000;
+	}
```
ColonelMoutarde:
```suggestion
	// default: wait max 20 seconds
	req.timeout = 20000;
	if (url_load_more.includes('&search=')) {
		// user queries could run a bit longer than usual
		req.timeout = 45000;
	}
```

ColonelMoutarde: the else is not required

Alkarex:
Isn't the `onerror` addressing the problem you faced @math-GH even without the explicit timeout?
Otherwise regarding the timeout, I am afraid of putting so low values (I would say 1 minute for basic requests).
I have not tested the PR yet, but when we decide that the request has failed, there should be a way for the user to do something, for instance manually click on the "load more" link

math-GH:
> Isn't the `onerror` addressing the problem you faced @math-GH even without the explicit timeout? Otherwise regarding the timeout, I am afraid of putting so low values (I would say 1 minute for basic requests). I have not tested the PR yet, but when we decide that the request has failed, there should be a way for the user to do something, for instance manually click on the "load more" link

Before the PR:
The spinning icon (to highlight the loading) will displayed forever, if there is an error

After the PR:
The spinning icon will change back to the "load more" button, if (1) an error appears or (2) the time is out. The user can click the "load more" button to restart the Ajax request.

Waiting and starring on a spinning icon for more than 30 seconds feels like waiting forever.

Alkarex:
But my question still stands: in the cases you faced (if you can reproduce them), isn't the `onerror` catching the cases when the request failed?
There are legitimate cases for very long requests, for instance on very slow networks and/or with large responses and/or with slow database requests, for which it does not help to re-issue the request and the best that can be done is wait

math-GH:
I cannot reproduce the very slow internet connection that I had.

I tried the following to enforce an network error: I blocked the domain in my development tool (in Firefox). But it does not enforce the onerror event.

grafik

As far as I understand, the onerror event will be triggert if the http answer is not a 2xx. But if there is no answer there will be no onerror trigger.

What can we do to finish this PR?

Frenzie:
If XHR doesn't do what we want fetch could be considered instead.
```js
fetch('https://developer.mozilla.org/')
  .then(function(response) {
    if (!response.ok) {
      throw new Error('Network response was not ok');
    }
    return response.json();
  })
  .then(function(data) {
    console.log(data);
  })
  .catch(function(error) {
    console.log('Error:', error);
  });
```
The catch should normally do the trick if it doesn't ever reach a response state.

Alkarex:
> As far as I understand, the onerror event will be triggert if the http answer is not a 2xx. But if there is no answer there will be no onerror trigger.

Hum, that is annoying... It would be nice to test the Fetch API indeed.

> What can we do to finish this PR?

I am afraid adding a hard timeout will lead to more problems than it solves. So maybe something like enabling the UI to re-issue the request after some time but without automatically killing the current one (which might still be in progress). Basically what you already have, without automatically killing the current request until the user clicks (and maybe tuning the UI to keep having the spinner).

Alkarex:
> So maybe something like enabling the UI to re-issue the request after some time but without automatically killing the current one (which might still be in progress). Basically what you already have, without automatically killing the current request until the user clicks (and maybe tuning the UI to keep having the spinner).

@math-GH What do you think?

PR #5501, opened 2023-06-29
fix many "Only booleans are allowed in an if condition"

ColonelMoutarde:

- By resetting the "booleansInConditions" directive to true, 370 errors are detected. I've corrected 20 of the errors that phpstan is reporting. There are still many...

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


cli/create-user.php
```diff
@@ -10,7 +10,7 @@
 }
 
 $usernames = listUsers();
-if (preg_grep("/^$username$/i", $usernames)) {
+if (is_bool(preg_grep("/^$username$/i", $usernames))) {
```
Alkarex: I would prefer:

```suggestion
if (preg_grep("/^$username$/i", $usernames) !== false) {
```

and likewise for other similar cases below

cli/i18n/I18nUsageValidator.php
```diff
@@ -42,7 +42,7 @@ public function validate(): bool {
 		foreach ($this->reference as $file => $data) {
 			foreach ($data as $key => $value) {
 				$this->totalEntries++;
-				if (preg_match('/\._$/', $key) && in_array(preg_replace('/\._$/', '', $key), $this->code, true)) {
+				if (is_int(preg_match('/\._$/', $key)) && in_array(preg_replace('/\._$/', '', $key), $this->code, true)) {
```
Alkarex: I would prefer something like the following (or `!= false`)

```suggestion
				if (preg_match('/\._$/', $key) === 1 && in_array(preg_replace('/\._$/', '', $key), $this->code, true)) {
```

p/api/greader.php
```diff
@@ -1069,7 +1069,7 @@ public static function parse() {
 								$include_target = $pathInfos[7];
 								if ($include_target != '' && !ctype_digit($include_target)) {
 									$include_target = empty($_SERVER['REQUEST_URI']) ? '' : $_SERVER['REQUEST_URI'];
-									if (preg_match('#/reader/api/0/stream/contents/feed/([A-Za-z0-9\'!*()%$_.~+-]+)#', $include_target, $matches)) {
+									if (preg_match('#/reader/api/0/stream/contents/feed/([A-Za-z0-9\'!*()%$_.~+-]+)#', $include_target, $matches) === false) {
```
Alkarex: This looks wrong

lib/Minz/ModelPdo.php
```diff
@@ -136,14 +141,13 @@ public function __construct(?string $currentUser = null, ?Minz_Pdo $currentPdo =
...
+			$db = Minz_Configuration::get('system')->db;
 
-		$db = Minz_Configuration::get('system')->db;
-
-		throw new Minz_PDOConnectionException(
+			throw new Minz_PDOConnectionException(
 				$ex->getMessage(),
 				$db['user'], Minz_Exception::ERROR
 			);
+		}
```
Alkarex: What happened here? This looks quite wrong

lib/Minz/Pdo.php
```diff
@@ -28,7 +28,7 @@ private function autoPrefix(string $sql): string {
 	}
 
 	protected function preSql(string $statement): string {
-		if (preg_match('/^(?:UPDATE|INSERT|DELETE)/i', $statement)) {
+		if (is_int(preg_match('/^(?:UPDATE|INSERT|DELETE)/i', $statement))) {
```
Alkarex: That is wrong for the return case of `0`

lib/Minz/Request.php
```diff
@@ -329,7 +331,7 @@ public static function serverIsPublic(string $address): bool {
 
 	private static function requestId(): string {
 		if (empty($_GET['rid']) || !ctype_xdigit($_GET['rid'])) {
-			$_GET['rid'] = uniqid();
+			$_GET['rid'] = uniqid('', true);
```
Alkarex: No need to waste entropy there

lib/lib_rss.php
```diff
@@ -116,14 +116,14 @@ function checkUrl(string $url, bool $fixScheme = true) {
 	if ($url == '') {
 		return '';
 	}
-	if ($fixScheme && !preg_match('#^https?://#i', $url)) {
+	if ($fixScheme && preg_match('#^https?://#i', $url) === false) {
```
Alkarex: Wrong, we mostly need to test for `0` here

lib/favicons.php
```diff
@@ -119,6 +121,7 @@ function download_favicon(string $url, string $dest): bool {
 			}
 		}
 	}
-	return ($favicon != '' && file_put_contents($dest, $favicon)) ||
+	$filePutContent = file_put_contents($dest, $favicon);
```
Alkarex: No, wrong order of logic: we should not write if `$favicon == ''`...

Alkarex:
I have fixed several regressions and other smaller details

Alkarex:
Let's not hurry making more changes of this type, especially because I believe that it leads to less readable and more error-prone code in many situations, for instance the calls to `preg_grep()`, `preg_match()`. Indeed, this original PR had several regressions, which I am guessing were at least partially due to the more complicated / less readable syntax.
I think we should relax the `booleansInConditions` rule to accept { 0, 1 } (and maybe null) in addition of { true, false }.
https://github.com/phpstan/phpstan-strict-rules/issues/221

lib/http-conditional.php
```diff
@@ -182,7 +182,7 @@ function httpConditional(int $UnixTimeStamp, int $cacheSeconds = 0, int $cachePr
  * Reference rfc2616-sec14.html#sec14.11
  */
 function _httpConditionalCallBack(string $buffer, int $mode = 5): string {
-	if (extension_loaded('zlib') && (!ini_get('zlib.output_compression'))) {
+	if (extension_loaded('zlib') && (ini_get('zlib.output_compression') === false)) {
```
Alkarex: I think this is wrong as well:
* https://php.net/function.ini-get
* https://php.net/zlib.configuration#ini.zlib.output-compression

```suggestion
	if (extension_loaded('zlib') && (ini_get('zlib.output_compression') == false)) {
```

ColonelMoutarde:
> Let's not hurry making more changes of this type, especially because I believe that it leads to less readable and more error-prone code in many situations, for instance the calls to `preg_grep()`, `preg_match()`. Indeed, this original PR had several regressions, which I am guessing were at least partially due to the more complicated / less readable syntax. I think we should relax the `booleansInConditions` rule to accept { 0, 1 } (and maybe null) in addition of { true, false }. [phpstan/phpstan-strict-rules#221](https://github.com/phpstan/phpstan-strict-rules/issues/221)

hello, the problem is polymorphism. it must be reduced as much as possible. a function must return only one type. 

See : 
https://www.phptutorial.net/php-oop/php-polymorphism/

if a native php function is polymorphic (for historical reasons), the best thing to do is to create a method that encapsulates and returns only what is necessary.

lib/Minz/ModelPdo.php
ColonelMoutarde: if I don't do this phpstan returns an error because $ex is unknown at this point

Alkarex: I have reverted without problem, as far as I can see. But the point was that the logic was broken with your changes.

PR #5531, opened 2023-07-17
Docker: Add DATA_PATH to cron env

zhaofengli:

- Adds `$DATA_PATH` to the cron environment

This fixes automatic feed updating with non-standard data paths.

1. Set a non-standard `DATA_PATH` and enable automatic feed updating with `CRON_MIN`
2. `cat /tmp/FreshRSS.log`
3. See successful result instead of `Failed requirements!`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -21,6 +21,7 @@ if [ -n "$CRON_MIN" ]; then
 		echo "export COPY_LOG_TO_SYSLOG=$COPY_LOG_TO_SYSLOG"
 		echo "export COPY_SYSLOG_TO_STDERR=$COPY_SYSLOG_TO_STDERR"
 		echo "export FRESHRSS_ENV=$FRESHRSS_ENV"
+		echo "export DATA_PATH=$DATA_PATH"
 	) >/var/www/FreshRSS/Docker/env.txt
 	sed </etc/crontab.freshrss.default \
 		-r "s#^[^ ]+ #$CRON_MIN #" | crontab -


ColonelMoutarde (approved):

Alkarex:
Hello,
It probably does not hurt to pass this variable, so let's do it. However, the expected way in Docker is to map the data path to a volume, which can be where you want, so I am a bit unsure of your use-case. Could you provide a bit more info?
https://github.com/FreshRSS/FreshRSS/tree/edge/Docker#docker-compose

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/pull/5423

zhaofengli:
> However, the expected way in Docker is to map the data path to a volume, which can be where you want, so I am a bit unsure of your use-case.

Sorry for the late reply. My use-case is to deploy FreshRSS on Kubernetes with [Litestream](https://github.com/benbjohnson/litestream)[^1]. Litestream runs as a sidecar container with a shared ephemeral directory mounted at `/data`. I could have used `/var/www/FreshRSS/data`, but `DATA_PATH` exists and I want to use a path consistent with other apps I deploy.

> Please add a line for you in [edge/CREDITS.md](https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md?rgh-link-date=2023-07-19T21%3A33%3A06Z)

Thanks, I'll add it the next time I contribute :smile:

[^1]: I also patch FreshRSS downstream to use a single SQLite database with table prefixes, but this is obviously too cursed to be upstreamed.

Alkarex:
Thanks for the explanations.

> Thanks, I'll add it the next time I contribute

You can just make a new PR for that

PR #3226, opened 2020-10-14
Docker: Use Apache remoteip

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3224
Log the client remote IP instead of the local proxy IP
https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html

PR #5549, opened 2023-07-26
Rework trusted proxies

### LINKED ISSUE #5502: [BUG] Mishandling of X-Forwarded-For ?
**Describe the bug**
Using nginx as proxy: When setting X-Forwarded-For as described in the docker readme, the REMOTE_ADDR in PHP gets set to the value being delivered by the forwarded for.
This renders the `trusted_sources` config unusable, as not the IP of the proxy server is checked, but of the client.
You'll then get a 403 Error when trying to connect.

**To Reproduce**

1. Using the current freshrss/freshrss:latest and nginx:lates docker image using the recommended config from the freshrss readme on docker.
2. Put `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;` in your nginx config
3. Get a phpinfo and check the REMOTE_ADDR

**Expected behavior**
Real REMOTE_ADDR  (i.e. container IP) should always be preferred to X-Forwarded-For ones.

**Additional context**
Might this be an issue like described here?
https://blog.ircmaxell.com/2012/11/anatomy-of-attack-how-i-hacked.html

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/5502
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/3226

New environment variable `TRUSTED_PROXY`: set to 0 to disable, or to a list of trusted IP ranges compatible with https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteiptrustedproxy

New internal environment variable `CONN_REMOTE_ADDR` to remember the true IP address of the connection (e.g. last proxy), even when using mod_remoteip.

Current working setups should not observe any significant change.


Alkarex:
To ease testing, I have pushed a Docker image `freshrss/freshrss:pr5549`

d-k-bo:
From what I can tell, the image works fine on my system, using Caddy's `reverse_proxy` directive without any additional options. I can access the server from an external network and the log still shows my real IP address.

Thank your for working on this!

lib/lib_rss.php
```diff
@@ -653,21 +653,43 @@ function checkCIDR(string $ip, string $range): bool {
...
+		return false;
+	}
+	$trusted = getenv('TRUSTED_PROXY');
+	if ($trusted != 0 && is_string($trusted)) {
+		$trusted = preg_split('/\s+/', $trusted, -1, PREG_SPLIT_NO_EMPTY);
+	}
+	if (empty($trusted)) {
+		$trusted = FreshRSS_Context::$system_conf->trusted_sources;
+	}
+	foreach (FreshRSS_Context::$system_conf->trusted_sources as $cidr) {
```
joestump, line 690: @Alkarex I've been trying to get this working with Authentik and staring at this logic for a bit. Am I crazy or does this completely override the `$trusted` logic on line 684? 

On line 684 `$trusted` gets set, then if it's `empty()` on line 687 it gets overridden with `FreshRSS_Context::$system_conf->trusted_sources`, but is ignored for the rest of the function as it defers to `FreshRSS_Context::$system_conf->trusted_sources` on this line.

I currently have `TRUSTED_PROXY` set to `192.168.1.34/32`, `$trusted` gets set, and then this falls through to `false` on line 695.

joestump (commented):
@Alkarex let me know if you'd like me to file a separate issue on this.

lib/lib_rss.php
Alkarex: Indeed, this should probably be something like:

```suggestion
	foreach ($trusted as $cidr) {
```

Do you feel like sending a PR?

joestump: @Alkarex that did the trick: https://github.com/FreshRSS/FreshRSS/pull/5853

PR #5570, opened 2023-08-06
Ready for year 2038

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/discussions/5569
Requires PHP on a 64-bit platform to take advantage of it.
https://en.wikipedia.org/wiki/Year_2038_problem
Note: at the first refresh for a user, it will take some time to alter the column types, but meanwhile, the other users are not blocked.


Frenzie (approved):
I'm somewhat surprised PostgreSQL was different from the others here.

ColonelMoutarde (approved):

PR #5572, opened 2023-08-07
fix long transactions blocking mysql

tqcq:
- The long transaction with more than 3 second is submitted automatically.

1. Add a time-consuming operation in the hook entry_before_insert
2. Import a feed with 3000+ articles
3. Use the command show processlist; to check the database for blocked operations.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -433,6 +433,9 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				/** @var array<string,bool> */
 				$newGuids = [];
 
+
+				$transactionStartTime = microtime(true);
+
 				// Add entries in database if possible.
 				/** @var FreshRSS_Entry $entry */
 				foreach ($entries as $entry) {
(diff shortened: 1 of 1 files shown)


Alkarex:
Hello,
Could you please explain the problem you observe more precisely?
This is with MySQL only, right? (Tests with SQLite and PostgreSQL welcome). And what is blocked during the process more precisely (e.g. a specific table)? And with what consequence (e.g. that specific user cannot access FreshRSS)?
I believe your suggested approach might create other problems, but we can try to find another way to fix. In particular, we have an intermediate `entrytmp` table (which is OK to block) before updating the `entry` table (for which the transaction should be fast).

Alkarex:
P.S. Please make sure you use test using the `edge` branch of FreshRSS

tqcq:
> Hello, Could you please explain the problem you observe more precisely? This is with MySQL only, right? (Tests with SQLite and PostgreSQL welcome). And what is blocked during the process more precisely (e.g. a specific table)? And with what consequence (e.g. that specific user cannot access FreshRSS)? I believe your suggested approach might create other problems, but we can try to find another way to fix. In particular, we have an intermediate `entrytmp` table (which is OK to block) before updating the `entry` table (for which the transaction should be fast).

When I enable the full-text crawling plugin, if the number of my articles reaches thousands, this transaction will last for hours, causing the _feed table to be locked.

I don't know about other databases and whether they lock after opening a transaction.

In addition, I want to know whether the edge branch is the current master branch?

tqcq:
@Alkarex I have re-created a request about the edge branch
#5574

PR #5576, opened 2023-08-09
Fix hash of articles with loadCompleteContent

Alkarex:
The detection of modified articles was wrong for feeds using loadCompleteContent. Indeed, the hash is supposed to be computed on the content provided by the server of the RSS feed, excluding further modifications.
Furthermore, read hash from database instead of recomputing it all the time.
Slightly related to https://github.com/FreshRSS/FreshRSS/pull/5574


Alkarex:
Ping @aledeg if you are around and if you have any comment

ColonelMoutarde (approved):

Alkarex:
Feedback & tests still welcome

PR #5580, opened 2023-08-10
fix: Fix the "Show all articles" option

### LINKED ISSUE #5579: [BUG] The option "Show all articles" acts as "Show only unread"
**Describe the bug**

In Configuration > Reading > Articles to display, when I set "Show all articles", it seems that only unread articles are displayed when I go back to the main stream.

**To Reproduce**

1. Go to Configuration > Reading
2. Choose "Show all articles" for the option "Articles to display"
3. Go to the main stream
4. Make sure you have read articles

**Expected behavior**

The read articles should be displayed while they are not.
This is the expected behaviour for "Show only unread" option, not "Show all articles"

- Device: PC
- OS: Fedora 38
- Browser: Firefox 117
- FreshRSS version: edge
- Database version: SQLite
- PHP version: PHP 8.1
- Installation type: Docker (dev env)

**Additional context**

~~Things changed a lot since I worked on this part, so I'm not sure how to fix that.
What I can tell is that it seems that the `normal` value of `default_view` is not handled, [only `adaptive` is](https://github.com/FreshRSS/FreshRSS/blob/edge/app/Models/Context.php#L211-L213).~~
Ok, that wasn't too complicated :D

On a related subject, someone [reported on Mattermost](https://framateam.org/freshrss/pl/zmm9mjquk3dyzyjsfcpaihxzyh) almost the same issue, except that even "read" articles don't show up.
In his case, anonymous users cannot see any article.

marienfressinaud:
Closes https://github.com/FreshRSS/FreshRSS/issues/5579

- Add `FreshRSS_Entry::STATE_ALL` to the current state when `default_view === 'all'`

I'm not too sure about this patch. I've checked and it was working fine in 1.21, but this code didn't change that much between these two versions. So it seems the issue was introduced elsewhere.

1. Go to Configuration > Reading
2. Choose "Show all articles" for the option "Articles to display"
3. Go to the main stream
4. Check that read and unread articles are displayed

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] ~~unit tests written (optional if too hard)~~
- [x] ~~documentation updated~~

### DIFF
--- a/app/Models/Context.php
+++ b/app/Models/Context.php
@@ -208,7 +208,9 @@ final class FreshRSS_Context {
 		self::$state = Minz_Request::paramInt('state') ?: self::$user_conf->default_state;
 		$state_forced_by_user = Minz_Request::paramString('state') !== '';
 		if (!$state_forced_by_user && !self::isStateEnabled(FreshRSS_Entry::STATE_READ)) {
-			if (self::$user_conf->default_view === 'adaptive' && self::$get_unread <= 0) {
+			if (self::$user_conf->default_view === 'all') {
+				self::$state |= FreshRSS_Entry::STATE_ALL;
+			} elseif (self::$user_conf->default_view === 'adaptive' && self::$get_unread <= 0) {
 				self::$state |= FreshRSS_Entry::STATE_READ;
 			}
 			if (self::$user_conf->show_fav_unread &&


Alkarex (approved):

Alkarex:
I have only quickly checked but it looks like this bug has been there for a long time (i.e. for many versions).
It was not working for me in 1.21.0 (nor in 1.20.0 nor in several older versions).

marienfressinaud:
Strange, it definitely works for me in 1.21 :thinking:

Alkarex:
You are right: it is because it required re-applying the settings for testing the behaviour correctly.
I believe it is a regression from https://github.com/FreshRSS/FreshRSS/pull/5251

Alkarex:
https://github.com/FreshRSS/FreshRSS/blob/0fe0ce894cbad09757d719dd4b400b9862c1a12a/app/Models/ConfigurationSetter.php#L55-L67

PR #5582, opened 2023-08-10
Default dark mode to auto

### LINKED ISSUE #5659: [Feature] Dark theme for login page (auto switch to light with system)
When I use FreshRSS in the night it would be good to have also a dark login site (site which appears when going to the FreshRSS-instance before I'm logged in) instead of the white login site.

Best would be the possibility to select an theme that automatically switch between light/dark with the system settings.

Alkarex:
Set default dark mode to automatic instead of disabled.
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/4843


math-GH:
I am not so happy with it, because only the Origine themes have an automatic dark mode.

Alkarex:
Well, at least it would work by default with the Origine theme, which is also the default theme.
Would it help to mention in each theme whether they are compatible with the Light/Dark modes?

Alkarex:
> Would it help to mention in each theme whether they are compatible with the Light/Dark modes?

Ping @math-GH

math-GH:
I have this PR still on my list.

ColonelMoutarde (approved):

Alkarex:
Ping @math-GH ;-)

math-GH:
> Would it help to mention in each theme whether they are compatible with the Light/Dark modes?

I think: yes :D

Alkarex:
image

app/i18n/en/conf.php
```diff
@@ -33,8 +33,9 @@
 	'display' => array(
 		'_' => 'Display',
 		'darkMode' => array(
-			'_' => 'Automatic dark mode (beta)',
+			'_' => 'Automatic dark mode',
 			'auto' => 'Auto',
+			'help' => 'For selected themes only',
```
math-GH: Would be "For subset themes only" better? There is always a theme selected (!) above.
(I mean that it would be better to have another word than "selected")

Alkarex: I think that it is the common way to say it. Maybe an English native can help there.
https://ludwig.guru/s/selected+items+only

Alkarex: *some* would also be fine for me

Frenzie: Or specific.

Alkarex: *For compatible themes only*
Better?

Frenzie: Works for me. 👍

PR #5594, opened 2023-08-16
Add xdebug for devs

ColonelMoutarde:

- add xdebug in project only for developpeurs

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


.devcontainer/Dockerfile
```diff
@@ -1,11 +1,13 @@
-FROM alpine:3.18
+FROM alpine:3.18 as dev
```
Alkarex: Still a single-stage build, right? So no need to have a name

.devcontainer/Dockerfile
```diff
@@ -1,11 +1,13 @@
...
 
 ENV TZ UTC
 SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 
 RUN apk add --no-cache \
 	tzdata \
 	apache2 php-apache2 \
+    && pecl install xdebug-3.2.2 \
+    && docker-php-ext-enable xdebug \
```
Alkarex, line 10: That does not build

Alkarex (changes requested):
Does not work for multiple reasons. See comments

tests/app/Services/ImportServiceTest.php
Alkarex, line 1: That does not seem to be related to this PR, is it? And fails the tests

ColonelMoutarde: Indeed, I don't understand what's going on.

PR #5606, opened 2023-08-22
Fix: sharing via clipboard for no https/localhost environments

### LINKED ISSUE #5595: [BUG] Copy to clipboard Sharing does not function
**Describe the bug**
The Clipboard feature in Sharing does not function.

**To Reproduce**
1. Go to Configuration > Sharing
2. Add a sharing method > [Clipboard](https://freshrss.github.io/FreshRSS/en/users/08_sharing_services.html)
3. Open an article
4. Click Share > Clipboard
5. Clipboard contents hasn't changed

**Expected behavior**
Clicking Share > Clipboard copies article URL to system clipboard

- Device: Macbook M1
- OS: macOS Ventura 13.4.1
- Browser: Firefox 116.0.2 [64-bit]
- FreshRSS version: 1.21.0 
- Installation type: Docker-compose on macOS Ventura Apple Silicon

Firefox is able to copy from other webpages including self-hoted webapps.

math-GH:
Closes #5595

- add a fallback if environment is not https or localhost

use the sharing option "Clipboard" ousite of https or localhost environment

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


p/scripts/main.js
```diff
@@ -1087,8 +1087,19 @@ function init_stream(stream) {
 		}
 
 		el = ev.target.closest('.item.share > button[data-type="clipboard"]');
-		if (el && navigator.clipboard) {	// Clipboard
-			navigator.clipboard.writeText(el.dataset.url);
+		if (el) { // Clipboard
+			if (navigator.clipboard) {
+				navigator.clipboard.writeText(el.dataset.url);
+			} else {
```
Alkarex, line 1100:
```suggestion
			} else if (document.execCommand) {
```

`document.execCommand` seems deprecated https://developer.mozilla.org/docs/Web/API/document/execCommand

math-GH: yes it is deprecated. But how could be the user informed, if it is not available? To have a "Clipboard" link without any positive result is not fine.

Frenzie: Yes, this PR is simply what I said in https://github.com/FreshRSS/FreshRSS/issues/5595#issuecomment-1685238478

The scare phrasing on MDN doesn't quite reflect reality though. Most of what it does doesn't have a replacement yet (though copy does) and bugs are still being actively fixed in all browsers.

p/scripts/main.js
```diff
@@ -1087,8 +1087,19 @@ function init_stream(stream) {
...
+				// fallback, if navigator.clipboard is not available f.e. if access is not via https or localhost
+				const inputElement = document.createElement('input');
+				inputElement.value = el.dataset.url;
+				inputElement.type = 'text';
```
Frenzie: This line isn't necessary, so I'd just leave it out.
```suggestion
```

Frenzie (approved):
lgtm

p/scripts/main.js
Alkarex: > But how could be the user informed, if it is not available

Maybe simply an `alert()`? In any case, it is better to test than having a JavaScript error.

Frenzie: As an aside, if you're worried about stuff like that note that the `navigator.clipboard.writeText()` promise can fail and that `document.execCommand('copy')` can return false.

p/scripts/main.js
```diff
@@ -1087,8 +1087,19 @@ function init_stream(stream) {
 		}
 
 		el = ev.target.closest('.item.share > button[data-type="clipboard"]');
-		if (el && navigator.clipboard) {	// Clipboard
-			navigator.clipboard.writeText(el.dataset.url);
+		if (el) { // Clipboard
+			if (navigator.clipboard) {
+				navigator.clipboard.writeText(el.dataset.url);
```
Frenzie: With reference to https://github.com/FreshRSS/FreshRSS/pull/5606/files#r1302688688
```suggestion
				navigator.clipboard.writeText(el.dataset.url)
					.catch(e => alert(e));
```

p/scripts/main.js
Frenzie: @Alkarex 
> In any case, it is better to test than having a JavaScript error.

What JS error are you referring to? It's already testing if document.execCommand exists. If copy fails it returns false.

p/scripts/main.js
Frenzie: PS As far as I'm concerned none of this stuff should be part of this PR, mind you. Just some potential future thoughts should you wish to look into the matter further.

I also noticed there doesn't really seem to be any feedback, so ideally it'd look more like this.

```suggestion
				navigator.clipboard.writeText(el.dataset.url)
					.then(() => {
						// UI feedback
					})
					.catch(e => alert(e)); // needn't be an alert really, could also use a similar mechanism as the positive feedback above
```

p/scripts/main.js
Alkarex: > What JS error are you referring to? It's already testing if document.execCommand exists.

The `if (document.execCommand)` is my suggestion

Frenzie: Oops, indeed it is.

Though while I do think it's good practice, it exist on basically any browser from the past 20 years, and there's pretty much no way they can actually drop it either. If it's gone anytime before the next decade I'll, uh, buy you a coffee or something (I don't want to eat my hat on the off chance it does happen). ;-D

p/scripts/main.js
math-GH: I like it to make the type of an `<input>` implicit. But you are right too. It works without this line.

Frenzie (approved):
Like I said I didn't think it was necessary for this PR, but still lgtm. ;-)

p/scripts/main.js
math-GH: Thanks. I took it into the code. Please have a look

math-GH:
> Like I said I didn't think it was necessary for this PR, but still lgtm. ;-)

I used it to learn new stuff ;)

p/scripts/main.js
```diff
@@ -1089,19 +1089,29 @@ function init_stream(stream) {
...
+						toggleClass(el, 'error');
+					});
 			} else {
 				// fallback, if navigator.clipboard is not available f.e. if access is not via https or localhost
 				const inputElement = document.createElement('input');
 				inputElement.value = el.dataset.url;
 				document.body.appendChild(inputElement);
 				inputElement.select();
-				document.execCommand('copy');
+				if (document.execCommand('copy')) {
```
```suggestion
				if (document.execCommand && document.execCommand('copy')) {
```

To account for the fact that `document.execCommand` is obsoleted. But as Frenzie wrote, not super important

PR #5612, opened 2023-08-26
Fix "server cannot write in /var/www/FreshRSS/p/api"

hyperreal64:

- Fixes "server cannot write in /var/www/FreshRSS/p/api" error in logs when using Docker/Podman.

1. Build a new Docker image from the Dockerfile in the repository.
2. Setup a FreshRSS Docker/Podman instance as per the manual.
3. Install an extension (e.g. TTRSS API) by extracting the archive into the `freshrss_extensions` Docker volume.
4. Restart the container.
5. Open the web UI (via reverse proxy or what have you) and try to enable the extension in Settings > Extensions > System Extensions by flipping the switch for the extension.
6. Check the logs. If the logs say "server cannot write in /var/www/FreshRSS/p/api", then the test has failed.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

No documentation seems to need to be updated for this change.

### DIFF
--- a/cli/access-permissions.sh
+++ b/cli/access-permissions.sh
@@ -14,8 +14,8 @@ fi
 # Based on group access
 chown -R :www-data .
 
-# Read files, and directory traversal
-chmod -R g+rX .
+# Read files, write files, and directory traversal
+chmod -R g+rwX .
 
 # Write access
 mkdir -p ./data/users/_/


Alkarex:
This extension is not really ready, even with the access rights fixed.
In any case, we should aim at having only `./data/` with write access, the rest being read-only (except for Web auto-updates)

Alkarex:
Related https://github.com/FreshRSS/FreshRSS/pull/1526

hyperreal64:
> This extension is not really ready, even with the access rights fixed. In any case, we should aim at having only `./data/` with write access, the rest being read-only (except for Web auto-updates)

Fair enough, my bad.

It seems like the issue is more that the TTRSS extension attempts to write to /p/api.

PR #5614, opened 2023-08-28
Create feverUtilTest

ColonelMoutarde:
Create feverUtilTest

- Create minium tests for Feverutil::class

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/tests/app/Utils/feverUtilTest
+++ b/tests/app/Utils/feverUtilTest
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+class feverUtilTest extends TestCase
+{
+	public function testDeleteKey(): void
+	{
+		$bool = FreshRSS_fever_Util::checkFeverPath();
+		$this->assertIsBool($bool);
+		$this->assertTrue($bool);
+	}
+
+	public function testKeyPathExpectContextException(): void
+	{
+		$this->expectException(FreshRSS_Context_Exception::class);
+		FreshRSS_fever_Util::getKeyPath('foo');
+	}
+
+	public function testUpdateKeyExpectContextException(): void
+	{
+		$this->expectException(FreshRSS_Context_Exception::class);
+		$this->expectExceptionMessage('System configuration not initialised!');
+		FreshRSS_fever_Util::updateKey('foo', 'bar');
+	}
+
+}


Alkarex:
Good initiative to introduce some tests 👍🏻  We do need some more :-)
However, I would much more welcome some tests that contribute to ensuring that the system is working / performing as expected by users or APIs (for instance, returning a valid response without any exception).
The tests in the current PR so far are focussed on internal details such as the type of internal exceptions, which I believe are more negative than positive, as they make internal changes more work demanding.

ColonelMoutarde:
> Good initiative to introduce some tests 👍🏻 We do need some more :-)
> However, I would much more welcome some tests that contribute to ensuring that the system is working / performing as expected by users or APIs (for instance, returning a valid response without any exception).
> The tests in the current PR so far are focussed on internal details such as the type of internal exceptions, which I believe are more negative than positive, as they make internal changes more work demanding.

Hi, I disagree.
You shouldn't oppose unit tests (which is what I did in this pr) and the functional tests you describe above. The two are complementary.

Here's a representation of the test pyramid: 

https://martinfowler.com/articles/practical-test-pyramid.html

What you describe corresponds to business tests with interactions, also known as integration tests.

You need a solid foundation of unit tests, to which you add layers of other types of testing.
Both are important, and both also entail unavoidable maintenance costs. that's the price of peace of mind.

tests/app/Utils/feverUtilTest
```diff
@@ -0,0 +1,29 @@
...
+	{
+		$this->expectException(FreshRSS_Context_Exception::class);
+		FreshRSS_fever_Util::getKeyPath('foo');
+	}
+
+	public function testUpdateKeyExpectContextException(): void
+	{
+		$this->expectException(FreshRSS_Context_Exception::class);
+		$this->expectExceptionMessage('System configuration not initialised!');
+		FreshRSS_fever_Util::updateKey('foo', 'bar');
```
Alkarex, line 26: To be more precise, this is an example of test, which I find more negative than positive.
It relies on some internal behaviour, which happens to be true at the moment, but which is not meant to be stable.
There is no value (or worse, a negative value), in testing the precise exception and exception message for internal behaviour.

For examples of positive unit tests in my book, see e.g. https://github.com/FreshRSS/FreshRSS/blob/edge/tests/app/Models/SearchTest.php , https://github.com/FreshRSS/FreshRSS/blob/edge/tests/app/Utils/passwordUtilTest.php

ColonelMoutarde: Tests need to be maintained over the lifetime of the application. I wasn't able to test the behavior where all is well, because FreshRSS doesn't respect the S.O.L.I.D. principles and in particular the dependency inversion principle. It's not possible to simply Mock a database access. 
So I did my best to test what was possible...
It's important to think deeply about the project's architecture, and to rely on tried-and-tested architectural principles to avoid making code evolutions risky.  

It's not too late, and I'm sure we can work together to produce high-quality code.

Alkarex: Let me rephrase: adding those tests do not contribute to higher quality code, but on the contrary, would make the code base more energy-demanding to maintain.
I very much welcome more tests. Just not this type, sorry...

PR #5620, opened 2023-08-30
Replace screenshot for token configuration   @Alwaysin

Alwaysin:
See https://github.com/FreshRSS/FreshRSS/discussions/5619


Alkarex:
I can see that we also need to fix the text in the UI. The token is per user, not only for the default user

Alwaysin:
Would replacing "default" by "current" do the trick? Or maybe "of your user" instead of "of the default user".
In french "de l’utilisateur par défaut" to "de l’utilisateur actuel" or "de votre utilisateur", but we'll have the other languages to update.

Frenzie:
Like this?

`Allows access to the RSS output for this user without authentication:`

Alkarex:
Maybe even in plural for outputs:

> Allows access to the RSS outputs for this user without authentication:

PR #5584, opened 2023-08-12
Support TT-RSS JSON import

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/5583
With new mechanism `FreshRSS_Feed::KIND_RSS_MAYBE` to better handle links that require autodiscovery (HTML or RSS)

PR #5629, opened 2023-09-03
Do not update feeds after import

robertdahlem:
Importing multiple files tends to create duplicate feeds when the feeds are updated between imports. This is because updating feeds might change `_feed.url` especially with PubSubHubBub feeds but also in other scenarios. (Because of the other scenarios it is not enough to set `pubsubhubbub_enabled` to `false` while importing). When the next import runs, the importer tries to match the URL with existing feeds, fails and creates a new feed.

Of course automatic feed updates by cron or alternatives should be deactivated during imports. The message near the new checkbox says so.

### DIFF
--- a/app/i18n/fr/feedback.php
+++ b/app/i18n/fr/feedback.php
@@ -57,8 +57,8 @@ return array(
 	),
 	'import_export' => array(
 		'export_no_zip_extension' => 'L’extension ZIP n’est pas présente sur votre serveur. Veuillez essayer d’exporter les fichiers un par un.',
-		'feeds_imported' => 'Vos flux ont été importés.  Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.'
-		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.  Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.',
+		'feeds_imported' => 'Vos flux ont été importés.	Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.',
+		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.	Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.',
 		'file_cannot_be_uploaded' => 'Le fichier ne peut pas être téléchargé !',
 		'no_zip_extension' => 'L’extension ZIP n’est pas présente sur votre serveur.',
 		'zip_error' => 'Une erreur est survenue durant le traintement du fichier ZIP.',
--- a/app/i18n/fr/feedback.php
+++ b/app/i18n/fr/feedback.php
@@ -57,8 +57,8 @@ return array(
 	),
 	'import_export' => array(
 		'export_no_zip_extension' => 'L’extension ZIP n’est pas présente sur votre serveur. Veuillez essayer d’exporter les fichiers un par un.',
-		'feeds_imported' => 'Vos flux ont été importés et vont maintenant être actualisés.  / Your feeds have been imported. If you are done importing, you can now click the <i>Update feeds</i> button.', // TODO
-		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.  / Your feeds have been imported, but some errors occurred. If you are done importing, you can now click the <i>Update feeds</i> button.', // TODO
+		'feeds_imported' => 'Vos flux ont été importés.  Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.'
+		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.  Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.',
 		'file_cannot_be_uploaded' => 'Le fichier ne peut pas être téléchargé !',
 		'no_zip_extension' => 'L’extension ZIP n’est pas présente sur votre serveur.',
 		'zip_error' => 'Une erreur est survenue durant le traintement du fichier ZIP.',
--- a/app/i18n/fr/feedback.php
+++ b/app/i18n/fr/feedback.php
@@ -57,8 +57,8 @@ return array(
 	),
 	'import_export' => array(
 		'export_no_zip_extension' => 'L’extension ZIP n’est pas présente sur votre serveur. Veuillez essayer d’exporter les fichiers un par un.',
-		'feeds_imported' => 'Vos flux ont été importés et vont maintenant être actualisés.',
-		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.',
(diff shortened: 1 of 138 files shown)


Alkarex:
A bit related to this draft https://github.com/FreshRSS/FreshRSS/pull/5584

Alkarex:
I have not understood what changes it makes to not automatically update the feeds after import? Could you please try to explain some more?

robertdahlem:
> I have not understood what changes it makes to not automatically update the feeds after import? Could you please try to explain some more?

Sorry for answering so late.

Take a look at `http://xkcd.com/rss.xml`. It 301s to `https://xkcd.com/rss.xml`, the HTTPS variant. TT-RSS ignores this because [fox is fox](https://community.tt-rss.org/t/please-consider-permanent-redirects-as-permament/470). FreshRSS instead takes note and updates `_feed.url` while updating the feeds.

Now import this:
```{
  "items": [
    {
      "categories": [
        "user/-/state/com.google/read"
      ],
      "guid": "SHA1:eebc13be8c417ff82a43b505d9f2a299aa678943",
      "origin": {
        "title": "xkcd.com",
        "feedUrl": "http://xkcd.com/rss.xml",
        "htmlUrl": "http://xkcd.com/"
      },
      "url": "https://xkcd.com/2677/",
      "title": "Some title",
      "content": "some stuff",
      "published": "2022-09-26 04:00:00"
    }
  ]
}
```
FreshRSS imports it, updates the feeds, sees the 301 and updates `_feed.url`.

Now import another file with the same `feedUrl` and a different `guid`. FreshRSS doesn't find a match for the `http` variant and creates a new feed. It looks like a duplicate but when you delete one of the duplicates, you lose articles.

This can only be avoided by NOT updating the feeds between multiple imports.

Alkarex:
Instead of a new option, I would better like that we never auto-refresh feeds after an import, and modify the i18n strings

from:

```php
'feeds_imported' => 'Your feeds have been imported and will now be updated',
'feeds_imported_with_errors' => 'Your feeds have been imported, but some errors occurred',
```

to something like:

```php
'feeds_imported' => 'Your feeds have been imported. If you are done importing, you may now click the <i>Update feeds</i> button.',
'feeds_imported_with_errors' => 'Your feeds have been imported, but some errors occurred. If you are done importing, you may now click the <i>Update feeds</i> button.',
```

Ping @Frenzie for a potentially better wording

Frenzie:
Would you want to mention something about reenabling cron?

Alkarex:
@Frenzie I am afraid cron-aspects go beyond what many users can do. In particular, it requires server-admin access. So I am afraid it might be more confusing than helpful, so I think we should leave that to the migration guide.

Note for later: the whole problem of feeds changing addresses should be addressed with a better logic during the import, because disabling the refresh of feeds only addresses a tiny part of the use cases.

Frenzie:
Right, I forgot for a second that mentioning cron was proposed in this PR rather than the current situation. :-)

I'd make it "If you're done importing, you can now click the <i>Update feeds</i> button." but that's merely a minor stylistic preference for being less formal.

robertdahlem:
> Instead of a new option, I would better like that we never auto-refresh feeds after an import, and modify the i18n strings

I'm happy with never auto-refreshing. Do you expect me to change my PR or will you do that yourself? Apart of dropping the most of my i18n changes it's just the deletion of the update line.

Frenzie:
Simply removing the actualize_feeds line sounds fine by me anyway.

Alkarex:
An update of your PR would be welcome, @robertdahlem

robertdahlem:
> An update of your PR would be welcome, @robertdahlem

I updated my changes to skip feed updates unconditionally and merged the changes to the i18n strings. Hope I did everything right with GitHub, this is my first PR.

app/i18n/fr/feedback.php
```diff
@@ -57,8 +57,8 @@
 	),
 	'import_export' => array(
 		'export_no_zip_extension' => 'L’extension ZIP n’est pas présente sur votre serveur. Veuillez essayer d’exporter les fichiers un par un.',
-		'feeds_imported' => 'Vos flux ont été importés et vont maintenant être actualisés.',
-		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.',
+		'feeds_imported' => 'Vos flux ont été importés et vont maintenant être actualisés.  / Your feeds have been imported. If you are done importing, you can now click the <i>Update feeds</i> button.', // TODO
+		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.  / Your feeds have been imported, but some errors occurred. If you are done importing, you can now click the <i>Update feeds</i> button.', // TODO
```
Alkarex:
```suggestion
		'feeds_imported' => 'Vos flux ont été importés.  Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.'
		'feeds_imported_with_errors' => 'Vos flux ont été importés mais des erreurs sont survenues.  Si vous avez fini vos importations, vous pouvez cliquer le bouton <i>Actualiser flux</i>.',
```

Alkarex:
Well done 👍🏻 
If have only done some minor fixes https://github.com/FreshRSS/FreshRSS/pull/5629/commits/127edaf314a9aa74b5017f5085e46c9b866ecc2c

Alkarex (approved):

Alkarex:
Help welcome https://github.com/FreshRSS/FreshRSS/discussions/5731#discussioncomment-7369289

robertdahlem:
> Help welcome [#5731 (comment)](https://github.com/FreshRSS/FreshRSS/discussions/5731#discussioncomment-7369289)

I see the problem:

In `importAction()` in `app/Controllers/importExportController.php` the call `Minz_Session::_param('actualize_feeds', true);`has been deleted to avoid feed updates after JSON imports. Unfortunately `importAction()` is called for all kinds of imports, including OPML where you probably want to do a feed update afterwards.

The decision what kind of import we are doing is in `importFile()` which returns `true` for "ok" and `false` for "error". Instead it should perhaps return `0` for "error", `1` for "ok, don't update feeds" (it was a JSON import) and `2` for "ok, update feeds" (it was an OPML import).

The big "but" is: you can import ZIP files, that include JSON and OPML. In this case the answer would be `3`, meaning "you can't have nice things :-)".

All in all I think it said "Your feeds have been imported. If you are done importing, you can now click the **Update feeds** button." after the import. Which is what you suggested in #5731. What exactly is wrong with the wording?

I read about needing to click **Update feeds** twice. This has nothing to do with this PR but what is done after clicking the button.

Alkarex:
The problematic wording is *This feed is empty. Please verify that it is still maintained.*

robertdahlem:
> The problematic wording is _This feed is empty. Please verify that it is still maintained._

I see.

Maybe in `importFile()` we should do something like `$force_refresh = true` in the OPML part when `$importService->lastStatus()` is `true` and call `Minz_Session::_param('actualize_feeds', true);` before `return $ok;` when `$force_refresh` is `true`.

This probably would avoid the problematic wording.

The corner case is still a ZIP file with OPML and JSON in it. But anyway: who has a ZIP file with OPML and JSON in it where the JSON files are only a part of the import? I would call that "asking for trouble".

PR #4649, opened 2022-09-27
Fix: my labels in anonym. mode

math-GH:
Closes #4305

Changes proposed in this pull request:

- show selected tags in tags list in anyom. mode
- do not show "my labels" link+dropbox, if the article has no selected tag

How to test the feature manually:

use anonym. mode
1.  article: do not show "my labels", if article has no assigned label
2. article: show "my labels", if article has assigned label + show dropdown with assigned labels

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #5650, opened 2023-09-10
Fix labels in anonymous mode

### LINKED ISSUE #4305: [BUG] anonym. mode: my labels error
**To Reproduce**
1. use FreshRSS in anonym. mode
2. open an article
3. click on "my labels"
4. see browser console (error displayed)
5. see the empty label list

grafik

**Expected behavior**

- No error
- show added labels

- Browser: Firefox 98
- FreshRSS version: 1.19.2 (see demo.freshrss.org), 1.20.0-dev

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/4305
Replaces https://github.com/FreshRSS/FreshRSS/pull/4649

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1339,15 +1339,22 @@ function loadDynamicTags(div) {
 
 		let html = '';
 		if (json && json.length) {
+			let nbLabelsChecked = 0;
 			for (let i = 0; i < json.length; i++) {
 				const tag = json[i];
 				if (context.anonymous && !tag.checked) {
 					// In anomymous mode, show only the used tags
 					continue;
 				}
+				if (tag.checked) {
+					nbLabelsChecked++;
+				}
 				html += '<li class="item"><label><input class="checkboxTag" name="t_' + tag.id + '" type="checkbox"' +
 						(context.anonymous ? ' disabled="disabled"' : '') +
-						(tag.checked ? ' checked="checked"' : '') + '> ' + tag.name + '</label></li>';
+						(tag.checked ? ' checked="checked"' : '') + '/> ' + tag.name + '</label></li>';
+			}
+			if (nbLabelsChecked === 0) {
+				html += '<li class="item">' + context.i18n.labels_empty + '</li>';
 			}
 		}
 		div.querySelector('.dropdown-menu').insertAdjacentHTML('beforeend', html);
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1349,9 +1349,11 @@ function loadDynamicTags(div) {
 				if (tag.checked) {
 					nbLabelsChecked++;
 				}
-				html += '<li class="item"><label><input class="checkboxTag" name="t_' + tag.id + '" type="checkbox"' +
-						(context.anonymous ? ' disabled="disabled"' : '') +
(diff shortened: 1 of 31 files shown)


p/scripts/main.js
```diff
@@ -1300,46 +1300,53 @@ function loadDynamicTags(div) {
...
 		let html = '';
 		if (json && json.length) {
 			for (let i = 0; i < json.length; i++) {
 				const tag = json[i];
+				if (context.anonymous && !tag.checked) {
+					// In anomymous mode, show only the used tags
+					continue;
+				}
 				html += '<li class="item"><label><input class="checkboxTag" name="t_' + tag.id + '" type="checkbox"' +
+						(context.anonymous ? ' disabled="disabled"' : '') +
```
math-GH: Please have a look into line 1201. The event listener does not make sense in context.anonymous = true on the checkbox.

math-GH:
If there is no label assigned then the popup list looks a bit broken:

grafik
The title is given but nothing else

p/scripts/main.js
Alkarex: The `if (checkboxTag) {` should do the job, doesn't it? In anonymous mode, this is false

math-GH: `if (ev.target.closest('.checkboxTag'))` is independend of context nor `disabled="disabled"`

Alkarex:
> If there is no label assigned then the popup list looks a bit broken:

What about now?

math-GH:
> > If there is no label assigned then the popup list looks a bit broken:
> 
> What about now?

IMHO it does not make sense to list all (incl. not used) labels in the anonym. mode

grafik


I would see follwing 2 options in case of that there is no label added:
- hide the "label" link (I guess that it would need some additional SQL statement)
- give the information "no label set" in the pop-up

p/scripts/main.js
Alkarex: Indeed, but there is no `.checkboxTag` in anonymous mode, so the rest is just skipped as it should, isn't it ?

Alkarex:
I have added a message when there are no labels.
Maybe it could be styled a bit better if you feel like it @math-GH 

image

I would like to keep the labels also in anonymous mode, as there are more features coming related to labels (such as automatic labelling).

p/scripts/main.js
math-GH: The class is still there 

grafik

grafik

math-GH:
No labels style improved.

p/scripts/main.js
Alkarex: Sorry, my mistake. Better now?

math-GH:
It looks fine now.

Does it make sense to keep the background color while hovering in this situation? There is no click action behind. As a user I would expect something there

grafik

As far as I can oversee it there is no CSS class for the anonym. mode, isn't it?

Alkarex:
> Does it make sense to keep the background color while hovering

I do not mind much, but I find it fine as it is, and I would rather limit the amount of code, especially when specific to a relatively niche feature

Frenzie:
Fwiw you might be able to do something with the new :has() selector, but wouldn't it make more sense to simply not show the checkbox instead of disabling it?

Issue #996, opened 2015-10-13
Smileys from Wordpress are too big


NerosTie:
This is a normal smiley: 

![tmp2](https://cloud.githubusercontent.com/assets/4572452/10459542/4c34fb6e-71d0-11e5-8614-8584ab7cf1aa.png)

This is a smiley, from the same article, in freshrss:

![tmp3](https://cloud.githubusercontent.com/assets/4572452/10459565/6b05909e-71d0-11e5-9cec-226657c3f1da.png)

(if you want to try: http://grisebouille.net/ailleurs-cest-pire/ )

Alkarex:
Hum, it will not be easy to solve globally.
The original image is the large one, which is reduced by styling. And we (SimplePie) are removing the styling.

``` html
<img style="height: 1em; max-height: 1em;" class="wp-smiley" alt=":)" src="http://grisebouille.net/wp-includes/images/smilies/simple-smile.png">
```

A quick fix would be to add a custom CSS (see the FreshRSS extension for that) with

``` css
.wp-smiley { height: 1em; max-height: 1em; }
```

marienfressinaud:
`class` seems to be removed as well so custom CSS would not work...

NerosTie:
Yes, it doesn't work with a custom css (I tried)

Alkarex:
Damn, you are right.
This should work, but is even less universal...:

``` css
img[alt=':)'] { height: 1em; max-height: 1em; }
```

NerosTie:
Yes, it works, but it's not really a good solution (but better than no solution at all!)

Wanabo:
Not tested, but try this for universal:

```
img[src*="/wp-includes/images/smilies/"] { max-height: 25px; }
```

https://css-tricks.com/attribute-selectors/

NerosTie:
@Wanabo it doesn't work (I don't know why...)

Wanabo:
You could try:

```
img[src*="/smilies/"] { max-height: 25px; }
```

Can you provide a link to an article on your fressrss website where those big smileys occur?

NerosTie:
@Wanabo It doesn't work either.

You can test with this blog: http://metropolife.net/  
Look at this post, she uses 3 different emojis: http://metropolife.net/what-i-have-learned-about-denmark/

Wanabo:
Sorry need a link to an article on your fressrss website where those big smileys occur. Only then I can fiddle with firebug to see what css will work.

[...]

PR #5653, opened 2023-09-11
Allow string in XPath tags

### LINKED ISSUE #5651: [BUG]item tags (static string) are not available in HTML + XPath mode
**Describe the bug**
Item tags are not available when using static string in HTML + XPath mode.

**To Reproduce**
1. Go to 'Add a feed'
2. Type of feed source option select 'HTML + XPath'
3. Fill in the item tags column with 'tag demo'
4. See error: Article tags does not add the custom tag 'tag demo'

**Expected behavior**
In HTML + XPath mode, item tags should support static string like other items (such as item author)

- Device: PC
- OS: windows 11
- Browser: edge 116
- FreshRSS version: 1.21.0
- Database version: MySQL 5.7.40
- PHP version: PHP 7.4
- Installation type: web

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5651

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -723,8 +723,10 @@ class FreshRSS_Feed extends Minz_Model {
 				}
 				$item['thumbnail'] = $xPathItemThumbnail == '' ? '' : @$xpath->evaluate('normalize-space(' . $xPathItemThumbnail . ')', $node);
 				if ($xPathItemCategories != '') {
-					$itemCategories = @$xpath->query($xPathItemCategories, $node);
-					if ($itemCategories !== false) {
+					$itemCategories = @$xpath->evaluate($xPathItemCategories, $node);
+					if (is_string($itemCategories) && $itemCategories !== '') {
+						$item['tags'] = [$itemCategories];
+					} elseif ($itemCategories instanceof DOMNodeList && $itemCategories->length > 0) {
 						$item['tags'] = [];
 						/** @var DOMNode $itemCategory */
 						foreach ($itemCategories as $itemCategory) {


Frenzie:
This suddenly reminded me of https://github.com/FreshRSS/FreshRSS/discussions/5644

Perhaps it would be relatively "simple" to add some kind of XSLT as another processing step, in the sense that XSLT is already in PHP? Just a random off-hand thought mind you, not a suggestion or anything.

Alkarex:
> Perhaps it would be relatively "simple" to add some kind of XSLT as another processing step, in the sense that XSLT is already in PHP? Just a random off-hand thought mind you, not a suggestion or anything.

Yes @Frenzie , adding XLTS would indeed be easy, both for processing inputs and outputs. Can you tell a bit more about your idea? (maybe in another discussion thread). I was actually considering XSLT to provide alternative outputs in addition of our RSS outputs.

Frenzie:
What I wrote above is the full extent of the thought really.

1. A pre-insertion step. First the CSS selector/XPath(/regular RSS feed?) we already have, followed by some XSLT to do whatever you might not be able to (sanely) do through XPath.
2. A pre-output step. Same as above but not permanent. I'm not sure if both would make sense or just one of the two.

I think you just added 3, do the XSLT on the entirety of a generated feed? Basically I just meant that with XSLT you wouldn't need to write an extension or an RSS Bridge bridge for many common use cases. Think of minor issues like #996, except that one happened to be easily fixable through CSS.

Although as an aside, reading the linked discussion in https://github.com/FreshRSS/FreshRSS/discussions/5644 again I don't think that's quite related to this thought. Or at least I was only thinking of the content of an entry.

PR #5656, opened 2023-09-13
Allow configuration of Base URL via website

### LINKED ISSUE #5246: [BUG] Instance update base_url
I migrated my dockerized FreshRSS instance, migration works fine but the base_url still points to the server where it was initially configured on

Migrate docker data to new server and run the instance on a new server

**Expected behavior**
base_url recognizing that it has been changed or a way to configure base_url from the webpage

samc1213:
Closes #5246 

- Add a configuration for base_url in the System Configuration page

1. Open System Configuration
2. Change "Base URL"
3. Refresh page, and ensure Base URL is updated
4. Check config.php and ensure `base_url` is updated

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


app/i18n/fr/admin.php
```diff
@@ -159,6 +159,7 @@
 	'system' => array(
 		'_' => 'Configuration du système',
 		'auto-update-url' => 'URL du service de mise à jour',
+		'base-url' => 'Base URL',	// TODO
```
```suggestion
		'base-url' => 'URL de la racine',
```

app/views/configure/system.phtml
```diff
@@ -54,6 +54,14 @@
 			</div>
 		</div>
 
+		<div class="form-group">
+			<label class="group-name" for="base-url"><?= _t('admin.system.base-url') ?></label>
+			<div class="group-controls">
+				<input type="text" id="base-url" name="base-url" value="<?= FreshRSS_Context::$system_conf->base_url ?>"
+					data-leave-validation="<?= FreshRSS_Context::$system_conf->limits['cookie_duration'] ?>"/>
```
Alkarex:
```suggestion
					data-leave-validation="<?= FreshRSS_Context::$system_conf->base_url ?>"/>
```

Alkarex:
Thanks, and welcome 👍🏻

app/views/configure/system.phtml
samc1213: Oops, good find, thank you

Alkarex:
I have added an automatic detection hint:
image

Alkarex (approved):

Frenzie (approved):
makes sense

samc1213:
> I have added an automatic detection hint: image

Love this idea - I tested it as well and it looks good to me.

math-GH:
I am a bit late to the party...
I would expect an improved documentation and would like to have here a warning message. It is very easy to destroy the application.

I changed `http://localhost/freshrss/p` to `http://localhost/freshrss` and it crashed the system.

grafik

It is to easy to fail here

Frenzie:
True, something like this might need a new UI element. Or at least I don't think we have something like an input disabled with an unlock button next to it anywhere.

I mean exactly the same thing as the show password button basically, just to get rid of the disabled attribute. You could additionally add a little warning underneath the title if that's not sufficiently saying "dangerous" all by itself.
image

Alkarex:
Right, indeed @math-GH . This went a bit too fast indeed, and should have waited for 1.23.
Temporary mitigation https://github.com/FreshRSS/FreshRSS/pull/5657

math-GH:
Let me think loud:
If the base URL is wrong, then the front end does not work. The admin cannot change it because (s)he cannot access.
If the base URL is correct, then it does not need to be changed.

What is the use case that needs the option? (I read the issue ticket but does not understand the use case)

Frenzie:
> If the base URL is wrong, then the front end does not work. The admin cannot change it because (s)he cannot access.
If the base URL is correct, then it does not need to be changed.

This is incorrect. I had a wrong base URL for years before I even noticed, unless I'm thinking of a different similar setting.

math-GH:
> > If the base URL is wrong, then the front end does not work. The admin cannot change it because (s)he cannot access.
> > If the base URL is correct, then it does not need to be changed.
> 
> This is incorrect. I had a wrong base URL for years before I even noticed, unless I'm thinking of a different similar setting.

That leads us back to the documentation: for what is the base URL used/needed?

Alkarex:
The base URL is used for the generation of various addresses, including relative addresses, some redirections, and for constructing absolute URLs as needed by e.g. API, WebSub

math-GH:
Any example for my setup that will work (default: `http://localhost/freshrss/p`)?

Alkarex:
@math-GH An example of place where you can see it in action is in one of the RSS outputs. Check the `<atom:link href="".../>`

Frenzie:
I believe in my case it wasn't wrong per se but the IP address instead of the domain which led to some subtle problem with something API-related iirc, like the inability to log on when entering the domain in a client.

marien-probesys:
What about guessing the recommended `base_url` with JavaScript instead? `window.location.href` would return a reliable URL, just having to strip the query part.

Alkarex:
> What about guessing the recommended `base_url` with JavaScript instead? `window.location.href` would return a reliable URL, just having to strip the query part.

Indeed Marien, as a tip to autofill the information in the configuration page. But that cannot be used the rest of the time (e.g. from API, WebSub, HTTP redirects) as the request is not passing through a JavaScript client environment.

PR #5658, opened 2023-09-15
Add xdebug for devs

ColonelMoutarde:

- Add xdebug for devs

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


.devcontainer/Dockerfile
```diff
@@ -3,12 +3,15 @@ FROM alpine:3.18
...
 
+COPY ./xdebug.ini /etc/php81/xdebug.ini
+
 RUN apk add --no-cache \
 	tzdata \
 	apache2 php-apache2 \
 	php php-curl php-gmp php-intl php-mbstring php-xml php-zip \
 	php-ctype php-dom php-fileinfo php-iconv php-json php-opcache php-openssl php-phar php-session php-simplexml php-xmlreader php-xmlwriter php-xml php-tokenizer php-zlib \
 	php-pdo_sqlite php-pdo_mysql php-pdo_pgsql \
+    php81-pecl-xdebug \
```
Alkarex, line 14: We might consider moving everything to PHP 8.2 for the development container

Alkarex: Let me make another PR for that

Alkarex: Hum, `composer` still requires `php81` explicitly https://pkgs.alpinelinux.org/package/v3.18/community/x86_64/composer , so let's wait for it

.devcontainer/xdebug.ini
```diff
@@ -0,0 +1,7 @@
+zend_extension=xdebug.so
+xdebug.mode=debug,coverage
+xdebug.start_with_request=yes
+xdebug.client_host=host.docker.internal
+xdebug.client_port=9003
+xdebug.idekey=PHPSTORM
```
Alkarex: We should probably leave that to the `DBGP_IDEKEY` environment setting

Alkarex:
I just had a quick look. The PR as it stands does not seem to be sufficient to make XDebug work out of the box in a development container.

Please test your PR from https://github.com/codespaces/new?ref=feature%2Fadd-xdebug-in-project&repo=444135317

In particular, the related extension is probably lacking to work in GitHub Codespaces
https://marketplace.visualstudio.com/items?itemName=xdebug.php-debug

In any case, could you please add a little bit of documentation in a new file here https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/ with a link from https://github.com/FreshRSS/FreshRSS/blob/edge/.devcontainer/README.md (which is automatically opened when the dev container starts).

.devcontainer/xdebug.ini
ColonelMoutarde: I don't know how to deal with this comment, can you help me?

Alkarex: This line is hardcoded for PhpStorm (I am not sure whether that IDE is used at all for running dev containers, by the way). Other values include `VSCODE`, etc. So a more configurable approach is via the environment variable `DBGP_IDEKEY`
See https://xdebug.org/docs/all_settings#idekey
And the environment variable could be defined in a `.env` (either be default or by the user) https://code.visualstudio.com/remote/advancedcontainers/environment-variables#_option-2-use-an-env-file

But it does not look super important, so maybe for now just removing this line should be enough, and we can come back to it when writing the documentation.

.devcontainer/xdebug.ini
```diff
@@ -0,0 +1,7 @@
+zend_extension=xdebug.so
+xdebug.mode=debug,coverage
+xdebug.start_with_request=yes
+xdebug.client_host=host.docker.internal
+xdebug.client_port=9003
+xdebug.idekey=DBGP_IDEKEY
```
Alkarex: I do not think that works. For the `DBGP_IDEKEY` environment variable to be used, I believe this line should be removed completely

```suggestion
```

ColonelMoutarde: in my IDE (phpstorm) I can put any key as long as it's a string, it's the same with other IDEs like netbeans.

"Session ID Arbitrary name of the debugging session. The default is netbeans-xdebug. This value is important to note in cases where you need to set the xdebug.idekey property in php.ini, as in some remote debugging scenarios."

I can do a "sed" on the key value if you like.
it's better to think of the values as defaults, I'm looking to easily override them in another commit

Alkarex: No need for `sed`, it should work with the environment variable after just deleting this line:

https://xdebug.org/docs/all_settings#idekey

> The default is based on the `DBGP_IDEKEY` environment setting

Alkarex:
Anything new here?

ColonelMoutarde:
Hi, I closed this PR, my time is short now. In January I will go back to
the feature.
Cordialement,
Luc SANCHEZ

Alkarex:
Ok, it would indeed be interesting to finalise adding some xdebug

ColonelMoutarde:
> Ok, it would indeed be interesting to finalise adding some xdebug

You're right.

PR #5669, opened 2023-09-23
Fix search using user queries

### LINKED ISSUE #5668: [BUG] "Feed DAO is not loaded in UserQuery" error when filtering search results by user query
**Describe the bug**
If I use the search field to filter results by user query (saved search) that has a feed or a category in it, the browser displays an error "Feed DAO is not loaded in UserQuery".

**To Reproduce**
1. Create a new user query with any expression, and select any feed or category in the Type list (see screenshot below). Note the id of the query, e.g. 1
2. Go to "Main stream" page and type in "S:1" in the search field
3. The browser displays an error "HTTP 500: Application problem (Feed DAO is not loaded in UserQuery)"

**Expected behavior**
The search results should be filtered according to the user query.

**Screenshots**
freshrss

**Environment information:**
- Device: PC
- OS: Windows 11
- Browser: Microsoft Edge 117
- FreshRSS version: 1.21.0
- Database version: SQLite
- PHP version: PHP 7.4.33
- Installation type: Image from Docker Hub freshrss/freshrss:latest

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5668

### DIFF
--- a/app/Models/UserQuery.php
+++ b/app/Models/UserQuery.php
@@ -125,7 +125,7 @@ class FreshRSS_UserQuery {
 	 */
 	private function parseCategory(int $id): void {
 		if ($this->category_dao === null) {
-			throw new FreshRSS_DAO_Exception('Category DAO is not loaded in UserQuery');
+			$this->category_dao = FreshRSS_Factory::createCategoryDao();
 		}
 		$category = $this->category_dao->searchById($id);
 		if ($category !== null) {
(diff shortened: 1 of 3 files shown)


app/Models/UserQuery.php
```diff
@@ -125,7 +125,7 @@ private function parseAll(): void {
 	 */
 	private function parseCategory(int $id): void {
 		if ($this->category_dao === null) {
-			throw new FreshRSS_DAO_Exception('Category DAO is not loaded in UserQuery');
```
Alkarex, line 128: Overkill to raise an exception when the same object can be created on demand

PR #5671, opened 2023-09-25
Add Japanese to font-family

superuser7777:
Please consider changing the CSS as shown below.
This modification improves display readability for Japanese people.

Add
`"Osaka",`

Change before
`font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "PingFang SC", "Microsoft YaHei", sans-serif;`

After change
`font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "Osaka", "PingFang SC", "Microsoft YaHei", sans-serif;`

The Chinese fonts such as "PingFang SC" and "Microsoft YaHei" can be displayed like Japanese, but the kanji are greatly broken down, making them difficult for Japanese people to read.

--
Thank you for the great software!
I've migrated all my RSS readers to FreshRSS!


Frenzie (approved):

Alkarex:
We should probably update all the themes, which have overridden the base font definition, e.g.:

https://github.com/FreshRSS/FreshRSS/blob/86d713478b25816340aebeeacca704f3aeee17af/p/themes/Alternative-Dark/adark.css#L52

Edit: Done https://github.com/FreshRSS/FreshRSS/pull/5671/commits/386463457ab4d3fc680ecf7067f5b0ebd83b2932

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/pull/1541

@hoilc if you are around, do you have any comment? For instance a negative consequence, we might have overlooked?

Alkarex:
I am wondering whether adding (only) "Osaka" is the most appropriate fix. It seems to be only for Apple. Isn't e.g. "MS Gothic" more common?

Alkarex:
Note for later: we should add a way to tweak font parameters independently of the default settings proposed by the themes (we might have an issue somewhere already)

math-GH (changes requested):
It is very nice to know that FreshRSS is used in Japan too :)

Please support us here with screenshots "before" and "after" the change.

Please find another default font families that is installed on Windows by default. Linux would be fine too.

math-GH:
the conflicted files can be deleted to solve the conflict

Alkarex:
@superuser7777 Your feedback would be welcome in https://github.com/FreshRSS/FreshRSS/pull/5895 , for instance related to font selection and calendar selection (which calendar system do you use?)

Alkarex:
Tests welcome:
* https://github.com/FreshRSS/FreshRSS/pull/9002

PR #5675, opened 2023-09-27
Test if set_time_limit exists

### LINKED ISSUE #4982: [BUG] PHP 8 disabled function behavior has changed
**Describe the bug**
Some feeds are displayed like having issues, are not refreshed. Impossible to refresh them or re-add them.

**To Reproduce**
- get a feed with this issue
- try to refresh it or re-add it

**Expected behavior**
I guess FreshRSS should fetch new posts.

**Screenshots**
image

image

image
Feed is valid and reachable.

I did try to delete it and re-add it, i just get a blank page:

image

- PHP version: 8.0.25 on a self hosted infomaniak website

Everything seems ok from hosting pov
image

**Additional context**
Anything I could provide, don't hesitate to ask.

V-E-O:
…abled functions

Closes #4982

-@set_time_limit to function_exists then set_time_limit

1.purge and optimize db
2.refresh and actualize feeds

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated (documentation not affected)

### DIFF
--- a/app/Controllers/entryController.php
+++ b/app/Controllers/entryController.php
@@ -170,7 +170,7 @@ class FreshRSS_entry_Controller extends FreshRSS_ActionController {
 		}
 
                 if (function_exists('set_time_limit')) {
-                        set_time_limit(300);
+                        @set_time_limit(300);
                 }
 
 		$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
@@ -191,7 +191,7 @@ class FreshRSS_entry_Controller extends FreshRSS_ActionController {
 	 */
 	public function purgeAction(): void {
                 if (function_exists('set_time_limit')) {
-                        set_time_limit(300);
+                        @set_time_limit(300);
                 }
 
 		$feedDAO = FreshRSS_Factory::createFeedDao();
--- a/app/Controllers/entryController.php
+++ b/app/Controllers/entryController.php
@@ -169,9 +169,9 @@ class FreshRSS_entry_Controller extends FreshRSS_ActionController {
 			Minz_Request::forward($url_redirect, true);
 		}
 
-                if (function_exists('set_time_limit')) {
-                        @set_time_limit(300);
-                }
+		if (function_exists('set_time_limit')) {
+			@set_time_limit(300);
+		}
 
 		$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
 		$databaseDAO->optimize();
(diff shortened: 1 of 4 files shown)


app/Controllers/entryController.php
```diff
@@ -169,7 +169,9 @@ public function optimizeAction(): void {
 			Minz_Request::forward($url_redirect, true);
 		}
 
-		@set_time_limit(300);
+                if (function_exists('set_time_limit')) {
+                        set_time_limit(300);
```
Frenzie: Is that truly the only reason it might fail? This way it seems that you're changing behavior compared to just leaving it with `@` to say that any failure doesn't really matter.
```suggestion
                        @set_time_limit(300);
```

V-E-O: @Frenzie set_time_limit only raise fatal error if it's disabled.

> Prior to PHP 8.0.0, it was possible for the @ operator to disable critical errors that will terminate script execution. For example, prepending @ to a call of a function which did not exist, by being unavailable or mistyped, would cause the script to terminate with no indication as to why.
from https://www.php.net/manual/en/language.operators.errorcontrol.php

Thus with function_exists (supported from PHP 4 to 8), @ decorator is not needed.
Any way, preserve the @ is also okay.

V-E-O: preserve @ decorator commit appended

app/Controllers/entryController.php
```diff
@@ -169,7 +169,9 @@ public function optimizeAction(): void {
 			Minz_Request::forward($url_redirect, true);
 		}
 
-		@set_time_limit(300);
+                if (function_exists('set_time_limit')) {
+                        @set_time_limit(300);
```
ColonelMoutarde:
```suggestion
                       $isTimeLimited = set_time_limit(300);
                       if ($isTimeLimited === false) {
                                    continue;
                       }
```

ColonelMoutarde: You can add the logic in function

V-E-O: @ColonelMoutarde After PHP8, with function set_time_limit disabled, set_time_limit (or even decorated with @) will cause fatal termination of script.

> Prior to PHP 8.0.0, it was possible for the @ operator to disable critical errors that will terminate script execution. For example, prepending @ to a call of a function which did not exist, by being unavailable or mistyped, would cause the script to terminate with no indication as to why.

from https://www.php.net/manual/en/language.operators.errorcontrol.php

ColonelMoutarde: Hi, https://www.php.net/manual/en/function.set-time-limit.php set_time_limit is available for all php versions from 4 to 8.
The @ is not necessary, especially as you're checking that the function is present just beforehand. ;)

Alkarex: > set_time_limit is available for all php versions from 4 to 8.

Partially wrong: any function can be disabled, and we have several examples of providers doing so for that specific function

> The @ is not necessary

Wrong: there might be a warning with PHP < 8 (the function would exist, but would be disabled). See https://php.watch/versions/8.0/disable_functions-redeclare

PR #5690, opened 2023-10-07
Try Catch for window.Notification

### LINKED ISSUE #5687: [BUG] Installed via Docker Error: "The Notification permission may only be requested in a secure context."
**Describe the bug**
When running FreshRSS directly from a Docker Container (like in your documentation) the service runs without SSL, what is ok.  But after login the click on a news item does not work. 

In the Console I can see an error message like this:

    The Notification permission may only be requested in a secure context.

I guess that the JS execution stops here because of the exception.

I know this is not a production ready configuration, but it would be nice if it would work just out-of-the-box (maybe catch the situation and print a warning that the notifications will not work because of not secure context etc.)

**To Reproduce**
1. Setup via Docker like described here https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/README.md
2. Login as Admin
3. It will show some feed items of a default feed
4. Click on the item and nothing will happen
5. When open the Console you will see the error

**Expected behavior**
It would be nice to work just out-of-the-box. Maybe catch the situation and print a warning that the notifications will not work because of not secure context etc.  Let the user just open the item.

I know this is not a critical issue because the service should run in SSL but I think this behavior could bother some people try to setup the App and see its not working properly on localhost ... 

**Screenshots**

Bildschirmfoto 2023-10-07 um 16 24 00

- Device: Macbook Pro
- OS: OSX 14.0
- Browser: Firefox 117
- FreshRSS version: Docker edge
- Database version: PostgresQL 16
- PHP version: 8.2.7

**Additional context**

Dockerfile

```yaml
version: "2.4"


volumes:
  data:
    driver: local
  extensions:
    driver: local
  db:
    driver: local


services:
  freshrss:
    image: freshrss/freshrss:edge
    # Optional build section if you want to build the image locally:
    build:
      # Pick #latest (stable release) or #edge (rolling release) or a specific release like #1.21.0
      context: https://github.com/FreshRSS/FreshRSS.git#edge
      dockerfile: Docker/Dockerfile-Alpine
    container_name: freshrss
    restart: unless-stopped
    logging:
      options:
        max-size: 10m
    volumes:
      # Recommended volume for FreshRSS persistent data such as configuration and SQLite databases
      - data:/var/www/FreshRSS/data
      # Optional volume for storing third-party extensions
      - extensions:/var/www/FreshRSS/extensions
      # Optional file providing custom global settings (used before a FreshRSS install)
      - ./config.custom.php:/var/www/FreshRSS/data/config.custom.php
      # Optional file providing custom user settings (used before a new user is created)
      - ./config-user.custom.php:/var/www/FreshRSS/data/config-user.custom.php
    ports:
      # If you want to open a port 8080 on the local machine:
      - "6080:80"
    environment:
      # A timezone http://php.net/timezones (default is UTC)
      TZ: Europe/Berlin
      # Cron job to refresh feeds at specified minutes
      CRON_MIN: '2,32'
      # 'development' for additional logs; default is 'production'
      FRESHRSS_ENV: production
      # Optional advanced parameter controlling the internal Apache listening port
      LISTEN: 0.0.0.0:80
      # Optional parameter, remove for automatic settings, set to 0 to disable,
      # or (if you use a proxy) to a space-separated list of trusted IP ranges
      # compatible with https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteiptrustedproxy
      # This impacts which IP address is logged (X-Forwarded-For or REMOTE_ADDR).
      # This also impacts external authentication methods;
      # see https://freshrss.github.io/FreshRSS/en/admins/09_AccessControl.html
      TRUSTED_PROXY: 172.16.0.1/12 192.168.0.1/16
      # Optional parameter, set to 1 to enable OpenID Connect (only available in our Debian image)
      # Requires more environment variables. See https://freshrss.github.io/FreshRSS/en/admins/16_OpenID-Connect.html
      OIDC_ENABLED: 0
      # Optional auto-install parameters (the Web interface install is recommended instead):
      # ⚠️ Parameters below are only used at the very first run (so far).
      # So if changes are made (or in .env file), first delete the service and volumes.
      # ℹ️ All the --db-* parameters can be omitted if using built-in SQLite database.
      FRESHRSS_INSTALL: |-
        --api_enabled
        --base_url ${BASE_URL}
        --db-base ${DB_BASE}
        --db-host ${DB_HOST}
        --db-password ${DB_PASSWORD}
        --db-type pgsql
        --db-user ${DB_USER}
        --default_user admin
        --language en
      FRESHRSS_USER: |-
        --api_password ${ADMIN_API_PASSWORD}
        --email ${ADMIN_EMAIL}
        --language en
        --password ${ADMIN_PASSWORD}
        --user admin
    depends_on:
      - freshrss-db

  freshrss-db:
    image: postgres:16
    container_name: freshrss-db
    hostname: freshrss-db
    restart: unless-stopped
    logging:
      options:
        max-size: 10m
    volumes:
      - db:/var/lib/postgresql/data
    environment:
      POSTGRES_DB: ${DB_BASE:-freshrss}
      POSTGRES_USER: ${DB_USER:-freshrss}
      POSTGRES_PASSWORD: ${DB_PASSWORD:-freshrss}
    command:
      # Examples of PostgreSQL tuning.
      # https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server
      # When in doubt, skip and stick to default PostgreSQL settings.
      - -c
      - shared_buffers=128MB
      - -c
      - work_mem=16MB
```

The "env" file

```shell
# Example of environment file for docker-compose
# Copy this file into your own `.env` file

# ================================
# FreshRSS
# ================================

ADMIN_EMAIL=admin@example.com

# Published port for development or local use (optional)
#PUBLISHED_PORT=6080

# =========================================
# For automatic FreshRSS install (optional)
# =========================================

ADMIN_PASSWORD=xxx 	
ADMIN_API_PASSWORD=xxx

# Address at which the FreshRSS instance will be reachable:
BASE_URL=https://freshrss.example.com

# Database server (not relevant if using default SQLite)
# Use the name of the Docker container if running on the same machine
DB_HOST=freshrss-db

# ===========================================================
# Database credentials (not relevant if using default SQLite)
# ===========================================================

# Database to use
DB_BASE=freshrss

# User in the freshrss database
DB_USER=freshrss

# Password for the defined user
DB_PASSWORD=xxx
```

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5687
Exceptions might be thrown in some cases
https://developer.mozilla.org/docs/Web/API/Notification/Notification


Frenzie (approved):
Seems rather unnecessary to me (it just posts a message to the log), but if you like. ^_^

Frenzie:
Actually the try/catch block still posts to the console, so I don't think it helps any.

Alkarex:
It is for catching `TypeError`, which might happen according to https://developer.mozilla.org/docs/Web/API/Notification/Notification
If we do not catch, some other elements of our `main.js` might stop.

Frenzie:
That would only happen if you put contradictory options in, in which case you'll notice it immediately. #5687 doesn't have anything to do with that.

Frenzie:
Simply put, there's no such thing as an exception that might be caused by Notification and there never has been.

Frenzie:
Something like this:
image

Alkarex:
I was referring to e.g. the following, but I admit that I have not looked into the details:

image

Frenzie:
That's pretty weird. Oh well.

Definitely not related to #5687 though. :-P

PR #5733, opened 2023-10-24
Avoid a warning on non-numeric TRUSTED_PROXY en var

### LINKED ISSUE #5732: [BUG] Minor warning in docker logs when setting a non-numeric TRUSTED_PROXY env var
**Describe the bug**
If you set a TRUSTED_PROXY env var with a non-numeric value, like "10.0.0.0/8", there is a warning in the log:
` ./Docker/entrypoint.sh: 15: [: Illegal number: 10.0.0.0/8`

**To Reproduce**
1. Deploy freshrss 1.22 through its docker image
2. pass a TRUSTED_PROXY env var, with value "10.0.0.0/8"
3. Run the container
4. See the first line in docker logs:

`./Docker/entrypoint.sh: 15: [: Illegal number: 10.0.0.0/8`

**Expected behavior**
There should be no warning, as this is an expected value for this env var, according to the documentation

- Device: Desktop
- OS: Ubuntu 22.04
- Browser: Firefox 118.0.2
- FreshRSS version: 1.22
- Database version: PostgreSQL 14.7
- PHP version: 8.2.7
- Installation type: docker image 1.22.0-arm (but should be the same with amd)

**Additional context**
This is because of the following line in entrypoint.sh file:
`if [ "$TRUSTED_PROXY" -eq 0 ]; then`

mossroy:
Fixes #5732

- use a double-bracket syntax, that does not complain when TRUSTED_PROXY env var is not numeric

How to test the feature manually: follow the steps from #5732

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -12,7 +12,7 @@ if [ -n "$LISTEN" ]; then
 fi
 
 if [ -n "$TRUSTED_PROXY" ]; then
-	if [[ "$TRUSTED_PROXY" == 0 ]]; then
+	if [ "$TRUSTED_PROXY" == "0" ]; then
 		# Disable RemoteIPHeader and RemoteIPTrustedProxy
 		find /etc/apache2/ -type f -name FreshRSS.Apache.conf -exec sed -r -i "/^\s*RemoteIP.*$/s/^/#/" {} \;
 	else
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -12,7 +12,7 @@ if [ -n "$LISTEN" ]; then
 fi
 
 if [ -n "$TRUSTED_PROXY" ]; then
-	if [ "$TRUSTED_PROXY" == "0" ]; then
+	if [ "$TRUSTED_PROXY" = "0" ]; then
 		# Disable RemoteIPHeader and RemoteIPTrustedProxy
 		find /etc/apache2/ -type f -name FreshRSS.Apache.conf -exec sed -r -i "/^\s*RemoteIP.*$/s/^/#/" {} \;
 	else
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -12,7 +12,7 @@ if [ -n "$LISTEN" ]; then
 fi
 
 if [ -n "$TRUSTED_PROXY" ]; then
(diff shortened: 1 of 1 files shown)


Docker/entrypoint.sh
```diff
@@ -12,7 +12,7 @@ if [ -n "$LISTEN" ]; then
 fi
 
 if [ -n "$TRUSTED_PROXY" ]; then
-	if [ "$TRUSTED_PROXY" -eq 0 ]; then
+	if [[ "$TRUSTED_PROXY" == 0 ]]; then
```
Alkarex: Please double-check that it works both with our Debian and Alpine images (which uses `ash`).
This line might have to be transformed to a standard POSIX equivalent, that does not require Bash.

Alkarex (changes requested):
POSIX

Docker/entrypoint.sh
mossroy: You're right, I had assumed bash.
I'll fix that

mossroy: This time I've tested with both Dockerfile and Dockerfile-Alpine

Alkarex (approved):

Alkarex:

Alkarex:
If you can, please test `freshrss/freshrss:edge`, which contains a few additional changes, before we release version 1.22.1.
https://github.com/FreshRSS/FreshRSS/blob/edge/CHANGELOG.md

mossroy:
I'm currently running it (to be precise, this one: https://hub.docker.com/layers/freshrss/freshrss/edge-arm/images/sha256-c62c0655f28d6e87f0433aeca0ad5fd51e14d3d2e2bf80f4e9360de35ed0909f).
I confirm the 2 issues I created are fixed in this image. And I don't see any visible regression.

Alkarex:
You mean that it works on ARM for you?

mossroy:
> You mean that it works on ARM for you?

Yes, I've deployed FreshRSS on ARM (armhf and arm64 architectures) since February 2023

Alkarex:
Sorry, I got confused this PR about proxies with our PRs about OpenID Connect (which is not yet working on ARM)

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/5792

Issue #5735, opened 2023-10-24
[BUG] Read articles won't display since 1.22.0 upgrade


Jucgshu:
**Describe the bug**
Hi guys, kudos on the new release. Since I upgraded to  1.22.0, I cannot display read articles when pressing the little red envelop icon, after I finished reading the articles.

**To Reproduce**
Simply press the display read / unread articles button.

**Expected behavior**
Display the already read articles as expected.

**Environment information (please complete the following information):**
- Device: Mobile & laptop
- OS: Archlinux
- Browser: Firefox 118.0.2
- FreshRSS version:  1.22.0
- Installation type: Docker (freshrss/freshrss)

Thanks !

Alkarex:
Hello,
Could you provide a screenshot illustrating the problem?

Jucgshu:
Sure, there you go (unread button just got pressed).

![Capture d’écran du 2023-10-24 20-08-23](https://github.com/FreshRSS/FreshRSS/assets/3890387/17b32522-bfc8-4a7f-9c89-f914cb2800c9)

Frenzie:
Pinging @math-GH

math-GH:
I am not sure that I understand 100% the reported issue.

I see, that with `Origine ` theme with dark mode (`Automatic dark mode (beta)` = `Auto`), it is not visible if the buttons are activated or not.

Could you please disable the dark mode (`Automatic dark mode (beta)` = `No`) and check your issue again?

Jucgshu:
Yup, same happens with the automatic dark mode disabled.

math-GH:
How is your setting here? 
![grafik](https://github.com/FreshRSS/FreshRSS/assets/1645099/e77206e6-23cb-44ae-a274-49e6b149346e)

Jucgshu:
Display unread articles.

math-GH:
Are you sure that you have unread articles? The main stream has 0 unread articles (no number is shown)
![grafik](https://github.com/FreshRSS/FreshRSS/assets/1645099/931f788a-da29-4462-bd20-cd7937296d4a)

Jucgshu:
Oww I missed the title of the issue. It is the *already* read articles I can't display since the upgrade when pressing the red icon. My bad. Will update 1st post.

Frenzie:
Oops, I misread. Sorry for the ping @math-GH ^_^

math-GH:
Then:
[...]

PR #5745, opened 2023-10-25
Fix origine dark mode

math-GH:
Ref #5735

Fix the active buttons in dark mode of Origine theme.
And it makes the error alert message better readable


Alkarex:
For future improvements, I find that the distinction between enabled/disabled buttons in Origine (dark AND light) could be further improved

PR #5719, opened 2023-10-23
Improved: Theme nord more improvement

math-GH:
Ref #5689

Changes proposed in this pull request:

some little improvements for the `Nord` theme that makes it a little bit better

PR #5689, opened 2023-10-07
improved: Nord theme

math-GH:
some improvements of the Nord theme.

PR #5754, opened 2023-10-26
More improvements/fixes of Nord theme

math-GH:
Ref #5689
Ref #5719

Some more improvements of `Nord` theme

### DIFF
--- a/p/themes/Nord/nord.css
+++ b/p/themes/Nord/nord.css
@@ -694,21 +694,27 @@ img.favicon {
 
 .item.feed.active {
 	background-color: var(--accent-bg);
-	font-weight: bold;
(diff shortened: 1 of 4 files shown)


math-GH:
Before:

1) in mobile view: the favicons had a grey background on the right hand side
2) the unread counter was not perfect positioned
grafik

After:
grafik

math-GH:
Before:
the color of the table border where missed, so it was just plain white.
grafik


After:
it has a dark color
grafik

Alkarex:
How much till its ready, more or less? I think we are ready for the release of 1.22.1

math-GH:
let's merge it here :)

PR #5756, opened 2023-10-26
SQL increase length of VARCHAR fields

### LINKED ISSUE #4986: [Feature] Increasing the maximum length of item titles
At the moment when the title of a feed item is longer than a certain length, it gets cut off.

I'd like to know if there's a way to increase the maximum allowed length of an item title. Is this already somehow possible? Or would it have to be a new feature?

**Additional context**
Well, for instance within Tumblr feeds there's very often items with titles that get cut in the middle because they're too long. [They do of course show up fully in the feed source.]

Alkarex:
Increase length of all fields, keeping the limits for:
* Unique indexes on UTF-8: 191 bytes for MySQL;
* Unique indexes on ASCII: 767 bytes for MySQL;
* Max URL for external tools: 32768 characters;
* Max VARCHAR: 65535 bytes for MySQL;

Follow-up of https://github.com/FreshRSS/FreshRSS/pull/5038
Fix https://github.com/FreshRSS/FreshRSS/issues/4986


Frenzie (approved):

amayatsky:
These `65535` scream CONST to me :D 
Also, that was really fast, thanks!
Sorry for bothering, but how is db migration achieved once the schema was changed?

Alkarex:
No migration at the moment. Will come soon. For now, you can export/import through CLI to a new user.
P.S. We had const in the past for other SQL-related values, and it was quite unreadable.

amayatsky:
As far as I understand, unless we run SQL strict mode the values will be just silently trimmed (alas with a warning)?
https://dev.mysql.com/doc/refman/8.0/en/char.html
Didn't check other engines though.
>export/import through CLI

Not an option for me on https://www.pikapods.com/ :(

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/5788

Alkarex:
Arg, I had forgotten that MySQL has a row limit of 65535 (sum of all columns). I will have to revert and/or change some fields

PR #5764, opened 2023-10-28
Prepare Alpine OIDC

Alkarex:
* Prepare syntax for OpenID Connect in Alpine.
* Update :newest Alpine development image to PHP 8.3
* Fix a little bug in test of OIDC_SCOPES


Alkarex:
```<IfModule !auth_openidc_module>``` still does not seem to work in Alpine at the moment (the module seems to be loaded but somehow not fully)

https://github.com/FreshRSS/FreshRSS/blob/e5ab91f4436770996818d8a28244eaa7b5ca79b1/Docker/FreshRSS.Apache.conf#L21

PR #5772, opened 2023-10-29
Export all environment variables to cron

### LINKED ISSUE #5770: [BUG] Database access not working for cron script
**Describe the bug**

Feeds are not updating and this error appears in the logs:

I recently upgraded to 1.22.0 and migrated from sqlite to postgres. Reading and marking existing posts read works, but the cron script is throwing errors about being unable to connect to the database:

```text
FreshRSS[2127]: [yaleman] [Sat, 28 Oct 2023 06:41:11 +1000] [error] --- Access to database is denied for ``: SQLSTATE[08006] [7] could not translate host name ":" to address: Name or service not known
```

And no updates are happening to feeds since the upgrade - until I exec into the container and run the actualize script manually.

**To Reproduce**

Using the following docker container:

```
    Image:           freshrss/freshrss:1.22.0-arm
    Image ID:        docker.io/freshrss/freshrss@sha256:108cfbb097b96f4b893752a257d39a676b122eacef4ba3527b5c769c6b0a1f96
```

Config db section:

```php
  'db' =>
  array (
    'type' => 'pgsql',
    'host' => getenv('FRESHRSS_POSTGRESQL_PORT_5432_TCP_ADDR').":".getenv('FRESHRSS_POSTGRESQL_PORT_5432_TCP_PORT'),
    'user' => getenv("DB_USER"),
    'password' => getenv("DB_PASSWORD"),
    'base' => getenv("DB_BASE"),
    'prefix' => false,
    'connection_uri_params' => '',
    'pdo_options' =>
    array (
    ),
  ),
```

The environment variables are confirmed to be working otherwise no data would appear - but the reader functionality and marking them read works.

**Expected behavior**

Feeds to be updated automatically with the cron timer.

**Screenshots**

n/a

- Device: linux
- OS: docker container
- Browser: n/a
- FreshRSS version: 1.22.0
- Database version: postgresql (docker postgres:15.4)
- PHP version: the one in the container (PHP 8.2.7 (cli) (built: Jun  9 2023 19:37:27) (NTS))
- Installation type: k8s, container

**Additional context**

The error would indicate that it's somehow loading the configuration but not getting the environment variables from the container? They need to come from there because it's running in kubernetes and are dynamically set on startup to point at the DB.

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5770
Note, the syntax complying with https://www.shellcheck.net/wiki/SC2002 does not seem to work in ash / Alpine

### DIFF
--- a/Docker/entrypoint.sh
+++ b/Docker/entrypoint.sh
@@ -34,13 +34,8 @@ if [ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ]; then
 fi
 
 if [ -n "$CRON_MIN" ]; then
-	(
-		echo "export TZ=$TZ"
-		echo "export COPY_LOG_TO_SYSLOG=$COPY_LOG_TO_SYSLOG"
-		echo "export COPY_SYSLOG_TO_STDERR=$COPY_SYSLOG_TO_STDERR"
-		echo "export FRESHRSS_ENV=$FRESHRSS_ENV"
-		echo "export DATA_PATH=$DATA_PATH"
-	) >/var/www/FreshRSS/Docker/env.txt
+	# shellcheck disable=SC2002
+	cat /proc/self/environ | tr '\0' '\n' | grep -vE '^(HOME|PATH|PWD|SHLVL|TERM|_)=' | sort -u | sed 's/^/export /' >/var/www/FreshRSS/Docker/env.txt
 	sed </etc/crontab.freshrss.default \
 		-r "s#^[^ ]+ #$CRON_MIN #" | crontab -
 fi


Alkarex:
I have made a test image `freshrss/freshrss:pr5772`

Docker/entrypoint.sh
```diff
@@ -34,13 +34,8 @@ if [ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ]; then
...
-	(
-		echo "export TZ=$TZ"
-		echo "export COPY_LOG_TO_SYSLOG=$COPY_LOG_TO_SYSLOG"
-		echo "export COPY_SYSLOG_TO_STDERR=$COPY_SYSLOG_TO_STDERR"
-		echo "export FRESHRSS_ENV=$FRESHRSS_ENV"
-		echo "export DATA_PATH=$DATA_PATH"
-	) >/var/www/FreshRSS/Docker/env.txt
+	# shellcheck disable=SC2002
+	cat /proc/self/environ | tr '\0' '\n' | grep -vE '^(HOME|PATH|PWD|SHLVL|TERM|_)=' | sort -u | sed 's/^/export /' >/var/www/FreshRSS/Docker/env.txt
 	sed </etc/crontab.freshrss.default \
```
Frenzie, line 39: This line is literally the syntax in question so something doesn't smell quite right. ;-)

Alkarex: Previous line, with cat

Frenzie: Yes, but this is the exact same situation. You could put a useless cat there.

So I'd assume it's something about `tr` then, not about Ash, though odd either way.

Alkarex: I think the difference is that in this PR, we access a process-dependent file

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/5792

Docker/entrypoint.sh
```diff
@@ -34,13 +34,8 @@ if [ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ]; then
...
 if [ -n "$CRON_MIN" ]; then
-	(
-		echo "export TZ=$TZ"
-		echo "export COPY_LOG_TO_SYSLOG=$COPY_LOG_TO_SYSLOG"
-		echo "export COPY_SYSLOG_TO_STDERR=$COPY_SYSLOG_TO_STDERR"
-		echo "export FRESHRSS_ENV=$FRESHRSS_ENV"
-		echo "export DATA_PATH=$DATA_PATH"
-	) >/var/www/FreshRSS/Docker/env.txt
+	# shellcheck disable=SC2002
+	cat /proc/self/environ | tr '\0' '\n' | grep -vE '^(HOME|PATH|PWD|SHLVL|TERM|_)=' | sort -u | sed 's/^/export /' >/var/www/FreshRSS/Docker/env.txt
```
Alkarex, line 38: Problematic for environment variables containing special characters such as `'`
Fixed in https://github.com/FreshRSS/FreshRSS/pull/5795

PR #5774, opened 2023-10-29
Improve table scrolling

math-GH:
When a table is to wide the table should be horizontal scroll-able and not the whole page.

If a table is scroll-able the thin scrollbars should be used (instead of the wider default scroll bars)

After:
grafik

grafik

- phtml + CSS (frss.css)

1. go to statics pages with table or go to display config page
2. have a thin screen
3. scroll the table horizontal

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/extension/index.phtml
+++ b/app/views/extension/index.phtml
@@ -45,7 +45,7 @@
 
 	<?php if (!empty($this->available_extensions)) { ?>
 		<h2><?= _t('admin.extensions.community') ?></h2>
-		<div class="table-wrapper">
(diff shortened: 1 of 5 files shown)


Frenzie (approved):
It's not quite clear to me why the diff is so large? But fine by me in principle.

math-GH:
> It's not quite clear to me why the diff is so large? But fine by me in principle.

because the table code had to be indented, because of the added `<div class="table-wrapper">`

Frenzie:
Right, mobile isn't showing any indentation here.

PR #5776, opened 2023-10-29
WIP Order Entries by date instead of insertion id

### LINKED ISSUE #2596: How do I tell FreshRSS to sort articles by publication date, not date added?
Every time I add new subscriptions, "Today" is filled with articles that are weeks or months old. And those new articles are in a nonsensical order (presumably some implementation detail):

image

I would like to be able to mark all old items as read with one click, the way I can with existing subscriptions.

Thanks.

Lcchy:
Closes #2596

- Return entries ordered by date and not by id in the SQL getter functions. 

As it is with this fix, the items seem properly ordered but the new page loading (when scrolling down) is repeating the first page with a slight offset. I suspect this has to do with how the $next_id or $firstId variables are set and used but I haven't been able to figure it out.

Any input or help would be more than welcome as I have no experience with this codebase nor Php. 

1. Add a new feed if you haven't done recently so that the entries are out of order in main view
2. Rebuild
3. Reload the main feed

Pull request checklist:

- [x ] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -1080,7 +1080,7 @@ SQL;
 			. ($type === 't' || $type === 'T' ? 'INNER JOIN `_entrytag` et ON et.id_entry = e.id ' : '')
 			. 'WHERE ' . $where
 			. $search
-			. 'ORDER BY e.id ' . $order
+			. 'ORDER BY e.date ' . $order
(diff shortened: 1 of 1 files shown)


Alkarex:
Sorry, this will not be accepted, as it will break several other things. There is a plan to support various sort orders, but that requires a bit more work than just changing the SQL query.

See https://github.com/FreshRSS/FreshRSS/issues/2596#issuecomment-1541934281

Lcchy:
I see, I suspected this to be the case. I will try and look into the steps you described, thanks!

PR #5782, opened 2023-10-30
New feature important feeds

Alkarex:
Take advantage of the feed priority information, to avoid missing updates from selected. important feeds (e.g. feeds with rare updates).

## Option
New option for the feed visibility / priority:
image

## UI
Important feeds can be styled differently (right now, double underline, but we might find something nicer later for each themes)
image

Also in aside feeds:
image

Also in feed management view:
image

# Behaviour
Even when the *mark as read on scroll* option is turned on, articles from important feeds will **not** be marked as read automatically during scroll (but can still be marked as read manually).
(We might make an option to tweak this behaviour, but I would rather avoid introducing a new options if possible)

image

image

* [x] We might do something similar for not marking as read important feeds during a "mark as as read" from the main view or [category view](https://github.com/FreshRSS/FreshRSS/pull/8067).


Alkarex:
I hope @math-GH will be so kind to polish a bit the UI (e.g. icon and alignment) when we get that far :-P

math-GH:
Interesting feature.

IMHO I would not separate the "important feeds" from the "main stream". It would make sense to have the important feeds also in the main stream, wouldn't it?
A checkbox would be better than a item of the select list.

And I would keep this order:
- main stream
- important feeds
- favorites
- labels
- categories

What do you think?

Alkarex:
The feeds would still be in the "main stream" and categories underneath.
It is a bit like a priority inbox in some systems like GMail. This PR is already working, except for calculating the number of unread articles, so you are welcome to give it a try to see how it behaves (I will post some more screenshots later).

Alkarex:
I have updated several things and reworked the description in the first post of this thread

math-GH:
What do you think about a pin icon for that feature?
grafik

Alkarex:
> What do you think about a pin icon for that feature?

It does look nice, and I kind of like it 🙂
However, I am not sure it best conveys the function. I was more thinking about something illustrating some kind of priority or urgency. In some email clients for instance, it is represented by an upward arrow like ⬆️

Frenzie:
An exclamation mark is also an obvious option.

math-GH:
I thought about both options (exclamation mark and "top" icon) too.

Why I did not chose one of them:
- exclamation mark: For me, it feels like a mental overload. There a lot of warnings (f.e. feed errors) and notifications so that one more icon that shout out for attention would not make me happy
- "top" icon: I do not like this icon because of the arrow that points to somewhere above.

The pin icon came into my mind because of the feature that articles from important feeds will not be marked as read automatically in the main stream (if the config is not enabled)

Alkarex:
Let's go with your option for now @math-GH and we can maybe get some feedback before the next release

math-GH:
* [x] Finding:
grafik

math-GH:
Does it make sense to have the "Important feeds" menu item if there is no feed assigned to the "Important feed" channel?

math-GH:
* [x] Toggeling the (un)read icon: The counter of the main stream is affected and not the "Important feed" counter
grafik

Alkarex:
> Does it make sense to have the "Important feeds" menu item if there is no feed assigned to the "Important feed" channel?

It is indeed cleaner to not show *important feeds* when there is no feed in that category. On the flip side, it makes this feature difficult to discover. (edit: same for *my labels*, which I consider always showing, also when there are no unread articles)

I will look at the counting problems

Alkarex:
Fixed :-)

Alkarex:
Any comment? If nothing blocks, I will soon merge to get wider testing and feedback :-)

math-GH:
Is it ok to collect feedback here? Or should I open a new issue or a discussion?

Alkarex:
As you prefer, @math-GH , I do not mind :-)

math-GH:
First feedback after using it for some days:

The mark as read function is not smart enough, because the user can mark the articles as read only in the "important feeds" channel.
I would suggest that the feed channel of an important feed needs an improvement: Right now, the user cannot mark the articles as read. Here it should be irrelevant if it is an "important feed" or not and the "mark as read" button & scrolling should work as usual.

In the category channel and main stream, there should be an addtional button something like "mark important feed articles as read"

What does this icon mean? It is just an image.
grafik

Alkarex:
> Right now, the user cannot mark the articles as read.

Articles from important streams can be marked as read individually by opening them, or by clicking the envelope.
They can also be marked as read when clicking "mark all as read" from the "Important feeds" view as well as from a feed view.

> Here it should be irrelevant if it is an "important feed" or not and the "mark as read" button & scrolling should work as usual

I am open to that; this was actually my first implementation, and I personally like it. I only changed to the current behaviour as an effort to simplify the understanding of the behaviour. More opinions welcome. I can make a PR.
Edit: Done in https://github.com/FreshRSS/FreshRSS/pull/5906

> In the category channel and main stream, there should be an addtional button something like "mark important feed articles as read"

I would rather avoid adding more buttons. This can be done from the "Important feeds" view or from the feeds view.

> What does this icon mean? It is just an image.

It was just an effort to help finding the relevant settings by repeating some of the same graphics. Changes welcome

Alkarex:
@math-GH Better? https://github.com/FreshRSS/FreshRSS/pull/5906

PR #5788, opened 2023-10-30
Rework SQL field length

Alkarex:
* Partial revert of https://github.com/FreshRSS/FreshRSS/pull/5756 to account for MySQL's total row size limit of 65535.
* We still increase the size of the fields, but not by as much
* Auto-updates for PostgreSQL and MySQL / MariaDB. Not possible for SQLite


Alkarex:
Arg, I had forgotten that MySQL has a row limit of 65535 (sum of all columns). I will have to revert and/or change some fields

Alkarex:
Merge a bit fast to avoid too much variation in the instances running `edge`

Issue #5707, opened 2023-10-15
[BUG]: Recurring 504 Timeouts and High Resource Usage with MariaDB Operations


Emporea:
I just have to create a new issue since its still problematic for me..

**Description**:
Upon utilizing FreshRSS with MariaDB (version 10.6) under the Docker edge tag, there’s still an evident high resource usage and subsequent 504 timeouts during specific operations. The FreshRSS database CPU utilization spikes to approximately only 20% (should use more in my opinion since this states its not really optimized) and high IO disk usage, which reaches around 1GB/s, have been observed to last for a few minutes during several actions.

**Problematic Actions**:
1. **Actualize Script Execution**: Triggering the script manifests in a substantial resource strain for ~ 4 minutes.
2. **Adding New Feeds**: While the feed does get added, the operation culminates in a 504.
3. **Viewing Idle Feeds**: Invariably results in a 504 and is never accessible.
4. **General Feed Statistics**: Unreachable due to a perpetual 504.

**Insights and Potentials**:
Although the functions eventually serve their intended purpose (aside from viewing idle feeds and statistics), the undue strain and resulting timeouts mar the usability and efficiency of the system. A deductive observation points towards the possibility that the database queries still might not be wholly optimized for use with MariaDB, potentially requiring a revisit and refinement to enhance performance, especially in the aforementioned contexts.
I hope I can be of any help.by providing logs or testing development builds.

**Previous Discussion**:
It’s notable that the 504 issue has been broached before, its been solved for "Marking all News as Read" which is now done in a matter of seconds, yet it still pervasively persists through these other actions. Alleviating this would drastically improve the overall user experience and system reliability.

**Potential Resolution**:
Optimizing how MariaDB processes queries, especially during intensive operations such as those highlighted, might hold the key to alleviating the resource burdens and the subsequent 504 timeouts.

Any insights, assistance, or pointers on this would be massively appreciated!

Kind Regards.

Alkarex:
Hello,
Please provide the usual information such as your FreshRSS version (please try with our edge version), how you installed FreshRSS (precise image in case of Docker install), size of your DB, etc.

Emporea:
Freshrss version:  1.21.1-dev
database size: ~2.5GB
new articles per day: ~6000

Never deleting articles (purge policy disabled, purge exceptions enabled)

docker-compose.yml:
```
freshrss_db:
    image: mariadb:10.6
    container_name: freshrss_db
    restart: unless-stopped
[...]

PR #5648, opened 2023-09-10
SQL: Same updateCacheUnreads for all DBs

Alkarex:
Use same SQL update request for MySQL / MariaDB than the one we already used for PostgreSQL / SQLite (i.e. using a sub-query).

Testing on a DB of 688MB with 270k entries, 199 feeds, 19 categories, using MySQL 8.1.0 and MariaDB 11.1.

The new SQL update using a sub-query took in average 0.02s, while the old SQL update using a join took in average 0.05s. SQL cache was properly invalidated between each run. The new SQL request is thus about twice faster on that test. `EXPLAIN` also shows an intrinsic advantage, and it might lead to fewer locks.

Another advantage of the SQL update using a sub-query is that it works identically in PostgreSQL, SQLite, MariaDB, MySQL, so we do need different versions anymore.

Contributes to https://github.com/FreshRSS/FreshRSS/issues/5008#issuecomment-1709755370

Edit: But with a `USE INDEX` hint for MySQL/MariaDB.

PR #5793, opened 2023-10-31
Update DB recommendations

Alkarex:
Favour PostgreSQL, based on tests such as:
https://github.com/FreshRSS/FreshRSS/pull/5648#issuecomment-1715843198
https://github.com/FreshRSS/FreshRSS/issues/5707

SQLite is generally very fine, although with the challenge that the DB model cannot easily be updated https://www.sqlite.org/lang_altertable.html#why_alter_table_is_such_a_problem_for_sqlite
(We should ship a command-line update mechanism)

I have verified that FreshRSS still works with `mysql:5.5.62`
https://hub.docker.com/layers/library/mysql/5.5.62/images/sha256-d404d78aa797c87c255e5ae2beb5d8d0e4d095f930b1f20dc208eaa957477b74?context=explore


Frenzie (approved):
What's that about SQLite being built into PHP?

Alkarex:
> What's that about SQLite being built into PHP?

I am not sure how to formulate that better. It does not require installing anything else like a DB server, as the support comes from PHP (with the appropriate extensions) itself.

Frenzie:
Maybe something more in the vein of easy setup?

Alkarex:
It is not even easy setup, it is no setup at all :-)

Frenzie:
Eh, you have to apt/apk install it and you have to set it up in FreshRSS. :-)

Alkarex:
> Eh, you have to apt/apk install it and you have to set it up in FreshRSS. :-)

Right, indeed for the package. I was more thinking from the Docker perspective.

I was curious of how much that package was adding (on `debian:12-slim` after Apache and PHP are installed):

```console
apt install php-sqlite3
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  php8.2-sqlite3
The following NEW packages will be installed:
  php-sqlite3 php8.2-sqlite3
0 upgraded, 2 newly installed, 0 to remove and 0 not upgraded.
Need to get 33.8 kB of archives.
After this operation, 145 kB of additional disk space will be used.
```

Regarding the setup in FreshRSS, there is nothing to do (besides picking SQLite, of course) :-)

image

Frenzie:
> I was more thinking from the Docker perspective.

Then maybe "included in Docker image"?

> Regarding the setup in FreshRSS, there is nothing to do (besides picking SQLite, of course) :-)

That's what I meant, yes. :-D

Alkarex:
I have changed a bit the explanations. SQLite being included in our Docker images is also mentioned in the Docker documentation, so that is probably enough.

PR #5802, opened 2023-11-01
Fix: Update page: Buttons

math-GH:
Regression from #5647

The needed CSS code were delivered just in Origine theme, but it is needed in all themes.

Before:
The "Updating..." button is shown, when an update is available but not yet started
grafik

After:
grafik

1. go to update page
2. search for an update
3. if there is an update available: no "updating..." button is shown before you started the update process

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/themes/Origine/origine.css
+++ b/p/themes/Origine/origine.css
@@ -291,24 +291,6 @@ form th {
 	text-decoration: none;
 }
 
-.btn-state1.hide {
-	display: none;
-}
-
-.btn-state2 {
-	display: none;
-}
-
-.btn-state2.show {
-	display: inline-block;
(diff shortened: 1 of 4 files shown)


Alkarex:
I just saw that we still have the same bug (tested with git) because the class is `"btn btn-state2"` and in my test with Firefox, btn seems to win:

image

math-GH:
I cannot reproduce it now.

Alkarex:
@math-GH Are you trying with a git-based setup?

math-GH:
ok, got it now. It is independent of the git-based setup.
The fix will follow.

PR #5805, opened 2023-11-01
document username:password@ syntax in proxy_help

### LINKED ISSUE #5497: [Feature] Proxy username and password
Proxies like SOCKS support usernames and passwords, which the current user interface doesn't provide a way to supply. For some configurations, this is necessary to use the proxy at all. In the case of Tor, it's a quick and easy way to force independent circuits (for FreshRSS, this would mean you could have two RSS feeds from the same domain coming from independent connections, whereas right now, the server hosting example.com/foo.rss and example.com/bar.rss could tell if requests to each are from the same connection).

jtracey:
Closes #5497

- adds `username:password@` to the `proxy_help` string, so that users are aware this is possible.

An alternative would be to add explicit fields for these, then just add them server-side with string manipulation, but the solution I propose here seems good enough to me, and is less likely to introduce bugs.

It's not clear to me whether the non-english strings should have `// TODO` comments added back, just let me know if they should.

1. Navigate to a page with proxy settings (i.e., adding or modifying a feed).
2. Observe the new help string, and set the proxy address accordingly. 
3. Connect to the feed via the proxy, now with a username and password.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


app/i18n/en/sub.php
```diff
@@ -144,7 +144,7 @@
 			'normal' => 'Show in its category',
 		),
 		'proxy' => 'Set a proxy for fetching this feed',
-		'proxy_help' => 'Select a protocol (e.g: SOCKS5) and enter the proxy address (e.g: <kbd>127.0.0.1:1080</kbd>)',
+		'proxy_help' => 'Select a protocol (e.g: SOCKS5) and enter the proxy address (e.g: <kbd>username:password@127.0.0.1:1080</kbd>)',
```
Alkarex: I am afraid that some users might think that providing the username and password is mandatory, so maybe we could indicate both syntax, like:

```suggestion
		'proxy_help' => 'Select a protocol (e.g: SOCKS5) and enter the proxy address (e.g: <kbd>127.0.0.1:1080</kbd> or <kbd>username:password@127.0.0.1:1080</kbd>)',
```

Alkarex:
Thanks for the PR!
You can mark the other languages with `// DIRTY` so that a translator might take care of it later

app/i18n/en/sub.php
jtracey: Good point, done.

app/i18n/fr/sub.php
```diff
@@ -144,7 +144,7 @@
 			'normal' => 'Afficher dans sa catégorie',
 		),
 		'proxy' => 'Utiliser un proxy pour télécharger ce flux',
-		'proxy_help' => 'Sélectionner un protocole (ex : SOCKS5) et entrer l’adresse du proxy (ex. : <kbd>username:password@127.0.0.1:1080</kbd>)',
+		'proxy_help' => 'Sélectionner un protocole (ex : SOCKS5) et entrer l’adresse du proxy (ex. : <kbd>127.0.0.1:1080</kbd> or <kbd>username:password@127.0.0.1:1080</kbd>)',	// DIRTY
```
```suggestion
		'proxy_help' => 'Sélectionner un protocole (ex : SOCKS5) et entrer l’adresse du proxy (ex. : <kbd>127.0.0.1:1080</kbd> ou <kbd>utilisateur:mot-de-passe@127.0.0.1:1080</kbd>)',
```

Alkarex:

PR #5089, opened 2023-02-10
Modernize Constants and use new constant for string 'currentUser'

ColonelMoutarde:
Changes proposed in this pull request:

- create constant for  'currentUser'

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #5808, opened 2023-11-02
add multi arch docker build

### LINKED ISSUE #4089: [Docker] Support for raspberry pi OS 64 bit
Tried both `freshrss:arm` and `freshrss:arm` and `:latest-arm`, both get:

```
WARNING: The requested image's platform (linux/arm) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```
Searching in [docker-tags](https://hub.docker.com/r/freshrss/freshrss/tags) it doesn't seem to be an ARM64 version.

b-reich:
Closes #4089

- use buildx for docker building
- use github action for building

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


b-reich:
Is publishing on github actions wantend?

b-reich:
https://github.com/b-reich/FreshRSS/actions/runs/6733089164/job/18301035008

looks like the build itself works

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,70 @@
...
+      - '[0-9]+.[0-9]+.[0-9]+'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build-docker-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
Alkarex:
```suggestion
        uses: actions/checkout@v4
```

Alkarex:
Thanks, that looks pretty good already 👍🏻 
I am having a look

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,70 @@
...
+
+on:
+  push:
+    branches:
+      - edge
+      - latest
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+  pull_request:
+  workflow_dispatch:
```
Alkarex: I think we should have on `release` instead

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,70 @@
+name: build-images
+
+on:
+  push:
+    branches:
+      - edge
+      - latest
```
Alkarex: Only for edge branch, not latest (which is done by `release` instead)

Alkarex:
For reference, the current Docker Hub build configuration:

image

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,70 @@
...
+            type=edge
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/0.') }}
+
+      - name: Build and push
+        id: docker_build_release
+        uses: docker/build-push-action@v5
+        with:
+          file: Docker/Dockerfile
```
Alkarex: It would be nice to also build `Docker/Dockerfile-Alpine`

.github/workflows/build-images.yml
b-reich: I dont mind, was just an idea

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,70 @@
...
+          platforms: linux/arm/v7,linux/arm64,linux/amd64
+          push: ${{ github.ref == 'refs/heads/latest' || startsWith(github.ref, 'refs/tags/') }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+  push-docker-readme:
```
Alkarex: This part is great, as I have not been able to push the correct Readme using solely Docker Hub...
We have a Docker readme at https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/README.md but we could make a better page

b-reich:
> For reference, the current Docker Hub build configuration:
> 
> image

in the end we should only have something like latest and latest-alpine i think. No need to tag the arm images separately. This is kinda a breaking change than.

EdJoPaTo:
Maybe the Dockerfile Labels should also be adapted for this. The `docker/metadata-action` sets a bunch of labels by itself so we shouldn't duplicate ourselves here?

https://github.com/FreshRSS/FreshRSS/blob/8b2f7848eb1054ee26048e3ffe76855410286407/Docker/Dockerfile-Alpine#L23-L33

.github/workflows/build-images.yml
b-reich: Done

.github/workflows/build-images.yml
b-reich: Done

.github/workflows/build-images.yml
b-reich: Using the Docker Readme now

.github/workflows/build-images.yml
EdJoPaTo: Just a random question, maybe there is a good answer here I am not seeing: Why multiple prebuilt images? Why not concentrate on one image and test that one instead of multiple ones? The image size difference is not that significant so personally I would concentrate on one image. (But thats personal taste, maybe you have different opinions and thats totally fine)

b-reich:
> Maybe the Dockerfile Labels should also be adapted for this. The `docker/metadata-action` sets a bunch of labels by itself so we shouldn't duplicate ourselves here?
> 
> https://github.com/FreshRSS/FreshRSS/blob/8b2f7848eb1054ee26048e3ffe76855410286407/Docker/Dockerfile-Alpine#L23-L33

The generated one from the metadata action are not used at the moment. this could be change @Alkarex what do you think?

Alkarex:
> No need to tag the arm images separately.

Indeed. I think it would be an acceptable breaking change.

> Why multiple prebuilt images?

Providing a Debian and an Alpine version is quite common. There are pros and cons with both (for instance build speed for Alpine, among other things), and having an alternative can be helpful (e.g. https://github.com/FreshRSS/FreshRSS/discussions/5742 )

I will come back for the other questions

b-reich:
https://github.com/b-reich/FreshRSS/actions/runs/6738089530/job/18316933656

the github action is nearly finish I think....
How to handle the tags is a thing now.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,86 @@
+name: build-images
+
+on:
+  push:
+    branches:
+      - edge
+  release:
+  workflow_dispatch:
```
Alkarex: Is that used?

Alkarex:
> The generated one from the metadata action are not used at the moment. this could be change @Alkarex what do you think?

Yes, it would probably be better to use the ones from the Action, indeed

.github/workflows/build-images.yml
b-reich: You mean the workflow_dispatch? 
I like the option to start the pipeline per hand.
Do you think it's not needed?

What do you prefer the on release or on tag? (The tag also happen on a release I think)

Alkarex: I prefer to use `release`, as this is the way new official versions are produced, and yes, I also believe a release will also trigger a tag event.
Yes, I meant the `workflow_dispatch`. Yes, starting the pipeline by hand is always handy (haha), but I am not sure whether `workflow_dispatch` is needed for that.

b-reich: Yeah its needed to define there
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

Docker/Dockerfile
```diff
@@ -23,18 +23,6 @@ ARG FRESHRSS_VERSION
...
-	org.opencontainers.image.authors="Alkarex" \
-	org.opencontainers.image.description="A self-hosted RSS feed aggregator" \
-	org.opencontainers.image.documentation="https://freshrss.github.io/FreshRSS/" \
-	org.opencontainers.image.licenses="AGPL-3.0" \
-	org.opencontainers.image.revision="${SOURCE_BRANCH}.${SOURCE_COMMIT}" \
-	org.opencontainers.image.source="https://github.com/FreshRSS/FreshRSS" \
-	org.opencontainers.image.title="FreshRSS" \
-	org.opencontainers.image.url="https://freshrss.org/" \
-	org.opencontainers.image.vendor="FreshRSS" \
-	org.opencontainers.image.version="$FRESHRSS_VERSION"
```
Alkarex: I would prefer to keep the static ones here, so that they exist when the builds are done locally, e.g. like
https://github.com/FreshRSS/FreshRSS/blob/8b2f7848eb1054ee26048e3ffe76855410286407/Docker/freshrss/docker-compose.yml#L12-L15

`authors` can be removed

By the way, if anyone knows a way to populate `revision` and/or `version` even in those more manual scenarios, let me know

b-reich: not sure what happen when you use the github action ones and the static ones.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,91 @@
...
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+      # - name: Login to DockerHub
+      #   uses: docker/login-action@v3
+      #   with:
+      #     username: ${{ secrets.DOCKERHUB_USERNAME }}
+      #     password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      # - name: Login to GitHub Container Registry
```
Alkarex: Any comments about how pros and cons of (also) using the GitHub Container Registry?
Probably seems overkill for now?

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,91 @@
...
+      uses: docker/setup-qemu-action@v3
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+      # - name: Login to DockerHub
+      #   uses: docker/login-action@v3
+      #   with:
+      #     username: ${{ secrets.DOCKERHUB_USERNAME }}
+      #     password: ${{ secrets.DOCKERHUB_TOKEN }}
```
Alkarex: I will take care or that

.github/workflows/build-images.yml
b-reich: For podman user?
And Dockerhub had in the pass some weird decisions I think.

Personaly I prefer the github registry

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,91 @@
...
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
+
+  push-docker-readme:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Update repo description
+      uses: peter-evans/dockerhub-description@v3
```
Alkarex: For third-party Actions, I would prefer to pin to a specific git hash instead of `@v3`, to increase trust

Docker/Dockerfile
Alkarex: Let's try to remove the static ones from GitHub Actions, so that we can see whether that is sufficient or whether they need to be overridden again

.github/workflows/build-images.yml
Alkarex: Ok. Let's keep it but comment out for now, so we can easily activate if needed.
I am not familiar with the use-case: Can't Podman users pull from Docker Hub?

Docker/Dockerfile
b-reich: I found some docu about. 
passing labels at build time will overwrite the static ones

.github/workflows/build-images.yml
b-reich: It can, but the default registry (dockerhub) is not set.

.github/workflows/build-images.yml
b-reich: I put it into a seperate action and pin it to a commit

.github/workflows/build-images.yml
b-reich: I enable it now.

.github/workflows/build-images.yml
EdJoPaTo: Newer podman versions try dockerhub on unspecified image tags. Personally I don't think it matters much as its the same image

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,77 @@
...
+              onlatest=false
+          - name: Alpine
+            file: Docker/Dockerfile-Alpine
+            flavor: |
+              latest=auto
+              suffix=-alpine
+              onlatest=true
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
```
EdJoPaTo: The checkout can be removed from the action as long as `docker/build-push-action` has no context option defined. It will pull the repo by itself and has no need for a checkout.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,77 @@
...
+      #     ghcr.io/${{ github.repository }}
+        tags: |
+          type=edge
+          type=semver,pattern={{version}}
+          type=semver,pattern={{major}}.{{minor}}
+          type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/0.') }}
+        labels: |
+          org.opencontainers.image.description="A self-hosted RSS feed aggregator"
+          org.opencontainers.image.documentation="https://freshrss.github.io/FreshRSS/"
+          org.opencontainers.image.url="https://freshrss.org/"
```
EdJoPaTo: Personally I do the login as a last step before the build for two reasons.
The login can be time critical, if I remember correctly it only lasts for an hour for the ghcr.

Also I order my actions based on what they generate for the following actions. The fewer information the following actions have the better. A login is something that could be used otherwise so expose as few actions afterwards with it as possible. Probably irrelevant most of the time but I still try to follow that when creating workflows.

```suggestion
    - name: Metadata
      id: meta
      uses: docker/metadata-action@v5
      with:
        flavor: ${{ matrix.flavor }}
        images: |
          docker.io/freshrss/freshrss
      #     ghcr.io/${{ github.repository }}
        tags: |
          type=edge
          type=semver,pattern={{version}}
          type=semver,pattern={{major}}.{{minor}}
          type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/0.') }}
        labels: |
          org.opencontainers.image.description="A self-hosted RSS feed aggregator"
          org.opencontainers.image.documentation="https://freshrss.github.io/FreshRSS/"
          org.opencontainers.image.url="https://freshrss.org/"

    - name: Login to DockerHub
      uses: docker/login-action@v3
      with:
        username: ${{ secrets.DOCKERHUB_USERNAME }}
        password: ${{ secrets.DOCKERHUB_TOKEN }}

    # - name: Login to GitHub Container Registry
    #   uses: docker/login-action@v3
    #   with:
    #     registry: ghcr.io
    #     username: ${{ github.repository_owner }}
    #     password: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,77 @@
+name: build-images
+
+on:
+  push:
+    branches:
+      - edge
+  release:
+  workflow_dispatch:
+
```
EdJoPaTo: Restrict (or add) permissions to the job. When the ghcr is used the permission to write to the packages is needed. Always specify as less as possible here. The restricted token is also the default for new repos so it only works without in older repos. See [token permission documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)
```suggestion

permissions:
  contents: read
  packages: write

```

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,77 @@
+name: build-images
+
+on:
+  push:
+    branches:
+      - edge
```
EdJoPaTo: Personally I like to build on all branches and just not push the result to the registry then. It ensures the container still builds. But that's probably a decision for the maintainers.

.github/workflows/build-images.yml
b-reich: the token is not used add the moment. But a good addition

EdJoPaTo: Maybe add the `packages: write` commented out like the ghcr login part is currently then?

.github/workflows/build-images.yml
Alkarex: I am concerned by our energy footprint, so I try to limit the automated builds to a minimum.
When working on making changes to one of the Docker images, they need to be tested locally anyway, so auto-build does not add a lot of value. And if the builds fail once in a while, it is not a big deal as it does not affect production and can be fixed or reverted with another PR.

.github/workflows/push-dockerhub-readme.yml
```diff
@@ -0,0 +1,23 @@
+name: push readme to dockerhub
```
Alkarex: I think I would better like to have it back in the `build-images.yml`, mostly to be more resource efficient (avoid having to spin a new VM and git clone once more)

b-reich: It was a seperated job
so it always create a new context on another machine

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,78 @@
...
+      #     ghcr.io/${{ github.repository }}
+        tags: |
+          type=edge
+          type=semver,pattern={{version}}
+          type=semver,pattern={{major}}.{{minor}}
+          type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/0.') }}
+        labels: |
+          org.opencontainers.image.description="A self-hosted RSS feed aggregator"
+          org.opencontainers.image.documentation="https://freshrss.github.io/FreshRSS/"
+          org.opencontainers.image.url="https://freshrss.org/"
```
Alkarex: So what was the conclusion? Can't those static labels be removed, as they are already defined in the Dockerfile?

.github/workflows/push-dockerhub-readme.yml
Alkarex: Can't we just add the steps to the existing job?

Alkarex: (It would be fine to react to the same events we already have)

Alkarex: But good point, with your current suggestion, this job might not have to run that often.
Let's keep it that way then.

.github/workflows/push-dockerhub-readme.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  push:
+    paths:
+      - Docker/README.md
+    branches:
+      - latest
```
Alkarex:
```suggestion
      - edge
```

Branch would be `edge` though

Alkarex:
Excellent job, this looks really nice 👍🏻 
We need a little update to https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/README.md#our-docker-image-variants
Next step from my side will be to disable the `edge` builds on Docker Hub to see how well it behaves, and then later disable the other tags.

.github/workflows/build-images.yml
b-reich: This overwrite the metadata ones. And the metadata ones overwrite the statics ones.
So we benefit from the right versions label (and other labels)

b-reich:
@Alkarex please take a look on the new readme

Docker/README.md
```diff
@@ -66,13 +66,16 @@ The [tags](https://hub.docker.com/r/freshrss/freshrss/tags) correspond to FreshR
 
 * `:latest` (default) is the [latest stable release](https://github.com/FreshRSS/FreshRSS/releases/latest)
 * `:edge` is the rolling release, same than our [git `edge` branch](https://github.com/FreshRSS/FreshRSS/tree/edge)
-* `:x.y.z` are [specific FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases)
-* `:arm` or `:*-arm` are the ARM `arm32v7` versions (e.g., for Raspberry Pi).
+* `:x.y.z` tags correspond to [specific FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases), allowing you to target a precise version for deployment
+* `:x.y` tags are tied to a specific major version and minor version number. For example, `:12.2` will automatically receive updates for any `12.2.x` releases, but will not update to `12.3.x`
+* `:x` tags track the latest release within a major version series. For instance, `:12` will update to include any `12.x` releases, but will exclude versions beyond `13.x`
+* `*-alpine` use apline as base-image instead debian
```
Alkarex:
```suggestion
* `*-alpine` use Linux Alpine as base-image instead of Debian
```

Docker/README.md
```diff
@@ -66,13 +66,16 @@ The [tags](https://hub.docker.com/r/freshrss/freshrss/tags) correspond to FreshR
 
 * `:latest` (default) is the [latest stable release](https://github.com/FreshRSS/FreshRSS/releases/latest)
 * `:edge` is the rolling release, same than our [git `edge` branch](https://github.com/FreshRSS/FreshRSS/tree/edge)
-* `:x.y.z` are [specific FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases)
-* `:arm` or `:*-arm` are the ARM `arm32v7` versions (e.g., for Raspberry Pi).
+* `:x.y.z` tags correspond to [specific FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases), allowing you to target a precise version for deployment
+* `:x.y` tags are tied to a specific major version and minor version number. For example, `:12.2` will automatically receive updates for any `12.2.x` releases, but will not update to `12.3.x`
+* `:x` tags track the latest release within a major version series. For instance, `:12` will update to include any `12.x` releases, but will exclude versions beyond `13.x`
```
Alkarex:
```suggestion
* `:x.y` tags are tied to a specific major version and minor version number. For example, `:1.23` will automatically receive updates for any `1.23.x` releases, but will not update to `1.24.x`
* `:x` tags track the latest release within a major version series. For instance, `:1` will update to include any `1.x` releases, but will exclude versions beyond `2.x`
```

I have changed to examples closer to what we have in FreshRSS

.github/workflows/build-images.yml
Alkarex: I am still confused: without these lines, what would `org.opencontainers.image.description` contain? Wouldn't it have what was defined in Dockefile?

Docker/README.md
```diff
@@ -66,13 +66,16 @@ The [tags](https://hub.docker.com/r/freshrss/freshrss/tags) correspond to FreshR
...
+* `*-alpine` use apline as base-image instead debian
+* Our Docker images are designed with multi-architecture support, accommodating a variety of Linux platforms including `linux/arm/v7`, `linux/arm64`, and `linux/amd64`.
   * For other platforms, see the [custom build section](#build-custom-docker-image)
 
 ### Linux: Debian vs. Alpine
 
-Our default image is based on [Debian](https://www.debian.org/). We offer an alternative based on [Alpine](https://alpinelinux.org/) (with the `:alpine` or `*-alpine` tag suffix).
```
Alkarex, line 75: Don't we have the `freshrss/freshrss:alpine` tag anymore?

b-reich: when I am correct we have `latest-alpine`or `v2-alpine` and so on

Alkarex:
Ah, for the record, I was a bit confused because there were some actions triggered by pushes to this PR:
* https://github.com/b-reich/FreshRSS/actions/runs/6756095787
But it is because the name of @b-reich 's PR branch is also `edge`

https://github.com/orgs/community/discussions/26409#discussioncomment-3251822

Docker/README.md
b-reich: grafik
I testet it... there is something odd

.github/workflows/build-images.yml
b-reich: grafik

b-reich: when I dont set in manualy for the actions its taken from the github repo
grafik

b-reich: The metadata action set it to `A free, self-hostable news aggregator… `I overwrite it to `A self-hosted RSS feed aggregator`because the static ones in the dockerfile would be overwritten by the metadata ones.

Docker/README.md
b-reich: grafik
working now

.github/workflows/push-dockerhub-readme.yml
```diff
@@ -0,0 +1,23 @@
...
+on:
+  push:
+    paths:
+      - Docker/README.md
+    branches:
+      - edge
+  workflow_dispatch:
+
+jobs:
+  push-docker-readme:
```
Alkarex:
```suggestion
  push-docker-readme:
    if: github.repository_owner == 'FreshRSS'
```

https://github.com/orgs/community/discussions/26409#discussioncomment-3251822
I do not know whether that condition can be put higher up, in the `on:` section

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,76 @@
...
+      - edge
+  release:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  # packages: write
+
+jobs:
+  build-container-image:
```
Alkarex:
```suggestion
  build-container-image:
    if: github.repository_owner == 'FreshRSS'
```

https://github.com/orgs/community/discussions/26409#discussioncomment-3251822

Maybe that could be done smarter. It would be OK to let the build occur, as long as there is no push attempt

Docker/README.md
Alkarex: So `freshrss:latest-alpine`, but no `freshrss:alpine`?

.github/workflows/build-images.yml
b-reich: I add some logic to prevent the push on non upstream repos

.github/workflows/push-dockerhub-readme.yml
b-reich: already implemented with the other push logic

.github/workflows/build-images.yml
Alkarex: Ok, thanks for the investigation 👍🏻

Docker/README.md
b-reich: yes

Alkarex: Can you think of a way to have `freshrss:alpine` ? (and maybe no `freshrss:latest-alpine` at all)
It seems to be the more common way, e.g. https://hub.docker.com/layers/library/httpd/alpine/images/sha256-e4e78032a4c6375e4e697d68f9e70306bce66ccdf09659d4c06b029d5fc60d41?context=explore

b-reich: done, but I find `latest-alpine` whould be "nice"

b-reich: grafik

Alkarex:
It looks almost good now :-)
I can spot one thing, which seems to be missing: we used to have the FreshRSS version in the `org.opencontainers.image.version` metadata:

https://github.com/FreshRSS/FreshRSS/blob/ecf1585d74c44954720acddce02d3827c93a06df/Docker/Dockerfile#L36

https://github.com/FreshRSS/FreshRSS/blob/ecf1585d74c44954720acddce02d3827c93a06df/Docker/hooks/build#L4

b-reich:
> It looks almost good now :-) I can spot one thing, which seems to be missing: we used to have the FreshRSS version in the `org.opencontainers.image.version` metadata:
> 
> https://github.com/FreshRSS/FreshRSS/blob/ecf1585d74c44954720acddce02d3827c93a06df/Docker/Dockerfile#L36
> 
> https://github.com/FreshRSS/FreshRSS/blob/ecf1585d74c44954720acddce02d3827c93a06df/Docker/hooks/build#L4

Do you think the default from the metadata action didnt feed the needs? It kinda match the labels now.
alpine:
grafik
debian:
grafik

Alkarex:
> Do you think the default from the metadata action didnt feed the needs? It kinda match the labels now.

It is very good for the releases, but we lack the FreshRSS version in the case of `edge`, don't we?

Alkarex:
Just for future inspiration (not for this PR), here is the default template for Docker Publish:
https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml
It is using e.g. https://github.com/sigstore/cosign , which I am not familiar with, but which sounds interesting

b-reich:
@Alkarex its look like the cuurrent docker builds at the moment have a broken version label.
grafik

Yes the edge build have the version `edge` or `edge-alpine`. But its a rolling release. So its kinda personal taste I think.

Alkarex:
> its look like the cuurrent docker builds at the moment have a broken version label.

Indeed 😱
Edit: That was a regression due to https://github.com/FreshRSS/FreshRSS/pull/5089

b-reich:
@Alkarex  you need to checkout the repo when you try to retrive the version

Alkarex:
> you need to checkout the repo when you try to retrive the version

Thanks. Right, I had read a bit too fast that comment https://github.com/FreshRSS/FreshRSS/pull/5808#discussion_r1381382672

As you can see, I am experimenting with an arg-based approach for labels. What do you think?

Alkarex:
Hum, I was hopping that `docker/metadata-action` could be skipped, but `tags: ${{ matrix.tags }}` does not seem to work directly...

Alkarex:
The output looks good to me now, although I would have liked to simplify a bit more if we could avoid `docker/metadata-action`.
Final thoughts @b-reich ?

EdJoPaTo:
Out of curiousity: why would you like to avoid using metadata-action? it seems helpful to me for generating the v1, v1.2, v1.2.3 tags and so on. Maybe I overlook something?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,94 @@
...
+      uses: docker/metadata-action@v5
+      with:
+        flavor: ${{ matrix.flavor }}
+        images: |
+          docker.io/freshrss/freshrss
+      #     ghcr.io/${{ github.repository }}
+        tags: ${{ matrix.tags }}
+        labels: |
+          org.opencontainers.image.revision=${{ github.ref_name }}.${{ github.sha }}
+          org.opencontainers.image.version=${{ env.FRESHRSS_VERSION }}
```
b-reich, line 66: this is already handle by the `org.opencontainers.image.revision` label which is the commit hash
dont know if there is benefit to know which branch is build from, when you have the commit

also thinks the container version label is fine with sometthing like `1.22.2` or `1.22.2-alpine`its specify the container version not the version of the application itself

b-reich:
Not sure why you would avoid using the metadata action?
Personally I wouldn't worry so much about the labels. Many people aren't even interested in them.

Alkarex:
> Out of curiousity: why would you like to avoid using metadata-action? it seems helpful to me for generating the v1, v1.2, v1.2.3 tags and so on. Maybe I overlook something?

Although it does not matter much, it would be mostly to remain as simple as possible, as we do not know how things will evolve in the future. For instance, we have been through periods during which the auto-build in Docker Hub was not available while waiting for a (initially long) *Sponsored OSS* process. During that period, I built and pushed images manually. The more tags there are, the more difficult it is to move to another build system, and the more work there is for the maintenance.

Another use-case: if we were to move back to building on Docker Hubs, there is poor support for multiple tags, which each take a new build, so I would not like to have 1, 1.22, 1.22.1 but only 1.22.1. Especially not when combined with multiple architectures and Debian + Alpine. It just becomes too many tags.

Furthermore, FreshRSS is not a library. And we do not have the resources to maintain several branches anyway. So the semver semantics only marginally applies to it. For instance, fixes in 1.22.1 may very well be in 1.23.0. So I am not sure why anyone would chose to track 1.22 instead of 1.22.1 or latest. And by offering semver tags, it might mislead some users into believing that we do have updates on old branches, which we do not.

> Personally I wouldn't worry so much about the labels. Many people aren't even interested in them.

Yes, and that is why I would probably like to err on the side of simplicity instead of having rarely-used features, which might end up having some drawbacks (see my explanation higher up). As soon as we introduce something, I take on me to actually make a good effort to maintain it over time, so I need to balance what we add with its actual value.

.github/workflows/docker-publish.yml
Alkarex: > if there is benefit to know which branch is build from

This is mostly because this is what we already have, and I find it a bit easier to read with something like `1.22.1.cc2878aed855463efe36cf6fb96d001dbb75612e`. But I would not be much against leaving it to only the git SHA, if this is what is most standard.

Edit: indeed, according to https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys : **org.opencontainers.image.revision** Source control revision identifier for the packaged software.
I will edit accordingly

> also thinks the container version label is fine with sometthing like 1.22.2 or 1.22.2-alpineits specify the container version not the version of the application itself

I would also be happy with `1.22.2-alpine`, but that does not work for e.g. `edge` builds

Alkarex: Switched back to default revision https://github.com/FreshRSS/FreshRSS/pull/5808/commits/3d110b8ec9af4ba9423890c1e018e99c8fbbaec1

EdJoPaTo:
> Although it does not matter much, it would be mostly to remain as simple as possible, as we do not know how things will evolve in the future. For instance, we have been through periods during which the auto-build in Docker Hub was not available while waiting for a (initially long) _Sponsored OSS_ process. During that period, I built and pushed images manually. The more tags there are, the more difficult it is to move to another build system, and the more work there is for the maintenance.
> 
> Another use-case: if we were to move back to building on Docker Hubs, there is poor support for multiple tags, which each take a new build, so I would not like to have 1, 1.22, 1.22.1 but only 1.22.1. Especially not when combined with multiple architectures and Debian + Alpine. It just becomes too many tags.

When I moved away from auto build on Docker Hub it supported multiple tags like 1, 1.2, 1.2.3, … But that is some time ago. I explicitly searched for a way to replicate that behavior on GitHub Actions so I know and understand the pain points you are speaking of. Thanks for clarifying :)


> Furthermore, FreshRSS is not a library. And we do not have the resources to maintain several branches anyway. So the semver semantics only marginally applies to it. For instance, fixes in 1.22.1 may very well be in 1.23.0. So I am not sure why anyone would chose to track 1.22 instead of 1.22.1 or latest. And by offering semver tags, it might mislead some users into believing that we do have updates on old branches, which we do not.

Personally I use the main version tag for my containers. I get all the new features and bugfixed without worrying about breaking changes (new major release). Chosing a minor release would be to manually update but get bugfix only releases which don't change much and update the rest manually (like you suggested pinning a git hash on a used action.

Just to give an idea here: Maybe only support major and bugfix version (1 and 1.2.3). Minor releases and bugfix releases would be kinda the same. When someone wants to pin them they can. If they only care for breaking changes they can use the major version tag. This reduces the amount of tags while still allowing for non-breaking tags like I mainly use.
After all its only a single line less in the metadata-action config.

Alkarex:
> Maybe only support major and bugfix version (1 and 1.2.3)

Ok, let's go for that

Alkarex:
Let's give it a spin 🚀
Thanks for your patience and constructive dialogue so far 👍🏻 

Seems to work fine for most of it:
https://github.com/FreshRSS/FreshRSS/actions/runs/6763929607/job/18381624080

image

image

There are only a few things that need to be ironed out:
https://github.com/FreshRSS/FreshRSS/actions/runs/6763929609/job/18381624094

Alkarex:
@b-reich and @EdJoPaTo : please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

Alkarex:
Everything seems fine now, including the push to the Docker Hub description https://hub.docker.com/r/freshrss/freshrss/

Ideas for future improvement:
* Look at https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml for inspiration, for instance regarding signing (let's first find out the pros and cons)
* Run https://github.com/FreshRSS/FreshRSS/actions/workflows/tests.yml first and kill the other actions if the tests fail

PR #5812, opened 2023-11-03
Added ability to mark articles as read on focus.

### LINKED ISSUE #5723: Mark article as read after focus using hotkeys n/p
Hi team, great work on the project. There is one item preventing me from switching to FreshRSS from Feedly and wondering how hard it would be to implement.

When using the focus next (n) / previous (p) shortcuts, would it be possible to mark the article as read (without opening the article).

I usually wake up with 1000's of unread items and base what I read on the title. I've looked into creating an extension for this functionality but don't believe the hooks are available to achieve it. Would appreciate anyone's feedback on this.

passbe:
Closes #5723 

- Allow a user to enable mark as read when focusing article.

1. Enable "Mark an article as read... when focused" under Configuration > Reading > Mark article as read.
2. Using shortcut "focus next without opening" confirm article is marked as read without opening.
3. Using shortcut "focus previous without opening" confirm article is marked as read without opening.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

Notes:

- Other mark as read attributes are not documented. More than happy to write some if we believe the option isn't clear to the end user.
- The automatic scrolling is a little jarring when using this feature. Out of scope of this PR but may consider further PR's to adjust in future.


Alkarex:
Thanks for the PR 👍🏻 
I have fixed some minor things ( `make fix-all` ) and also used the batch mode for marking as read, so that when when using the shortcuts to quickly go through the list of articles, there aren't too many network requests made (save resources, and better performance). You are welcome to test

Alkarex:
Added compatibility with https://github.com/FreshRSS/FreshRSS/pull/5782 and also with the already existing *keep unread* concept (if you manually mark an article as unread, it will not automatically be marked as read when moving over it - scroll or focus - as long as you are in the same view)

passbe:
@Alkarex thank you for this. I've tested the feature and it works as expected. My only comment would be the lag in the interface between selecting the next item and it changing from read to unread in the UI. Something I may look to tweak in future. Appreciate your review.

Alkarex:
> My only comment would be the lag in the interface between selecting the next item and it changing from read to unread in the UI

Let's address that in another PR.
The current batch process timing is: wait *1s, repeat max every second.*
It should be: *wait 0s, repeat max every second.*

PR #5816, opened 2023-11-04
perf(image): replace js images lazy loading by html `loading` attribute

### LINKED ISSUE #2371: Consider using native loading="lazy"

sad270:
Closes #2371

- Replace js image lazy loading by `loading="lazy"` html attribute
- Add `loading="lazy"` on all rss content image if lazy-loading configuration is true (even if the "display_post is enabled)

1. Enable Lazy loading from setting
2. Open a feed with images
3. Check if `loading="lazy"` attribute is set
4. You can also open "Network" tab of your browser developer options and check if the feed images is only load when the feed is open, and on scroll.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/helpers/javascript_vars.phtml
+++ b/app/views/helpers/javascript_vars.phtml
@@ -15,7 +15,6 @@ echo htmlspecialchars(json_encode(array(
 		'auto_mark_scroll' => !!$mark['scroll'],
 		'auto_load_more' => !!FreshRSS_Context::$user_conf->auto_load_more,
 		'auto_actualize_feeds' => !!Minz_Session::param('actualize_feeds', false),
-		'does_lazyload' => !!FreshRSS_Context::$user_conf->lazyload ,
 		'sides_close_article' => !!FreshRSS_Context::$user_conf->sides_close_article,
 		'sticky_post' => !!FreshRSS_Context::isStickyPostEnabled(),
 		'html5_notif_timeout' => FreshRSS_Context::$user_conf->html5_notif_timeout,
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -515,13 +515,9 @@ function validateEmailAddress(string $email): bool {
  * @param string $content is the text we want to parse
  */
 function lazyimg(string $content): string {
-	return preg_replace([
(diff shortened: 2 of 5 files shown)


lib/lib_rss.php
```diff
@@ -515,13 +515,9 @@ function validateEmailAddress(string $email): bool {
...
-	return preg_replace([
-			'/<((?:img|iframe)[^>]+?)src="([^"]+)"([^>]*)>/i',
-			"/<((?:img|iframe)[^>]+?)src='([^']+)'([^>]*)>/i",
-		], [
-			'<$1src="' . Minz_Url::display('/themes/icons/grey.gif') . '" data-original="$2"$3>',
-			"<$1src='" . Minz_Url::display('/themes/icons/grey.gif') . "' data-original='$2'$3>",
-		],
+	return preg_replace(
+		'/<((?:img|iframe)[^>]+?)([^>]*)>/i',
+		'<$1 loading="lazy"$2>',
```
Alkarex, line 520: We need to check upstream that we filter out existing `loading` attributes (that might already be the case), to make sure we do not end up with two identical attributes, which would be invalid

app/views/helpers/javascript_vars.phtml
```diff
@@ -15,7 +15,6 @@ echo htmlspecialchars(json_encode(array(
 		'auto_mark_scroll' => !!$mark['scroll'],
 		'auto_load_more' => !!FreshRSS_Context::$user_conf->auto_load_more,
 		'auto_actualize_feeds' => !!Minz_Session::param('actualize_feeds', false),
-		'does_lazyload' => !!FreshRSS_Context::$user_conf->lazyload ,
```
Alkarex, line 18: This removes the option of **not** having lazy loading, which is not good

Alkarex (changes requested):
Thanks for the PR!
The use-case of **not** using lazy-loading must be preserved (descriptions of the need can be found in past issues, for instance on slow connections, to make sure all images are loaded before starting to read)

Alkarex:
Since we are at the beginning of a new FreshRSS version, it would be a good time to complete and merge this PR ;-)

PR #5827, opened 2023-11-06
Do not show message Add some feeds

Alkarex:
I found this message misleading, as a view is most of the time empty due to a filter, not due to a lack of feeds.

image


Frenzie (approved):

math-GH:
let me point to the discussion in the past: #4039

Alkarex:
Good memory, @math-GH 🤓
Are we good to merge then? We can of course improve later by having a smarter system, depending on whether there are feeds or not in the current view

math-GH (approved):

PR #5834, opened 2023-11-09
Fix/project should not use invalid return types

ColonelMoutarde:

- The project should not use invalid return types

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Models/LogDAO.php
+++ b/app/Models/LogDAO.php
@@ -12,9 +12,9 @@ final class FreshRSS_LogDAO {
 	public static function lines(?string $logFileName = null): array {
 		$logs = [];
 		$handle = @fopen(self::logPath($logFileName), 'r');
-		if ($handle) {
+		if ($handle !== false) {
 			while (($line = fgets($handle)) !== false) {
-				if (preg_match('/^\[([^\[]+)\] \[([^\[]+)\] --- (.*)$/', $line, $matches)) {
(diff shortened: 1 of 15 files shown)


lib/SimplePie/SimplePie/Cache/File.php
```diff
@@ -117,7 +117,7 @@ public function save($data)
 	/**
 	 * Retrieve the data saved to the cache
 	 *
-	 * @return array Data for SimplePie::$data
+	 * @return array|false Data for SimplePie::$data
```
Alkarex, line 120: Please do not change code in SimplePie. I have another pending task to reduce the diff with upstream

app/Models/LogDAO.php
```diff
@@ -12,9 +12,9 @@ public static function logPath(?string $logFileName = null): string {
 	public static function lines(?string $logFileName = null): array {
 		$logs = [];
 		$handle = @fopen(self::logPath($logFileName), 'r');
-		if ($handle) {
+		if ($handle !== false) {
 			while (($line = fgets($handle)) !== false) {
-				if (preg_match('/^\[([^\[]+)\] \[([^\[]+)\] --- (.*)$/', $line, $matches)) {
+				if (is_int(preg_match('/^\[([^\[]+)\] \[([^\[]+)\] --- (.*)$/', $line, $matches))) {
```
Alkarex: This looks wrong... Is that tested?

PR #5841, opened 2023-11-09
Add contribution

b-reich:
Add missing contribution (#5808) 

It's somewhat difficult to link @EdJoPaTo, as they are neither the author of the commit nor the author of the pull request.

Is there a reason to link the pr instead of the commits itself?


Alkarex:
Thanks 👍🏻 Linking to commits does not work much better, as we typically squash & merge PRs

PR #5842, opened 2023-11-09
Fix updateTagName

### LINKED ISSUE #5840: [BUG] Renaming Label with error
**Describe the bug**
`SQL error FreshRSS_TagDAO::updateTagName["HY093",null,null] ` when renaming a label

**To Reproduce**
1. Go to 'Label management'
2. Rename a label
3. The notification shows, that the renaming was successful
4. Check the label list: it was not renamed
5. check the log: see the error log entry

**Expected behavior**
If there is an error, than I expect an error notification instead a success message.
Of course I would expect that there is no error ;)

**Screenshots**
grafik

grafik

grafik

- FreshRSS version: 1.22.2-dev
- Database version: MySQL + SQlite

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5840
Regression from https://github.com/FreshRSS/FreshRSS/pull/5328

### DIFF
--- a/app/Models/TagDAO.php
+++ b/app/Models/TagDAO.php
@@ -57,15 +57,16 @@ SQL;
 	public function updateTagName(int $id, string $name) {
 		// No category of the same name
 		$sql = <<<'SQL'
-UPDATE `_tag` SET name=? WHERE id=?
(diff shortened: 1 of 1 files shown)


math-GH (approved):
tested. Works fine now.

PR #5859, opened 2023-11-11
Fix newlines not always displaying properly in description

### LINKED ISSUE #5844: [BUG] YouTube description formatting not maintained in external RSS Readers
**Describe the bug**
Descriptions for YouTube videos don't include linebreaks when viewing FreshRSS articles in a third-party RSS reader. They are just a single line of text which readers wrap to fill the content space. Viewing articles from the FreshRSS web view does not have this problem.

**To Reproduce**
I have synced my FreshRSS feed with NetNewsWire on macOS and iOS as well as Vienna RSS on macOS. The issue appears on every platform. Copy/pasting all the text from the description in a plaintext editor indicates there's just a single line of text.

**Expected behavior**
The text from the video description should match the formatting on YouTube itself, regardless of whether you're viewing from the web or a third-party RSS reader.

**Screenshots**
Here is what a YouTube video from FreshRSS currently looks like on NetNewsWire for macOS. It's very hard to read and parse what's going on:
BugReport1

**Environment information**
- FreshRSS version: 1.22.1
- Installation type: Docker
- Installed extensions: [Clickable Links](https://github.com/kapdap/freshrss-extensions)

FireFingers21:
Closes #5844 

- Modify `$description` and `<figcaption>` so that newlines are properly rendered when using both web view and API

1. Subscribe to a [youtube.com](https://www.youtube.com) feed
2. Formatting should be unchanged in web view but newlines should now appear properly in RSS readers such as NetNewsWire

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -185,7 +185,7 @@ HTML;
 				continue;
 			}
 			$credit = $enclosure['credit'] ?? '';
-			$description = nl2br($enclosure['description'] ?? '');
+			$description = nl2br($enclosure['description'] ?? '', true);
 			$length = $enclosure['length'] ?? 0;
 			$medium = $enclosure['medium'] ?? '';
 			$mime = $enclosure['type'] ?? '';
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -185,7 +185,7 @@ HTML;
 				continue;
 			}
 			$credit = $enclosure['credit'] ?? '';
-			$description = $enclosure['description'] ?? '';
+			$description = nl2br($enclosure['description'] ?? '');
 			$length = $enclosure['length'] ?? 0;
 			$medium = $enclosure['medium'] ?? '';
 			$mime = $enclosure['type'] ?? '';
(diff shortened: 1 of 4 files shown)


Frenzie (approved):

app/Models/Entry.php
```diff
@@ -222,7 +222,7 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
 				$content .= '<p class="enclosure-credits">© ' . $credit . '</p>';
 			}
 			if ($description != '') {
-				$content .= '<figcaption class="enclosure-description">' . $description . '</figcaption>';
+				$content .= '<figcaption>' . $description . '</figcaption>';
```
```suggestion
				$content .= '<figcaption class="enclosure-description">' . $description . '</figcaption>';
```

Let's keep the class, which can be used also for semantic reasons

app/Models/Entry.php
```diff
@@ -185,7 +185,7 @@ public function content(bool $withEnclosures = true, bool $allowDuplicateEnclosu
 				continue;
 			}
 			$credit = $enclosure['credit'] ?? '';
-			$description = $enclosure['description'] ?? '';
+			$description = nl2br($enclosure['description'] ?? '');
```
```suggestion
			$description = nl2br($enclosure['description'] ?? '', true);
```

Use XML-compatible mode when we can

Alkarex:
Thanks 👍🏻

PR #5884, opened 2023-11-15
Added: Display option for "My labels"

### LINKED ISSUE #5882: My labels can't be toggled in article display settings
> You could do that, but it's probably a lot easier in settings → display → disable article tags. ;-)

it's still visible after removing tickmarks from the settings on your screenshot above

_Originally posted by @igor17721 in https://github.com/FreshRSS/FreshRSS/issues/5869#issuecomment-1811649718_

math-GH:
Closes #5882

1. toggle the checkbox in Display configs
2. see the "my labels" in article footer in normal/global view

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/i18n/fr/conf.php
+++ b/app/i18n/fr/conf.php
@@ -41,6 +41,7 @@ return array(
 			'bottom_line' => 'Ligne du bas',
 			'display_authors' => 'Auteurs',
 			'entry' => 'Icônes d’article',
+			'myLabels' => 'My labels',	// TODO
 			'publication_date' => 'Date de publication',
 			'related_tags' => 'Tags de l’article',
 			'sharing' => 'Partage',
--- a/app/i18n/fr/conf.php
+++ b/app/i18n/fr/conf.php
@@ -41,7 +41,6 @@ return array(
 			'bottom_line' => 'Ligne du bas',
 			'display_authors' => 'Auteurs',
 			'entry' => 'Icônes d’article',
-			'myLabels' => 'My labels',	// TODO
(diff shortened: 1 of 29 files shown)


app/i18n/fr/conf.php
```diff
@@ -41,6 +41,7 @@
 			'bottom_line' => 'Ligne du bas',
 			'display_authors' => 'Auteurs',
 			'entry' => 'Icônes d’article',
+			'myLabels' => 'My labels',	// TODO
```
```suggestion
			'myLabels' => 'Mes étiquettes',
```

We already have the translations for that string, though

math-GH: yes, it is still in draft ;)

math-GH:
Why is this PR still in draft?
We need to discuss a bit here.
The toggle is easy. But it could be also the complete disabling of the label features.

If there are 0 labels and the new checkbox is disabled, then...
This list will keep empty "forever", because only with the "add label" feature articles could be tagged
grafik

The label management could be obsolote, if there are 0 labels
grafik

Alkarex:
Indeed. Good points @math-GH . I do not have strong feelings on that, but I would rather limit the addition of options, and it would indeed give some inconsistencies

math-GH:
Lets merge it as it is.
The labels section in the normal view: I checked it. It is already solved. 0 labels = "My labels" will not shown.
The label management needs an bigger improvement...

Frenzie (approved):
lgtm, though as stated can't we just reuse the existing "My labels"?

PR #5830, opened 2023-11-08
Use strict_types

ColonelMoutarde:
Changes proposed in this pull request:

- Little's optimisations and booleans in conditions

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #5889, opened 2023-11-17
Fix Minz_Session::param visibility

Alkarex:
Regression from https://github.com/FreshRSS/FreshRSS/pull/5830


Alwaysin:
Confirmed working!

PR #5895, opened 2023-11-20
Jalali Calendar support

### LINKED ISSUE #8379: [Feature] Persian (jalali) calendar

is there anyway to use persian (jalali) calendar in freshrss?
I have to my browser language to persian and in google the dates are shown in jalali.
is there anyway to replicate this?
does frehsrss use the web standar date format (or something like that) which if I remember correctly would automatically convert dates to local calendar.

### Use case

_No response_

instead of Gregorian calendar I want to use Jalali calendar date for dates uses in webui.

I am of course willing to test any solution.

qasedak:
Jalali is the official calendar in Iran and Afghanistan, so it would be nice to support jalali calendar for their time zones.

- Jalali calendar for Asia/Tehran and Asia/Kabul time-zones
- suitable date format for Persian and Pashto languages
- use of Vazirmant font in RTL for default theme


qasedak:
here is FreshRSS Displaying Jalali dates
Screenshot 2023-11-20 at 14-48-15 (44) جریان اصلی · FreshRSS

p/themes/Origine/origine.rtl.css
```diff
@@ -84,7 +84,7 @@
 html, body {
 	background-color: var(--background-color-light);
 	color: var(--font-color);
-	font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "PingFang SC", "Microsoft YaHei", sans-serif;
+	font-family: "Vazirmatn", "OpenSans", "Cantarell", "Helvetica", "Arial", "PingFang SC", "Microsoft YaHei", sans-serif;
```
Frenzie: The RTL files are automatically generated, so a processing directive would be better.

https://rtlcss.com/learn/getting-started/why-rtlcss/#Processing-Directives

Perhaps something like:
```suggestion
	font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "PingFang SC", "Microsoft YaHei", sans-serif /*rtl:prepend:"Vazirmatn", */;
```

p/themes/base-theme/frss.rtl.css
```diff
@@ -1,4 +1,5 @@
 @charset "UTF-8";
+@import url('https://fonts.googleapis.com/css2?family=Vazirmatn:wght@100;200;300;400;500;600;700;800;900&display=swap');
```
Frenzie, line 2: Same, but with something like:

```suggestion
/*rtl:raw:@import url('https://fonts.googleapis.com/css2?family=Vazirmatn:wght@100;200;300;400;500;600;700;800;900&display=swap');*/
```

Frenzie (commented):
Thanks, that looks cool!

p/themes/Origine/origine.rtl.css
Frenzie: Er, in the non-RTL file that is, don't apply these suggestions as is. ;-)

p/themes/base-theme/frss.rtl.css
Alkarex: I would rather avoid a query to a Google server if possible. Could this font be exported to be used locally?

Alkarex:
For internal developers: We need to find out what to do with those fonts. We might not be able to satisfy all use cases with an identical declaration (cf. e.g. https://github.com/FreshRSS/FreshRSS/pull/5671). Maybe a menu to change the font independently of the theme, or a list of fonts that depends on the language... What do you think @math-GH @Frenzie ?

p/themes/base-theme/frss.rtl.css
qasedak: Yes it could be used locally. But first i need to look into themes structure so i can implement it better...

math-GH:
It is very nice to see that FreshRSS is used around the globe. It is very important that the community gives input from their language/local needs.

What could help me to understand the current situation better? @qasedak Could you please paste a screenshot here how it looks like before the PR without this specific font family?

p/themes/base-theme/frss.rtl.css
math-GH: have a look into the CSS `@font-face`.
Font files are stored in the directory `/p/themes/fonts`

qasedak:
@math-GH , the font family is no big deal it's just a more appealing font for Arabic, Persian, ... 
here is the screenshot:
Screenshot 2023-11-21 at 09-35-54 جریان اصلی · FreshRSS

Frenzie:
> or a list of fonts that depends on the language

A UI language dependency sounds problematic to me. Even as a correlation that's quite weak.

We should probably distinguish between fonts that might be installed on the OS (is there any harm?) and web fonts (better to avoid downloading them unnecessarily).

But I would start to write it a bit differently, i.e.

```css
font-family:
  /* generic */
  Helvetica, Tex Gyre Pagella, blah blah,
  /* Japanese */
  Mac Japanese Font, Win Japanese font,
  /* Persian */
  etc.
```

> Maybe a menu to change the font independently of the theme

Yes, ultimately that makes the most sense to me, though I'll point out this can also quite easily be done with the CustomCSS.

app/Models/Entry.php
```diff
@@ -336,10 +337,25 @@ public function link(): string {
...
+				$num = 'fa';
+				$dateFormat = 'j F Y در H:i';
+			} elseif ($userLanguage == 'ps') {
+				$num = 'fa';
+				$dateFormat = 'j p Y در H:i';
+			} else {
+				$num = 'en';
+				$dateFormat = 'Y/n/j - H:i';
+			}
+			return jdate($dateFormat, $this->date, '', $userTimezone, $num);
```
ColonelMoutarde, line 406: Suggestion: use switch case...

app/Models/Entry.php
```diff
@@ -336,10 +337,25 @@ public function link(): string {
 	 * @return string|int
 	 */
 	public function date(bool $raw = false) {
+		$userTimezone = FreshRSS_Context::$user_conf->timezone;
+		$userLanguage = FreshRSS_Context::$user_conf->language;
 		if ($raw) {
 			return $this->date;
+		} else if ($userTimezone == 'Asia/Tehran' || $userTimezone == 'Asia/Kabul') {
```
ColonelMoutarde:
```suggestion
		} elseif ($userTimezone === 'Asia/Tehran' || $userTimezone === 'Asia/Kabul') {
```

Alkarex: I am not knowledgeable enough. A few points for which I have doubts:
* are all users of this timezone using this calendar?
* isn't this calendar used in other timezones, for instance when people travel?

In other words, I believe we might need a new selector option in the UI

app/views/helpers/calendars/jalali/jdf.php
```diff
@@ -0,0 +1,647 @@
...
+/**
+ * @فارسی : توابع زمان و تاریخ هجری شمسی (جلالی) در پی اچ پی
+ * @name: Hijri_Shamsi,Solar(Jalali) Date and Time Functions
+ * @Author : Reza Gholampanahi & WebSite : http://jdf.scr.ir
+ * @License: GNU/LGPL _ Open Source & Free : [all functions]
+ * @Version: 2.76 =>[ 1399/11/28 = 1442/07/04 = 2021/02/16 ]
+ */
+
+/*	F	*/
+function jdate($format, $timestamp = '', $none = '', $time_zone = 'Asia/Tehran', $tr_num = 'fa') {
```
Alkarex, line 12: Is there a version of this library (or equivalent) that exists on https://packagist.org so that we could depend on it through `composer`?

Alkarex: E.g. one of https://packagist.org/?query=Jalali ?

app/Models/Entry.php
Frenzie: > are all users of this timezone using this calendar?

Indeed, that goes both ways. It could make sense as a default but it's otherwise more or less identical to the font issue.

qasedak: In Iran and Afghanistan it's the official calendar. Also I like to add Hijri calendar system which is common in Arab and Islamic countries. If we add an option (like regional format in windows) for users to select their preferred calendar system it would be a better way I think.

app/Models/Entry.php
qasedak: like this?
```
if ($raw) {
    return $this->date;
} else {
    switch ($userTimezone) {
        case 'Asia/Tehran':
        case 'Asia/Kabul':
            switch ($userLanguage) {
                case 'fa':
                    $num = 'fa';
                    $dateFormat = 'j F Y در H:i';
                    break;
                case 'ps':
                    $num = 'fa';
                    $dateFormat = 'j p Y در H:i';
                    break;
                default:
                    $num = 'en';
                    $dateFormat = 'Y/n/j - H:i';
                    break;
            }
            return jdate($dateFormat, $this->date, '', $userTimezone, $num);
        default:
            return timestamptodate($this->date);
    }
}
```
I think it doesn't improve it

ColonelMoutarde: it's much easier to read

app/views/helpers/calendars/jalali/jdf.php
qasedak: @Alkarex I think this is the one:
https://packagist.org/packages/jalalidate/jdf
or 
https://packagist.org/packages/alifti/jdf

app/Models/Entry.php
Alkarex: Maybe picking another library then? https://packagist.org/?query=Jalali%20Hijri
https://github.com/fisharebest/ext-calendar seems to be quite popular and well maintained.
Could you check whether that could work for you?

qasedak:
I've been searching for a better replacement for JDF library that I've used here and i found https://github.com/hanifhefaz/dcter , which supports both Arabic calendar and Persian Calendar so I'll update my code soon. sorry I've been busy.

PR #5901, opened 2023-11-23
Fix types for extensions

Alkarex:
To accompany https://github.com/FreshRSS/Extensions/pull/185


Alkarex:
Let's merge not to block to much the work on the extensions.
P.S. It also calls for a new FreshRSS release soon (there are already a couple of important fixes already done and waiting for a release)

PR #5902, opened 2023-11-24
Pass FreshRSS version to JS client side for extensions

### LINKED ISSUE #5542: Pass FreshRSS version to client side
having the version on the client side could be nice, it would allow to act accordingly.

I don't think there is a way to mark an extension as only compatible with version <= x => ?

_Originally posted by @nicofrand in https://github.com/FreshRSS/FreshRSS/issues/5537#issuecomment-1640256145_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5542


Alkarex:
Ping @nicofrand

nicofrand:
Thank you @Alkarex !

nicofrand:
Hi!

Should that be available in 1.23.1? I am trying `window.FRESHRSS_VERSION` but that's undefined?

Frenzie:
context.version

The `FRESHRSS_VERSION` you're referring to is the PHP constant. ;-)

nicofrand:
Ah indeed, thanks!

PR #5930, opened 2023-12-06
Fix font priority for languages using Han characters

### LINKED ISSUE #5670: Chinese fonts are prioritized when displaying Japanese
https://github.com/FreshRSS/FreshRSS/blob/86d713478b25816340aebeeacca704f3aeee17af/p/themes/base-theme/base.css#L10C36-L10C36

Please consider changing the CSS as shown below.
This modification improves display readability for Japanese people.

Add
`"Osaka",`

Change before
`font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "PingFang SC", "Microsoft YaHei", sans-serif;`

After change
`font-family: "OpenSans", "Cantarell", "Helvetica", "Arial", "Osaka", "PingFang SC", "Microsoft YaHei", sans-serif;`

The Chinese fonts such as "PingFang SC" and "Microsoft YaHei" can be displayed like Japanese, but the kanji are greatly broken down, making them difficult for Japanese people to read.

--
Thank you for the great software!
I've migrated all my RSS readers to FreshRSS!

kasimircash:
Closes #5670

The presence of the simplified Chinese fonts 'PingFang SC' and 'Microsoft YaHei' in several theme's font-family lists means that if they are installed on a users system they will be favoured over other fonts for displaying text in other languages that use 'Han characters' such as Japanese and traditional Chinese leading to broken text in those languages.

By removing 'PingFang SC' and 'Microsoft YaHei' from the font-family list of these themes we allow the browser to use the HTML lang attribute to display text in a font that is correct for the language if one is present on the users system.

We now simply leverage browser behaviour to provide a much fuller and more robust check for default language fonts.

1. Enter Display page of Configuration.
image

2. From language menu select '日本語' then click submit button to display Japanese.
image

3. From language menu select '简体中文' then click submit button to display simplified Chinese.
image

4. From language menu select '正體中文' then click submit button to display traditional Chinese.
image

# Pull request checklist:

- [X] clear commit messages
- [x] code manually tested
- [X] documentation updated

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -107,6 +107,7 @@ People are sorted by name so please keep this order.
 * [Julien-Pierre Avérous](https://github.com/javerous): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:javerous), [Web](https://www.sourcemac.com/)
 * [Justin Tracey](https://github.com/jtracey): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:jtracey), [Web](https://unsuspicious.click)
 * [Kaibin Yang](https://github.com/SkyYkb): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:SkyYkb), [Web](https://kaibinyang.com/)
+* [Kasimir Cash](https://github.com/KasimirCash): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:KasimirCash)
 * [Kevin Papst](https://github.com/kevinpapst): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:kevinpapst), [Web](http://www.kevinpapst.de/)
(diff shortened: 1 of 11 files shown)


Frenzie:
Cross-ref to #1541.

Frenzie (approved):
Makes sense to me in theory, but it does leave me a little confused about what to do if one wanted to use a custom font for both.

kasimircash:
@Frenzie

That's a very good point and I did explore the topic of setting custom fonts for individual languages, the issue being that fonts that use Han character glyphs don't play nicely in a single list so separation is necessary. So, if a theme wanted to use custom fonts the lang() selector could be used as follows to achieve that separation:

```css
body:lang(ja) {
	font-family: "Osaka", "Meiryo", sans-serif;
}
body:lang(zh-cn) {
	font-family: "PingFang SC", "Microsoft YaHei", sans-serif;
}
body:lang(zh-tw) {
	font-family: "PingFang TC", "Microsoft JhengHei", sans-serif;
}
```
I didn't go ahead with something like that as it seemed to be somewhat outside the main issue of not rendering Japanese with a Chinese font.

kasimircash:
@Frenzie

I've now signed the commits. Could you re-review, please?

Frenzie (approved):

PR #5942, opened 2023-12-14
Add filter actions (auto mark read) at category and global levels

### LINKED ISSUE #2788: Feature Request: Global Filters
Nice and simple one – a global 'Mark as read' filter.

So I can add rules like `intitle:'deals' OR author:'deals'` to all feeds, rather than individually per feed.

### LINKED ISSUE #3497: [Feature Request] Mark articles as read with filter actions on category level
Hello,

Sorry if it's not really the right place to ask my question. Is it possible to ban/hide some words ? Like "Bons plans", "promo", etc ? I searched in administration but not found, maybe I missed the setting ?

Thank you

### LINKED ISSUE #3837: [Feature] Enable reusable/global 'Filter actions'
Hi !

Long time user of FreshRSS here !

I recently discovered the Filter Actions setting in the feeds setting : it is wonderful to be able to automatically mark as read items in feeds based on those filters.

However, I've come to notice that I'm using the same filter for many (like ~10) feeds .. all related to the main interest. For example, I'd use the same Actions filters for Tech feeds, another one for Photog feeds etc ...
And what is frustrating is that whenever I enrich the Filter actions for a given feed, I then need to update all the other one. THey do not sync :)

What would be wonderful would be to enable the creation of Action filter sets that you can use
* either at category level or even better : at feed setting level, 
* or at feed level, set to inherit this Filter Actions and oervcharge it with additional filter entries.
To do so, I believe that a Filter Actions setup page would be needed, where you define a filter action name and the filter content.
You should be able to enrich it from this page and all feeds (or categories) using it would inherit the changes.
In case the user attemps to remove a filter in use : either block it or warn that the feed using it will not use it anymore.
What do you think ?

Thanks ... and again, a huge thanks for  FreshRSS !

GLLM
-

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/3497
fix https://github.com/FreshRSS/FreshRSS/issues/2788
fix https://github.com/FreshRSS/FreshRSS/issues/3837

## At category level
image

## At global level
image


app/Controllers/categoryController.php
```diff
@@ -80,45 +80,80 @@ public function createAction() :void {
 
 	/**
 	 * This action updates the given category.
-	 * @todo Check whether this function is used at all
-	 * @see FreshRSS_subscription_Controller::categoryAction() (consider merging)
```
Alkarex, line 84: Merged with `FreshRSS_subscription_Controller::categoryAction()` (to avoid partially duplicated and dead code)

PR #5955, opened 2023-12-20
Fix login

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5953
Regression due to https://github.com/FreshRSS/FreshRSS/pull/5946

PR #5959, opened 2023-12-21
Fix notifications

Alkarex:
The notification about wrong login was not working. Noticed while working on https://github.com/FreshRSS/FreshRSS/pull/5955
This was due to timing of when the notification is retrieved.
Simplified code to make the logic easier and more robust.

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -77,10 +77,10 @@ if (_t('gen.dir') === 'rtl') {
 <?php
 	$msg = '';
 	$status = 'closed';
-	if (!empty($this->notification)) {
-		$msg = $this->notification['content'];
-		$status = $this->notification['type'];
-
+	$notif = Minz_Request::getNotification();
+	if (!empty($notif)) {
+		$msg = $notif['content'];
+		$status = $notif['type'];
 		invalidateHttpCache();
 	}
 ?>
--- a/app/layout/simple.phtml
+++ b/app/layout/simple.phtml
@@ -59,9 +59,10 @@
 <?php
 	$msg = '';
 	$status = 'closed';
-	if (!empty($this->notification)) {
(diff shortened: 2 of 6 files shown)


Frenzie (approved):
Looks good, but I haven't tested it.

app/layout/layout.phtml
```diff
@@ -77,10 +77,10 @@ if (_t('gen.dir') === 'rtl') {
 <?php
 	$msg = '';
 	$status = 'closed';
-	if (!empty($this->notification)) {
-		$msg = $this->notification['content'];
-		$status = $this->notification['type'];
-
+	$notif = Minz_Request::getNotification();
+	if (!empty($notif)) {
```
ColonelMoutarde, line 81:
```suggestion
	if ($notif !== null && array_key_exists('content', $notif) && array_key_exists('type', $notif)) {
```

app/layout/simple.phtml
```diff
@@ -59,9 +59,10 @@
 <?php
 	$msg = '';
 	$status = 'closed';
-	if (!empty($this->notification)) {
-		$msg = $this->notification['content'];
-		$status = $this->notification['type'];
+	$notif = Minz_Request::getNotification();
+	if (!empty($notif)) {
```
ColonelMoutarde, line 63:
```suggestion
	if ($notif !== null && array_key_exists('content', $notif) && array_key_exists('type', $notif)) {
```

app/layout/layout.phtml
Alkarex: The check is already done by type declaration (this code already passes PHPStan Level 9). And if anything, `array_key_exists` does not seem like the correct test as it accepts `null`, which we do not want.
We could consider using `??` below for providing default values, but I do not think it is necessary for now.

PR #5969, opened 2023-12-23
Authentik oAuth setup documentation updates

laxmanpradhan:
Closes #

- Updated the Redirect URL documents to include port number. Required for authentik at least, however I suspect this is a FreshRSS issue as I have seafile oAuth setup the same way and the redirect URL does not require the port. Without the port number Authentik gives a redirect_url error. 
- Many small clarifications and improvements to make setup easier 

1. Create a new FreshRSS docker install follow the new instructions for oAuth, will be much easier to follow. 

Pull request checklist:

- [ x] clear commit messages
- [ x] documentation updated


Alkarex:
Let's try to fix the problem you are describing:

> Not sure how to setup OIDC into an existing installation because you get stuck in a loop. The last step after everything works, is to login to FreshRSS and under FreshRSS > Settings > Authentication, you need to change the user to HTTP. But this setting is only available if you have OIDC enabled. But if you enable OIDC and try to login you have to login with a user that already has HTTP enabled. I got it to work because I was setting up a fresh install and so it brought me to the page to create the user, I made one with the same name as my oAuth admin user and then I was able to immediately set it to HTTP in the settings before logging out.

Changing to HTTP should be possible after https://github.com/FreshRSS/FreshRSS/pull/5969/commits/0140469e711e5b74d9a388b9237f15f9c81a1f55

Alkarex:
Could you please check @laxmanpradhan ?

PR #5971, opened 2023-12-23
Added additional media:content for thumbnails

### LINKED ISSUE #5925: Improve images for a range of clients
Some RSS feeds output from FreshRSS do not contain images that are seen by some clients. This is a proposed change.

Example:
Add a feed with separate images, e.g. one scraped from a website by XPATH
Add an RSS source, e.g from Wordpress for comparison, with images inside the article descriptions,
View FreshRSS's default anonymous output feed in MailChimp or view in Feedbro browser (Firefox)

**Actual behaviour**
Missing images that are tagged only as "media:thumbnail"

**Expected behaviour**
All available images should show.  Scraped images are listed in the feed but in a form that these readers don't recognise ("media:thumbnail"). Images from a Wordpress RSS source in the same output feed will show, probably because they appear inside the "description" tag.

- Browser: Firefox 119. Mailchimp - as of today
- FreshRSS version: 1.22.1
- PHP version:: PHP 7.4.3
- Cpanel self hosted

**Additional context**
There appears a difference in interpretation of the spec: https://www.rssboard.org/media-rss :
<media:content url= ...> tag is a "primary element", of the form that FreshRSS already outputs for an 'enclosure'.
<media:thumbnail url=...> tag is an "optional element", a sub-element of "media:content
" or "item"

While the FreshRSS implementation is in spec, it's not the same interpretation as these 2 readers, and probably some others

**Suggestion**
In rss.phtml add the enclosure code for thumbnails

`
			if (!empty($thumbnail['url'])) {
  			// https://www.rssboard.org/media-rss#media-thumbnails				
					echo "\t\t\t", '<media:thumbnail url="' . $thumbnail['url']
						. (empty($thumbnail['width']) ? '' : '" width="' . $thumbnail['width'])
						. (empty($thumbnail['height']) ? '' : '" height="' . $thumbnail['height'])
						. (empty($thumbnail['time']) ? '' : '" time="' . $thumbnail['time'])
						. '" />', "\n";
   				// added for readers such as MailChimp + Feedbro
				  echo "\t\t\t", '<media:content medium="image" url="' . $thumbnail['url']
					. (empty($thumbnail['width']) ? '' : '" width="' . $thumbnail['width'])
					. (empty($thumbnail['height']) ? '' : '" height="' . $thumbnail['height'])
					. '" />', "\n";
			}
`

With this code both above services can see the images from the scraped content.

Other readers can read this OK (Feedly, Newsblur etc), so this seems like perhaps a more compatible solution - or drop the "media:thumbnail" code, but FreshRSS seems to use this itself.  Having 2 tags of different formats for the same image doesn't create multiple images - at least in my month of testing.   The W3C and RSS advisory board validators also don't object.

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5925
Ref https://github.com/keynet/FreshRSS/pull/1
Ping @keynet

### DIFF
--- a/app/views/index/rss.phtml
+++ b/app/views/index/rss.phtml
@@ -49,6 +49,14 @@ foreach ($this->entries as $item) {
 						. (empty($thumbnail['height']) ? '' : '" height="' . $thumbnail['height'])
 						. (empty($thumbnail['time']) ? '' : '" time="' . $thumbnail['time'])
 						. '" />', "\n";
+					/*  Added mostly for MailChimp + Feedbro which don't support <media:thumbnail> 
+						different to media actually inside enclosures below. 
+						See https://mailchimp.com/help/rss-merge-tags/
+					*/
+					  echo "\t\t\t", '<media:content medium="image" url="' . $thumbnail['url']
+						. (empty($thumbnail['width']) ? '' : '" width="' . $thumbnail['width'])
+						. (empty($thumbnail['height']) ? '' : '" height="' . $thumbnail['height'])
+						. '" />', "\n";
 				}
 				$enclosures = $item->enclosures(false);
 				if (is_iterable($enclosures)) {


Alkarex:
Replaced by https://github.com/FreshRSS/FreshRSS/pull/5972

Issue #5982, opened 2023-12-26
[BUG]Export/Import always times out after update


vdbhb59:
**Describe the bug**
The export/Import functionality is always timing out upon update to V1.23.0

**To Reproduce**
1. Go to 'Export/Import feature'
2. Click on 'Export'
3. Timeout happens

**Expected behavior**
Export should start for the OPML, instead of timing out.

**Screenshots**
![image](https://github.com/FreshRSS/FreshRSS/assets/60728004/9552fe9c-75f3-4688-a30b-69321fce24be)

**Environment information (please complete the following information):**
- Device: Laptop / Smartphone
- OS: Windows 11 / Zorin 17 / Android 13
- Browser: [Librewolf 121.0-1 / Mull 121.X]
- FreshRSS version: [1.23.0]
- Database version: [5.7.44 - MySQL Community Server (GPL)]
- PHP version: [PHP 8.1]
- Installation type: [e.g. cPanel Shared Hosting]

**Additional context**
Another BUG is also being raised of feed posts not loading beyond 24 hours time separately.
Extensions currently active are:
'TinyTinyRSS API' => true,
'Invidious Video Feed' => true,
'Black List' => false,
'RSS-Bridge' => true,
'Comics in feed' => false

This same error comes in log here too: SQL error searchById|category|toGReader|include ["HY000",2014,"Cannot execute queries while other unbuffered queries are active. Consider using PDOStatement::fetchAll(). Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute."]

Alkarex:
Would you be able to add a debug line in the code?

Insert the following line between lines 106 and 107 (inside the `if`):

```php
Minz_Log::warning(json_encode(debug_backtrace(), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES));
```

of `./FreshRSS/app/Models/Feed.php`:

https://github.com/FreshRSS/FreshRSS/blob/e9689645383d37231afa9dbbcd9bd765b804046d/app/Models/Feed.php#L104-L110

Likely the same root problem than https://github.com/FreshRSS/FreshRSS/issues/5981

Alkarex:
After the test above (I am interested in the logs, if possible), would you be able to try the `edge` version? I have made some changes, which might hopefully help https://github.com/FreshRSS/FreshRSS/pull/5986

[...]

Issue #5981, opened 2023-12-26
[BUG] Feeds not loading beyond certain time (24 hours)


vdbhb59:
**Describe the bug**
The feeds / posts are not loading beyond the last 24 hours since update to 1.23.0

**To Reproduce**
1. Go to 'Home page/feed page'
2. Scroll down to 'show feeds / auto load'
3. See feeds not loading beyond 24 hours.

**Expected behavior**
Feeds always auto load upon scrolling down, all the time, till the end of scrolling.

**Screenshots**
![image](https://github.com/FreshRSS/FreshRSS/assets/60728004/995e93a8-89b2-4353-87c7-5b00eed38a10)

**Environment information (please complete the following information):**
- Device: Laptop / Smartphone
- OS: Windows 11 / Zorin 17 / Android 13
- Browser: [Librewolf 121.0-1 / Mull 121.X]
- FreshRSS version: [1.23.0]
- Database version: [5.7.44 - MySQL Community Server (GPL)]
- PHP version: [PHP 8.1]
- Installation type: [e.g. cPanel Shared Hosting]

**Additional context**
Extensions currently active are:
'TinyTinyRSS API' => true,
'Invidious Video Feed' => true,
'Black List' => false,
'RSS-Bridge' => true,
'Comics in feed' => false,

Logs give me this:
SQL error searchById|category|toGReader|include ["HY000",2014,"Cannot execute queries while other unbuffered queries are active. Consider using PDOStatement::fetchAll(). Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute."] 

More details: This is happening only in the following themes > Origine & Origine-Compact. No other theme is giving this error.

Alkarex:
Hum, the `toGReader()` is not supposed to be called from this view. Could you please try to disable your FreshRSS extensions and compare?

vdbhb59:
Without extensions it is working intermittently. I removed all extensions, even the only ones which are shown under compatible. Then it is not giving this issue every time, but otherwise it is back to same 24 hours stoppage. Also, it is happening mostly for those 2 themes Origine & Origine-Compact only with and without extensions.

Alkarex:
Would you be able to add a debug line in the code?

Insert the following line between lines 106 and 107 (inside the `if`):

```php
[...]

PR #5986, opened 2023-12-27
More robust assignment of categories to feeds

Alkarex:
Several minor cases, none of which should really be necessary
Might help:
https://github.com/FreshRSS/FreshRSS/issues/5981
https://github.com/FreshRSS/FreshRSS/issues/5982


Alkarex:
Merging a bit fast to get broader feedback

PR #5657, opened 2023-09-14
Sensitive base-url

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/5656#issuecomment-1719802075
Make read-only while waiting for a better approach (which can wait till release 1.23)

image

PR #5992, opened 2023-12-28
fix: `base_url` becomes empty after saving configuration

yzqzss:

Since #5657 sets the base URL input box to `disabled="disabled"`, the browser will not send the `base-url` parameter. As a result, when the submit button is clicked, the base URL will be cleared.

before:

https://github.com/FreshRSS/FreshRSS/assets/30341059/a9760aca-8742-4668-9f43-9ac0fc72dbe8

fixed:

https://github.com/FreshRSS/FreshRSS/assets/30341059/70e1f3f2-d1de-402f-b40f-f40634311f62

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 	* Fix crash regression when enabling extensions [#5979](https://github.com/FreshRSS/FreshRSS/pull/5979)
 	* Fix crash regression during export when using MySQL [#5988](https://github.com/FreshRSS/FreshRSS/pull/5988)
 	* More robust assignment of categories to feeds [#5986](https://github.com/FreshRSS/FreshRSS/pull/5986)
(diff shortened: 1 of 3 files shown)


Frenzie:
Fwiw there's also the `readonly` attribute for `<input>` elements.

Frenzie (approved):
Seems sensible enough.

Alkarex:
Regression from https://github.com/FreshRSS/FreshRSS/pull/5657

> Fwiw there's also the `readonly` attribute for `<input>` elements.

Indeed @Frenzie , I was not aware of the difference in POST behaviour

https://developer.mozilla.org/docs/Web/HTML/Attributes/readonly#attribute_interactions

Alkarex:
Well spotted @yzqzss , thanks!

PR #5994, opened 2023-12-29
Docker Debian disable unused PHP modules

### LINKED ISSUE #5993: [Feature] Disable unused PHP modules in Docker image
PHP in the Docker image loads several modules that are not needed and (1) consume memory and (2) enlarge the attack surface security-wise.

```
RUN phpdismod calendar exif ftp gettext gmp iconv intl phar posix readline shmop sockets \
              sysvmsg sysvshm tokenizer xmlreader xmlwriter xsl

```

None

**Additional context**
None

Alkarex:
Debian has several PHP modules enabled by default, which we do not use.
Aligns Debian image with the PHP modules of the Alpine image

fix https://github.com/FreshRSS/FreshRSS/issues/5993

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -36,6 +36,7 @@ LABEL \
 
 RUN a2dismod -q -f alias autoindex negotiation status && \
 	a2dismod -q auth_openidc && \
+	phpdismod calendar exif ffi ftp gettext mysqli posix readline shmop sockets sysvmsg sysvsem sysvshm xsl && \
 	a2enmod -q deflate expires headers mime remoteip setenvif && \
 	a2disconf -q '*' && \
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
exif too? sounds potentially useful

Alkarex:
> exif too? sounds potentially useful

Not something we are using in the core at the moment, but are you thinking about extensions, maybe?

Frenzie:
Yup! I don't know if any extensions make us of it though. :-)

Alkarex:
This PR just aligns the Debian image with what we already have in Alpine. So if we wanted e.g. Exif, we would need to enable in Alpine too, so that calls for another PR if there is a demand :-)

Issue #6008, opened 2023-12-31
[QUERY]Update page shows last version unknown


vdbhb59:
In the image below, as you can see there is the line stating: "**An update of FreshRSS is available: Version unknown.**" & below it states "**Last check = unknown**".
I went through the code, and there is a file called "last_update.txt", which seems to be missing for my deployment of the instance.
Can you guide me, what path should this file be in and what should the file contain to get the above 2 points showing correct data, please?
https://github.com/FreshRSS/FreshRSS/blob/77108ea19e725022e6168753d19037d781b4f8fe/app/Controllers/updateController.php#L6

Commented here: https://github.com/FreshRSS/FreshRSS/commit/9c104ee6bb81b6bf82b45355ef59709ce783488a#commitcomment-135967818

![293474965-14de2c99-3e31-405b-8843-b0ec1a4cb0aa](https://github.com/FreshRSS/FreshRSS/assets/60728004/f98c5ae2-db79-4d9b-97d8-87f213e5b28b)

Alkarex:
Please provide information about your setup (OS, versions, type of deployment...)

vdbhb59:
CentOS 7
Manual ZIP installation
Version 1.23.1-Stable
PHP 8.0.30
All permissions in place

Info: The file "last_update.txt" may have been deleted in error, while I was trying to delete a folder, but I am not fully sure. I believe if I just recreate the text file and put in the version number, this issue will get fixed automatically.

Alkarex:
And what happens when you click "Start update"?

vdbhb59:
> Info: The file "last_update.txt" may have been deleted in error, while I was trying to delete a folder, but I am not fully sure. I believe if I just recreate the text file and put in the version number, this issue will get fixed automatically.

@Alkarex updated above about file I may have deleted in error.

> And what happens when you click "Start update"?

It just goes round and round (updating) and then does nothing, but instead times out.
`This request takes too long to process, it is timed out by the server. If it should not be timed out, please contact administrator of this web site to increase 'Connection Timeout'.`

No logs in UI, nor in /data/users/**/log file.

![image](https://github.com/FreshRSS/FreshRSS/assets/60728004/9c312750-cccd-4374-b330-65e53fcb9e11)

Alkarex:
Since you seem to have command-line access to your server, I can propose that you upgrade to using git to control FreshRSS versions (it will still be possible to update through the Web interface with that method):

[...]

PR #6013, opened 2024-01-02
docs: Update 02_Prerequisites.md

math-GH:
see discussion here: https://github.com/FreshRSS/FreshRSS/issues/6008#issuecomment-1874003898


docs/en/admins/02_Prerequisites.md
```diff
@@ -7,8 +7,9 @@ You need to verify that your server can run FreshRSS before installing it. If yo
 | Software      | Recommended             | Also Works With         |
 | ------------- | ----------------------- | ----------------------- |
 | Web server    | **Apache 2.4+**         | nginx, lighttpd         |
-| PHP           | **PHP 7.4+**            |                         |
+| PHP           | **PHP 7.4+**            | Soft require Apache 2.4+ (but repair minimal compatibility with Apache 2.2) |
```
Frenzie: Most of that is part of the commit message which shouldn't be copied here verbatim.

I'm not sure if the 2.2 thing should be mentioned, but if so perhaps something like:
```suggestion
| PHP           | **PHP 7.4+**            | Apache 2.4+ (minimal compatibility with Apache 2.2) |
```

Frenzie: Or perhaps simply: `2.2+ minimum, 2.4+ recommended`

math-GH: It is just copy&pasted from the changelog of 1.23.0 https://github.com/FreshRSS/FreshRSS/releases/tag/1.23.0

Frenzie: Exactly ;-)

docs/en/admins/02_Prerequisites.md
```diff
@@ -7,8 +7,9 @@ You need to verify that your server can run FreshRSS before installing it. If yo
...
 | PHP modules   | Required: libxml, cURL, JSON, PDO_MySQL, PCRE and ctype.<br />Required (32-bit only): GMP <br />Recommended: Zlib, mbstring, iconv, ZipArchive<br />*For the whole modules list see [Dockerfile](https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/Dockerfile-Alpine#L7-L9)* | |
+| PHP settings  | **memory_limit: 128MB+**, **max_execution_time: 30s+** |                        |
```
Alkarex, line 12: I am not sure of those values. Should be tested a bit. And we modify the `max_execution_time` a few places in the code

docs/en/admins/02_Prerequisites.md
math-GH: by the way: I see that the notice is placed in the wrong row.... -.-

docs/en/admins/02_Prerequisites.md
```diff
@@ -6,9 +6,10 @@ You need to verify that your server can run FreshRSS before installing it. If yo
 
 | Software      | Recommended             | Also Works With         |
 | ------------- | ----------------------- | ----------------------- |
-| Web server    | **Apache 2.4+**         | nginx, lighttpd         |
+| Web server    | **Apache 2.4+**         | nginx, lighttpd<br />minimal compatibility with Apache 2.2    |
```
Alkarex: Good wording 👍🏻

docs/en/admins/02_Prerequisites.md
Alkarex: I suggest to keep this line for another PR, so that we can merge the part with Apache already

Alkarex: I still think we should first merge the Apache info and come back to the PHP requirements in another PR after some actual tests of the needed memory

math-GH: Or we could use the numbers, that works fine for sure. My current settings are:
max_execution_time: 60 s
memory_limit: 256 MB

Alkarex: Sure, but it also works with `memory_limit: 1024 MB` for instance, so that would be misleading. Ideally, it should be the minimal memory required, but I have not tested what it is, and it will also vary depending on the feeds and articles (256 MB might not be enough for some extreme use cases), although there has been efforts to keep memory requirements low. And 60s for sure will not work for some big use cases (I have some myself requiring higher values).

Alkarex: I will make a few tests on my personal instance (which does not host large use cases, but has many different types of feeds so it should have some good degree of representativeness). 16MB seems to work fine so far.
Help welcome

Frenzie: I've been running on 128 (the Debian default) with no issues fwiw, but I also think lower should be just fine.

Alkarex: For the record, why this should be done carefully / why it matters: this type of information is used for instance to calculate the number of possible concurrent users on a server, thereby also the cost for operating a service, and I think it might soon also be used for CSRD-like green assessment

Alkarex: At 16MB, it has worked mostly fine for my test, besides one specific function of the API, which could by the way be good to optimize:

https://github.com/FreshRSS/FreshRSS/blob/4b29e666b06762b4b36438c9370c38bc43121f78/p/api/greader.php#L719

Alkarex: https://github.com/FreshRSS/FreshRSS/issues/6136

math-GH: thanks for this discussions.

Let me explain why it is important for me to have this numbers in the documentation:
I use a cheap webspace. The memory_limit is one of the most important prerequisits to understand whether the application can run on my machine or not.
So I want to share here: which minimum memory_limit is needed for a first run and which memory_limit is suggested. Yes, more is better ;)

Alkarex: Exactly, @math-GH , so putting wrong numbers is bad. And finding the correct numbers requires testing.
Contribution https://github.com/FreshRSS/FreshRSS/pull/6137

Alkarex: I am testing something in the lines of: 16MB minimum (after https://github.com/FreshRSS/FreshRSS/pull/6137), more than 64MB recommended. Help welcome.

Alkarex: The export/import functions of labelled/starred articles currently do not scale at all, as they put everything in memory (and multiple copies) https://github.com/FreshRSS/FreshRSS/issues/6139

math-GH:
let's close it. We can add the information later to the documentation.

PR #6028, opened 2024-01-05
2297 cli parameters consistency

### LINKED ISSUE #2297: CLI parameters consistency
It is very strange to get syntax inconsistency in parameter name: `--api_password` seems not natural as we usually use dash *or* underscore, but not both of them, we should convert all parameter to dash syntax: `--api-password`.

see #2046 to see an anwser from @Alkarex

kasimircash:
Closes #2297

We want to standardise flags used with CLI commands around the use of '-' rather than '\_' while maintaining backward compatibility. To that end I've deprecated all flags using '\_' in favour of an equivalent using '-' and introduced a check for the use of deprecated flags as well as logic to replace any used deprecated flags with their replacements. I've also scooped up some existing logic for CLI flag retrieval and validation and wrapped all this up into a parseCliParams() function. This way we can start to isolate some logic that otherwise is scattered throughout individual CLI commands.

If we're happy with the changes proposed here, then I could look at raising some related tickets such as extending use of parseCliParams() to other relevant CLI commands and looking at other opportunities for standardisation around validating CLI inputs as there is currently quite a bit of variation in how individual CLI commands handle similar work e.g. validating options associated with flags or updating config.

1. Use either cli/do-install, cli/reconfigure, cli/create-user or cli/update-user with updated flags (e.g. --default-user admin --allow-anonymous) to see successful behaviour.
image
image

2. Use deprecated flags (e.g. --default_user admin --allow_anonymous) to see successful behaviour with a deprecation warning.
image
image

3. Use unknown flags (e.g. --defaultuser admin --allow_anon)to trigger expected warning and failure.
image
image

# Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


cli/_cli.php
```diff
@@ -75,6 +75,48 @@ function performRequirementCheck(string $databaseType): void {
 	}
 }
 
+/**
+ * Parses perameters used with FreshRSS' CLI commands.
+ * @param array{'valid':array<string,string>,'deprecated':array<string,string>} $parameters An array of 'valid': An
+ * array of perameters as keys and their respective getopt() notations as values. 'deprecated' An array with
```
aledeg:
```suggestion
 * array of parameters as keys and their respective getopt() notations as values. 'deprecated' An array with
```

There are multiple instances of `parameters`. Please replace them by `parameters`.

aledeg (changes requested):
I haven't tested it but it seems alright. Good job.

cli/_cli.php
kasimircash: That's fixed now.

kasimircash: Can't believe I forgot how to spell parameters...

Alkarex:
Hello and thanks for the PR 👍🏻 
If possible, please allow edits; example from another PR:

image

cli/_cli.php
```diff
@@ -104,3 +145,68 @@ function validateOptions(array $input, array $params): bool {
...
+ */
+function checkforDeprecatedParameterUse(array $options, array $params): bool {
+	$deprecatedOptions = array_intersect($options, $params);
+	$replacements = array_map(static fn($option) => array_search($option, $params, true), $deprecatedOptions);
+
+	if (0 === count($deprecatedOptions)) {
+		return false;
+	}
+
+	trigger_error("The FreshRss CLI option(s): " . implode (', ', $deprecatedOptions) .
```
Alkarex:
```suggestion
	trigger_error("The FreshRSS CLI option(s): " . implode (', ', $deprecatedOptions) .
```

Alkarex: Although the native function `trigger_error(..., E_USER_DEPRECATED)` is nice, it generates some outputs, which I found a bit misleading. Users could tend to believe that is is related to some internal code, due to the `PHP Deprecated: ` prefix, and ` in /var/www/FreshRSS/cli/_cli.php on line 183` suffix (which I would like to strip).

```
PHP Deprecated:  The FreshRss CLI option(s): base_url, default_user are deprecated and will be removed in a future release. Use: base-url, default-user instead in /var/www/FreshRSS/cli/_cli.php on line 183
```

I think we should use a normal STDERR output like we do for some other messages.

kasimircash: That all makes sense. I'll switch it over.

kasimircash:
@Alkarex, @aledeg Thanks for the feedback.

PR #6036, opened 2024-01-12
Standardise command line option parsing

kasimircash:

Continuing on from the work done in #6028, parsing of options entered on the command line is now standardised across all relevant commands (those that accept options) giving us a single source of truth and potentially clearing the way to further standardisation.

Additionally: we now validate short as well as long options, all existing short options now have a long equivalent, and it would be trivial to add short equivalents for any existing long options. Relevant documentation (e.g. the cli README as well as any --help or usage outputs) has been updated to take account of these changes.

I've also scooped up a fix to check.translation.php which was failing to run due to not having access to the I18N_PATH constant and now runs as expected.

1. To see short option validation use an unknown short option with any command e.g. 
```shell
cli/do-install.php --default-user admin -q
```
Previous behaviour: 
image

New behaviour:
image

2. All other behaviour remains unchanged and so can be tested by using commands and verifying that the result is as would have previously been expected e.g.
```shell
cli/user-info.php --user foo --json
```
will still provide information for user 'foo' in json format etc.

# Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/cli/export-sqlite-for-user.php
+++ b/cli/export-sqlite-for-user.php
@@ -5,19 +5,26 @@ require(__DIR__ . '/_cli.php');
 
 performRequirementCheck(FreshRSS_Context::systemConf()->db['type'] ?? '');
 
-$params = [
-	'user:',
-	'filename:',
-];
+$parameters = array(
+	'long' => array(
+		'user' => ':',
+		'filename' => ':',
+	),
+	'short' => array(),
+	'deprecated' => array(),
+);
 
-$options = getopt('', $params);
+$options = parseCliParams($parameters);
 
-if (!validateOptions($argv, $params) || empty($options['user']) || empty($options['filename']) || !is_string($options['user']) || !is_string($options['filename'])) {
+if (!empty($options['invalid'])
+	|| empty($options['valid']['user']) || empty($options['valid']['filename'])
+	|| !is_string($options['valid']['user']) || !is_string($options['valid']['filename'])
+) {
 	fail('Usage: ' . basename(__FILE__) . ' --user username --filename /path/to/db.sqlite');
 }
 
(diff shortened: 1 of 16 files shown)


Alkarex:
Looks good at a quick glance, thanks 👍🏻 
The PR is still marked as *draft*; is it on purpose?

kasimircash:
@Alkarex Ah. No, draft status was an accident on my part. Sorry about that.

cli/export-sqlite-for-user.php
```diff
@@ -5,19 +5,26 @@
...
-
-if (!validateOptions($argv, $params) || empty($options['user']) || empty($options['filename']) || !is_string($options['user']) || !is_string($options['filename'])) {
+$parameters = array(
+	'long' => array(
+		'user' => ':',
+		'filename' => ':',
+	),
+	'short' => array(),
+	'deprecated' => array(),
+);
```
Alkarex: Minor syntax detail: Could you please use the `[  ]` array syntax instead of `array( )` for the new code?
(Multiple instances in this file and other modified files)

kasimircash: Sure. Just to clarify, shall I replace all instances of `array( )` with `[ ]` in the modified files or just the ones I've introduced?

Alkarex: Only your own changes and lines immediately above or below, for now.
We will probably run an automated replacement for the rest in another PR.

kasimircash: I think that takes care of it.

Alkarex (approved):

Alkarex:
Thanks 👍🏻

PR #5962, opened 2023-12-22
improve phpdoc

ColonelMoutarde:
Changes proposed in this pull request:

- improve phpdoc

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

PR #5504, opened 2023-07-03
Chore/processing of depreciations and updating code to php72 minimum

ColonelMoutarde:
Changes proposed in this pull request:
- processing of depreciation and updating code to php72 minimum

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #6037, opened 2024-01-13
PHPStan prepare exceptions

Alkarex:
Take advantage of
https://phpstan.org/blog/bring-your-exceptions-under-control

Minimum changes to pass `tooWideThrowType` and `implicitThrows`.

Revert some mistakes from:
https://github.com/FreshRSS/FreshRSS/pull/5504
Preparation needed before new PRs of the same type:
https://github.com/FreshRSS/FreshRSS/pull/5962

Fix several wrong PHPDocs and catches:

> Method ... has ...Exception in PHPDoc @throws tag but it's not thrown.

> Dead catch - ...Exception is never thrown in the try block.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -50,6 +50,13 @@ parameters:
 		strictCalls: true
 		switchConditionsMatchingType: true
 		uselessCast: true
+	exceptions:
+		check:
+			missingCheckedExceptionInThrows: false	# TODO pass
+			tooWideThrowType: true
+		implicitThrows: false
+		checkedExceptionClasses:
+			- 'Minz_Exception'
 	ignoreErrors:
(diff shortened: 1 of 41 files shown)


phpstan.neon
```diff
@@ -50,6 +50,13 @@ parameters:
 		strictCalls: true
 		switchConditionsMatchingType: true
 		uselessCast: true
+	exceptions:
+		check:
+			missingCheckedExceptionInThrows: false	# TODO pass
+			tooWideThrowType: true
+		implicitThrows: false
+		checkedExceptionClasses:
+			- 'Minz_Exception'
```
Alkarex, line 60: This is the goal of this PR

PR #6041, opened 2024-01-15
Fix manipulate.translation ignore behaviour

kasimircash:
When using `cli/manipulate.translation.php` we would expect that `cli/manipulate.translation.php -a ignore -l en -k install.auth.form` would add an ignore note to the key `install.auth.form` and that `cli/manipulate.translation.php -a ignore -l en -k install.auth.form -r` would remove the ignore note from `install.auth.form`. Instead, this behaviour is reversed. This same reversed behaviour is present when using `-a ignore_unmodified` with `cli/manipulate.translation` as well.

This fix gives us the behaviour from `cli/manipulate.translation.php` that we would expect e.g. the reverse of current behaviour.

I've also modified the tests for `ignore` and `ignore_unmodified` in I18nDataTest.php as they were checking for the incorrect output.

1. Use `cli/manipulate.translation.php -a ignore -l en -k install.auth.form` to ignore a key.

2. Check the translation file (in this case `FreshRSS/app/i18n/en/install.php`) to confirm the key is ignored.
image

3. Use `cli/manipulate.translation.php -a ignore -l en -k install.auth.form -r` to remove the ignore note.

4. Confirm the key has been unignored.
image

# Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] unit tests written

### DIFF
--- a/cli/i18n/I18nData.php
+++ b/cli/i18n/I18nData.php
@@ -288,27 +288,27 @@ class I18nData {
 	}
 
 	/**
-	 * Ignore a key from a language, or reverse it.
+	 * Ignore a key from a language, or revert an existing ignore on a key.
 	 */
-	public function ignore(string $key, string $language, bool $reverse = false): void {
+	public function ignore(string $key, string $language, bool $revert = false): void {
 		$value = $this->data[$language][$this->getFilenamePrefix($key)][$key];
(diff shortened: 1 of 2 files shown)


Alkarex:
Pinging @aledeg if you are around

PR #5269, opened 2023-04-06
Phpstan Level6 for View.php

ColonelMoutarde:
Changes proposed in this pull request:

- Phpstan Level6 for View.php

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #6046, opened 2024-01-18
Native array return type forgotten

Alkarex:
Native type forgotten from
https://github.com/FreshRSS/FreshRSS/pull/5269
https://github.com/FreshRSS/FreshRSS/pull/5213


Alkarex:
Merging a bit fast, as I am working on a dependency, which conflicts due to change in function signature

PR #6052, opened 2024-01-21
New feature: shareable user query 

### LINKED ISSUE #3066: Share feeds between users?
Hi,

First of all, thanks for this very nice software! I have a slightly particular use case though, which involves a kind of cooperation between users.

Ideally, I'd like any user to be able to pick feeds from any other user's account and add it to its account (my users all share the same interestes), to help improve discoverability of RSS feeds.

Is there anything which could help me achieve this? Or is there any "dirty hack" solution which could achieve this (or alternatively pre-filling a list of RSS feeds for new user accounts)?

Otherwise, what would it require (in terms of development) to add such a feature in FreshRSS?

Thanks!
Best,

### LINKED ISSUE #3178: Making a feed public?
Hi there,

FreshRSS has this very useful ability to create a single RSS feed out of multiple feeds. But is there a way I can make this new feed public, so that it can be viewed without having to login? For instance via a feed reader that doesn't have the ability to login? I also see that there's this "allow unsafe automatic login.." format that can be used to access the reader, but the feeds don't seem to be accessible in this fashion. So is there a way around this?

Thanks!

### LINKED ISSUE #6086: [BUG] User Queries settings page hangs if recursive query added
**Describe the bug**
The user queries settings page will hang if the user adds a recursive user query.

**To Reproduce**
1. Add a user queries. The offending query only used `intitle`
2. Add a second user query that uses the previously created user query. i.e: `search:"Query A"`
3. The user queries page should now hang and prevent the user from editing or removing any user queries.

- FreshRSS version: 1.22.1 (also tested on 1.23.1)
- Database version: SQLite 3.34.1
- PHP version: 8.3
- Installation type: Manual

**Additional context**
Manually editing the user's `config.php` to remove the recursive user query will fix the problem

### LINKED ISSUE #6097: [Feature] Anonymous reading from the API
I'm trying to figure out if it is possible for the fever or google api to allow anonymous reading of the default user feed (no authentification required) but prevent deletion/moditication/... api calls  ? I'm specifically talking about the api only, not the web-based interface which I know have an option for this.

**Additional context**
We have a student made app for our university in which we would like to integrate a news section that is an aggregation of all of the department RSS news feeds, the app is freely accessed by all students and doesn't require any login and we would like to keep it that way. We are exploring the use of freshRSS as the backend actual aggregator that do the heavy job, and use the api to get the news from the app but seemingly, we cannot find a public read-only access api for the aggregated articles.

Alkarex:
Share the output of a user query by RSS / HTML / OPML with other people through unique URLs.

Replaces the global admin token, which was the only option (but unsafe) to share RSS outputs with other people.

Also add a new HTML output for people without an RSS reader.
    
fix https://github.com/FreshRSS/FreshRSS/issues/3066#issuecomment-648977890
fix https://github.com/FreshRSS/FreshRSS/issues/3178#issuecomment-769435504

image


Alkarex:
Not completely done, but ready for tests and comments.
Lacking in particular:
* [x] checkbox to enable/disable sharing of a given user query (will also fix the fact that existing user queries need to be saved once before sharing works)
* [x] paging in HTML outputs
* [x] OPML sharing
* [ ] ~~Display user labels?~~
* [x] Fix article tags
* [x] Additional user search
* [x] a bit of documentation (help welcome)
* [x] HTTP Conditional requests (HTTP 304, caching)

Alkarex:
I realised that the current implementation of the user queries was doing lots of redundant SQL queries, also when the user queries were not used. So I did some more refactoring to fix that.

app/Models/BooleanSearch.php
```diff
@@ -74,7 +74,11 @@ private function parseUserQueryNames(string $input): string {
 					$name = trim($matches['search'][$i]);
 					if (!empty($queries[$name])) {
 						$fromS[] = $matches[0][$i];
-						$toS[] = '(' . trim($queries[$name]->getSearch()->getRawInput()) . ')';
+						if ($allowUserQueries) {
+							$toS[] = '(' . trim($queries[$name]->getSearch()->getRawInput()) . ')';
+						} else {
+							$toS[] = '';
+						}
```
ColonelMoutarde, line 81:
```suggestion
						$toS[] = '';
						if ($allowUserQueries) {
							$toS[] = '(' . trim($queries[$name]->getSearch()->getRawInput()) . ')';
						}
```

ColonelMoutarde: Reduce cyclomatic complexity

app/Models/BooleanSearch.php
```diff
@@ -117,7 +117,11 @@ private function parseUserQueryIds(string $input): string {
 					$id = (int)(trim($matches['search'][$i])) - 1;
 					if (!empty($queries[$id])) {
 						$fromS[] = $matches[0][$i];
-						$toS[] = '(' . trim($queries[$id]->getSearch()->getRawInput()) . ')';
+						if ($allowUserQueries) {
+							$toS[] = '(' . trim($queries[$id]->getSearch()->getRawInput()) . ')';
+						} else {
+							$toS[] = '';
+						}
```
ColonelMoutarde, line 124: same

app/Models/BooleanSearch.php
Alkarex: Sorry, that is wrong (you would append twice...)

Alkarex:
image

app/Models/BooleanSearch.php
ColonelMoutarde: Sorry, you’r right

Alkarex:
I have addressed the latest details, so all done here for now (several improvements in particular of the HTML output view are left to future work).
Any comment? Otherwise I will merge shortly.

ColonelMoutarde (approved):

docs/en/users/user_queries.md
```diff
@@ -0,0 +1,63 @@
...
+From the configuration page of the user queries,
+it is possible to share the output of the user queries with external users,
+in the formats HTML, RSS, and OPML:
+
+![Share user query](../img/users/user-query-share.png)
+
+> ℹ️ Note that the sharing as OPML is only available for user queries based on all feeds, a category, or a feed.
+> Sharing by OPML is **not** available for queries based on user labels or favourites or important feeds,
+> to avoid leaking some feed details in an unintended manner.
+
```
math-GH, line 42: Let us think to give a note here, that this feature will be available with 1.24 (it will note, that it is not yet available)

app/views/helpers/configure/query.phtml
```diff
@@ -18,15 +17,53 @@
...
+				</label>
+				<?php if ($this->query->sharedUrlOpml() !== ''): ?>
+				<ul>
+					<li><a href="<?= $this->query->sharedUrlOpml() ?>"><?= _i('link') ?> <?= _t('conf.query.share.opml') ?></a></li>
+				</ul>
+				<?php endif; ?>
+			</div>
+			<p class="help"><?= _i('help') ?> <?= _t('conf.query.share.help') ?></a></p>
+			<p class="help"><?= _i('help') ?> <?= _t('conf.query.help') ?></a></p>
+		</div>
```
math-GH, line 53: could we add a check here, if the API is disabled?
grafik

(btw: the text in the settings needs an update, while it is not only used for the API)

math-GH:
HTML:

There is a pagination. When I click on "Next"
grafik

an "empty" page appears
grafik

math-GH:
There is no log entry if the user query is not available for sharing (`Service Unavailable!` nor if the token is wrong ('User query not found!')

Alkarex:
> an "empty" page appears

Inded (Edit: Now that I think about it, what I did in the main page is to fetch 1 article more than demanded to know whether there is more or not)

> There is no log entry if the user query is not available for sharing

Yes, this is by design as this can better be seen and reacted upon from the Web server logs (e.g. with Fail2Ban or similar) as I would rather avoid giving the possibility to an external client to too easily flood the FreshRSS user logs. We return different HTTP status codes (404, 422, 503...), which are visible in the Web server logs.

app/views/helpers/configure/query.phtml
Alkarex: It is indeed now used for more than the mobile apps.
Note that WebSub, which is also an API, has its own internal option

app/i18n/en/conf.php
```diff
@@ -120,20 +120,33 @@
...
 			'order' => 'Sort by date',
 			'search' => 'Expression',
+			'shareOpml' => 'Enable sharing by OPML of corresponding categories and feeds',
+			'shareRss' => 'Enable sharing by HTML &amp; RSS',
 			'state' => 'State',
 			'tags' => 'Display by label',
 			'type' => 'Type',
 		),
 		'get_all' => 'Display all articles',
+		'get_all_labels' => 'Display articles with any label',
```
math-GH, line 130: Is it obsolete? I cannot find any place where it is used

math-GH: Strange, I cannot find all this "get_...." labels used somewhere....
@Alkarex Could you help me please?

Alkarex: It is because they are used by string concatenation (which I would like to get rid of precisely because it makes validation and maintenance quite challenging). And there is another layer of property with another name for the getter, which makes it even more difficult to search:

https://github.com/FreshRSS/FreshRSS/blob/0ffcf41f9371b9321856d2c709cf7bcb4f7c3c60/app/views/configure/queries.phtml#L57

app/views/helpers/index/tags.phtml
```diff
@@ -0,0 +1,42 @@
...
+	<?php else: // API public access ?>
+		<?php foreach ($firstTags as $tag): ?>
+		<li class="item tag"><a class="link-tag" href="<?= $this->html_url . '&search=%23' . str_replace(' ', '+', htmlspecialchars_decode($tag, ENT_QUOTES)) ?>" title="<?= _t('gen.action.filter') ?>">#<?= $tag ?></a></li>
+		<?php endforeach; ?>
+	<?php endif; ?>
+
+	<?php if (!empty($remainingTags)): // more than 7 tags: show dropdown menu ?>
+		<li class="item tag">
+			<div class="dropdown">
+				<div id="dropdown-tags2-<?= $this->entry->id() ?>" class="dropdown-target"></div>
```
math-GH, line 22: This makes an regression :/
If the tags are shown above and below an article the ID needs to be different from eachother (above: `tags` and below `tags2`). 
With `tags.html`, It is always `tags2`. Any idea how to handle it?

ping @Alkarex

Alkarex: Quick suggestion: https://github.com/FreshRSS/FreshRSS/pull/6990

math-GH: It would fix it but it does not make me happy because of the very large number that is added.

It would be useful if a variable could be set in normal.html (i.e. $foo = 1 for header tags and $foo = 2 for footer tags). And in tags.html `$foo` is used. But I cannot find a solution that makes it possible to have the variable setter in normal.phtml

Alkarex: What is the issue with the number? It is rarely used

PR #6068, opened 2024-01-26
Fix: Favicon with protocol-relative URLs have duplicate slashes.

thomasrenes:
The URL of favicons with protocol-relative URL's already start with "//" so by adding it again it will result in "http:////url" and the icon will never be retrieved.


Alkarex (approved):

Alkarex:
Thanks 👍🏻

PR #6072, opened 2024-01-27
Improve day headline

math-GH:
A bit more modern layout. A better HTML structure (the date is now in the same column as the date for the feeds). A simpler CSS.

Before:
grafik

After:
grafik


Frenzie:
If we assume the date there is something one would want to see at a glance it doesn't make sense to hide it somewhere off to the side. I think this is very similar in effect to just removing it completely.

But I'm not opposed to removing it completely.

math-GH:
I agree to remove the date in the headline. It is already invisible in the mobile view.

It would have the benefit to have a more "light" headline line.

What do you think @Alkarex ?

Alkarex:
I do not really mind :-)

math-GH:
date is now removed

Alkarex:
I find the alignment of the feed title a bit strange. Is it on purpose?

image

math-GH:
I agree. On a wide screen it does not look nice. I changed it. Is it now better?

Alkarex:
The alignment is different, but still a bit strange to me :-)
image

Maybe it should be centred, or aligned to the right, or event to the left. I do not really mind, though

Alkarex:
Anything new regarding the alignment here?

math-GH:
> Anything new regarding the alignment here?

I am not really happy with one of this results. I do not have any better idea right now. Let's keep it as it is. Maybe I will come back later. Is it okay?

PR #6099, opened 2024-02-14
Command Line Parser Concept

kasimircash:
Continuing on from #6036, parsing of command line options has now been cleaned up and wrapped into a `CommandLineParser` Class.
Options on commands are now represented by an `Option` object and are added via the calls to the `addOption()` and `addRequiredOption()` methods on `CommandLineParser` where they are given a name and their properties can be set.

We now have the following concepts associated with an option
- Aliases: Options can be given short and long alias' via `Option`'s constructor and a deprecated alias via the `deprecatedAs()` method.
- Value Taken: The existing behaviour of each option either requiring a value, having an optional value or taking no value is maintained but is now set via one of; `withValueRequired()`, `withValueOptional()` or `withValueNone()`. For `withValueOptional()` a default value can also be set for when the option is used but given no value, e.g. an option with `withOptionalValue('true')` will be given the input 'true' if it is used like a flag.
- Type Taken: Options now have a type they accept that is set via one of; `typeOfString()`, `typeOfBool()`, `typeOfInt()` or `typeOfArrayOfString()`. Inputs for that option must be castable to that type e.g. `typeOfInt()` must be all digits, `typeOfBool()` must be 'true'/'false', 'on'/'off', '1'/'0' etc. An error will be raised if the input is not castable to the wanted type.

Additionally, a default input value can be set as the final argument in `addOption()`. This will be given to the option as input if it does not have any user input. This allows us to have options that will always return a value.

User input is retrieved and parsed via the `parse()` method on `CommandLineParser`. `parse()` must be given a class string to use for constructing it's output, at the moment the we only ever pass `stdClass::class` to it so that we can utilise dynamic properties.
Calls to `parse()` return an object with any error messages generated inside an `errors` property, a usage message for the command under a `usage` property and any valid inputs cast to the correct type under a property with the same name as the option the input was given to.

Invalid option checking on short options has been improved. The short option check introduced in #6036 could return a false positive in a certain edge case, this no longer happens.

We now also handle an option being set multiple times by a user and so having multiple values by returning only the last valid input unless the option has an array type in which case we return all valid inputs as an array.

The only alteration to existing behaviour, other than the extra validation, is that options that address boolean values in config such as `--allow-anonymous`, `--disable-update` etc. have been made to optionally accept boolean values. This allows us to maintain current behaviour of those commands (a flag that implies `true`) while also allowing us to also give them an explicit `true` or `false` should we wish to. Meaning we can now turn these option off from the command line.

Overall this means that we will now have a single abstraction that handles usage messages,  input errors, type checking and type casting, standardising them and so saving us from having to reimplement them inside every command. It also becomes trivial to add options to an existing command and puts things in a good position for any further expansion or refactoring of the command line tools.

All behaviour should remain the same as before with the exception of the above mentioned changes to flags addressing boolean values in config, which can be tested by giving one those options a value e.g. `--disable-update=false` and confirming config has been appropriately updated.

# Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [X] unit tests written
- [ ] documentation updated


cli/CommandLineParser.php
```diff
@@ -0,0 +1,277 @@
...
+		foreach ($userInputs as $input) {
+			preg_match($regex, $input, $matches);
+
+			if(!empty($matches['short'])) {
+				$foundAliases = array_merge($foundAliases, str_split($matches['short']));
+			}
+			if(!empty($matches['long'])) {
+				$foundAliases[] = $matches['long'];
+			}
+		}
```
ColonelMoutarde:
```suggestion
		foreach ($userInputs as $input) {
			preg_match($regex, $input, $matches);
			if (array_key_exists('short', $matches) === false && $matches['short'] !== null) {
				$foundAliases = array_merge($foundAliases, str_split($matches['short']));
			}
			if (array_key_exists('long', $matches) === false && $matches['long'] !== null) {
				$foundAliases[] = $matches['long'];
			}
		}
```
empty()

[empty()](https://www.php.net/manual/fr/function.empty.php) do a lot of things, depending on the type of the parameter.

You always have a better function to do it: * count() for array * > 0 or !== 0 for int * strlen() for string

cli/CommandLineParser.php
```diff
@@ -0,0 +1,277 @@
...
+	 */
+	private function appendUnknownAliases(array $input, $output) {
+		$valid = [];
+		foreach ($this->options as $option) {
+			$valid = array_merge($valid, $option->getAliases());
+		}
+
+		$sanitizeInput = $this->getAliasesUsed($input, $this->makeInputRegex());
+		$unknownAliases = array_diff($sanitizeInput, $valid);
+		if (empty($unknownAliases)) {
```
ColonelMoutarde:
```suggestion
		if ($unknownAliases === []) {
```

cli/CommandLineParser.php
Alkarex: I think `empty` is the better variant here, both in terms of functionality and readability. The suggested alternative is harder to read, more error prove, and incidently slower.

cli/CommandLineParser.php
Alkarex: I can also better like the original, which is more robust

Alkarex:
Thanks @KasimirCash, nice work 👍🏻 
I have added a `.php` extension to one of the new files in order to make sure that out other tests and rules would apply to it too (PHPStan, PHPCS, EditorConfig, etc.)
https://github.com/FreshRSS/FreshRSS/pull/6099/commits/7175edc04c2c348c084e96d9790df6d92a5ad2db

cli/Option.php
```diff
@@ -0,0 +1,100 @@
...
+	public const VALUE_REQUIRED = 'required';
+	public const VALUE_OPTIONAL = 'optional';
+
+	private string $valueTaken = self::VALUE_REQUIRED;
+	/** @var array{type:string,isArray:bool} $types */
+	private array $types = ['type' => 'string', 'isArray' => false];
+	private string $optionalValueDefault = '';
+	private ?string $deprecatedAlias = null;
+
+	public function __construct(private readonly string $longAlias, private readonly ?string $shortAlias = null) {
```
Alkarex: We are still supporting PHP 7.4 a little longer (https://github.com/FreshRSS/FreshRSS/discussions/5474). This line does not seem compatible. Could you please fix?

Tip on how to test:

```sh
docker build --pull -f Docker/Dockerfile-Oldest --tag freshrss/freshrss:oldest .

docker run --rm -e FRESHRSS_ENV=development -v $PWD:/var/www/FreshRSS freshrss/freshrss:oldest sh -c 'bin/composer update && bin/composer test'

docker build --pull -f Docker/Dockerfile-Newest --tag freshrss/freshrss:newest .

docker run --rm -e FRESHRSS_ENV=development -v $PWD:/var/www/FreshRSS freshrss/freshrss:newest sh -c 'bin/composer update && bin/composer test'
```

cli/CommandLineParser.php
```diff
@@ -0,0 +1,277 @@
...
+
+		return $this;
+	}
+
+	/**
+	 * @template T
+	 * @param class-string<T> $target
+	 * @return T
+	 */
+	public function parse(string $target) {
```
Alkarex: This function seems to be always called with `stdClass::class`, isn't it?
The parameter could probably be removed then, to make the calls simpler (`$parser->parse(stdClass::class)` is not very obvious)

cli/CommandLineParser.php
```diff
@@ -0,0 +1,277 @@
...
+		global $argv;
+
+		$output = new $target();
+		$output->errors = [];
```
Alkarex: If you could find a way to avoid dynamic properties, that would be nice. I am trying to get rid of them throughout our codebase, as they are difficult to maintain over time (how to know which ones exist, how to refactor, no IDE autocompletion, no error checks, etc.) compared to more strongly-typed approaches.

Maybe by having some predefined children of the class `Option`?

tests/cli/cli-parser-test.php
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env php
+<?php
+declare(strict_types=1);
+
+require(__DIR__ . '/../../vendor/autoload.php');
+require(__DIR__ . '/CommandLineParserTest.php');
+
+$static_method = getenv('CLI_PARSER_TEST_STATIC_METHOD');
+
+// @phpstan-ignore-next-line
```
Alkarex: Alternative welcome, if possible

tests/cli/CommandLineParserTest.php
```diff
@@ -0,0 +1,213 @@
...
+		return $parser->parse(stdClass::class);
+	}
+
+	private function runCommand(string $static_method, string $options = ''): stdClass {
+		$command = __DIR__ . '/cli-parser-test.php';
+
+		$result = shell_exec("CLI_PARSER_TEST_STATIC_METHOD='$static_method' $command $options 2>/dev/null");
+		$result = is_string($result) ? unserialize($result) : '';
+
+		/** @var stdClass $result */
```
Alkarex: There is probably a need of an explicit PHP cast and not only an annotation (this might explain the PHPStan failures with PHP 8.1).

ColonelMoutarde (approved):

cli/Option.php
kasimircash: I've dealt with the issue with promoted properties along with a few other pieces related to 7.4 under [af0bbcb](https://github.com/FreshRSS/FreshRSS/pull/6099/commits/af0bbcbd61399a617198ca9c2ff7fdf44b487225).

cli/CommandLineParser.php
kasimircash: With [09552c7](https://github.com/FreshRSS/FreshRSS/pull/6099/commits/09552c7e164f1f3599d09a29f0b21ece4c918eff) properties are now set in a 'Definition' `Class` for all commands, this 'Definition' `Class` is now given to the `parse()` method on `CommandLineParser` letting us move away from using dynamic properties.

I had intended to go ahead with something along these lines anyway. I ended up using the `$parser->parse(stdClass::class)` format as it felt like it worked a little more cleanly with how things stand. Although, I certainly agree that moving away from dynamic properties is desirable.

cli/CommandLineParser.php
kasimircash: See https://github.com/FreshRSS/FreshRSS/pull/6099#discussion_r1494907030

tests/cli/cli-parser-test.php
kasimircash: I can't see a good alternative to this as things currently stand. If we moved away from using `getopt()` we could eliminate the `cli-parser-test.php` file altogether and so simplify `CommandLineParserTest` a fair bit. Until then I think we may be stuck with it.

tests/cli/CommandLineParserTest.php
kasimircash: This has been dealt with in [25a673a](https://github.com/FreshRSS/FreshRSS/pull/6099/commits/25a673a8f7429deb1e8d50a596eb24d1bb07c48b) as part of the move away from dynamic properties.

Alkarex: This looks quite good, thanks 👍🏻

cli/CommandLineParser.php
ColonelMoutarde: My version is more verbose, as it is more explicit about what the "if ( )" should test for.
empty() 

empty() does many things, depending on the type of parameter. You always have a better function to do it: * count() for array * > 0 or !== 0 for int * strlen() for string.

The following values are considered empty:

"" (an empty string)
0 (0 as an integer)
0.0 (0 as a float)
"0" (0 as a string)
NULL
FALSE
array() (an empty array)

In this case, the "if" refers to an array(), but you have to read the rest of the code to understand it. You might as well have wanted to know if a string was not empty. With my version, you're sure of what you're looking for and what you're expecting.

cli/CommandLineParser.php
ColonelMoutarde: see https://github.com/FreshRSS/FreshRSS/pull/6099#discussion_r1495970870

cli/CommandLineParser.php
kasimircash: I'd say these are all good points. I think `empty()` is more readable in this case but, as you point out, it can leave ambiguity as to what we are checking for and being more explicit about that is almost always a good thing.

I can't say I really favour solution one or the other here, so I'm happy to go along with whatever the consensus is.

Alkarex:
@KasimirCash I had a number of suggested changes, so I have made a PR to your PR :-)
Please check https://github.com/KasimirCash/FreshRSS/pull/1

* More strongly typed, more object oriented
    * Example: `$options = new ActualizeUserDefinition();` instead of `$options = $parser->parse(ActualizeUserDefinition::class);`
* Addressed PHPStan ignore
* Avoid some variable variables
* Simplify and shorten a bit

tests/cli/cli-parser-test.php
Alkarex: Fixed in https://github.com/KasimirCash/FreshRSS/pull/1

cli/CommandLineParser.php
Alkarex: Improved a bit more in https://github.com/KasimirCash/FreshRSS/pull/1
Example: `$options = new ActualizeUserDefinition();` instead of `$options = $parser->parse(ActualizeUserDefinition::class);`

cli/CommandLineParser.php
Alkarex: Simplified in https://github.com/KasimirCash/FreshRSS/pull/1

Alkarex:
Changes https://github.com/KasimirCash/FreshRSS/pull/1 merged here for easier feedback

kasimircash:
> Changes [KasimirCash#1](https://github.com/KasimirCash/FreshRSS/pull/1) merged here for easier feedback

Sorry about the delay. Okay, so I've had a read through this and everything looks really good to me.

Alkarex:
I have just made some light renaming: https://github.com/FreshRSS/FreshRSS/pull/6099/commits/163ad9a56d50bc252839cdd9670c897a14f9f61f
I was not so found of the many new classes used only once and defined locally (not in their own file), so I have used a bit of syntactic sugar and updated them to anonymous classes, which also helps with regard to the naming of the classes, which I have tried to make a bit more consistent.

Example:

```php
$cliOptions = new class extends CliOptionsParser {
	public string $user;

	public function __construct() {
		$this->addRequiredOption('user', (new CliOption('user')));
		parent::__construct();
	}
};
```

instead of:

```php
final class ActualizeUserDefinition extends CommandLineParser {
	public string $user;

	public function __construct() {
		$this->addRequiredOption('user', (new Option('user')));
		parent::__construct();
	}
}

$options = new ActualizeUserDefinition();
```

Alkarex:
Merged 🚀  Thanks again!

kasimircash:
Thank you for the extensive feed back. It's been really valuable.

cli/create-user.php
```diff
@@ -1,38 +1,93 @@
...
+			'noDefaultFeeds',
+			(new CliOption('no-default-feeds'))->withValueNone()->deprecatedAs('no_default_feeds')
+		);
+		parent::__construct();
+	}
+};
 
-$usernames = listUsers();
-if (preg_grep("/^$username$/i", $usernames)) {
-	fail('FreshRSS warning: username already exists “' . $username . '”', EXIT_CODE_ALREADY_EXISTS);
```
Alkarex, line 16: Removing this test caused a regression, fixed in https://github.com/FreshRSS/FreshRSS/pull/6214

PR #6112, opened 2024-02-18
Web scraping support encodings such as EUC-JP

### LINKED ISSUE #6106: [BUG] HTTP+XPATH scraping retrieves garbled feeds from non UTF-8 contents
**Describe the bug**
Scraping HTML encoded in non-UTF-8 with XPATH gives unreadable results.

**To Reproduce**
1. Add feed: https://www.xfree.ne.jp/support/news.php (EUC-JP encoded)
2. Got unreadable result

**Expected behavior**

I want results that follow the encoding information of the content (e.g. `<meta charset="EUC-JP">`).
From the [example site](https://www.xfree.ne.jp/support/news.php), the following output is obtained, just as it would be seen in a browser.

> - 2024/02/01    サービスの事業会社変更に伴う利用規約改定のお知らせ
> - 2024/01/30    「SSH接続」など9種類の機能追加のお知らせ
> - 2024/01/30    契約情報やサーバー設定に対して用途を記載できるメモ欄追加のお知らせ
> - 2024/01/16    申し込みの増加に伴う『シン・クラウド for Free』の一時的な新規受付停止のお知らせ
> ...

**Screenshots**

Add feed:
image

Got unreadable result:
image

- Device: PC
- OS: Windows 10 Version 22H2 (Build 19045.4046)
- Browser: Vivaldi 6.5.3206.63
- Server OS: Ubuntu 22.04
- FreshRSS version: 1.23.1
- Database version: SQLite
- PHP version: PHP 8.2.7
- Installation type: Docker

**Additional context**

### LINKED ISSUE #6133: Invalid char in most of articles [FULL CONTENT via CSS selector] (SimplePie problem converting to UTF-8)
hi,
I'm not a dev, so please forgive my lack of jargon. Freshrss is wonderful :)
Unfortunately, I'm experiencing inconsistent character encoding issues.. It's possible that I set FreshRSS incorrectly. In my case, the website I'm getting the full article content from uses ISO-8859-2 encoding. From what I understand, freshrss only uses UTF-8 and there is no option to choose the encoding for a specific feed.

**My settings for RSS**
Title: TwojePC.pl
Description: info about HW
Website URL: http://twojepc.pl
Feed URL: https://twojepc.pl/news_rss.php
Type of feed source: RSS / Atom (default)
Article CSS selector on original website: tr:nth-child(2)>td.txt
CSS selector of the elements to remove: table
"Always clear cache" option checked

**To Reproduce**
1. Create an RSS feed as described in "My RSS settings" above;
2. click the "Update feeds" button;
3. Go back to Main stream;
4. Check all feeds (articles with full text) for character encoding.
5. See error: Most articles will have character encoding errors.

**Expected behavior**
Articles should have the correct characters. If the character conversion fails, there should be information about it on the application log page.

**Screenshots**
Screenshot_wrong_characters

**more info**
- Hosting SSD NVMe (provider hitme.pl my plan E-BIZNES)
- OS: Linux (debian - no info about version)
- Browser: Firefox 123
- FreshRSS version: 1.23.1 
- Database version: mysql  Ver 15.1 Distrib 10.6.16-MariaDB, for Linux (x86_64) using readline 5.1
- PHP version: 8.1
- Installation type: Apache 2.4

**Additional context**
I think that twojepc.pl is not the only website that has information about character encoding in the meta tag. For better protection against unexpected encoding, font, and other similar issues, it is recommended to implement a new feature: In the "Manage" section for each RSS feed, a new advanced option could be added: "Enable feed's meta tag for content." This would allow users to customize how FreshRSS retrieves article content.

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6106

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -379,7 +379,7 @@ function stripHtmlMetaCharset(string $html): string {
 
 /**
  * Set an XML preamble to enforce the HTML content type charset received by HTTP.
- * @param string $html the row downloaded HTML content
+ * @param string $html the raw downloaded HTML content
  * @param string $contentType an HTTP Content-Type such as 'text/html; charset=utf-8'
  * @return string an HTML string with XML encoding information for DOMDocument::loadHTML()
  */
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -369,6 +369,15 @@ function cleanCache(int $hours = 720): void {
 }
 
 /**
+ * Remove the charset meta information of an HTML document, e.g.:
+ * `<meta charset="..." />`
+ * `<meta http-equiv="Content-Type" content="text/html; charset=...">`
+ */
+function stripHtmlMetaCharset(string $html): string {
+	return preg_replace('/<meta\s[^>]*charset\s*=\s*[^>]+>/i', '', $html, 1) ?? '';
+}
+
+/**
  * Set an XML preamble to enforce the HTML content type charset received by HTTP.
  * @param string $html the row downloaded HTML content
  * @param string $contentType an HTTP Content-Type such as 'text/html; charset=utf-8'
@@ -381,7 +390,7 @@ function enforceHttpEncoding(string $html, string $contentType = ''): string {
 		return $html;
 	}
 	$httpCharsetNormalized = SimplePie_Misc::encoding($httpCharset);
-	if ($httpCharsetNormalized === 'windows-1252') {
+	if (in_array($httpCharsetNormalized, ['windows-1252', 'US-ASCII'], true)) {
 		// Default charset for HTTP, do nothing
 		return $html;
 	}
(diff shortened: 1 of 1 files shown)


lib/lib_rss.php
```diff
@@ -397,7 +406,20 @@ function enforceHttpEncoding(string $html, string $contentType = ''): string {
...
+		if (is_string($utf8)) {
+			$html = stripHtmlMetaCharset($utf8);
+			$httpCharsetNormalized = 'UTF-8';
+		}
+	}
+	if ($httpCharsetNormalized === 'UTF-8') {
+		// Save encoding information as XML declaration
+		return '<' . '?xml version="1.0" encoding="' . $httpCharsetNormalized . '" ?' . ">\n" . $html;
+	}
+	// Give up
```
Frenzie, line 421: Glancing at this I was wondering if it might make more sense to give up with the charset from the `<meta>` element?

Alkarex: `DOMDocument::loadHTML()` will already sniff information from `<meta charset`. This is the reason why we have to strip this information after conversion to UTF-8.
Furthermore, we cannot always write `<?xml encoding`, in particular for character encodings that are not supersets of ASCII, such as UTF-32.
So as far as this function goes, I do not believe we can do much more.

PR #6114, opened 2024-02-18
No warning for muted feeds

### LINKED ISSUE #6113: Warning icons for disabled feeds
> I am myself running several instances, each with multiple broken feeds, and without problem.

I keep around a number of feeds that once were for various reasons, as a random example `http://feeds.feedburner.com/carbonbrief`. (Once in a while one returns after a hiatus of years!) That shows up as a lot of warnings that I'm not particularly interested in from the main interface.

_Originally posted by @Frenzie in https://github.com/FreshRSS/FreshRSS/issues/6089#issuecomment-1951379351_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6113

### DIFF
--- a/app/views/subscription/index.phtml
+++ b/app/views/subscription/index.phtml
@@ -47,7 +47,7 @@
 
 						$error_class = '';
 						$error_title = '';
-						if ($feed->inError()) {
+						if ($feed->inError() && !$feed->mute()) {
 							$error_class = ' error';
 							$error_title = _t('sub.feed.error');
 						}
--- a/app/views/subscription/index.phtml
+++ b/app/views/subscription/index.phtml
@@ -47,7 +47,7 @@
 
 						$error_class = '';
 						$error_title = '';
-						if ($feed->inError() && !$feed->mute()) {
+						if ($feed->inError()) {
 							$error_class = ' error';
 							$error_title = _t('sub.feed.error');
 						}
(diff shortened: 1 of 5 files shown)


app/views/subscription/index.phtml
```diff
@@ -47,7 +47,7 @@
 
 						$error_class = '';
 						$error_title = '';
-						if ($feed->inError()) {
+						if ($feed->inError() && !$feed->mute()) {
```
Frenzie: Subscription management is quite different from just the UI like in `index/global.html`. Which isn't to say that it definitely should show them, but the considerations are different.

Alkarex: Yes, the consideration is different, but I find it more consistent and also pleasing with this change. Don't you think?

Frenzie: I agree about aesthetically pleasing, but not about consistent. If "muting" and "unmuting" seemingly toggles the error status I find that inconsistent. As such the current display makes sense to me:
image

Alkarex: It was to be consistent with the main UI. The error status of muted feeds is not reliable as it is never updated, so it stays to the last status. I do not really mind reverting this page though

Frenzie: To me it makes sense that in management you'd want more info than in the main UI. In this case it's an implicit clue about why you muted it. For the EFIC example above I'd otherwise wonder why I muted that and I'd unmute it right back.

Put another way, there's muting because it's definitely dead and there's muting because you don't want updates at this time but you don't want to delete it (yet).

Alkarex: Ok, reverted

Alkarex: Good now?

Frenzie: Should be good, yes.

PR #6132, opened 2024-02-27
Minz remove use of deprecated variable

Alkarex:
and related improvements to clarify inheritance
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/6130


app/FreshRSS.php
```diff
@@ -23,8 +23,6 @@ public function init(): void {
 			Minz_Session::init('FreshRSS');
 		}
 
-		Minz_ActionController::$defaultViewType = FreshRSS_View::class;
```
Alkarex, line 26: This was the deprecated code

Issue #6136, opened 2024-02-28
Memory optimisation of API function


Alkarex:
Rewrite this code so that it produces JSON output as stream instead of making a full copy in memory:

https://github.com/FreshRSS/FreshRSS/blob/4b29e666b06762b4b36438c9370c38bc43121f78/p/api/greader.php#L697-L719

> At 16MB, it has worked mostly fine for my test, besides one specific function of the API, which could by the way be good to optimize:

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/pull/6013#discussion_r1506779881_

PR #6137, opened 2024-02-29
Reduce API memory consumption

Alkarex:
`echo json_encode(...)` is very memory demanding for large responses, so optimised.
Contributes to https://github.com/FreshRSS/FreshRSS/issues/6136
https://github.com/FreshRSS/FreshRSS/pull/6013#discussion_r1506779881

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -5,6 +5,24 @@ if (version_compare(PHP_VERSION, FRESHRSS_MIN_PHP_VERSION, '<')) {
 	die(sprintf('FreshRSS error: FreshRSS requires PHP %s+!', FRESHRSS_MIN_PHP_VERSION));
 }
 
+if (!function_exists('array_is_list')) {
+	/**
+	 * Polyfill for PHP <8.1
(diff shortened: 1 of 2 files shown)


Alkarex:
With 16MB of PHP memory, before this patch I could retrieve about ~1000 articles from API (which is precisely the limit used by some clients, so we are just at a tipping point).

```sh
curl -s -H "Authorization:GoogleLogin auth=alice/..." \
  'https://freshrss.example.net/api/greader.php/reader/api/0/stream/contents/reading-list?n=1000' | jq .
```

After this patch, increased to ~1600. The difference will be more significant for users with large articles (my test is mostly composed of short articles).
It is still not as good as what should be done in https://github.com/FreshRSS/FreshRSS/issues/6136 but it contributes to it.

Frenzie:
I feel compelled to mention <https://jsonlines.org/>, although at least in this context it's quite irrelevant. :-)

Alkarex:
> I feel compelled to mention https://jsonlines.org/, although at least in this context it's quite irrelevant. :-)

Yes, that is much better for anything large and for stream processing

PR #6155, opened 2024-03-04
version_compare raise exception if param is not a string

cn-tools:
In file "metadata.json" of extensions it's possible that the data of version is not a string. In this case the check of the version raises an exception. see [issue](https://github.com/cn-tools/cntools_FreshRssExtensions/issues/10)

Ex. [Auto-Refresh-Extension](https://github.com/Eisa01/FreshRSS---Auto-Refresh-Extension)

Maybe the file [generate.php](https://github.com/FreshRSS/Extensions/blob/master/generate.php) could fix this issue with additional line '$metadata['version'] = strval($metadata[' version']);' before line 58 instead of this PR.

Pull request checklist:

- [x] clear commit messages (i hope so)
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/extension/index.phtml
+++ b/app/views/extension/index.phtml
@@ -61,11 +61,11 @@
 						<td>
 							<?= $ext['description'] ?>
 							<?php if (isset($this->extensions_installed[$ext['name']])) { ?>
-								<?php if (version_compare($this->extensions_installed[$ext['name']], $ext['version']) >= 0) { ?>
+								<?php if (version_compare(strval($this->extensions_installed[$ext['name']]), strval($ext['version'])) >= 0) { ?>
 									<span class="alert alert-success">
 										<?= _t('admin.extensions.latest') ?>
 									</span>
-								<?php } elseif ($this->extensions_installed[$ext['name']] != $ext['version']) { ?>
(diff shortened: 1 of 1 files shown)


Alkarex:
Good idea to make a fix in `generate.php`. Could you please do so? And then we can assess whether we also need something more in this PR

app/views/extension/index.phtml
```diff
@@ -61,11 +61,11 @@
 						<td>
 							<?= $ext['description'] ?>
 							<?php if (isset($this->extensions_installed[$ext['name']])) { ?>
-								<?php if (version_compare($this->extensions_installed[$ext['name']], $ext['version']) >= 0) { ?>
+								<?php if (version_compare(strval($this->extensions_installed[$ext['name']]), strval($ext['version'])) >= 0) { ?>
 									<span class="alert alert-success">
 										<?= _t('admin.extensions.latest') ?>
 									</span>
-								<?php } elseif ($this->extensions_installed[$ext['name']] != $ext['version']) { ?>
+								<?php } elseif (strval($this->extensions_installed[$ext['name']]) != strval($ext['version'])) { ?>
```
Alkarex:
```suggestion
								<?php } elseif (strval($this->extensions_installed[$ext['name']]) !== strval($ext['version'])) { ?>
```

PR #6162, opened 2024-03-05
URL: queryID instead of ID

math-GH:
I need some help.

While I worked on #6047 the view buttons (normal view, global view and reader view) will be displayed on every page, also on the configuration settings.

That situation creates a bug that is connected with the URL-parameters.

Step to follow the issue on the current `edge` branch:
- go to user query page
- open config of a user query
- click on save
- the URL will change: more parameters will be added, f.e. the `id` parameter

The view buttons (not in `edge` but in #6047, where them are global visible) will take all URL parameters. In this case: the `id` parameter is added to. When you click on the view button, the error will be shown: `Error 404 - Not found. You are looking for a page that doesn’t exist`

Screenshots made with #6047
grafik

grafik

grafik

grafik

This PR here suggests a renaming of the `id` parameter of the user query configs. More renaming would be necessary.

Is there a better solution for `Minz_Url::display($mode->getUrlParams()` in:

```
foreach ($readingModes as $mode) {
				?>
				<a class="<?= $mode->getId() ?> btn <?php if ($mode->isActive()) { echo 'active'; } ?>" title="<?=
					$mode->getTitle() ?>" href="<?= Minz_Url::display($mode->getUrlParams()) ?>">
					<?= $mode->getName() ?>
				</a>
				<?php
			}
```

### DIFF
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -331,8 +331,8 @@ class FreshRSS_configure_Controller extends FreshRSS_ActionController {
 		$this->view->feeds = FreshRSS_Context::feeds();
 		$this->view->tags = FreshRSS_Context::labels();
 
-		if (Minz_Request::paramTernary('id') !== null) {
-			$id = Minz_Request::paramInt('id');
+		if (Minz_Request::paramTernary('queryId') !== null) {
+			$id = Minz_Request::paramInt('queryId');
 			$this->view->query = $this->view->queries[$id];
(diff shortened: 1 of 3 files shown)


Frenzie:
Is it possible that it says `id= 1` as opposed to `id=1`?

Alkarex:
I have not looked into the problem yet, but my initial intuition is that we should try to fix the fact that there is a parameter at all, which is moved from one action to the other

PR #6181, opened 2024-03-08
SimplePie workaround for Atom cdata type bug

### LINKED ISSUE #6180: [BUG]The subscription source has garbled characters.
**Describe the bug**
The subscription source has garbled characters.
https://peterwunder.de/projects/achievements/atom.xml

**To Reproduce**
It can be reproduced after subscription.

**Expected behavior**
The content is displayed normally.

**Screenshots**
image

- OS: Windows 11 23H2
- Browser: Chrome 122.0.6261.95
- FreshRSS version: 1.23.1
- PHP version: PHP 8.2.7

Alkarex:
Accept feeds with wrong `<content type="CDATA">` instead of `<content type="html">`
fix https://github.com/FreshRSS/FreshRSS/issues/6180


Frenzie (approved):
I was going to ask if that's really necessary, but then I saw your comment about the library. Should it give some kind of log warning for that case though?

Alkarex:
I am more pondering whether this should be merged at all or not. There are many other types of bugs we are coping with, without a log entry. This is more a role for a feed validator. If the source bug is actually https://github.com/lkiesow/python-feedgen/pull/123, it might soon be less relevant (assuming people are updating their libraries...)

Frenzie:
Yes, not merging it makes sense to me.

PR #6202, opened 2024-03-14
fixed HTML: <fieldset> + <legend>

math-GH:
`The <legend> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a caption for the content of its parent [<fieldset>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset).`
(source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend?retiredLocale=de)

This PR cleans it up. Sometimes the `<legend>` is just a `<h2>`headline. Often a wrapping `<fieldset>` improves the semantic.

There is no change in most themes because they do not show the `<fieldset>` frame.

### DIFF
--- a/app/views/configure/integration.phtml
+++ b/app/views/configure/integration.phtml
@@ -13,7 +13,7 @@
 	<form method="post" action="<?= _url('configure', 'integration') ?>" class="draggableList">
 		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
 		<template id="simple-share">
-			<formgroup class="group-share dragbox">
+			<fieldset class="group-share dragbox">
 				<legend draggable="true">##label##</legend>
 				<input type="hidden" id="share_##key##_type" name="share[##key##][type]" value="##type##" data-leave-validation="" />
 				<div class="form-group" id="group-share-##key##">
@@ -27,10 +27,10 @@
 						<button type="button" class="remove btn btn-attention" title="<?= _t('conf.sharing.remove') ?>"><?= _t('gen.action.remove') ?></button>
 					</div>
 				</div>
-			</formgroup>
+			</fieldset>
 		</template>
 		<template id="advanced-share">
-			<formgroup class="group-share dragbox">
+			<fieldset class="group-share dragbox">
 				<legend draggable="true">##label##</legend>
 				<input type="hidden" id="share_##key##_type" name="share[##key##][type]" value="##type##" data-leave-validation="" />
 				<input type="hidden" id="share_##key##_method" name="share[##key##][method]" value="##method##" />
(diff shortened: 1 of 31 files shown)


app/views/configure/integration.phtml
```diff
@@ -13,7 +13,7 @@
 	<form method="post" action="<?= _url('configure', 'integration') ?>" class="draggableList">
 		<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
 		<template id="simple-share">
-			<formgroup class="group-share dragbox">
+			<fieldset class="group-share dragbox">
```
math-GH, line 16: I apologize very much that I added the no-HTML tag `<formgroup>`  here by accident in the past. 🤣

app/views/feed/add.phtml
```diff
@@ -3,7 +3,7 @@
 	/** @var FreshRSS_View $this */
 	if ($this->feed !== null) {
 ?>
-<main class="post">
+<main class="post bookmarklet">
```
math-GH, line 6: I added this class to have a little reminder that this `add.phtml` is just the form for the bookmarklet and not the default feed adding form.
I would prefer to rename this file to make it clear.

app/views/index/logs.phtml
```diff
@@ -10,56 +10,54 @@
...
+			<?php $this->logsPaginator->render('logs_pagination.phtml', 'page'); ?>
+			
+			<div class="form-group form-actions">
+			<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
+				<input type="hidden" name="clearLogs" />
+				<div class="group-controls">
+				<button type="submit" class="btn btn-attention"><?= _t('index.log.clear') ?></button>
+				</div>
+			</div>
+		</form>
```
math-GH, line 57: adding the log table into the form for better semantic

Alkarex:
In the logs view, the messages are centred, which is probably not wanted?

math-GH:
> In the logs view, the messages are centred, which is probably not wanted?

Good catch.
But why is this line in Origine.css needed?
```
form td, form th {
  font-weight: normal;
  text-align: center;
}
```

math-GH:
It is fixed now.
I found 2 tables that are effected by this CSS rule: The config-articleicons table (Config: Display) and the logs table. I hope I have not overseen any tables. The new CSS rule makes it clear for what the line is used for.

Frenzie:
Are we only talking about tables in the FreshRSS config UI? Because if not of course feed content can also have tables. :-)

math-GH:
> Are we only talking about tables in the FreshRSS config UI? Because if not of course feed content can also have tables. :-)

Tables yes, but hopefully not a table in a form (CSS: `form td`).

Frenzie:
I think form would or should be filtered out, but overriding the font-weight for `<th>` doesn't really strike me as desirable except in specific circumstances regardless.

PR #6206, opened 2024-03-14
Improved: Smoother scrolling while focussing articles

math-GH:
Before:
Opening/Closing a article was very "harsh". It focused the article immediately (it "jumps" to it). Same with the navigation via the navigation buttons.

After:
It scrolls smoothly to the next article 

- just a little CSS line

1. normal view: use the navigation buttons and navigate through the articles
2. activate in Reading config: ` Stick the article to the top when opened ` and open/close articles. It jumps smoothly

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -53,6 +53,10 @@
 		url('../fonts/OpenSans.woff') format('woff');
 }
 
+html {
+	scroll-behavior: smooth;
+}
+
 html, body {
 	margin: 0;
 	padding: 0;
--- a/p/themes/base-theme/frss.rtl.css
+++ b/p/themes/base-theme/frss.rtl.css
@@ -53,6 +53,10 @@
 		url('../fonts/OpenSans.woff') format('woff');
 }
 
+html {
+	scroll-behavior: smooth;
+}
+
 html, body {
 	margin: 0;
 	padding: 0;


Alkarex:
Let's try and see whether we get any feedback about this change

Frenzie:
I don't like it. :-/ I find it unbearably slow.

Alkarex:
I do not really mind either way, personally. On the slight minus side, animations tend to consume more power.
There is no way to indicate a speed with that CSS rule, is there?

Alwaysin:
Maybe I need to get used to it but indeed it feels strange and slow.

Alwaysin:
Update: this is unbearably slow when you have a long article, you are at the end of it and use the 'c' shortcut to close it.

WillPresley:
I don't mind it for the most part, but as mentioned by others already, the longer the article, the 'worse' this feels as far as UX goes.

For those who want it, maybe it should be set in Configuration > Reading first. Then either a CSS variable set or just the needed CSS inlined to the markup.

Alkarex:
There is another problem: when opening a long article, scrolling a bit, and then opening the next article, a bug occur: some articles **below** the current article get marked as read.
Let's revert for now

math-GH:
A side note:

The smooth scrolling depends on the browser settings. If it is enabled, then smooth scrolling works, if disabled it is the same behavior as before.

Firefox: 
grafik

math-GH:
> There is another problem: when opening a long article, scrolling a bit, and then opening the next article, a bug occur: some articles **below** the current article get marked as read. Let's revert for now

Is it an issue with enabled ` Stick the article to the top when opened `? Or only with disabled option?

Alkarex:
I only tested with my own settings, which have *Stick the article to the top when opened* enabled

PR #6214, opened 2024-03-19
Fix CLI install regressions

Alkarex:
* It was not possible to pass a blank prefix
* Docker crash when user already exists due to not returning proper exit code anymore. Regression from https://github.com/FreshRSS/FreshRSS/pull/6099


Alkarex:
Merging a bit fast as it prevents some systems from working

PR #6220, opened 2024-03-20
Improve issue template

math-GH:
Using the issue template form:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

New form:
grafik

It will replace the old very text based form:
grafik

Benefit: better UX while writing the bug issue template

How to test the form before committing this PR: see this repository: https://github.com/math-GH/issue_forms/issues


Frenzie (approved):
Better could be debatable (e.g., having to set `-` or something instead of removing irrelevant sections) but sure, why not.

Frenzie:
Actually, what happens with this if you leave something that isn't required empty? Does the system filter it out so it doesn't pollute the issue report?

math-GH:
> Actually, what happens with this if you leave something that isn't required empty? Does the system filter it out so it doesn't pollute the issue report?

Try it out here ;) https://github.com/math-GH/issue_forms/issues

grafik

Alkarex:
Installation type could include git. Web server type would also be good

Frenzie:
I find the "no response" a tad disappointing really, but oh well, it's not too bad.

math-GH:
Less "no response" fields. Added more env information. Made the FreshRSS version mandatory.

grafik

PR #6224, opened 2024-03-21
Counted all favorites

math-GH:
IMHO the total number of favorite articles is useless. There is no similar thing that shows up the absolut number of articles. Furthermore the number cannot be styled.

This PR deletes the number. It reduces 1 SQL query in each page load of a view.

Before:
grafik

After:
grafik


Frenzie:
I don't think it's comparable to the total number of articles. The total number of articles is a meaningless consequence of a few settings while the number of favorites is something that was quite purposefully caused by the user. But I don't mean that as an argument to keep that number, or at the very least not in that location.

math-GH:
> I don't think it's comparable to the total number of articles. The total number of articles is a meaningless consequence of a few settings while the number of favorites is something that was quite purposefully caused by the user. But I don't mean that as an argument to keep that number, or at the very least not in that location.

Thanks for your opinion. Nice argumentation, that could be similar to the labels, isn't it?

math-GH:
It became a little refactoring.

`countUnreadReadFavorites()` is splitted now into `countUnreadFavorites()` and `countFavorites()` (countFavorites() is only used by user-info.php). The return values are now simple `int` instead of arrays.

Frenzie:
> Thanks for your opinion. Nice argumentation, that could be similar to the labels, isn't it?

Yes, I would agree that conceptually favorite is a special label.

math-GH:
ping @Alkarex

Alkarex:
Quick note already: the reason for the single function for counting read+unread favourites is that counting both is similar in terms of costs (time) than counting only one. So calling two distinct functions takes twice as many resources

Alkarex:
I agree that is might be inconsistent with the rest of the UI, but I personally quite like this information as it is quite central to show my backlog. If removed from there, it would be good to have it easily accessible another way

math-GH:
> I agree that is might be inconsistent with the rest of the UI, but I personally quite like this information as it is quite central to show my backlog. If removed from there, it would be good to have it easily accessible another way

Nice point of view. It means, that you use the favorites as "read it later" backlog?

Let me think loud:
favorites: the number of articles has no/less value, but the unread number.
read it later: the unread number is less important but the number of articles.

What if there would be a switch where you can chose: "favorites" (show the unread articles as number) or "read it later" (show the amount of articles as number)

Alkarex:
> you use the favorites as "read it later" backlog?

Yes, and I use labels to keep articles in a categorised way

Alkarex:
> What if there would be a switch where you can chose: "favorites" (show the unread articles as number) or "read it later" (show the amount of articles as number)

We could experiment with something like that

Alkarex:
Related: https://github.com/FreshRSS/FreshRSS/issues/6580

PR #6240, opened 2024-03-27
Empty title guid or first words

### LINKED ISSUE #6218: Feed items with no title
**Describe the bug**
Feed items that have no title parse with a random string of the URL instead of just displaying the first line of text.  (As far as I know title is not required in the RSS spec.)

**To Reproduce**
Subscribe to a feed that doesn’t have titles for some or all feed items such as scripting.com’s feed. Then look at the feed in the reader.

**Expected behavior**
Ignore the empty title and just put body text in there. 

**Screenshots**
Current behavior:
IMG_2921

Example of expected behavior (NetNewsWire on iPhone)
IMG_2920

FreshRSS running via Docker on a RasPi 400. Accessed either through the web client on Chrome or through NetNewsWire syncing on my phone.

math-GH:
Closes #6218 

If a feed has no title then the GUID/URL is shown (example feed: https://mastodon.social/tags/FreshRSS.rss):
grafik

After:
show the first 75 characters. (config via configs, if necessary)
grafik

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
I would rather avoid a new option and just have one behaviour, which works well.
First we need to distinguish the Web UI and the API.
In the Web UI, I think it would be fine to:
* Show the title if not blank
* Otherwise show the first words if not blank
* Otherwise show the GUID

Frenzie:
> I would rather avoid a new option and just have one behaviour, which works well.

Though imo the current behavior works better… well, besides on such a Mastodon feed with meaningless ids.

math-GH:
Now it is simplified

thecptrgod:
That's a good idea for what the interface should look like. (Sorry this is late...been slammed all week.)

Frenzie:
> Though imo the current behavior works better… well, besides on such a Mastodon feed with meaningless ids.

On second thought the overlap of URLs looking nice combined with missing titles is probably minuscule.

app/Models/Entry.php
```diff
@@ -116,7 +116,29 @@ public function guid(): string {
...
 	public function title(): string {
-		return $this->title == '' ? $this->guid() : $this->title;
+		$title = '';
+
+		if ($this->title === '') {
+			// used while fetching the article from feed and store it in the database
+			$title = $this->guid();
+		} else {
+			// used while fetching from the database
+			if ($this->title != $this->guid) {
```
```suggestion
			if ($this->title !== $this->guid) {
```

constants.php
```diff
@@ -40,6 +40,9 @@
 // Maximum log file size in Bytes, before it will be divided by two
 defined('MAX_LOG_SIZE') or define('MAX_LOG_SIZE', 1048576);
 
+// amount of characters of text shown if feed has no title (option: firstWords)
```
Alkarex:
```suggestion
// Amount of characters of text shown if feed has no title
```

app/Models/Entry.php
```diff
@@ -117,7 +117,29 @@ public function guid(): string {
...
+				$content = trim(strip_tags($this->content(false)));
+				$title = trim(mb_substr($content, 0, MAX_CHARS_EMPTY_FEED_TITLE, 'UTF-8'));
+
+				if (strlen($content) > strlen($title)) {
+					$title .= '…';
+				}
+
+				if ($title === '') {
+					$title = $this->guid();
+				}
```
```suggestion
				if ($title === '') {
					$title = $this->guid();
				} elseif (strlen($content) > strlen($title)) {
					$title .= '…';
				}
```

PR #6246, opened 2024-03-30
Add a way to modify CSP rules within an extension

aledeg:
This will allow to change CSP rules to authorize the use of external scripts. We might need to add some safeguard since it will be virtually possible to load any script even malicious one.

- Add support to extension CSP rules

1. Modify the CSP rules in an extension as described in the documentation
2. Validate that the new rule is appended to the existing rules

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


math-GH:
Would it make sense to make it more transparent if an extension is adding external sources to the CSP?

Alkarex:
> Would it make sense to make it more transparent if an extension is adding external sources to the CSP?

Extensions can do worse things than editing the CSP, so while that could be nice if done in a light way, I do not think it is urgent

PR #6249, opened 2024-03-31
Move the action controller in the right folder

aledeg:
Before, the action controller was located in the model folder. Now, the action controller is located in the controller folder.

- See title

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


Alkarex:
Hum, this would break all the extensions, which provide their own controller, e.g.

https://github.com/flusio/xExtension-Flus/blob/6f0a63a70f2802a49b05ae15acad95835c717631/Controllers/billingController.php#L6

I would not mind moving it, but I am not sure it is worth it.
My guess is that the original thoughts were that the (abstract class) `FreshRSS_ActionController` is not a controller in itself but more like a class than can be used to implement a controller. In other words, a model for a controller, even though that is a bit of a shaky concept :-)

aledeg:
I haven't thought about that. It boggles me but I can live with that.

PR #6261, opened 2024-04-04
Make reader icon location configurable

joshka:
Putting the Icons for read/unread and bookmark above the header can break a user's reading flow.

This makes the location of these icons configurable to be placed in one of the following locations:
- above the title
- in the same line as the header
- in the same line as the authors and date

Closes #

- Configuration for icons
- Change icon location in reader view

1. Open settings | config | reading
2. Change option for Icons to each
3. Open reader view and notice the location of the icons changing

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

there didn't seem to be an obvious place for tests / docs for this.

### DIFF
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -177,6 +177,12 @@ return array(
 				'header' => 'In header',
 				'none' => 'None',
 			),
+			'icons' => array(
+				'_' => 'Icons',
+				'above_title' => 'Above title/tags',
+				'with_authors' => 'In authors and date row',
+				'header' => 'In header',
+			),
 			'feed_name' => array(
 				'above_title' => 'Above title/tags',
 				'none' => 'None',
--- a/app/views/configure/reading.phtml
+++ b/app/views/configure/reading.phtml
@@ -136,6 +136,16 @@
 				</div>
 			</div>
 			<div class="form-group">
+				<label class="group-name" for="show_icons"><?= _t('conf.reading.article.icons') ?></label>
+				<div class="group-controls">
+					<select name="show_icons" id="show_icons" data-leave-validation="<?= FreshRSS_Context::userConf()->show_icons ?>">
+						<option value="a" <?= FreshRSS_Context::userConf()->show_icons === 'a' ? ' selected="selected"' : '' ?> data-input-visible="true"><?= _t('conf.reading.article.icons.with_authors') ?></option>
+						<option value="h" <?= FreshRSS_Context::userConf()->show_icons === 'h' ? ' selected="selected"' : '' ?> data-input-visible="true"><?= _t('conf.reading.article.icons.header') ?></option>
+						<option value="t" <?= FreshRSS_Context::userConf()->show_icons === 't' ? ' selected="selected"' : '' ?> data-input-visible="true"><?= _t('conf.reading.article.icons.above_title') ?></option>
+					</select>
+				</div>
+			</div>
+			<div class="form-group">
 				<label class="group-name" for="show_tags"><?= _t('conf.reading.article.tags') ?></label>
 				<div class="group-controls">
 					<select class="select-input-changer" name="show_tags" id="show_tags" data-name="show_tags_max" data-leave-validation="<?= FreshRSS_Context::userConf()->show_tags ?>">
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -121,6 +121,7 @@ class FreshRSS_configure_Controller extends FreshRSS_ActionController {
 			FreshRSS_Context::userConf()->show_tags = Minz_Request::paramString('show_tags') ?: '0';
 			FreshRSS_Context::userConf()->show_tags_max = Minz_Request::paramInt('show_tags_max');
 			FreshRSS_Context::userConf()->show_author_date = Minz_Request::paramString('show_author_date') ?: '0';
+			FreshRSS_Context::userConf()->show_icons = Minz_Request::paramString('show_icons') ?: 't';
 			FreshRSS_Context::userConf()->show_feed_name = Minz_Request::paramString('show_feed_name') ?: 't';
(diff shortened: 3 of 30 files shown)


math-GH:
Thanks for improving the reading view.

Could you please describe what you mean with the users reading flow and how it is connected with these icons?

This is the current situation (above title/tags): The icons are on the left side
grafik

`in header`:
I am not a fan of having the icons on the end of headline. I prefer the left side before the headline
grafik

`in author and date row`:
IMHO I do not see that this makes sense to have the icons here. Also here, I would prefer the left side before the authors list.

grafik

app/i18n/en/conf.php
```diff
@@ -177,6 +177,12 @@
 				'header' => 'In header',
 				'none' => 'None',
 			),
+			'icons' => array(
+				'_' => 'Icons',
```
math-GH, line 181: IMHO: `Icons` is to general. Something like "mark (un)read, favorite buttons" would describe it more precisely

app/i18n/en/conf.php
```diff
@@ -177,6 +177,12 @@
...
+				'above_title' => 'Above title/tags',
+				'with_authors' => 'In authors and date row',
+				'header' => 'In header',
```
math-GH, line 184: `In headline` or `In headline row` could be the better description. Maybe a renaming of the key is needed too.

app/views/configure/reading.phtml
```diff
@@ -135,6 +135,16 @@
 					</select>
 				</div>
 			</div>
+			<div class="form-group">
+				<label class="group-name" for="show_icons"><?= _t('conf.reading.article.icons') ?></label>
+				<div class="group-controls">
+					<select name="show_icons" id="show_icons" data-leave-validation="<?= FreshRSS_Context::userConf()->show_icons ?>">
+						<option value="a" <?= FreshRSS_Context::userConf()->show_icons === 'a' ? ' selected="selected"' : '' ?> data-input-visible="true"><?= _t('conf.reading.article.icons.with_authors') ?></option>
+						<option value="h" <?= FreshRSS_Context::userConf()->show_icons === 'h' ? ' selected="selected"' : '' ?> data-input-visible="true"><?= _t('conf.reading.article.icons.header') ?></option>
+						<option value="t" <?= FreshRSS_Context::userConf()->show_icons === 't' ? ' selected="selected"' : '' ?> data-input-visible="true"><?= _t('conf.reading.article.icons.above_title') ?></option>
```
math-GH, line 144: I prefer to use more readable value keys (authors = above, t = title etc.)

app/i18n/en/conf.php
joshka: I copied this one from other settings which put things in the same area.

I'm not 100% sure what the impact of renaming keys would be - I think it's generally good to keep consistent names where possible, and the fact that there's translations for this already is a good argument to use the same terms.

It might be worthwhile to define the areas more specifically however as something like
- above title
- title
- subtitle
- footer
or something like that.

app/views/configure/reading.phtml
joshka: I copied this from the existing settings that use this approach. I also would have preferred these settings to be meaningful, but chose to keep it consistent rather than doing it right.

joshka:
> Could you please describe what you mean with the users reading flow and how it is connected with these icons?

I'm transitioning off feedly, which looks somewhat like this (oddly appropriate article lol)

image

When I'm scanning through a feed, I'm interested often in seeing the headline first for any article as in a Visual UX hierarchy this is the most important thing. I found myself pausing whenever I go to the next article in FreshRSS as there's a big red icon that takes the first visual cue away from the title.

image

When thinking about how the hierarchy of elements in a UX, the icons for marking the article read/unread or bookmark have a belongs-to relationship with the article, and hence subjectively belong in the authors and dates line. I felt however this perspective was perhaps too controversial to necessarily be correct for everyone's perspective and noted that the general approach in FreshRSS was to provide options for such things.

TL;DR
I want those icons not to be the highest priority thing in an article. This would make my specific reading workflow faster (hit next, repeatedly only stopping to read things that I truly care about based on the title).

app/i18n/en/conf.php
joshka: I agree - I wasn't sure what to put for this. Maybe instead even `Article actions`. This would allow these to be replaced with text in some future change / (or theme).

math-GH: I see `Article icons` is used in the Display config 

grafik

app/i18n/en/conf.php
math-GH: It is from the dropdown above.
grafik

But there it is fine enough to say "header".
In your case all place options are "in the header" ;) It needs to be more specific.

math-GH:
great to see this example. Now I understand it much better. We are both on the same site, that the reading view needs a better article actions tool bar that is also on the bottom of the article (similar to the normal view).

Would it be fine for you if I take some hours/days to think about it? I would create a new PR with a proof of new header/footer layout for the reading view. It could help us for more discussions there. I would appreciate your feedback there. Fine for you?

joshka:
> Would it be fine for you if I take some hours/days to think about it? I would create a new PR with a proof of new header/footer layout for the reading view. It could help us for more discussions there. I would appreciate your feedback there. Fine for you?

Yep - No rush. Ping me with the new PR. I was fixing a personal paper cut. P.S. having the devcontainer made it really easy to contribute a fix.

app/i18n/en/conf.php
joshka: This might be a better UI for the configuration of these things in the reader view too.

app/i18n/en/conf.php
joshka: One downside is that there are translations available for the existing locations mentioned, so changing this to different wording means you lose all the translations. Also, the location options used match the existing locations used in the other options. I figure consistency is key here.

joshka:
Closing out in preference to #6297

PR #6267, opened 2024-04-06
Add core extensions: UserCSS, UserJS

### LINKED ISSUE #227: [Feature] Save CustomCSS settings to user data

I faced the issue of not being able to write CustomCSS settings and solved it in #182.
However, if the extentsions directory has user data, it becomes more complicated to manage. Because the extensions directory is managed outside of the application, but the user data is variable data that the application writes.

I think CustomCSS settings should be stored in the database or in each user's data directory, because they are user data.

For example, [Readable](https://github.com/printfuck/xExtension-Readable) stores settings in `data/users/<username>/config.php`.
If it needs to be saved in an independent file, `data/users/<username>/custom.css` would be better.

Is it a technical difficulty to store CSS in the same way as general user data because it is a file delivered over HTTP?

**Additional context**

hkcomori:
Closes FreshRSS/Extensions#227

- Add UserCSS, which is based on [CustomCSS](https://github.com/FreshRSS/Extensions/tree/master/xExtension-CustomCSS) and changes the file storage location to the user's directory
- Add UserJS, which is based on [CustomJS](https://github.com/FreshRSS/Extensions/tree/master/xExtension-CustomJS) and changes the file storage location to the user's directory

1. Confirmed that extensions can be enabled
1. Confirmed that CSS and JS can be written from the settings page
1. Confirmed that the written contents are saved in `data/users/extensions/<extension-name>`.
1. Confirmed that the written contents can be read by reopening the settings page
1. Confirmed that the written CSS and JS are referenced from `<head>`.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


lib/core-extensions/xExtension-UserCSS/README.md
Alkarex, line 1: You can remove the `xExtension-` prefix for core extensions

lib/core-extensions/xExtension-UserJS/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "UserJS",
+	"author": "hkcomori, Frans de Jonge",
+	"description": "Apply user JS.",
+	"version": "0.0.1",
```
Alkarex: Please bump major version number

Alkarex:
Thanks! We will first issue FreshRSS 1.24 relatively soon and then address bigger changes such as this one in the edge version towards the next release

lib/core-extensions/xExtension-UserJS/metadata.json
hkcomori: OK, I'llchange to 1.0.0.

lib/core-extensions/xExtension-UserCSS/README.md
hkcomori: I removed it.

Alkarex:
I have done some refactoring. Please double-check.

lib/core-extensions/UserCSS/mobile_resolution.png
Alkarex, line 1: Let's reconsider whether those images are relevant or could be removed

lib/core-extensions/UserCSS/desktop_resolution.png
Alkarex, line 1: Let's reconsider whether those images are relevant or could be removed

lib/core-extensions/UserCSS/README.md
```diff
@@ -0,0 +1,89 @@
...
+A FreshRSS extension which give ability to create user-specific CSS rules to apply in addition of the actual theme.
+
+To use it, enable it on the extension panel in FreshRSS. You can add CSS rules by clicking on the manage button.
+
+## Changelog
+
+- 1.0.0 change rules location to user data folder; promoted to core extension
+- 0.0.1 initial version
+
+## Examples
```
Alkarex: We could consider adding something in [our documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs) for core extensions

lib/Minz/Extension.php
```diff
@@ -154,6 +154,23 @@ private function setType(string $type): void {
...
+	}
+
+	/** Return whether a user-specific, extension-specific, file exists */
+	protected final function hasFile(string $filename): bool {
+		return file_exists($this->getExtensionUserPath() . '/' . $filename);
+	}
+
+	/** Return the user-specific, extension-specific, file content, or null if it does not exist */
+	protected final function getFile(string $filename): ?string {
+		$content = @file_get_contents($this->getExtensionUserPath() . '/' . $filename);
```
Alkarex, line 183: In the future, we could try to see whether we could reduce the permissions for the extensions, so they can only call inherited functions

lib/core-extensions/UserCSS/README.md
hkcomori: OK, let's move the description to that document.
Please let me know a few things.

1. Can I add it to [here](https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/users/05_Configuration.md) in the "Extensions" section?
2. Is it acceptable to add only English, because I cannot use French?
3. Can I reduce some of the configuration examples, as I don't think we need many of them?
4. Can I remove the README for this extension?

hkcomori:
> I have done some refactoring. Please double-check.

Thanks, nice refactoring.
I found some descriptions that are no longer used and I removed them.

I think it is a good idea to make some functions `protected`.
But I don't know if all of existing third-party extensions will not be broken.
Can we assume that they are guaranteed by unit tests?

Alkarex:
> Can we assume that they are guaranteed by unit tests

All the extensions of https://github.com/FreshRSS/Extensions are tested automatically, at least

Alkarex:
I have made a new command to automatically test all known third-party FreshRSS extensions https://github.com/FreshRSS/Extensions/pull/228

lib/core-extensions/UserCSS/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "UserCSS",
```
math-GH: Could we give a better name here?
Something like "Tweak UI with CSS" or "UI Tweaker"? Please with spaces instead of CamelCases.

Alkarex: I do not really mind. For me something with CSS in the name would be most obvious, such as "Custom CSS"

Frenzie: I like CSS Tweaks and JS Tweaks, but User CSS / UserCSS and User JS / UserJS are decent enough names.

hkcomori: I don't mind if it is not named "UserCSS", but I did not put a space because this name will be used for the configuration file directory name.
I don't want to include spaces in any file paths.

Although the specifications will be changed, could we use `$this->getEntrypoint` instead of `$this->getName()` for the configuration file directory name?

Alkarex: @hkcomori That is a reasonable suggestion, thanks

hkcomori: Some AI suggestions:

- CSS Customizer
- CSS Personalizer
- Flexible CSS Tuner
- CSS Overwriter
- CSS Remixer

Frenzie: Those suggestions are worse to much worse imho. :-)

hkcomori: OK, I changed the directory name to entrypoint and added a space to the name.

lib/core-extensions/UserCSS/README.md
hkcomori: I updated docs.

lib/core-extensions/UserCSS/mobile_resolution.png
hkcomori: I removed README.
I have judged that this image is unnecessary because it refers to features that are already built into FreshRSS, such as the ability to hide icons.

lib/core-extensions/UserCSS/desktop_resolution.png
hkcomori: I removed README.
I have judged that this image is unnecessary because it refers to features that are already built into FreshRSS, such as the ability to hide icons.

lib/Minz/Extension.php
```diff
@@ -157,7 +157,7 @@ private function setType(string $type): void {
 	/** Return the user-specific, extension-specific, folder where this extension can save user-specific data */
 	protected final function getExtensionUserPath(): string {
 		$username = Minz_User::name() ?: '_';
-		return USERS_PATH . "/{$username}/extensions/{$this->getName()}";
+		return USERS_PATH . "/{$username}/extensions/{$this->getEntrypoint()}";
```
math-GH: Will this cause a new entry in the update.php that changes existing paths?

Alkarex: Could, if anyone is motivated in doing so. Might be a bit tricky to do safely, though. Another option could be to have a migration code in

https://github.com/FreshRSS/FreshRSS/blob/7aaed6092f09b73b2deb7e32a1fed9b776c0202d/lib/Minz/Extension.php#L83

hkcomori: If we expect version jumps, I think it is better method suggested by @Alkarex.
But we can't have a common migration code in `init()`, because it will be overwritten by each extension.
So how about adding a new method that is called just before `init()`?

Alkarex: Ideally, the extensions should call the parent method `parent::init()` when overriding. We could fix those we know like I did in https://github.com/FreshRSS/Extensions/pull/228/

hkcomori: Ok, then I will implement the migration code in `init()`.

hkcomori: It was also necessary to add migration code in `handleConfigureAction()` to support the use case of opening the configuration view before enabling the extension.

PR #6268, opened 2024-04-06
Fix paging posts per page

### LINKED ISSUE #6062: [BUG] loads n-1 articles of "Number of articles per page"
**Describe the bug**
If set to 10, it loads 9 articles. If set to 5 (min number) it loads 4 articles in normal view

grafik

**To Reproduce**
1. set the number in `Display` option
2. useful: disable `Load more articles at the bottom of the page`
3. go to normal view
4. count the loaded articles (it will be n-1 articles)
5. load more articles (it will be n-1 articles each loading)

**Expected behavior**
the amount of loaded articles is as the set number

- FreshRSS version: 1.23.2-dev (also reproduce-able on demo.freshrss.org)

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6062
Due to strange scoping of anonymous function trying to update a static property

### DIFF
--- a/app/Controllers/indexController.php
+++ b/app/Controllers/indexController.php
@@ -247,7 +247,7 @@ class FreshRSS_index_Controller extends FreshRSS_ActionController {
 	 * @return Traversable<FreshRSS_Entry>
 	 * @throws FreshRSS_EntriesGetter_Exception
 	 */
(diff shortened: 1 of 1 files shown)


math-GH (approved):
Works fine now :)

PR #6271, opened 2024-04-06
HTTP Get allow UTF-8 even when charset is far from top

### LINKED ISSUE #5586: [BUG] CSS Selector Retrieves Garbled Text
**Describe the bug**
I am currently subscribing to an RSS feed with the following address: https://rsseverything.com/feed/36a9c080-82fd-44ed-83e8-b286df24e249.xml. I’ve been attempting to use a CSS selector in FreshRSS to extract the article content. I used the selector div.news_text, but the content I obtained was incorrect. When I access the article webpage through my browser, it displays correctly. Despite trying various methods, I haven’t been able to resolve the issue. Could you please assist me? Thank you.

**To Reproduce**
error

**Expected behavior**
Get the content of the article normally

**Screenshots**
IMG_1796
IMG_1797

- Device: iPad Air 5
- OS: iPadOS
- Browser: Safari
- FreshRSS version: 1.21.0
- PHP version: PHP 7.4.33
- Installation type: Docker

**Additional context**

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5586

The case was an HTML document with 15k whitespace then 1.2k of scripts before the `<meta charset="utf-8">` (far from the 1024 bytes suggested by the spec..., and too far for DOMDocument)

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -446,7 +446,7 @@ function enforceHttpEncoding(string $html, string $contentType = ''): string {
 	if ($httpCharset == '') {
 		// No charset defined by HTTP
 		if (preg_match('/<meta\s[^>]*charset\s*=[\s\'"]*UTF-?8\b/i', substr($html, 0, 2048))) {
-			// Favour UTF-8 even if defined too deep in HTML for DOMDocument
+			// Detect UTF-8 even if declared too deep in HTML for DOMDocument
 			$httpCharset = 'UTF-8';
 		} else {
 			// Do nothing
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -444,8 +444,14 @@ function stripHtmlMetaCharset(string $html): string {
 function enforceHttpEncoding(string $html, string $contentType = ''): string {
 	$httpCharset = preg_match('/\bcharset=([0-9a-z_-]{2,12})$/i', $contentType, $matches) === 1 ? $matches[1] : '';
 	if ($httpCharset == '') {
-		// No charset defined by HTTP, do nothing
-		return $html;
+		// No charset defined by HTTP
+		if (preg_match('/<meta\s[^>]*charset\s*=[\s\'"]*UTF-?8\b/i', substr($html, 0, 2048))) {
+			// Favour UTF-8 even if defined too deep in HTML for DOMDocument
+			$httpCharset = 'UTF-8';
+		} else {
+			// Do nothing
+			return $html;
+		}
 	}
 	$httpCharsetNormalized = SimplePie_Misc::encoding($httpCharset);
 	if (in_array($httpCharsetNormalized, ['windows-1252', 'US-ASCII'], true)) {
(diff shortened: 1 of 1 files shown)


lib/lib_rss.php
```diff
@@ -444,8 +444,14 @@ function stripHtmlMetaCharset(string $html): string {
 function enforceHttpEncoding(string $html, string $contentType = ''): string {
 	$httpCharset = preg_match('/\bcharset=([0-9a-z_-]{2,12})$/i', $contentType, $matches) === 1 ? $matches[1] : '';
 	if ($httpCharset == '') {
-		// No charset defined by HTTP, do nothing
-		return $html;
+		// No charset defined by HTTP
+		if (preg_match('/<meta\s[^>]*charset\s*=[\s\'"]*UTF-?8\b/i', substr($html, 0, 2048))) {
+			// Favour UTF-8 even if defined too deep in HTML for DOMDocument
```
Frenzie: Maybe something more like this? "Favor" strikes me as misleading. It's just that it's a better default in 2024 than iso-8859-1 or whatever the PHP default is.
```suggestion
			// Default to UTF-8 if defined too deep in HTML for DOMDocument
```

Alkarex: What I meant is that we do some additional efforts for UTF-8, by searching some more in the HTML content

Alkarex: PHP has no default for that, but HTTP has: Windows-1252 (similar to ISO-8859-1)

Alkarex: New rewording suggestion

Frenzie: Oh wait, I read the code here completely in reverse because it doesn't relate to the OP at all afaict.

OP: "15k whitespace then 1.2k of scripts "

This code: 2048 characters (or probably even just bytes)

Alkarex: This PR also trims the whitespace first

lib/lib_rss.php
```diff
@@ -565,8 +571,11 @@ function httpGet(string $url, string $cachePath, string $type = 'html', array $a
 		// TODO: Implement HTTP 410 Gone
 	} elseif (!is_string($body) || strlen($body) === 0) {
 		$body = '';
-	} elseif ($type !== 'json') {
-		$body = enforceHttpEncoding($body, $c_content_type);
+	} else {
+		$body = trim($body, " \n\r\t");
```
Frenzie: Is overriding the defaults presumed to be better for performance?

Alkarex: Do you mean compared to letting DOMDocument decide, or what overriding more precisely, and what defaults?
In any case, some overriding is needed to convey the HTTP charset information to DOMDocument

Frenzie: No, I mean compared to simply (which also does vertical tabs and nul bytes, presumably uselessly of course):
```suggestion
		$body = trim($body);
```

Alkarex: Ah, for `trim()`, alright, I was afraid of breaking some [encoding BOMs](https://en.wikipedia.org/wiki/Byte_order_mark)

Frenzie: Ah, you mean like UTF-32 or something? Maybe a quick comment to that effect would be useful.

Alkarex: UTF-32BE starts with `\00`, for instance. But vertical tab does not seem to be a problem after all

Alkarex: Done

Frenzie (approved):
lgtm!

PR #6274, opened 2024-04-08
SimplePie replace iframe allow attribute

Alkarex:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#allow
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy#iframes

Besides security, the `allow autoplay` atttribute is especially problematic on mobile (Firefox on Android) as it asks to open the YouTube app as soon as the article is opened.
So we allow what is needed for a YouTube video to work fine, but not autoplay.

Example of code before:

```html
<iframe data-original="https://www.youtube.com/embed/??????feature=oembed"
allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share"
referrerpolicy="strict-origin-when-cross-origin"
allowfullscreen=""
sandbox="allow-scripts allow-same-origin"></iframe>
```

Example of feed https://www.lexpress.fr/arc/outboundfeeds/rss/alaune.xml

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -350,7 +350,7 @@ function customSimplePie(array $attributes = [], array $curl_options = []): Simp
 	$simplePie->add_attributes([
 		'audio' => ['controls' => 'controls', 'preload' => 'none'],
 		'iframe' => [
-			'allow' => 'fullscreen web-share',
+			'allow' => 'accelerometer; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share',
 			'sandbox' => 'allow-scripts allow-same-origin',
 		],
 		'video' => ['controls' => 'controls', 'preload' => 'none'],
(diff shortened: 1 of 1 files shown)


lib/lib_rss.php
```diff
@@ -335,46 +335,49 @@ function customSimplePie(array $attributes = [], array $curl_options = []): Simp
...
-		'audio' => array('controls' => 'controls', 'preload' => 'none'),
-		'iframe' => array('sandbox' => 'allow-scripts allow-same-origin'),
-		'video' => array('controls' => 'controls', 'preload' => 'none'),
-	));
-	$simplePie->set_url_replacements(array(
+		'onkeypress', 'onkeydown', 'onkeyup', 'onselect', 'onchange', 'seamless', 'sizes', 'srcset']));
+	$simplePie->add_attributes([
+		'audio' => ['controls' => 'controls', 'preload' => 'none'],
+		'iframe' => [
+			'allow' => 'accelerometer; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share',
```
Alkarex, line 353: Relevant line, which does **not** include *autoplay*

PR #6277, opened 2024-04-09
fix hotkey trigger for clipboard sharing option

### LINKED ISSUE #6276: [Bug] Numeric hotkey do not trigger clipboard sharing
### Describe the bug

The clipboard sharing option isn't triggered by the corresponding numeric hotkey.

### To Reproduce

1. Enable clipboard sharing.
2. Open the share menu on any feed entry.
3. Use the corresponding numeric hotkey to trigger clipboard sharing.
4. Nothing happens.

### Expected behavior

The menu should close and the link of the feed entry should be copied to the clipboard.

### FreshRSS version

1.23.2-dev

### Environment information

- Database version: SQLite
- PHP version: 7.4
- Installation type: manual
- Web server type: Apache
- Device: any
- OS: any
- Browser: any

### Additional context

_No response_

jakopo87:
Clipboard sharing uses  `<button>` instead of `<a>`, so the hotkey does not work because the selector does not find it.

Closes #6276 

- Fix the selector used to enumerate sharing option;

1. Enable clipboard sharing option;
2. Use the corresponding numeric key of the clipboard sharing;
3. Share dropdown should close and entry link copied to the clipboard;

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

Additional info:
The selector `.dropdown-menu .item a, .dropdown-menu .item button` can be used too, but it's more verbose than the one proposed.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -697,7 +697,7 @@ function auto_share(key) {
 	if (!share) {
 		return;
 	}
-	const shares = share.parentElement.querySelectorAll('.dropdown-menu .item a');
+	const shares = share.parentElement.querySelectorAll('.dropdown-menu .item [data-type]');
 	if (typeof key === 'undefined') {
 		// Display the share div
 		location.hash = share.id;


Frenzie (approved):
That'll be a recent regression. It was changed to `<button>` because it wasn't really a link. :+1:

math-GH:
If it is not a link than do not use `<a>` ;) Will need another solution maybe

math-GH:
> Additional info: The selector `.dropdown-menu .item a, .dropdown-menu .item button` can be used too, but it's more verbose than the one proposed.

I prefer these selectors. It keeps the code cleaner&clearer. The `data-` attributes should be used in CSS as less a possible.

Alkarex:
> I prefer these selectors.

@math-GH Can you please made a *code suggestion* from the review page?

math-GH:
~~The solution made here with this PR does not work for Firefox. Also `.dropdown-menu .item a, .dropdown-menu .item button` does not work.~~

~~The `Print` sharing is broken too (it is also a `button`)~~

Sorry, my fault. But it is very important to know and understand: If the `Default System Sharing` is added to the list, it will not be displayed in all cases (f.e. invisible in Firefox on Win10, but visible in Firefox on Android). It will keeps its hotkey. 

In my case, I pressed the `4` but it was the hidden `Default System Sharing` and not the desired `Clipboard`.
grafik

p/scripts/main.js
```diff
@@ -697,7 +697,7 @@ function auto_share(key) {
 	if (!share) {
 		return;
 	}
-	const shares = share.parentElement.querySelectorAll('.dropdown-menu .item a');
+	const shares = share.parentElement.querySelectorAll('.dropdown-menu .item [data-type]');
```
math-GH, line 700:
```suggestion
	const shares = share.parentElement.querySelectorAll('.dropdown-menu .item.share a, .dropdown-menu .item.share button');
```

math-GH: Another idea could be to add a `share-action` class on the `<button>` and `<a>` in `\app\views\helpers\index\normal\entry_bottom.phtml`. It would make the JavaScript line better: 
`const shares = share.parentElement.querySelectorAll('.dropdown-menu .item.share .share-action');`

Frenzie: You could write it as `.dropdown-menu .item.share :is(a, button)` nowadays.

math-GH: yes, but I do not like to travers on tags, I prefer to travers ids and classes

Frenzie: I think the original proposal with `[data-type]` does that better than adding a classname. `[data-x]` is clearly meant for JS and not for styling.

PR #6289, opened 2024-04-12
Improved: Reading view: author

### LINKED ISSUE #6296: [Bug] Reading view: Search for an author by clicking on the author name
### Describe the bug

If the user clicks on an author user name (with firstname and family name) the search engine cannot find any results

### To Reproduce

(Screens from demo.freshrss.org)
Click on the author's name:
grafik

Search engine searches but cannot find any article:
grafik

Search for the first name: it works:
grafik

Search for the familiy name: it works:
grafik

### Expected behavior

minimum expected, that the article is found where the user comes from

### FreshRSS version

1.23.2-dev (latest)

### Environment information

1) demo.freshrss.org
2) XAMPP on Win10 (PHP8.0)

### Additional context

_No response_

math-GH:
Reader view:

Before:
Author information was smaller then the date (90%).
Just the name was given while the normal view has a "by:" pre-string
grafik

After:
Same font size as the date.
"By:" is given.
grafik

Before:
no author is given but 2 dots are shown:
grafik

After:
no author and just 1 dot to separate the date
grafik

- the `margin: 0 0 10px` moved from the themes into `frss.css`

1. use the reading view
2. have an article with author(s)
3. have an article without author(s) 

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Frenzie (approved):
The size thing is presumably on purpose, but I agree it seems a bit unexpected and unnecessary.

math-GH:
> The size thing is presumably on purpose, but I agree it seems a bit unexpected and unnecessary.

The line was already on `0.9rem` and the author `90%` of it. 
IMHO 0.9rem is small enough.

math-GH:
This PR will fix #6296 too

PR #6291, opened 2024-04-12
Improve performance #5835

### LINKED ISSUE #5835: Improve performance of batch mark as read
@passbe:

> My only comment would be the lag in the interface between selecting the next item and it changing from read to unread in the UI

Let's address that in another PR.
The current batch process timing is: wait *1s, repeat max every second.*
It should be: *wait 0s, repeat max every second.*

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/issues/5812#issuecomment-1803540240_

math-GH:
Closes #5835

- 1s => 0s waiting

1. mark an article as read
2. see that it is faster now

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Frenzie:
A timeout of 0 is meaningless. ;-) That's equivalent to removing the entire asBatch boolean and related functionality. The batching is for performance reasons. See #2199.

math-GH:
0.1s would be better than 1s or 0s?

Alkarex:
When the `mark_read()` is called, check the `asBatch` parameter. The idea is not to flood the server with requests but pool them (especially in the case of marking as read during scroll).
Manually clicking on the envelope to mark an article as read is *not* using the batch logic, so the code you have changed is not even called.
How do you mark an article as read? And please check in the Web browser debugger which code path is triggered.

Alkarex:
Ah right, now I see that it was to address https://github.com/FreshRSS/FreshRSS/issues/5835
But the proposal was to make the logic smarter, not to remove it :-)

Frenzie:
> Manually clicking on the envelope to mark an article as read is _not_ using the batch logic, so the code you have changed is not even called.

That sounds a bit suspect. That means that clicking 10 icons in a row isn't batched while pressing j or k 10 times in a row is. (Which is indeed the case.)  It would seem that either the former should be too or the latter shouldn't either.

PR #6299, opened 2024-04-14
improved: Gap between header and first image

math-GH:
In some rare cases the first image is not wrapped into a `<p>` so that there is no gap between the header and the image (the `<p>` has some margin).

Before:
grafik

After:
grafik

- `padding-bottom` added

1.  a useful feed could be: https://gizmodo.com/rss
2. normal view/reading view
3. see the gap over the image

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Frenzie (approved):
And when there's a `<p>` with margin the margins collapse I presume?

Issue #3649, opened 2021-05-30
[BUG] Security Header


math-GH:
**Describe the bug**
Some security headers are missed. Mostly they are important and improve in general the security while using this tool.
![image](https://user-images.githubusercontent.com/1645099/120119031-c38cbb00-c195-11eb-8c0b-7d6488f29588.png)

General information about the headers: https://owasp.org/www-project-secure-headers/#x-frame-options

**To Reproduce**
used Firefox Addon: https://addons.mozilla.org/en-US/firefox/addon/cookies-and-headers-analyser/

**Expected behavior**
secure headers are set

**Environment information (please complete the following information):**
 - Browser: Firefox 88 + Addon
 - FreshRSS version: 1.18.0

math-GH:
any thoughts about it?

Frenzie:
I believe most of those are irrelevant and not supported, but it would definitely be worthwhile to investigate why it seems to have ideas about CSP that should be incorrect. (Cf. https://github.com/FreshRSS/FreshRSS/issues/1075)

math-GH:
Thanks to point to the CSP discussion.

I think that the x-frame-options and x-xss-protextion are relevant. Why do you think that these flags are irrelevant ?

Frenzie:
You asked for thoughts, so since @Alkarex seems to be occupied I thought I'd give a sign of life, that's all. ;-)

But as to the former, see the site you linked yourself:
<https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html>
> X-Frame-Options Deprecated While the X-Frame-Options header is supported by the major browsers, it was never standardized and has been deprecated in favour of the frame-ancestors directive from the CSP Level 2 specification.

As to the latter, it's simply not supported as I said:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

Alkarex:
Sorry for the delay (my agenda should get better in a few days).
We do have CSP and `X-Content-Type-Options`, but only in the sub-directory `https://freshrss.example.net/i/` as the root is only a redirection (we could have default strict values at the root too, to please such scanners):

https://github.com/flusio/FreshRSS/blob/7802fd33a627dc7c582df871cfa613d9be8f8788/app/Controllers/indexController.php#L87-L92

https://github.com/FreshRSS/FreshRSS/blob/edge/app/FreshRSS.php#L130

[...]

PR #6302, opened 2024-04-14
set X-Frame-Options

math-GH:
Ref. #3649

Set the `X-Frame-Options`

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options


Alkarex:
This header seems obsolete

math-GH:
> This header seems obsolete

As far as I understand it is obsolete if `[frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors)` is used.

Do you prefer to use it instead of the `X-Frame-Options`?

Frenzie:
Yes, `frame-ancestors` is to be preferred afaict.

PR #6303, opened 2024-04-14
Referrer-Policy: same-origin

math-GH:
Ref. https://github.com/FreshRSS/FreshRSS/issues/3649

Set the `Referrer-Policy`

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy


Alkarex:
1. I am not sure disabling Referer is a good idea for the ecosystem. The default behaviour (`strict-origin-when-cross-origin`) seems better to me
2. In this PR, the main place where this would be relevant (the main view) is not addressed for setups not using our default `.htaccess` (e.g. people insisting on using nginx). We should compare with the way we define `Content-Security-Policy` in PHP

math-GH:
> 1. I am not sure disabling Referer is a good idea for the ecosystem. The default behaviour (`strict-origin-when-cross-origin`) seems better to me
> 

I personally prefer `same-origin` because of this:
grafik

`same-origin` keeps my very private used server and its URL privat. The target server thinks I am visiting the article directly.
`strict-origin-when-cross-origin` will tell the target server where I come from.


>     2. In this PR, the main place where this would be relevant (the main view) is not addressed for setups not using our default `.htaccess` (e.g. people insisting on using nginx). We should compare with the way we define `Content-Security-Policy` in PHP

I took the same lines of `Content-Security-Policy` in PHP and .htaccess
Have I have overseen a line?

Frenzie:
Perhaps it makes sense in `config.default.php` or similar? I'm probably more sympathetic to the no referrer point of view myself, but it's definitely a thing where you can imagine people wanting more of a "website" behavior than an "app" behavior.

Alkarex:
Sorry for the delay 🙈  Time is flying...

Alkarex:
Doh, I forgot we already had:
* https://github.com/FreshRSS/FreshRSS/pull/1198
* https://github.com/FreshRSS/FreshRSS/issues/955

Issue #6014, opened 2024-01-02
[Feature]GUI: Clear the logs button at the top instead of at the bottom


vdbhb59:
In the GUI, under the Logs section, the "Clear the logs" action button is placed at the bottom of the page.
Can this button be placed at the top instead? That would be a good option.

There is no specific reason behind it, but it would just be a good to have option, rather than scroll down and then click that button.

math-GH:
Thanks for your idea.

Could you please describe your usecase of the Clear the logs button? How often do you use it?

vdbhb59:
> Thanks for your idea.
> 
> Could you please describe your usecase of the Clear the logs button? How often do you use it?

Hi Mate,
I use it almost every time I am logged in via a browser, be it laptop or mobile. That makes it anywhere between 5-10 times daily. I also use `find/xargs/rm -rf` from my terminal when connected to shared hosting, which is once every two or three days.

Alkarex:
For the record, the logs are automatically truncated using `MAX_LOG_SIZE`, which can be defined in `./FreshRSS/constants.local.php`
Of course, it is good to regularly look at the logs, but not for the sake of disk space alone.

vdbhb59:
I do have it setup in size. I do this daily, as I do like to keep tab on things.
But this ask would be a good to have. Not compulsory, but helps. :+1:

math-GH:
@vdbhb59 Does the sticky button fulfill your needs? It was shipped with 1.24.0

vdbhb59:
Let me check once I am back. Around 4 days.

math-GH:
I close this issue. The sticky buttons are shipped already and it should solve this issue

PR #6304, opened 2024-04-14
Sticky buttons

math-GH:
Inspired by #6014
Pre-work done by #6202

Keep the submit button always reachable. 

have a look onto the logs
grafik

have a look onto the feed configuration
grafik

1. go to `display` option and enable the sticky buttons (last option)
2. see the sticky buttons at the bottom of window

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
I think we should just make the new behaviour the default, without new option (I would like to limit the number of options, as it increases the amount of tests and translations and complexity)

math-GH:
> I think we should just make the new behaviour the default, without new option (I would like to limit the number of options, as it increases the amount of tests and translations and complexity)

reverted ✅

Alkarex:
Nice work! 👍🏻

PR #6305, opened 2024-04-14
Improved: Important feeds icon: bell instead of pin

math-GH:
In #5782 we were looking for an icon for the important feeds section.

Here a draft for a bell icon instead of the pin icon

grafik

- (TBD)
-
-

1. (TBD)
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


math-GH:
What do you think? Is the bell a better icon? Is it worth to change?

What me inspired to use the bell icon: some tools use the bell to highlight "important" messages

Alkarex:
I like it, but I also like the current one, so I do not really mind :-)
On the negative side, it might lead some users to believe that there are special notifications there

Frenzie:
The pin makes a bit more conceptual sense to me (you pinned them as important) while I associate the bell more with messages, but strictly visually speaking it looks decent enough.

math-GH:
great. I agree.

PR #6334, opened 2024-04-18
Fix updated entry filters

### LINKED ISSUE #6331: Apply *mark as read* to updated articles too
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/6330

<sup>Originally posted by **hkcomori** April 18, 2024</sup>
When 'Mark updated articles as unread' triggered, the filter action is not applied.
In my usecase, I set it for articles not wanted to see, so I expect only articles that were not filtered to be unread.

How about your use cases?</div>

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6331

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -573,6 +573,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 						$existingHash = $existingHashForGuids[$entry->guid()];
 						if (strcasecmp($existingHash, $entry->hash()) !== 0) {
 							//This entry already exists but has been updated
+							$entry->_isUpdated(true);
 							//Minz_Log::debug('Entry with GUID `' . $entry->guid() . '` updated in feed ' . $feed->url(false) .
 								//', old hash ' . $existingHash . ', new hash ' . $entry->hash());
 							$entry->_isFavorite(null);	// Do not change favourite state
@@ -585,6 +586,11 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 							if (!($entry instanceof FreshRSS_Entry)) {
 								// An extension has returned a null value, there is nothing to insert.
 								continue;
+							}
+
+							$entry->applyFilterActions($titlesAsRead);
+							if ($readWhenSameTitleInFeed > 0) {
+								$titlesAsRead[$entry->title()] = true;
 							}
 
 							if (!$entry->isRead()) {
@@ -601,6 +607,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 							$entryDAO->updateEntry($entry->toArray());
 						}
 					} else {
+						$entry->_isUpdated(false);
 						$id = uTimeString();
 						$entry->_id($id);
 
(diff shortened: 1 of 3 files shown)


app/Controllers/feedController.php
```diff
@@ -587,6 +588,11 @@ public static function actualizeFeeds(?int $feed_id = null, ?string $feed_url =
 								continue;
 							}
 
+							$entry->applyFilterActions($titlesAsRead);
+							if ($readWhenSameTitleInFeed > 0) {
+								$titlesAsRead[$entry->title()] = true;
+							}
+
```
hkcomori, line 595: `entry_auto_unread` is called before the filter marks articles as read.
I think not calling it for filtered articles because they will be marked as read finally.

Alkarex: The action filters are designed to perform more things than just marking as read (although not everything is implemented yet). Already read articles will be skipped a bit later:

https://github.com/FreshRSS/FreshRSS/blob/0ffcf41f9371b9321856d2c709cf7bcb4f7c3c60/app/Models/Entry.php#L661

But let me know if you spot any bug in the logic

hkcomori: Oh, not that part.
I think the following part calls `entry_auto_unread` before filter action.

https://github.com/FreshRSS/FreshRSS/blob/094ebf92f7c7fd9af5cb72a27198cffd23cdd4be/app/Controllers/feedController.php#L580-L583

Then apply the filter action:

https://github.com/FreshRSS/FreshRSS/blob/094ebf92f7c7fd9af5cb72a27198cffd23cdd4be/app/Controllers/feedController.php#L591

Articles matching filter action are marked as read finally even though `entry_auto_unread` was called.
But when `entry_auto_unread` was called, I will think articles have been marked as unread (in fact it was marked as read).
So, I think it would be better if `entry_auto_unread` only calls for articles that it finally marks as unread.

Please let me know if my understanding is wrong.

hkcomori: Sorry, please forget my comment.

Hook call behavior is a different topic and seems to be consistent with existing policy.
The existing code calls `entry_auto_unread` for no read status change too (unread -> unread).

It seems there is not need to change, because no problems have occurred now.

Alkarex: Thanks for commenting again. I had not had time yet to investigate and was about to do so :-)

app/Controllers/feedController.php
```diff
@@ -587,6 +588,11 @@ public static function actualizeFeeds(?int $feed_id = null, ?string $feed_url =
 								continue;
 							}
 
+							$entry->applyFilterActions($titlesAsRead);
```
Alkarex, line 591: Regression: marking as read for duplicated titles should not apply to updated articles.
Fixed in https://github.com/FreshRSS/FreshRSS/pull/6664

Alkarex:
@hkcomori If you are around, additional tests welcome after https://github.com/FreshRSS/FreshRSS/pull/6664

hkcomori:
@Alkarex Okay, I can't take the time to test specifically, but I'll watch out for any problems while using edge for a while.

Alkarex:
> I'll watch out for any problems while using edge for a while.

Excellent, thanks

Issue #6326, opened 2024-04-17
[i18n] Translation Chinese (简体中文 zh-cn) 🇨🇳 needs translators


math-GH:
Dear FreshRSS-Community,

please help us to improve our Chinese (zh-cn) translations.

While the core developer and many other contributors develop our [FreshRSS ](https://freshrss.org/) app we cannot support the translations in all 20+ languages. 

## Actual Status
Right now about 90% of our text is translated into Chinese (zh-cn).

## How to support us
How to do it:
1) it is very simple
2) all files are stored in `https://github.com/FreshRSS/FreshRSS/tree/edge/app/i18n/zh-cn` ([Link](https://github.com/FreshRSS/FreshRSS/tree/edge/app/i18n/zh-cn))
3) each untranslated text has an inline comment `//  TODO`
4) When the text is translated, just delete the comment
5) If the English term fits as well, please change the comment from `//  TODO` to `//  IGNORE`
6) open a Pull Request and link to this ticket

## Thanks
Thanks a lot. It will help our very international community of users and developers. Your submitted PRs will be published with our next release version [1.24.0](https://github.com/FreshRSS/FreshRSS/milestones) or later

math-GH:
Ping @hoilc @Felix2yu @yzqzss @aidistan @nanhualyq

Could you assist us once more with translating this into Chinese (zh-cn)?
您能再次协助我们将其翻译成中文（zh-cn）吗？

PR #6317, opened 2024-04-17
i18n improved: dotted path -> dot-notation

math-GH:
Ref. #5662
see discussion there https://github.com/FreshRSS/FreshRSS/pull/5662#discussion_r1568434706

Changes proposed in this pull request:

- `dotted path` -> `dot-notation`
- `dotted path` -> `dot-notated`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #6335, opened 2024-04-18
update i18n for zh_CN

nanhualyq:
#6326


math-GH:
I fixed the conflict, because #6317 renamed `dot paths` => `dot notation`.
@nanhualyq Could you please check that the translation is still valid?

nanhualyq:
> I fixed the conflict, because #6317 renamed `dot paths` => `dot notation`. @nanhualyq Could you please check that the translation is still valid?

updated!

PR #6355, opened 2024-04-19
Feature/http response code

ColonelMoutarde:
- http code is now replaced by Enum

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/Models/HttpResponseCode.php
```diff
@@ -0,0 +1,131 @@
...
+	public const MESSAGE_501 = 'Not Implemented';
+	public const MESSAGE_502 = 'Bad Gateway';
+	public const MESSAGE_503 = 'Service Unavailable';
+	public const MESSAGE_504 = 'Gateway Timeout';
+	public const MESSAGE_505 = 'HTTP Version Not Supported';
+	public const MESSAGE_506 = 'Variant Also Negotiates';
+	public const MESSAGE_507 = 'Insufficient Storage';
+	public const MESSAGE_508 = 'Loop Detected';
+	public const MESSAGE_510 = 'Not Extended';
+	public const MESSAGE_511 = 'Network Authentication Required';
```
math-GH: This messages are unused at all.

Could be used maybe in:
[app/Controllers/errorController.ph](https://github.com/FreshRSS/FreshRSS/pull/6355/files#diff-b45443632ee435eff9b2f65c74176c014168f72730248209f3701a4ac785f7c8)

Alkarex:
Plenty of dead code there. In think a short subset could potentially be done after we update the requirements to PHP 8.1+ with enum https://php.net/language.types.enumerations

app/Controllers/apiController.php
```diff
@@ -41,7 +41,7 @@ public static function updatePassword(string $apiPasswordPlain) {
 	 */
 	public function updatePasswordAction(): void {
 		if (!FreshRSS_Auth::hasAccess()) {
-			Minz_Error::error(403);
+			Minz_Error::error(FreshRSS_HttpResponseCode::FORBIDDEN);
```
Frenzie: Granted, my brain developed its own HTTP status code parser literal decades ago, but I find the former immeasurably clearer than the latter.

Frenzie: That's not necessarily a comment on the principle but it's definitely a comment on the execution. :-)

Frenzie: I just realized I'm not offering a concrete solution, but a much clearer pattern exists in the form of for example `HTTP_400_BAD_REQUEST`, `HTTP_422_UNPROCESSABLE_ENTITY`, etc.

See https://www.django-rest-framework.org/api-guide/status-codes/

ColonelMoutarde: I have updated following your recommendations

ColonelMoutarde:
> Plenty of dead code there. In think a short subset could potentially be done after we update the requirements to PHP 8.1+ with enum https://php.net/language.types.enumerations

If we follow the roadmap, it won't be for another 2 years... that's a long time, and a small refacto at that point will be easy.

ColonelMoutarde:
> > Plenty of dead code there. In think a short subset could potentially be done after we update the requirements to PHP 8.1+ with enum https://php.net/language.types.enumerations
> 
> If we follow the roadmap, it won't be for another 2 years... that's a long time, and a small refacto at that point will be easy.

[Now use real Enum](https://github.com/FreshRSS/FreshRSS/pull/6355/commits/0652c7cecaea75e612f2f0e3c2ec869b3b610451)

app/Controllers/apiController.php
```diff
@@ -40,7 +40,7 @@ public static function updatePassword(string $apiPasswordPlain): string|false {
 	 */
 	public function updatePasswordAction(): void {
 		if (!FreshRSS_Auth::hasAccess()) {
-			Minz_Error::error(403);
+			Minz_Error::error(FreshRSS_HttpResponseCode::HTTP_403_FORBIDDEN->value);
```
Alkarex: I suggest to not use `->value` but instead change `Minz_Error::error(...)` to accept an `FreshRSS_HttpResponseCode`

ColonelMoutarde: Ok, but in this case, the 404 is not the default value ? Minz_Error:: error() has 404 by default.

Frenzie: I think he just wants it without the `->value` at the end if reasonably possible in Minz_Error. ^_^

ColonelMoutarde: I've thought things through and tested things, getting rid of “->value” isn't possible.
For example:
```php

var_dump(FreshRSS_HttpResponseCode::HTTP_500_INTERNAL_SERVER_ERROR->value);
var_dump(FreshRSS_HttpResponseCode::HTTP_500_INTERNAL_SERVER_ERROR);
var_dump(FreshRSS_HttpResponseCode::class);

result
int(500)
enum(FreshRSS_HttpResponseCode::HTTP_500_INTERNAL_SERVER_ERROR)
string(25) "FreshRSS_HttpResponseCode"
```
I think it's impossible not to use “->value”.

ColonelMoutarde: I try another approach

ColonelMoutarde: In this approach, no default value

app/Controllers/subscriptionController.php
```diff
@@ -13,7 +13,8 @@ class FreshRSS_subscription_Controller extends FreshRSS_ActionController {
 	#[\Override]
 	public function firstAction(): void {
 		if (!FreshRSS_Auth::hasAccess()) {
-			Minz_Error::error(403);
+			Minz_Error::error(FreshRSS_HttpResponseCode::HTTP_
+                        3_FORBIDDEN);
```
Frenzie: typo

Frenzie (approved):
Fine by me, but it looks like PHPStan is less happy.

app/Controllers/subscriptionController.php
ColonelMoutarde: resolved

ColonelMoutarde:
> Fine by me, but it looks like PHPStan is less happy.

All tests successfull

PR #6369, opened 2024-04-21
Refactoring: Rename dotpath into dotnotation

math-GH:
Follow up of #6317

renaming from dot path to dot notation

(for a better review: go through each commit. I tried to bundle it in each commit)

It will affect exported OPMLs that were already exported and hat dotpath information. But it was published as dev version, so I would say that it is okay to rename the attribute.

### DIFF
--- a/app/Services/ImportService.php
+++ b/app/Services/ImportService.php
@@ -161,7 +161,7 @@ class FreshRSS_Import_Service {
 				case strtolower(FreshRSS_Export_Service::TYPE_XML_XPATH):
 					$feed->_kind(FreshRSS_Feed::KIND_XML_XPATH);
 					break;
-				case strtolower(FreshRSS_Export_Service::TYPE_JSON_DOTPATH):
+				case strtolower(FreshRSS_Export_Service::TYPE_JSON_DOTNOTATION):
 					$feed->_kind(FreshRSS_Feed::KIND_JSON_DOTNOTATION);
 					break;
 				case strtolower(FreshRSS_Export_Service::TYPE_JSONFEED):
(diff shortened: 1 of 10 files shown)


app/Services/ImportService.php
```diff
@@ -161,8 +161,8 @@ private function createFeed(array $feed_elt, FreshRSS_Category $category, bool $
 				case strtolower(FreshRSS_Export_Service::TYPE_XML_XPATH):
 					$feed->_kind(FreshRSS_Feed::KIND_XML_XPATH);
 					break;
-				case strtolower(FreshRSS_Export_Service::TYPE_JSON_DOTPATH):
-					$feed->_kind(FreshRSS_Feed::KIND_JSON_DOTPATH);
+				case strtolower(FreshRSS_Export_Service::TYPE_JSON_DOTNOTATION):
```
Alkarex, line 164: It might be sufficient to add another case line here to support OPML exports with the previous name

Alkarex: Done in https://github.com/FreshRSS/FreshRSS/pull/6369/commits/cf17221f778f80c653b255798624fdb550e95b0d
Test welcome

Alkarex (approved):

app/Services/ImportService.php
math-GH: I am not so a big fan of having legacy code because of changed feature within a dev release, but it is ok.

Alkarex: Tested to work.
`edge` is our rolling release, and there are likely many users of the JSON feature already, so avoiding breaking changes is always good. I have myself this feature deployed on multiple instances already.

PR #6371, opened 2024-04-22
Remove noise parameters in query URL

Alkarex:
Remove undeeded `name` parameters from saved user query


app/Controllers/configureController.php
```diff
@@ -398,6 +398,8 @@ public function queryAction(): void {
 			if (!empty($params['shareOpml']) && ctype_digit($params['shareOpml'])) {
 				$queryParams['shareOpml'] = (bool)$params['shareOpml'];
 			}
+			unset($queryParams['name']);
+			unset($queryParams['rid']);
```
Frenzie: Fine by me, but there seems to be a problem?

Alkarex: Fixed

PR #6373, opened 2024-04-23
Fix: overflow expanding title in entry header

math-GH:
## Before
Config article icons (default):
grafik

(A) mouse over an article expands the title with overlapping the date
grafik

If some icons are disabled
grafik

(B) The title expanding is buggy and in some cases the title does not expand - it shrinks
grafik

If the summary is enabled
grafik

(C) The title cannot be expanded by mouse over
grafik

## After
(A) Similar as before
grafik

(B) The title does not shrinks while hovering
grafik

(C) The title expands
grafik

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -51,8 +51,8 @@
 		endif;
 	?></li><?php
 	endif; ?>
-
-	<li class="item title<?= (($topline_thumbnail !== 'none') || $topline_summary) ? ' multiline' : '' ?>" dir="auto"><a target="_blank" rel="noreferrer" href="<?= $this->entry->link() ?>" class="item-element"><?= $this->entry->title() ?><?php
+	<li class="item titleAuthorSummaryDate">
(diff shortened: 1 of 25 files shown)


Alkarex:
Ups, regression: the date is always visible, also on small screen, making titles nearly impossible to read on mobile

math-GH:
> Ups, regression: the date is always visible, also on small screen, making titles nearly impossible to read on mobile

I forgot to commit a little change :/

Alkarex:
Another regression from here I think: the hover colour of the date in dark theme (Origine) is darker than non-hover, which is likely wrong:

image

math-GH:
That is not an regression, it is by purpose. The font color gets an opacity for a better title overlapping view.

Maybe we could improve the opacity

Alkarex:
> That is not an regression, it is by purpose. The font color gets an opacity for a better title overlapping view.

Ah right, but when there is no overlap, it looks strange. I believe it should have the same colour than the title

Alwaysin:
@math-GH This PR broke the suspension points in Dark theme by removing the dark filter on them.
Before:
chrome_z2Q0q2GFt6
After, whiter and brighter:
chrome_7K12cpxMi2

math-GH:
First I could confirm it. Font color is `#888888` but the points are `#cccccc`

After trying something I cannot reproduce it. Strange.

colinstu12:
Looking forward to this fix! Thanks! Ran into this issue myself and didn't think to report.

PR #6390, opened 2024-04-27
Improved: title hover: date opacity 0.3 -> 0.5

math-GH:
Improves https://github.com/FreshRSS/FreshRSS/pull/6373#issuecomment-2080063273

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1287,7 +1287,7 @@ a.website:hover .favicon {
 }
 
 .flux:not(.current):hover .flux_header .item .date {
-	opacity: 0.3;
+	opacity: 0.5;
 }
 
 .flux:not(.current):hover .item .title {
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1286,10 +1286,6 @@ a.website:hover .favicon {
 	z-index: 2;
 }
 
-.flux:not(.current):hover .flux_header .item .date {
-	opacity: 0.5;
-}
-
 .flux:not(.current):hover .item .title {
 	background-color: inherit;
 }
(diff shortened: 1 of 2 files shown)


Alkarex:
But now it fails in light mode :-P

image

The date should not become less visible when the article is highlighted, and it should probably have the same colour than the title

math-GH:
Thanks for your feedback. Would you suggest to revert the date behavior (do not change the font color while hovering)?

Alkarex:
> Would you suggest to revert the date behavior (do not change the font color while hovering)?

I have not looked at the implementation in details, but I think the date should have the same colour as the title, and if the title overlaps with the date, then use another mechanism to either hide the date completely or use a background for the title.

math-GH:
> > Would you suggest to revert the date behavior (do not change the font color while hovering)?
> 
> I(...) I think the date should have the same colour as the title, 

that is not the current behaviour. I have not checked each theme in detail, but f.e. the Origine theme has a grey font color of the date.

> and if the title overlaps with the date, then use another mechanism to either hide the date completely

it would hide the date also if the title does not overlap the date column.

>  or use a background for the title.

It is already implemented. But I cannot add too much padding between the last letter of the title and the border, because if could cut the last letters.

Alkarex:
>> I think the date should have the same colour as the title,

> that is not the current behaviour

Indeed, my bad, I was not remembering correctly.

Even with this PR, the date becomes *less* visible when hovering compared to when not hovering, which is what troubles me. It is a regression compared to version 1.23.1. It would be nice to fix if possible.

math-GH:
Back to the behavior of 1.23.1 while hovering.

This screenshot shows the issue while hovering and having the same font color of author and date:
grafik

Alkarex:
Yes, there is a possible improvement there, but let's come back to it later. Maybe a CSS background opacity gradient or something else to mitigate the issue.

PR #6297, opened 2024-04-13
Reading view: action icons position

math-GH:
inspired by #6261 from @joshka

### (default): above title
grafik

grafik

### in authors and date row
grafik

grafik

### top line article icons configs respected
grafik

How to test the feature manually:

1. go to config `reading`, section `Articles: header/footer`
2. edit the option
3. see in reading view

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #6395, opened 2024-04-27
New: Sharing articles from the article title line

math-GH:
It adds a column for sharing articles directly in the normal view article row

grafik

config via display config (default: not selected)
grafik

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
The reading view is untouched yet. It will prevent conflicts with #6297. Sharing will be added later.

Alkarex:
I like the idea, but I would like to suggest a more powerful solution: make a menu from the title line as you have, but allowing multiple functions, the sharing one being a sub-menu.
Examples of functions (which can be added in other PRs) include: mark previous articles as read, reload full content, search similar articles, etc.

app/views/helpers/index/normal/entry_header.phtml
```diff
@@ -52,26 +53,72 @@
...
+				<a class="dropdown-toggle" href="#dropdown-share2-<?= $this->entry->id() ?>" title="<?= _t('index.share') ?>">
+					<?= _i('share') ?>
+				</a>
+
+				<ul class="dropdown-menu">
+					<li class="dropdown-header"><?= _t('index.share') ?> <a href="<?= _url('configure', 'integration') ?>"><?= _i('configure') ?></a></li><?php
+						$id = $this->entry->id();
+						$link = $this->entry->link();
+						$title = $this->entry->title() . ' · ' . ($this->feed === null ? '' : $this->feed->name());
+						foreach (FreshRSS_Context::userConf()->sharing as $share_options) {
```
Alkarex: Implementation consideration: this adds a lot of data in the output (consume resources server side, at network level, and client side) for a non-essential shortcut feature. To save resources and improve performances, I think this should be left to a dynamically-generated menu like we have for the configuration drop-down menu of the feeds in the aside bar.

Example:
https://github.com/FreshRSS/FreshRSS/blob/329fd4bcf6504c74e3906e51c6fc2124bc09cc02/app/layout/aside_feed.phtml#L153-L189

Or (probably even better) use an HTML template element https://developer.mozilla.org/docs/Web/HTML/Element/template like:
https://github.com/FreshRSS/FreshRSS/blob/329fd4bcf6504c74e3906e51c6fc2124bc09cc02/app/views/configure/integration.phtml#L15-L31

math-GH:
@Alkarex Thanks for your feedback.

Let me explain my roadmap that I have in mind:
General idea is to have the features "sharing" and "my labels" in the header too. It is not a good UX to have both just in the footer. The options are already there in the table "article actions" (as disabled checkboxes). Let the users decide if they need it or not. If the user does not need it it will not have an HTML overhead. If the user decides to have this option in the header and not in the footer anymore: plus minus zero overhead ;)
If the user decides to have the sharing menu in the header and disable other buttons (f.e. the reading icon and the "open on website" icon), then there is enough space.

TODO: I was thinking to use the templates too, because this menu is created in each article (also before this PR). I wanted to do this refactoring after this PR to keep focus on this feature and do not mix it with the refactoring of the existing sharing menu.

The submenu thing: yes, I am thinking about it too, but there is some more brain power needed ;) Right now the code does not have any sub menu features. A sub menu needs one more click/interaction than just a button (I can remember some discussion about having 1 more click to more-or-less important features). I think it will be important to have this sub menu feature and more features as you described. We could play here with CSS container queries too, so that it is usable on small and big screens.

My roadmap in bullet points to summarize it:

1. waiting for merging this PR (need to wait a bit for release of 1.24.0 and maybe a bit more for bugfixes of 1.24.0 -> 1.24.1) and keep this code free of merging conflicts
2. refactor the menu to use a template instead of duplicated HTML code
3. add "my labels" feature in similar way
4. (maybe) thinking about having the article tags in the header too (I can remember there are some feature requests)
5. optimize the UI and "compress" actions into a sub menu

While doing it we should not forget to have a good integration into the reading view too.

I am bit excited to implement it :)

math-GH:
> 2. refactor the menu to use a template instead of duplicated HTML code
> 3. add "my labels" feature in similar way

A search could be add there too: #6595

math-GH:
@Alkarex Could we merge this PR?

Alkarex:
There seems to be a regression in the theme (title colours):
image

Alkarex:
And please double-check my light refactoring https://github.com/FreshRSS/FreshRSS/pull/6395/commits/6e55595005ce8904cbcc0af4a37b99a5abcd09bd

math-GH:
> There seems to be a regression in the theme (title colours)

fixed. it was issued by merge conflicts.

math-GH:
> And please double-check my light refactoring [6e55595](https://github.com/FreshRSS/FreshRSS/commit/6e55595005ce8904cbcc0af4a37b99a5abcd09bd)

Nice! Thanks!

arseru:
This new feature is really really great, thanks a lot! Do you know how could I achieve the same sharing functionality you added for Whatsapp, but for Telegram?

PR #6421, opened 2024-05-01
Fix user query filter display

### LINKED ISSUE #6360: [Bug] User Queries: different displays - depends on selecting main stream or not
### Describe the bug

While trying to reproduce #5674 I found following bug:

main stream selected: user query box is "empty" (`no filter`)
grafik

selecting another stream: user query box lists the parameters
grafik

### To Reproduce

1. Go to User Queries
2. chose the type
3. submit
4. see the box in the user queries overview page

### Expected behavior

minimum the `expression` is listed in the overview.

`no filter` message only if nothing is selected like this example (screen shot from demo.freshrss.org, that has an older version than the current 1.23.2-dev):
grafik

### FreshRSS version

1.23.2-dev (latest)

### Environment information

- Database version: SQLite
- PHP version: PHP 8.0

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6360

### DIFF
--- a/app/Models/UserQuery.php
+++ b/app/Models/UserQuery.php
@@ -155,22 +155,22 @@ class FreshRSS_UserQuery {
 
 	/**
 	 * Check if the user query has parameters.
-	 * If the type is 'all', it is considered equal to no parameters
 	 */
 	public function hasParameters(): bool {
-		if ($this->get_type === 'all') {
-			return false;
+		if ($this->get_type !== 'all') {
+			return true;
 		}
 		if ($this->hasSearch()) {
 			return true;
 		}
-		if ($this->state) {
+		if (!in_array($this->state, [
+				0,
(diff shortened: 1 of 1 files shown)


math-GH:
Quite better as before PR.

What I think is not good working right now:
"Empty filter" text, if there is no expression and the main stream chosen.
grafik


If you select this query f.e. in the view of favorites, than it would jump to the main stream.
If you would select this query in the main stream, but you have the "show only unread articles, then it would show unread and read articles after reloading.

One more thought: Does it make sense to keep the "no filter" text? I would say that every query is a valid filter.

If the expression is mandatory then we should mark it as mandatory (via require attribute)

Alkarex:
The search expression is not mandatory. We could indeed remove the *no filter*. Let's keep that for another PR, as it is less urgent for 1.24.0.

PR #6423, opened 2024-05-02
Fix CSS selector preview

### LINKED ISSUE #6266: [Bug] Preview of CSS selector uses the one saved and not the one entered
### Describe the bug

I find it misleading that when previewing a CSS selector, it uses the one that is saved and not the one entered. When one doesn't really know what to input there and goes by trial and error, it is very cumbersome to each time save the settings and go back to the config to be able to preview the selector.

Plus, I'm not sure if "CSS selector of the elements to remove" is taken into account in the previewing, and if it is the case it probably also needs to be saved.

### To Reproduce

1. Go to a feed settings
2. Change the value of "Article CSS selector on original website"
4. Click the preview button
6. See the old value used

### Expected behavior

Using the newly entered value

### FreshRSS version

latest edge

### Environment information

- Database version: MariaDB 10.11
- PHP version: PHP 8.2
- Installation type: git
- -Web server type: Apache
- Device: Laptop
- OS: Win11
- Browser: Chrome 124

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6266

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -1121,7 +1121,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 		$feed_id = Minz_Request::paramInt('id');
 		$content_selector = Minz_Request::paramString('selector');
 
-		if (!$content_selector) {
+		if ($content_selector === '') {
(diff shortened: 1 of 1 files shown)


math-GH (approved):
works fine as expected :)

PR #6426, opened 2024-05-02
Fix CSS selector encoding

### LINKED ISSUE #6229: [BUG] Article selector is HTML escaped
**Describe the bug**
The article CSS selector is being passed to CSSXPath with some characters encoded as XML entities. The selector `[id="main"]` becomes `[id=&quot;main&quot;]` which is converted to `.//*[@id="&quot;main&quot;")]`

**To Reproduce**
1. Input an Article CSS Selector with double quotes
2. Submit due to #6228
3. Click the preview button by the selector
4. CSSXpath has now been passed a mangled selector (I debugged by adding an `error_log` statement to Translator.php)

**Expected behavior**
The selector should be unescaped by the time it reaches CSSXpath.

- Device: ThinkPad A485
- OS: Void Linux
- Browser: Firefox 124.0
- FreshRSS version: 1.23.2-dev 1c0b8a7
- Database version: PostgreSQL 16.2
- PHP version: 8.2.7
- Installation type: Docker self hosted

### LINKED ISSUE #6266: [Bug] Preview of CSS selector uses the one saved and not the one entered

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6229
fix https://github.com/FreshRSS/FreshRSS/issues/6266#issuecomment-2090432818

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -1148,6 +1148,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			return;
 		}
 		$feed->_pathEntries($content_selector);
+		$feed->_attribute('path_entries_filter', Minz_Request::paramString('selector_filter', true));
 
 		//Fetch & select content.
 		try {
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -748,7 +748,8 @@ HTML;
 			}
 
 			$content = '';
-			$nodes = $xpath->query((new Gt\CssXPath\Translator($feed->pathEntries()))->asXPath());
+			$cssSelector = htmlspecialchars_decode($feed->pathEntries(), ENT_QUOTES);
+			$nodes = $xpath->query((new Gt\CssXPath\Translator($cssSelector))->asXPath());
 			if ($nodes != false) {
 				$path_entries_filter = $feed->attributeString('path_entries_filter');
 				foreach ($nodes as $node) {


math-GH:
A point of view from UI/UX: If the "eye" button uses both inputs then both inputs should have its own `<fieldset>` with its own buttons. One button could be the "eye" button then. (others: save, cancel)
grafik

mrnoname1000:
From some quick testing this seems to fix both bugs!

math-GH:
While testing I found this bug:

if the last element of this list is empty (last character is a comma):
grafik

grafik

Alkarex:
Indeed @math-GH , this UX could be improved. There are also all the other parameters below, which are not used by the preview. But let's merge the fixed bugs already.

> if the last element of this list is empty (last character is a comma)

Let me try to make another PR for that

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6428

PR #6428, opened 2024-05-02
CSS selector trim

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/pull/6426#issuecomment-2091159868

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -749,11 +749,13 @@ HTML;
 
 			$content = '';
 			$cssSelector = htmlspecialchars_decode($feed->pathEntries(), ENT_QUOTES);
+			$cssSelector = trim($cssSelector, ', ');
 			$nodes = $xpath->query((new Gt\CssXPath\Translator($cssSelector))->asXPath());
 			if ($nodes != false) {
-				$path_entries_filter = $feed->attributeString('path_entries_filter');
+				$path_entries_filter = $feed->attributeString('path_entries_filter') ?? '';
+				$path_entries_filter = trim($path_entries_filter, ', ');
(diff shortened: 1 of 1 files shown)


math-GH:
yes and not enough.

Because this situation (`ul>li, ,,,, ul`) triggers the warning too
grafik

Alkarex:
Right, but then it is a fix to be done in the library https://github.com/PhpGt/CssXPath/
I am not sure the current PR is especially relevant, then.

math-GH:
Ping @g105b 
Maybe he could take a look if it is fixable in CssXPath

math-GH:
> Right, but then it is a fix to be done in the library https://github.com/PhpGt/CssXPath/ I am not sure the current PR is especially relevant, then.

okay. This PR works fine while trimming the last comma. It is better than before

math-GH (approved):

g105b:
Hello!

I would be happy to help with changes to CssXPath. As far as I can see from the linked thread, the issue you are facing is that the CssXPath library throws an exception if the last character of the selector is a comma. Is my understanding correct?

It would be useful to compare this behaviour to what happens within the web context. For example, open your browser's developer tools and execute this: `document.querySelector("p,");` - in Firefox and Chrome, this throws an exception "Uncaught DOMException: Document.querySelector: 'p,' is not a valid selector".

Do you think it's still worth making a change to the CssXPath library?

Alkarex:
Indeed, I do not believe we should attempt to support invalid CSS selectors.

g105b:
I agree. If the specification states that an exception should be thrown with an invalid selector, we should throw the expected exception, and then a library like this can catch that exception.

PR #6439, opened 2024-05-07
Fix OPML import CURLOPT_PROXYTYPE

### LINKED ISSUE #6438: [Bug] Export Import OPML file: PROXY_TYPE is not imported
### Describe the bug

Export - Import feed settings via OPML lose the proxy address.

### To Reproduce

Feed settings. Have a proxy with protocol and address.
grafik

Export it via OPML file.

`frss:CURLOPT_PROXYTYPE` is in the OPML file

grafik

Import this OPML file (before: delete to feed, so that you can import it again)

Address is given, but the protocol is lost:
grafik

save the settings via `save` button.

Address is lost too, because the protocol select list was empty

grafik

### Expected behavior

after import: proxy address is restored

### FreshRSS version

1.23.2-dev (latest)

### Environment information

- Database version: SQLite
- PHP version: PHP 8.0

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6438

### DIFF
--- a/app/Services/ImportService.php
+++ b/app/Services/ImportService.php
@@ -284,7 +284,7 @@ class FreshRSS_Import_Service {
 				$curl_params[CURLOPT_PROXY] = $feed_elt['frss:CURLOPT_PROXY'];
 			}
 			if (isset($feed_elt['frss:CURLOPT_PROXYTYPE'])) {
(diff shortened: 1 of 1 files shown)


math-GH (approved):
tested. it works fine

PR #6476, opened 2024-05-21
Fix JSON Dot direct array

Alkarex:
Allow consuming a JSON array in which entries are top level
https://github.com/FreshRSS/FreshRSS/issues/6474


K4LCIFER:
Thanks! One note: shoudln't it be `$` instead of `.` to reference the root array? I was looking at [this](https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html), and it mentions:
> `$` 	The root object or array.

Alkarex:
> shoudln't it be `$`

I have just added support for it as well

PR #6523, opened 2024-06-01
Remove Referer header from SimplePie feed request

willhbr:
This seems unnecessary, and the equivalent behaviour was removed for other downloads in [this PR](https://github.com/FreshRSS/FreshRSS/pull/4372).

### DIFF
--- a/lib/SimplePie/SimplePie/File.php
+++ b/lib/SimplePie/SimplePie/File.php
@@ -106,7 +106,6 @@ class SimplePie_File
 				curl_setopt($fp, CURLOPT_FAILONERROR, 1);
 				curl_setopt($fp, CURLOPT_TIMEOUT, $timeout);
 				curl_setopt($fp, CURLOPT_CONNECTTIMEOUT, $timeout);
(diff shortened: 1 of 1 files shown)


Alkarex:
Sorry for the delay, due to the refactoring of using SimplePie (big mess!) https://github.com/FreshRSS/FreshRSS/pull/4374
Our SimplePie fork is now maintained on https://github.com/FreshRSS/simplepie so feel free to resubmit there, or even better upstream https://github.com/simplepie/simplepie (but the upstream repo is very slow).
Regarding this PR, it would be nice to test with plenty of feeds. We have seen a bunch, which are sensitive to user-agent, cookies, and what not, which should in principle not be needed. I have not tested specifically for Referer recently though, but I would not be surprised to break some feeds by removing it.

Alkarex:
Custom HTTP headers added https://github.com/FreshRSS/FreshRSS/pull/6820
Ready to remove Referer

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/6811

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6822

PR #6541, opened 2024-06-04
add image and description to user query share

### LINKED ISSUE #6538: image and description on user query share
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/6537

<sup>Originally posted by **TomW1605** June  4, 2024</sup>
hi
is there a way to add a description and feed image to the rss feed from a user query?

i am using FreshRSS to process the patreon feeds of some podcasts to split out normal and patreon episodes so they are different feeds in my podcast app. unfortunatly they dont have the image or description of the input feed and that makes them quite hard to tell apart.

i understand this is not at all the intended usecase for FreshRSS but other than this issue it is something it works extreemly well for so if there was a way to do this it would be basicaly perfect.
thanks</div>

TomW1605:
Closes #6538

- add a field for feed description to user queries
- add a field for feed image url to user queries
- put these fields in the shared rss feed

1. add a user query
1. go to the user query config
1. add a description and image url
1. enable sharing by rss
1. save
1. load the rss from the link
1. the description and image in the rss should match the valuse entered in step 3

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/index/rss.phtml
+++ b/app/views/index/rss.phtml
@@ -9,7 +9,7 @@
 	<channel>
 		<title><?= $this->rss_title ?></title>
 		<link><?= $this->html_url ?></link>
-		<description><?= _t('index.feed.rss_of', $this->rss_title) ?></description>
+		<description><?= $this->description ?></description>
 		<pubDate><?= date('D, d M Y H:i:s O') ?></pubDate>
 		<lastBuildDate><?= gmdate('D, d M Y H:i:s') ?> GMT</lastBuildDate>
 		<atom:link href="<?= $this->rss_url ?>" rel="self" type="application/rss+xml" />
--- a/app/views/index/rss.phtml
+++ b/app/views/index/rss.phtml
@@ -3,7 +3,7 @@
 	/** @var FreshRSS_View $this */
 ?>
 <?= '<?xml version="1.0" encoding="UTF-8" ?>'; ?>
-<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:media="http://search.yahoo.com/mrss/"
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:media="http://search.yahoo.com/mrss/" xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd"
 	<?= $this->rss_base == '' ? '' : ' xml:base="' . $this->rss_base . '"' ?>
 >
 	<channel>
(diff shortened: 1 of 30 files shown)


Alkarex:
Looks good 👍🏻 
You can use our CLI to add new translations:

```sh
cli/manipulate.translation.php -a add -k 'conf.query.image_url' -v 'Image URL'
```

TomW1605:
thanks, ill remember that for next time (and thanks for neatening up that if)

Alkarex:
You are welcome :-)
About the process: we are just about to release FreshRSS 1.24.1 (final tests welcome!) so there is a feature freeze for now. As soon as it is out, this PR can be merged in our `edge` rolling-release branch.

app/views/index/rss.phtml
```diff
@@ -3,16 +3,22 @@
...
 >
 	<channel>
 		<title><?= $this->rss_title ?></title>
 		<link><?= $this->html_url ?></link>
-		<description><?= _t('index.feed.rss_of', $this->rss_title) ?></description>
+		<description><?= $this->description ?></description>
 		<pubDate><?= date('D, d M Y H:i:s O') ?></pubDate>
 		<lastBuildDate><?= gmdate('D, d M Y H:i:s') ?> GMT</lastBuildDate>
 		<atom:link href="<?= $this->rss_url ?>" rel="self" type="application/rss+xml" />
+		<itunes:image href="<?= $this->image_url ?>"></itunes:image>
```
Alkarex: Is the `<itunes:image />` needed at all when there already is the `<image />` one?

TomW1605: im honestly not sure. i added it because it was in the feed i was working off. my only testing podcast player is Overcast and it works without it but also the dev seems to have a very good understanding of the podcast RSS standards and makes sure Overcast can handle things so im not sure other apps will be as robust. there seem to be a whole range of "itunes" tags that are variously used in different feeds. there is also one for description `<itunes:summary>`.

edit: i just tested with apple podcasts and it also works without the itunes tag, so i think it would be reasonable to remove it. i cant test any android podcast apps since i only have an iphone.

Alkarex: Yes please for the test. I would rather add it again later if we identify a need

TomW1605: cool, ill remove it and the itunes xml definition

app/views/index/rss.phtml
```diff
@@ -9,10 +9,15 @@
 	<channel>
 		<title><?= $this->rss_title ?></title>
 		<link><?= $this->html_url ?></link>
-		<description><?= _t('index.feed.rss_of', $this->rss_title) ?></description>
+		<description><?= $this->description ?></description>
```
Alkarex, line 12: Regression fixed in https://github.com/FreshRSS/FreshRSS/pull/6548

PR #6544, opened 2024-06-05
Pass PHPStan level 9

Alkarex:
Passing PHPStan Level 9 https://phpstan.org/user-guide/rule-levels

### DIFF
--- a/lib/Minz/Url.php
+++ b/lib/Minz/Url.php
@@ -26,7 +26,7 @@ class Minz_Url {
 
 		$url_string = '';
 
-		if ($absolute) {
+		if ($absolute !== false) {
 			$url_string = Minz_Request::getBaseUrl();
 			if (strlen($url_string) < strlen('http://a.bc')) {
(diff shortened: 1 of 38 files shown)


lib/Minz/Url.php
```diff
@@ -94,27 +96,23 @@ private static function printUri(array $url, string $encodage): string {
...
-				if (!is_string($key) || (!is_string($param) && !is_int($param))) {
+				if (!is_string($key) || (!is_string($param) && !is_int($param) && !is_bool($param))) {
 					continue;
 				}
 				$uri .= $separator . urlencode($key) . '=' . urlencode((string)$param);
 				$separator = $and;
 			}
 		}
 
-		if (!empty($url['#']) && is_string($url['#'])) {
```
Alkarex, line 105: @math-GH If you can, please check that this has not caused any regression since https://github.com/FreshRSS/FreshRSS/pull/4416

PR #6555, opened 2024-06-12
Sticky Footer Regression dark mode

math-GH:
Regression #6304

Before:
Origine dark mode had a light background.

After:
Dark background


Alkarex:
I thought it was on purpose :-P

PR #6615, opened 2024-07-08
Initial support for PHP 8.4

Alkarex:
Sole fix needed so far seems to be related to https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

See also upstream PR https://github.com/PhpGt/CssXPath/pull/227

We are also hitting a PHP bug with `min()` function, reported upstream https://github.com/php/php-src/issues/14873

### DIFF
--- a/Docker/Dockerfile-Newest
+++ b/Docker/Dockerfile-Newest
@@ -5,11 +5,11 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories && \
 	apk add --no-cache \
 	tzdata \
-	apache2 php83-apache2 \
+	apache2 php84-apache2 \
 	apache-mod-auth-openidc \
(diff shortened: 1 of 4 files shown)


ColonelMoutarde:
hi, for help with version upgrades there's rector_php.
i use it at work, it's very handy
https://github.com/rectorphp/rector

Alkarex:
> hi, for help with version upgrades there's rector_php. i use it at work, it's very handy https://github.com/rectorphp/rector

It does not seem to support PHP 8.4, and not even PHP 8.3, does it?

> Rector now supports upgrades from PHP 5.3 to 8.2

ColonelMoutarde:
Support for php8.4 will arrive later this year. 
You can already get rid of all the things that will no longer be supported or obsolete with php 7.4/8.1.
we'll have less work to do for 8.4.

Alkarex:
This line fails with PHP 8.4.0alpha1 when OPCache is enabled, while the rest seems fine, at a first glance
https://github.com/FreshRSS/FreshRSS/blob/58c297f823b4a1fdfc203499cf7662a76079c6f4/app/Models/Feed.php#L333

PR #6642, opened 2024-07-19
add OIDCSessionMaxDuration and OIDCSessionInactivityTimeout

### LINKED ISSUE #6640: [Bug] OIDC not save session
### Describe the bug

FreshRSS via OIDC constantly loses the session, i.e. if you close the tab and come back in thirty minutes or an hour later, you will have to log in again.

Is it possible to make the session last longer? Like in FreshRSS without OIDC, for example.

### To Reproduce

1. Log in to FreshRSS via OIDC
2. Close the tab for an hour or two.
3. Here comes the bug, you have to log in again.

### Expected behavior

FreshRSS via OIDC constantly loses the session, i.e. if you close the tab and come back in thirty minutes or an hour later, you will have to log in again.

Is it possible to make the session last longer? Like in FreshRSS without OIDC, for example.

### FreshRSS version

 1.24.2-dev 

### Environment information

- Database version: PostgreSQL 16
- PHP version: [e.g. PHP 8.1]
- Installation type: Docker
-Web server type: Apache > nginx
- Device: Web [Desktop]
- OS: Win11, MacOS 15
- Browser: Firefox 128

### Additional context

_No response_

crisukbot:
Closes #6640

- OIDC Saves Session (cookies)

1. Install on instance
2. Login and wait
3. No need to log in every 5 minutes via OIDC.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


Alkarex:
Thanks! 
A few things to improve: 
* Compatibility with our Alpine images (currently assumes Debian)
* Fix risk of having multiple times those two configuration lines when the container is restarted
* Do not affect containers without OIDC enabled

Hint: I think some initial configuration should be put in our various Dockerfiles(s) to ease addressing the points above. 

Try to see if you can fix some of it, otherwise I will come back to it later

Docker/entrypoint.sh
```diff
@@ -23,7 +23,8 @@ fi
 
 if [ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ]; then
 	# Debian
-	(which a2enmod >/dev/null && a2enmod -q auth_openidc) ||
+	(which a2enmod >/dev/null && a2enmod -q auth_openidc)
+ 	(echo "OIDCSessionMaxDuration ${OIDC_SESS_MAX_DURATION:-2592000}" >> /etc/apache2/mods-available/auth_openidc.conf && echo "OIDCSessionInactivityTimeout ${OIDCSessionInactivityTimeout:-86400}" >> /etc/apache2/mods-available/auth_openidc.conf) ||
```
Alkarex: I believe this will still append the configuration lines multiple times, each time the container is restarted

Docker/entrypoint.sh
```diff
@@ -26,6 +26,9 @@ if [ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ]; then
 	(which a2enmod >/dev/null && a2enmod -q auth_openidc) ||
 		# Alpine
 		(mv /etc/apache2/conf.d/mod-auth-openidc.conf.bak /etc/apache2/conf.d/mod-auth-openidc.conf && echo 'Enabling module auth_openidc.')
+	OIDC_CONF=$(test -e /etc/apache2/mods-available/auth_openidc.conf && echo /etc/apache2/mods-available/auth_openidc.conf || echo /etc/apache2/conf.d/mod-auth-openidc.conf)
+	echo "OIDCSessionMaxDuration ${OIDC_SESS_MAX_DURATION:-2592000}" >> "$OIDC_CONF"
+	echo "OIDCSessionInactivityTimeout ${OIDCSessionInactivityTimeout:-86400}" >> "$OIDC_CONF"
```
Alkarex: I believe this will still append the configuration lines multiple times, each time the container is restarted, won't it?

Alkarex: I am taking a look

Alkarex:
Would this approach work for you instead? https://github.com/FreshRSS/FreshRSS/pull/6642/commits/3ff95a5da4a77b27923490625dfb18e7bc8bc0d3
And let's wait to see if there is a demand for environment variables.

Alkarex:
Could you please check that it works for you, before we merge?

Alkarex:
> when OIDC is enabled, anonymous browsing is not available because OIDC redirects to Authentik

Indeed, so far, anonymous browsing is only available when using *Web forms* as login method, because otherwise the whole `/i/` subfolder gets protected by HTTP Basic Auth. The easiest would likely be to add another entry-point for anonymous reading, but that would nevertheless require some work.
At the moment, what you can try is the Web view of *user queries*, which should be available https://freshrss.github.io/FreshRSS/en/users/user_queries.html#share-your-user-queries

Alkarex:
> This solution doesn't work.

Arg, surprising. Did you rebuild the container to test, or how did you do?
That file is copied to `/etc/apache2/sites-available/` during build.

Alkarex:
Thanks. Could you please try this other variant?
And I would be interested in more debugging to understand why the configuration lines in `Docker/FreshRSS.Apache.conf` did not seem to work as they should have

Alkarex:
Merging, but tests / feedback still welcome

Alkarex:
Follow-up https://github.com/FreshRSS/FreshRSS/pull/6730

PR #6650, opened 2024-07-22
New: sharing service archive.is added

math-GH:
Closes #6635

- `archive.is` added (it is the same as `archive.ph` just another domain)

1. add `archive.is` to the sharing services
2. share a link via `archive.is`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/i18n/fr/gen.php
+++ b/app/i18n/fr/gen.php
@@ -204,7 +204,7 @@ return array(
 	),
 	'share' => array(
 		'Known' => 'Sites basés sur Known',
-		'archiveIS' => 'archive.is',	// TODO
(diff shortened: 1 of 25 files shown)


app/i18n/fr/gen.php
```diff
@@ -204,6 +204,7 @@
 	),
 	'share' => array(
 		'Known' => 'Sites basés sur Known',
+		'archiveIS' => 'archive.is',	// TODO
```
```suggestion
		'archiveIS' => 'archive.is',	// IGNORE
```

And for (most or all) other languages as well

math-GH: I forgot it 🙈

PR #6659, opened 2024-07-26
Add default API CORS HTTP Headers

Alkarex:
To allow interacting with our APIs from a JavaScript application.
So far limited to the APIs: Greader, User queries.
Fix https://github.com/FreshRSS/FreshRSS/discussions/6654#discussioncomment-10131144


VincenzoKoestler (commented):
Just tested it out and it seems to work great, I can access the json feed from my javascript application without issues so far.

PR #6661, opened 2024-07-29
Strong type array parameter helper

Alkarex:
Also useful for extensions (including one I am writing)


Alkarex:
Tested to work fine. Quick merge due to dependency

PR #5470, opened 2023-06-15
Refactor lastSeen and markReadAsGone

Alkarex:
Make the logic a bit more robust and explicit.
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/5404

PR #6663, opened 2024-07-30
Fix markAsReadUponGone regression

Alkarex:
Regression from https://github.com/FreshRSS/FreshRSS/pull/5470
Was not working anymore when the feed was empty.
Plus simplification of the logic when the feed is not empty

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -515,7 +515,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 				} else {
 					$newGuids = $feed->loadGuids($simplePie);
 					$entries = $feed->loadEntries($simplePie);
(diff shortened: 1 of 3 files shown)


app/Controllers/feedController.php
```diff
@@ -515,7 +515,7 @@ public static function actualizeFeeds(?int $feed_id = null, ?string $feed_url =
 				} else {
 					$newGuids = $feed->loadGuids($simplePie);
 					$entries = $feed->loadEntries($simplePie);
-					$feedIsEmpty = $simplePiePush !== null && empty($newGuids);
+					$feedIsEmpty = $simplePiePush === null && empty($newGuids);
```
Alkarex, line 518: Regression was there https://github.com/FreshRSS/FreshRSS/pull/5470

PR #6664, opened 2024-07-30
Fix regression on mark duplicate titles as read for modified articles

Alkarex:
When the option *mark as read if an identical title already exists in the top n newest articles*, that should not apply to updated articles, as they risk being marked as read due to themselves having the same title.
Identifying the different sub-cases would require a more complicated logic, so disable for now.
Regression due to https://github.com/FreshRSS/FreshRSS/pull/6334

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -583,7 +583,8 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 								continue;
 							}
 
-							$entry->applyFilterActions($titlesAsRead);
+							// NB: Do not mark updated articles as read based on their title, as the duplicate title maybe be from the same article.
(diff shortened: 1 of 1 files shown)


GitWo262:
With 1.24.3 I still have this problem 
freshrss_double

Alkarex:
@GitWo262 It does not seem to be related to this PR, and in your case, articles are visibly from two different feeds. This mechanism only applies to the articles of the same feed

GitWo262:
> @GitWo262 It does not seem to be related to this PR, and in your case, articles are visibly from two different feeds. This mechanism only applies to the articles of the same feed

Sorry then I am wrong. I thought the feature is working overall all feeds.

Alkarex:
I believe we have a ticket somewhere to add this feature at category and global levels

PR #6666, opened 2024-07-30
Minor update whitespace PHPCS rules

Alkarex:
To simplify our PHPCS configuration, apply more rules, and be clearer about what is added or removed compared with PSR12.
Does not change our current conventions, but just a bit more consistent.

Useful commands:

```sh
make test-all
make fix-all
# or individually:
composer run-script phpcs
composer run-script phpcbf
```

### DIFF
--- a/app/Controllers/entryController.php
+++ b/app/Controllers/entryController.php
@@ -79,31 +79,31 @@ class FreshRSS_entry_Controller extends FreshRSS_ActionController {
 			} else {
 				$type_get = $get[0];
 				$get = (int)substr($get, 2);
-				switch($type_get) {
-				case 'c':
-					$entryDAO->markReadCat($get, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
-				case 'f':
-					$entryDAO->markReadFeed($get, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
-				case 's':
-					$entryDAO->markReadEntries($id_max, true, null, FreshRSS_Feed::PRIORITY_IMPORTANT,
-						FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
-				case 'a':
-					$entryDAO->markReadEntries($id_max, false, FreshRSS_Feed::PRIORITY_MAIN_STREAM, FreshRSS_Feed::PRIORITY_IMPORTANT,
-						FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
-				case 'i':
-					$entryDAO->markReadEntries($id_max, false, FreshRSS_Feed::PRIORITY_IMPORTANT, null,
-						FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
-				case 't':
-					$entryDAO->markReadTag($get, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
-				case 'T':
-					$entryDAO->markReadTag(0, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
-					break;
+				switch ($type_get) {
+					case 'c':
+						$entryDAO->markReadCat($get, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
+						break;
+					case 'f':
+						$entryDAO->markReadFeed($get, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
+						break;
+					case 's':
+						$entryDAO->markReadEntries($id_max, true, null, FreshRSS_Feed::PRIORITY_IMPORTANT,
+							FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
+						break;
+					case 'a':
+						$entryDAO->markReadEntries($id_max, false, FreshRSS_Feed::PRIORITY_MAIN_STREAM, FreshRSS_Feed::PRIORITY_IMPORTANT,
+							FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
+						break;
+					case 'i':
+						$entryDAO->markReadEntries($id_max, false, FreshRSS_Feed::PRIORITY_IMPORTANT, null,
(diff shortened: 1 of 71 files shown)


app/Controllers/entryController.php
```diff
@@ -79,31 +79,31 @@ public function readAction(): void {
 			} else {
 				$type_get = $get[0];
 				$get = (int)substr($get, 2);
-				switch($type_get) {
-				case 'c':
```
Alkarex, line 83: Most modified lines are due to wrongly indented `switch` cases.

Alkarex:
@math-GH FYI, in VS Code, the extension https://marketplace.visualstudio.com/items?itemName=ValeryanM.vscode-phpsab will offer live feedback about those PHPCS rules, including whitespace:

image

image

Alkarex:
Any comment @Frenzie ?

app/FreshRSS.php
```diff
@@ -74,10 +74,10 @@ private static function initAuth(): void {
 			if (!FreshRSS_Context::hasSystemConf() || !(FreshRSS_Auth::isCsrfOk() ||
 				(Minz_Request::controllerName() === 'auth' && Minz_Request::actionName() === 'login') ||
 				(Minz_Request::controllerName() === 'user' && Minz_Request::actionName() === 'create' && !FreshRSS_Auth::hasAccess('admin')) ||
-				(Minz_Request::controllerName() === 'feed' && Minz_Request::actionName() === 'actualize'
-					&& FreshRSS_Context::systemConf()->allow_anonymous_refresh) ||
-				(Minz_Request::controllerName() === 'javascript' && Minz_Request::actionName() === 'actualize'
-					&& FreshRSS_Context::systemConf()->allow_anonymous)
+				(Minz_Request::controllerName() === 'feed' && Minz_Request::actionName() === 'actualize' &&
```
Frenzie, line 77: This is less immediately obvious, but probably no big deal.

app/Controllers/importExportController.php
```diff
@@ -38,9 +38,15 @@ public function indexAction(): void {
 	 */
 	private static function megabytes(string $size_str) {
 		switch (substr($size_str, -1)) {
-			case 'M': case 'm': return (int)$size_str;
-			case 'K': case 'k': return (int)$size_str / 1024;
-			case 'G': case 'g': return (int)$size_str * 1024;
+			case 'M':
```
Frenzie, line 41: I personally prefer indented case inside switch statements. But when I set up the rules I tried to make them as close to the existing style as possible, not to add in my own opinion. ;-)

cli/i18n/I18nData.php
```diff
@@ -329,5 +329,4 @@ public function getReferenceLanguage(): array {
 	private function getFilenamePrefix(string $key): string {
 		return preg_replace('/\..*/', '.php', $key) ?? '';
 	}
-
```
Frenzie: This strikes me as a potentially somewhat unfortunate thing to enforce.

p/api/fever.php
```diff
@@ -12,6 +11,8 @@
  * See https://github.com/dasmurphy/tinytinyrss-fever-plugin
  */
 
+declare(strict_types=1);
```
Frenzie: Meh, reduces clarity by hiding it further down.

app/Controllers/indexController.php
```diff
@@ -266,10 +266,9 @@ public static function listEntriesByContext(?int $postsPerPage = null): Traversa
 		}
 
 		foreach ($entryDAO->listWhere(
-					$type, $id, FreshRSS_Context::$state, FreshRSS_Context::$order,
-					$postsPerPage ?? FreshRSS_Context::$number, FreshRSS_Context::$offset, FreshRSS_Context::$first_id,
-					FreshRSS_Context::$search, $date_min)
-				as $entry) {
+			$type, $id, FreshRSS_Context::$state, FreshRSS_Context::$order,
+			$postsPerPage ?? FreshRSS_Context::$number, FreshRSS_Context::$offset, FreshRSS_Context::$first_id,
+			FreshRSS_Context::$search, $date_min) as $entry) {
```
Frenzie: Also clarity reduction.

Frenzie (approved):
> Any comment @Frenzie ?

I left some stray thoughts. I think some of the phpcs rules are overzealous resulting in reduced clarity but it's not a big deal either.

app/Controllers/importExportController.php
Alkarex: Yes, me too, and it is also more common to have indented cases in switches, I believe

app/FreshRSS.php
Alkarex: This one is a rule that asks to have boolean operators either at the end of the line or at the beginning, but consistently within the same statement

cli/i18n/I18nData.php
Alkarex: Ok, I will exclude a few more rules

app/FreshRSS.php
Frenzie: Fwiw, I think

bla
&& bla

is clearer than

bla &&
bla

But other people have a different opinion.

Alkarex: Yes, both are allowed, but not:

```
aa &&
bb
&& cc
```

Frenzie: But there was no mixture here.

Alkarex: Yes, there was a mixture, with trailing `||` on the first line, for instance

Frenzie: Ah see, that's the problem with putting them at the end, lol

app/Controllers/indexController.php
Alkarex: Disabling `Squiz.ControlStructures.ForEachLoopDeclaration.SpacingBeforeAs` completely would prevent checking spaces in `foreach ($list as $item) {}`, and since the problem occurs only once in the whole codebase, I came with another workaround

cli/i18n/I18nData.php
Alkarex: Fixed

p/api/fever.php
Alkarex: Rule about headers order disabled

app/Controllers/indexController.php
Frenzie: Oh, I didn't mean to insist too strongly. :-)

PR #6672, opened 2024-07-31
Fix parentheses for complex `OR` boolean search expressions

Alkarex:
Search expressions combining some non-uniform parentheses and `OR` expressions where failing.
For instance:
* `(ab) OR cd OR ef OR (gh)` was wrongly parsed as `(ab) (cd OR ef) OR (gh)`

Improved the parser, also for expressions involving negations, and add several unit tests.

### DIFF
--- a/app/Models/BooleanSearch.php
+++ b/app/Models/BooleanSearch.php
@@ -17,7 +17,6 @@ class FreshRSS_BooleanSearch {
 
 	/** @param 'AND'|'OR'|'AND NOT' $operator */
 	public function __construct(string $input, int $level = 0, string $operator = 'AND', bool $allowUserQueries = true) {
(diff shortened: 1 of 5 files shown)


Frenzie (approved):
Could be an idea to add a couple of bad testcases to ensure predictable failure behavior? (E.g., mismatched parentheses.)

Alkarex:
> Could be an idea to add a couple of bad testcases to ensure predictable failure behavior? (E.g., mismatched parentheses.)

Indeed. At the moment, the parser never fails (unfortunately) but that is clearly something to improve, by failing more visibly when there is for instance mismatched parentheses

Alkarex:
Also fixed `ab OR !(cd OR ef)`, which was parsed as `ab !(cd OR ef)`

Issue #6236, opened 2024-03-26
[BUG] Setting the filter action breaks the feed source settings


hkcomori:
**Describe the bug**
Setting a filter to mark as read unnecessary articles will break the feed source settings.

**To Reproduce**
1. Add feed https://www.hoxsin.co.jp/news/:
    - feed source type: HTML + XPATH
    - item: `//ul[@class="article-list"]/li`
    - item title: `descendant::a`
    - item contents: `.`
    - item link: `descendant::a/@href`
    - item datetime: `descendant::time`
    - custom date/time format: `Y/m/d`
    - item unique id: `descendant::a/@href`
2. At this point, the feed can be fetched right.
3. Set and save filter action:
    - `intitle:'委託者証拠金'`
4. At this point, the feed can no longer be fetched and the feed source settings are broken.

**Expected behavior**
Feed settings are remained, unnecessary articles are filtered and marked as read, and unfiltered articles can be fetched in an unread state.

**Screenshots**

**Environment information (please complete the following information):**
- OS: Windows 10 Version 22H2 (Build 19045.4046)
- Browser: Vivaldi 6.5.3206.63
- Server OS: Ubuntu 22.04
- FreshRSS version: 1.23.1
- Database version: SQLite
- PHP version: PHP 8.2.7
- Installation type: Docker

**Additional context**

math-GH:
tested on edge 1.23.2-dev: Only 1 article is fetched with this parameters, before I set the filter

![grafik](https://github.com/FreshRSS/FreshRSS/assets/1645099/006cb71e-26ce-4598-a182-8c54230702e0)


Could you please check the parameters?

hkcomori:
Sorry, my parameter was wrong.

```diff
-//ul[@class="article-list"]
+//ul[@class="article-list"]/li
```

But even with the correct parameters and `freshrss/freshrss:edge`, this issue was reproduced.

math-GH:
Thank you for the corrected config. Now it works for me too.
 Could you please explain what you mean with "Set and save filter action: intitle:'委託者証拠金'"? Where do you save it?

hkcomori:
Under "Filter actions", enter `intitle:'委託者証拠金'` into "Mark an article as read..." and "Submit".

![image](https://github.com/FreshRSS/FreshRSS/assets/11016036/47bbc2d2-06be-430a-8e9f-89eeac6ce189)

math-GH:
I can confirm.

How the error looks like:

[...]

PR #6706, opened 2024-08-12
Regex search

### LINKED ISSUE #3549: [Feature] extended keyword filtering
Hi,

is it possible to use **wildcards or some kind of regular expressions in keywords** when filtering? The problem is that I want to get all posts of a certain topic. The topic is not only defined by a single keyword, but by a list of keywords and by keywords that can be similar.

For example for getting posts about Jakob Ludwig Felix Mendelssohn Bartholdy, I want to use a filter like this
'(Jakob )(Ludwig )Felix (Mendelssohn |M. )Bartholdy'
instead of
'Jakob Ludwig Felix Mendelsohn Bartholdy' OR 'Jakob Ludwig Felix M. Bartholdy' OR 'Ludwig Felix Mendelsohn Bartholdy' OR 'Ludwig Felix M. Bartholdy' OR ... 'Felix Bartholdy'

Another question/feature request: Is it possible to **combine logical AND and OR**?
For example to get posts about a person in certain citys, I want to use a filter like
(Berlin OR D(ü|ue)sseldorf OR Frankfurt OR Leipzig) 'Felix (Mendelsohn )Bartholdy'

Kind regards,
abelbabel

### LINKED ISSUE #6750: [Bug] `inurl:` search term doesn't work with quotes
### Describe the bug

The `inurl:'term'` format always produces no results.

### To Reproduce

1. Find a single word `term` with `inurl:term` that finds matches
2. Change it to `inurl:'term'`
3. Notice it now has no matches

### Expected behavior

`inurl:term` and `inurl:'term'` should find the same thing for a single word match and `inurl:'term'` should run the search with the `'` characters removed.

### FreshRSS version

1.24.2-dev  (current `:edge` docker image)

### Environment information

- Database version: Postgres 16.4
- PHP version: 8.3.10
- Installation type: Docker
- Web server type: Apache
- Device: Linux server
- OS: (irrelevant) OpenSuse MicroOS
- Browser: Firefox 129.0.1

### Additional context

A build of the latest `edge` exhibits the same behavior

### LINKED ISSUE #6761: [Bug] quoted tag search terms don't work
### Describe the bug

Using quotes on a search term for a tag doesn't work.

### To Reproduce

1. Find a feed and term that can be searched for with a filter like `#term1`
2. Change the filter to `#'term1'`
3. Notice nothing is matched anymore

### Expected behavior

Searching for `#term1` or `#'term1'` (or `#"term1"`) should all work the same, except quoting allows the term to contain spaces.

### FreshRSS version

1.24.2-dev  (current `:edge` docker image)

### Environment information

- Database version: SQLite
- PHP version: 8.3.10
- Installation type: Docker
- Web server type: Apache
- Device: Linux server
- OS: (irrelevant) OpenSuse MicroOS
- Browser: Firefox 129.0.1

### Additional context

This is an identical bug to https://github.com/FreshRSS/FreshRSS/issues/6750, except for the tag search/filter (`#`) rather than the URL search/filter (`inurl`).

### LINKED ISSUE #6775: [Bug] Filters not working correctly when one of the words is an URL
### Describe the bug

Example article URL that contains the text that should be filtered: https://www.genbeta.com/actualidad/jornada-laboral-cuatro-dias-fracasa-japon-pais-necesita-mejorar-productividad-reduciendo-horas-trabajo

The target text is in the body, so we use free-text filter, and the aim is to filter by: 'Vía | Xataka'
The problem seems to be that in the body, part of the keywords are text, but another part is actually an URL, like this: Vía | [Xataka](https://www.xataka.com/empresas-y-economia/japon-necesitaba-acabar-su-modelo-trabajo-excesivo-esperanza-era-jornada-laboral-cuatro-dias-esta-fracasando)

`<p>Vía | <a href="https://www.xataka.com/empresas-y-economia/japon-necesitaba-acabar-su-modelo-trabajo-excesivo-esperanza-era-jornada-laboral-cuatro-dias-esta-fracasando">Xataka</a></p>`

So the filter does not give any results.

I noticed that if I search for an URL in free-text mode, it does find matches inside the article body, even if the URL is not explicitly shown as text in the body, it does search hyperlinks in the body, such as 'https://www.xataka.com'.
So I tried filtering by 'Vía | https://www.xataka.com' but it also does not work.
It also finds the keywords that contain an hyperlink correctly, but if you add any previous preceding or succeeding keyword to the search, it gives no results.

I hope you can reproduce the issue with this feed or any other feed in this circumstances.
Thank you in advance for your help.

### To Reproduce

1. Add a test feed article like the one provided, or one with 
2. Go to 'Search bar'
3. Search for 'Vía | Xataka'
4. See error, it does not filter the mentioned article that contains that text.

### Expected behavior

I expected the article to get filtered.

### FreshRSS version

1.24.2

### Environment information

- Browser: [Firefox 129]

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/3549

Support regex filters and regex search (working with PostgreSQL and MySQL and SQLite).

Works like any regular expression. Must be enclosed in `/ /` 

Currently supporting the prefix `intitle:` or no prefix at all (will add `author:` etc. soon)

Case sensitive by default, but can be made case-insensitive with `i` modifier like `/Alice/i`

Supports multiline mode with `m` modifier like `/^Alice/m`

Example of filter to keep only the entries, which title ends with `.0` from https://github.com/FreshRSS/FreshRSS/releases.atom , marking as read the entries not matching this filter:
image

image

Requires PHP 8.0+ due to use of new function `str_contains`, as planned requirement for FreshRSS 1.25.0+ https://github.com/FreshRSS/FreshRSS/pull/6711


mtalexan:
I'm running this branch in a custom docker-compose build based on the `Docker-Alpine`. I see you pretty much only have `intitle:` and the prefix-less (context) matching implemented so far so I'm focusing on testing regex features and combos of the regex feature with other pre-existing filter types.

---

I'd recommend being explicit about what type of regex patterns you support.  The most common standard types are Posix (very limited), Extended, and Perl, though there are lots of variations like Java, RE2 (Golang), etc that slightly tweak the syntax, usually primarily only of more advanced features. Maybe if you add a link in the documentation to the regex engine's definition that would be helpful as well. 

Working:
- `intitle` with regex
- `intitle` with case insensitive regex
- content with regex
- content with case insensitive regex
- `intitle` with regex and content with non-regex
- `intitle` with case insensitive regex and non-regex
- `intitle` with regex ORed with `intitle` regex
- `intitle` without regex
- content without regex
- intitle regex and tag (non-regex) pattern
- character exclusion regex patterns (i.e. `[^a-z]`)
- Wildcard match patterns (`.*` between words)
- Start of boundary patterns (`/^Israel/`)
- One-or-more wildcard match patterns (`[^a-z]+`)
- End of boundary match patterns (`routing\.$`)
- Alternatives matching, though see the issues with grouping. `messages?( routing|$)`

Not working:
- grouping in regex patterns (sort of, see below)
- multi-line mode (see below)
- (user error?) explicit `and` between patterns. i.e. `intitle:/election/i and intitle:/year/i`
- (pre-existing?) `inurl` (non-regex) can't match anything (i.e. `inurl:'p88#')
- (pre-existing?) matches are applied to the raw text, not rendered HTML. i.e. `'hit 100,000 downloads'` where the `100,000 downloads` part is a hyperlink fails to match.
- (not supported?) Look ahead patterns (`/election (?!year)/i`)

## Bugs

### Pattern Parenthesis

It's common to use parentheses in regex patterns to form conditional groups.  For example: `intitle:/messages( routing|$)/i`  to match either `messages` at the end of a line, or `messages routing` anywhere in the title.  

The parenthesis currently require manual escaping however, they're being captured by the filter logic outside the regex.  This also appears to be an existing bug in how the non-regex filters work as well, but it's much more significant for regex patterns because parentheses are used so frequently in them.

For example, this fails to match anything `intitle:/messages( routing|$)/i`, but this `intitle:/messages\( routing|$\)/i` matches two different articles, one with `messages routing` in the middle of the title, and another that ends with `messages`.  I also verified that literal parentheses aren't being matched by unescaped parentheses in the pattern, `intitle:/(Zachary/i` fails to match a title containing `(Zachary Cohen/CNN)` in it. In fact, matching a literal parenthesis requires escaping it from both the filter engine (`\(`) and the regex pattern (`[(]`), resulting in needing to specify `[\(]` and `[\)]`.

I did verify that this same parenthesis bug affects quoted non-regex patterns though too.  Doing `infile:'(Zachary'` is also failing to match the title with `(Zachary Cohen/CNN)` in it, and I have to do `infile:'\(Zachary'` to get it to match.  It's just less severe there because literal parentheses are much less common than needing to use a parenthesis in a regex pattern.


### Multi-line mode

I'm not sure how to test multi-line mode exactly.  It works fine if I put the `m` on an otherwise working case sensitive or insensitive pattern, but it doesn't add any functionality to do so.  With text like this in the middle of the content of an item:
```
Better than Signal? Looks promising.
A few bugs and UX issues but great foundation. Love that it’s public domain.
```

The pattern `/Looks promising\./` works, as does `/looks promising\./i` and I can put an `m` on the end of either and it still works (`/Looks promising\./m` or `/looks promising\./im`).   
But if I try to match across the lines: `/Looks promising\..*A few bugs and UX issues/m` it fails to match.  Or if I try to explicitly match the newlines it also doesn't work: `/Looks promising\.(\r)?\n(\r)?A few/im` (though maybe there's some escaping issues here).  Here's no way to insert a literal newline in the ad-hoc search box that I've been doing most of my testing with that I'm aware of, so I'd have to have some way to match past a line ending.  

I suspect you want `.` to match line endings as well, and probably need to set a special option for the regex engine to let `.` match those end-of-line terminators when multi-line mode is enabled.  For some reason that's frequently something regex engines require you to separately enable from multi-line mode, even though they'd really only ever expect to be used together.

Alkarex:
Hello @mtalexan and thanks for testing 👍🏻 
Quick answer already:
* there is no explicit `AND` operator; just use a space instead.
* The regex engine is not fully consistent and depends on which database you are using:
    * [PHP preg_match](https://php.net/function.preg-match) for auto-mark-as-read and auto-favourite, as well as for searching with an SQLite database
    * [For PostgreSQL search](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP)
    * [For MySQL search](https://dev.mysql.com/doc/refman/8.4/en/regexp.html#function_regexp-like)
* Searches are indeed applied to the raw HTML content
* The multiline mode allows searching for instance for a text at the start or end of a line in the raw HTML content

mtalexan:
Excellent, thanks for the clarification.  For reference, I'm testing with PostgreSQL.

- Lookahead is working, it just needed the parentheses escaped.
- Literal parentheses can be escaped with `\\(`, i.e. an escaped `\\`, but for some reason a single `\` not preceeding a `(` never needs to be escaped.
- Multi-line matching is working, it's just usually hard to find a use for since multi-line content usually has HTML or something included that's harder to match against. The `.` is non-newline-sensitive like you'd expect/want.
- Literal `/` in a pattern can successfully be escaped (`\/`)

It seems only parentheses have weird escaping rules.  Pattern escape characters like `\d` don't need the `\` part escaped. Only a `(` or `)` in the regex needs to be escaped, or a `\` before a `(` or `)` that is supposed to be in the regex pattern (i.e. `\\(` and `\\)` -> `\(` and `\)` in the pattern).  That's incredibly strange behavior because it means `\` isn't a consistent escape character like almost every other parser ever written, instead there's weird multi-character matching that must be going on instead of nested interpretation(?).  

The super strange behavior with parenthesis and escaping definitely need to be documented. No one will ever figure out how to use it without a lot of trial-and-error, and it's such a standard feature of regex patterns to use parentheses you can guarantee almost everyone will encounter the issue. I'd suggest dropping the mention in the documentation on filtering along with the links to the regex syntaxes supported by the different databases. 

---

Is there anything else you'd like me to test?  
The existing non-regex patterns still work the same, a solid sample of the range of all supported regex features work for both case-insensitive and multi-line mode, and can be used in combination with other regex and/or non-regex patterns.

It looks like the same code path is used for all cases of filters being used, so I assume there's not much point in re-testing with per-feed filters, label filters, or saved user filters.  Correct?

mtalexan:
Oh, forgot to mention I'd tested negation of a regex filter in my original set of tests.  It's working fine too.

mtalexan:
As I work on using the regexes more, I'm finding more and more cases where it's not clear if there was a regex error, or if the pattern is matching unexpected things.  For example, if I use the word-boundary escape character (`\y`), it doesn't seem to be working and the only way I can determine that is because the same top 10 things appear in my view as if I had no search filter.  

This for example doesn't work: `intitle:/\yart/i, and it causes 10 items to appear in the view that don't match anything close to what I'm looking for (the case-insensitive prefix "art", making use of the `\y` constraint escape for a word boundary).  Recognizing the top list of items are unchanged despite the whole list blanking and re-loading, and none of them match what I'm trying to search for isn't exactly easy to determine.

---

Related: it appears [constraint escapes](https://www.postgresql.org/docs/current/functions-matching.html#POSIX-CONSTRAINT-ESCAPES-TABLE) aren't working. Class-Shorthand escapes and Backreferences are working however.

Alkarex:
Thanks for the additional tests and feedback, @mtalexan 👍🏻 
Yes, the regex applied to other operators such as `author:XXX` will work identically.

Indeed, we do not have any feedback to the user when the search expression is wrong, for instance with wrong parentheses, invalid regex, etc. But we should add that.
And the whole parsing of the search query has grown from very simple rules to more complicated ones, and everything should be re-written with a proper parser. The question is how much should be done now, to avoid delaying the landing of the regex features (which we could brand as `beta` for now).

Adding an error feedback in particular should not be too difficult, but should still be left to a follow-up PR.

mtalexan:
> Thanks for the additional tests and feedback, @mtalexan 👍🏻 Yes, the regex applied to other operators such as `author:XXX` will work identically.
> 
> Indeed, we do not have any feedback to the user when the search expression is wrong, for instance with wrong parentheses, invalid regex, etc. But we should add that. And the whole parsing of the search query has grown from very simple rules to more complicated ones, and everything should be re-written with a proper parser. The question is how much should be done now, to avoid delaying the landing of the regex features (which we could brand as `beta` for now).
> 
> Adding an error feedback in particular should not be too difficult, but should still be left to a follow-up PR.

I'm definitely on board with splitting the work up. Having access to regex parsing at all is a major feature that shouldn't be held up by reduced usability.

mtalexan:
Did you mean to include the `\b` (non-printing bell character) in the regex example in filtering Readmes? `intitle:/^Lo+l\b/i`

Alkarex:
Added `author:` and `inurl:` and some documentation (excluding the parentheses aspects for now). Additional tests welcome.

Example : `author:/^Alice Doe$/im` as we split multiple authors in different lines

Alkarex:
> Did you mean to include the `\b` (non-printing bell character) in the regex example in filtering Readmes? `intitle:/^Lo+l\b/i`

I meant `\b` like in [word boundary](https://www.regular-expressions.info/wordboundaries.html). Should be tested, by the way

Alkarex:
Hum, PostgreSQL does not support `\b`
https://www.postgresql.org/docs/current/functions-matching.html#POSIX-LIMITS-COMPATIBILITY

mtalexan:
> > Did you mean to include the `\b` (non-printing bell character) in the regex example in filtering Readmes? `intitle:/^Lo+l\b/i`
> 
> I meant `\b` like in [word boundary](https://www.regular-expressions.info/wordboundaries.html). Should be tested, by the way

I thought that might be what you meant. PostgreSQL uses the TCL-style word boundaries instead (`\y`, `\Y`, `\m`, `\M`), not the Perl ones (`\b`). The PostgreSQL documentation calls them "constraint escapes", and they aren't working.

I reported the [word boundary escapes aren't working](https://github.com/FreshRSS/FreshRSS/pull/6706#issuecomment-2314227234), but I need to test the start and end of window matchers to see if they're also broken (the only other "constraint escapes" supported by PostgreSQL).

app/Models/Search.php
```diff
@@ -226,9 +266,26 @@ public function getSearch(): ?array {
...
+		return $this->not_search_regex;
+	}
+
+	/**
+	 * Filter an array of string to keep only the valid regexes
+	 * @param array<string> $regexes
+	 * @return array<string>
+	 */
+	private static function sanitizeRegexes(array $regexes): array {
+		return array_filter($regexes, static fn(string $regex) => @preg_match($regex, '') !== false);
```
Alkarex: This does not work for PostgreSQL-specific regex syntax

Alkarex:
Right, the problem comes from the way I sanitize regex expressions, which does not work for PostgreSQL-specific regex syntax:

https://github.com/FreshRSS/FreshRSS/blob/afcdea0a426b9540d945d3246a702410959382b4/app/Models/Search.php#L286-L288

Edit: PostgreSQL example for later for myself: `intitle:/^Qui\M/`

Alkarex:
I have disabled the regex sanitization for now. A question is whether it can be misused. Maybe not in a significant way (SQL ingestions should still not be possible).

Alkarex:
Another category of tests we should do relates to the proper handling of UTF-8. For instance, for the PHP regex engine, involving `\R` on multi-bytes strings (cf. https://github.com/FreshRSS/FreshRSS/issues/6236#issuecomment-2028492958). We should maybe have the `//u` modifier by default.

mtalexan:
> Another category of tests we should do relates to the proper handling of UTF-8. For instance, for the PHP regex engine, involving `\R` on multi-bytes strings (cf. [#6236 (comment)](https://github.com/FreshRSS/FreshRSS/issues/6236#issuecomment-2028492958)). We should maybe have the `//u` modifier by default.

Do you want me to run some tests on this, or are you saying it still needs to be enabled/implemented first?

Alkarex:
You are very welcome to give it a try

app/Models/Entry.php
```diff
@@ -678,11 +698,21 @@ public function matches(FreshRSS_BooleanSearch $booleanSearch): bool {
 						$ok &= stripos($this->link, $url) !== false;
 					}
 				}
+				if ($ok && $filter->getIntitleRegex()) {
```
mtalexan: I think you meant this to be `getInurlRegex`

app/Models/EntryDAO.php
```diff
@@ -960,7 +1013,7 @@ public static function sqlBooleanSearch(string $alias, FreshRSS_BooleanSearch $f
 			if ($filter->getSearch() !== null) {
```
mtalexan, line 1046: Looks like you're missing the `getNotInurlRegex` case here

app/Models/EntryDAO.php
```diff
@@ -27,6 +27,34 @@ public static function sqlIgnoreConflict(string $sql): string {
...
+	/** @param array<int|string> $values */
+	protected static function sqlRegex(string $expression, string $regex, array &$values): string {
+		$matches = static::regexToSql($regex);
+		if (isset($matches['pattern'])) {
+			$matchType = $matches['matchType'] ?? '';
+			if (!str_contains($matchType, 'i')) {
+				// Case-sensitive matching
+				$matchType .= 'c';
+			}
+			$values[] = $matches['pattern'];
```
mtalexan: Looks like you haven't implemented multi-line matching for MySQL yet. Just adding this comment to track when it's completed.

app/Models/Entry.php
```diff
@@ -636,21 +636,41 @@ public function matches(FreshRSS_BooleanSearch $booleanSearch): bool {
...
 					foreach ($filter->getNotIntitle() as $title) {
 						$ok &= stripos($this->title, $title) === false;
 					}
 				}
+				if ($ok && $filter->getNotIntitleRegex()) {
+					foreach ($filter->getNotIntitleRegex() as $title) {
+						$ok &= preg_match($title, $this->title) === 0;
+					}
+				}
 				if ($ok && $filter->getTags()) {
```
mtalexan, line 669: It looks like it would be relatively easy to add tag regex support too. I know I've had cases where I wanted to match a list of tags and it's a a lot easier to do `#/\(tag1|tag2|tag3|tag4|tag5\)/i` than it is to do `#tag1 OR #tag2 OR #tag3 OR #tag4 OR #tag5`, especially if there are some shared sub-string between each of those tags.

app/Models/EntryDAOPGSQL.php
```diff
@@ -23,6 +23,32 @@ public static function sqlIgnoreConflict(string $sql): string {
...
 	}
 
+	#[\Override]
+	protected static function sqlRegex(string $expression, string $regex, array &$values): string {
+		$matches = static::regexToSql($regex);
+		if (isset($matches['pattern'])) {
+			$matchType = $matches['matchType'] ?? '';
+			if (str_contains($matchType, 'm')) {
+				// newline-sensitive matching
+				$matches['pattern'] = '(?m)' . $matches['pattern'];
```
mtalexan, line 33: It looks like metasyntax could be specified for Postgres, but only if you don't use `/../m` and include the `m` in the set of metasyntax options manually if you need it.  As a power user it's good to know, but I'm uncertain whether or not it might be worth mentioning in documentation somewhere.

app/Models/EntryDAOSQLite.php
```diff
@@ -28,6 +28,27 @@ public static function sqlIgnoreConflict(string $sql): string {
...
+
+	#[\Override]
+	protected function registerSqlFunctions(string $sql): void {
+		if (!str_contains($sql, ' REGEXP ')) {
+			return;
+		}
+		// https://php.net/pdo.sqlitecreatefunction
+		// https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators
+		$this->pdo->sqliteCreateFunction('regexp',
+			function (string $pattern, string $text): bool {
```
mtalexan, line 45: It appears SQLite support isn't implemented yet. Just adding this comment so it's obvious when it is.

mtalexan:
Fun fact, `inurl:` is going to cause some difficulties.  Currently `inurl:` without regex doesn't work if you use quotes (https://github.com/FreshRSS/FreshRSS/issues/6750).  With regex now using the `/` as a regex "quoting" character, doing a URL match on the literal `/term/` is more difficult and existing patterns will break because `inurl:'/term/'` doesn't work, and the former literal `inurl:/term/` will now trigger a regex match instead.  Now the only way to do it is `inurl:/\/term\//`

mtalexan:
@Alkarex  I just realized something you mentioned before (emphasis mine):  

> [PHP preg_match](https://php.net/function.preg-match) for **_auto-mark-as-read and auto-favourite_**, as well as for searching with an SQLite database

This could present a pretty major problem if the regex engine and syntax for a filter run from the search box is different than the one used for auto-labeling and auto-marking-as-read. Currently the only way to test the filters for auto-labeling and auto-mark-as-read is to copy-paste it into the regular search box and see what comes up.  

I'm not familiar, is `preg_match` an option for all 3 database types and that's why it's used for auto-labeling and auto-mark-as-read rather than a database-specific syntax? Or is all the content being stored in an internal SQLite database during loading and auto-* matching before (possibly) being transferred to the alternative database? Or does searching on the SQLite database require PHP to do the filtering instead of the database, unlike other database types, and the auto-* matching occurs before the content reaches a database?

mtalexan:
If it's feasible, it seems like using the database-specific regex engine is almost always less powerful than using the PHP `preg_match`.  
Reading about it now, `preg_match` uses the PCRE library, which is literally the library for Perl-regex (but also including a few optional additions not implemented in Perl itself).  
Postgres has a modified form of the same (written from scratch), but with a more limited set of extended functions (e.g. `(?=...)`), and some modifications to a few of the escape codes (e.g. `\y` vs `\b`).  
MySQL only supports Extended-POSIX patterns with the addition of named character classes.  

Of all of them, PCRE is far and away more powerful, and avoids the non-standard Postgres syntax variations.  

---

Using the actual database engine for doing the search would appear to only really be useful for extremely massive datasets using an appropriately capable database so the engine itself can optimize the search.  To me it seems like it's more important to be able to test the auto-* matching against existing datasets than to have faster regex matching in manually entered/saved filters for huge datasets.  
If possible, I'd propose implementing regex thru the `preg_match` for all cases, then a follow up can look at providing an option to enable database-specific regex filtering (whether it's a toggle for the whole search filter, a global toggle, or a per-regex flag).

app/Models/Entry.php
Alkarex: Well spotted!

app/Models/EntryDAO.php
Alkarex: It looks like there might be a need for two different approaches for MySQL and for MariaDB.

MariaDB does not seem to support `REGEXP_LIKE` https://mariadb.com/kb/en/regexp/
but MariaDB has a new PCRE engine https://mariadb.com/kb/en/pcre/

app/Models/EntryDAOSQLite.php
Alkarex: Hum, SQLite is supposed to work already

app/Models/EntryDAO.php
Alkarex: Thanks! Fixed

Alkarex:
I would rather not embark on filtering client side (in PHP), except for SQLite which has a built-in way to do so.
Not only the performances would suffer, but the code would also be a lot more complicated with some parts to run server side and some parts to run client side

app/Models/EntryDAOSQLite.php
mtalexan: Sorry, that was from code inspection. I was looking for this block to do matching against the `im` like the others. I haven't migrated my database to check it yet though.  
This was just a comment to mark what was and wasn't included based on inspecting the code changes.

mtalexan:
> I would rather not embark on filtering client side (in PHP), except for SQLite which has a built-in way to do so. Not only the performances would suffer, but the code would also be a lot more complicated with some parts to run server side and some parts to run client side

Understandable. I only brought it up in case there was some architectural function I didn't see that would make it easy/feasible.  

---

Assuming you do the MariaDB split from MySQL, that would give 3 different regex engines: PCRE, a custom Tcl-style variant of Perl, and Extended POSIX + named character classes.  
You'll certainly want to make it big and bold in the filter document that labeling and auto-mark-read uses PHP pre-filtering regex engine while the search box uses the database regex engine and those syntaxes might not match. Putting together what you said then, the per-database(-ish) breakdown is:
- PHP (prefiltering) : PCRE
- SQLite (uses PHP): PCRE
- MariaDb : PCRE
- Postgres : custom Tcl-style variant of Perl syntax
- MySQL : Extended POSIX syntax + named character classes (no multiline support)

Notably, Extended POSIX + named character classes is a subset of the other types, and the custom Postgres format is similar to PCRE but more limited and with somw different identifiers (advanced use is often incompatible). It might be worth mentioning in the documemt that lays out the supported database types too since this could drive the database selection decision.

For me personally, it means I'll be switching my database to SQLite so I can have a consistent regex engine, but luckily there's no real harm for my use case since I only expect a few 10s of thousands of articles before purging starts kicking in. The database decision is primarily driven by that regex engine mismatch with Postgres though.

mtalexan:
To my future self:

> I'm not familiar, is preg_match an option for all 3 database types and that's why it's used for auto-labeling and auto-mark-as-read rather than a database-specific syntax? Or is all the content being stored in an internal SQLite database during loading and auto-* matching before (possibly) being transferred to the alternative database? Or does searching on the SQLite database require PHP to do the filtering instead of the database, unlike other database types, and the auto-* matching occurs before the content reaches a database?

It's using PHP to pre-filter incoming content as it's being added to the database so the associated labels and read state can be added at the same time. SQLite happens to use a PHP function you provide for it to do the filtering with when doing a database search.

Alkarex:
> luckily there's no real harm for my use case since I only expect a few 10s of thousands of articles

Note that SQLite's performance is surprisingly good, even with many articles, with or without regex

mtalexan:
@Alkarex  it appears the tag searching has the same issues as the `inurl:`, it doesn't work if quotes are included. https://github.com/FreshRSS/FreshRSS/issues/6761

Alkarex:
Tags fixed (not tested). Tests welcome

Alkarex:
And regex added for tags

app/Models/Entry.php
Alkarex: Done. Tests welcome

Alkarex: P.S. In multiline

app/Models/EntryDAO.php
Alkarex: Implemented

docs/en/users/10_filter.md
```diff
@@ -115,6 +117,31 @@ Finally, parentheses may be used to express more complex queries, with basic neg
...
+
+Supports multiline mode with `m` modifier like: `/^Alice/m`
+
+> ℹ️ `author:` is working with one author per line, so the multiline mode may advantageously be used, like: `author:/^Alice Dupont$/im`
+>
+> ℹ️ `#` is likewise working with one tag per line, so the multiline mode may advantageously be used, like: `#/^Hello World$/im`
+
+Example to search entries, which title starts with the *Lol* word, with any number of *o*: `intitle:/^Lo+l/i`
+
+As opposed to normal searches, HTML special characters are not escaped in regex searches, to allow searching HTML code, like: `/Hello <span>world<\/span>/`
```
mtalexan, line 134: This also has the effect that I have to use the HTML encoding of any character literals I'm looking for in the body as well, right? So `&amp;` and `&quot;` to find the post-rendered characters `&` and `"` for example.

Might it be worth doing this only if you have an `r` character on the regex?  Like `/Hello <span>world<\/span>/r`?

Alkarex: Indeed, the search is supposed to match the HTML content verbatim. No need to extra encode anything though, just exactly at it appears in the original HTML document.

We could consider adding some modifiers, but let's see if there is a sufficient need first. Another modifier I am considering is an automated transformation from PCRE syntax (subset) to the other variants such as PostgreSQL.

mtalexan: I think you might run into issues trying to translate between the regex versions. If you settle on the PCRE syntax there are a lot of advanced `(?...)` features that aren't present in Postgres.  Postgres also has explicit start and end word boundary markers, while PCRE has only a singular word boundary marker.  Then there's MySQL, which supports only a very small subset of regex features.  

I could see maybe trying to do something with the constraint escapes for Postgres, but that's about the only thing I can see that could possibly be translated and it would probably just confuse the situation with the databases having different regex formats.

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6785

arseru:
Hi @Alkarex, I was trying to find issues/discussions where the regex filters were mentioned, and I ended up here. I have an issue where I'm trying to assign labels automatically to incoming feed items by using intitle regex filters, but it has some erratic behaviour that it's difficult for me to identify, and I was hoping you could give me a hint of what's happening.

So I tried the same as you did in your first comment for the FreshRSS releases feed:

image

If I use the filter `intitle:/\.0$/`, I get no results:

image

Any idea why is this happening in my instance? I'm using the default [Docker deploy](https://github.com/FreshRSS/FreshRSS/tree/edge/Docker).

Alkarex:
> Any idea why is this happening in my instance? I'm using the default [Docker deploy](https://github.com/FreshRSS/FreshRSS/tree/edge/Docker).

What version of FreshRSS are you running? Try `freshrss/freshrss:edge` and make sure it was pulled recently

arseru:
> What version of FreshRSS are you running?

Version 1.24.3.

> Try `freshrss/freshrss:edge` and make sure it was pulled recently

Awesome! With this one (1.25.0-dev), it works as expected:

image

Problem solved then, thanks a lot for the quick reply :D

EDIT: duh, I just realized from your first comment that regex search is only available from FreshRSS 1.25.0+... sorry for not checking that before.

Alkarex:
@mtalexan Patch to allow parentheses in regex searches https://github.com/FreshRSS/FreshRSS/pull/6926
Tests welcome

Alkarex:
* https://github.com/FreshRSS/FreshRSS/pull/8141

PR #6709, opened 2024-08-14
Update about running tests

Alkarex:
Including Docker tests of minimum / maximum PHP versions


Alkarex:
Merging fast to avoid conflict with upcoming PR. Follow-up feedback / additional PRs still welcome

PR #6711, opened 2024-08-14
Upgrade to PHP 8.1

Alkarex:
As discussed in https://github.com/FreshRSS/FreshRSS/discussions/5474

https://www.php.net/releases/8.0/en.php
https://www.php.net/releases/8.1/en.php

Upgrade to available native type declarations
https://php.net/language.types.declarations

Upgrade to https://phpunit.de/announcements/phpunit-10.html which requires PHP 8.1+ (good timing, as version 9 was not maintained anymore)

Upgrade `:oldest` Docker dev image to oldest Alpine version supporting PHP 8.1: Alpine 3.16, which includes PHP 8.1.22.


Frenzie (approved):
Sounds good :-)

PR #6718, opened 2024-08-16
Fix HTTP cache of user queries

### LINKED ISSUE #6717: Chaching of API Queries
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/6707

<sup>Originally posted by **VincenzoKoestler** August 13, 2024</sup>
Hey everyone, I realized when using a query like https://domain/api/query.php?user=XY&t=XY=greader&nb=1500 it seems to somehow get cached or at least does not refresh automatically when new articles get added to the feeds? Is there anything I am doing wrong or some setting I am missing? 
There is no data stored in the local browser chache so it seems to be an issue from freshRSS?

Thanks for any ideas and help in resolving this issue :)</div>

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6717

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -631,11 +631,7 @@ function invalidateHttpCache(string $username = ''): bool {
 		Minz_Session::_param('touch', uTimeString());
 		$username = Minz_User::name() ?? Minz_User::INTERNAL_USER;
 	}
-	$ok = @touch(DATA_PATH . '/users/' . $username . '/' . LOG_FILENAME);
-	//if (!$ok) {
-		//TODO: Display notification error on front-end
-	//}
-	return $ok;
+	return FreshRSS_UserDAO::ctouch($username);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


lib/lib_rss.php
```diff
@@ -631,11 +631,7 @@ function invalidateHttpCache(string $username = ''): bool {
 		Minz_Session::_param('touch', uTimeString());
 		$username = Minz_User::name() ?? Minz_User::INTERNAL_USER;
 	}
-	$ok = @touch(DATA_PATH . '/users/' . $username . '/' . LOG_FILENAME);
-	//if (!$ok) {
-		//TODO: Display notification error on front-end
```
Frenzie, line 636: The error notification is taken care of with the refactor?

Frenzie (approved):

lib/lib_rss.php
Alkarex: Yes, the Boolean return value is tested by the callers when it makes sense, for instance:
https://github.com/FreshRSS/FreshRSS/blob/fe423e19ee326e4f8d87f23ec4eb10635b76ee60/app/actualize_script.php#L108-L112

PR #6719, opened 2024-08-16
Fix HTTP cache of main page

Alkarex:
Revert bug introduced by https://github.com/FreshRSS/FreshRSS/pull/3096 in which username was always null, preventing HTTP cache from working and always returning 200 instead of 304.
Discovered when checking https://github.com/FreshRSS/FreshRSS/pull/6718

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -373,7 +373,6 @@ SQL;
 	 * @return int|false affected rows
 	 */
 	public function markRead($ids, bool $is_read = true) {
-		FreshRSS_UserDAO::touch();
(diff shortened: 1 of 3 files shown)


app/Models/EntryDAO.php
```diff
@@ -373,7 +373,6 @@ protected function updateCacheUnreads(?int $catId = null, ?int $feedId = null):
 	 * @return int|false affected rows
 	 */
 	public function markRead($ids, bool $is_read = true) {
-		FreshRSS_UserDAO::touch();
```
Alkarex, line 376: Removed one redundant call to `touch()` due to recursivity

PR #6723, opened 2024-08-19
Fix HTTP cache for modified time when content is not modified

### LINKED ISSUE #6712: [Feature] bad IMS/INM values?
I read this post (and the previous posts about RSS readers) and FreshRSS gets some critique:

_Quote: "FreshRSS/1.24.0. This one latched bad IMS/INM values back in *May* and is still sending them. Hella broken."_ 

https://rachelbythebay.com/w/2024/08/02/fs/ 

I know too little about this, to know if this a serious thing that needs to be fixed (maybe it is only a specific user problem)?

None! I love FreshRSS and if this is a chance to make it better, than I am all for it.

Alkarex:
Fix case when server is sending a new modified time while the content was not actually modified.
fix https://github.com/FreshRSS/FreshRSS/issues/6712

### DIFF
--- a/lib/SimplePie/SimplePie.php
+++ b/lib/SimplePie/SimplePie.php
@@ -1728,11 +1728,17 @@ class SimplePie
 
 					$md5 = $this->cleanMd5($file->body);
 					if ($this->data['md5'] === $md5) {
+						// FreshRSS
 						if ($this->syslog_enabled)
 						{
 							syslog(LOG_DEBUG, 'SimplePie MD5 cache match for ' . SimplePie_Misc::url_remove_credentials($this->feed_url));
 						}
-						$cache->touch();
+						$this->data['headers'] = $file->headers;
+						$this->data['mtime'] = time();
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
I don't think that fixes the linked issue as such, even my hypothesis about `Retry-After` is incorrect. But no matter. this behavior is clearly better regardless. :+1: 

The feed in question had plenty of updates after all, so this particular condition would've occurred only very few times, unless it purposefully kept changing the `Last-Modified` constantly even without updates.

Alkarex:
Indeed, this PR is not related to `Retry-After`, and yes, the problematic condition is likely quite rare

Alkarex:
Will have to be re-implemented in https://github.com/FreshRSS/FreshRSS/pull/4374 , https://github.com/FreshRSS/simplepie/pull/11

Alkarex:
Help welcome to test this PR https://github.com/FreshRSS/FreshRSS/pull/4374

PR #6730, opened 2024-08-23
Fix OIDC session params definition

### LINKED ISSUE #6378: OIDC / oAuth expiry variable to prevent logging in every day
Problem:
- FreshRss doesn't seem to hang onto the authentication token for very long and requires repeat login with oAuth every day
- Most apps would allow you to specify this or at least set it longer be default
- Setting the expiry time in Authentik to 30d does not affect this
- Setting the expiration time in the FreshRss admin settings to the max value also doesn't change this
- Having to login with my SSO password every since day is not very good UX, no other app requires this

Background:
- I've been using the oAuth with Authentik for quite some time now (I helped clarify the docs on how to do this setup #5969 ). I have several other services that use oAuth. Each of these will remain logged in for days with no issue. (immich, linkwarden, seafile all use oAuth via my Authentik setup and have no problem staying logged in for days, maybe forever)
- Authentik has a setting to change the expiry in the provider screen, and I have set it to "days=30" which has not made any difference
- On the Authentik side, there is really nothing different than any other oAuth provider I have setup. Once logged in it works perfectly, it just requires re-login all the time. 
- This happens on my computer when accessing it via browser and also on mobile when accessing through browser

Technical:
- I have `freshrss/freshrss:1.23.1` running in docker
- I have persistent volumes for /data and /extensions

Solution:
- I don't think this issue is on the Authentik side, I think FreshRss is not saving the authentication token or has a setting somewhere to delete it after a certain amount of time
- some sort of expiry option in FreshRss or at least setting the default expiry to 30d would be reasonable

pando85:
Closes #6378

- standardize environment variable names
- group all in the same configuration file
- use mod_auth_openidc default values
- add documentation

1.  I tested in my cluster with my IDM configuration but apache2 configuration and changes can be easily revised.

Pull request checklist:

- [ x] clear commit messages
- [ x] code manually tested
- [ x] documentation updated


pando85:
I fixed the sanity checks in bash.

Docker/FreshRSS.Apache.conf
```diff
@@ -53,7 +57,6 @@ CustomLog "|/var/www/FreshRSS/cli/sensitive-log.sh" combined_proxy
 		OIDCXForwardedHeaders ${OIDC_X_FORWARDED_HEADERS}
 	</IfDefine>
 
-	# Can be overridden e.g. in /var/www/FreshRSS/p/i/.htaccess
```
Alkarex, line 56: Out of curiosity: isn't it working?

Alkarex (approved):

Docker/FreshRSS.Apache.conf
pando85: Ouch, nice caught! I don't know what I did sorry. Fixed in 4f697d37bfa15a1b189d4765ddf1c5e29f314680

Alkarex: If you have a chance to test that it actually works, that would be nice. I do not have an OIDC setup at the moment, to experiment

PR #6752, opened 2024-08-30
Update package-lock.json

math-GH:
made `npm update` that updated the `package-lock.json`.
Maybe it is useful to commit this update to the repository too


Frenzie:
I see a lot of potentially suspicious ESLint updates. I seem to recall there was something with that. Btw, the dependabot automatically creates PRs for updates every once in a while.

PR #6753, opened 2024-08-30
improved: 'My Labels' field be searchable

### LINKED ISSUE #6595: Opt to have the 'My Labels' field  be searchable

While adding an item as a Favourite, for example, it can be a best practice to sort entries using the "My Labels" option at the bottom of a feed item. Over time, I have created dozens of labels, some longer in words than others, nevertheless there are many, depending on topic and context. 

It is a bit frustrating and inefficient that in order to pick one or multiple labels, I have to scroll down an ever longer list. Sure, I can type the name of the label if I can remember the exact word, but more than once I have typed it incorrectly or without a 's' or with a 's' and what ends up happening is that a new label is created. I have set a system where I stick to a particular heuristic but this cannot be applied in every use case. In any case, I have spent what seems as too much time having to go back and relabel the entry before then having to go into Settings and deleting unnecessary labels, in the case I guessed or remembered incorrectly. I cannot be the only one with this issue.   

A fairly simple fix to address this would be to make the "My Labels' field searchable. Then once opened, typing the first couple of letters would bring up the relevant/matching entries to select from. Deleting the contents of the field by the user would reset the search, ready to type for the next item to select/click on, and so on. 

There is nothing else to try that I have thought of, except scrolling down to find the entry, then scrolling down some more to find the next one, then to the next one, if applicable, etc. Alternatively, there might be an option that enables an existing option for this but if so, I have clearly missed it, and if that is the case, I apologize. I did a search here and the main hit I got was regarding "searching tags," which is not what this is about. 

**Additional context**

I believe I have described the issue sufficiently, but please, let me know if more info is needed.

math-GH:
Closes #6595

- use `<datalist>` for having a suggested labels list while typing

grafik

`datalist `is state of the art: https://caniuse.com/?search=datalist

1. go to `my label` dropdown list below an article
2. type something into the text field
3. already existing labels will be listed
4. chose an item
5. text will be taken
6. click the add button/press enter
7. the chosen label will be checked

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
marked as draft to collect feedback from you.

ping @daffydock: does it solve your handling issue?

Alkarex:
Looks good for me, minus a few code syntax details 👍🏻

math-GH:
fixed. Ready for review

Issue #3380, opened 2021-01-17
[Suggestion] Improve User Queries Interface


rom-1:
1. Add a Menu Item for "Edit User Queries" in the Main Menu (#dropdown-query) next to "Bookmark current query" OR put an [edit icon](https://fontawesome.com/v4.7.0/icon/pencil) next to every user query in this dropdown menu
1. in the user queries editor: the hyperlink "filter" on the top right below the headline should be a more prominent real **button**, maybe with a text like "run", "show" or "go", or maybe it can stay "filter" if it becomes a real button. It should be placed at the bottom, next to save/cancel.
1. in the user queries editor: the "type" box should be a multi-select box. A user might want to search for "iphone" on 2 or 3 feeds at the same time.
1. the word "state" and "type" are not fitting, but I lack better words. Maybe leave "state" as it is. But "type" feels plain wrong. Even "nothing" (empty) would be better than "type", as the content of the box is quite self explaining.
1. it would be great to have some kind of syntax help link next to every input box, in which the user can input complex user queries. This means on the main page next to the search box, and in the user query editor next to the "expression" input box. Maybe put a small [question mark icon](https://fontawesome.com/v4.7.0/icon/question-circle) there, with a link to: https://freshrss.github.io/FreshRSS/en/users/03_Main_view.html --> "with the search field" (actually, there should be an extra page "query syntax" in documentation).

_Originally posted by @rom-1 in https://github.com/FreshRSS/FreshRSS/pull/3379#discussion_r559207171_

aledeg:
@rom-1 Could you tell me what theme you are using? I suspect that you are using one of the following:
- Blue Lagoon
- Screwdriver
Those themes remove the header of the drop-down menus. I guess that's the reason you've added the first item in your list. Could switch themes to test the menu entry and tell us if that's what you had in mind?

aledeg:
See #3447

rom-1:
@aledeg you are correct, I am using Blue Lagoon as a theme as I like it's minimal button style. This is why I did not have a headline, and could not see the settings button.

I switched themes, and the button in the headline is a good alternative to the idea I had in mind. Just for clarification, this is a basic mockup of what I had in mind:
![2021-02-14_14h22_09](https://user-images.githubusercontent.com/37835829/107878103-bc84ba00-6ed0-11eb-8c3d-a4c7895fdf76.png)

aledeg:
[...]

PR #6755, opened 2024-08-31
Improved: Reorg slider buttons

math-GH:
Ref. #3380 (number 2)

- added a trash icon
- moved the `remove` button upper
- the filter and statistic link looks now like a button
- this 3 buttons are now in the same row
- changed the icon for the `filter` link/button: from `mouse cursor `icon to the `eye `icon

Before:
grafik

After:
grafik

1. open the feed config
2. check the filter button
3. check the statistic button
4. check the remove/delete button

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

Todos left:
- [ ] style the trash icon button in the other themes too (`.btn-attention`)
- [ ] config category
- [ ] config user query
- [ ] config label management


math-GH:
please give feedback before I will do the left open tasks

Alkarex:
I do not really mind. One the one hand, this looks fine; on the other hand, I am used to have the *delete whatever* at the bottom of settings, also in other systems than FreshRSS

math-GH:
> I do not really mind. One the one hand, this looks fine; on the other hand, I am used to have the _delete whatever_ at the bottom of settings, also in other systems than FreshRSS

The delete botton is in this case (feed config) not at the bottom of the settings. It is below the first form box. In other cases the delete button is at the bottom of all settings (i.e. category config)

Frenzie:
I'm not a fan of any of these changes I'm afraid. :-/ Moving delete to the top especially. It's this big in your face red button (because it's dangerous; that shouldn't change) that you almost never want to click on, except in rare cases when you're actually looking for it. But also styling links like buttons is just not something I care much for in general. It depends on the specifics of course, but what's the reason for styling these links as buttons, making it look like they perform an action?

> The delete botton is in this case (feed config) not at the bottom of the settings. It is below the first form box. In other cases the delete button is at the bottom of all settings (i.e. category config)

If that's considered problematic I'd take that as an argument to move it down to maintenance.

Frenzie:
> except in rare cases when you're actually looking for it.

PS I personally want to click on it quite frequently because I add feeds from people posting issues here, but I don't believe such an atypical use case should guide design.

Issue #6223, opened 2024-03-21
remove simple.phtml and use only `layout.phtml`


math-GH:
it could be good to remove simple.phtml and use only `layout.phtml` if possible

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/pull/6221#discussion_r1533503350_

PR #6765, opened 2024-09-02
Proof: simple.phtml is not used anymore

math-GH:
Ref #6223

This PR is just a proof that simple.phtml is not necessary anymore

- `simple.phtml` references deleted
- `simple.phtml` content deleted for better testing
- title has a text that gives a text that simple.phtml was referenced here

1. have a user, that has to validate his/her e-mail (activate email validation + create a new user. This user tries to login. Do not click on the validation link in the email)


math-GH:
This draft PR should help to understand if and where the simple.phtml is still needed.

(sharebyemail extension uses simple.phtml too but without any consequences. Try this extension too)

Alkarex:
Thanks for the investigation 👍🏻  Let's go ahead and remove it, then, right?

Frenzie:
What's used instead? Or put another way, this is definitely not a bug and can safely be removed? :-)

JamBalaya56562:
Thanks for the investigation here. I opened #8998 with a production-ready version on current `edge`: it removes the remaining `simple` layout references, deletes `app/layout/simple.phtml`, and relies on the shared `layout.phtml` plus the existing email-validation route guard.

I think this draft can be closed in favor of #8998 if maintainers are happy with that PR.

PR #6767, opened 2024-09-03
Use curl to fetch extensions list

### LINKED ISSUE #6744: [Bug] Unable to load extension list over IPv6
### Describe the bug

When attempting to view extensions in the GUI, FreshRSS attempts to [load available extensions](https://github.com/FreshRSS/FreshRSS/blob/9b2bb50e410607fc67ecafbdaf85ffc4830723e2/app/Controllers/extensionController.php#L45):

```
protected function getAvailableExtensionList(): array {
	$extensionListUrl = 'https://raw.githubusercontent.com/FreshRSS/Extensions/master/extensions.json';
	$json = @file_get_contents($extensionListUrl);
	...
```

I have not tested this across other installation types, but in my environment, this always fails if https://raw.githubusercontent.com resolves to an IPv6 address. Logging into the same Kubernetes pod, and installing some additional tools, I was able to verify that I can reach that address over IPv6 (ping succeeds, curl is able to download the file over IPv6), and also that it is the `file_get_contents` call that fails.

Disabling IPv6 globally, or overriding DNS resolution for that particular address to force it onto IPv4 A records only is a viable workaround.

### To Reproduce

1. Go to _Configuration -> Extensions_
2. Wait
3. Connection times out

### Expected behavior

Extensions settings page loads.

### FreshRSS version

1.23.1

### Environment information

- Database version: PostgreSQL 14.1.0
- PHP version: PHP 8.2.7
- Installation type: Kubernetes, based on official Docker image version 1.23.1
- Web server type: Apache

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6744

### DIFF
--- a/app/Controllers/extensionController.php
+++ b/app/Controllers/extensionController.php
@@ -44,10 +44,10 @@ class FreshRSS_extension_Controller extends FreshRSS_ActionController {
 	 */
 	protected function getAvailableExtensionList(): array {
 		$extensionListUrl = 'https://raw.githubusercontent.com/FreshRSS/Extensions/master/extensions.json';
-		$json = @file_get_contents($extensionListUrl);
+		$json = httpGet($extensionListUrl, CACHE_PATH . '/extension_list.json', 'json');
 
 		// we ran into problems, simply ignore them
-		if ($json === false) {
+		if ($json === '') {
 			Minz_Log::error('Could not fetch available extension from GitHub');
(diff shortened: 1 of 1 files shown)


Alkarex:
Related improvement https://github.com/FreshRSS/FreshRSS/pull/6768

PR #6770, opened 2024-09-03
improved: Default styles for pre and code 

### LINKED ISSUE #4689: default styles for <pre> and <code>
Hi,

in Origine there is some styling for `<pre>` and `<code>`, when it is placed in an article
https://github.com/FreshRSS/FreshRSS/blob/8f31827e035686f46546a9c01fe5d48b3eed8a81/p/themes/Origine/origine.css#L801-L823

I would suggest to set a default style to all themes (and not define it just in Origine).

Which style (light or dark background? Mono spaced font?) would you expect?

math-GH:
Closes #4689

nothing spectacular. Just a bit housekeeping. 

- move some essential CSS lines for `<pre>` and `<code>` into `frss.css`

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -838,19 +838,14 @@ kbd {
 }
 
 .content pre {
-	margin: 10px auto;
-	padding: 10px 20px;
-	overflow: auto;
 	background: var(--background-color-dark);
 	color: var(--font-color-light);
-	font-size: 0.9rem;
 	border-radius: 3px;
 }
 
 .content code {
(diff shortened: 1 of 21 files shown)


Frenzie:
Sorry, a bit late, but I tend to prefer `whitespace: pre-wrap` for `<pre>`. Perhaps a thought for another time if people agree. :-)

Alkarex:
@Frenzie I do not really mind. Maybe we should collect a few real-life examples to decide what makes most sense

Frenzie:
Real life examples galore, anything longer that would cause horizontal scrolling. Whether you prefer that or mind that is a personal preference I suppose.

PR #6774, opened 2024-09-04
Fix XPath for HTML documents with broken root

### LINKED ISSUE #6773: [Bug] Retrieve content from a specific website
### Describe the bug

For some time now, no article content has been retrieved from the website [TV-Spielfilm](https://www.tvspielfilm.de/tv-programm/rss/tipps.xml), error message `The selector didn’t match anything. As a fallback the original feed text will be displayed instead.`. I also get the same error message in the [Demo](https://demo.freshrss.org/i).

### To Reproduce

Feed URL: https://www.tvspielfilm.de/tv-programm/rss/tipps.xml
Type: RSS/Atom
Advanced:
Article CSS Selector: .broadcast-detail

### Expected behavior

content from main page should be replace the feed content.

### FreshRSS version

1.24.3-dev

### Environment information

- PHP version: 7.4.33
- Installation type: git
- Web server type: Apache
- OS: Debian Linux
- Browser: Vivaldi

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6773

The default `.//` prefix for the XPath does not work for HTML documents, which have content after the end of their main node.
But `//` works in both cases.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -785,7 +785,7 @@ HTML;
 			$content = '';
 			$cssSelector = htmlspecialchars_decode($feed->pathEntries(), ENT_QUOTES);
 			$cssSelector = trim($cssSelector, ', ');
-			$nodes = $xpath->query((new Gt\CssXPath\Translator($cssSelector))->asXPath());
+			$nodes = $xpath->query((new Gt\CssXPath\Translator($cssSelector, '//'))->asXPath());
 			if ($nodes != false) {
 				$path_entries_filter = $feed->attributeString('path_entries_filter') ?? '';
 				$path_entries_filter = trim($path_entries_filter, ', ');


Alkarex:
Example of test:

```php
<?php
$htmlString = <<<HTML
<!DOCTYPE html>
<html>
	<body>
	</body>
</html>
<div>div1</div>
<div>div2</div>
HTML;

$html = new DOMDocument();
$html->loadHTML($htmlString);
$xpath = new DOMXPath($html);

$divs = $xpath->query('.//div');
foreach ($divs as $div) {
	echo 'Matched with .// ' . $div->nodeValue . PHP_EOL;
}

$divs = $xpath->query('//div');
foreach ($divs as $div) {
	echo 'Matched with // ' . $div->nodeValue . PHP_EOL;
}
```

Returns:

```
Matched with // div1
Matched with // div2
```

Issue #6776, opened 2024-09-05
UI support SeaMonkey


Alkarex:
### Describe the bug

Our current CSS for the *origine* (default) theme works mostly in [SeaMonkey](https://www.seamonkey-project.org), which uses an older Gecko engine, except for a wrong overlay of the date:

![image](https://github.com/user-attachments/assets/b4507232-ec3f-469c-99c5-ef27b86b749b)

It would be nice to fix this minor bug at least for the default theme.

Likewise, I have spotted some minor JavaScript glitches (edit: https://github.com/FreshRSS/FreshRSS/pull/6777)

Footnote: Found when testing https://github.com/FreshRSS/FreshRSS/discussions/6159#discussioncomment-10561267

polybjorn:
The overlap comes from the `:has(~.date)` selectors in `frss.css` (lines 1507, 1585, 1603, and a media-query variant at 2552) that reserve space for the date column. SeaMonkey's Gecko is too old for `:has()` so the rules never match and the date sits on top of the title.

Quickest fix is to render a `has-date` class on `.flux_header` like the existing `has-thumbnail` / `has-summary` and key those rules on the class instead. Tested on my own instance, works in both SeaMonkey and LibreWolf, ~11 lines plus the regenerated rtl.

A bigger change might be worth considering though. Converting the absolute-positioning layout for `.titleAuthorSummaryDate` to CSS Grid would handle the missing-date case naturally, so no sibling-state queries are needed at all (`:has` or class). The `min-width: calc(100% - 155px)` click-target hack on the title and the `:has(.multiline)` rule at 1516 also become unnecessary. Same compatibility win, cleaner end state.

One small thing that's separable from either change. The `.multiline` class on `.title` only exists so the parent can query for it via `:has`. Could just be moved to the `<li>` directly.

Any reason not to go the grid route?

Frenzie:
It wasn't supported (over) a decade ago when the UI was mostly written. I don't see any reason not to use it where it makes sense.

PR #6777, opened 2024-09-05
JavaScript form validation compatibility older browsers

Alkarex:
Restore compatibility with older browsers (e.g. Firefox and Chrome older than 2020) and simplify code at the same time
https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter
Contributes to https://github.com/FreshRSS/FreshRSS/issues/6776
Was introduced by https://github.com/FreshRSS/FreshRSS/pull/4370 (to be retested a bit more, though)


Alkarex:
If you can @math-GH , let me know if you can think of possible regressions there

math-GH:
> If you can @math-GH , let me know if you can think of possible regressions there

I am not deep enough to see the impacts on older browsers.

PR #6797, opened 2024-09-11
Fix unsafe login

### LINKED ISSUE #6796: [Bug] Auth with unsecure URL applies htmlspecialchars() escape before username/password comparison
### Describe the bug

Hello.

I tried the "unsecure" method to login : https://subdomain.domain.fr/frss/p/i/?c=auth&a=login&u=USER&p=PASSWORD

It works perfectly if password is without special char inside like "Toto1234" but changing it to "Toto&1234" will not work anywork and gives a "invalid ID" error message.

How could I use this method with a more secure password ?

Thanks

### To Reproduce

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

### Expected behavior

Use a password with special char inside.
Tested with "&".

### FreshRSS version

1.24.3

### Environment information

PHP 8.2
MariaDB 10.11
Apache 2.4.59

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6796

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -187,8 +187,8 @@ class FreshRSS_auth_Controller extends FreshRSS_ActionController {
 				Minz_Request::forward(['c' => 'auth', 'a' => 'login'], false);
 			}
 		} elseif (FreshRSS_Context::systemConf()->unsafe_autologin_enabled) {
-			$username = Minz_Request::paramString('u');
-			$password = Minz_Request::paramString('p');
+			$username = Minz_Request::paramString('u', specialchars: true);
+			$password = Minz_Request::paramString('p', specialchars: true);
 			Minz_Request::_param('p');
 
 			if ($username === '') {


app/Controllers/authController.php
```diff
@@ -187,8 +187,8 @@ public function formLoginAction(): void {
 				Minz_Request::forward(['c' => 'auth', 'a' => 'login'], false);
 			}
 		} elseif (FreshRSS_Context::systemConf()->unsafe_autologin_enabled) {
-			$username = Minz_Request::paramString('u');
-			$password = Minz_Request::paramString('p');
+			$username = Minz_Request::paramString('u', specialchars: true);
+			$password = Minz_Request::paramString('p', specialchars: true);
```
Frenzie, line 191: This works, but it's confusing.

Elsewhere it says something intuitive like:

https://github.com/FreshRSS/FreshRSS/blob/31c8846791d4b5316fbc790202f79545c012f9c2/lib/Minz/Request.php#L78

But the guilty function reverses that:

https://github.com/FreshRSS/FreshRSS/blob/31c8846791d4b5316fbc790202f79545c012f9c2/lib/Minz/Request.php#L114

Alkarex: I am not sure to understand what you mean, @Frenzie .
And `paramArrayString()` is not used here

Frenzie: paramArrayString and all the other old functions say

```
if specialchars 
  return htmlspecialchars($result)
else
  result $result
```

This new addition `paramStringNull` says:
```
if specialchars
  return $result
else
  return htmlspecialchars($result)
```

I find this very unexpected, and it's the reason I didn't see where the problem came from at a glance.

Alkarex: Right, I see what you mean, and that is indeed unfortunate, and should be made uniform.
The original function is the following one:

https://github.com/FreshRSS/FreshRSS/blob/e5320759ebe5bd820e2f815265ecd289f899842f/lib/Minz/Request.php#L45-L58

Alkarex: But let's address than in another PR

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/6800

PR #6800, opened 2024-09-11
Fix inversed encoding logic in paramArray

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6797#discussion_r1754661634
Also fix the possibility to use `<'&">` in shortcuts, and some minor encoding bugs in user queries

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -40,7 +40,7 @@ class Minz_Request {
 	 * Read the URL parameter
 	 * @param string $key Key name
 	 * @param mixed $default default value, if no parameter is given
-	 * @param bool $specialchars special characters
+	 * @param bool $specialchars `true` to return special characters, `false` (default) to XML-encode them
 	 * @return mixed value of the parameter
 	 * @deprecated use typed versions instead
 	 */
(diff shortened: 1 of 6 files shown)


lib/Minz/Request.php
```diff
@@ -61,15 +61,21 @@ public static function hasParam(string $key): bool {
...
+	/**
+	 * @param bool $specialchars `true` to return special characters, `false` (default) to XML-encode them
+	 * @return array<string|int,string|array<string,string|int|bool>>
+	 */
 	public static function paramArray(string $key, bool $specialchars = false): array {
 		if (empty(self::$params[$key]) || !is_array(self::$params[$key])) {
 			return [];
 		}
-		return $specialchars ? Minz_Helper::htmlspecialchars_utf8(self::$params[$key]) : self::$params[$key];
+		return $specialchars ? self::$params[$key] : Minz_Helper::htmlspecialchars_utf8(self::$params[$key]);
```
Alkarex, line 72: That is the main change. The rest are fixes and documentation

Frenzie: To my mind this is exactly backwards with how it's named. :-/

Frenzie: By which I don't mean I'd be happy with `nospecialchars` because that would still be unexpected to read.

Imo the original is perfect and logical and this reversal is just asking for exactly the trouble that we saw.

Frenzie: Perhaps `encodespecialchars` would disambiguate sufficiently (because clearly you have a different view) while reading properly and easily.

Alkarex: This PR only fixes an anomaly in the new functions `paramArray*()` compared with the others. In other words, to align with what we have had for a long time. See e.g. this function back from version 1.0:

https://github.com/FreshRSS/FreshRSS/blob/4c99fd689f0778cbf6a987471ebf72b3787739c2/lib/Minz/Request.php#L30-L41

Now I understand the possible confusion, which is why I have added some documentation. Changing the name of the parameter would be an option, but I would rather not change the default value.

Frenzie: Gosh, I had even misread the original from a decade ago because of this problem. :-/

I find `if specialchars DO NOT specialchars` utterly confusing.

Frenzie: So yes, I'd definitely change the name to something else. Something that conveys a meaning like `leavespecialchars`, although I'm not 100% happy with that suggestion.

Frenzie: retain/preserve/keep are probably all better

Frenzie: `ignore`?

Alkarex: For another PR, but examples of parameter name could be: `plaintext, unsafeHtml, unescaped...`
(P.S. sorry, got interrupted, and this comment was for 20 minutes ago, before some of yours)

Alkarex: At the moment, `specialchars` is to be understood as *give me the special characters* :-)

Alkarex: Merging, but let's continue in another PR if we want to change the parameter name

Frenzie: > At the moment, `specialchars` is to be understood as _give me the special characters_ :-)

I understand that, but I also think anyone who's known `htmlspecialchars()` for 25 years (or even just a couple) would be likely to misread it as "apply htmlspecialchars()". :-)

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/6809

PR #6809, opened 2024-09-14
Rename param specialchars to plaintext

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6800#discussion_r1756435762

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -187,8 +187,8 @@ class FreshRSS_auth_Controller extends FreshRSS_ActionController {
 				Minz_Request::forward(['c' => 'auth', 'a' => 'login'], false);
 			}
 		} elseif (FreshRSS_Context::systemConf()->unsafe_autologin_enabled) {
-			$username = Minz_Request::paramString('u', specialchars: true);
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Thanks! I vaguely wonder if plaintext might be associated with ASCII, which would imply a different kind of escaping, but perhaps that's old fashioned.

<https://en.wikipedia.org/wiki/Plain_text>
> In principle, plain text can be in any [encoding](https://en.wikipedia.org/wiki/Character_encoding), but occasionally the term is taken to imply [ASCII](https://en.wikipedia.org/wiki/ASCII). As [Unicode](https://en.wikipedia.org/wiki/Unicode)-based encodings such as [UTF-8](https://en.wikipedia.org/wiki/UTF-8) and [UTF-16](https://en.wikipedia.org/wiki/UTF-16) become more common, that usage may be shrinking.

I nevertheless retain a minor preference for something like unescaped or raw, but either way it addresses the confusion.

Alkarex:
I would also have better like something related to escaping, but I try to avoid negations (which would be: false for escaped, true for unescaped)

Alkarex:
Related https://github.com/FreshRSS/FreshRSS/pull/6821

PR #6816, opened 2024-09-19
Improved: Subscription management page

math-GH:
Before
grafik

After:
grafik

1. feeds list has less padding to the left hand side. It is aligned to the manage category button 
2. 2+ line feed names are aligned to the favicon
3. same like 2
4. feeds with error have a triangle warning symbol now (similar to the normal view feed navigation bar)
5. muted feeds has now its sign  (similar to the normal view feed navigation bar)
6. `Add an RSS feed` -> `Add a feed` (because FreshRSS supports more than "just" RSS feeds)
7. a bit more space between the icon on the button's text
8. mouse title `manage` added (similar to feed manage button)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
There seems to be a regression in the main view, with repeated icons:
image

math-GH:
fixed

Frenzie:
Looks nice.

> feeds with error have a triangle warning symbol now (similar to the normal view feed navigation bar)

As in maybe something else needs to be cleaned up or as in it wasn't there?

Alkarex:
I think the feeds, which are disabled AND with error, should have the disabled icon shown, just like in the main view.
I do not really mind if the error icon is also shown for those feeds or not.

math-GH:
Improved. The mute icon is now overruling the warning icon. Additionally the mouse hover title shows now the information what is the issue with this feed.

grafik

math-GH:
> > feeds with error have a triangle warning symbol now (similar to the normal view feed navigation bar)
> 
> As in maybe something else needs to be cleaned up or as in it wasn't there?

I do not understand it 🙈

Frenzie:
> I do not understand it 🙈

If it's purely additive I didn't say anything, but if it replaces another icon that other icon might need to be removed. :-)

math-GH:
the triangle warning icon and muted icon are now added. No icon was shown before.
It is in sync now to the icons in the normal view feed sidebar.

app/views/subscription/index.phtml
```diff
@@ -59,29 +59,33 @@
 							$empty_title = _t('sub.feed.empty');
 						}
 						$mute_class = $feed->mute() ? ' mute' : '';
+
+						$title = $error_title !== '' ? '⚠ ' . $error_title . '&#13;' : '';
```
Alkarex, line 63:
```suggestion
						$title = $error_title !== '' ? '⚠ ' . $error_title . "\n" : '';
```

Doesn't that work? (not tested) I would prefer, rather than an entity

math-GH: it does not work (Firefox on Windows):
grafik

Alkarex: Please make sure to try with `"\n"` (double quotes) or `PHP_EOL` and not single quotes.
Seems to work for me.

math-GH: ok. I got it. Thanks for clarification.

Yes it works, because it adds a real line break into the `title` attribute. It means that the HTML tag will have a line break inside of its tag.
IMHO I am not a big fan of having breaks in HTML attributes. And it messes a bit the PHP code with single quotes and double quotes characters.
grafik

Currently, the code is similar to:
https://github.com/FreshRSS/FreshRSS/blob/20c1d06e6258b211fb208fc885ad4e2d6e4a4094/app/layout/aside_feed.phtml#L129-L132
(the line breaks are a bit on different place, because there is an additional first line in `aside_feed.phtml` that is not there in subscription management page.)

While searching for a line break in `title` attribute Stackoverflow recommends mostly `'&#13;'`

I do not see a benefit of using `"\n"` instead of `'&#13;'`

Alkarex: It does not matter much, so it can stay as you prefer. But what about `PHP_EOL` ?
I try to avoid entities and favour native characters whenever possible, which I find typically easier to read, lighter, and less error-prone.

Alkarex:
Ah, one last consistency thing: the behaviour in the subscription page looks good now, but in the main view, a muted (disabled) feed that is also in error has the error icon instead of the muted one.

app/views/subscription/index.phtml
Alkarex, line 63:
```suggestion
						$title = $error_title === '' ? '' : "⚠ {$error_title}\n";
```

Actually, this looks like a good situation for having double-quotes

math-GH:
> Ah, one last consistency thing: the behaviour in the subscription page looks good now, but in the main view, a muted (disabled) feed that is also in error has the error icon instead of the muted one.

fixed.

Alkarex:
Regression https://github.com/FreshRSS/FreshRSS/issues/6850

PR #6817, opened 2024-09-20
Skip toggle password button on keyboard navigation

timendum:
Set `tabindex` of the  toggle password button in the login form to "-1", so that when the user do keyboard navigation (i.e. use tab to move between fields) skip the toggle button.

In the end, the user will be presented:

- username
- password
- keep me logged
- login button (but can login any time with enter)

[Documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) suggest to use any value different than 0 or -1.

- A small change in the usability of the login form

1. Go to the login form
2. Use tab to move between form element

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/auth/formLogin.phtml
+++ b/app/views/auth/formLogin.phtml
@@ -25,7 +25,7 @@
 			<label for="passwordPlain"><?= _t('gen.auth.password') ?></label>
 			<div class="stick">
 				<input type="password" id="passwordPlain" required="required" />
-				<button type="button" class="btn toggle-password" data-toggle="passwordPlain"><?= _i('key') ?></button>
+				<button type="button" class="btn toggle-password" data-toggle="passwordPlain" tabindex="-1"><?= _i('key') ?></button>
 			</div>
 			<input type="hidden" id="challenge" name="challenge" />
 			<noscript><strong><?= _t('gen.js.should_be_activated') ?></strong></noscript>


Alkarex:
Hello,
I believe it is useful to keep the toggle button tabbable. But we could discuss the order of tabindex.
What do you think @Frenzie , @math-GH ?

@timendum Could you please put a few more words on why the current flow is problematic for you?

Frenzie:
Indeed, I wouldn't be inclined to do this. While presumably the primary use case is mobile with a less accurate touch keyboard, simply following the visual order as is makes for the principle of least surprise.

timendum:
The "toggle password" button break the flow for me, I just type user and password (or do auto-fill with a password manager), decide if keep me logged or not, and that's it.

The "let me see the password" is really a corner case, rare.  

And also not very a good experience in my opinion, for every "tab" pressed the cursor goes one row "down" but not on "toggle password". This is not expected nor very clear for me.

A tabindex like:
- 0 = user
- 1 = password
- 2 = "remember me"
- 3 = login butto
- 4 = show password

could be ok in my opinion, but Mozilla documentation discourages it.

Maybe a new row with a dedicated checkbox and label for the "show password" would be a better UX?

Out of curiosity, I did a quick survey:

- Facebook (and Instagram) do like FreshRSS, "show password" before login button, inside password field
- Google as a dedicated row for "show password"
- Reddit and Gitlab have no "show password"

math-GH:
It is on purpose that the button is tab-able.
The button should not break the login flow: username - tab - password - enter -> logged in. It is noting special, isn't it?
If you want to check the "keep me logged in" checkbox just double tap (you are using the tab already, just do it twice ;) ).
(side note: we could add a config option that this checkbox is checked/unchecked by default)
In case of using a password manager: it should be possible to config the actions. Mostly it should do automatically the `enter` after adding the password (and it does not care about the show password button). If your password manager is doing the checkbox checking too, than add one more tab to the flow.

And one more option is possible:
If you do not like the show password button at all: you can hide the buttons via the customCSS extension

IMHO: I do not see any need to change the current tab-flow

Alkarex:
Sorry that we cannot accept your PR this time @timendum 🫤

PR #6751, opened 2024-08-30
Improved: refactor the sharing menu to use a template instead of duplicated HTML code

math-GH:
Ref. https://github.com/FreshRSS/FreshRSS/pull/6395#issuecomment-2084595602

Changes proposed in this pull request:

- use `<template>` instead of duplicated HTML dropdown menu
- use `<template>` instead of `<script>` (feed aside navigation bar: label menu, feed menu)
- views: normal view, global view (reader view does not have the sharing button)

How to test the feature manually:

1. use the sharing button in the feed header (do not forget to activate the button)
2. use the sharing button in the feed footer
3. check different sharing services PLUS mail, print, clipboard

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

PR #6825, opened 2024-09-22
Fix:  (#6815) share menu shortcut

### LINKED ISSUE #6815: [Bug] Shortcut "s" for sharing does not work
### Describe the bug

In former times, I was able to share articles to my Wallabag installation, using the hotkey "s"-

### To Reproduce

Go to any article and press "s".

### Expected behavior

Open the configured sharing service.

### FreshRSS version

1.25.0-dev 

### Environment information

- Database version: MariaDB 10.6.18
- PHP version: 8.1.29
- Installation type: Git
-Web server type: Apache
- Device: Desktop
- OS: Arch Linux
- Browser: Firefox

### Additional context

_No response_

math-GH:
Closes #6815
Regression #6751

- refactoring: add `function show_share_menu(el)` so that the new menu template is called with the shortcut too.
- improved: scroll the share button into the view if out of view
-

1. go to normal view and press the share shortcut (default: `s`)
2. it opens the `first`/`topmost` sharing menu of the article

check too:
- share menu icon is enabled/disabled in the top line (config page: display)
- share menu icon is enabled/disabled in the bottom line (config page: display)
- scroll the share button out of view (above, below): it scrolls the button into the view
- keep the share button in the view: no scrolling

Known edge case bug: If the top line sharing icon is disabled and the bottom line sharing icon is enabled and the share button is visible on the bottom view edge than the sharing menu opens but it does not scroll into the view (reason: the scrolling feature checks if the share button is in the view not the share menu.). 

In general we should discuss if it is better to open the menus in the bottom line to the top instead to the bottom.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -738,9 +738,13 @@ function auto_share(key) {
 		// Display the share div
 		location.hash = share.id;
 		// Force scrolling to the share div
-		const scrollTop = needsScroll(share.closest('.bottom'));
+		const scrollTop = needsScroll(share.closest('.horizontal-list'));
 		if (scrollTop !== 0) {
-			document.scrollingElement.scrollTop = scrollTop;
+			if (share.closest('.horizontal-list.flux_header')) {
+				share.nextElementSibling.nextElementSibling.scrollIntoView({ behavior: "smooth", block: "start" });
+			} else {
+				share.nextElementSibling.nextElementSibling.scrollIntoView({ behavior: "smooth", block: "end" });
+			}
 		}
 		// Force the key value if there is only one action, so we can trigger it automatically
 		if (shares.length === 1) {
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -698,29 +698,31 @@ function user_filter(key) {
 }
 
 function show_share_menu(el) {
-	const itemId = el.closest('.flux').id;
-	const templateId = 'share_article_template';
-	const id = itemId;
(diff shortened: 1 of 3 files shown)


p/scripts/main.js
```diff
@@ -697,19 +697,54 @@ function user_filter(key) {
...
+	const titleText = title;
+	const div = el.parentElement;
+	const dropdownMenu = div.querySelector('.dropdown-menu');
+	const template = document.getElementById(templateId).innerHTML
+		.replace(/--entryId--/g, id)
+		.replace(/--link--/g, link)
+		.replace(/--titleText--/g, titleText)
+		.replace(/--websiteName--/g, websiteName)
+		.replace(/--articleAuthors--/g, articleAuthorsText);
+	if (!dropdownMenu) {
```
Alkarex, line 704: Could this test be located at the top of the function, to avoid doing a lot of wasted work?

Alkarex: I was imagining an early `return` statement to avoid a big `if`, but it is also fine like this

math-GH: Done.

Then a refactoring is needed here too:
https://github.com/FreshRSS/FreshRSS/blob/24d10feebd0c9802b1082e88bf165902885591aa/p/scripts/main.js#L914-L936

Should I do it with this PR or a new PR?

math-GH: ok. You merged it already :D

PR #6827, opened 2024-09-23
Docker entrypoint optimize find sed

Alkarex:
Single `find` and single `sed` command instead of 3


Frenzie (approved):
I'd actually advertise it as being easier to read. The presumably submillisecond optimization hardly matters in this context. ^_^

PR #6868, opened 2024-10-05
Support: new form for feed problems

math-GH:
This form could help us for better feed issue investigation.
Related: https://github.com/FreshRSS/FreshRSS/discussions/6287

grafik
grafik

### DIFF
--- a/.github/ISSUE_TEMPLATE/feed-report-form.yml
+++ b/.github/ISSUE_TEMPLATE/feed-report-form.yml
@@ -12,7 +12,7 @@ body:
     attributes:
       label: Feed URL
       description: Share the URL of the feed
-      placeholder: https://…
+      placeholder: https://...
     validations:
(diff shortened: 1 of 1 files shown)


Alkarex:
Good idea. An OPML export of the feed settings could also be interesting in some cases, especially for Web scraping

math-GH:
> Good idea. An OPML export of the feed settings could also be interesting in some cases, especially for Web scraping

I would not cover the Web Scraping feature yet. Currently that are rare cases. I do not want to over engineer the issue forms, just having a better form than the Bug Report form that does not fit to the feed problems.

The Web Scrapping feature is often a support thing and rarely an issue (and so it should be a discussion thread instead of an issue ticket).

If we feel that we need a Web Scrapping form in the future than we can create it there too :) But let's keep it lean(er) now.

PR #6881, opened 2024-10-10
Apache protect more non-public folders and files

Alkarex:
https://github.com/FreshRSS/FreshRSS/discussions/6879#discussioncomment-10906338

### DIFF
--- a/.devcontainer/.htaccess
+++ b/.devcontainer/.htaccess
@@ -1,11 +1,1 @@
-# Apache 2.2
-<IfModule !mod_authz_core.c>
-	Order	Allow,Deny
-	Deny	from all
-	Satisfy	all
-</IfModule>
-
-# Apache 2.4
-<IfModule mod_authz_core.c>
-	Require all denied
-</IfModule>
+Require all denied
--- a/.devcontainer/.htaccess
+++ b/.devcontainer/.htaccess
@@ -1,0 +1,11 @@
+# Apache 2.2
+<IfModule !mod_authz_core.c>
+	Order	Allow,Deny
+	Deny	from all
+	Satisfy	all
+</IfModule>
+
+# Apache 2.4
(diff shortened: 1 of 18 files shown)


Frenzie (approved):

Alkarex:
@Frenzie I have added another rule for the root https://github.com/FreshRSS/FreshRSS/pull/6881/commits/087668e042d69ae78804b7fb8083f50d304ced0d

Frenzie:
Could that interfere with updates? (Or risk losing customizations.) It might be better as a sample file akin to the constants.

Alkarex:
> Could that interfere with updates? (Or risk losing customizations.)

What scenario could it be, do you believe?

> It might be better as a sample file akin to the constants.

I would rather have something that is a bit more secure out of the box, if possible

Frenzie:
> What scenario could it be, do you believe?

For example a mod_rewrite on shared hosting as per the issue that prompted this PR, HTTP authentication, custom mimetypes, enabling compression, or dozens of other things neither of us can think of off the cuff. You don't want such customizations to get in the way of an upgrade or risk of being overwritten.

Alkarex:
Right, so a case when users have a custom `./FreshRSS/.htaccess` ?

That is a good point.

By the way, the place for a custom `.htaccess` related to HTTP authentication is `./FreshRSS/p/i/.htaccess`

Frenzie:
> Right, so a case when users have a custom ./FreshRSS/.htaccess ?

Yes, a similar problem theoretically exists anywhere but only for the root does it strike me as potentially problematic.

Frenzie (approved):

Alkarex:
OIDC regression fixed in https://github.com/FreshRSS/FreshRSS/pull/6893

Alkarex:
Reworked in https://github.com/FreshRSS/FreshRSS/pull/7008

PR #6886, opened 2024-10-11
Add move to next unread Label on mark as read.

### LINKED ISSUE #6867: Allow Mark as read to move to the next peer on Labels
### Describe the bug

Unlike categories and feeds, marking everything in a Label as read isn't moving to the next peer Label.

### To Reproduce

1. Go to the first label with unread articles in it
2. Have a second label following it with its own unique unread articles in it
3. Scroll to the bottom of the article list and click the "mark all as read" checkbox
4. Notice the view doesn't move to the next label with unread articles

### Expected behavior

The mark all as read moves to the next peer label just like how it moves to the next unread feed when marking a feed as read, or to the next category when marking a category as read.

### FreshRSS version

1.25.0-dev (current edge)

### Environment information

- Database version: SQLite
- PHP version: 8.3.10
- Installation type: Docker (using official Alpine Dockerfile from edge)
-Web server type: Apache
- Device: Android Pixel Fold
- OS: OpenSUSE MicroOS Server
- Browser: Firefox 131.0

### Additional context

This is the same when using the Mark as Read button from the top bar.  

Maybe there's some worry about knowing what the next peer label is with unread articles in it since the same articles can appear in multiple labels and if you justarked those as read in the current label, the next peer label with unread articles can change after the "mark as read"?

mtalexan:
The Labels, unlike the Feeds and Categories, don't move to the next unread when "move to next unread on mark all as read" user feature is enabled.
Labels are more complex than Feeds and Categories because Entries can be in more than Label at a time. So when marking all Entries in the Label as read, it can cause other Labels to end up with all their Entries marked as read as well. The calculation of what the next Label/Feed/Category is to jump to normally happens when generating the link for the "Mark as Read" buttons, but it can't for Labels.

To address the problem for Labels, use a placeholder value during the pre-calculation of the "Mark as Read" button link. When that placeholder value is encountered during the "Mark as Read" action, the next Label with unread Entries will be calculated immediately after the mark as read action has been processed.

Fix all the translations of the 'jump_next' text to remove the '(feed or categories' part that no longer applies.
Attempt to fix the inconsistent Russian, Italian, and Polish translations of 'jump_next' text, which phrased the '(feed or categories)' part differently.

---- 

Open questions:
- Should all the translations of `jump_next` be changed to the equivalent of `"jump to next"` instead of `"jump to next sibling"`?
- Are the Italian, Polish, and Russian translation changes of `jump_next` correct?
- Is using the default undefined `nextGet=a` acceptable or should some other value be used?
- When there are no other labels unread should it return to the Main Feed (currently implemented) or should it stay on the current label with everything already read, or something else?
- Is there a better approach for this? It's questionable MVC design to put this logic in the Controller, but the Model currently has no way to know and supply the answer until after the Controller has performed its actions.
- Are there unit tests that can/should be written?

----

Closes #6867 

- When "move to next unread on mark as read" is enabled, marking a Label as read will move to the next unread label after the current one
- If there are no other Labels, it returns to the Main Feed
- Supports later labels indirectly being marked as read and skipped during the "next unread"
- Move to tags after the current as the first preference, but wrap back to the start of the list if necessary.
- Remove the '(feed or category)' part of the `jump_next` string from all translations
- (?) Correct inconsistent Italian `jump_next` translation (I don't speak this language)
- (?) Correct inconsistent Polish `jump_next` translation (I dont speak this language)
- (?) Correct inconsistent Russian `jump_next` translation (I don't speak this language)

Technical:

- When calculating the `nextGet` value for the 'mark as read' links of Labels, use the "unknown" value (`nextGet=a`) to indicate it should be calculated after the `markAsRead` action.
- Find the next unread "tag" (label) after performing the `markAsRead` on tags if `nextGet=a`
- The Italian, Polish, and Russian translations of the `jump_next` didn't use the `(feed or category)` text at the end that could be easily identified and removed, but instead said `"jump to next feed or category"`.
  - Italian and Polish I used Google Translate to try and change these to `"jump to next sibling"` but this needs to be checked
  - Russian apparently doesn't have a generic word for "sibling" (according to Google Translate), so I just left it as `"jump to next"`

1. Do not enable the "move to next unread on mark as read" user feature
2. Select a label with unread entries in it
3. Verify the URL for the "mark as read" button and link at the end of the list both have `nextGet=t_...` (pointing back to the current tag/label ID)
4. Click the button
5. Verify it returns to the current tag/label with nothing unread left
- [x] Case: Current behavior for disabled "move to next unread on mark as read" preserved

7. Enable the "move to next unread on mark as read" user feature
8. Select a label with unread entries in it
9. Verify the URL for the "mark as read" button and link at the end of the list both have 'nextGet=a`
10. Click the button
11. Verify it moves to the next tag/label with unread entries
- [x] Case: Simple move to next unread label after current

13. Select the only label with any unread entries in it
14. Click the Mark as Read button
15. Verify it returns to the Main Feed
- [x] Case: Nothing else unread

17. Select a label where the only unread entry is also the only unread entry in the label immediately after it
18. Click the Mark as Read button 
19. Verify it moves to the next label that still has something unread in it
- [x] Case: Multiple labels change unread state at the same time

21. Select a label with no unread labels after it, but unread labels before it
22. Click the Mark as Read button
23. Verify it moves to the first unread label in the overall list (wraps)
- [x] Case: Wrap the mark as read check

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] (N/A) unit tests written (optional if too hard)
- [x] (N/A) documentation updated


Alkarex:
Looks like a good start 👍🏻 
Let us know if you need any help to finalise the PR.
Try:
```sh
make fix-all
make test-all
```

mtalexan:
@Alkarex It sounds like you're OK with the way things are in terms of these questions I had, correct?
> - Is using the default undefined nextGet=a acceptable or should some other value be used?
> - When there are no other labels unread should it return to the Main Feed (currently implemented) or should it stay on the current label with everything already read, or something else?
> - Is there a better approach for this? It's questionable MVC design to put this logic in the Controller, but the Model currently has no way to know and supply the answer until after the Controller has performed its actions.

app/Controllers/entryController.php
```diff
@@ -102,6 +102,41 @@ public function readAction(): void {
 						break;
 					case 't':
 						$entryDAO->markReadTag($get, $id_max, FreshRSS_Context::$search, FreshRSS_Context::$state, $is_read);
+						//marking all entries in a tag as read can result in other tags also having all entries marked as read,
+						//so the next unread tag calculation is deferred by passing next_get = 'a' instead of the current get ID.
+						if ($next_get === 'a' && $is_read) {
+							$tagDAO = FreshRSS_Factory::createTagDao();
+							$found_tag = 0;
+							foreach ($tagDAO->listTags() ?: [] as $tag) {
```
Alkarex: This adds an SQL query (times 2, a bit below), so let's see if we can find a cheaper way to achieve the same

Alkarex:
> Is using the default undefined nextGet=a acceptable or should some other value be used?

I need to set me into this part of the code a bit more to answer. Let's come back to details optimisations a bit later.

> When there are no other labels unread should it return to the Main Feed (currently implemented) or should it stay on the current label with everything already read, or something else?

I do not have a strong feeling about it, but maybe showing the view with all labels could be one option.

> Is there a better approach for this? It's questionable MVC design to put this logic in the Controller, but the Model currently has no way to know and supply the answer until after the Controller has performed its actions.

In particular, I would like to avoid the (double) additional SQL query if possible (at least keep the results in a variable), so let's look at the options. Ideas welcome. I cannot dig into it right now.

app/Controllers/entryController.php
mtalexan: I saw it did that, but couldn't come up with an alternative. Effectively we need the list of tags with unread articles in it, starting with the current tag and warpping back to the beginning. The current tag won't show up in the list of unread since we just marked it all as read, so I had to get the entire list just to figure out the label order, then query each label to find which has unread.

Maybe if we had a way to get a 2-dimemsional array of tags and their unread counts I could do it with a single query, but I don't see any way to achieve the order-based search for the next unread than the looping thru the list to find a matching ID, and then continuing the (wrapping) loop looking for the next unread.

mtalexan:
> > Is using the default undefined nextGet=a acceptable or should some other value be used?
> 
> I need to set me into this part of the code a bit more to answer. Let's come back to details optimisations a bit later.

The basic mechanic is for each mark as read button on the screen to query for what it's `nextGet` should be when it's generating the link URL in the JavaScript. That `nextGet` is usually where we detect if we're on a feed or category and determine what the next should be if we mark the current as read. It's largely implementing focus movement logic so it makes sense there. Labels however can have multiple simultaneously marked as read if the only unread entries are also the only unread in another Label. So we can't really precalculate the next unread label without duplicating the markAsRead database action. I opted instead to have it defer the determination until after the button was pressed and the current entries had actually been marked as read.

app/Controllers/entryController.php
Alkarex: Ok. Please at least save `$tagDAO->listTags()` in a variable to avoid the second query

mtalexan:
- Changed the fallback case if there's no Labels with unread entries so it falls back to the overall Label list (`nextGet=T`).
- Only query the `tagDAO->listTags()` once and use it in each foreach-loop.
- Fixed the very broken Makefile `fix*` and `test*` targets
  - Dependencies were pretty messed up for the `vendor/bin/php*`. Commands and were symlinking a non-existent `../vendor/bin/php*` to `bin/php*` as part of a `vendor/bin/php*` target (!?)
  - The "`vendor/bin/php*`" targets were all separately calling the `bin/composer install ...` even though a single call generated them all.
  - `vendor/bin/php*` commands were being used instead of the `bin/php*` that were symlinked.
  - The `bin/composer` dependency was missing from some targets

Also verified the `make fix-all` and `make test-all` pass now.

mtalexan:
Oops forgot to check whether we're still in the fallback case before setting `nextGet=T`

Alkarex:
Would you mind moving the Makefile changes to another PR?

mtalexan:
Moved makefile changes to https://github.com/FreshRSS/FreshRSS/pull/6913.
I rebased out the Makefile change commit into the other MR, and rebased to drop it here.  I don't know of a good way to do the same in a merge.

app/Controllers/entryController.php
mtalexan: Done.

app/i18n/ru/conf.php
```diff
@@ -220,7 +220,7 @@
 		),
 		'hide_read_feeds' => 'Скрывать категории и ленты без непрочитанных статей (не работает с «Показывать все статьи»)',
 		'img_with_lazyload' => 'Использовать режим «ленивой загрузки» для загрузки картинок',
-		'jump_next' => 'перейти к следующей ленте или категории',
+		'jump_next' => 'перейти к следующей	',
```
Alkarex:
```suggestion
		'jump_next' => 'перейти к следующей',
```

app/Controllers/entryController.php
```diff
@@ -102,6 +102,45 @@ public function readAction(): void {
...
+									}
+									// Give up if reached our first tag again
+									if ($tag->id() == $get) {
+										break;
+									}
+								}
+							}
+							// if we still haven't found any unread tags, fallback to the full tag list
+							if ($next_get === 'a') {
+								$next_get = "T";
```
Alkarex:
```suggestion
								$next_get = 'T';
```

Alkarex:
Thanks 👍🏻 
Please add a line for you https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #6896, opened 2024-10-15
Fix PHPCS ControlSignature

Alkarex:
Regression from https://github.com/FreshRSS/FreshRSS/pull/6666
We were not checking anymore for whitespace rules in e.g. `if (true) {`


Alkarex:
Found due to https://github.com/FreshRSS/FreshRSS/pull/6886 (needed there)

PR #6913, opened 2024-10-17
Fix the Makefile for test and fix targets.

### LINKED ISSUE #6912: [Bug] `make fix-all` and `make test-all` are broken
### Describe the bug

Running `make fix-all` or `make test-all` in a VSCode devcontainer tries to symlink in php files from outside the repo that don't exist, and is missing the `bin/composer` dependency on some of the targets that call it.

### To Reproduce

1. Open the `edge` branch
2. Enter the devcontainer
3. Run either `make fix-all` or `make test-all`
4. See the build errors

### Expected behavior

It completes the build successfully

### FreshRSS version

1.25.1-dev (current edge)

### Environment information

- Database version: N/A
- PHP version: N/A
- Installation type: N/A
- Web server type: N/A
- Device: N/A
- OS: N/A
- Browser: N/A

### Additional context

_No response_

mtalexan:
The bin/composer dependency was missing from half the targets that needed it.
The vendor/bin/* targets are all created by a single call to composer, but were all duplicating the composer call.
The php* tools defined a vendor/bin/php* target, but then proceeded to create the bin/php* target instead by symlinking from an incorrect path. php* tools were referenced as vendor/bin/php* even though they were symlinked to bin/php*.

Closes #6912 

- Fixed the very broken Makefile fix* and test* targets
  - Dependencies were pretty messed up for the vendor/bin/php*. Commands and were symlinking a non-existent ../vendor/bin/php* to bin/php* as part of a vendor/bin/php* target (!?)
  - The "vendor/bin/php*" targets were all separately calling the bin/composer install ... even though a single call generated them all.
  - vendor/bin/php* commands were being used instead of the bin/php* that were symlinked.
  - The bin/composer dependency was missing from some targets

1. Open a vscode devcontainer
2. run the Makefile `test` and `fix` targets
3. Confirm they work

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] (N/A) unit tests written (optional if too hard)
- [x] (N/A) documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,32 +59,29 @@ stop: ## Stop FreshRSS container if any
 ## Tests and linter ##
 ######################
 .PHONY: test
-test: vendor/bin/phpunit ## Run the test suite
-	$(PHP) vendor/bin/phpunit --bootstrap ./tests/bootstrap.php ./tests
+test: bin/phpunit ## Run the test suite
+	$(PHP) bin/phpunit --bootstrap ./tests/bootstrap.php ./tests
 
 .PHONY: lint
-lint: vendor/bin/phpcs ## Run the linter on the PHP files
-	$(PHP) vendor/bin/phpcs . -p -s
+lint: bin/phpcs ## Run the linter on the PHP files
+	$(PHP) bin/phpcs . -p -s
 
 .PHONY: lint-fix
-lint-fix: vendor/bin/phpcbf ## Fix the errors detected by the linter
-	$(PHP) vendor/bin/phpcbf . -p -s
+lint-fix: bin/phpcbf ## Fix the errors detected by the linter
+	$(PHP) bin/phpcbf . -p -s
 
 bin/composer:
 	mkdir -p bin/
 	wget 'https://raw.githubusercontent.com/composer/getcomposer.org/1a26c0dcb361332cb504e4861ed0f758281575aa/web/installer' -O - -q | php -- --quiet --install-dir='./bin/' --filename='composer'
 
-vendor/bin/phpunit: bin/composer
+# building any of these builds them all
+vendor/bin/phpunit vendor/bin/phpcs vendor/bin/phpcbf vendor/bin/phpstan &: bin/composer
 	bin/composer install --prefer-dist --no-progress
-	ln -s ../vendor/bin/phpunit bin/phpunit
 
-vendor/bin/phpcs: bin/composer
-	bin/composer install --prefer-dist --no-progress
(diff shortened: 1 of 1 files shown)


Alkarex:
Just to understand the original issue better: did you hit anything that was not working?

I have just tried again from a clean Dev Container (e.g. on GitHub codespaces, but same effect locally), and I could execute `make test-all` without any problem

image

Alkarex:
Ah sorry, you tried `test` and `fix` as opposed to the `test-all` and `fix-all`.

`test` and `fix` are older, with some legacy scenarios, and should indeed be cleaned. They need to be able to run locally with no PHP installed and by relying on a Docker image, so that should be tested with this PR.

mtalexan:
I opened the DevContainer completely fresh and ran `make fix-all` and it failed for me. 

The first error was that there was no `bin/composer` when calling `bin/composer run-script fix`. That was a pretty clear the `composer-fix` target was being called, and is quite obviously missing a dependency on the `bin/composer` target.
I went thru all the places `bin/composer` was being called and made sure they had a dependency listed for it.

After the `make fix-all` was working, I ran the `make test-all`. It errored out trying create a symlink from `../vendor/bin/phpcbf` to `bin/` right after it had successfully run thru a composer install (based on the output it was printing). Sure enough, it's running from the root of the repo clone, so `..` is a directory outside the repo and doesn't have anything of the sort. I assumed since the `bin/composer install` had just been called on the prior line of the makefile, it probably wanted `vendor/bin/phpcbf` that had just been installed, not something else, so I fixed the reference. 
Assuming it might be a repeated problem, I looked to see if `../vendor` was used anywhere else, and found that all the PHP tools were doing it the same way and fixed them all. Since they were all also basically identical targets except for what was being linked, I corrected how the `bin/composer install` was called so it wasn't re-called for every tool (the first call creates all of them, that's just how it works), and combined the targets in the Makefile so it's easy to add new ones.

mtalexan:
I could keep going on this a bit if necessary. It looks like there's still one target in the file that refers to something in `..` (so it won't exist), and a bunch of the target's should really be marked `.PHONY` since they are a command to run and don't have a an actual file with the given name.

Alkarex:
> I opened the DevContainer completely fresh and ran `make fix-all` and it failed for me.

Ah indeed, error if starting with `make fix-all` without doing a `make test-all` first.

> I could keep going on this a bit if necessary.

Yes, that would be most welcome.

Alkarex (approved):

Alkarex:
Let's merge this PR, which is already an improvement

PR #6926, opened 2024-10-21
Allow regex parentheses

Alkarex:
While waiting for a new better search parser, auto-escape parentheses in regex expressions to allow them like in `/^(ab|cd)/`, including escaped parentheses such as `/^(ab|cd) [(] \) (ef|gh)/`

### DIFF
--- a/app/Models/BooleanSearch.php
+++ b/app/Models/BooleanSearch.php
@@ -37,6 +37,7 @@ class FreshRSS_BooleanSearch {
 		if ($level === 0) {
 			$input = $this->parseUserQueryNames($input, $allowUserQueries);
 			$input = $this->parseUserQueryIds($input, $allowUserQueries);
+			$input = self::escapeRegexParentheses($input);
 			$input = trim($input);
 		}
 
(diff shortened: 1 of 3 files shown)


Alkarex:
Ping @mtalexan

mtalexan:
I'll give it a try in the next couple days. I'll need to do a bunch of regex pattern conversion, but if it works for one case it will probably work for all.

mtalexan:
Switched to this branch and converted all my patterns yesterday, and everything still seems to be labeling and marking as read properly today. Looks good to me.

Alkarex:
Excellent 🙂
Thanks for the feedback

mtalexan:
Dang I may have spoken too soon.
It looks like this broke User Searches when used as part of a per-feed "mark as read".

I have a User Search named "Sports" that includes a lot of regex patterns, including many with parentheses in them. 
I have some feeds that include `search:'Sports'` in their mark-as-read pattern.

This was working before, and things about Sports we're automatically being marked as read as soon as they came in for those feeds.

Now I see those sports entries showing up and not being marked as read.  I can run a manual `search:'Sports'` and see it's picking up the articles I expected to be marked as read, so I know the User Search is correct, and I can see the syntax of the mark-as-read in my feed settings is very simple (just the one command), but the items still aren't being marked as read.

I'm guessing the escaping of the parentheses in the filtering is happening before the `search:` items are dynamically resolved to their patterns, so the parentheses in the `search:` patterns aren't being escaped?

Alkarex:
@mtalexan Good catch! Please try https://github.com/FreshRSS/FreshRSS/pull/6936

PR #6931, opened 2024-10-22
Web export  SQLite

Alkarex:
https://github.com/FreshRSS/FreshRSS/discussions/6930

Simple case of a system using SQLite as database:
image

Case of a system using e.g. PostgreSQL but making regular exports as SQLite:
image

Note that there is not yet a Web interface to initiate a DB export to SQLite (like is possible in CLI)

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -31,6 +31,22 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 	public function indexAction(): void {
 		$this->view->feeds = $this->feedDAO->listFeeds();
 		FreshRSS_View::prependTitle(_t('sub.import_export.title') . ' · ');
+		$this->listSqliteArchives();
+	}
+
+	private function listSqliteArchives(): void {
+		$this->view->sqliteArchives = [];
+		$files = glob(USERS_PATH . '/' . Minz_User::name() . '/*.sqlite') ?: [];
+		foreach ($files as $file) {
+			$archive = [
+				'name' => basename($file),
+				'size' => filesize($file),
+				'mtime' => filemtime($file),
+			];
+			if ($archive['name'] != '' && $archive['size'] != false && $archive['mtime'] != false) {
+				$this->view->sqliteArchives[] = $archive;
+			}
+		}
 	}
 
 	private static function megabytes(string $size_str): float|int|string {
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -34,21 +34,6 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 		$this->listSqliteArchives();
 	}
 
-	private function listSqliteArchives(): void {
-		$this->view->sqliteArchives = [];
-		$files = glob(USERS_PATH . '/' . Minz_User::name() . '/*.sqlite') ?: [];
-		foreach ($files as $file) {
-			$archive = [
-				'name' => basename($file),
-				'size' => filesize($file),
-				'mtime' => filemtime($file),
-			];
-			if ($archive['name'] != '' && $archive['size'] != false && $archive['mtime'] != false) {
-				$this->view->sqliteArchives[] = $archive;
-			}
-		}
-	}
-
 	private static function megabytes(string $size_str): float|int|string {
 		switch (substr($size_str, -1)) {
 			case 'M':
@@ -709,5 +694,40 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 			default:
 				return 'application/octet-stream';
 		}
+	}
+
+	private const REGEX_SQLITE_FILENAME = '/^(?![.-])[0-9a-zA-Z_.@ #&()~\-]{1,128}\.sqlite$/';
+
+	private function listSqliteArchives(): void {
+		$this->view->sqliteArchives = [];
+		$files = glob(USERS_PATH . '/' . Minz_User::name() . '/*.sqlite') ?: [];
+		foreach ($files as $file) {
+			$archive = [
+				'name' => basename($file),
+				'size' => @filesize($file),
+				'mtime' => @filemtime($file),
+			];
+			if ($archive['size'] != false && $archive['mtime'] != false && preg_match(self::REGEX_SQLITE_FILENAME, $archive['name'])) {
+				$this->view->sqliteArchives[] = $archive;
+			}
+		}
+	}
+
+	public function sqliteAction(): void {
+		if (!Minz_Request::isPost()) {
+			Minz_Request::forward(['c' => 'importExport', 'a' => 'index'], true);
+		}
+		$sqlite = Minz_Request::paramString('sqlite');
+		if (!preg_match(self::REGEX_SQLITE_FILENAME, $sqlite)) {
+			Minz_Error::error(404);
+			return;
+		}
+		$path = USERS_PATH . '/' . Minz_User::name() . '/' . $sqlite;
+		if (!file_exists($path) || @filesize($path) == false) {
+			Minz_Error::error(404);
+			return;
+		}
+		$this->view->sqlitePath = $path;
+		$this->view->_layout(null);
 	}
 }
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -700,7 +700,7 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 
 	private function listSqliteArchives(): void {
 		$this->view->sqliteArchives = [];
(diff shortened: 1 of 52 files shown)


Frenzie:
Wow, that was fast. I'd barely even read that issue yet. :smile:

rezad1393:
thank you for the fast pull.
I hope it becomes a great addition to this great app.

Alkarex:
Let's merge this one already, while waiting for another PR to add a mechanism to produce SQLite exports from the Web interface, but this will surely require several configuration options and care regarding disk usage etc.

rezad1393:
is the backup need to be run as a user having access to the host cli?
because in the new release I dont see this anywhere but in changelog.

Alkarex:
@rezad1393 For setups using SQLite, there is nothing to do and it just works.
For systems using other databases, the SQLite export must be done by CLI (so far).

rezad1393:
> @rezad1393 For setups using SQLite, there is nothing to do and it just works. For systems using other databases, the SQLite export must be done by CLI (so far).

I use the sqlite (and switch to git setup based on your suggestion) but I dont see this GUI mock-up anywhere.

Alkarex:
Did you try in the export/import section? http://localhost/i/?c=importExport

rezad1393:
oh.
I was seasrching for it in setting and admin part. my bad.
when I try it I get this error
```


Secure Connection Failed

An error occurred during a connection to myhost:1081.

    The page you are trying to view cannot be shown because the authenticity of the received data could not be verified.
```

I use a self signed https connection to my LAN hosted freshrss instance.

Alkarex:
That error is unrelated to FreshRSS. You of course need to change `http://localhost/` to whatever address you are using for your FreshRSS

rezad1393:
> That error is unrelated to FreshRSS. You of course need to change `http://localhost/` to whatever address you are using for your FreshRSS

I use a LAN client for hosting freshrss and I know its ip and name. and use a self signed ssl cert for it.
I use that for address and use it daily.

my mistake was not specifying my issue more clearly.
I can reach the import/export page.
the issue is that when I try to use it and click download, I get that https error.

Alkarex:
Could you check what you have in your FreshRSS logs and in your Web server logs?
And check also the exact URL leading to this error in your Web browser networking tools.

rezad1393:
```
2025/01/01 18:15:58 [error] 661#661: *16065 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 155099136 bytes) in /mnt/data/freshrss/app/views/importExport/sqlite.phtml on line 9" while reading response header from upstream, client: 192.168.1.101, server: odroid, request: "POST /i/?c=importExport&a=sqlite HTTP/2.0", upstream: "fastcgi://unix:/run/php-fpm/php-fpm.sock:", host: "odroid:1081"
2025/01/01 18:15:58 [error] 661#661: *16065 upstream prematurely closed FastCGI request while reading upstream, client: 192.168.1.101, server: odroid, request: "POST /i/?c=importExport&a=sqlite HTTP/2.0", upstream: "fastcgi://unix:/run/php-fpm/php-fpm.sock:", host: "odroid:1081"

```

rezad1393:
ok I had to increase the memory_limit in php.ini to be bigger than sqlite size.

is this an issue of php that needs a memory limit for a file download?
is this the correct behavior?

if I had 1GB sqlite and wanted to download it php needs a 1GB memory limit?

rezad1393:
btw can a date be added to exported file so that it is not a generic db.sqlite ?

Alkarex:
> is this an issue of php that needs a memory limit for a file download?

This might depend on your setup. It is not needed with Apache and PHP as module as recommended (and used by our Docker image, for instance).

rezad1393:
> > is this an issue of php that needs a memory limit for a file download?
> 
> This might depend on your setup. It is not needed with Apache and PHP as module as recommended (and used by our Docker image, for instance).

weird.
I use nginx.

Alkarex:
Edit: Wrong

Alkarex:
@rezad1393 Just to be sure, could you please try to add a line `ob_end_clean();` in https://github.com/FreshRSS/FreshRSS/blob/7a5ce0fe20d63746b7f5b435ba5eef0a9b7db2d6/app/views/importExport/sqlite.phtml like below, and tell whether it makes any difference?

```php
<?php
declare(strict_types=1);
ob_end_clean(); // Added
header('Content-Type: application/vnd.sqlite3');
header('Content-Disposition: attachment; filename="' . basename($this->sqlitePath) . '"');
header('Cache-Control: private, no-store, max-age=0');
header('Last-Modified: ' . gmdate(DateTimeInterface::RFC7231, @filemtime($this->sqlitePath) ?: 0));
header('Content-Length: ' . (@filesize($this->sqlitePath) ?: 0));
readfile($this->sqlitePath);
```

Alkarex:
Buffering issues should be fixed for more setups with https://github.com/FreshRSS/FreshRSS/pull/7230

rezad1393:
> Buffering issues should be fixed for more setups with #7230

so after that merge I can lower back the php memory limit to previous value ?

Alkarex:
> > Buffering issues should be fixed for more setups with #7230
> 
> so after that merge I can lower back the php memory limit to previous value ?

Yes. Please try with `freshrss/freshrss:edge` or `freshrss/freshrss:edge-alpine` (new version just pushed)

rezad1393:
> > > Buffering issues should be fixed for more setups with #7230
> > 
> > 
> > so after that merge I can lower back the php memory limit to previous value ?
> 
> Yes. Please try with `freshrss/freshrss:edge` or `freshrss/freshrss:edge-alpine` (new version just pushed)

if it is ok I wait for next release.

Alkarex:
> if it is ok I wait for next release.

`freshrss/freshrss:edge` *is* our continuous release and is ready. If you refer to the *slow* release channel `freshrss/freshrss`, it is too late for feedback or for fixing anything at that time

rezad1393:
I usually used the zip version that is released (like FreshRSS 1.25.0 [Latest](https://github.com/FreshRSS/FreshRSS/releases/latest)) 
but now I use git with version release (same version but with git commands)

is the edge stable? 
usually software developers don't recommend non-numbered release mechanisms.

"continuous release" sounds too much like rolling release linuxes (archlinux) to be stable.

Frenzie:
It's stable enough to test if this particular feature works the way you expect. :-)

Of course it's not to be randomly updated every day if you just want things to keep working as they are.

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/7330#issuecomment-2665729063

> > How stable "edge" is?

> *edge* is our rolling release, stable enough for many production setups. Fast releases, meaning that in case of regressions (rare), they get priority to be rapidly fixed.
> To be compared with the slow releases (*latest*), which are less frequent, a bit more tested, but without fix in case of bug/regression until the next slow release.

Importantly, the best time to provide feedback is during the PR, second best is when it is merged in *edge*, and too late after a *latest* release.

rezad1393:
> [#7330 (comment)](https://github.com/FreshRSS/FreshRSS/issues/7330#issuecomment-2665729063)
> 
> > > How stable "edge" is?
> 
> > _edge_ is our rolling release, stable enough for many production setups. Fast releases, meaning that in case of regressions (rare), they get priority to be rapidly fixed.
> > To be compared with the slow releases (_latest_), which are less frequent, a bit more tested, but without fix in case of bug/regression until the next slow release.
> 
> Importantly, the best time to provide feedback is during the PR, second best is when it is merged in _edge_, and too late after a _latest_ release.

I read that edge thing and I see that it is different from my own understanding of edge in other projects.
usually edge meant main or master which was latest git commit.
but when I look at freshrss github page the edge IS the master.
which means edge is not beta it is alpha release at best.
I ask because I am afraid of data loss.

Alkarex:
> which means edge is not beta it is alpha release at best.

This is not the definition of alpha. And the development is done in PRs, not in *edge*.

> I ask because I am afraid of data loss.

This is to be addressed with backups anyway.

The key part is that if you - the person who is asking a feature or fix - are not contributing to testing the changes, it is unlikely that there will be any difference between PR → edge → latest.

rezad1393:
> > which means edge is not beta it is alpha release at best.
> 
> This is not the definition of alpha. And the development is done in PRs, not in _edge_.
> 
> > I ask because I am afraid of data loss.
> 
> This is to be addressed with backups anyway.
> 
> The key part is that if you - the person who is asking a feature or fix - are not contributing to testing the changes, it is unlikely that there will be any difference between PR → edge → latest.

I have tested with edge (git checkout) and even reported a bug here https://github.com/FreshRSS/FreshRSS/pull/6931#issuecomment-2567011743 .
I just don't use edge for daily usage.

PR #6934, opened 2024-10-22
Remove new window target in normal view

tydavis:
As of v1.24.3, clicking the title of an article in normal view opens a
new tab in the browser. Remove the link reference to expand the article
within the reader, while still allowing for direct external view via
the "See on original website" link at the end of the row.

### DIFF
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -56,7 +56,7 @@
 	?></li><?php
 	endif; ?>
 	<li class="item titleAuthorSummaryDate">
-		<a target="_blank" rel="noreferrer" href="<?= $this->entry->link() ?>" class="item-element title<?= (($topline_thumbnail !== 'none') || $topline_summary) ? ' multiline' : '' ?>" dir="auto"><?= $this->entry->title() ?><?php
+		<?= $this->entry->title() ?><?php
 		if ($topline_display_authors):
 			?><span class="author"><?php
 			$authors = $this->entry->authors();


Alkarex:
> As of v1.24.3, clicking the title of an article in normal view opens a new tab in the browser

That is not the expected behaviour. This could be a sign of a FreshRSS extension, which is misbehaving. Please try again after disabling your FreshRSS extensions.

tydavis:
@Alkarex This is present without extensions (vanilla install). The `target="_blank"` anchor is present in the code currently: https://github.com/FreshRSS/FreshRSS/blob/edge/app/views/helpers/index/normal/entry_header.phtml#L59

My guess is that this was duplicated from the "See on website" anchor on L90.

EDIT: for clarity, when the target link is removed, the application behaves as expected ("Clicking the article title in the Normal view expands the article in Normal view") instead of viewing the article on the original website.

Alkarex:
The `<a target="_blank"` is meant for systems without JavaScript and looks correct. I cannot reproduce the problem you are facing.
Can you provide some details about your Web browser and FreshRSS setup? Any JavaScript error in the console? Could you try with another browser without any Web browser extension? Would you be able to post a screen recording here to make sure we talk about the same thing?

tydavis:
Looks like my "Stop The Madness" plugin (Safari, also Firefox & Chrome) is forcing HTML5 behavior to be honored -- per https://html.spec.whatwg.org/#navigable-target-names  `target="_blank"` always opens a new window. 

Why does the list entry need to contain a link when, without it, the application performs exactly as expected?

Attached are shots of the demo site and my edited local site.
Screenshot 2024-10-23 at 10 30 43
Screenshot 2024-10-23 at 10 29 20

Alkarex:
> Why does the list entry need to contain a link when, without it, the application performs exactly as expected?

Because our Web UI also works to a good extent without JavaScript at all, and the non-JavaScript behaviour is to open links in a new tab. I will try something else in another PR

Alkarex:
Could you try https://github.com/FreshRSS/FreshRSS/pull/6937 ?

app/views/helpers/index/normal/entry_header.phtml
```diff
@@ -56,7 +56,7 @@
 	?></li><?php
 	endif; ?>
 	<li class="item titleAuthorSummaryDate">
-		<a target="_blank" rel="noreferrer" href="<?= $this->entry->link() ?>" class="item-element title<?= (($topline_thumbnail !== 'none') || $topline_summary) ? ' multiline' : '' ?>" dir="auto"><?= $this->entry->title() ?><?php
+		<?= $this->entry->title() ?><?php
```
math-GH, line 59: simply: no way.
The CSS classes are needed

math-GH (changes requested):

math-GH:
> > Why does the list entry need to contain a link when, without it, the application performs exactly as expected?
> 
> Because our Web UI also works to a good extent without JavaScript at all, and the non-JavaScript behaviour is to open links in a new tab. I will try something else in another PR

~~I am thinking about a better non-JS solution: Instead of having an external link the title here could have a #anchor that makes the following article visible (via CSS `:target`). It would make the `target:"_blank"` cruft~~

Edit: The reason for this `<a>` is the middle mouse click, that opens the link in a new tab

tydavis:
@Alkarex - #6937 reproduced the "new tab" behavior. I'm curious about the anchor with CSS target option

tydavis:
>Edit: The reason for this <a> is the middle mouse click, that opens the link in a new tab
Do we hide the "see on original website" link for mobile / smaller screens?

tydavis:
I turned off my extension for the domain and behavior returned to what you described. For reference, https://underpassapp.com/StopTheMadness/

tydavis:
Sorry for raising an issue / PR.

math-GH:
> > Edit: The reason for this  is the middle mouse click, that opens the link in a new tab
Do we hide the "see on original website" link for mobile / smaller screens?

not by default. I wished you can tweak it with the CSS extension. If needed I can help

PR #6937, opened 2024-10-23
Add compatiblity with browser extension 6934

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6934


math-GH:
IMHO we should not fix it in this way for this very specific case.
An extension should do it for everyone who needs that fix.

Alkarex:
Let's hear first whether it helps or not @tydavis

PR #6957, opened 2024-10-31
Change SQL update query

### LINKED ISSUE #5707: [BUG]: Recurring 504 Timeouts and High Resource Usage with MariaDB Operations

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5707
Related to https://github.com/FreshRSS/FreshRSS/pull/8207 for PostgreSQL / SQLite


app/Models/FeedDAOMySQL.php
```diff
@@ -0,0 +1,33 @@
...
+	public function updateCachedValues(int ...$feedIds): int|false {
+		// Performance problem when using `WHERE id_feed IN (...)`, so cannot take advantage of `$feedIds` for now with MySQL / MariaDB
+		$sql = <<<'SQL'
+			UPDATE `_feed`
+			JOIN (
+				SELECT id_feed,
+					COUNT(id) AS total_entries,
+					SUM(CASE WHEN is_read = 0 THEN 1 ELSE 0 END) AS unread_entries
+				FROM  `_entry`
+				-- WHERE id_feed IN (...)
```
Alkarex: We could make an option for that, but really I would prefer something that works well in every MariaDB / MySQL setup...

app/Models/FeedDAOPGSQL.php
```diff
@@ -10,40 +10,4 @@ public function sqlResetSequence(): bool {
...
-	public function updateCachedValues(int ...$feedIds): int|false {
-		// Faster than the MySQL version
-		if (empty($feedIds)) {
-			$whereFeedIds = 'true';
-			$whereEntryIdFeeds = 'true';
-		} else {
-			$whereFeedIds = 'id IN (' . str_repeat('?,', count($feedIds) - 1) . '?)';
-			$whereEntryIdFeeds = 'id_feed IN (' . str_repeat('?,', count($feedIds) - 1) . '?)';
-		}
-		$sql = <<<SQL
```
Alkarex, line 24: Promoted (with minor compatibility changes) to the SQL used by all databases

PR #6959, opened 2024-10-31
Fixed: article footer dropdowns (my labels, article tags, sharing)

math-GH:
Closes #

Before:
grafik
grafik

After:
grafik
grafik

1: only mobile view: the arrow is not at its place
2: only mobile view: better border (each theme decides about `border-radius` and the width of the dropdown is now better calculated)
3: Article tags: Head line added. Now it is in sync to the `my labels` and `share` drowdowns
4: the dropdown menus are now left hand side aligned
5: my labels menu: long labels with a word break are now aligned to the checkbox

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


p/themes/base-theme/frss.rtl.css
```diff
@@ -2421,10 +2437,8 @@ html.slider-active {
 	}
 
 	.dropdown .dropdown-menu {
-		width: 94%;
-		border-radius: 0;
-		right: 3%;
-		left: 3%;
+		right: 0.5rem !important;
+		left: 0.5rem !important;
```
Alkarex, line 2441: Any possibility of avoiding `!important` here?

math-GH: unfortunatelly not :/
It is needed to overrule the following lines that are more specific

grafik

PR #6961, opened 2024-11-01
Rebuild Feed .ico and .txt on cache clear

### LINKED ISSUE #6927: [Bug] Issue with updating favicon when website changes
### Describe the bug

This isn't a big deal, but I noticed my employer's RSS feed had a GitHub favicon, and it turned out to be because they hadn't specified the website URL, so it was defaulting to the github page of the library that generated the feed.

I pointed it out and they fixed it, but then it took me a bunch of trial and error to get FreshRSS to update to the new icon. 

I tried refreshing the feed and clearing the cache, but then I noticed that the Website URL field in FreshRSS still had the old value. So I manually corrected that and then cleared the cache and refreshed again, but the icon *still* didn't update. Finally, on a whim, I added a `?` to the end of the Feed URL and was moderately surprised to see that fixed the issue. 

Even more surprising, when I took the `?` away, it went back to the old favicon, even though nothing in the feed or settings would now point to it. 

So, I'm guessing this is some kind of caching issue? Like maybe the icon is pulled from the Website URL, then cached with the key based on the Feed URL?

### To Reproduce

1. Create an RSS feed with a website that has one favicon, add it to FreshRSS
2. Edit the feed to point to a new website with a new favicon
3. Update FreshRSS settings to also know about the new website
4. Observe that the icon does not change
5. Make an inconsequential change to the Feed URL in FreshRSS, such as adding a `?` to the URL
6. Observe that the icon now changes
7. Revert step 5
8. Observe that the icon also reverts

### Expected behavior

I'd expect the icon to be solely tied to the Website URL and for the Feed URL to have no effect on it when the Website URL is set.

(I'd also prefer to see the website url updated automatically, but that's somewhat beside the point.)

### FreshRSS version

1.24.3

### Environment information

- Database version: SQLite (I think. It doesn't seem to say anywhere in the settings?)
- PHP version: 8.3.12
- Installation type: lscr.io/linuxserver/freshrss Docker
- Web server type: Headers say `server: openresty`, but the logs mention a couple of nginx config files, so I'm not sure. Maybe both? It's whatever comes with the docker image. I also have Nginx Proxy Manager in front of it, handling SSL and port mapping.
- Device: Macbook pro
- OS: Unraid (Slackware Linux)
- Browser: Firefox 131.0.2

### Additional context

_No response_

patHyatt:
Closes #6927 

 - Modify Feed.faviconPrepare signature to allow forced retrieval of the favicon   
 - Add faviconRebuild which clears file cache and re-retrieves icons   
 - Call faviconRebuild when clearCache is fired

1. Add feed
2. Manage feed -> Clear Cache
3. See original files being deleted, and new timestamps for files replaced

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [N/A] unit tests written (optional if too hard)
- [N/A] documentation updated

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -1007,7 +1007,7 @@ class FreshRSS_Feed extends Minz_Model {
 	}
 
 	private function faviconRebuild(): void {
-		$this->faviconDelete($this->hash());
+		FreshRSS_Feed::faviconDelete($this->hash());
 		$this->faviconPrepare(true);
 	}
 
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -225,7 +225,7 @@ class FreshRSS_Feed extends Minz_Model {
 		return $this->nbNotRead;
 	}
 
-	public function faviconPrepare(): void {
+	public function faviconPrepare(bool $force = false): void {
 		require_once(LIB_PATH . '/favicons.php');
 		$url = $this->website;
 		if ($url == '') {
@@ -235,7 +235,7 @@ class FreshRSS_Feed extends Minz_Model {
 		if (@file_get_contents($txt) !== $url) {
 			file_put_contents($txt, $url);
 		}
-		if (FreshRSS_Context::$isCli) {
+		if (FreshRSS_Context::$isCli || $force) {
 			$ico = FAVICONS_DIR . $this->hash() . '.ico';
 			$ico_mtime = @filemtime($ico);
 			$txt_mtime = @filemtime($txt);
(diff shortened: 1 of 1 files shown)


app/Models/Feed.php
```diff
@@ -1006,7 +1006,13 @@ public function cacheFilename(string $url = ''): string {
 		}
 	}
 
+	private function faviconRebuild() : void {
```
Alkarex:
```suggestion
	private function faviconRebuild(): void {
```

Alkarex:
Thanks 👍🏻 

Issue #6949, opened 2024-10-27
[Bug] Some User Query searches in mark-as-read not being marked as read


mtalexan:
### Describe the bug

Sometime between 24a4fcb5c3a001ee9810d7140f107160472fc1d5 (merge of https://github.com/FreshRSS/FreshRSS/pull/6886) and 6c58b90518c30a165db1ee8ea34e0f3d12c6c898 (merge of https://github.com/FreshRSS/FreshRSS/pull/6936), `search:` being used as part of per-feed mark-as-read stopped working consistently.

After updating and converting all escaped parentheses in all regex patterns to be unescaped (for the relevant PR), some, but not all, of the `search:` are not being applied from the per-feed mark-as-read.  This is verifiable as a mark-as-read bug because a few days later the exact mark-as-read pattern can be copy-pasted into the manual search box for the feed and unread entries published after the update are found matching the pattern (SQLite backend is used so the syntax is the same, and every line of the mark-as-read feed is manually ended with ` OR` specifically so it can be easily copy-pasted).   

This can be narrowed down to a few specific User Queries, but there appears to be nothing unique about the patterns that work vs the ones that don't.

Patterns that work properly list no unread Entries in the manual search but do list read Entries.  Patterns that don't work list unread Entries in the manual search.  

### To Reproduce

1. Define some User Queries to match a common set of terms
2. List `search:` for those User Queries in the per-feed mark-as-read
3. As a matter of convenience end every line in the mark-as-read with ` OR` so it can by copy-pasted into the manual search box to run the equivalent.
4. Note the date and time the mark-as-read was updated for the feed
5. Wait a few days
6. Copy the mark-as-read filter pattern for the feed
7. Open the manual search for just unread entries on the one feed
8. Paste in the mark-as-read filter pattern
9. Notice there are a number of unread entries matching the mark-as-read pattern that have been added since the mark-as-read pattern was updated (error)

### Expected behavior

The pattern in the per-feed mark-as-read marks the matching entries as read.  Copy-pasting the mark-as-read filter to the manual search of the feed should find no new unread entries from after the mark-as-read filter was updated. 

### FreshRSS version

1.25.0-dev

### Environment information

- Database version: SQLite
- PHP version: 8.3.10
- Installation type: Docker
- Web server type: whatever is built into the Alpine Dockerfile.
- Device: PC or Android Pixel Fold
- OS: N/A (OpenSUSE MicroOS)
- Browser: Firefox 131.0.2

[...]

PR #6972, opened 2024-11-01
Fix regex search for tags

Alkarex:
Fix regex search for tags, when regex contains parentheses
https://github.com/FreshRSS/FreshRSS/issues/6949#issuecomment-2452363378
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/6926


mtalexan:
I've been testing this and it seems to be the fix. I was trying a couple variations of changes:
- Set "unique by link" instead of the default by GUID
- Use the contents of the User Query directly other tha n as `search:...`
- Just this change

It turns out this change alone was all that was needed to fix the filtering in per-feed mark-as-read rules (#6949)

Alkarex:
Thanks for the feedback @mtalexan

PR #6981, opened 2024-11-07
[Feature] 6975: Redirect to shortcut page on pressing '?'

### LINKED ISSUE #6975: [Feature] Redirect to shortcut page on pressing '?' (or show shortcut info)
Using shortcuts is useful and nice. But remembering them is not easy. 

In other web apps I am used to a popup/information about shortcuts when pressing `?`. 
It would be very useful if a `?` press would either (1) show a popup with shortcuts or (2) redirect to the shortcuts page.

Of course I can navigate to the shortcut page using the mouse. But who uses the mouse these days... 😜

**Additional context**
Many other (web) apps support the `?` button. Intuitively I tried it on FreshRSS too, but no luck 😊.

oshaposhnyk:
Closes #6975

- Added redirect to the configure.shortcut page when pressing Shift + ?

1. Open any page
2. Press Shift + ?
3. Check if the redirect to the configure.shortcut page has occurred

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -973,9 +973,7 @@ function init_shortcuts() {
 	});
 
 	document.addEventListener('keydown', ev => {
-		if (ev.target.closest('input, select, textarea') ||
-				(document.activeElement && document.activeElement.isContentEditable) ||
-				ev.ctrlKey || ev.metaKey || (ev.altKey && ev.shiftKey)) {
+		if (ev.ctrlKey || ev.metaKey || (ev.altKey && ev.shiftKey) || ev.target.closest('input, select, textarea')) {
 			return;
 		}
 
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -963,22 +963,17 @@ function init_column_categories() {
 }
 
 function init_shortcuts_helper() {
-	document.addEventListener("keydown", function (event) {
-		if (event.shiftKey && event.key === "?") {
-			const activeElement = document.activeElement;
-			const isInputField = activeElement.tagName === "INPUT" ||
-				activeElement.tagName === "TEXTAREA" ||
-				activeElement.isContentEditable;
-			if (!isInputField) {
-				window.location.href = context.urls.shortcut.replace(/&amp;/g, '&');
-			}
-		}
-	});
+	const activeElement = document.activeElement;
+	const isInputField = activeElement && (
+		["INPUT", "TEXTAREA", "SELECT"].includes(activeElement.tagName) ||
+		activeElement.isContentEditable
+	);
+	if (!isInputField) {
+		window.location.href = context.urls.shortcut.replace(/&amp;/g, '&');
+	}
 }
 
 function init_shortcuts() {
-	init_shortcuts_helper();
-
 	Object.keys(context.shortcuts).forEach(function (k) {
 		context.shortcuts[k] = (context.shortcuts[k] || '').toUpperCase();
 		if (context.shortcuts[k].indexOf('&') >= 0) {
@@ -1095,6 +1090,9 @@ function init_shortcuts() {
 			}
 			ev.preventDefault();
 			return;
+		}
+		if (ev.shiftKey && ev.key === "?") {
+			init_shortcuts_helper();
 		}
 
 		if (ev.altKey || ev.shiftKey) {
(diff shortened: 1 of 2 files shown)


p/scripts/main.js
```diff
@@ -962,7 +962,23 @@ function init_column_categories() {
 	};
 }
 
+function init_shortcuts_helper() {
+	document.addEventListener("keydown", function (event) {
+		if (event.shiftKey && event.key === "?") {
+			const activeElement = document.activeElement;
+			const isInputField = activeElement.tagName === "INPUT" ||
```
Frenzie: It may be easier to read along these lines, but then again it may be faster as written.

More important note the check that activeElement isn't null and including `SELECT` with the list of elements.

The activeElement thing may be outdated but it wouldn't hurt either, cf. `https://github.com/ten1seven/what-input/issues/38` for example.
```js
const isInputField = activeElement && (
                     ["INPUT", "TEXTAREA", "SELECT"].includes(activeElement.tagName) ||
                     activeElement.isContentEditable
                     );
```

p/scripts/main.js
```diff
@@ -962,7 +962,23 @@ function init_column_categories() {
...
+				activeElement.isContentEditable;
+			if (!isInputField) {
+				window.location.href = context.urls.shortcut.replace(/&amp;/g, '&');
+			}
+		}
+	});
+}
+
 function init_shortcuts() {
+	init_shortcuts_helper();
```
Frenzie: Doesn't the entire thing belong a few lines lower in `document.addEventListener('keydown', ev => {`?

p/scripts/main.js
```diff
@@ -1080,6 +1091,9 @@ function init_shortcuts() {
 			ev.preventDefault();
 			return;
 		}
+		if (ev.shiftKey && ev.key === "?") {
+			init_shortcuts_helper();
```
Frenzie: If you'll permit me one final nitpick, I'd probably just insert it here, but if not the function name is a bit deceptive now. ^_^

Alkarex: See https://github.com/FreshRSS/FreshRSS/pull/6981/commits/015db724e71020c581a1f39a7ada84b23f2ec0b2

p/scripts/main.js
```diff
@@ -1080,6 +1081,10 @@ function init_shortcuts() {
 			ev.preventDefault();
 			return;
 		}
+		if (ev.key === '?' && ev.shiftKey) {
```
Alkarex: On ~~some~~ most keyboard layouts, <kbd>Shift</kbd> must always be pressed to produce a <kbd>?</kbd>
We could consider removing the `ev.shiftKey` test for uniformity

p/scripts/main.js
```diff
@@ -973,7 +973,8 @@ function init_shortcuts() {
 	});
 
 	document.addEventListener('keydown', ev => {
-		if (ev.target.closest('input, textarea') ||
+		if (ev.target.closest('input, select, textarea') ||
+				(document.activeElement && document.activeElement.isContentEditable) ||
```
Alkarex: We might not need that line as I do not think we have anything editable

oshaposhnyk: What about search input? Will the redirect to the shortcuts page not be performed if the user types the '?' character in this input?

Alkarex: Search input is an `<input>` element, which we exclude

p/scripts/main.js
Alkarex: What do you think @Frenzie ?

Frenzie: The shift test feels redundant or perhaps even negative for keyboard layouts that don't require pressing shift to produce a question mark.

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/6981/commits/ee30c8a291125d39c8a047d84c8e21d7041b792a

math-GH:
Regression:
The shortcut leads to an error in anonym. mode.

The shortcut itself is not documented/set in the shortcut config itself.

Side note:
The shortcut `shift+?` works on a German keyboard too. To reach the `?` char we have to press `Shift+?` (`?` is on the same key like `ß`). But it is not a problem.

PR #6983, opened 2024-11-07
Optimize code:

ColonelMoutarde:
before
count(...)' is used in a loop and is a low performing construction. 
after
Foreach instead (easier to read and support)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Frenzie (approved):
I'd disagree with more readable, but I don't mind either way. ;-)

ColonelMoutarde:
> I'd disagree with more readable, but I don't mind either way. ;-)
;-) count(...)' is used in a loop and is a low performing construction.

Frenzie:
Sure, but are there ever more than about 4 parameters? ;-) But agreed it's a good habit in case someone copies it as an example where it might actually matter.

PR #6990, opened 2024-11-11
Fix tag ID uniqueness

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/pull/6052#discussion_r1837266309

### DIFF
--- a/app/views/helpers/index/tags.phtml
+++ b/app/views/helpers/index/tags.phtml
@@ -22,8 +22,9 @@
 	<?php if (!empty($remainingTags)): // more than 7 tags: show dropdown menu ?>
 		<li class="item tag">
 			<div class="dropdown">
-				<div id="dropdown-tags2-<?= $this->entry->id() ?>" class="dropdown-target"></div>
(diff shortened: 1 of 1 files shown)


app/views/helpers/index/tags.phtml
```diff
@@ -22,8 +22,9 @@
 	<?php if (!empty($remainingTags)): // more than 7 tags: show dropdown menu ?>
 		<li class="item tag">
 			<div class="dropdown">
-				<div id="dropdown-tags2-<?= $this->entry->id() ?>" class="dropdown-target"></div>
-				<a class="dropdown-toggle" href="#dropdown-tags2-<?= $this->entry->id() ?>"><?= _i('down') ?></a>
+				<?php $dropId = 'dropdown-tags2-' . $this->entry->id() . '-' . hrtime(true); ?>
```
math-GH:
```suggestion
				<?php $dropId = 'dropdown-tags-' . hrtime(true); ?>
```
could be enough

PR #7000, opened 2024-11-15
Avoid race condition in users' homedir creation

andrey-utkin:

- Avoid race condition in users' homedir creation

Haven't tested this for now, it might take an instrumented test, or a long-running stress-test.

This is for now just me sharing a topic of concern. I just noticed that mkdir usage was too "optimistic", while really strictness is in order. 

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -229,6 +229,10 @@ class FreshRSS_user_Controller extends FreshRSS_ActionController {
 
 		$ok = self::checkUsername($new_user_name);
 		$homeDir = join_path(DATA_PATH, 'users', $new_user_name);
+		// create basepath if missing
+		if (!is_dir(join_path(DATA_PATH, 'users'))) {
+			$ok &= mkdir(join_path(DATA_PATH, 'users'), 0770, true);
+		}
 		$configPath = '';
 
 		if ($ok) {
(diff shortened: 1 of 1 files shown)


app/Controllers/userController.php
```diff
@@ -243,10 +247,12 @@ public static function createUser(string $new_user_name, ?string $email, string
 			$ok &= !file_exists($configPath);
 		}
 		if ($ok) {
-			if (!is_dir($homeDir)) {
-				mkdir($homeDir, 0770, true);
+			// $homeDir must not exist beforehand,
+			// otherwise it might be multiple remote parties racing to register one username
+			$ok = mkdir($homeDir, 0770, true);
```
Alkarex, line 252:
```suggestion
			$ok = @mkdir($homeDir, 0770, true);
```

Would otherwise generate a warning if the directory already exists

andrey-utkin: It makes sense to silence these warnings, but I don't think we want to silence them at the call level here.
- It is a good practice to set `display_errors = Off` in php.ini in production
- The official docker image has that setting
- Minz FrontController sets that setting so, unless `FRESHRSS_ENV` is set to `development`.
- Silencing individual standard function calls would break the way `FRESHRSS_ENV=development` is expected to work
- Silencing individual standard function calls adds cognitive load to comprehend the code

Alkarex: The thing is that we catch the error already, and react to it, so there is not much more to do about it. With the same reasoning, you would have to re-throw every error again in a `try/catch` construct.
I do run with `FRESHRSS_ENV=development` and my expectation is to be visually alerted (and break) in the case of error conditions not accounted for, which is not the case here.
But the situation is likely so rare that I do not really mind

Alkarex:
@andrey-utkin Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

app/Controllers/userController.php
andrey-utkin: I see what you mean - the situation is handled in code, so it's not something the admin has to be notified and do anything about.

PR #7007, opened 2024-11-19
Add extension hook simplepie_after_init

### LINKED ISSUE #7006: [Feature] Extension events for post feed data fetch
I'm trying to implement something suggested in #6556  
A `sleep` or some sort of pause to the update is a simple solution but has little to no flexibility.  
I'm thinking on processing the potential headers some sites send to improve how many feeds can be updated in a given period of time.  

An event `feed_after_actualize` or `simplepie_after_init` which should allow me to get the headers of the request made to the feed.  

Since the extension would already be using `feed_before_actualize` to cancel the update, also could be to manually call the feed's URL to process the headers in this event.  

**Additional context**
None.

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7006

Note: the `simplepie_*` hooks are only for feeds using SimplePie by pull, i.e. normal RSS/Atom feeds. This excludes WebSub (push), and the various HTML or JSON Web scraping methods.

### DIFF
--- a/docs/en/developers/03_Backend/05_Extensions.md
+++ b/docs/en/developers/03_Backend/05_Extensions.md
@@ -178,7 +178,8 @@ The following events are available:
 * `nav_menu` (`function() -> string`): will be executed if the navigation was built.
 * `nav_reading_modes` (`function($reading_modes) -> array | null`): **TODO** add documentation.
 * `post_update` (`function(none) -> none`): **TODO** add documentation.
-* `simplepie_before_init` (`function($simplePie, $feed) -> none`): **TODO** add documentation.
+* `simplepie_after_init` (`function(\SimplePie\SimplePie $simplePie, FreshRSS_Feed $feed, bool $result) -> none`): Triggered after fetching an RSS/Atom feed with SimplePie. Useful for instance to get the HTTP response headers (e.g. `$simplePie->get_headers();`).
+* `simplepie_before_init` (`function(\SimplePie\SimplePie $simplePie, FreshRSS_Feed $feed) -> none`): Triggered before fetching an RSS/Atom feed with SimplePie.
 
 ### Injecting CDN content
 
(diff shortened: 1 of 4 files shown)


docs/en/developers/03_Backend/05_Extensions.md
```diff
@@ -178,7 +178,10 @@ The following events are available:
 * `nav_menu` (`function() -> string`): will be executed if the navigation was built.
 * `nav_reading_modes` (`function($reading_modes) -> array | null`): **TODO** add documentation.
 * `post_update` (`function(none) -> none`): **TODO** add documentation.
-* `simplepie_before_init` (`function($simplePie, $feed) -> none`): **TODO** add documentation.
+* `simplepie_after_init` (`function(\SimplePie\SimplePie $simplePie, FreshRSS_Feed $feed, bool $result) -> none`): Triggered after fetching an RSS/Atom feed with SimplePie. Useful for instance to get the HTTP response headers (e.g. `$simplePie->get_headers();`).
```
pe1uca: The function `get_headers()` would be missing.  
Currently they'd need to be fetched with `$simplePie->data['headers']`

Here's the assignment and in the rest of the class they're accessed the same way  
https://github.com/FreshRSS/FreshRSS/blob/966f211202bc2ed6bf56b64e1ea3c4804e93c404/lib/simplepie/simplepie/src/SimplePie.php#L1877

Alkarex: Thanks; doc fixed

pe1uca:
The hook is missing from this array which without seems it's not fired.  
https://github.com/FreshRSS/FreshRSS/blob/0f4bcdad9b3bde1cf7724069e774eff198a7169c/lib/Minz/ExtensionManager.php#L24

Alkarex:
Sorry, fixed!

pe1uca:
Here's the extension that will make use of this hook in case it can work as a test https://github.com/pe1uca/xExtension-RateLimiter/

Alkarex:
@pe1uca Could you please consider adding your extension to our list?
* https://github.com/FreshRSS/Extensions/blob/master/repositories.json
* https://github.com/FreshRSS/Extensions/blob/master/README.md

pe1uca:
For sure!  
Just will give it a final test and improvement, since right now it assumes simplepie/simplepie#905 has been merged, which by the looks of it won't be, and instead the fix will be simplepie/simplepie#909, so it'll be better to have it working without those at the moment.

Alkarex:
If it takes too long time, we can merge one of them in our own fork

Issue #7011, opened 2024-11-20
[Feature Request] Add "Exit Reader View" Button to Quickly Return to Landing Page


bmmmm:
**Is your feature request related to a problem? Please describe.**  
Not exactly a problem, but a usability improvement. While immersed in reading feeds in the reader view, it’s sometimes cumbersome to navigate back to the landing page or main overview of FreshRSS.

**Describe the solution you’d like**  
I propose adding a floating button (positioned at the bottom, right, or left side of the interface) that allows users to exit the reader view and return directly to the landing page or main feed overview. This would streamline the transition and enhance usability.  

**Describe alternatives you’ve considered**  
Currently, I manually click the title bar (URL) and remove the trailing path to navigate back to the landing page. 

Example visualization:  
![Example Visualization](https://github.com/user-attachments/assets/e74e5961-ed6d-43b5-af5a-c62c0914a7ec)

**Additional context**  
The left and right margins of the reader interface remain largely unused. It might be interesting to explore using these areas for navigation or quick actions.  I am thinking of "using" them for navigation / modification could be cool? 😊

BIG THANKS to the project team!!!1💖💖

Abdallah-Abdaljalil:
Hi @bmmmm, I'm Interested in working on this issue. I would highly appreciate it if you could assign this issue to me.

patman62:
Doesn't clicking the Normal View (or Global View) take you back to where you were before? As a relatively newcomer to FreshRSS the UX seems a little intimidating to noobs and this proposal would not improve that. Of course, if it was opt-in in the display settings that would be ok.

Abdallah-Abdaljalil:
> Doesn't clicking the Normal View (or Global View) take you back to where you were before? As a relatively newcomer to FreshRSS the UX seems a little intimidating to noobs and this proposal would not improve that. Of course, if it was opt-in in the display settings that would be ok.

From what I understand, when the user is in reading mode, it shows all the content under a section (Like Planet GNOME), I guess. So if you were to scroll far whilst reading and you would like to return to the main page, it would be easier to have a button that takes you there straight away that is located on the open space on the left and right. Because, from my experience, the user has to go back to the top to get back to normal or global view or whatever. I don't know if I am making sense here or if I am wrong, but I would appreciate a response to indicate what you think.

bmmmm:
hello everyone,
[...]

PR #7029, opened 2024-11-27
Implement showing and hiding header depending on scroll

Frenzie:
References #7011.

This is just a quick poc, but it shows a possible approach.

1. Scroll down.
2. Scroll up.
3. Repeat.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1956,11 +1956,12 @@ function init_confirm_action() {
 
 function init_scroll_header() {
 	const header = document.querySelector('header.header');
+	const headerHeight = parseInt(getComputedStyle(header).height);
 	if (header) {
 		header.previousScroll = 0;
 		window.addEventListener('scroll', function () {
 			const currentScroll = window.scrollY;
-			if (currentScroll > parseInt(getComputedStyle(document.querySelector('.header')).height) && currentScroll > header.previousScroll) {
+			if (currentScroll > headerHeight && currentScroll > header.previousScroll) {
 				header.classList.add('hide');
 			} else {
 				header.classList.remove('hide');
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1956,8 +1956,8 @@ function init_confirm_action() {
 
 function init_scroll_header() {
 	const header = document.querySelector('header.header');
-	const headerHeight = parseInt(getComputedStyle(header).height);
 	if (header) {
+		const headerHeight = parseInt(getComputedStyle(header).height);
 		header.previousScroll = 0;
 		window.addEventListener('scroll', function () {
 			const currentScroll = window.scrollY;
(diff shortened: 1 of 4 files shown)


Alkarex:
Looks good. It is already an improvement as it is. Maybe the toolbar should be included as well in what is shown when scrolling up
image

Probably related to https://github.com/FreshRSS/FreshRSS/pull/6047 but might be good while waiting for that larger PR to be completed

Frenzie:
If you're happy with it as is I should probably fix the indentation in `app/layout/header.phtml`?

Alkarex:
What do you think @math-GH ? It seems like a fine improvement while waiting for a more comprehensive rework

math-GH:
Nice!

Please check:
When you scroll just 2px down, the header will disappear, but will leave an empty space:
grafik

Frenzie:
> When you scroll just 2px down, the header will disappear, but will leave an empty space:

Adjusted by adding `currentScroll > parseInt(getComputedStyle(document.querySelector('.header')).height)`.

Edit: though if that's static maybe it can simply be stored.

p/scripts/main.js
```diff
@@ -1954,6 +1954,23 @@ function init_confirm_action() {
 	document.querySelectorAll('button.confirm').forEach(function (b) { b.disabled = false; });
 }
 
+function init_scroll_header() {
+	const header = document.querySelector('header.header');
+	const headerHeight = parseInt(getComputedStyle(header).height);
+	if (header) {
```
```suggestion
	if (header) {
		const headerHeight = parseInt(getComputedStyle(header).height);
```

Alkarex:
Following up on https://github.com/FreshRSS/FreshRSS/pull/7029#issuecomment-2514400799 , I still think the toolbar should be included as well in what is shown when scrolling up. Don't you think so?

Frenzie:
The other PR completely changes related code. As you said, it's better to wait until after.

math-GH:
Please check the global view. The header over lays the window with the articles when a feed/category is open

Frenzie:
scrollY becomes 0, so from the logic's perspective it's scrolled up. It's not immediately clear if there's a good way around that because you'd get the same effect from pressing home. Of course the overlay could also simply be given a higher z-index to hide the effect.

Alkarex:
I may have merged a bit too fast here as I observe numerous UI regressions, in particular on mobile, where the toolbar appears far too often, in particular when expanding an article, but also when scrolling normally.
And on desktop using the normal view, the bar hides access to the main view:

image

So I suggest to revert and rework https://github.com/FreshRSS/FreshRSS/pull/7064

Alkarex:
I was hopping to make a quick fix, but still not so good https://github.com/FreshRSS/FreshRSS/pull/7063

PR #7030, opened 2024-11-27
New button to delete errored feeds from a category

### LINKED ISSUE #7025: [Feature] Bulk remove feeds with errors
I recently imported an OPML file with about 400 feeds in it. About half of them are defunct. I love that FreshRSS provides an indication when the feed url can't be reached, but it would take a lot of time and effort to go through and individually remove each bad feed.

I would be very helpful if I could get a button next to the `Show only feeds with errors` button that let's me remove all feeds with errors.

Screenshot 2024-11-26 at 10 09 35 AM

If there is some way I can do this in the database I'm open to it. Also, I would consider some type of tool the checks all the links my OPML file and removes broken ones.

### LINKED ISSUE #7026: [Bug] Cannot delete muted feeds
### Describe the bug

In Subscription Management, when "Delete muted feeds" is clicked, a message appears that says "Category cannot be updated" and muted feeds are not deleted.

### To Reproduce

1. Go to Subscription Management
2. Click the gear icon next to a category that contains at least one muted feed.
3. Click the red button that says "Delete muted feeds"
4. Click "OK" to confirm the browser dialog that appears that says "Are you sure you want to perform this action? It cannot be cancelled!"

### Expected behavior

Muted feeds removed

### FreshRSS version

1.24.3

### Environment information

- Database version: SQLite
- PHP version: 8.2.15
- Installation type: Manual
-Web server type: Apache
- Device: MacBook Pro
- OS: MacOS 14.5
- Browser: Chrome Version 131.0.6778.86 (Official Build) (arm64)

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7025
fix https://github.com/FreshRSS/FreshRSS/issues/7026

image


app/i18n/en/gen.php
```diff
@@ -18,6 +18,8 @@
 		'back_to_rss_feeds' => '← Go back to your RSS feeds',
 		'cancel' => 'Cancel',
 		'create' => 'Create',
+		'delete_all_feeds' => 'Delete all feeds',
+		'delete_errored_feeds' => 'Delete feeds in error',
```
math-GH:
```suggestion
		'delete_errored_feeds' => 'Delete feeds with errors',
```

Alkarex: I hesitated a bit. "Errored feeds" is likely a bit too nerdy. A feed "with errors" is not necessarily the same as a feed in an error state

Alkarex: More feedback / argumentations welcome

Frenzie: In (an) error state?

math-GH: "Feeds with errors" 
grafik

grafik

But it is not consistent used:
grafik

grafik

And to be honest: I do not like "errors" and "problems" naming. What about "broken" or "stopped/stopped working"? Has "warnings" or an "issue"?

Alkarex: *broken* would also be fine for me

Frenzie: I don't see any issue with "has encountered a problem," but I'd prefer *stopped working* over *broken*, since *broken* puts the onus very much on the feed itself which is not a given.

Alkarex: For a button, I would rather avoid a full sentence and keep it very short. Let's use "feeds with errors" in this PR as this is what we already have in other messages. I hope it is OK.
I will make another PR for changing the *always reachable*, which looks like a wrong French to English translation

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/7049

PR #7031, opened 2024-11-27
improved: article footer for small displays

math-GH:
Before:
on small screens the article footer links for `my labels`, `article tags`, `share` could have a line break

grafik

After:
Do not show the text of the links, if the main stream is smaller than 600px.
grafik

- CSS. Use of `@container` queries.

1. normal view
2. change the size of the window
3. see the article footer

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/views/helpers/index/normal/entry_bottom.phtml
+++ b/app/views/helpers/index/normal/entry_bottom.phtml
@@ -39,7 +39,7 @@
 		<div class="item-element dropdown dynamictags">
 			<div id="dropdown-labels-<?= $this->entry->id() ?>" class="dropdown-target"></div>
 			<a class="dropdown-toggle" href="#dropdown-labels-<?= $this->entry->id() ?>">
-				<?= _i('label') ?><?= _t('index.menu.tags') ?>
+				<?= _i('label') ?><span class="dropdown-label"><?= _t('index.menu.tags') ?></span>
 			</a>
 		</div>
 	</li><?php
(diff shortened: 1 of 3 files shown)


Alkarex:
I have not checked: what is the behaviour on Web browser not supporting `@container` ?

Frenzie:
I think this isn't quite the right solution. It's almost entirely identical to a regular media query, still subject to what the user has or hasn't enabled, while the power of container queries is that you could stick it on the individual elements.

That's not an objection to this form, but I'd just stick it under `@media (max-width: 840px) {` in that case.

math-GH:
> I have not checked: what is the behaviour on Web browser not supporting `@container` ?

It should happen nothing, because the text will not be hidden in that case. It would look like "before".

math-GH:
> I think this isn't quite the right solution. It's almost entirely identical to a regular media query, still subject to what the user has or hasn't enabled, while the power of container queries is that you could stick it on the individual elements.
> 
> That's not an objection to this form, but I'd just stick it under `@media (max-width: 840px) {` in that case.

Okay, let me explain a bit what happend while I tried to find a solution. I invested a couple of hours to investigate what works and what does not work and what fit best:

Idea 1: use the Container query directly at the individual element. It is not easy to implement, because it does not work in inline elements. But I managed it. But: The dropdown menu does not work anymore because the container cuts everything that would overlay outside the container. The menu will be cut.

Idea 2: use the default media query as suggested but it is not the best solution. Between 840px (media query) and 600px (container break point) is enough space to display the text. 

That is why I chose this solution by purpose that you can see in this PR

Frenzie:
> I have not checked: what is the behaviour on Web browser not supporting `@container` ?

Forgot to say, but it's just as if it wasn't there.

Alkarex:
Let's go for that but feel free to propose another approach @Frenzie

PR #7033, opened 2024-11-28
Fix API use rawurldecode

Alkarex:
Labels or categories containing a `+` were failing.
And avoid returning everything if the label/category filter is not found.
Compatibility with FocusReader


Alkarex:
Merging a bit fast to better allow testing API clients

PR #7036, opened 2024-11-29
Add number of feeds in subscription view

### LINKED ISSUE #7035: [Feature] Show amount of feeds in category
I have a lot of feeds in some categories, would like to see how many are in each.  

Show the amount of feeds in a category in the `Subscription management` screen

**Additional context**

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7035

image


lib/lib_rss.php
```diff
@@ -191,12 +191,8 @@ function escapeToUnicodeAlternative(string $text, bool $extended = true): string
...
 }
 
-/** @param int|float $n */
-function format_number($n, int $precision = 0): string {
-	// number_format does not seem to be Unicode-compatible
-	return str_replace(' ', ' ',	// Thin non-breaking space
-		number_format((float)$n, $precision, '.', ' ')
-	);
+function format_number(int|float $n, int $precision = 0): string {
+	return number_format((float)$n, $precision, '.', ' ');	// Thin non-breaking space
```
Alkarex, line 221: `number_format` does seem to be Unicode-compatible now

Alkarex:
@math-GH Maybe you will want to fine tune a bit the style

math-GH:
I’m not a fan of implementing a solution in the core, nor am I fond of this particular style.

I would propose to create an extension for it. It would be a simple JavaScript that counts the items of each category.

That is the current global view The counter has a complete different meaning here
grafik


Why I do not like the style: The numbers are used in the other cases for the "unread" counter.

And a question left:
What is the information behind "501" feeds in this category vs. "299" feeds vs. "8" feeds?

Alkarex:
Fair point for the style being used in other contexts. But in a view, which purpose is to manage feeds in categories, showing their number seems like a reasonable basic expectation.

So other style suggestions welcome.

When categories are used to organise larger amounts of feeds, showing their numbers is interesting. I have for instance a case of organising many companies.
Furthermore, some categories may have a dynamic number of feeds, and seeing changes can be useful.

P.S. I can for instance remember in the past having exported to OPML to get this information

math-GH:
What if it gives more information than just the number of feeds?
- number of feeds
- number of scrapped websites (feed type that uses XPath)
- number of "error" feeds
- number of muted feeds

=> 4 numbers + icon for each => give numbers in 1 rows below the headline, above the feed list (or have a fixed footer in each box, that includes the "add feeds" link too)

And the "Display position" of the category (if given) could be useful too.

math-GH:
Let me link #4672 that wants to improve the subscription page too

Alkarex:
@math-GH Sounds nice 👍🏻

bhj:
The simple solution is to show the count in parentheses after the title, rather than right-aligned in a badge

Alkarex:
Something like that? Help to polish the styling welcome + maybe i18n.

🗞️ feeds · 🔇 muted · ⚠️ errors · 🕸️ scraping · 🎣 full content retrieval

image

math-GH:
I like this solution. Great!
What would I expect as a user: each icon should be a simple filter button. Click on it and show only this entries. Please use here `<button>` incl. a `title`

Currently I am not so super happy while using UTF8 icons instead of . But I know that the triangle symbol and muted symbol are used already as UTF8. So let's keep it for the first shot.
The feed count icon: I would purpose to use the RSS icon (I know it is not 100% accurate but self explaining). Or what about this: 🧾 or 📰
Icon propose for web scrapping: ⛏️ (a pick)
Icon for full content retrieval: I cannot see what kind of icon it is. This icon could be used too: ♨️(it has no real meaning but it is striking) 😆 
Or the scissors could represent it as well? ✂️

math-GH:
What does this number mean?
grafik

Alkarex:
Number of unread articles, like in the main view

math-GH:
> Number of unread articles, like in the main view

Is it interesting in this view?

Alkarex:
> Is it interesting in this view?

I would say a bit interesting, yes, and also for consistency with the main view, but I do not really mind

Alkarex:
Would you feel like pushing some of your suggestions, @math-GH ?

> Or the scissors could represent it as well? ✂️

I like it

math-GH:
> > Is it interesting in this view?
> 
> I would say a bit interesting, yes, and also for consistency with the main view, but I do not really mind

I would not expect that number here. Would it increase the used resource to calculate the number?

In my opinion another number is more interesting if used: The `display position` number. It would explain the order of the categories.

Alkarex:
> Would it increase the used resource to calculate the number?

No, this information comes for free in this view.

I do not really mind, so if you have energy for polishing this PR, please go ahead as you believe is best

hkcomori:
I think making a list of display position numbers is a good idea.
Would it be better to merge #7341 into this view or separate it into a another view?
In my opinion, the latter is more simple.
If the latter, it is not required in this view, because a list of display position numbers is provided in #7341 view (of course, both views could provide it).

aledeg:
As stated [here](https://github.com/FreshRSS/FreshRSS/pull/4672#issuecomment-3242045358), I think it would be better to have a list of feeds and categories in a table where we can apply some filters (ex: show muted feeds, show feeds with an overriding configuration, ...).

This list could have mode information than what is given in the current form.

Alkarex:
> As stated [here](https://github.com/FreshRSS/FreshRSS/pull/4672#issuecomment-3242045358), I think it would be better to have a list of feeds and categories in a table where we can apply some filters (ex: show muted feeds, show feeds with an overriding configuration, ...).
> 
> This list could have mode information than what is given in the current form.

Sounds like an promising idea

Issue #6989, opened 2024-11-11
Upgrade to PHPStan 2.0


Alkarex:
https://github.com/phpstan/phpstan/releases/tag/2.0.0
https://github.com/phpstan/phpstan/blob/2.0.x/UPGRADING.md
And to Level 10, either at the same time or as a following step https://phpstan.org/user-guide/rule-levels

PR #7052, opened 2024-12-03
Update dev dependencies

Alkarex:
I was expecting some of them (e.g. PHPStan) as part of the last Dependabot round, so here is a global update


Frenzie:
I think dependabot has an open PR limit. Maybe that's related?

You can check up a bit under https://github.com/FreshRSS/FreshRSS/network/dependencies

It says it hasn't detected anything since September.
image

You can however click check for updates, which I just did. Mainly to see if it'll detect anything relating to Composer.

Frenzie:
It lists some problems. I guess it's not capable of updates on 1.x when 2.x exists, or something along those lines.

https://github.com/FreshRSS/FreshRSS/network/updates/926901955
```
updater |   Problem 1
updater |     - Root composer.json requires phpstan/phpstan >= 1.12.7, == 2.0.3, found phpstan/phpstan[2.0.3] but these were not loaded, likely because it conflicts with another require.
updater |   Problem 2
updater |     - phpstan/phpstan-phpunit is locked to version 1.4.0 and an update of this package was not requested.
updater |     - phpstan/phpstan-phpunit 1.4.0 requires phpstan/phpstan ^1.11 -> found phpstan/phpstan[1.11.0, ..., 1.12.12] but it conflicts with your root composer.json require (>= 1.12.7, == 2.0.3).
updater |   Problem 3
updater |     - phpstan/phpstan-strict-rules is locked to version 1.6.1 and an update of this package was not requested.
updater |     - phpstan/phpstan-strict-rules 1.6.1 requires phpstan/phpstan ^1.12.4 -> found phpstan/
```

Alkarex:
Good guess. Related to https://github.com/FreshRSS/FreshRSS/issues/6989

PR #7055, opened 2024-12-03
Parentheses in quoted search

Alkarex:
Allow parentheses in quoted search like `author:"Bob (Team1)"`
Related to https://github.com/FreshRSS/FreshRSS/pull/7054

PR #7056, opened 2024-12-03
Avoid Unicode escape of authors in HTML UI

### LINKED ISSUE #2770: [1.15.3] Slashes in author name are UTF-8 encoded in Search byline
My setup:
- Server: Debian 10
- FreshRSS: 1.15.3
- `AllowEncodedSlashes On` is set in the Apache vhost (per the setup docs)

I have a workstation which turns out does not have a full UTF-8 character set (my problem) and specifically is missing the "fullwidth solidus" (slash) character in the UTF-8 encodings: https://www.fileformat.info/info/unicode/char/ff0f/index.htm

The reddit RSS feeds supply the `author` element with the username in low-ASCII slashes (char 47), however it appears that FreshRSS is mistakenly UTF-8 encoding those in the author Search injected byline - but not in the Submitted author line in the same post. Visually, with my broken font:

---
slashfull
---

The HTML for each element looks like so, which might be hard to see as it will render with your local font correctly (assuming you don't have a broken font too :)), but the encoded chars are `ef bc 8f` (UTF-8 U+FF0F) before and after the "u" in `/u/DigitalGoomba` in this example:

top:
```
<div class="author">By: <em><a href="./?search=author%3A%EF%BC%8Fu%EF%BC%8FDigitalGoomba">／u／DigitalGoomba</a></em>
```
bottom:
```
submitted by   <a href="https://www.reddit.com/user/DigitalGoomba"> /u/DigitalGoomba </a>
```

I'm of the initial opinion that the encoding of the forward slash into it's UTF-8 variant is a mistake in the Search injected byline, and that they (the author elements) should be presented as low-ASCII characters on the Search byline the same as they are in the Submitted byline.

Thanks!

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/2770
Related to https://github.com/FreshRSS/FreshRSS/pull/7055

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -640,7 +640,7 @@ class FreshRSS_Feed extends Minz_Model {
 				foreach ($authors as $author) {
 					$authorName = $author->name != '' ? $author->name : $author->email;
 					if (is_string($authorName) && $authorName !== '') {
-						$authorNames .= escapeToUnicodeAlternative(strip_tags($authorName), true) . '; ';
+						$authorNames .= html_only_entity_decode(strip_tags($authorName)) . '; ';
 					}
 				}
 			}


Alkarex:
Example of feed to test: https://theconversation.com/articles.atom?language=en

Stress tests much welcome for this PR (multiple feeds, feed formats, and also API clients)

PR #7059, opened 2024-12-04
Fix initial scroll for some browsers

aarnej:
For some browsers (I tested desktop Edge and mobile Safari), setting document.scrollingElement.scrollTop to zero does not seem to be enough to reset the scroll position at start. Setting history.scrollRestoration = 'manual' seems to fix it for these browsers. Firefox seems to work without this fix but works also with it.

1. Open FreshRSS and scroll down some feed that has entries
2. Press the reload button in the browser
3. Verify that scroll position is reset to beginning of the feed and that no unexpected entries were marked as read if read-on-scroll is enabled in FreshRSS config.
4. Repeat for different browsers

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -2028,6 +2028,7 @@ function init_normal() {
 }
 
 function init_main_beforeDOM() {
+	history.scrollRestoration = 'manual';
 	document.scrollingElement.scrollTop = 0;
 	init_shortcuts();
 	if (['normal', 'reader', 'global'].indexOf(context.current_view) >= 0) {


p/scripts/main.js
```diff
@@ -2028,6 +2028,7 @@ function init_normal() {
 }
 
 function init_main_beforeDOM() {
+	history.scrollRestoration = 'manual';
```
Frenzie, line 2031: Makes sense. The sample on MDN does this fwiw:

```js
if (history.scrollRestoration) {
  history.scrollRestoration = 'manual';
}
```

Frenzie (approved):

p/scripts/main.js
Alkarex: It is probably a good idea to use the same construct, then, don't you think?

Frenzie: I suppose the goal would be to avoid accidentally fooling some other piece of code. You can safely assign history.just_about_anything.

Alkarex:
Thanks @aarnej 👍🏻 

aarnej:
> Thanks @aarnej 👍🏻 Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

Thanks a lot all! I added a PR to update credits: https://github.com/FreshRSS/FreshRSS/pull/7065

PR #7060, opened 2024-12-04
API support edit multiple tags

### LINKED ISSUE #7057: [Bug] GReader API: edit-tag w/ multiple labels
### Describe the bug

Hey again :) This is similar to #7017, but dealing with items. When multiple labels are specified (i.e. the `a` or `r` param is repeated) only the last param is applied. From what I [can](https://feedhq.readthedocs.io/en/latest/api/reference.html#edit-tag) [gather](https://web.archive.org/web/20120204201356/http://code.google.com/p/google-reader-api/wiki/ApiEditTags), it should be possible to add/remove multiple labels at once.

It's a minor thing, but I noticed multiple `i` params are already supported, so ideally this could be improved to more closely align with "the spec".

### To Reproduce

curl 'https://example.com/api/greader.php/reader/api/0/edit-tag' \
  --data-raw 'i=tag%3Agoogle.com%2C2005%3Areader%2Fitem%2F0006265cbcf0ab7c&a=user%2F-%2Flabel%2Ftestlabel1&a=user%2F-%2Flabel%2Ftestlabel2'

### Expected behavior

Both labels are applied

### FreshRSS version

1.24.3

### Environment information

- Installation type: Docker

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7057

### DIFF
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -785,9 +785,11 @@ final class GReaderAPI {
 	}
 
 	/**
-	 * @param array<string> $e_ids
+	 * @param array<string> $e_ids IDs of the items to edit
+	 * @param array<string> $as tags to add to all the listed items
+	 * @param array<string> $rs tags to remove from all the listed the items
 	 */
-	private static function editTag(array $e_ids, string $a, string $r): never {
+	private static function editTag(array $e_ids, array $as, array $rs): never {
 		foreach ($e_ids as $i => $e_id) {
 			if (!ctype_digit($e_id) || $e_id[0] === '0') {
 				$e_ids[$i] = hex2dec(basename($e_id));	//Strip prefix 'tag:google.com,2005:reader/item/'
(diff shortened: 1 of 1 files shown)


p/api/greader.php
```diff
@@ -785,9 +785,11 @@ private static function streamContentsItems(array $e_ids, string $order): never
 	}
 
 	/**
-	 * @param array<string> $e_ids
+	 * @param array<string> $e_ids IDs of the items to edit
+	 * @param array<string> $as tags to add to all the listed items
+	 * @param array<string> $rs tags to remove from all the listed the items
```
bhj: Small typo - should be `the listed items`

bhj (commented):
Nice, did some basic tests and it's working well!

Not exactly related, but in the comment on line 1099, the bit about `The parameter may
be repeated...` is incorrect I think?

Alkarex:
> Not exactly related, but in the comment on line 1099, the bit about `The parameter may be repeated...` is incorrect I think?

@bhj I guess this might be taken from:

* https://github.com/mihaip/google-reader-api/blob/master/wiki/ApiStreamItemsIds.wiki#input
* https://web.archive.org/web/20120203181807/http://code.google.com/p/google-reader-api/wiki/ApiStreamItemsIds#Input
* http://ranchero.com/downloads/GoogleReaderAPI-2009.pdf (page 12)

At the moment, repeated `s` values are not supported by FreshRSS (which supports only one at a time). It would also be a bit challenging to implement. But let me know if this is something you see in other client and/or server implementations.

bhj:
> At the moment, repeated `s` values are not supported by FreshRSS

Right, the comment is confusing since it doesn't match the implementation. If it's more of a TODO I think that could be made more clear for future code divers.

Alkarex:
Good idea. Would you feel like sending a PR?

PR #7063, opened 2024-12-05
Alternative on_scroll_header

Alkarex:
Draft of alternative for https://github.com/FreshRSS/FreshRSS/pull/7029
But still not so good

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -785,6 +785,7 @@ function auto_share(key) {
 let box_to_follow;
 
 function onScroll() {
+	on_scroll_header();
 	if (!box_to_follow) {
 		return;
 	}
@@ -818,7 +819,7 @@ let lastScroll = 0;	// Throttle
 let timerId = 0;
 function debouncedOnScroll() {
 	clearTimeout(timerId);
-	if (lastScroll + 500 < Date.now()) {
+	if (lastScroll + 500 <= Date.now()) {
 		lastScroll = Date.now();
 		onScroll();
 	} else {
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -785,6 +785,7 @@ function auto_share(key) {
 let box_to_follow;
 
 function onScroll() {
+	on_scroll_header();
```
Frenzie, line 788: That sucks terribly imho because the point of this function is not to batter the server with requests.

p/scripts/main.js
```diff
@@ -1954,21 +1955,25 @@ function init_confirm_action() {
...
-			} else {
-				header.classList.remove('hide');
-			}
-			header.previousScroll = currentScroll;
-		});
 	}
+	const currentScroll = window.scrollY;
+	if (currentScroll <= header.headerHeight || currentScroll < header.previousScroll - 4) {
+		header.classList.remove('hide');
+	} else if (currentScroll > header.previousScroll + 8) {
```
Frenzie, line 1973: I also strongly dislike this tbh. :-/

Alkarex: It is not working so well anyway, but without it, there were far too many shifts between showing/hiding the bar

Frenzie:
Could you describe what it is you're trying to accomplish or avoid exactly?

p/scripts/main.js
Alkarex: No, the point was to avoid very fast firing events taking a lot of CPU

Frenzie:
I hadn't seen https://github.com/FreshRSS/FreshRSS/pull/7029#issuecomment-2520086201, never mind.

Alkarex:
This PR was mainly to keep a trace of my quick attempt to fix a few problems:

* Toolbar being shown much too often during scrolls, especially on mobile
    * More precisely, the state shown/hidden would change too easily
* Avoid very fast firing on-scroll events

I stopped when I realised there were several other regressions
https://github.com/FreshRSS/FreshRSS/pull/7029#issuecomment-2520086201

p/scripts/main.js
Frenzie: Fair enough. (My poor server suffers more from the lot of CPU thing.) Scroll events already seem to be limited to some 60 fps despite the polling rate being higher afaik. Personally I'd much rather not have it than have it laggy.

If you're worried that scroll events might exceed the framerate you could do something like this with `requestAnimationFrame`, maybe coupled with a limitation to e.g. 30 fps. That is, crucially, closer to 50 ms than to 500 ms.

```js
let ticking = false;

function onScrollRequestAnimationFrame() {
  if (!ticking) {
    window.requestAnimationFrame(() => {
      // logic
      ticking = false;
    });
    ticking = true;
  }
}
```

PR #7066, opened 2024-12-05
Fix invalid login message

### LINKED ISSUE #7061: [Bug] Login with wrong user name: error message "The web server seems to be incorrectly configured for cookies required for PHP sessions!"
### Describe the bug

(https://mastodon.social/@ts-new@hub.tschlotfeldt.de/113596516493945482 made me aware of this bug)

Login procedure: If the user name is NOT correct the error message is strange/wrong in that case.

(Not correct means also that the user name is case sensitiv!)

The error message: "The web server seems to be incorrectly configured for cookies required for PHP sessions!"

grafik
(screen shot from demo.freshrss.org - user name -> I tried `demos` instead of `demo`)

### To Reproduce

1. Go to login page
2. try to login with a non existing user
3. see the misleading error message

(reproduce-able on demo.freshrss.org too)

### Expected behavior

I would expect a message like: "Wrong user name or password".
And I would not expect that there is an issue with the case sensitivity of the user name

### FreshRSS version

1.25.0-dev

### Environment information

- Database version: ~SQLite~ MySQL
- PHP version: PHP 8.1
- Installation type: git
-Web server type:  Apache

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7061

### DIFF
--- a/app/Controllers/javascriptController.php
+++ b/app/Controllers/javascriptController.php
@@ -78,5 +78,6 @@ class FreshRSS_javascript_Controller extends FreshRSS_ActionController {
 			$this->view->salt1 .= $alphabet[random_int(0, 63)];
 		}
 		$this->view->nonce = sha1('' . mt_rand());
+		Minz_Session::_param('nonce', $this->view->nonce);
 	}
 }


math-GH:
I cannot reproduce the bug on my testing system 🙈

After merging, @marienfressinaud please update demo.freshrss.org . I was finding the bug there.

Alkarex:
> I cannot reproduce the bug on my testing system

Try to clear your cookies

math-GH:
Thanks. Now I can reproduce it.
And the fix looks fine.

Frenzie:
Of possible note that you can still distinguish between the page `Error 403 - Forbidden: Login is invalid` and the notification `Login is invalid` (= account exists).

Alkarex:
Yes Frenzie, it could be an improvement to make that more fuzzy

marienfressinaud:
I'm not sure if you still need the demo to be updated, but it is now :)

math-GH:
> I'm not sure if you still need the demo to be updated, but it is now :)

`Error during context user init!` when I opened the demo

:/

marienfressinaud:
Indeed, the reset script wasn't working anymore because the "demo" user wasn't deleted properly (fixed in https://github.com/FreshRSS/xExtension-Demo/commit/031510ecf3e0a420237163fa5624373db0fe78e1). The demo should work now.

PR #7037, opened 2024-11-30
Fix CSS selector for removing elements

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7014
Case of negative filter matching the positive filter

Issue #7014, opened 2024-11-21
[Bug] CSS selector to remove elements is not removing elements properly


Tywele:
### Describe the bug

I have a CSS selector in the config field to remove a certain element from an article and it's not being removed.

### To Reproduce

1. Using this feed: `https://www.queer.de/rss/newsticker2.0.komplett.xml`
2. Add the article CSS selectors: `.artikel>p, .artikel>blockquote, .artikel>.bildkasten`
3. Add the CSS selectors to remove elements: `p[data-sanitized-class="leistenbanner"]` or `.artikel>p[data-sanitized-class="leistenbanner"]` or `.artikel>.leistenbanner` or `.leistenbanner` (none of them work)
4. See the element still being fetched in the article

### Expected behavior

The element should not appear in the fetched article

### FreshRSS version

1.25.0-dev

### Environment information

Installation type: Docker

### Additional context

Screenshot of website with uBlock Origin disabled to see the element I want to remove
![Screenshot of website with uBlock Origin disabled to see the element I want to remove](https://github.com/user-attachments/assets/d558b65a-55dd-4d46-a716-87a5a49aa997)

Screenshot of the article in my FreshRSS feed to see that the element is still in the article
![Screenshot of the article in my FreshRSS feed to see that the element is still in the article](https://github.com/user-attachments/assets/4d4bb870-b38b-4f1f-a74b-c9473a4bbb18)

Alkarex:
Could you please export the feed settings as OPML as the bottom of the feed settings?

Tywele:
https://drive.proton.me/urls/XHQVXHZ5CG#o20KuHA6q6kv

Github didn't let me attach an XML file to a comment so here is Proton Drive link to the file.

Frenzie:
You can always zip it or simply add .txt at the end.

Edit: or for something relatively small, just paste it.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<opml xmlns:frss="https://freshrss.org/opml" version="2.0">
  <head>
    <title>FreshRSS</title>
    <dateCreated>Thu, 21 Nov 2024 10:28:05 +0100</dateCreated>
  </head>
  <body>
[...]

PR #7073, opened 2024-12-07
Sanitize before CSS manipulation

### LINKED ISSUE #7039: Hint about sanitized CSS classes
> The hint about sanitized CSS classes should be removed then if you can't use them there. That's why I tried them.

Let me address that in another issue / PR. On the road at the moment, but we might have a discrepancy between filtering on the content provided by the RSS feed and filtering on the content scraped from the Web site.

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/issues/7014#issuecomment-2509020304_

Alkarex:
Allows using the same CSS filters for content coming from RSS feeds and from Web scraping
fix https://github.com/FreshRSS/FreshRSS/issues/7039
https://github.com/FreshRSS/FreshRSS/issues/7014#issuecomment-2508987606
https://github.com/FreshRSS/FreshRSS/pull/7037

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -844,7 +844,13 @@ HTML;
 				$base = (parse_url($url, PHP_URL_SCHEME) ?? 'https') . ':' . $base;
 			}
 
-			$content = '';
+			unset($xpath, $doc);
+			$html = sanitizeHTML($html, $base);
+			$doc = new DOMDocument();
+			$doc->loadHTML($html, LIBXML_NONET | LIBXML_NOERROR | LIBXML_NOWARNING);
+			$xpath = new DOMXPath($doc);
+
+			$html = '';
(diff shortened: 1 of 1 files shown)


app/Models/Entry.php
```diff
@@ -844,7 +844,13 @@ public function getContentByParsing(string $url = '', int $maxRedirs = 3): strin
 				$base = (parse_url($url, PHP_URL_SCHEME) ?? 'https') . ':' . $base;
 			}
 
-			$content = '';
+			unset($xpath, $doc);
+			$html = sanitizeHTML($html, $base);
+			$doc = new DOMDocument();
+			$doc->loadHTML($html, LIBXML_NONET | LIBXML_NOERROR | LIBXML_NOWARNING);
```
Alkarex, line 850: UTF-8 encoding fixed in https://github.com/FreshRSS/FreshRSS/pull/7081

PR #7083, opened 2024-12-10
Fix CSS filter help message

Alkarex:
https://github.com/FreshRSS/FreshRSS/issues/7014#issuecomment-2531610211

### DIFF
--- a/app/i18n/cs/sub.php
+++ b/app/i18n/cs/sub.php
@@ -64,7 +64,7 @@ return array(
 		'css_path' => 'Přepínač CSS článku na původních webových stránkách',
 		'css_path_filter' => array(
 			'_' => 'selektor CSS prvků, které mají být odstraněny',
-			'help' => 'Selektor CSS může být seznam, například: <kbd>.footer, .aside, p[data-sanitized-class~="menu"]</kbd>',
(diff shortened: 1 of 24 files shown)


Frenzie (approved):

Tywele:
I'm still confused about this change:
grafik

Why did you remove the dots before the classnames? This way they don't properly represent CSS selectors anymore IMO. I think a much better hint would be something like `.footer, .header, div, #adbanner, p[data-sanitized-class~="menu"]` to show that pretty much all CSS selectors and additionally special syntax unique to FreshRSS also works.

Frenzie:
`footer` selects `<footer>`, `.footer` selects `class=footer`. Both are valid selectors.

Alkarex:
We could though add again an example of class selector like `.content`, maybe instead of `footer` which is a bit redundant since we already have `aside` in this example. PR welcome

Frenzie:
Agreed, something like `.content` or `.entry-content` sounds like a good example.

Tywele:
> `footer` selects `<footer>`, `.footer` selects `class=footer`. Both are valid selectors.

Yes I know I just couldn't think of a common class name in articles and instead used `div` as an element selector 😅 
And the current hint only represents element selectors and the special syntax and no class or id selectors which is what I was trying to say with my comment.

PR #7077, opened 2024-12-08
windows.open noopener

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/6862
Rework https://github.com/FreshRSS/FreshRSS/pull/2506

PR #7089, opened 2024-12-11
Always prevent default regardless of window.open() return status

Frenzie:
Fixes regression noted in https://github.com/FreshRSS/FreshRSS/pull/7077#discussion_r1879016226

1. Use a key like space to open new windows
2. Press space
3. Only a window should open without any scrolling

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1099,7 +1099,7 @@ function init_shortcuts() {
 
 			const link_go_website = document.querySelector('.flux.current a.go_website');
 			if (link_go_website) {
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -1100,9 +1100,7 @@ function init_shortcuts() {
 			const link_go_website = document.querySelector('.flux.current a.go_website');
 			if (link_go_website) {
 				const newWindow = window.open(link_go_website.href, '_blank', 'noopener');
```
```suggestion
				window.open(link_go_website.href, '_blank', 'noopener');
```

Fix CI failure

PR #7107, opened 2024-12-15
Add support for custom stylesheet URL in UserCSS extension

Rambomst:

- This update introduces the ability for users to specify a custom stylesheet URL in the UserCSS extension.
- It includes configuration fields, backend handling, and CSP updates to allow external stylesheets while maintaining security. Existing functionality remains untouched.

1. Go to the User CSS settings area.
2. Set a custom stylesheet URL.
3. Load the FreshRSS main page and view the page source to see the custom stylesheet URL has been loaded.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


Frenzie (approved):
lgtm

math-GH:
~what is the use case?~

link: https://github.com/FreshRSS/Extensions/pull/261

Alkarex:
I am a bit worried of having something like that in the core, which dynamically modifies CSP security headers.
What would be a typical location for your CSS file, in your personal use case, @Rambomst ?

Rambomst:
> I am a bit worried of having something like that in the core, which dynamically modifies CSP security headers. What would be a typical location for your CSS file, in your personal use case, @Rambomst ?

I have the CSS file sitting in another docker container on the same server though its referenced using a full domain `https://example.com/my.css`.

Since this is a user based plugin and the CSS URL entered will only be displayed to the current logged in user after they themselves have set the URL, would it be a huge concern? Perhaps a warning message about the potential perils of not using a domain under your control?

Alkarex:
We should in particular double-check that this is not used for instance on the login screen, if used by the default user

math-GH:
My 2 cents here:
I am not a fan of having this security change inside a default delivered extension.
I understand the use case but to honest: If you have hundrets/thousands of CSS lines added why do not develop a new own theme? It is very simple to use the CSS files from exisiting themes and create new CSS files.

But open a potential stream into the web just for a very very very very rare use case makes me a bad gut feeling

Rambomst:
@math-GH 
> but to honest: If you have hundrets/thousands of CSS lines added why do not develop a new own theme?

The stylesheet I am using is actually shared by a few applications so that's why I needed something like this over a custom theme.

> But open a potential stream into the web just for a very very very very rare use case makes me a bad gut feeling

The only time anything is changed is if a user sets a URL otherwise the CSP values remain unchanged, but I also get where you are coming from. Originally I just had this as a custom extension but when I opened a PR to add my extension to the list of third party extensions available I was told to do this instead and that PR there hasn't been merged.

Frenzie:
> It is very simple to use the CSS files from exisiting themes and create new CSS files.

That's talking from the perspective of a server administrator. From the perspective of a user that isn't very simple but completely impossible.

> But open a potential stream into the web just for a very very very very rare use case makes me a bad gut feeling

Imho the complete opposite of rare among the audience most likely to use FreshRSS.

Frenzie (approved):
I'd like to take this opportunity to vehemently object to making things worse for the purposes of "security".

Frenzie:
@Rambomst 
> The only time anything is changed is if a user sets a URL otherwise the CSP values remain unchanged, but I also get where you are coming from. Originally I just had this as a custom extension but when I opened a PR to add my extension to the list of third party extensions available I was told to do this instead and that PR there hasn't been merged.

Imo it should either be part of CustomCSS or part of the server config and it doesn't make sense as a separate extension, but my expectation would be that if you disagreed you would've said so rather than jumping straight into implementing it in CustomCSS. ;-)

PR #7108, opened 2024-12-15
Add ability to bypass image lazy loading for networks

Cyberax:
Eagerly-loaded images are great for the local network, it feels awesome when web comic pages can be opened within a fraction of a second.

However, eagerly-loaded images suck when you're on a phone with cellular data connectivity. The front page loading can take multiple seconds.

This change adds a way to disable lazy loading for certain IP ranges (your home network).

1. Enable image lazy loading, and test that images are indeed loaded lazily. The easiest way is to check the browser's network monitoring console.
2. Add your local network (e.g. 10.0.0.0/8 or 192.168.0.0/16) to the exclusion list (Settings -> Reading -> Misc).
3. Check that images are loaded eagerly.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

PS: I'm sorry, I'm not very familiar with PHP and my code might not be well-organized.


Alkarex:
Please check whether the following function could be used instead of the additional library:
https://github.com/FreshRSS/FreshRSS/blob/3b87372061b16d172899b29fe9b5414b9194abd4/lib/lib_rss.php#L690

Alkarex:
Another idea: I think this could also be done purely in JavaScript client side, and the preference could also be saved client side

Cyberax:
> Please check whether the following function could be used instead of the additional library

This indeed works! I reworked my patch to use it.

I tried to look into doing it purely on the JS side, but I don't think it's possible. We still need to pass the client's IP back to the JS somehow, so the code would end up the same.

Frenzie:
> Another idea: I think this could also be done purely in JavaScript client side, and the preference could also be saved client side

Having to set it up on each client individually seems like it'd somewhat defeat the purpose.

Cyberax:
> Having to set it up on each client individually seems like it'd somewhat defeat the purpose.

You need to set it up only once for your FreshRSS installation, by adding the CIDRs of your local networks (that shouldn't change). We can pre-populate it with RFC1918 networks, though.

Unfortunately, I don't really see an easy way to change this. FreshRSS server can't know all your local CIDRs, especially if it's started from Docker. And the browser client can't know that it's inside the FreshRSS' home network.

Frenzie:
> You need to set it up only once for your FreshRSS installation

Of course, that was a reply to the "purely in JavaScript" idea. You either need a server component or it'd need to be set up again in each and every single individual browser.

Cyberax:
> You either need a server component or it'd need to be set up again in each and every single individual browser.

I'm sorry, but you don't need to do that with my patch. It's set up only once, in the same config area where the lazy loading is set up. It will then work across any device within your home network.

Frenzie:
I know, that's a statement in support of the conceptual approach in this PR.

Frenzie:
Seeing the lack of enthusiasm, close this one, merge the extension?

Alkarex:
Related https://github.com/FreshRSS/FreshRSS/pull/5816 (help welcome)

> Seeing the lack of enthusiasm, close this one, merge the extension?

Yes, could be an approach. Is there an extension PR?

Frenzie:
D'oh… I accidentally commented on the wrong PR. I meant to say that about the CSP header one, not about this one. :facepalm:

math-GH:
> Seeing the lack of enthusiasm, close this one, merge the extension?

I agree.
I see this PR as a very very very special use case

PR #7144, opened 2024-12-26
Search in all feeds

### LINKED ISSUE #690: [feature] add a way to make the search on all feeds, all categories
At the moment, the search applies only where we are located. It would be nice to have a switch to a global search.
Maybe with the use of a keyword like "full:true" or "full:false" (default) in the search box.

### LINKED ISSUE #2131: Global search
When a search is performed from the freshrss mainstream, it seams the search is limited to only mainstream enabled feeds. Is it possible to perform a global search across all feeds including those grouped in a separate category and not displayed in the mainstream?

### LINKED ISSUE #6245: [Feature] Global Search Function is needed
## Current Situation

When searching for keywords, it seems like to search for feeds shown in the home page. I know we can add filter command to search for sepcific feeds. However, when the number of feeds is relatively large, it is unconvenient.

## Feature

I think that default search function should work globally with nothing set up which follows our intuition better.

Alkarex:
* Search in all feeds above `PRIORITY_MAIN_STREAM` with `&get=a` (we had this one already)
* Search in all feeds above `PRIORITY_CATEGORY` with `&get=A` (new)
* Search in all feeds above `PRIORITY_ARCHIVED` with `&get=Z` (new)

fix https://github.com/FreshRSS/FreshRSS/issues/690
fix https://github.com/FreshRSS/FreshRSS/issues/2131
fix https://github.com/FreshRSS/FreshRSS/issues/6245
fix https://github.com/FreshRSS/FreshRSS/discussions/7143

### DIFF
--- a/app/i18n/it/conf.php
+++ b/app/i18n/it/conf.php
@@ -138,7 +138,7 @@ return array(
 			'tags' => 'Tag',
 			'type' => 'Tipo',
 		),
-		'get_A' => 'Show all feeds, also those showed in their category',	// TODO
+		'get_A' => 'Show all feeds, also those shown in their category',	// TODO
 		'get_Z' => 'Show all feeds, also archived ones',	// TODO
 		'get_all' => 'Mostra tutti gli articoli',
 		'get_all_labels' => 'Mostra gli articoli con qualsiasi etichetta',
--- a/app/i18n/it/conf.php
+++ b/app/i18n/it/conf.php
@@ -138,6 +138,8 @@ return array(
 			'tags' => 'Tag',
 			'type' => 'Tipo',
 		),
+		'get_A' => 'Show all feeds, also those showed in their category',	// TODO
+		'get_Z' => 'Show all feeds, also archived ones',	// TODO
 		'get_all' => 'Mostra tutti gli articoli',
 		'get_all_labels' => 'Mostra gli articoli con qualsiasi etichetta',
 		'get_category' => 'Mostra la categoria “%s” ',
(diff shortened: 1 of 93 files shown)


Alkarex:
image

Alkarex:
@tao4mike

Alkarex:
image

app/i18n/it/conf.php
```diff
@@ -138,6 +138,8 @@
 			'tags' => 'Tag',
 			'type' => 'Tipo',
 		),
+		'get_A' => 'Show all feeds, also those shown in their category',	// TODO
+		'get_Z' => 'Show all feeds, also archived ones',	// TODO
```
UserRoot-Luca:
```suggestion
		'get_Z' => 'Mostra tutti i feed, anche quelli archiviati',
```

app/i18n/it/conf.php
```diff
@@ -138,6 +138,8 @@
 			'tags' => 'Tag',
 			'type' => 'Tipo',
 		),
+		'get_A' => 'Show all feeds, also those shown in their category',	// TODO
```
UserRoot-Luca:
```suggestion
		'get_A' => 'Mostra tutti i feed, anche quelli mostrati nella loro categoria',
```

math-GH:
Does the documentation need this information too?

Alkarex:
> Does the documentation need this information too?

It would be good, somewhere, probably on the page related to user queries, although it is quite explicit from the UI

Alkarex:
Let's merge, as I have some new features coming with overlapping code

Issue #4761, opened 2022-10-22
Sort order per feed


pniewiadowski:
**Describe the solution you’d like**
Currently we can choose sort order globally in Reading submenu, for all feeds we have in FreshRSS.
I'd like to be able to choose sort order per feed. So feed A would be order "Newest first" and feed B ordered "Older first"

Do you think it would be possible?

aledeg:
I think it would be possible to have such configuration and to apply it on a feed itself. However, when reading those feeds in an configuration other than a single feed (category, favorite, ...), it won't be possible to mix and match the sort orders.

pniewiadowski:
Would it then be good to have both, global and feed level sort option? Feed level when it is possible to sort that way, and global for other configuration like category, favorite and so on.

aledeg:
I think so. That's something I am thinking about for the longest time but for the view used to display the feed :).

aledeg:
As always, help welcomed :)

benfishbus:
This is the feature I've been looking for in FreshRSS. I thought surely it's in there somewhere, I'm just overlooking it. Normally my preferred sort order is Newest First. But for reading comics (either individually or as a category) I prefer them Oldest First.

Alkarex:
@benfishbus It is not a feature we have yet, but that would be a nice addition. Help with a PR welcome

PeterVavercak:
I would like to take the issue.

I feel like it would make sense add option of local sorting for idividual feeds when globally option of sorting by feed name is chosen.

Alkarex:
Work in progress:
* https://github.com/FreshRSS/FreshRSS/pull/8234

Tests welcome

Issue #4280, opened 2022-03-20
[Feature] Store a date and time in the DB to record when an entry was marked as read.


mdemoss:
I made an attempt at doing this myself, but wasn't able to understand the Minz framework well enough to work out how to modify the DB schema safely such that upgrades work properly.

I need this change for the [shuffle sort](https://github.com/FreshRSS/FreshRSS/pull/2463) I've been working on, but I could see it being useful for implementing other features as well. It could be used to build a _recently read_ page, to undo bulk _mark as read_ actions, and to produce more interesting statistics.

mdemoss:
If this feature becomes available I'll open another PR for my updated version of shuffle sort.

mdemoss:
Is there a previous instance of safely adding a column to the DB schema I could take a look at?

edit: the most recent addition of a column I found is `kind` https://github.com/Alkarex/FreshRSS/commit/1fe66ad020ca8f0560bb9c6e311852ed77228f78

Alwaysin:
I believe there are more steps needed to implement "undo bulk mark as read actions"?
I reckon it wasn't the main subject, only used as an exemple in the description.
Shall we open an issue for this?

Alkarex:
> I believe there are more steps needed to implement "undo bulk mark as read actions"? I reckon it wasn't the main subject, only used as an exemple in the description. Shall we open an issue for this?

Yes, please do. We only lack the filtering on the *user modified date*.

The process will be something like:
1. Click on the toggle button to show the read articles (making sure the toggle for the unread articles is off)
2. Sort by *User modified 9→1*
3. Filter by *user modified date*, for instance to the last 3 hours by typing [`userdate:PT3H`](https://github.com/FreshRSS/FreshRSS/pull/8093)
4. Click in the drop-down menu *Mark selection as unread*

Alkarex:
@Alwaysin Done in https://github.com/FreshRSS/FreshRSS/pull/8093
Tests welcome

PR #7149, opened 2024-12-27
Implement custom order-by

### LINKED ISSUE #161: Please add better sorting options for new feeds
I realize this issue will be closed as soon as I open it, but I think it's worth putting this back on the front burner even for a short time. 

I am a new FreshRSS user. I've spent the better part of this week playing with all of the options out there for self-hosted RSS feeds and FreshRSS is the one I keep coming back to. While I subscribe to Bazqux for my day-to-day feed reading I wanted something I could use to consolidate my YouTube subscriptions into a topic-based system where I and not the YouTube algorithm decide what gets put in front of me. 

The problem I'm having is the same one that's been raised by many others: adding a new feed to the mix brings all of the articles from that feed to the top of the list regardless of the date they were posted. It seems as though this is by design but this is not a good user experience and not what any of the alternative readers are doing with new feeds. 

I have, with mixed success, used some MySQL code to reorder the database but it's a pain to have to do that every time I add a new group of feeds. And my code is not always able to do this reliably. 

It would be great to give the user the choice about how to sort new feeds into the database. Some may like this behavior but I think there's an equal number who don't.

### LINKED ISSUE #1771: Streams do not sort by date correctly
Streams appear to sort by feed first, then by date, so instead of a mix of articles, you get clumps.

[See this screenshot.](https://i.imgur.com/qI4evfg.jpg)

### LINKED ISSUE #2083: Need help for changing the feed sort order from lastSeen to the publish date
Hello,

I am deploying a public feed aggregator and I will frequently add new feeds. The current sorting of the feed by lastSeen will look strange to the visitors.

I would like to change the sort order of the feed to be by the publishing date of the post (date) instead of by the date it was seen by the aggregator (lastSeen).

I searched the code and didn't find where to do that. Help will be appreciated.

Thanks.

### LINKED ISSUE #2119: Episodes from a newly added feed show up under the today category
I don't know how to better describe it. Perhaps this is per design. See screenshot below:

image

I have added the feed `VASAviation` today and even though there are episodes from last month all episodes appear in the today group.

### LINKED ISSUE #2596: How do I tell FreshRSS to sort articles by publication date, not date added?

### LINKED ISSUE #3204: Can I sort my feed by date of publication by default?
HI there, I am new to rss feeds and so on. I've found this app, hosted it on my server and added a couple of feeds. The app is great but I noticed that in main feed articles are not sorted by date of publication but by date of adding them to the app... is there a way to switch this behaviour to use date of publication?

screenshot

P.S. I've read the contributing guidelines and about your chat... but haven't succeeded to register there since I don't know the French and got there some errors I don't understand... So I decided to contact you here

### LINKED ISSUE #4405: [BUG] Items not keeping sort by date after a new subscription is added
**Describe the bug**
I have several subscriptions. Whenever I add a new one, it fetches all latest RSS entries in this feed.
But in the main stream, it only shows this latest subscription, instead of "mixing" its publications with the existing feeds.

**To Reproduce**
1. Have several different feeds
2. Add a new feed
3. Go to main stream and you'll only see the latest feed posts, despite some of them being older than existing ones on your main stream

**Expected behavior**
The entries in the main stream are sorted by published date, not by "subscription added date" (presumably, when FreshRSS fetched the article is the date it uses to sort by "newest").

**Screenshots**
image

Latest docker image at the time of writing.

**Additional context**
I am sorry if there is an option in the interface to sort this out, I just didn't find it :\

### LINKED ISSUE #5529: Please add better sorting options when adding new feeds
I realize this issue will be closed as soon as I open it, but I think it's worth putting this back on the front burner even for a short time.

I am a new FreshRSS user. I've spent the better part of this week playing with all of the options out there for self-hosted RSS feeds and FreshRSS is the one I keep coming back to. While I subscribe to Bazqux for my day-to-day feed reading I wanted something I could use to consolidate my YouTube subscriptions into a topic-based system where I and not the YouTube algorithm decide what gets put in front of me.

The problem I'm having is the same one that's been raised by many others: adding a new feed to the mix brings all of the articles from that feed to the top of the list regardless of the date they were posted. It seems as though this is by design but this is not a good user experience and not what any of the alternative readers are doing with new feeds.

I have, with mixed success, used some MySQL code to reorder the database but it's a pain to have to do that every time I add a new group of feeds. And my code is not always able to do this reliably.

It would be great to give the user the choice about how to sort new feeds into the database. Some may like this behavior but I think there's an equal number who don't.

### LINKED ISSUE #5864: Problem with filtering by date
I installed today freshRSS and added my feeds, but I'm facing this problem: the feeds are first grouped and then sorted by date.

E.g.:

> feed 1 - title - 11 nov 2023
> feed 1 - title - 10 nov 2023
> feed 1 - title - 09 nov 2023
> feed 1 - title - 08 nov 2023
> feed 2 - title - 11 nov 2023
> feed 2 - title - 10 nov 2023
> feed 2 - title - 09 nov 2023
> feed 2 - title - 08 nov 2023
> and so on...

What I want is to have all in cronological order:

> feed 1 - title - 11 nov 2023
> feed 2 - title - 11 nov 2023
> feed 1 - title - 10 nov 2023
> feed 2 - title - 10 nov 2023
> feed 1 - title - 09 nov 2023
> feed 2 - title - 09 nov 2023
> feed 1 - title - 08 nov 2023
> feed 2 - title - 08 nov 2023

If I click the "older/newest first" button the situation doesn't change, feeds are inverted, but always grouped.

[Here is a short video recording of the issue](https://screenrec.com/share/HGYvbo5hys).

Am I missing something?

### LINKED ISSUE #7239: Feeds are not ordered by date
### Describe the bug

Image

The feeds are not being displayed in order with the most recent one in top. Instead, they are completely mixed up. Sometimes the feeds of a certain blog are grouped (independently of when they were published). 

### To Reproduce

No need to do anything. This is apparently the standard setting and I cannot find a way to change it.

### Expected behavior

I was expecting feeds to be ordered purely by date.

### FreshRSS version

1.25.0

### System information

- Database version: freshrss native (SQLite?)
- PHP version: unknown
- Installation type: Coolify
-Web server type: Ubuntu
- Device: Macbook pro
- OS: macOS Sequoia 15.2
- Browser: Brave 1.74.48

### Additional context

_No response_

Alkarex:
Add option to sort results by received date (existing, default), publication date, title, URL (link), random.

fix https://github.com/FreshRSS/FreshRSS/issues/1771
fix https://github.com/FreshRSS/FreshRSS/issues/2083
fix https://github.com/FreshRSS/FreshRSS/issues/2119
fix https://github.com/FreshRSS/FreshRSS/issues/2596
fix https://github.com/FreshRSS/FreshRSS/issues/3204
fix https://github.com/FreshRSS/FreshRSS/issues/4405
fix https://github.com/FreshRSS/FreshRSS/issues/5529
fix https://github.com/FreshRSS/FreshRSS/issues/5864
fix https://github.com/FreshRSS/Extensions/issues/161

URL parameters:
* `&sort=id` (current behaviour, sorting according to newest received articles)
* `&sort=date` (publication date, which is not indicative of how new an article is)
* `&sort=title`
* `&sort=link`
* `&sort=rand` (random order - which disables infinite scrolling, at least for now)

combined with `&order=ASC` or `&order=DESC`

image

## Implementation notes

The sorting criteria by *received date* (id), which is the default, and which was the only one before this PR, is the one that has the best sorting characteristics:
* *uniqueness*: no entries have the exact same received date
* *monotonicity*: new entries always have a higher received date
* *performance*: this field is efficiently indexed in database for fast usage, including for paging (indexing could also be done to other fields, but with lower effective performance)

In contrary, sorting criteria such as by *publication date*, by *title*, or by *link* are neither unique nor monotonic. In particular, multiple articles may share the same *publication date*, and we may receive articles with a *publication date* far in the future, and then later some new articles with a *publication date* far in the past.

To understand why sorting by *publication date* is problematic, it helps to think about sorting by *title* or by *link*, as sorting by *title* and by *publication date* share more or less the same characteristics.

### Problem 1: new articles

New articles may be received in the background after what is shown on screen, and before the next user action such as *mark all as read*. Due to the lack of *monotonicity* when sorting by e.g. *publication date* or *title*, users risk marking as read a batch of articles containing some fresh articles without seeing them.

Mitigation: A parameter `idMax` tracks the maximum ID related to a batch of actions such as *mark all as read* to exclude articles received after those that are displayed.

### Problem 2: paging / pagination

When navigating articles, only a few articles are displayed, and a new "page" of articles needs to be received from the database when scrolling down or when clicking the button to show more articles. When sorting by e.g. *publication date* or *title*, it is not trivial to show the next page without re-showing some of the same articles, and without skipping any. Indeed, views are often with additional criteria such as showing only unread articles, and users may mark some articles as read while viewing them, hereby removing some articles from the previous pages. And like for *Problem 1*, new articles may have been received in the background. Consequently, it is not possible to use `OFFSET` to implement pagination (so the patches suggested by a few users were wrong due to that, in particular).

Mitigation: `idMax` is also used (just like for *Problem 1*) and a *Keyset Pagination* approach is used, combining an unstable sorting criterion such as *publication date* or *title*, together with *id* to ensure stable sorting. (So, 2 sorting criteria + 1 filter criteria)

See e.g. https://www.alwaysdeveloping.net/dailydrop/2022/07/01-keyset-pagination/

### Problem 3: performance

Sorting by anything else than *received date* (id) is doomed to be slow(er) due to the combination of 3 criteria (see *Problem 2*). An `OFFSET` approach (which is not possible anyway as explained) would be even slower. Furthermore, we have no SQL index at the moment, but they would not necessarily help much due to the multiple sorting criteria needed and involving some `OR` logic which is difficult to optimise for databases.

The nicest syntax would be using tuples and corresponding indexes, but that is poorly supported by MySQL https://bugs.mysql.com/bug.php?id=104128

Mitigation: a compatibility SQL syntax is used to implement *Keyset Pagination*

### Problem 4: user confusion

Several users have shown that they do not fully understand the difference between *received date* and *publication date*, and particularly not the pitfalls of *publication date*.

Mitigation: the menus to mark-as-read *before 1 day* and *before 1 week* are disabled when sorting by anything else than *received date*. Likewise, the separation headers *Today* and *Yesterday* and *Before yesterday* are only shown when sorting by *received date*.

Again here, to better understand why, it helps to think about sorting by *title* or by *link*, as sorting by *title* and by *publication date* share more or less the same characteristics.

* [ ] We should write a Q&A and/or documentation about the problems associated to *sorting by publication date*: risks of not noticing new publication, of inadvertently marking them as read, of having some articles with a date in the future hanging at the top of the views (vice versa when sorting in ascending order), performance, etc.

### Problem 5: APIs

Sorting by anything else than *received date* breaks the guarantees needed for a successful synchronisation via API.

Mitigation: sorting by *received date* is ensured for all API calls.


app/Models/EntryDAO.php
```diff
@@ -1146,21 +1150,42 @@ protected function sqlListEntriesWhere(string $alias = '', ?FreshRSS_BooleanSear
...
+		if ($id_min !== '0') {
 			$search .= 'AND ' . $alias . 'id >= ? ';
-			$values[] = $date_min . '000000';
+			$values[] = $id_min;
+		}
+
+		if ($continuation_id !== '0' && in_array($sort, ['date', 'link', 'title'], true)) {
+			$sign = $order === 'ASC' ? '>' : '<';
+			// Keyset pagination (Compatibility syntax due to poor performance of tuple syntax in MySQL https://bugs.mysql.com/bug.php?id=104128)
+			$search .= "AND ({$alias}{$sort} {$sign} ? OR ({$alias}{$sort} = ? AND {$alias}id {$sign}= ?)) ";
```
Alkarex, line 1185: This is the most important line of this PR

Alkarex:
Back-end and UI done

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
 		'reader_view' => 'Modalità di lettura',
 		'rss_view' => 'Feed RSS',
 		'search_short' => 'Cerca',
+		'sort' => array(
+			'_' => 'Sorting criteria',	// TODO
```
UserRoot-Luca: @Alkarex I'm sorry but I don't know how to translate it into Italian

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'date_asc' => 'Publication date 1→9',	// TODO
```
UserRoot-Luca:
```suggestion
			'date_asc' => 'Data di pubblicazione 1→9',
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'date_desc' => 'Publication date 9→1',	// TODO
```
UserRoot-Luca:
```suggestion
			'date_desc' => 'Data di pubblicazione 9→1',
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'id_asc' => 'Freshly received last',	// TODO
```
UserRoot-Luca:
```suggestion
			'id_asc' => 'Dal meno recente',
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'id_desc' => 'Freshly received first',	// TODO
```
UserRoot-Luca:
```suggestion
			'id_desc' => 'Dal più recente',
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'link_asc' => 'Link A→Z',	// TODO
```
UserRoot-Luca:
```suggestion
			'link_asc' => 'Link A→Z',	// IGNORE
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'link_desc' => 'Link Z→A',	// TODO
```
UserRoot-Luca:
```suggestion
			'link_desc' => 'Link Z→A',	// IGNORE
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'rand' => 'Random order',	// TODO
```
UserRoot-Luca:
```suggestion
			'rand' => 'Ordine casuale',
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'title_asc' => 'Title A→Z',	// TODO
```
UserRoot-Luca:
```suggestion
			'title_asc' => 'Titolo A→Z',	
```

app/i18n/it/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'title_desc' => 'Title Z→A',	// TODO
```
UserRoot-Luca:
```suggestion
			'title_desc' => 'Titolo Z→A',
```

app/i18n/it/index.php
Frenzie: `Sorteercriteria` is `critères de tri(age)` in French. A quick search suggests LibreOffice translates that to `Criteri di ordinamento`.

app/i18n/nl/index.php
```diff
@@ -60,6 +60,18 @@
...
+			'_' => 'Sorting criteria',	// TODO
+			'date_asc' => 'Publication date 1→9',	// TODO
+			'date_desc' => 'Publication date 9→1',	// TODO
+			'id_asc' => 'Freshly received last',	// TODO
+			'id_desc' => 'Freshly received first',	// TODO
+			'link_asc' => 'Link A→Z',	// TODO
+			'link_desc' => 'Link Z→A',	// TODO
+			'rand' => 'Random order',	// TODO
+			'title_asc' => 'Title A→Z',	// TODO
+			'title_desc' => 'Title Z→A',	// TODO
```
Frenzie:
```suggestion
			'_' => 'Sorteercriteria',
			'date_asc' => 'Publicatiedatum 1→9',
			'date_desc' => 'Publicatiedatum 9→1',
			'id_asc' => 'Nieuw ontvangen laatst',
			'id_desc' => 'Nieuw ontvangen eerst',
			'link_asc' => 'Link A→Z',	// IGNORE
			'link_desc' => 'Link Z→A',	// IGNORE
			'rand' => 'Willekeurige volgorde',
			'title_asc' => 'Titel A→Z',
			'title_desc' => 'Titel Z→A',
```

app/i18n/it/index.php
UserRoot-Luca: @Frenzie I also looked it up but the translation is not convincing to me because the term `ordinamento` is a term that in Italian is used to refer to a set of laws

Frenzie: I couldn't say. What does Windows (or whichever OS) use for the `Sort by` menu if you right click in a file manager, `Ordina per`?

image

Alkarex:
I have added some [implementation notes](https://github.com/FreshRSS/FreshRSS/pull/7149#issue-2760772548) higher up

app/i18n/it/index.php
UserRoot-Luca:
```suggestion
			'_' => 'Ordina per',
```
@Frenzie Thank you I hadn't thought of translating it that way

@Alkarex Thank you for correcting my error in the code I edited earlier

app/i18n/fr/index.php
```diff
@@ -60,6 +60,18 @@
 		'reader_view' => 'Vue lecture',
 		'rss_view' => 'Flux RSS',
 		'search_short' => 'Rechercher',
+		'sort' => array(
+			'_' => 'Critère de tri',
+			'date_asc' => 'Date de publication 1→9',
+			'date_desc' => 'Date de publication 9→1',
+			'id_asc' => 'Reçus récemment en premier',
+			'id_desc' => 'Reçus récemment en dernier',
```
Alkarex:
```suggestion
			'id_desc' => 'Reçus récemment en premier',
```

app/i18n/fr/index.php
```diff
@@ -60,6 +60,18 @@
 		'reader_view' => 'Vue lecture',
 		'rss_view' => 'Flux RSS',
 		'search_short' => 'Rechercher',
+		'sort' => array(
+			'_' => 'Critère de tri',
+			'date_asc' => 'Date de publication 1→9',
+			'date_desc' => 'Date de publication 9→1',
+			'id_asc' => 'Reçus récemment en premier',
```
Alkarex:
```suggestion
			'id_asc' => 'Reçus récemment en dernier',
```

ekr1:
Awesome work, @Alkarex - I just went down the rabbit hole of the copious discussions asking for this and found out that the rather large change is almost ready and you're spending what looks like your holidays on this. Thanks for your efforts! 👍

Alkarex:
I have changed a bit the title of the separation headers to reduce ambiguity:

image

(There used to be only "Today")

image

app/i18n/it/index.php
```diff
@@ -27,6 +27,11 @@
 	),
 	'feed' => array(
 		'empty' => 'Non ci sono articoli da mostrare.',
+		'received' => array(
+			'before_yesterday' => 'Received before yesterday',	// TODO
```
UserRoot-Luca:
```suggestion
			'before_yesterday' => 'Ricevuto prima di ieri',
```
I translated trying to keep the sense of the original text, but in Italian it makes more sense to keep the actual sentence `Meno recenti`

@Alkarex This translation is fine if it is also used for days perecedent to the day before yesterday
In case it is used only for a single day it is better to use this translation `Ricevuto l’altro ieri`

app/i18n/it/index.php
```diff
@@ -27,6 +27,11 @@
...
+			'today' => 'Received today',	// TODO
```
UserRoot-Luca:
```suggestion
			'today' => 'Ricevuto oggi',
```

app/i18n/it/index.php
```diff
@@ -27,6 +27,11 @@
...
+			'yesterday' => 'Received yesterday',	// TODO
```
UserRoot-Luca:
```suggestion
			'yesterday' => 'Ricevuto ieri',
```

app/i18n/it/index.php
Alkarex: it is also used for days preceding the day before yesterday

app/i18n/it/index.php
```diff
@@ -27,6 +27,11 @@
 	),
 	'feed' => array(
 		'empty' => 'Non ci sono articoli da mostrare.',
+		'received' => array(
+			'before_yesterday' => 'Ricevuto prima di ieri',
```
UserRoot-Luca, line 31:
```suggestion
			'before_yesterday' => 'Meno recenti',
```
@Alkarex Thank you for the answer to my question then it is more correct to use the old phrase because in Italian it is not very common to use this `Ricevuto prima di ieri`

I used this translation `Ricevuto prima di ieri` first because I was not sure of the context now that I know I can make a more “correct” translation

app/layout/nav_menu.phtml
```diff
@@ -212,22 +213,43 @@
...
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'DESC' && FreshRSS_Context::$sort === 'date' ? 'true' : 'false' ?>">
+					<a href="<?= Minz_Url::display($url_order, amend: ['params' => ['sort' => 'date', 'order' => 'DESC']]) ?>"><?= _t('index.menu.sort.date_desc') ?></a></li>
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'DESC' && FreshRSS_Context::$sort === 'link' ? 'true' : 'false' ?>">
+					<a href="<?= Minz_Url::display($url_order, amend: ['params' => ['sort' => 'link', 'order' => 'DESC']]) ?>"><?= _t('index.menu.sort.link_desc') ?></a></li>
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'DESC' && FreshRSS_Context::$sort === 'title' ? 'true' : 'false' ?>">
+					<a href="<?= Minz_Url::display($url_order, amend: ['params' => ['sort' => 'title', 'order' => 'DESC']]) ?>"><?= _t('index.menu.sort.title_desc') ?></a></li>
+				<hr />
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'ASC' && FreshRSS_Context::$sort === 'id' ? 'true' : 'false' ?>">
```
math-GH:
```suggestion
				<li class="item separator" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'ASC' && FreshRSS_Context::$sort === 'id' ? 'true' : 'false' ?>">
```

app/layout/nav_menu.phtml
```diff
@@ -212,22 +213,43 @@
...
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'ASC' && FreshRSS_Context::$sort === 'date' ? 'true' : 'false' ?>">
+					<a href="<?= Minz_Url::display($url_order, amend: ['params' => ['sort' => 'date', 'order' => 'ASC']]) ?>"><?= _t('index.menu.sort.date_asc') ?></a></li>
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'ASC' && FreshRSS_Context::$sort === 'link' ? 'true' : 'false' ?>">
+					<a href="<?= Minz_Url::display($url_order, amend: ['params' => ['sort' => 'link', 'order' => 'ASC']]) ?>"><?= _t('index.menu.sort.link_asc') ?></a></li>
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$order === 'ASC' && FreshRSS_Context::$sort === 'title' ? 'true' : 'false' ?>">
+					<a href="<?= Minz_Url::display($url_order, amend: ['params' => ['sort' => 'title', 'order' => 'ASC']]) ?>"><?= _t('index.menu.sort.title_asc') ?></a></li>
+				<hr />
+				<li class="item" role="radio" aria-checked="<?= FreshRSS_Context::$sort === 'rand' ? 'true' : 'false' ?>">
```
math-GH:
```suggestion
				<li class="item separator" role="radio" aria-checked="<?= FreshRSS_Context::$sort === 'rand' ? 'true' : 'false' ?>">
```

app/i18n/it/index.php
Alkarex: I suggest we stay with the more literal translation for now. We may revise those headers later (I think we have a ticket somewhere)

Fayyaadh:
> Add option to sort results by received date (existing, default), publication date, title, URL (link), random.
> 
> fix #1771 fix #2083 fix #2119 fix #2596 fix #3204 fix #4405 fix #5529 fix #5864 fix [FreshRSS/Extensions#161](https://github.com/FreshRSS/Extensions/issues/161)
> 
> URL parameters:
> 
> * `&sort=id` (current behaviour, sorting according to newest received articles)
> * `&sort=date` (publication date, which is not indicative of how new an article is)
> * `&sort=title`
> * `&sort=link`
> * `&sort=rand` (random order - which disables infinite scrolling, at least for now)
> 
> combined with `&order=ASC` or `&order=DESC`
> 

Hi there

Thank you so much for all your work on this.

Can you perhaps elaborate a bit on how to use this?

I went to the main feed page and added the query parameter for `&sort=date` to the URL but nothing happened.

Is the feature not yet released in the latest Docker image?

Thanks for the help.

Alkarex:
@Fayyaadh What Docker image are you running, more precisely?
Make sure you are using our rolling release and are using one of our official images such as `freshrss/freshrss:edge` or `freshrss/freshrss:alpine-edge`

Fayyaadh:
> @Fayyaadh What Docker image are you running, more precisely? Make sure you are using our rolling release and are using one of our official images such as `freshrss/freshrss:edge` or `freshrss/freshrss:alpine-edge`

My docker compose file just has `freshrss/freshrss` without any tag.

I guess I'll need to change that?

Frenzie:
Yes, this is only in edge.

Fayyaadh:
Ok I pulled the new docker image and the feature now works!

Is there any way to get it to persist though?

So if I click away from the main feed and then go back, the query parameter `&sort=date` is no longer there.

mdemoss:
Possibly this either complements or obviates https://github.com/FreshRSS/FreshRSS/issues/4280 which I requested for my implementation of a random order-by which I called "shuffle."

To implement a random order, do pagination, and not have items "jump around" I think it may be necessary to track which items have been marked read since the view of items was requested. Recording the time when an item is marked read may still be the best way to do that.

Alkarex:
> which I requested for my implementation of a random order-by which I called "shuffle."

Quick idea to do that: you can also define a hash function (e.g. `md5()`) with a salt, apply it to the items ID, and you will have a new, stable, randomized order. If you want to reshuffle, just change the salt.

mdemoss:
In my first implementation of the concept I used a cryptographic hash (sha1) but in the [second, current implementation](https://github.com/mdemoss/FreshRSS/blob/beeac788d56e440735b1751daca193e0c796089f/app/Models/EntryDAO.php#L929) I use a "multiplicative hash." 

Should be faster and more portable SQL.

CRYPTO-GENIK:
I just wanted to chime in with a huge thank you to everyone that helped put this feature together!
It absolutely should be the default setting to sort by publication date and this is a huge feature to the app so 
Again -- Thank you! Thank you!

Let's get persistence of last sort or a default config option added and pushed out to everyone!

brevilo:
> Let's get persistence of last sort or a default config option added and pushed out to everyone!

This! Is there a separate issue to track?

Other than that, a huge thank you - as always.

shadowlocked:
Is there any existing way to address the default setting in config? For journalists and researchers, anything other than sorting by publication date is not a good default.

Alkarex:
When possible, please either open a new question or use an existing open issue or question.

> Is there any existing way to address the default setting in config? 

@shadowlocked Yes there is, and more than one way. Since this PR, by changing the corresponding preference in your config file. Alternatively, thanks to the sorting criteria being kept while you navigate between articles / feeds / categories, you can just change once to your favourite sorting criteria. This means that you can even bookmark your FreshRSS by including the sorting criteria in the URL.

See:
* https://github.com/FreshRSS/FreshRSS/issues/8085#issuecomment-3392435306
* https://github.com/FreshRSS/FreshRSS/issues/4761

For the first approach, check the sort order in your `data/users/shadowlocked/config.php` on the model of:

https://github.com/FreshRSS/FreshRSS/blob/ec1f5ee61bfeb9e8ed6a3c1e069b82d9f26f64e6/config-user.default.php#L56

> For journalists and researchers, anything other than sorting by publication date is not a good default.

*Publication date* is highly misunderstood by many users, which is also a reason why it has not been promoted more. Articles may have publication dates far in the past or in the future (or no date at all), either for legitimate reasons or due to source bugs. A consequence is that the sorting view is not stable when using *publication date*, since new articles will often be inserted in the middle of existing articles, which is quite confusing. This makes it a poor and error-prone default.
The only stable and reliable criteria for sorting by freshness is the received date, which very often will be quite similar to the publication date.

But with that been said, yes, the default sorting criteria can nevertheless already be changed, and there are issues open to contributors for additional UI and options.

Alkarex:
Work in progress:
* https://github.com/FreshRSS/FreshRSS/pull/8234

Tests welcome

PR #7161, opened 2024-12-29
New: About page with env information

math-GH:
Added a box with environment information.

It should help for writing a bug report with environment information.

Information are available only for admins (to be discussed!)

grafik

1. be an admin
2. go to `about` page
3. click on the the `environment information` "link" to open the information box

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


math-GH:
What do you think? Are this information sensitive so that only admins should see them?

app/views/index/about.phtml
```diff
@@ -24,6 +24,24 @@
...
+			<summary><?= _t('index.about.bug_reports.environment_information') ?></summary>
+			<dl>
+				<dt><?= _t('index.about.bug_reports.environment_information.version_frss') ?></dt>
+				<dd><?= FRESHRSS_VERSION ?></dd>
+				<dt><?= _t('index.about.bug_reports.environment_information.version_php') ?></dt>
+				<dd><?= PHP_VERSION ?></dd>
+				<dt><?= _t('index.about.bug_reports.environment_information.database') ?></dt>
+				<dd><?= FreshRSS_Context::systemConf()->db['type'] ?></dd>
+				<dt><?= _t('index.about.bug_reports.environment_information.server_software') ?></dt>
+				<dd><?= $_SERVER['SERVER_SOFTWARE'] ?></dd>
```
math-GH: From PHPStan:
`Parameter #1 (mixed) of echo cannot be converted to string.`

I do not have any idea how to solve it correctly

app/views/index/about.phtml
```diff
@@ -24,6 +24,24 @@
...
+				<dt><?= _t('index.about.bug_reports.environment_information.browser') ?></dt>
+				<dd><?= $_SERVER['HTTP_USER_AGENT'] ?></dd>
```
math-GH: From PHPStan:
`Parameter #1 (mixed) of echo cannot be converted to string.`

I do not have any idea how to solve it correctly

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
...
 		'_' => 'Informazioni',
 		'agpl3' => '<a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL 3</a>',	// IGNORE
+		'bug_reports' => array(
+			'environment_information' => array(
+				'_' => 'Environment information',	// TODO
+				'browser' => 'Browser',	// TODO
+				'database' => 'Database',	// TODO
+				'server_software' => 'Server software',	// TODO
+				'version_frss' => 'FreshRSS version',	// TODO
+				'version_php' => 'PHP version',	// TODO
```
UserRoot-Luca:
```suggestion
				'version_php' => 'Versione PHP',
```

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
 	'about' => array(
 		'_' => 'Informazioni',
 		'agpl3' => '<a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL 3</a>',	// IGNORE
+		'bug_reports' => array(
+			'environment_information' => array(
+				'_' => 'Environment information',	// TODO
+				'browser' => 'Browser',	// TODO
+				'database' => 'Database',	// TODO
+				'server_software' => 'Server software',	// TODO
+				'version_frss' => 'FreshRSS version',	// TODO
```
UserRoot-Luca:
```suggestion
				'version_frss' => 'Versione FreshRSS',
```

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
 	'about' => array(
 		'_' => 'Informazioni',
 		'agpl3' => '<a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL 3</a>',	// IGNORE
+		'bug_reports' => array(
+			'environment_information' => array(
+				'_' => 'Environment information',	// TODO
+				'browser' => 'Browser',	// TODO
+				'database' => 'Database',	// TODO
+				'server_software' => 'Server software',	// TODO
```
UserRoot-Luca:
```suggestion
				'server_software' => 'Software server',
```

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
 	'about' => array(
 		'_' => 'Informazioni',
 		'agpl3' => '<a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL 3</a>',	// IGNORE
+		'bug_reports' => array(
+			'environment_information' => array(
+				'_' => 'Environment information',	// TODO
+				'browser' => 'Browser',	// TODO
+				'database' => 'Database',	// TODO
```
UserRoot-Luca:
```suggestion
				'database' => 'Database',	// IGNORE
```

Frenzie:
> What do you think? Are this information sensitive so that only admins should see them?

It's probably better to err on the side of caution.

app/views/index/about.phtml
Frenzie: Maybe it'll like this?
```suggestion
				<dd><?= $_SERVER['SERVER_SOFTWARE'] ?? '' ?></dd>
```

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
 	'about' => array(
 		'_' => 'Informazioni',
 		'agpl3' => '<a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL 3</a>',	// IGNORE
+		'bug_reports' => array(
+			'environment_information' => array(
+				'_' => 'Environment information',	// TODO
```
UserRoot-Luca:
```suggestion
				'_' => 'Informazioni sul sistema',
```
@math-GH I had to change the sentence a little bit to keep the sense and translated it this way `System information`

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
...
+				'browser' => 'Browser',	// TODO
```
UserRoot-Luca:
```suggestion
				'browser' => 'Browser',	// IGNORE
```

Frenzie:
> It should help for writing a bug report with environment information.

There's a strong overlap with installation check btw.

app/i18n/it/index.php
math-GH: @Frenzie : What do you think? What is a better choice for the English term? `Environment information` or `System information`?

math-GH:
> > It should help for writing a bug report with environment information.
> 
> There's a strong overlap with installation check btw.

Good point! The idea is to have a copy&paste ready to go solution. That is why the box is placed directly under the Github link in the about page

UserRoot-Luca:
> What do you think? Are this information sensitive so that only admins should see them?

The only non-sensitive information is the version of FreshRSS and maybe even the version of PHP

UserRoot-Luca:
> > > It should help for writing a bug report with environment information.
> > 
> > 
> > There's a strong overlap with installation check btw.
> 
> Good point! The idea is to have a copy&paste ready to go solution. That is why the box is placed directly under the Github link in the about page

The problem is that by leaving the information on the server accessible to everyone you risk providing easy information for “Hackers”

app/views/index/about.phtml
Alkarex:
```suggestion
				<dd><?= is_string($_SERVER['SERVER_SOFTWARE'] ?? null) ? $_SERVER['SERVER_SOFTWARE'] : '' ?></dd>
```

Alkarex: This would only help for `string|null` but not for mixed, which cannot always be casted to a string

app/views/index/about.phtml
Alkarex:
```suggestion
				<dd><?= is_string($_SERVER['HTTP_USER_AGENT'] ?? null) ? $_SERVER['HTTP_USER_AGENT'] : '' ?></dd>
```

app/views/index/about.phtml
Frenzie: Why would it be mixed? I think phpstan just takes everything in $_SERVER as mixed which isn't accurate factually speaking afaik. Some are strings, some are arrays, but string values won't randomly be arrays.

app/i18n/it/index.php
Frenzie: I'd go with system information since it's widely used for the purpose. 

Incidentally, I'd add `php_uname()`, probably limited to s, r, v, and m (because hostname is a bit sensitive). There've been too many arm-exclusive issues not to.

```php
php_uname('s') . ' ' . php_uname('r') . ' ' . php_uname('v') . ' ' . php_uname('m')
```

https://www.php.net/manual/en/function.php-uname.php

app/views/index/about.phtml
Alkarex: @Frenzie I agree. I believe PHPStan's reasoning is that upon arriving at this line, nothing prevents some code from before to have called something like `$_SERVER['SERVER_SOFTWARE'] = [5 => 'Bad idea', 2 => 3];`
But I would rather like preventing manual assignments to `$_SERVER` and keep the string guarantee. Any idea on how to do that would be welcome.

Alkarex: * https://github.com/phpstan/phpstan/issues/8497
* https://github.com/phpstan/phpstan/issues/10870
* https://github.com/phpstan/phpstan/issues/12252

Alkarex: If we can define an ignore rule that works well, that would be fine with me. Help welcome

math-GH: checks are green now. Thanks :)

app/i18n/it/index.php
math-GH: > Incidentally, I'd add `php_uname()`, probably limited to s, r, v, and m (because hostname is a bit sensitive). There've been too many arm-exclusive issues not to.
> 
> ```
> php_uname('s') . ' ' . php_uname('r') . ' ' . php_uname('v') . ' ' . php_uname('m')
> ```
> 
> https://www.php.net/manual/en/function.php-uname.php

Do you mean it is better than `$_SERVER['SERVER_SOFTWARE']`?

math-GH:
> Incidentally, I'd add `php_uname()`, probably limited to s, r, v, and m (because hostname is a bit sensitive). There've been too many arm-exclusive issues not to.
> 
> ```
> php_uname('s') . ' ' . php_uname('r') . ' ' . php_uname('v') . ' ' . php_uname('m')
> ```
> 
> https://www.php.net/manual/en/function.php-uname.php

Do you mean it is better than `$_SERVER['SERVER_SOFTWARE']`?

app/i18n/it/index.php
```diff
@@ -14,6 +14,16 @@
 	'about' => array(
 		'_' => 'Informazioni',
 		'agpl3' => '<a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPL 3</a>',	// IGNORE
+		'bug_reports' => array(
+			'environment_information' => array(
+				'_' => 'System information',	// TODO
```
UserRoot-Luca:
```suggestion
				'_' => 'Informazioni sul sistema',
```

Frenzie:
> Do you mean it is better than $_SERVER['SERVER_SOFTWARE']?

Neither better nor worse but completely different.

```
 mode is a single character that defines what information is returned:

    'a': This is the default. Returns the same information as the individual modes 's', 'n', 'r', 'v', 'm' separated by spaces.
    's': Operating system name. eg. FreeBSD.
    'n': Host name. eg. localhost.example.com.
    'r': Release name. eg. 5.1.2-RELEASE.
    'v': Version information. Varies a lot between operating systems.
    'm': Machine type. eg. i386.
```

Alkarex:
Added database information:

image

PR #7190, opened 2025-01-07
Feature/http response code

ColonelMoutarde:
- http code is now replaced by Enum

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/Models/Log.php
+++ b/app/Models/Log.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 class FreshRSS_Log extends Minz_Model {
 
 	private string $date;
+	/** @property 'error'|'warning'|'notice'|'debug'|'info' $level */
 	private string $level;
 	private string $information;
 
@@ -20,6 +21,10 @@ class FreshRSS_Log extends Minz_Model {
 		$this->date = $date;
 	}
 	public function _level(string $level): void {
+		if (!in_array($level, ['error', 'warning', 'notice', 'debug', 'info'], true)) {
+			$this->level = 'info';
+			return;
+		}
 		$this->level = $level;
 	}
 	public function _info(string $information): void {
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -56,7 +56,7 @@ class Minz_Log {
 					$level_label = 'info';
 			}
 
-			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . $information . "\n";
+			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . str_replace(["\r", "\n"], ' ', $information) . "\n";
 
 			if (defined('COPY_LOG_TO_SYSLOG') && COPY_LOG_TO_SYSLOG) {
 				syslog($level, '[' . $username . '] ' . trim($log));


ColonelMoutarde:
Hello,
Can you add my PR to the 1.26.0 release?

lib/Minz/Error.php
```diff
@@ -4,30 +4,32 @@
...
+	 * Launches an error
+	 * @param FreshRSS_HttpResponseCode $code error type
+	 * @param string|array<'error'|'warning'|'notice',array<string>> $logs error logs broken down by form
+	 *      > $logs['error']
+	 *      > $logs['warning']
+	 *      > $logs['notice']
+	 * @param bool $redirect indicates whether to force redirection (logs will not be transmitted)
+	 * @throws Minz_ConfigurationException
+	 */
+	public static function error(FreshRSS_HttpResponseCode $code, string|array $logs = [], bool $redirect = true): void {
```
Frenzie: This looks like a breaking change. Is that intentional? Based on the old code I'd expect to see something like:

```suggestion
	public static function error(FreshRSS_HttpResponseCode $code = FreshRSS_HttpResponseCode::HTTP_404_NOT_FOUND, string|array $logs = [], bool $redirect = true): void {
```

lib/Minz/Error.php
```diff
@@ -4,30 +4,32 @@
...
  * Sous licence AGPL3 <http://www.gnu.org/licenses/>
-*/
+ */
 
 /**
  * The Minz_Error class logs and raises framework errors
  */
-class Minz_Error {
+class Minz_Error
+{
```
Frenzie: A bit random, none of the other files around here do that.

Frenzie (approved):
lgtm in principle ^_^

app/Controllers/authController.php
```diff
@@ -120,7 +120,7 @@ public function formLoginAction(): void {
 
 			if ($nonce === '') {
 				Minz_Log::warning("Invalid session during login for user={$username}, nonce={$nonce}");
-				header('HTTP/1.1 403 Forbidden');
+				header(FreshRSS_HttpResponseCode::descriptionFromCode(FreshRSS_HttpResponseCode::HTTP_403_FORBIDDEN));
```
Alkarex: This does not look nicer to me. I can better like the string, which is a well-known standard, but even with the code version, I think we might be able to do it better

ColonelMoutarde: I've improved the code to make it less wordy

ColonelMoutarde:
Hi, @Alkarex,
so this PR will be merged? I've been updating it for a long time. I'd like it to be present, it makes the code easier to read :D

Have a good day !

lib/http-conditional.php
```diff
@@ -142,13 +142,13 @@ function httpConditional(int $UnixTimeStamp, int $cacheSeconds = 0, int $cachePr
 	}
 
 	if ($is412) { //rfc2616-sec10.html#sec10.4.13
-		header('HTTP/1.1 412 Precondition Failed');
+		header(FreshRSS_HttpResponse::description(FreshRSS_HttpResponse::HTTP_412_PRECONDITION_FAILED));
```
Alkarex: This file is a standalone library, which does not know about FreshRSS_HttpResponse...

Alkarex:
> Hi, @Alkarex, so this PR will be merged? I've been updating it for a long time. I'd like it to be present, it makes the code easier to read :D
> 
> Have a good day !

Sorry for the delay. A part of the reason is that I am not sure what to think 🤔
To me, it makes the code harder to read, not easier, and with a learning curve instead of a well-known standard. Example:

image

So I was hopping to find time to come up with a better syntax than the above, but I have not come to it yet.

So as it is now, it adds about 100 lines of code for no clear value, or maybe even negative value.

Ideally, I would be pleased with some constants at PHP level, but even there, I am not sure it is really worth it.

There are also some errors (see example of comment higher up).

Inverle:
Wouldn't it be better to use https://www.php.net/manual/en/function.http-response-code.php with [enums](https://github.com/symfony/http-foundation/blob/7.2/Response.php#L24) that describe HTTP status codes?

ColonelMoutarde:
> > Hi, @Alkarex, so this PR will be merged? I've been updating it for a long time. I'd like it to be present, it makes the code easier to read :D
> > Have a good day !
> 
> Sorry for the delay. A part of the reason is that I am not sure what to think 🤔 To me, it makes the code harder to read, not easier, and with a learning curve instead of a well-known standard. Example:
> 
> image
> 
> So I was hopping to find time to come up with a better syntax than the above, but I have not come to it yet.
> 
> So as it is now, it adds about 100 lines of code for no clear value, or maybe even negative value.
> 
> Ideally, I would be pleased with some constants at PHP level, but even there, I am not sure it is really worth it.
> 
> There are also some errors (see example of comment higher up).

The use of backedEnum is one of the great advances of php 8.x 

In this case: 'HTTP/1.1 412 Precondition Failed' at the slightest typo, the return code no longer makes sense.

 Even though I've been coding for almost 10 years, I don't always remember all the values associated with HTTP codes. 

IDEs are perfectly capable of understanding ENUMS and offering autocompletion, which isn't possible with a string of cards. 
It's true that “FreshRSS_” makes the code heavier to read, but without the use of namespaces and [PSR4](https://www.php-fig.org/psr/psr-4/), we're doomed to make do. 

All modern frameworks, Symfony, Laravels... offer this type of syntactic sugar, as the risk of typo errors is reduced.

Another contributor has validated my PR, and I think it can add a little something and help the code to be more “user-friendly”.

If you want me to correct my PR so that the enum is not present in the lib OK. I'll make the change quickly.

ColonelMoutarde:
> Wouldn't it be better to use https://www.php.net/manual/en/function.http-response-code.php with [enums](https://github.com/symfony/http-foundation/blob/7.2/Response.php#L24) that describe HTTP status codes?

I'm sorry, I don't understand the point of this function.

https://www.php.net/manual/en/function.http-response-code.php

This is not the purpose of my PR ;)

Alkarex:
> Wouldn't it be better to use https://www.php.net/manual/en/function.http-response-code.php

Yes, that looks like an option. At least we do not have to provide a list of message strings, and it avoids very verbose constructs like:

```
header(FreshRSS_HttpResponse::description(FreshRSS_HttpResponse::HTTP_204_NO_CONTENT);
```

> I don't understand the point of this function

Indeed, the documentation of this function is really not clear. Does it set the status message? Which status codes are supported? Etc.

I remember having looked at the PHP source code to try to get a better insight:

https://github.com/php/php-src/blob/e3715cddf075e2dcfb44aae2c8fa2a36069fc965/main/http_status_codes.h#L25C45-L76

But it does the job.

There are some pitfalls though. When trying to send any code not supported by PHP (and not documented, as far as I can see), a 500 error will be sent instead.

E.g. the following will send 500 instead of 418:

```php
http_response_code(418);
```

But combined with an enum of only the HTTP status codes that we use, it could be OK.

Inverle:
@Alkarex

> E.g. the following will send 500 instead of 418:

On PHP 8.3.6 dev server, it doesn't send 500. It simply sends 418.

```php
<?php http_response_code(418) ?>
```

```
root@vmware:~# curl -I 192.168.0.88:1337
HTTP/1.1 418 Unknown Status Code
```

apache2 with PHP 8.2.28 inside freshrss container (in this case you are right):
```
root@vmware:/var/www/freshrss.local/p# curl -I 192.168.0.88:8080/test.php
HTTP/1.1 500 Internal Server Error
```
nginx:
```
root@vmware:/var/www/freshrss.local/p# curl -I freshrss.local/test.php
HTTP/1.1 418
```

PR #7194, opened 2025-01-07
Simplify the Caddy configuraton

Roan-V:
Most of the Caddy config was overly complicated, this greatly simplifies the setup!

I am currently running this for personal use, so can confirm it still works

I am unable to get this same config working without https though, so I am pretty confused on that, if somebody else would be willing to confirm that'd be great
I very much doubt that these changes would be the issue for that, but trying to get freshrss working behind a proxy was enough of an headache already (I am more than glad that I got it working at all), and this unnecessarily complicated config didn't help 😅

As a bonus, I am pretty sure this is all that should be needed for running on a subdomain instead of in a subfolder
Not adding to the PR, because I haven't tested it
``` caddy
subdomain.example.com {
	reverse_proxy freshrss:80  # Enables the reverse proxy for the configured program:port
}
```

1. Replace existing Caddy config with this updated version
2. See that it ~hopefully~ still works just fine

**Pull request checklist:**
- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/docs/en/admins/Caddy.md
+++ b/docs/en/admins/Caddy.md
@@ -21,7 +21,7 @@ To set up FreshRSS behind a reverse proxy with Caddy and using a subfolder, foll
 
     Replace `example.com` with your actual domain and the four instances of `subfolder` with the subfolder where you want FreshRSS to be hosted.
 
-> **_NOTE:_** Ensure that the Docker container name for FreshRSS (freshrss in this example) matches the name used in the Caddyfile configuration.
+    > **_NOTE:_** Ensure that the Docker container name for FreshRSS (freshrss in this example) matches the name used in the Caddyfile configuration.
 
 2. **Update FreshRSS Configuration:**
 
--- a/docs/en/admins/Caddy.md
+++ b/docs/en/admins/Caddy.md
@@ -8,42 +8,36 @@ To set up FreshRSS behind a reverse proxy with Caddy and using a subfolder, foll
 
     Update your Caddyfile with the following configuration:
 
-    ```plaintext
-    example.com {
-        redir /freshrss /freshrss/i/
-        route /freshrss* {
-            uri strip_prefix /freshrss
-            reverse_proxy freshrss:80 {
-                header_up Host {host}
-                header_up X-Real-IP {remote}
-                header_up X-Forwarded-Proto {scheme}
-                header_up X-Forwarded-Host {host}
-                header_up X-Forwarded-For {remote}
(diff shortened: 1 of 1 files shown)


Roan-V:
Oh, added note for anyone else struggling with getting freshrss working behind a proxy

This threw me for a good long loop: base_url in config.php is apparently case sensitive....

I was playing with having the redirect also change the subfolder name to FreshRSS instead of freshrss (just for looks tbh) which massively undermined my testing efforts, because getting the wanted results or not changed if I had the backslash in the URL or not
This also means that you **cannot** manually change the capitalization of the URL when navigating, you get to the page just fine, but you can't login anymore

Can you tell I had fun with this? 🥲

Frenzie:
Fwiw, using `subfolder` seems less clear to me than using a realistic folder name like `freshrss`.

Roan-V:
> Fwiw, using `subfolder` seems less clear to me than using a realistic folder name like `freshrss`.

tbh, I mostly did that to remove ambiguity with the `freshrss` that corresponds to the docker service
Don't want people changing the wrong thing

But either is fine with me

Roan-V:
> docs/en/admins/Caddy.md:26:1 MD029/ol-prefix Ordered list item prefix [Expected: 1; Actual: 2; Style: 1/2/3]
docs/en/admins/Caddy.md:36:1 MD029/ol-prefix Ordered list item prefix [Expected: 2; Actual: 3; Style: 1/2/3]
docs/en/admins/Caddy.md:50:1 MD029/ol-prefix Ordered list item prefix [Expected: 3; Actual: 4; Style: 1/2/3]

Wondering why the test is failing, list item 1 is still there
The blockquote for the note is breaking it up maybe?

Problem for tomorrow :)

Frenzie:
Presumably it'll work if you add a few spaces in front.

Roan-V:
Ah I see, makes sense
Not too familiar with markdown yet

Alkarex:
Thanks. If you can, it would be nice to add the subdomain option.

Roan-V:
> Thanks. If you can, it would be nice to add the subdomain option.

https://github.com/FreshRSS/FreshRSS/pull/7197 :)


Extra note:
I am pretty sure the root url configured in `base_url` doesn't matter
`base_url' => 'https://subdomain.example.com/` works, but `base_url' => 'https://subdomain.gobbledegook.com/` also works just fine, same for the subfolder configuration

Not really worth mentioning in the docs, but wanted to add that in case it's useful information to somebody stumbling onto this PR

Issue #7202, opened 2025-01-09
Batch reload articles for error feeds


yunanwg:
**Is your feature request related to a problem? Please describe.**
I am always frustrated when there are multiple error feeds due to various reasons, I will have to go to individual feed settings, click "Reload articles" to resolve the errors.

**Describe the solution you’d like**
Ideally the "Update Feeds" button can also help reload error feeds.

**Describe alternatives you’ve considered**
Add a command, or a specific button in settings to help do this in batch.

**Additional context**
This has happened to me several times, as I have a large majority of feeds coming from locally deployed services (i.e. [RSSHub](rsshub.app)) with local DNS rewrites. One common issue is that my local DNS was down due to upgrades, and this will totally bring everything down in FreshRSS.

Thanks!

Frenzie:
They'll automatically retry next time, won't they?

yunanwg:
Thanks for the prompt reply @Frenzie.

My observation is that yes, but only if there are new entries coming from that feeds. Saying there is an error feed that sends only one item per month, if I don't click on "Reload Article", I will have to wait for a month to let go of the Warning sign.

I think this could have been more user friendly in terms of UI/UX, to avoid having red warning signs everywhere on the interface, whereas the feeds are actually pingable.

Frenzie:
Alright, so not so much batch reload error feeds as regular feed updates should also get rid of the error sign if there are no new entries.

yunanwg:
That can definitely do the work too, and makes sense for user experience.

Alkarex:
I have tried to reproduce the problem, but everything looks fine on my side. So help welcome to reproduce the issue. And @yunanwg , I would need more information from your FreshRSS logs and syslog. There should be some lines related to those feeds.

The scenario I tried is the following:
1. Add a working feed
2. Wait for it to be refreshed automatically → all is good
3. Remove the feed from the server
4. Wait for it to be refreshed automatically → leads to a 404 error and the feed is shown as having an error
5. Wait for it to be refreshed automatically → the feed is still in error
6. Restore the feed on the server (without changing its content nor its modified time)
7. Wait for it to be refreshed automatically → the feed is good again

PR #7204, opened 2025-01-09
Reduce undeeded use of elvis operator ?:

Alkarex:


Frenzie (approved):
Any particular reason?

Alkarex:
> Any particular reason?

I bumped into a couple of examples while searching what could cause https://github.com/FreshRSS/FreshRSS/issues/7202#issuecomment-2579870795

We were returning for instance `array|false` but never used the `false` value for anything. So I wanted to simplify the types and the code. For the only case where the `false` return value was used, I think `null` is cleaner and allows the stricter and faster `??` operator.

PR #7212, opened 2025-01-11
Fix stats for MySQL

### LINKED ISSUE #7210: stats page broken
The stats page doesn't display anything at all in current edge. `init_notifications()` balks and the script fails after that.
```
Uncaught TypeError: notification is null
[main.js:1693:2]
```

_Originally posted by @Frenzie in https://github.com/FreshRSS/FreshRSS/issues/7206#issuecomment-2582204855_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7210
MySQL returns decimal for some columns, which get converted to string instead of integer in PHP


Frenzie:
Confirmed fixed.

PR #7219, opened 2025-01-13
Docker Compose Traefik 3

Alkarex:
Use `:3` instead of `3.0` to avoid having to update the documentation too often.
See https://hub.docker.com/_/traefik
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/6401


Frenzie (approved):
Probably makes sense.

PR #7230, opened 2025-01-16
Ensure no PHP buffer for SQLite download

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/pull/6931#issuecomment-2567037406


Frenzie (approved):
Interesting

Alkarex:
For the record, I could reproduce the problem with our Alpine image but not with our Debian image. I am not sure why

PR #7234, opened 2025-01-16
Hide base theme

Alkarex:
Avoid:
image

And apply same sanity filter on non-empty name for extensions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,6 +15,9 @@ indent_style = tab
 [*.js]
 indent_style = tab
 
+[*.json]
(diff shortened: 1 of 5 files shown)


math-GH:
what was the root cause of this bug?

Frenzie:
Presumably it just doesn't have a screenshot?

Edit: never mind, the base theme to build on. In that case it'll be from getting rid of truthy/falsy.

Alkarex:
> what was the root cause of this bug?

I have not tracked down exactly the code change that made it visible

PR #7259, opened 2025-01-25
Doc force-https

Alkarex:
Add some comments to the *Force HTTPS* mechanism
https://github.com/FreshRSS/FreshRSS/discussions/7252#discussioncomment-11951183

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -390,6 +390,10 @@ function customSimplePie(array $attributes = [], array $curl_options = []): \Sim
 	if (is_array($force)) {
 		$https_domains = array_merge($https_domains, $force);
 	}
+
+	// Filter lines containing a # / ; as comments
+	$https_domains = array_filter($https_domains, fn(string $v) => !preg_match('/\\s*[\/#;]/', $v));
+
 	$simplePie->set_https_domains($https_domains);
 	return $simplePie;
 }
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -391,8 +391,9 @@ function customSimplePie(array $attributes = [], array $curl_options = []): \Sim
 		$https_domains = array_merge($https_domains, $force);
 	}
 
-	// Filter lines containing a # / ; as comments
-	$https_domains = array_filter($https_domains, fn(string $v) => !preg_match('/^\\s*[\/#;]/', $v));
+	// Remove whitespace and comments starting with # / ;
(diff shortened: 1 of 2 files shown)


lib/lib_rss.php
```diff
@@ -390,6 +390,10 @@ function customSimplePie(array $attributes = [], array $curl_options = []): \Sim
 	if (is_array($force)) {
 		$https_domains = array_merge($https_domains, $force);
 	}
+
+	// Filter lines containing a # / ; as comments
+	$https_domains = array_filter($https_domains, fn(string $v) => !preg_match('/\\s*[\/#;]/', $v));
```
Frenzie: It somehow seems both a bit overcomplicated and too simple at the same time? :-)

When going for complicated I'd expect something along these lines:
```php
$v = preg_replace('/\s*[\/#;].*$/', '', $v);
return !empty(trim($v));
```

Alkarex: What case are you worried about?
`/ # ;` are all invalid characters in domain names anyway

Alkarex: But good that you are catching my attention on it: I had forgotten the `^` that I wanted to have

Frenzie: > What case are you worried about?

Reading the code and wondering why it does something imo unexpected wrt comments.

Of course if you meant it with a ^ then it's not doing anything particularly unexpected.

Frenzie: In other words, I think a comment means "everything following character x" or "a line starting with x" (roughly the same thing anyway), not "this line randomly happens to contain character x".

Frenzie: > Of course if you meant it with a ^ then it's not doing anything particularly unexpected.

But that is where the overcomplicated comes in. I don't see a need for the whitespace there.

Alkarex: Indeed. The whitespace is only because I wanted to anchor the regex at the start (which I had initially forgotten), so it looks like that now: `/^\\s*[\/#;]/`
But if you do not like it, it would also be fine for me to just search for `/#;`

Frenzie: Doesn't matter much really, as long as that `^` is there. It's just a slightly weird in between situation to me, where it doesn't support end of line comments yet has logic to support leading whitespace. :-)

Alkarex: Reworked with a more proper support for comments, also end-of-line comments

Frenzie: Sorry, haha. ^_^

Frenzie (approved):

PR #7270, opened 2025-01-29
Pass PHPStan checkBenevolentUnionTypes

Alkarex:
https://phpstan.org/config-reference#checkbenevolentuniontypes


cli/CliOptionsParser.php
```diff
@@ -24,6 +24,7 @@ public function __construct() {
 
 	private function parseInput(): void {
 		$getoptInputs = $this->getGetoptInputs();
+		// @phpstan-ignore argument.type
```
Alkarex, line 27: Maybe related to https://github.com/phpstan/phpstan-src/pull/3655

PR #7274, opened 2025-01-30
Shortcuts for adding labels

### LINKED ISSUE #2087: Add shortcut to add a label
Now that we have this great feature, we should add a shortcut to quickly add an article to a label, in the same principle as that of sharing, with numbers.

math-GH:
Closes #2087

- new shortcut for the `my labels` menu added (default: `L`
- opens the `my labels` list of the active article
- `1` - `9` selects its item from the list
- `0`: Cursor jumps into the input field

(similar behavior as for the sharing menu. Sorry, only the first 9 items are available with 1 key press)

1. open an article
2. press `L`: it opens the `my labels` list
3. press `0` - `9` to active its item

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


app/i18n/nl/conf.php
```diff
@@ -321,6 +321,7 @@
 		'load_more' => 'Laad meer artikelen',
 		'mark_favorite' => 'Markeer als favoriet',
 		'mark_read' => 'Markeer als gelezen',
+		'mylabels' => 'My labels',	// TODO
```
```suggestion
		'mylabels' => 'Mijn labels',
```

Alkarex:
Detail about i18n: I know there are also some advantages, but I am not found of duplicating the strings.
*'My labels'* has already been translated, so it is a pity to make a new, untranslated duplicate.

https://github.com/FreshRSS/FreshRSS/blob/d1eb85e374c064146c0519d4bf6ca7e936a8014c/app/i18n/en/index.php#L93

So could the existing `index.menu.tags` be used instead of a new `conf.shortcut.mylabels` ?

(P.S. It would be fine for me to rename `index.menu.tags` to `index.menu.user_labels` for improved clarity)

app/i18n/it/conf.php
```diff
@@ -321,6 +321,7 @@
 		'load_more' => 'Carica altri articoli',
 		'mark_favorite' => 'Segna come preferito',
 		'mark_read' => 'Segna come letto',
+		'mylabels' => 'My labels',	// TODO
```
UserRoot-Luca:
```suggestion
		'mylabels' => 'Le mie etichette',
```

app/views/helpers/javascript_vars.phtml
```diff
@@ -46,6 +46,7 @@ echo htmlspecialchars(json_encode([
 		'last_entry' => @$s['last_entry'],
 		'collapse_entry' => @$s['collapse_entry'],
 		'load_more' => @$s['load_more'],
+		'mylabels' => @$s['mylabels'],
```
Alkarex, line 49: Note for either this PR or another PR: we should replace this syntax with something like `$s['mylabels'] ?? ''` on all these lines

math-GH: I created a ticket for it to do it outside of this PR

PR #7275, opened 2025-01-30
i18n: [de] Über -> Info

math-GH:
For the German translation: I would say that "Info" is a better wording for "About" instead of "Über"

grafik

### DIFF
--- a/app/i18n/de/gen.php
+++ b/app/i18n/de/gen.php
@@ -176,7 +176,7 @@ return array(
 		'zh-tw' => '正體中文',	// IGNORE
 	),
 	'menu' => array(
-		'about' => 'Über',
+		'about' => 'Info',
 		'account' => 'Account',	// IGNORE
 		'admin' => 'Administration',	// IGNORE
 		'archiving' => 'Archivierung',
--- a/app/i18n/de/index.php
+++ b/app/i18n/de/index.php
@@ -12,7 +12,7 @@
 
 return array(
 	'about' => array(
-		'_' => 'Über',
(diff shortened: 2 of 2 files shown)


Frenzie:
Some random thoughts about "About" in English that don't directly pertain to the PR. About is standard phrasing in software. Info seems more generic. About is about this software, while info is info about what?

Those thoughts may or may not apply to Über vs Info. Info is a perfectly fine localization of About at least in principle.

Alkarex:
What is for instance Firefox using for that menu in German?

Here is the French variant (*À propos*):

image

Or VSCode?

image

Frenzie:
> Or VSCode?

It's the same in essentially all software on all platforms (by which I also mean the likes of Android and iOS).

Alkarex:
> It's the same in essentially all software on all platforms (by which I also mean the likes of Android and iOS).

Yes, I would just be interested in seeing a couple of examples in German 🤓

math-GH:
Good that we have this discussion. 

I came to this idea, while I had to open the "About" section in my Windows10.
It is translated into "Info" in German.
grafik

And I can remember that it was usual used "everywhere".

I know that it is translated often today to "Über" because it is the direct translation of "About".

The Android app "Binary Eye" is made by a German: The "about" is translated to "Info" too.

Another possible option:
`About` -> `About FreshRSS` -> `Über FreshRSS` would be fine. Then we should do it for every language.

Frenzie:
Right, so then `Info` is simply a decent German localization. In English that Windows entry is called `About`.

PR #7278, opened 2025-02-01
i18n: de: Lieblingsartikel -> Favoritenartikel

math-GH:
This translation has the same German wording for "Favorites" now.


Frenzie (approved):
Lieblingsartikel seems more poetic. :-) (But perhaps it's more about articles of clothing or some such than about newsfeed articles).

PR #7287, opened 2025-02-01
Improve notifications: notificationID

### LINKED ISSUE #6409: [Feature] Move or remove 'Mark as read' popup
When quickly going trough feeds I frequently hit 'Mark as read'. The result is that I get a popup that is imposed OVER the 'Mark as read' button.

No popup or a different place for the popup.

Waiting, or clicking the x. For powerusers not really ideal.

**Additional context**
image

Here you see the popup is OVER the action buttons
image

math-GH:
Closes #6409

Notification banners could have a `notificationID` as additional CSS class.
So it will be possible to design the notification banners individually via the `User CSS` extension

- add `notificationID` as optional parameter for the notification setters
- some first notification IDs given (especially `readAction` as demanded in #6409)
-

1. trigger a notification
2. see the additional CSS class in the notification banner

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -84,14 +84,16 @@
 <?php
 	$msg = '';
 	$status = 'closed';
+	$notificationID = '';
 	$notif = Minz_Request::getNotification();
 	if (!empty($notif)) {
 		$msg = $notif['content'];
 		$status = $notif['type'];
+		$notificationID = $notif['notificationID'];
 		invalidateHttpCache();
 	}
 ?>
-<div role="dialog" id="notification" class="notification <?= $status ?>" aria-describedby="dialogMsg">
+<div role="dialog" id="notification" class="notification <?= $status ?> <?= $notificationID ?>" aria-describedby="dialogMsg">
 	<span class="msg" id="dialogMsg"><?= $msg ?></span>
 	<button class="close" title="<?= _t('gen.action.close') ?>"><?= _i('close') ?></button>
 </div>
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -84,16 +84,16 @@
 <?php
 	$msg = '';
 	$status = 'closed';
-	$notificationID = '';
+	$notificationName = '';
 	$notif = Minz_Request::getNotification();
 	if (!empty($notif)) {
 		$msg = $notif['content'];
 		$status = $notif['type'];
-		$notificationID = $notif['notificationID'];
+		$notificationName = $notif['notificationName'];
 		invalidateHttpCache();
 	}
 ?>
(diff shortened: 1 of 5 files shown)


math-GH:
Does it make sense to invest the time for setting each notificationID for each notification?
Only the `Minz_Request::good()` is used 52 times.

app/layout/layout.phtml
```diff
@@ -84,14 +84,16 @@
...
 	$notif = Minz_Request::getNotification();
 	if (!empty($notif)) {
 		$msg = $notif['content'];
 		$status = $notif['type'];
+		$notificationID = $notif['notificationID'];
 		invalidateHttpCache();
 	}
 ?>
-<div role="dialog" id="notification" class="notification <?= $status ?>" aria-describedby="dialogMsg">
+<div role="dialog" id="notification" class="notification <?= $status ?> <?= $notificationID ?>" aria-describedby="dialogMsg">
```
Frenzie: Given the context, something like `notificationType` might be better than ID?

Frenzie (approved):

app/layout/layout.phtml
math-GH: `type` is already used for `good`/`bad` CSS class:

https://github.com/FreshRSS/FreshRSS/blob/4ab47defd55676d294046ad140d962d115dee0f7/app/layout/layout.phtml#L85-L94

I would say that it is an notification identifier (ID). Better wording is welcome :)

Frenzie: Notification name?

Frenzie: Just basically anything that doesn't look like `class="<? $id ?>"`. :-P

Frenzie: Notification class.

math-GH: Thanks! I chose `notificationName`

lib/Minz/Request.php
```diff
@@ -385,33 +385,33 @@ private static function requestId(): string {
...
-				'notification' => [ 'type' => $type, 'content' => $content ],
+				'notification' => [ 'type' => $type, 'content' => $content, 'notificationName' => $notificationName ],
 			];
 		Minz_Session::_param('requests', $requests);
 		Minz_Session::unlock();
 	}
 
-	public static function setGoodNotification(string $content): void {
-		self::setNotification('good', $content);
+	public static function setGoodNotification(string $content, string $notificationName = 'xx'): void {
```
Alkarex:
```suggestion
	public static function setGoodNotification(string $content, string $notificationName = ''): void {
```

Is the `xx` needed?

math-GH: good catch. It was over from the testing.

PR #7292, opened 2025-02-02
GitHub: Issue forms updated

math-GH:
The issue forms will be updated:

- use the new `type` (Bug, Feature, Task)
- use the YAML instead of markdown for the feature request form. It will provide a better form with some mandatory fields.
- feature request form: added: a field for the use case
- bug report: fixed a small issue for the system information


.github/ISSUE_TEMPLATE/bug_report_form.yaml
```diff
@@ -43,7 +44,7 @@ body:
     attributes:
       label: System information
       description: Please complete the following information
-      value: "- Database version: [e.g. Mysql 5.7, SQLite]\n- PHP version: [e.g. PHP 8.1]\n- Installation type: [e.g. Docker, Docker image source, git, Yunohost]\n-Web server type: [e.g. Apache, nginx]\n- Device: [e.g. iPhone13]\n- OS: [e.g. Ubuntu 22.04, Win10, MacOS14]\n- Browser: [e.g. Firefox 124]"
+      value: "- Database version: [e.g. Mysql 5.7, SQLite]\n- PHP version: [e.g. PHP 8.1]\n- Installation type: [e.g. Docker, Docker image source, git, Yunohost]\n- Web server type: [e.g. Apache, nginx]\n- Device: [e.g. iPhone13]\n- OS: [e.g. Ubuntu 22.04, Win10, MacOS14]\n- Browser: [e.g. Firefox 124]"
```
Alkarex:
```suggestion
      value: "- Database version: [e.g. SQLite, PostgreSQL 17]\n- PHP version: [e.g. PHP 8.2]\n- Installation type: [e.g. Docker, Docker image source, git, Yunohost]\n- Web server type: [e.g. Apache, nginx]\n- Device: [e.g. iPhone13]\n- OS: [e.g. Ubuntu 24.04, Win11, MacOS14]\n- Browser: [e.g. Firefox 134]"
```

A few updates

math-GH: Any reasons why MySQL is not anymore on the list?

Frenzie: They're just examples, I don't think it matters much one way or the other? (But shouldn't it say SQLite 3.48.0)

Alkarex: It is just a list of examples, not an exhaustive list, so it does not really matter, but MySQL has been replaced by MariaDB by most OS distributions (e.g. Debian, Ubuntu, Fedora, Red Hat, Arch...), at least by default. It has poorer searching (e.g. regex) and indexing options than PostgreSQL (and poorer than MariaDB), and oftentimes have poorer performance than SQLite (https://github.com/FreshRSS/FreshRSS/pull/5648#issuecomment-1715843198, https://github.com/FreshRSS/FreshRSS/issues/5707, https://github.com/FreshRSS/FreshRSS/pull/6957), so it is likely the least recommendable option

Alkarex: (@Frenzie was faster :-P) Regarding the version of SQLite, as you prefer. It might be difficult for users to find out (besides via our own *About* system) and is in most cases dependent to the PHP version.

math-GH: When the current `edge` is published as 1.26.0 I will add the information where to find the system environment information for finding the information fast

Alkarex: Let's try to publish 1.26.0 very soon

math-GH:
Oh no, there is an issue with the feature request from:

grafik

But I do not have any idea what it means and how to fix it. For me it looks fine

Frenzie:
I think the indentation may be wrong on/from line 21.

PR #7314, opened 2025-02-07
Mark as read button: config for the size

### LINKED ISSUE #2016: Improving the "mark all as read" big button

### LINKED ISSUE #7310: [Feature] Disable the mark all as read button

The "mark all as read" button on the bottom of the main page is something I press by accident before reading all my feeds. Which is not wanted behaviour for me.

### Use case

_No response_

Can we please disable this button in the settings somewhere?

_No response_

### Additional context

_No response_

math-GH:
Closes #7310 and #2016

default: big
grafik

small:
grafik

none:
grafik

benefit:
user can decide which button fits best. The big button has advantages ans disadvantages.


Alkarex:
(Let's try to release 1.26.0 soon, so focussing mostly on regressions)

app/Controllers/configureController.php
```diff
@@ -131,6 +131,7 @@ public function readingAction(): void {
 			FreshRSS_Context::userConf()->lazyload = Minz_Request::paramBoolean('lazyload');
 			FreshRSS_Context::userConf()->sides_close_article = Minz_Request::paramBoolean('sides_close_article');
 			FreshRSS_Context::userConf()->sticky_post = Minz_Request::paramBoolean('sticky_post');
+			FreshRSS_Context::userConf()->mark_read_button = Minz_Request::paramStringNull('mark_read_button') ?? 'big';
```
math-GH: I do not understand the issue here:

```
 Check failure on line 134 in app/Controllers/configureController.php
GitHub Actions / tests

Property FreshRSS_UserConfiguration::$mark_read_button ('big'|'none'|'small') does not accept string.
```

Alkarex: It is because you have not checked that the string is only 'big'|'none'|'small'.
You may do so with e.g. `in_array()`

math-GH: Thanks for the help.
Would it make sense to have a function similar to `Minz_Request::paramStringNull` that have the check implemented too?

Alkarex: > Would it make sense to have a function similar to `Minz_Request::paramStringNull` that have the check implemented too?

I did a quick test but did not found how to do that in a way that is understood by PHPStan, so keeping it simple for now, but help welcome with another proposition that works.

PR #7416, opened 2025-03-07
Fix issue where $keepMax is defaulted to const when non-zero value is…

BryanButlerGit:
… compared.

Fix issue where $keepMax is defaulted to const when non-zero value is compared.

Fix an issue where $keepMax var is set to ARCHIVING_RETENTION_COUNT_LIMIT when a user sets "Archiving" -> "Maximum number of articles to keep per feed" to a non-zero value for a given category.

[EXAMPLE]:
User sets value to 50 (int), This matches the !== comparison and thus is overridden by ARCHIVING_RETENTION_COUNT_LIMIT.


Alkarex:
Well spotted, thanks!

PR #7419, opened 2025-03-08
Add check for Apache mod_filter to ensure "AddOutputFilterByType" works.

dezponia:
`AddOutputFilterByType` has been moved from `mod_deflate` to `mod_filter` in Apache 2.3.7. Without the `mod_filter` module loaded Apache will error out reading the `p/.htaccess` file.

Apache documentation about the `AddOutputFilterByType` directive for reference:
https://httpd.apache.org/docs/2.4/mod/mod_filter.html#addoutputfilterbytype

This issue was encountered when installing FreshRSS 1.26.0 in a shared hosting environment where compression is applied at the reverse proxy, not the Apache web server.

Add check in .htaccess for Apache `mod_filter` to ensure `AddOutputFilterByType` works.

It may also be necessary to update the Apache `LoadModule` configuration to include `mod_filter` for the Docker container files. I haven't tested running this with Docker.

1. Setup Apache 2.4 with `mod_deflate`, but without `mod_filter` loaded.
2. Install FreshRSS 1.26.0 and try to run it.
3. Notice it doesn't work and the Apache error logs complains about missing the module for `AddOutputFilterByType`
4. Apply this patched .htaccess and try again, notice that it now works as Apache knows to ignore the `AddOutputFilterByType` directive due to missing `mod_filter`. If both `mod_filter` and `mod_deflate` are loaded it will also work.

**Pull request checklist:**

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -15,7 +15,7 @@ RUN rm -f /etc/apache2/conf.d/languages.conf /etc/apache2/conf.d/info.conf \
 		/etc/apache2/conf.d/status.conf /etc/apache2/conf.d/userdir.conf && \
 	sed -r -i "/^\s*LoadModule .*mod_(alias|autoindex|negotiation|status).so$/s/^/#/" \
 		/etc/apache2/httpd.conf && \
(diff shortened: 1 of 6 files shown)


Alkarex:
Oh, I forgot to thank you here. And please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #7426, opened 2025-03-14
Clarify MINZ usage

glyn:
MINZ (archived, read-only) is not a dependency of FreshRSS, which would be very concerning.

Instead, FreshRSS copied MINZ and has evolved it since then under the same license.

Ref: https://github.com/FreshRSS/FreshRSS/discussions/7425

## Move credits

This PR moves the credits from the web UI to the README.

### DIFF
--- a/README.fr.md
+++ b/README.fr.md
@@ -255,17 +255,18 @@ et [l’API Fever](https://freshrss.github.io/FreshRSS/fr/users/06_Fever_API.htm
 # Bibliothèques incluses
 
 * [SimplePie](https://simplepie.org/)
-* [MINZ](https://framagit.org/marienfressinaud/MINZ)
 * [php-http-304](https://alexandre.alapetite.fr/doc-alex/php-http-304/)
 * [lib_opml](https://framagit.org/marienfressinaud/lib_opml)
+* [bcrypt.js](https://github.com/dcodeIO/bcrypt.js)
 * [PhpGt/CssXPath](https://github.com/PhpGt/CssXPath)
 * [PHPMailer](https://github.com/PHPMailer/PHPMailer)
 * [Chart.js](https://www.chartjs.org)
 
-## Uniquement pour certaines options ou configurations
+# Crédits additionels
 
-* [bcrypt.js](https://github.com/dcodeIO/bcrypt.js)
-* [phpQuery](https://github.com/phpquery/phpquery)
+* Basé sur une version modifiée du [framework MINZ](https://framagit.org/marienfressinaud/MINZ)
+* Certaines [icônes](https://gitlab.gnome.org/Archive/gnome-icon-theme-symbolic) sont issues du [projet GNOME](https://www.gnome.org/)
+* Polices : [*Open Sans*](https://fonts.google.com/specimen/Open+Sans), [*Lato*](https://www.latofonts.com/lato-free-fonts/), [*Spectral*](https://github.com/productiontype/spectral)
 
 # Alternatives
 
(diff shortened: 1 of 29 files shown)


Frenzie:
The others should be marked as TODO. The CLI tool can be used to automate that.

Alkarex:
Thanks 👍🏻

@marienfressinaud A lot of it seems to be quite obsolete. Furthermore, there are other, more important packages we use and that are not in the list. See https://github.com/FreshRSS/FreshRSS/blob/edge/README.md#included-libraries

Maybe we should remove this string and move the remaining relevant things to the readme.

marienfressinaud (approved):
This makes totally sense :+1:

Alkarex:
> Some design elements come from <a href="http://twitter.github.io/bootstrap/">Bootstrap</a> although FreshRSS doesn’t use this framework.

@marienfressinaud I do not think this is relevant anymore, is it?

> <a href="https://gitlab.gnome.org/Archive/gnome-icon-theme-symbolic">Icons</a> come from the <a href="https://www.gnome.org/">GNOME project</a>.

@marienfressinaud I am not sure how much of that is left. Would you know?

> <em>Open Sans</em> font police has been created by <a href="https://fonts.google.com/specimen/Open+Sans">Steve Matteson</a>.

This font is still used, but we have a couple more used by our different themes.

Frenzie:
> Some design elements come from [Bootstrap](http://twitter.github.io/bootstrap/) although FreshRSS doesn’t use this framework.

I've always read it as that the general concept of `class="form-group"` etc. is strongly inspired by Bootstrap but otherwise has nothing to do with Bootstrap at all.

glyn:
Thanks for the feedback. I've pushed some changes to move the credits to the README and delete them from the web UI.

marienfressinaud:
> > Some design elements come from [Bootstrap](http://twitter.github.io/bootstrap/) although FreshRSS doesn’t use this framework.
> 
> I've always read it as that the general concept of `class="form-group"` etc. is strongly inspired by Bootstrap but otherwise has nothing to do with Bootstrap at all.

The look and colors were also very inspired by Bootstrap 2. You can compare the buttons at https://getbootstrap.com/2.3.2/base-css.html#buttons for example. I wrote the CSS myself though, so I'm not sure it's pertinent to keep the reference to Bootstrap (especially as it changed a lot since then).

> > [Icons](https://gitlab.gnome.org/Archive/gnome-icon-theme-symbolic) come from the [GNOME project](https://www.gnome.org/).

I don't think the icons have changed much, so I'd say they all still come from GNOME (for instance, [the "read view" icon](https://gitlab.gnome.org/Archive/gnome-icon-theme-symbolic/-/blob/master/gnome/scalable/actions/view-dual-symbolic.svg?ref_type=heads))

> > _Open Sans_ font police has been created by [Steve Matteson](https://fonts.google.com/specimen/Open+Sans).
> 
> This font is still used, but we have a couple more used by our different themes.

I found Lato, Open Sans and Spectral under https://github.com/FreshRSS/FreshRSS/tree/edge/p/themes/fonts

@glyn I noticed that you re-introduced the credit to Minz in the README, you can remove it :)

Alkarex:
A few more changes applied

glyn:
With @Alkarex's changes, I believe this PR is ready to merge. Please advise. /cc @marienfressinaud, @Frenzie

Frenzie (approved):

Alkarex:
Let's go for that for now, and do more changes in another PR if needed.

Alkarex:
@glyn Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

glyn:
> @glyn Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

Done! Thanks @Alkarex.

PR #7475, opened 2025-04-01
Catch extension exceptions in override

Alkarex:
https://github.com/FreshRSS/Extensions/pull/300#issuecomment-2768578464


hkcomori:
Thank you. PHPStan no longer gives errors about exceptions.

However, FreshRSS/Extensions#300 expects `Minz_ActionException` to provide a validation error message to the result page.
This PR returns only HTTP 500 and no error message is served to the result page.
Errors are shown on the log page, but users may have difficulty finding them.

If `Minz_ActionException` in FreshRSS/Extensions#300 is the wrong usage, what about the idea to provide another exception to serve error message on the web page? (eg. `Minz_ValidationException`)

Alkarex:
Thanks for the test. Please try again

hkcomori:
Thanks for the change. Makes it easier to find the error.

PR #7478, opened 2025-04-01
Partial revert Referrer-Policy

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/6303#issuecomment-2768907702
Was already implemented conditionally
https://github.com/FreshRSS/FreshRSS/pull/1198

https://github.com/FreshRSS/FreshRSS/blob/6ab57c1e69db1617e90b281095245bfefc9df795/app/layout/layout.phtml#L36


Alkarex:
Merging fast to avoid regression, but feedback and additional changes in another PR welcome

PR #7496, opened 2025-04-06
Web scraping forbid security headers in cURL

Alkarex:
Prevent using `Remote-User`, `X-WebAuth-User` during Web scraping.


Frenzie (approved):
I'm not sure if I fully understand how they'd end up there, but sure.

PR #7528, opened 2025-04-28
HTTP Auth disallow multiple headers

Alkarex:
When using HTTP Auth methods (including OpenID Connect), exactly 1 HTTP header should be received, not more.

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -813,6 +813,12 @@ function checkTrustedIP(): bool {
 }
 
 function httpAuthUser(bool $onlyTrusted = true): string {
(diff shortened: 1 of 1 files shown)


alheureux88:
I've finally updated my freshrss instance and this commit broke the authentication for me. Would there be a way to allow multiple header with a config?

My use case is that I have multiple service behind a reverse proxy. My authentication add multiple headers because some application use different one and so freshrss receives multiple too. Maybe they all need to be the same value?

Inverle:
> I've finally updated my freshrss instance and this commit broke the authentication for me. Would there be a way to allow multiple header with a config?
> 
> My use case is that I have multiple service behind a reverse proxy. My authentication add multiple headers because some application use different one and so freshrss receives multiple too. Maybe they all need to be the same value?

Which authentication proxy do you use? Would it be possible to send only one header for FreshRSS? (either `X-WebAuth-User` or `Remote-User`)

PR #7545, opened 2025-05-02
Fix file serving for symlinked extensions

### LINKED ISSUE #6040: Extension installation via symlink not working
I finally switching from my own modified Freshrss with "hardcoded" QuickCollapse" button to the normal FreshRSS with the QuickCollapse Extension.
In order to be able to easily update, I checked out this Extension repo next to the FreshRSS folder and added the Extension as a symlink:

```
www-data@***:~/FreshRSS/extensions$ ls -la
total 16
drwxrwxr-x  2 www-data www-data 4096 Jan 11 12:13 .
drwxr-xr-x 14 www-data www-data 4096 Jan 11 12:00 ..
-rw-rw-r--  1 www-data www-data    6 Nov  9 16:19 .gitignore
-rw-rw-r--  1 www-data www-data  197 Nov  9 16:19 README.md
lrwxrwxrwx  1 www-data www-data   50 Jan 11 12:13 xExtension-QuickCollapse -> ../../FreshRSSExtensions/xExtension-QuickCollapse/
```

This seemd to work fine at first. The extension appeared in the config.
But it didn't work.

When analyzing the issue with the Browser Dev Tools i noticed that the javascript wasn't loaded.
When opening `<domain>/freshrss/ext.php?f=xExtension-QuickCollapse%2Fstatic%2Fscript.js&t=js&1704970022` instead of the javascript file i only get:
File is not supported

I have now removed the symlink and just copied the folder over to the extensions folder …
Now it works fine.

It would be nice if this would be possible if it isn't too difficult to fix.

Btw: I think it would make sense to ignore anything in the extensions folder, not only stuff startig with "x". But I don't know if it is worth it's own issue in the FreshRSS repo.

Edit: This was with FreshRSS 1.22.1. After installing the extension I have now realized there was a new release.
I have now updated to 1.23.1 but not tested this issue again.

Inverle:
Fixes #6040, in addition to not recursively deleting symlink contents, but instead just the symlink.

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -922,6 +922,14 @@ function recursive_unlink(string $dir): bool {
 		return true;
 	}
 
+	if (is_link($dir)) {
+		if (PHP_OS_FAMILY === "Windows") {
+			return rmdir($dir);
+		}
+
+		return unlink($dir);
+	}
+
 	$files = array_diff(scandir($dir) ?: [], ['.', '..']);
 	foreach ($files as $filename) {
 		$filename = $dir . '/' . $filename;
--- a/p/ext.php
+++ b/p/ext.php
@@ -11,7 +11,7 @@ function get_absolute_filename(string $file_name): string {
 
 	$third_party_extension = realpath(THIRDPARTY_EXTENSIONS_PATH . '/' . $file_name);
 	if (false !== $third_party_extension) {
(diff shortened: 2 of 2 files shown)


Alkarex (approved):

Alkarex:
Thanks 🙂

PR #7546, opened 2025-05-02
Add CLI tool for clearing the cache of all feeds

Inverle:
After switching my FreshRSS installation from the `latest` branch to the `edge` branch, all of my feed favicons disappeared and I could only restore them by going manually through every feed and clearing the cache on it.

But it didn't feel like the best way to do it, so I did that with this tool.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -101,6 +101,9 @@ cd /usr/share/FreshRSS
 #  and the number of: 5) categories, 6) feeds, 7) read articles, 8) unread articles, 9) favourites, 10) tags,
 #  11) language, 12) e-mail.
 
+./cli/clear-all-cache.php [--user=<user>] [--all-users]
+# Clear the cache of all feeds for a specific user or all users.
+
 ./cli/import-for-user.php --user username --filename /path/to/file.ext
 # The extension of the file { .json, .opml, .xml, .zip } is used to detect the type of import.
 
--- a/cli/clear-all-cache.php
+++ b/cli/clear-all-cache.php
@@ -1,0 +1,81 @@
+#!/usr/bin/env php
+<?php
+declare(strict_types=1);
+require(__DIR__ . '/_cli.php');
+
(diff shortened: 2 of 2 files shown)


Alkarex:
Quick note: the favicons should repair themselves without any intervention after running a global refresh of the feeds.
A refresh of the page in the Web browser may be needed as well.
It is a side-effect of https://github.com/FreshRSS/FreshRSS/pull/7505

Inverle:
> Quick note: the favicons should repair themselves without any intervention after running a global refresh of the feeds.

Unfortunately I couldn't refresh them due to rate limit, so I haven't tried that.
So this is a quicker option if you can't refresh your feeds.

Frenzie:
They automatically repaired themselves for me but it did look odd for a bit until they did.

In any case the tool sounds potentially useful to me in general. :-)

Alkarex:
I have not had time to review here just yet, but I do not think this helps with the favicon change observed when upgrading to 1.26.2, since new files are used. So at best it will have no effect, and risks delaying the automatic healing

Inverle:
@Alkarex

> I have not had time to review here just yet, but I do not think this helps with the favicon change observed when upgrading to 1.26.2, since new files are used. So at best it will have no effect, and risks delaying the automatic healing

I know that the script may have to be optimized, so that it doesn't request the same favicon multiple times, when multiple users have the same feed added, but I only wrote it since I didn't know refreshing feeds would fix the favicons.
Though, for me, it fixed all favicons instantly after using it. (I have tested the script)

If the script is working as I intended, then it may be useful after updating an instance with a lot of users to 1.26.2 so that the users don't ever see any of their favicons missing. (I may be wrong here)

Inverle:
> Unfortunately I couldn't refresh them due to rate limit, so I haven't tried that.

A simpler solution that would probably also work is running:
```
sqlite3 ./data/users/test/db.sqlite 'UPDATE feed SET lastUpdate = 0'
 ```
and then refreshing feeds

PR #7554, opened 2025-05-05
Fix mark as read request showing popup due to onbeforeunload

### LINKED ISSUE #7410: [Bug] When browsing feeds, mark as read often pop up a Stay or Leave page in browser
### Describe the bug

When browsing feeds in FreshRSS (latest version) and clicking on "Mark all as read jump to next unread sibling", there is often a popup stating "This page is asking you to confirm that you want to leave" with a "Stay on page" or "Leave page" button.

This happens on both Firefox and Chrome and not everytime. I don't know what trigger this pop up to come up as I simply scroll the feeds and click the Mark as read button (I don't enter any changes).

### To Reproduce

I don't know how to effectively reproduce the issue yet.

### Expected behavior

_No response_

### FreshRSS version

1.26.0

### System information

- Database version: SQLite 3.45.3
- PHP version: 8.3.15
- Installation type: Docker from Linuxserver.io
- Web server type: nginx/1.26.2
- Device: Laptop
- OS: Fedora 41
- Browser: Firefox 135

### Additional context

_No response_

Inverle:
Fixes #7410 (introduced in 4b1a2cc)

- Uses `visibilitychange` event instead of `onbeforeunload`
- Refactors `send_mark_read_queue` to use `fetch()`


p/scripts/main.js
```diff
@@ -2076,9 +2076,6 @@ function init_normal() {
 		if (sidebar) {	// Save sidebar scroll position
 			sessionStorage.setItem('FreshRSS_sidebar_scrollTop', sidebar.scrollTop);
 		}
-		if (mark_read_queue && mark_read_queue.length > 0) {
-			return false;
```
Alkarex, line 2105: The goal of this line is to prevent leaving the page if there are pending requests (such as several articles to be marked as read) which have not completed.
Leaving too early would break some of the user's actions

Alkarex:
Decreasing the interval can be discussed, but the negative side is that more HTTP requests are sent, spending more resources both client side (thing e.g. battery and data consumption) as well as server side (think e.g. rate limits).
The other `return false` is by design, as I have explained a bit higher up - does it make sense?

Inverle:
> Decreasing the interval can be discussed, but the negative side is that more HTTP requests are sent, spending more resources both client side (thing e.g. battery and data consumption) as well as server side (think e.g. rate limits). The other `return false` is by design, as I have explained a bit higher up - does it make sense?

Okay, thanks for your input, do you think 8a8f395 is a better solution?
This solution also makes sure that the entries are marked as read, when the user leaves the page, but without showing a popup.
I have set the interval back to 1000.

p/scripts/main.js
```diff
@@ -2077,7 +2077,8 @@ function init_normal() {
 			sessionStorage.setItem('FreshRSS_sidebar_scrollTop', sidebar.scrollTop);
 		}
 		if (mark_read_queue && mark_read_queue.length > 0) {
-			return false;
+			clearTimeout(send_mark_read_queue_timeout);
+			send_mark_queue_tick(null);
```
Alkarex: I do not think that works.
It is not allowed to start new tasks in an `onbeforeunload` event, so I think it will be ignored

Alkarex: What we could try to do instead is the same logic but when the user clicks on one of our own links, before the navigation and therefore before the `onbeforeunload` event

Inverle: > I do not think that works.

I have tested this, and it works even after the user closes their tab manually or the browser.
Though it probably doesn't work when you close your browser on mobile

(what I mean is that the browser is waiting for the request to finish before unloading, even after the tab appears closed)

Alkarex: Interesting.
My guess is that Web browsers might give a very short amount of time, for instance to save state to local storage or something like that, but I though initiating new requests was completely blocked.
A reference to the relevant documentation would be welcome

Inverle: I guess it's less about the request, but more about the blocking code that's running inside the `beforeunload` event

Now, that I looked deeper into it, I found this: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon#avoid_unload_and_beforeunload

So we should probably use either `pagehide` or `visibilitychange` combined with either `sendBeacon` or `fetch` with `keepalive: true`

Even though this current approach seemed to work for me

Inverle: https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/

image

article is from 2015, as seen in the url.

p/scripts/main.js
```diff
@@ -2097,15 +2092,18 @@ function init_normal() {
...
 	faviconNbUnread();
 
-	window.onbeforeunload = function (e) {
-		const sidebar = document.getElementById('sidebar');
-		if (sidebar) {	// Save sidebar scroll position
-			sessionStorage.setItem('FreshRSS_sidebar_scrollTop', sidebar.scrollTop);
-		}
-		if (mark_read_queue && mark_read_queue.length > 0) {
-			return false;
+	document.addEventListener("visibilitychange", () => {
```
Alkarex, line 2102: I am quite in doubt that this will work to guarantee that the actions are completed.
Could we have a bit more information there, such as how it compares with `onbeforeunload` in terms of robustness?

Alkarex: Just to be sure @Inverle , you are trying to address the case of https://github.com/FreshRSS/FreshRSS/issues/7410#issuecomment-2705411269 namely that some entries are in the process of getting marked as read (due to auto-marking-as-read during scroll) while the user navigates away?

p/scripts/main.js
```diff
@@ -219,27 +219,23 @@ function send_mark_read_queue(queue, asRead, callback) {
...
-	};
-	req.onload = function (e) {
-		if (this.status != 200) {
-			return req.onerror(e);
-		}
-		const json = xmlHttpRequestJson(this);
-		if (!json) {
-			return req.onerror(e);
-		}
+	fetch('.?c=entry&a=read' + (asRead ? '' : '&is_read=0'), {
```
Alkarex: Could you please explain what has changed here? Only using the `fetch` API instead of `XMLHttpRequest`?

Inverle: Yes, we need to use `fetch` in order to use the `keepalive: true` option

p/scripts/main.js
Inverle: > Just to be sure @Inverle , you are trying to address the case of [#7410 (comment)](https://github.com/FreshRSS/FreshRSS/issues/7410#issuecomment-2705411269) namely that some entries are in the process of getting marked as read (due to auto-marking-as-read during scroll) while the user navigates away?

Yes, you can also test by clicking on an article to mark it as read[1], and while the loading spinner is still present, quickly press Ctrl-W to close the tab, then go back after a while to see it has been marked as read.

[1] Assuming you have this option enabled

image

Inverle: > I am quite in doubt that this will work to guarantee that the actions are completed. Could we have a bit more information there, such as how it compares with `onbeforeunload` in terms of robustness?

Do you mean by commenting the code? I posted a picture earlier that shows the differences.
The bugs where it says `n (bug)` appear to be fixed by now even on WebKit, so that makes `visibilitychange` tick all of the boxes.
The `visibilitychange` event gets triggered whenever the user navigates away or into the tab. Such as when the browser window gets minimized, when the tab gets closed, when the user switches to another tab etc.

After the event gets triggered, `document.visibilityState` can be checked for either `visible` or `hidden`
Then when it's `hidden` calling `fetch` with `keepalive: true` should guarantee the actions are completed after navigating away.

p/scripts/main.js
Alkarex: Ok, sounds good. Let's target 1.27.0 though (one more patch version 1.26.3 before)

p/scripts/main.js
```diff
@@ -219,27 +219,23 @@ function send_mark_read_queue(queue, asRead, callback) {
...
 	}
-	const req = new XMLHttpRequest();
-	req.open('POST', '.?c=entry&a=read' + (asRead ? '' : '&is_read=0'), true);
-	req.responseType = 'json';
-	req.onerror = function (e) {
-		for (let i = queue.length - 1; i >= 0; i--) {
-			const div = document.getElementById('flux_' + queue[i]);
-			div.querySelectorAll('a.read > .icon').forEach(icon => {
-				icon.outerHTML = div.classList.contains('not_read') ? context.icons.unread : context.icons.read;
-			});
```
Alkarex, line 230: This code was not relevant, right?

Inverle: Yes I forgot to move it to the new `catch` block

p/scripts/main.js
```diff
@@ -219,27 +219,23 @@ function send_mark_read_queue(queue, asRead, callback) {
...
+		},
+		body: JSON.stringify({
+			ajax: true,
+			_csrf: context.csrf,
+			id: queue,
+		}),
+		keepalive: true
+	}).then(resp => {
+		if (!resp.ok) {
+			return Promise.reject(resp);
```
Alkarex: Checking the support for the Fetch API and for async/await, I think the two are close enough that we should jump to the much cleaner async/await style.
That is to say `try ... await fetch() ... throw resp; ... await resp.json() ... catch...`
Would you feel like giving it a try?

Inverle: Yes, I'll do that.

Inverle: I should make `mark_favorite` and labels request also use `fetch(..., {keepalive: true})`, right?

PR #7559, opened 2025-05-06
Provide theme class for CSS

Inverle:
Adds a theme class into the HTML to be used inside CSS. Use cases include user-supplied CSS (via UserCSS) and extensions.

For example this is what I'm doing in my User CSS:
```css
	/* toggle sidebar */

	@media (prefers-color-scheme: dark) {
		body:not(.theme-pafat):not(.theme-ansum) {
			.show-sidebar img,
			.close-sidebar img {
				filter: invert(1) brightness(2);
			}
		}
	}
	
	body.theme-nord,
	body.theme-swage,
	body.theme-mapco,
	body.theme-flat,
	body.theme-dark-pink,
	body.theme-dark,
	body.theme-alternative-dark {
		.show-sidebar img,
		.close-sidebar img {
			filter: invert(1) brightness(2);
		}
	}
```

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(str_replace(' ', '-', FreshRSS_Context::userConf()->theme)) ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(preg_replace('/[\s\._]/', '-', FreshRSS_Context::userConf()->theme)) ?>">
 <?php
 	if (!Minz_Request::paramBoolean('ajax')) {
 		flush();
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(str_replace([' ', '_', '.', '<', '>', '&', '"', '\''], '-', html_entity_decode(FreshRSS_Context::userConf()->theme))) ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(str_replace([' ', '_', '.', '<', '>', '&', '"', '\''], '-', htmlspecialchars_decode(FreshRSS_Context::userConf()->theme))) ?>">
 <?php
 	if (!Minz_Request::paramBoolean('ajax')) {
 		flush();
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(preg_replace('/[\s\._]/', '-', FreshRSS_Context::userConf()->theme) ?? '') ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= htmlspecialchars(strtolower(str_replace([' ', '_', '.'], '-', FreshRSS_Context::userConf()->theme))) ?>">
 <?php
 	if (!Minz_Request::paramBoolean('ajax')) {
 		flush();
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(str_replace(' ', '-', FreshRSS_Context::userConf()->theme)) ?>">
 <?php
 	if (!Minz_Request::paramBoolean('ajax')) {
 		flush();
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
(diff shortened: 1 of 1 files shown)


app/layout/layout.phtml
```diff
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(preg_replace('/[\s\._]/', '-', FreshRSS_Context::userConf()->theme) ?? '') ?>">
```
Alkarex: * Please consider the faster function `str_replace()` if no regex feature is actually needed;
* Please use `htmlspecialchars()` or include at least `<&">` in your existing black-list of characters, or use a white-list instead

Alkarex (changes requested):

app/layout/layout.phtml
Inverle: > * Please consider the faster function `str_replace()` if no regex feature is actually needed;

I didn't realize that str_replace accepted arrays for the `search` parameter, so that's why I used regex.

> * Please use `htmlspecialchars()` or include at least `<&">` in your existing black-list of characters, or use a white-list instead

htmlspecialchars isn't really needed, because the value that is set in the user's `config.php` is already sanitized, but let's do that, just to be sure.

For example, when I set my theme as `<b>test</b>`, this is what gets put in `config.php`:
```php
  'theme' => '&lt;b&gt;test&lt;/b&gt;',
```

app/layout/layout.phtml
```diff
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= htmlspecialchars(strtolower(str_replace([' ', '_', '.'], '-', FreshRSS_Context::userConf()->theme))) ?>">
```
```suggestion
	<body class="<?= Minz_Request::actionName() ?> theme-<?=strtolower(str_replace([' ', '_', '.', '<', '>', '"', '&'], '-', FreshRSS_Context::userConf()->theme)) ?>">
```

Probably a bit lighter this way

Alkarex: Hum, we should probably prevent a few more CSS special chars
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_selectors
Edited above

Inverle: > Hum, we should probably prevent a few more CSS special chars

No, that won't be needed.

See this in a browser (the selector chars can be escaped with `\`):
```css
<div class="xyz||test">
test
</div>
<style>
.xyz\|\|test {
  background-color: red;
}
</style>
```

Alkarex: Ok, reverted in my suggestion higher up

app/layout/layout.phtml
```diff
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(str_replace([' ', '_', '.', '<', '>', '&', '"', '\''], '-', html_entity_decode(FreshRSS_Context::userConf()->theme))) ?>">
```
Alkarex: We do not use HTML entities, except for the few XML ones.
So I do not think it is necessary, but if it is, the lighter and safer `htmlspecialchars_decode()` should be sufficient

Inverle:
Whitespace is replaced to prevent injection of other CSS classes, for example when a user sets their theme to `a[any whitespace character]dropdown-menu`, any page they visit is going to be blank, due to injection of the `dropdown-menu` class.

The `dropdown-menu` class has the `display: none` property inside of it, among others.

app/layout/layout.phtml
```diff
@@ -63,7 +63,7 @@
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
 	</head>
-	<body class="<?= Minz_Request::actionName() ?>">
+	<body class="<?= Minz_Request::actionName() ?> theme-<?= strtolower(str_replace([' ', '_', '.', '<', '>', '&', '"', '\'', "\t", "\r", "\n", "\v", "\f"], '-', htmlspecialchars_decode(FreshRSS_Context::userConf()->theme))) ?>">
```
Alkarex: I found several other problems so I have just made https://github.com/FreshRSS/FreshRSS/pull/7565

After that PR, it should be sufficient here with the much simpler:
```suggestion
	<body class="<?= Minz_Request::actionName() ?> theme-<?= FreshRSS_Context::userConf()->theme ?>">
```

Inverle: @Alkarex

Does that also replace all of the characters as previously with `-`?

Including whitespace, so that CSS classes can't be injected, as well as `_` and `.` so that the class name looks better? (and in lowercase)

Inverle: I'm considering moving the theme class to `<html>` instead, like the `darkMode_auto` class for consistency
The CSS selector would then look like `:root.theme-origine` similarly to `:root.darkMode_auto`

Alkarex: Yes, whitespace should be safe already.
Regarding `_` and other cosmetic stuff, I would rather avoid, as it may create conflicts if two theme IDs only different by one of those things

Alkarex: This is really mostly of being safe with users mounting a volume with their own themes, otherwise the class becomes exactly the folder name of our themes, with a `theme-` prefix.
We can go for the `strtolower()` if you insist, but I am a bit worried every-time we remove case-sensitivity on a file system :-)

Alkarex: > I'm considering moving the theme class to `<html>` instead, like the `darkMode_auto` class for consistency The CSS selector would then look like `:root.theme-origine` similarly to `:root.darkMode_auto`

Sounds good 👍🏻

Inverle:
I decided to go with `theme_` instead of `theme-`, once again for consistency.

Alkarex (approved):

Alkarex:
Looks good to me. Let me know if you would better like it with `strtolower()` of whether it is good enough as it is

Inverle:
> Looks good to me. Let me know if you would better like it with strtolower() of whether it is good enough as it is

It's good

PR #7564, opened 2025-05-07
Implement loading spinner for marking as favorite/read

### LINKED ISSUE #2304: [Feature request] Asynchronous actions to provide instant visual feedback to users
It would be nice for the user to have an instant feedback on their action.

Here are a few examples:

1. User opens an item in their feed, then press the envelop (or associated keyboard shortcut) to mark this item as unread
2. User clicks on the star (or use associated keyboard shortcut) to mark an item as favorite

Expected results:

1. The item is immediately marked as unread (the envelop icon and the item color change to reflect this)
2. The item is immediately marked as favorite (the star icon changes to reflect this)

Acual results:

1. There is a delay before the item is marked as unread
2. There is a delay before the item is marked as favorite

Now, I understand the delay is caused because when clicking on the envelop, the page sends a request to the server, waits for the server to acknowledge and make sure everything is fine, and then finally provides visual feedback.

In the case where something goes wrong (for instance the server cannot be contacted), a flow such as the following could be used:

1. User clicks on the envelop to mark item as unread
2. The visual feedback is immediately made (envelop icon changes, background color changes)
3. The page realizes the server cannot be contacted and times out (or any other communication error)
4. The item state is reverted in the page (since it could not be updated on the server) and an error popup is displayed.

This is actually something that is implemented in [Selfoss](https://github.com/SSilence/selfoss) (another open source feed reader), so for implementation details, tips might be found in its source code :)

### LINKED ISSUE #7563: [Feature] Change "favorite" icon for a spinner when processing the request

There's no feedback the action to favorite an article is being processed and makes me believe the site has frozen.  

### Use case

Not sure what part of my setup makes the action to take a bit of time, I'm guessing it's the sqlite DB being locked during updating the feeds, since I notice this issue more at the times I have the cron setup.  

It's more noticeable on feeds with small articles where I can easily use the keyboard to go "J" -> "F".  

Change the "start" icon to some kind of spinner which makes the user know the action is being processed.  

Change to postgres to see if updating the feeds doesn't slow down the UI requests.  

### Additional context

This is also noticeable when going quickly through the articles with "J" and just watching the "unread" envelope not change for a bit for all the articles passed.

Inverle:
Closes #7563

Spinner icon is https://github.com/n3r4zzurr0/svg-spinners/blob/main/svg-css/90-ring.svg ([MIT licensed](https://github.com/n3r4zzurr0/svg-spinners/blob/main/LICENSE))

Here is how it looks:
favoritespinner

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -8,8 +8,9 @@
 		$dir = ' dir="rtl"';
 		$class = 'rtl ';
 	}
+	$class .= 'theme_' . FreshRSS_Context::userConf()->theme;
 	if (FreshRSS_Context::userConf()->darkMode !== 'no') {
-		$class .= 'darkMode_' . FreshRSS_Context::userConf()->darkMode;
+		$class .= ' darkMode_' . FreshRSS_Context::userConf()->darkMode;
 	}
 ?>
 <!DOCTYPE html>
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -62,9 +62,9 @@
 <?php } else { ?>
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
-		<link rel="preload" href="../themes/icons/spinner.svg" as="image">
-		<link rel="preload" href="../themes/icons/starred.svg" as="image">
-		<link rel="preload" href="../themes/icons/non-starred.svg" as="image">
+		<link rel="preload" href="../themes/icons/spinner.svg" as="image" />
+		<link rel="preload" href="../themes/icons/starred.svg" as="image" />
+		<link rel="preload" href="../themes/icons/non-starred.svg" as="image" />
 	</head>
 	<body class="<?= Minz_Request::actionName() ?>">
 <?php
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -62,9 +62,6 @@
 <?php } else { ?>
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
-		<link rel="preload" href="../themes/icons/spinner.svg" as="image" />
-		<link rel="preload" href="../themes/icons/starred.svg" as="image" />
-		<link rel="preload" href="../themes/icons/non-starred.svg" as="image" />
 	</head>
 	<body class="<?= Minz_Request::actionName() ?>">
 <?php
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -62,6 +62,9 @@
 <?php } else { ?>
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
+		<link rel="preload" href="../themes/icons/spinner.svg" as="image">
+		<link rel="preload" href="../themes/icons/starred.svg" as="image">
+		<link rel="preload" href="../themes/icons/non-starred.svg" as="image">
 	</head>
 	<body class="<?= Minz_Request::actionName() ?>">
 <?php
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -361,7 +361,13 @@ function mark_favorite(div) {
 	}
 	pending_entries[div.id] = true;
 
+	let originalIcon;
+
 	div.querySelectorAll('a.bookmark > .icon').forEach(icon => {
+		originalIcon = {
+			src: icon.getAttribute('src'),
+			alt: icon.getAttribute('alt')
+		};
 		icon.src = context.icons.spinner;
 		icon.alt = '⏳';
 		icon.classList.add('spinner');
@@ -372,7 +378,8 @@ function mark_favorite(div) {
 	req.responseType = 'json';
 	req.onerror = function (e) {
 		div.querySelectorAll('a.bookmark > .icon').forEach(icon => {
-			icon.outerHTML = div.classList.contains('not_read') ? context.icons.unread : context.icons.read;
+			icon.src = originalIcon.src;
+			icon.alt = originalIcon.alt;
 		});
 		delete pending_entries[div.id];
 		badAjax(this.status == 403);
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -351,10 +351,20 @@ function mark_favorite(div) {
 	}
 	pending_entries[div.id] = true;
 
+	const icon = a.querySelector('img');
+	setTimeout(() => {
+		if (pending_entries[div.id]) {
+			icon.src = '../themes/icons/spinner.svg';
+			icon.classList.add('spinner');
+		}
+	}, 200);
+
 	const req = new XMLHttpRequest();
 	req.open('POST', url, true);
 	req.responseType = 'json';
 	req.onerror = function (e) {
+		icon.src = icon.alt === '☆' ? '../themes/icons/non-starred.svg' : '../themes/icons/starred.svg';
+		icon.classList.remove('spinner');
 		delete pending_entries[div.id];
 		badAjax(this.status == 403);
 	};
@@ -366,6 +376,7 @@ function mark_favorite(div) {
 		if (!json) {
 			return req.onerror(e);
 		}
+		icon.classList.remove('spinner');
 		let inc = 0;
 		if (div.classList.contains('favorite')) {
 			div.classList.remove('favorite');
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -351,18 +351,21 @@ function mark_favorite(div) {
 	}
 	pending_entries[div.id] = true;
 
-	const icon = a.querySelector('img');
-	const originalIcon = icon.getAttribute('src');
-
-	icon.src = '../themes/icons/spinner.svg';
-	icon.classList.add('spinner');
+	let originalIcon;
+	div.querySelectorAll('a.bookmark > .icon').forEach(icon => {
+		originalIcon = icon.src;
+		icon.src = context.icons.spinner;
+		icon.classList.add('spinner');
+	});
 
 	const req = new XMLHttpRequest();
 	req.open('POST', url, true);
 	req.responseType = 'json';
 	req.onerror = function (e) {
-		icon.src = originalIcon;
-		icon.classList.remove('spinner');
+		div.querySelectorAll('a.bookmark > .icon').forEach(icon => {
+			icon.src = originalIcon;
+			icon.classList.remove('spinner');
+		});
 		delete pending_entries[div.id];
 		badAjax(this.status == 403);
 	};
@@ -374,7 +377,6 @@ function mark_favorite(div) {
 		if (!json) {
 			return req.onerror(e);
 		}
-		icon.classList.remove('spinner');
 		let inc = 0;
 		if (div.classList.contains('favorite')) {
 			div.classList.remove('favorite');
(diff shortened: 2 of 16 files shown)


Inverle:
Sometimes it's buggy when the icon can't be loaded, but that should usually follow with an error like this:

image

p/scripts/main.js
```diff
@@ -351,10 +351,21 @@ function mark_favorite(div) {
...
 	pending_entries[div.id] = true;
 
+	const icon = a.querySelector('img');
+	const originalIcon = icon.getAttribute('src');
+	setTimeout(() => {
+		if (pending_entries[div.id]) {
+			icon.src = '../themes/icons/spinner.svg';
+			icon.classList.add('spinner');
+		}
+	}, 200);
```
Frenzie: Doesn't a timeout of 200 ms kind of defeat the point? 

I'd think maybe something like:
```js
requestAnimationFrame(() => {
  requestAnimationFrame(() => {
	if (pending_entries[div.id]) {
			icon.src = '../themes/icons/spinner.svg';
			icon.classList.add('spinner');
		}
  });
});
```

Inverle: It's kind of pointless to be showing a spinner animation for like 20ms if the user has a fast enough connection to the server, this does exactly what I want.
Should I change anything?
(perhaps lowering the timeout would be sufficient)

Frenzie: I find 200 ms an almost unbearably long time. Waiting so long doesn't help to make it not feel slow. To me it's a bit of a wasted effort if it doesn't make the feeling that there's a network behind it go away. Maybe I'm just weird. :-)

> (perhaps lowering the timeout would be sufficient)

Yup, that's the gist of what I said above. Most requests finish in 15-25 ms, so waiting more than 1 frame skips a change in that edge case. It still makes it feel ever so slightly unresponsive though.

More constructively, maybe it could do something else entirely to indicate that you activated it. For example, if you do something along the general lines of `filter: brightness(50%)` you have immediate activation feedback.[1] Then you'd be spending a few frames on the animation, and some 100-200 ms later you could turn it into a spinner if the request isn't done yet.

[1] I mean that it can be done with a CSS property more so than any particular resulting visual effect, but the brightness adjustment idea does seem simple and to the point. (But probably not for dark mode.)

Frenzie: PS But mind, @Alkarex might disagree with me.

app/layout/layout.phtml
```diff
@@ -62,6 +62,9 @@
 <?php } else { ?>
 		<meta name="robots" content="noindex,nofollow" />
 <?php } ?>
+		<link rel="preload" href="../themes/icons/spinner.svg" as="image">
+		<link rel="preload" href="../themes/icons/starred.svg" as="image">
+		<link rel="preload" href="../themes/icons/non-starred.svg" as="image">
```
```suggestion
		<link rel="preload" href="../themes/icons/spinner.svg" as="image" />
		<link rel="preload" href="../themes/icons/starred.svg" as="image" />
		<link rel="preload" href="../themes/icons/non-starred.svg" as="image" />
```

I like to keep as much compatibility with various parsers as possible

p/themes/Alternative-Dark/adark.css
```diff
@@ -274,6 +274,10 @@ th {
 	box-shadow: none;
 }
 
+.spinner {
+	filter: invert(1) brightness(2) !important;
```
Alkarex: Is it possible to avoid the `!important` ?

app/layout/layout.phtml
Frenzie: Does `prefetch` work in this context? Concretely, to put it at the bottom of the priority ladder. `preload` says something like "start downloading this while we haven't even finished parsing the page" which doesn't seem to make much sense for this use case.

Alkarex: I think it would help for the first click changing the status to a spinner or to an icon, both of which we might not yet have retrieved

p/themes/Alternative-Dark/adark.css
Inverle: not sure, I put it there because the spinning icon was appearing darker after hovering it on some theme, and it was fine after adding !important

app/layout/layout.phtml
Frenzie: Sure, but that's what I mean. `prefetch` should still do that most of the time, while `preload` seems a bit rushed.

Inverle: > `preload` says something like "start downloading this while we haven't even finished parsing the page" which doesn't seem to make much sense for this use case.

When emulating a slow connection, I noticed that the spinning icon didn't appear right away after loading the page, so I put the preload tags, and the icon started appearing correctly.
(I haven't tested with prefetch)

p/themes/Alternative-Dark/adark.css
Alkarex: It is probably just a rule priority issue. It would be nice to remove it if possible, as it makes further overriding challenging.
And it would also be good to have a default rule in `p/themes/base-theme/frss.css` (and when identical, it can be avoided in the different themes)

Inverle: > It is probably just a rule priority issue. It would be nice to remove it if possible, as it makes further overriding challenging.

I'll see if there are any issues after removing it, and try to fix them in a different way.

> And it would also be good to have a default rule

What do you mean exactly?

Alkarex:
@Inverle I have fixed the support of multiple icons, in particular the bottom one that was broken (when opening an article and using the star icon at the bottom). Please test a bit more
https://github.com/FreshRSS/FreshRSS/pull/7564/commits/e549dfdaf9cd2993416939e496bda4e9451f8769

p/themes/Alternative-Dark/adark.css
Alkarex: > What do you mean exactly?

Nevermind, I forgot we do not have default dark mode rules in `frss.css` so far (would be nice to add in the future)

Inverle:
> Sometimes it's buggy when the icon can't be loaded

This is due to me testing on nginx (probably caching issue), this doesn't happen inside the FreshRSS apache2 container.
It also doesn't need \<link rel="preload"\> to function properly.

Inverle:
> @Inverle I have fixed the support of multiple icons, in particular the bottom one that was broken (when opening an article and using the star icon at the bottom). Please test a bit more [e549dfd](https://github.com/FreshRSS/FreshRSS/commit/e549dfdaf9cd2993416939e496bda4e9451f8769)

I didn't realize there was a bookmark icon at the bottom too

Alkarex:
By the way, the same should be done for read/unread

Edit: Done

p/themes/Alternative-Dark/adark.css
Alkarex: Fixed

Inverle:
Read and unread icons have their javascript var in javascript_vars.phtml but bookmark icon doesn't, despite the fact that different themes may have different starred.svg/non-starred.svg icons, just as is the case with read.svg/unread.svg

Alkarex:
> Read and unread icons have their javascript var in javascript_vars.phtml but bookmark icon doesn't, despite the fact that different themes may have different starred.svg/non-starred.svg icons, just as is the case with read.svg/unread.svg

Let's keep this improvement for another PR. Help welcome

Frenzie:
Cheers.

One thing I failed to realize when taking a peek a few days ago is that we already have a different spinner in certain circumstances, for example loading more articles and updating FreshRSS. Should it be replaced by this one or vice versa?

Inverle:
> Cheers.
> 
> One thing I failed to realize when taking a peek a few days ago is that we already have a different spinner in certain circumstances, for example loading more articles and updating FreshRSS. Should it be replaced by this one or vice versa?

spinner
loader.gif doesn't really work well

https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/base-theme/loader.gif

Alkarex:
> Should it be replaced by this one

Indeed, probably a good idea

Alwaysin:
I find this moving icon very distracting. Especially as my server is not very fast and it runs a bit of time.

Inverle:
Maybe there should be an option to disable this

> Especially as my server is not very fast and it runs a bit of time.

That's one of the use cases for it

Alwaysin:
> Maybe there should be an option to disable this
> 

I second this.
Or is there a CSS or JS rule I could use to prevent it for spinning/displaying?

Frenzie:
> Maybe there should be an option to disable this

Or just a static icon along the lines of https://icons.getbootstrap.com/icons/arrow-clockwise/ for example?

Frenzie:
> Maybe there should be an option to disable this

Or just a static icon along the lines of https://icons.getbootstrap.com/icons/arrow-clockwise/ for example?

Alkarex:
In the `<img alt` , I am using ⌛

Inverle:
@Alwaysin

try putting this in your UserJS:
```js
"use strict";

window.addEventListener("load", () => {
  document.querySelectorAll(".bookmark, .read").forEach(el => el.addEventListener("click", () => {
    window.oldIconSrc = el.querySelector('img').getAttribute('src');
  }));

  new MutationObserver((mutationsList) => {
    for (let mutation of mutationsList) {
      if (mutation.type === "attributes" && mutation.attributeName === "src") {
        if (mutation.target.src.includes('themes/icons/spinner.svg')) {
          mutation.target.src = window.oldIconSrc;
          mutation.target.classList.remove('spinner');
        }
      }
    }
  }).observe(stream, {
    attributes: true,
    subtree: true
  });
});
```

this should restore the original behavior
edit: may be slightly buggy, but you won't see a spinning icon anymore

Alwaysin:
Thank you. I have this instead of the spinning icon:
chrome_HeXpTKBMPj
At least it does not move now ^_^

Inverle:
> At least it does not move now ^_^

I edited the previous code, it should work better but it's not guaranteed

Inverle:
A static icon doesn't look bad either, as @Frenzie suggested, so for now you could replace p/themes/icons/spinner.svg with https://icons.getbootstrap.com/icons/arrow-clockwise/

Alkarex:
https://framagit.org/nicofrand/xextension-threepanesview/-/issues/28

WGH-:
This looks OK when the spinner appears as a reaction to an explicit action (i.e. clicking on read/unread/star icon).

But this is too distracting when you're just browsing through the feed (e.g. via next hotkey). I ended up making this hack:

```diff
diff --git a/p/scripts/main.js b/p/scripts/main.js
index c0e31ef9..fdfa9b6d 100644
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -319,11 +319,13 @@ function mark_read(div, only_not_read, asBatch) {
        }
        pending_entries[div.id] = true;

-       div.querySelectorAll('a.read > .icon').forEach(icon => {
-               icon.src = context.icons.spinner;
-               icon.alt = '⏳';
-               icon.classList.add('spinner');
-       });
+       if (!asBatch) {
+               div.querySelectorAll('a.read > .icon').forEach(icon => {
+                       icon.src = context.icons.spinner;
+                       icon.alt = '⏳';
+                       icon.classList.add('spinner');
+               });
+       }

        const asRead = div.classList.contains('not_read');
        const entryId = div.id.replace(/^flux_/, '');
```

ch3lmi:
fully agree with @WGH- , way too distracting when reading through, should only be visible for explicit actions

Alkarex:
PR welcome to make an option

Frenzie:
Like I said before, I think a static (or in any case less active) icon would obviate the need for an option?

Alkarex:
> Like I said before, I think a static (or in any case less active) icon would obviate the need for an option?

Ok, fair enough. PR welcome

PR #7577, opened 2025-05-10
Fix newest articles not shown

### LINKED ISSUE #7412: [Bug] Freshly received sorting: The latest news received is not displayed at the top
### Describe the bug

The new received article do not appear at the top without pressing the Show read button twice.

### To Reproduce

1. Go to 'a category'
2. Click on 'Update feeds button'
3. The newly received article does not appear at the top.

### Expected behavior

If the freshly received sorting enabled the new received article appear at the top. 

### FreshRSS version

1.26.0

### System information

- PHP version: PHP 8.4.4
- Installation type: zip
- Web server type: nginx
- Device: Google Pixel 9
- OS: Android 15
- Browser: Google Chrome

### Additional context

_No response_

Alkarex:
Case when processing was faster than 1 second.
fix https://github.com/FreshRSS/FreshRSS/issues/7412
Regression from https://github.com/FreshRSS/FreshRSS/pull/7149

### DIFF
--- a/app/Controllers/indexController.php
+++ b/app/Controllers/indexController.php
@@ -63,7 +63,7 @@ class FreshRSS_index_Controller extends FreshRSS_ActionController {
 		FreshRSS_View::prependTitle($title . ' · ');
 
 		if (FreshRSS_Context::$id_max === '0') {
-			FreshRSS_Context::$id_max = time() . '000000';
+			FreshRSS_Context::$id_max = uTimeString();
 		}
 
 		$this->view->callbackBeforeFeeds = static function (FreshRSS_View $view) {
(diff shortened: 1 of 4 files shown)


Inverle:
Not sure how to test the fix, because the bug happens randomly.

andris155:
> Not sure how to test the fix, because the bug happens randomly.

I will test it for at least 1-2 weeks. This bug happens for every 2nd day for me before.

Inverle:
I guess I will apply the fix and see if it ever happens again

Alkarex:
An easy way is to import a feed via OPML, refresh, then delete it, then import it via OPML again, refresh again.
Due to cache, it will be sub-second on a fast-enough computer.

Frenzie:
I just received a notification that said `There are -1 new articles`. Could that be related to this change?

Alkarex:
> I just received a notification that said `There are -1 new articles`. Could that be related to this change?

No, this bug existed before, but I have not taken the time to investigate. Help welcome

andris155:
It looks like works perfectly now. Thanks.

Frenzie:
> No, this bug existed before, but I have not taken the time to investigate.

Alright. I don't think I've ever seen it before.

PR #7578, opened 2025-05-12
Add ClawCloud Run button

ClawCloud-Ron:

- Add ClawCloud Run button to Readme

1. Click the button to deploy FreshRSS on ClawCloud Run.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


Alkarex:
Hello,
Quick suggestion: it looks like you are using Docker but with some configurations like `/config` that differ from our official documentation and official Docker images. While this is not a problem on the technical side, this gives confusion, in particular when users are asking for some support here. Furthermore, there are some third-party Docker images that do not offer the same features than our official images, in particular regarding OpenID Connect and other Web-server related configurations.
So if possible, could you consider using the same than our official Docker images https://hub.docker.com/r/freshrss/freshrss ?

ClawCloud-Ron:
Thank you for your feedback. We have changed the image of FreshRSS in ClawCloud Run's App Store to `freshrss/freshrss:latest`.

https://github.com/ClawCloud/Run-Template/blob/main/template/freshrss/index.yaml#L31

PR #7608, opened 2025-05-23
CSS fixes for Swage

### LINKED ISSUE #7465: [Bug] Sort button with Swage disappears on small screen
### Describe the bug

Hi,

New sort options are very useful, thanks a lot fort that update !

But with the Swage theme, when the screen is to small (on a phone for example), the button to chose sort by date, title, etc. disappears.

Is there any way to solve it ? Perhaps by using User CSS ?

Image

### To Reproduce

Use normal view with Swage theme and zoom on the page if needed

### Expected behavior

Keep the button with sort options

### FreshRSS version

1.26.0

### System information

PC, Windows 11, Firefox 134
Same problem with an Eink tablet running Firefox for Android

### Additional context

_No response_

Inverle:
- Fixes for Swage
	- Buttons not showing correctly #7465
	- Dropdown placement (some dropdowns couldn't fit in viewport)
	- Icons not appearing in some places due to color i.e. label management and subscription management page
	- Made `.form-group` and button inside search dropdown hover colors slightly darker, for better visibility
	- Nav menu style improvements on mobile
		- Smaller if not on main page
		- Align settings icon on configuration pages to the right
		- Support nav menu for feed statistics page
	- Text alignment in search dropdown
	- Ensure `input, select, textarea` don't overflow the page
	- Login link placement in anonymous view
	- Other fixes for
		- Add via bookmarklet page
		- About page
		- Shared HTML query page
		- Register page
	- Text visibility for debug log
- Provide classes `layout.phtml` within `<html>`:
	- `logged_in` if applicable
	- `controller_$NAME` if applicable
	- `file_$NAME` if applicable
- Some other fixes

### the PR is still dependent on these changes, and isn't complete without them
- ~~Named some elements with IDs in `nav_menu.phtml` to make them easier selectable in Swage~~
- ~~Fixes for frss.css~~
	- ~~It wasn't possible to scroll fully to the bottom in aside menu and configure menu on mobile~~
	- ~~Align text on expandable sections to the left on mobile~~
	- ~~Show `not-allowed` cursor on disabled buttons instead of a `pointer`~~

Tested on Chrome, Firefox and GNOME Web


Inverle:
QuickCollapse button for example is still broken though, but only on mobile

Alkarex:
For Swage, the changes need to be done in https://github.com/FreshRSS/FreshRSS/blob/edge/p/themes/Swage/swage.scss
and then run:

```sh
sass swage.scss:swage.css
```

As a side note, I would like to get rid of SCSS at some point to use only standard CSS, using [CSS nesting](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting). It should be possible to make an automatic conversion from our current SCSS files used by some of our themes. Help welcome.

Inverle:
> [make dropdowns full width on mobile](https://github.com/FreshRSS/FreshRSS/pull/7608/commits/37fa68129cc0f8ed2b289ba3497191deead6e5b7)

turns out I didn't fix them correctly

app/layout/nav_menu.phtml
```diff
@@ -4,13 +4,13 @@
...
 	<?php if ($actual_view === 'normal' || $actual_view === 'reader') { ?>
-		<div class="group">
+		<div class="group" id="nav_menu_toggle_aside">
 			<a class="btn toggle_aside" href="#aside_feed"><?= _i('category') ?></a>
 		</div>
 	<?php } ?>
 
 	<?php if (FreshRSS_Auth::hasAccess()) { ?>
-	<div id="nav_menu_actions" class="group">
+	<div class="group" id="nav_menu_actions">
```
math-GH: I prefer to have the `id` first before `class` because `id` is unique and stronger compared to `class`

math-GH (changes requested):

p/themes/Swage/swage.css
```diff
@@ -1,1182 +1,1274 @@
 :root {
-	--color-text-light: #fcfcfc;
-	--color-text-light-darker: #969696;
-	--color-text-dark: #181621;
-	--color-text-nav: rgb(0, 71.6947368421, 139);
-	--color-text-aside: #22303d;
```
math-GH: Please check the text indentation and respect the given.

p/themes/Swage/swage.css
```diff
@@ -1,1182 +1,1274 @@
...
+  --color-text-aside: #22303d;
+  --color-text-alert: #fa8052;
+  --color-text-good: #5eaabf;
+  --color-text-bad: #b0425b;
+  --color-text-bad-lighter: rgb(195.8181818182, 97.1818181818, 119.5991735537);
+  --color-text-shadow-light: #f5f3f3;
+  --color-box-shadow-light: rgb(221.8181818182, 215.1818181818, 215.1818181818);
+  --color-border-light: #f5f3f3;
+  --color-border-light-darker: rgb(221.8181818182, 215.1818181818, 215.1818181818);
+  --color-border-grey: #e3e3e3;
```
math-GH: Could you please give examples why you change so many colors? 
Would it be possible to have cleaner numbers instead of having 10  positions after decimal point?

p/themes/Swage/swage.css
```diff
@@ -1,1182 +1,1274 @@
...
+  font-family: Helvetica, Arial, sans-serif;
+}
+
+body:not(.normal) nav.configure a.dropdown-toggle {
+  margin-left: 4px;
+}
+
+#slider-content #feed_update fieldset .form-group:last-child {
+  margin-bottom: 0.4rem;
+}
```
math-GH: Please use `rem` instead of `px` (`4px` = `0.25rem`

p/themes/Swage/swage.css
```diff
@@ -1,1182 +1,1274 @@
...
-	content: url(../Swage/icons/error.svg);
+  display: inline-block;
+  padding-right: 7px;
+  width: 16px;
+  content: url(../Swage/icons/error.svg);
+}
+
+main.post > div.form-group.form-actions > div.group-controls > ul > li > a {
+  margin-left: 1rem;
 }
```
math-GH: It looks complicated. Why is it needed?

math-GH (changes requested):

p/themes/Swage/swage.css
Alkarex: I think that is due to the current SASS default conversion when running `sass swage.scss:swage.css`. It may require an option to the command line. Suggestion welcome

Alkarex: By the way, as written above:

> As a side note, I would like to get rid of SCSS at some point to use only standard CSS, using [CSS nesting](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting). It should be possible to make an automatic conversion from our current SCSS files used by some of our themes. Help welcome.

which would also solve such issues

p/themes/Swage/swage.css
Inverle: Used on the import/export page for mobile

After:

image

Before:

image

p/themes/Swage/swage.css
Inverle: Yes, it is due to SASS, I changed one color variable inside the .scss file, and that changed more colors.
It may be enough to just change the color `--color-background-light` in `.form-group:hover`, where I specifically need it:

https://github.com/FreshRSS/FreshRSS/blob/37fa68129cc0f8ed2b289ba3497191deead6e5b7/p/themes/Swage/swage.scss#L235-L243

I think it's fine as is though.

p/themes/Swage/swage.css
math-GH: good catch. Let's find a better solution. Please check the import button on the same page below "Dynamic OPML" section. The `import` button has a `margin` too. 

This is the CSS line for the button:
```
.form-group.form-actions .btn {
  margin: 0 10px;
}
```

would 
```
.form-group.form-actions .btn,
.form-group.form-actions .a {
...
```
work too?

p/themes/Swage/swage.css
math-GH: yes ,I agree to get rid of SCSS (it is a task outside this PR). But before we should check the compatibility with older devices and browser. CSS nesting is quite new

Alkarex: Yes for another PR.
CSS nesting is "Baseline 2023", so I think it is good enough to migrate some of our non-default themes. In case of problem, a user can fall back to another theme

Inverle:
I broke the actualize button from appearing after moving changes from css to scss, so I'll have to fix that too.

p/themes/Swage/swage.css
Inverle: yes, done

app/layout/nav_menu.phtml
Inverle: done

p/themes/Swage/swage.css
Inverle: I'm not sure why that happened

p/themes/Swage/swage.css
Alkarex: * https://github.com/FreshRSS/FreshRSS/issues/7611

Inverle:
todo:
- [x] ~~sign in button is incorrectly positioned on anonymous view~~
- [x] ~~the search is hidden for some reason (on anonymous view)~~
- [x] ~~about page is broken~~
- [x] ~~add feed via bookmarklet page is broken~~

math-GH:
to be honest: this PR blewed up to much more than just CSS fixes for Swage. 

I would suggest to split the PR and keep the focus here on Swage

Inverle:
> to be honest: this PR blewed up to much more than just CSS fixes for Swage.

Yes.
- Fixes for Swage
- Fixes for frss.css that also affect Swage
- Added necessary ids and class names in layout.phtml and nav_menu.phtml to use them in Swage
	- also moved an if block in nav_menu.phtml

> I would suggest to split the PR and keep the focus here on Swage

Which parts?

math-GH:
This parts could be a separate PR because it will affect also the other themes or the other themes could be benefit from it too:

>     * Provide `logged_in` class within `<html>` if applicable, and a controller class
> 
>     * Fixes for frss.css
>       
>       * It wasn't possible to scroll fully to the bottom in aside menu and configure menu on mobile
>       * Align text on expandable sections to the left on mobile
>       * Show `not-allowed` cursor on disabled buttons instead of a `pointer`
> 
>     * Some other fixes
> 
>     * Made more navigation buttons accessible within anonymous view
>

Inverle:
> * Provide `logged_in` class within `<html>` if applicable, and a controller class

It doesn't look like any other themes aside from Swage can benefit from those changes

> * Fixes for frss.css

I guess this does affect all themes, not just Swage

> * Some other fixes

Those are just for Swage (adding borders, margins and paddings in some places)

> * Made more navigation buttons accessible within anonymous view

Yes, this part would make sense, but I will also make the user queries dropdown visible (I haven't made it visible yet, I'll have to also hide the `bookmark current query` button inside the user queries dropdown)

Before:

image


After (still have to make user queries visible):

image

Inverle:
Done #7616

Inverle:
todo:
- [x] ~~fix shared html query page~~

app/layout/layout.phtml
```diff
@@ -2,19 +2,27 @@
 	declare(strict_types=1);
 	/** @var FreshRSS_View $this */
 	FreshRSS::preLayout();
-	$class = '';
+	$class = [];
+	if (Minz_Request::controllerName() !== '') {
+		$class[] = 'controller_' . Minz_Request::controllerName();
+	} else {
+		$class[] = 'file_' . pathinfo(is_string($_SERVER['SCRIPT_NAME'] ?? null) ? $_SERVER['SCRIPT_NAME'] : '', PATHINFO_FILENAME);
+	}
```
math-GH, line 10: for what is it needed?
It looks very complicated. How is it connected to the CSS fixes for Swage?

math-GH (changes requested):

app/layout/nav_menu.phtml
```diff
@@ -4,12 +4,11 @@
 ?>
 <nav class="nav_menu">
 	<?php if ($actual_view === 'normal' || $actual_view === 'reader') { ?>
-		<div class="group">
+		<div id="nav_menu_toggle_aside" class="group">
 			<a class="btn toggle_aside" href="#aside_feed"><?= _i('category') ?></a>
 		</div>
 	<?php } ?>
 
-	<?php if (FreshRSS_Auth::hasAccess()) { ?>
```
math-GH: is not it part of https://github.com/FreshRSS/FreshRSS/pull/7616?

app/layout/nav_menu.phtml
```diff
@@ -65,7 +64,8 @@
 			<a class="dropdown-close" href="#close">❌</a>
 		</div>
 
-		<div class="dropdown">
+		<?php if (FreshRSS_Auth::hasAccess()) { ?>
```
math-GH: is not it part of https://github.com/FreshRSS/FreshRSS/pull/7616?

app/layout/nav_menu.phtml
```diff
@@ -181,11 +181,11 @@
 			<a class="dropdown-close" href="#close">❌</a>
 		</div>
 		</form>
+		<?php } ?>
```
math-GH: is not it part of https://github.com/FreshRSS/FreshRSS/pull/7616?

math-GH (changes requested):

p/themes/base-theme/frss.css
```diff
@@ -2474,6 +2478,20 @@ html.slider-active {
 		text-align: center;
 	}
 
+
+	body:not(.normal):not(.reader):not(.global) .aside {
+		overflow-y: scroll;
+	}
+
```
math-GH: Which problem does it solve?

p/themes/base-theme/frss.css
```diff
@@ -2474,6 +2478,20 @@ html.slider-active {
...
+	nav.configure > .dropdown li.item.dropdown-section:last-child,
+	nav#aside_feed > ul li.item.nav-section:last-child {
+		margin-bottom: 6rem;
+	}
+
+	.form-advanced-title {
+		text-align: left;
+	}
```
math-GH: where is it used?

p/themes/base-theme/frss.css
```diff
@@ -2474,6 +2478,20 @@ html.slider-active {
...
 	}
 
+
+	body:not(.normal):not(.reader):not(.global) .aside {
+		overflow-y: scroll;
+	}
+
+	nav.configure > .dropdown li.item.dropdown-section:last-child,
+	nav#aside_feed > ul li.item.nav-section:last-child {
+		margin-bottom: 6rem;
```
math-GH: Which problem does it solve?
As far as I can oversee it could risk a scrolling where it is not needed.

math-GH (changes requested):

app/layout/layout.phtml
Inverle: If you look in swage.scss where I'm using `:root.file_*` then you'll see that I'm using it specifically for `:root.file_query` (which is the HTML user query sharing page `/api/query.php?user=...&t...=&f=html`)

After:

image

Before:

image

And then I also need the controller class e.g. `:root.controller_feed`, because there are both `/i/?c=feed&a=subscription` and `/i/?c=feed&a=add` (used for bookmarklet add) endpoints

`/i/?c=feed&a=add` after:

image

`/i/?c=feed&a=add` before:

image


edit: those classes exist because selecting just the action name from the `body` class name is not sufficient

app/layout/nav_menu.phtml
Inverle: Yes, it's no longer supposed to be in this PR

app/layout/nav_menu.phtml
Inverle: Yes, it's no longer supposed to be in this PR

app/layout/nav_menu.phtml
Inverle: Yes, it's no longer supposed to be in this PR

p/themes/base-theme/frss.css
Inverle: > It wasn't possible to scroll fully to the bottom in aside menu and configure menu on mobile

It's no longer supposed to be in this PR.

p/themes/base-theme/frss.css
Inverle: > Align text on expandable sections to the left on mobile

It's no longer supposed to be in this PR.

p/themes/base-theme/frss.css
Inverle: > It wasn't possible to scroll fully to the bottom in aside menu and configure menu on mobile

It's no longer supposed to be in this PR.

> As far as I can oversee it could risk a scrolling where it is not needed.

It's in the media query for mobile, and for specific elements, so I don't think there are any risks.

app/layout/layout.phtml
Inverle: I just noticed that the search field is missing from the API query page, so this may need more work.

p/themes/base-theme/frss.css
math-GH: could you please check your commits and pushs? Currently this code change is still part of this PR

app/layout/layout.phtml
Inverle: Needs a separate PR to solve, as this is also broken for other themes. e.g. Origine shows search bar only when not on mobile

(unless I just match the behavior of Origine for now)

p/themes/base-theme/frss.css
Inverle: @math-GH I reverted the changes from this PR that are now in #7616. But #7616 needs to be merged first, because lack of necessary changes in nav_menu.phtml now broke some parts of this PR.

(I didn't exactly foresee that this would happen)

math-GH:
Please check the reader view:

before:
grafik

this PR:
the buttons are not placed well now
grafik

Inverle:
> Please check the reader view:
> 
> before: grafik
> 
> this PR: the buttons are not placed well now grafik

Make sure to also apply the changes from #7616

math-GH:
> > Please check the reader view:
> > before: grafik
> > this PR: the buttons are not placed well now grafik
> 
> Make sure to also apply the changes from #7616

I prefer independent PRs. In this case: the other PR (7616) is still in discussion.

Inverle:
> I prefer independent PRs. In this case: the other PR (7616) is still in discussion.

I was saying that this PR can't be tested without the changes from the other PR, which is why you were seeing a broken nav menu.

Inverle:
Known bugs:

* (in settings after scrolling down)

image

* Articles don't load automatically when scrolling down in stream for some reason

p/themes/Swage/swage.scss
```diff
@@ -382,6 +457,12 @@ form {
 	border: none;
 	background-color: var(--color-background-nav-darker);
 
+	.item > a,
+	.item > .as-link,
+	.item > span {
+		width: 96% !important;
```
Alkarex, line 417: If possible, it would be better to avoid the `!important` like here and use a more precise qualification instead.
But may wait for later

Inverle: Still WIP until I fix the issues mentioned in https://github.com/FreshRSS/FreshRSS/pull/7608#issuecomment-3156668298

(and maybe do some more refactoring of the CSS)

Alkarex: Ok, I have just marked the PR as draft then

Inverle:
> Known bugs:
> 
>     * (in settings after scrolling down)
> 
> 
> image
> 
>     * Articles don't load automatically when scrolling down in stream for some reason

both of the issues were caused by `overflow-x: hidden` in `<html>` added in previous commit

Inverle:
Anonymous view when more buttons were shown (mobile view)
image
Now:
image
A little empty

Inverle:
I consider it ready but testing welcome

Alkarex:
Yes, I would like to take time to look a bit closer to the changes in `layout.phtml` but would first like to release FreshRSS 1.27.0.
This PR can land in a minor `1.27.1` or whichever is the next release.

app/layout/layout.phtml
Inverle: Bookmarklet page header looks like this now

image

Inverle:
> Yes, I would like to take time to look a bit closer to the changes in layout.phtml

I did just find one issue, fixing it right now

Inverle:
Resolves the same `x dropdown-menu` issue as in #7722 (tested with darkMode field from user config)

Inverle:
What's the specific concern with layout.phtml?
@Alkarex

math-GH:
It looks now ok for me.

math-GH (approved):

Alkarex (approved):

Alkarex:
The SCSS annoyances will vanish when we get to implement:
* https://github.com/FreshRSS/FreshRSS/issues/7611

PR #7613, opened 2025-05-25
Fix remove last share

### LINKED ISSUE #7612: [Bug] Unable to delete all sharing methods – always at least one remaining
### Describe the bug

When adding a sharing method in FreshRSS, it cannot be fully deleted. Once a sharing method is added, you must add a new one before the delete option appears for the previous one. However, no matter what, there is always at least one sharing method that cannot be removed. The system does not allow all sharing methods to be deleted, leaving one remaining permanently.

### To Reproduce

1. Go to Settings → Sharing
2. Add a new sharing method
3. Attempt to delete it – notice that delete option is not available unless you create another sharing method first
4. After creating another, delete one of the entries
5. Observe that at least one sharing method always remains and cannot be removed

### Expected behavior

Users should be able to delete all sharing methods if desired, including the last one. There should not be a requirement to always keep at least one sharing method.

### FreshRSS version

1.26.2

### System information

- Database version: sqlite 3.40.1
- PHP version: 8.2.28
- Web server type: Nginx
- OS: Debian 12
- Browser: Chrome 136

### Additional context

https://github.com/user-attachments/assets/a12c1cbe-cea9-41b7-8a74-26320b83017d

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7612

### DIFF
--- a/app/Controllers/configureController.php
+++ b/app/Controllers/configureController.php
@@ -178,7 +178,7 @@ class FreshRSS_configure_Controller extends FreshRSS_ActionController {
 		FreshRSS_View::appendScript(Minz_Url::display('/scripts/draggable.js?' . @filemtime(PUBLIC_PATH . '/scripts/draggable.js')));
 
 		if (Minz_Request::isPost()) {
-			$share = $_POST['share'] ?? null;
+			$share = $_POST['share'] ?? [];
 			if (is_array($share)) {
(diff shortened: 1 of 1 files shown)


Inverle:
I found this bug yesterday while working on #7608 but forgot about it

PR #7616, opened 2025-05-25
Fixes for frss.css

Inverle:
- Fixes for frss.css
	- It wasn't possible to scroll fully to the bottom in aside menu and configure menu on mobile (moved from #7608)
	- Align text on expandable sections to the left on mobile (moved from #7608)
	- ~~Show not-allowed cursor on disabled buttons instead of a pointer (moved from #7608)~~
	- Wrap text in search dropdown for mobile
	- Fixed page overflowing on mobile in feed settings
- Named some elements with IDs in nav_menu.phtml to make them easier selectable in Swage (moved from #7608)


Inverle:
Navigation buttons in anonymous view

After (actualize button is shown too of course, when enabled):

image

Before: 

image

How they usually look when signed in:

image

Alkarex:
Any comment on the CSS @math-GH ?

Alkarex (approved):

p/themes/base-theme/frss.css
```diff
@@ -2474,6 +2478,24 @@ html.slider-active {
 		text-align: center;
 	}
 
+
+	body:not(.normal):not(.reader):not(.global) .aside {
```
Frenzie: Wouldn't that be better as:

```css
body:not(.normal, .reader, .global) .aside
```

Frenzie: Though of course there's also the obvious question if the type of `.aside` in question can't simply be targeted by `.aside.some-other-class` (presumably to be added).

p/themes/base-theme/frss.css
```diff
@@ -2474,6 +2478,24 @@ html.slider-active {
 		text-align: center;
 	}
 
+
```
Frenzie: Nit: stray extra newline.

p/themes/base-theme/frss.css
Inverle: Simplified the selector to
```css
.aside:not(.aside_feed)
```

Frenzie (approved):

math-GH:
> Any comment on the CSS @math-GH ?

will do it.
There are some changes that needs more investigation to understand the impact on all other themes.

A comment regarding showing more buttons:
In general (it is not the fault of this PR): It will be more visible now that there are "read" and "unread" articles. The user can read but cannot mark an article as read. It could be confusing to have a articles that the user read 1 minute ago is still marked as "unread". The "read/unread" icon of each article will not be displayed in the anonymous view. So the reader could not understand which articles are "read" or "unread". Similar to the "stared" articles. 

If there are no user queries: what will be displayed in the dropdown in the anonymous view? With this PR it will be an empty dropdown list (because the "add query" list item will no be displayed).

I understand the idea behind this PR but it will make the UX (currently) worse instead of better.

(review of frss.css will follow later)

Inverle:
@math-GH any updates? also #7608

> It will be more visible now that there are "read" and "unread" articles. The user can read but cannot mark an article as read. It could be confusing to have a articles that the user read 1 minute ago is still marked as "unread"

Not a problem, as long as the user understands they're in anonymous view.
(though I understand it could be confusing)

> The "read/unread" icon of each article will not be displayed in the anonymous view. So the reader could not understand which articles are "read" or "unread". Similar to the "stared" articles.

I think this is obvious there's one unread article in this screenshot and one starred?

image

> If there are no user queries: what will be displayed in the dropdown in the anonymous view? With this PR it will be an empty dropdown list (because the "add query" list item will no be displayed).

image

I wouldn't say that this is an empty dropdown list, but it could be improved (maybe it could say `No user queries` similar to the `My labels` menu?)

math-GH:
> @math-GH any updates? also #7608
> 
> > It will be more visible now that there are "read" and "unread" articles. The user can read but cannot mark an article as read. It could be confusing to have a articles that the user read 1 minute ago is still marked as "unread"
> 
> Not a problem, as long as the user understands they're in anonymous view. (though I understand it could be confusing)

I would say that the "anonymous view" is used mostly by non registered users. It means they do not have any experiences with the system. The UI and UX should be understandable without any experiences.




> 
> > The "read/unread" icon of each article will not be displayed in the anonymous view. So the reader could not understand which articles are "read" or "unread". Similar to the "stared" articles.
> 
> I think this is obvious there's one unread article in this screenshot and one starred?

IMHO each article should have its own read/unread icon to show its status.
The background color should not be used as only markup. There are users with low contrast view (people with color disability, people with low contrast screens or low light screens etc)

Articles could starred and read/unread. Starred overrides the read/unread color (in Origine theme. I did not check the other themes)

> 
> image
> 
> > If there are no user queries: what will be displayed in the dropdown in the anonymous view? With this PR it will be an empty dropdown list (because the "add query" list item will no be displayed).
> 
> image
> 
> I wouldn't say that this is an empty dropdown list, but it could be improved (maybe it could say `No user queries` similar to the `My labels` menu?)

It is just the title line now. I would suggest to hide the User Query button if there is no User Query.

Inverle:
@math-GH
> I would say that the "anonymous view" is used mostly by non registered users. It means they do not have any experiences with the system. The UI and UX should be understandable without any experiences.

You're right. I reverted these changes, I may work on this later.

The PR has only CSS changes now.

Alkarex:
Now there is a little PR conflict

PR #7626, opened 2025-05-31
Fix user self-deletion

### LINKED ISSUE #7625: [Bug] Deleting account doesn't work
### Describe the bug

After typing in password and clicking `Remove`

Image

This prompt gets shown:

Image

After clicking OK, nothing happens and no request gets sent.

### To Reproduce

1. Create new test user account
2. Log into the account
3. Go into Profile
4. Try deleting account and see nothing happens

### Expected behavior

The account should get deleted after clicking OK

### FreshRSS version

1.26.3-dev

### System information

-

### Additional context

_No response_

Inverle:
Closes #7625

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -65,7 +65,6 @@ function init_crypto_form() {
 						const s = bcrypt.hashSync(document.getElementById('passwordPlain').value, json.salt1);
 						const c = bcrypt.hashSync(json.nonce + s, strong ? bcrypt.genSaltSync(4) : poormanSalt());
 						document.getElementById('challenge').value = c;
-						submit_button.disabled = false;
 						if (!s || !c) {
 							openNotification('Crypto error!', 'bad');
 						} else {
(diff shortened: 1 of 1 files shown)


p/scripts/extra.js
```diff
@@ -70,13 +65,15 @@ function init_crypto_form() {
...
 							openNotification('Crypto error!', 'bad');
 						} else {
 							crypto_form.removeEventListener('submit', crypto_form.onsubmit);
 							crypto_form.submit();
 						}
 					} catch (ex) {
+						submit_button.disabled = false;
 						openNotification('Crypto exception! ' + ex, 'bad');
 					}
 				}
```
Alkarex:
```suggestion
						if (!s || !c) {
							openNotification('Crypto error!', 'bad');
						} else {
							crypto_form.removeEventListener('submit', crypto_form.onsubmit);
							crypto_form.submit();
						}
					} catch (ex) {
						openNotification('Crypto exception! ' + ex, 'bad');
					}
				}
				submit_button.disabled = false;
```

A single `submit_button.disabled = false;` should be sufficient, just after the `else`

Alkarex (approved):

Alkarex:
Could catch 👍🏻 
I have not checked for how long this has been broken

p/scripts/extra.js
Inverle: You're right

Inverle:
Maybe since 5ca961b?

PR #7627, opened 2025-05-31
Docker: Alpine 3.22

Alkarex:
https://alpinelinux.org/posts/Alpine-3.22.0-released.html
PHP 8.3.21, Apache 2.4.63

### DIFF
--- a/Docker/Dockerfile-Alpine
+++ b/Docker/Dockerfile-Alpine
@@ -1,4 +1,4 @@
-FROM alpine:3.21
+FROM alpine:3.22
 
 ENV TZ=UTC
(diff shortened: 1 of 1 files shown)


Alkarex:
Waiting a bit for updating the [Dev Container](https://github.com/FreshRSS/FreshRSS/blob/edge/.devcontainer/Dockerfile) due to
https://gitlab.alpinelinux.org/alpine/aports/-/issues/17215
although we could manually force to PHP 8.4 like in https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/Dockerfile-Newest

Issue #1416, opened 2017-01-26
Consider Cache-Control: Immutable


Alkarex:
https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/
https://bitsup.blogspot.dk/2016/05/cache-control-immutable.html

Frenzie:
> 

    “just 25 network requests”

    You know something is wrong when a statement like this is said with the expectation of sounding impressive.

:rofl: 

The comments there are gold. I look forward to their contrived explanation for this behavior:

> But when you press refresh we in the past have revalidated even fresh things (i.e. with a high max age) to check to see if they were still the most current.

math-GH:
Is it still actual?

Alkarex:
Yes

Inverle:
https://github.com/FreshRSS/FreshRSS/pull/7646/commits/69ef23427358e74fb06f95a93b6fbaa139a02e6b

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/7552

PR #7646, opened 2025-06-03
Implement custom feed favicons

### LINKED ISSUE #3789: Custom Favicons
The possibility to assign custom favicons to feeds, to help separation.
The possibility to create sub-categories within the main categories. 

A favicon upload option, for each feed on the Subscription Manager.  (and an option to not refresh the favicons)

### LINKED ISSUE #6503: [Feature] Allow Customize Favicon override for each feed
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/6502

<sup>Originally posted by **johnsturgeon** May 27, 2024</sup>
I get several RSS feeds from reddit and the favicon is always the same (resulting in the SAME favicon for every feed).  It would be great if I could override the favicon on a feed by feed basis.

I did see a similar question here #3875 (...)</div>

Inverle:
Closes #3789, #6503

Icon setting when no custom icon is set yet:

image

- `Change...` button opens a file dialog, and after selecting a file shows the chosen icon in the preview on the left. `Submit` must be clicked after selecting the icon.
- `Reset to default` changes the preview icon to the default one, and also requires `Submit` to be clicked to apply the changes.

Full list of changes:
- CSP now includes `blob:` in `img-src` for
   - `indexAction()` and `feedAction()` in `subscriptionController.php`
   - all of the view actions in `indexController.php`
- Introduce new attribute `customFavicon (boolean)` for feeds that indicates if the feed has a custom favicon
   - `hashFavicon()` in `Feed.php` is dependent on this attribute
      - `hashFavicon()` has a new parameter called `skipCache (boolean)` that allows the reset of the favicon hash for the Feed object
      - `resetFaviconHash()` just calls `hashFavicon(skipCache: true)`
- `f.php` URLs now have the format of `/f.php?h=XXXXX&t=cachebuster`, where the `t` parameter is only used for serving custom favicons
   - if `t` parameter is set, `f.php` returns a `Cache-Control: immutable` header
- `stripos` and `strpos` were changed to `str_contains` in various places (refactor)
- JS for handling the custom favicon configuration logic is in `extra.js` inside `init_update_feed()` which is called when feed configuration is opened from the aside or when the subscription management page with the feed is loaded
   - Server-side code for uploading the icon in `subscriptionController.php` under `feedAction()`
   - Errors that may occur during the setting of a custom favicon:
      - Unsupported image file type (handled only server-side with `isImgMime()`)
      - When the file is bigger than 1 MiB (default), handled both client-side and server-side
      - Standard feed error when `updateFeed()` fails
- JS vars `javascript_vars.phtml` are no longer escaped with `htmlspecialchars()`, instead with json encoding,
- CSS for disabled buttons was added
- Max favicon file size is configurable with the `max_favicon_upload_size` option in `config.php` (not exposed via UI)
- Custom favicons are currently deleted only when they are either reset to the default icon, or the feed gets deleted. They do not get deleted when the user deletes their account without removing their feeds first.
- ` faviconPrepare()` and `faviconRebuild()` are not allowed to be called when the `customFavicon` attribute is `true`
- New i18n strings:
   - `'sub.feed.icon' => 'Icon'`
   - `'sub.feed.change_favicon' => 'Change…'`
   - `'sub.feed.reset_favicon' => 'Reset to default'`
   - `'sub.feed.favicon_changed_by_ext' => 'The icon has been set by the <b>%s</b> extension.'`
   - `'feedback.sub.feed.favicon.too_large' => 'Uploaded icon is too large. The maximum file size is <em>%s</em>.'`
   - `'feedback.sub.feed.favicon.unsupported_format' => 'Unsupported image file format!'`
- Extension hook `custom_favicon_hash`
   - `setCustomFavicon()` method
   - `resetCustomFavicon()` method
   - `customFaviconExt` and `customFaviconDisallowDel` attributes
   - example of usage: https://github.com/FreshRSS/Extensions/pull/337
- Extension hook `custom_favicon_btn_url`
   - Allows extensions to implement a button for setting a custom favicon for individual feeds by providing an URL. The URL will be sent a POST request with the `extAction` field set to either `query_icon_info` or `update_icon`, along with an `id` field which describes the feed's ID.


Inverle:
todo:
- [x] ~~test on postgresql, mysql and sqlite~~: works on all of them
- [x] ~~check if the new `customFavicon` column gets added after applying the PR to an existing installation~~: works on postgresql, mysql and sqlite <--- PR is no longer using a new column (outdated TODO entry)
- [x] ~~make custom favicon display correctly on shared HTML query page~~: breaks when bookmarking just one feed, and is not related to this PR (#7652)
- [x] ~~use a view for the new action so that there is no ` [notice] --- File not found: /views/javascript/originalIconUrl.phtml` message~~
- [x] ~~consider better styling for disabled button~~

Alkarex:
I am wondering what is most demanded between uploading a favicon and providing an URL.
Maybe you could ask for feedback in e.g., https://github.com/FreshRSS/FreshRSS/issues/3789, https://github.com/FreshRSS/FreshRSS/issues/6503

Inverle:
> I am wondering what is most demanded between uploading a favicon and providing an URL.

We could have both (though what would be the use case for providing an URL instead of just downloading the desired favicon and uploading it?)

Alkarex:
> though what would be the use case for providing an URL instead of just downloading the desired favicon and uploading it?

I am not sure, but it could be interesting to have a bit of user feedback

app/Models/CategoryDAO.php
```diff
@@ -93,7 +93,7 @@ protected function autoUpdateDb(array $errorInfo): bool {
 			if ($errorInfo[0] === FreshRSS_DatabaseDAO::ER_BAD_FIELD_ERROR || $errorInfo[0] === FreshRSS_DatabaseDAOPGSQL::UNDEFINED_COLUMN) {
 				$errorLines = explode("\n", (string)$errorInfo[2], 2);	// The relevant column name is on the first line, other lines are noise
 				foreach (['kind', 'lastUpdate', 'error', 'attributes', 'customFavicon'] as $column) {
-					if (stripos($errorLines[0], $column) !== false) {
+					if (strpos($errorLines[0], $column) !== false) {
```
Inverle: `stripos` is case-insensitive, while `strpos` is case-sensitive. The reason for this change is that the function was picking up `error` first due to the presence of `ERROR: ` in the error message before `customFavicon`, which likely prevented `customFavicon` from being added.

This function probably needs further improvement.

app/Models/CategoryDAO.php
```diff
@@ -90,8 +92,8 @@ protected function autoUpdateDb(array $errorInfo): bool {
 		if (isset($errorInfo[0])) {
 			if ($errorInfo[0] === FreshRSS_DatabaseDAO::ER_BAD_FIELD_ERROR || $errorInfo[0] === FreshRSS_DatabaseDAOPGSQL::UNDEFINED_COLUMN) {
 				$errorLines = explode("\n", (string)$errorInfo[2], 2);	// The relevant column name is on the first line, other lines are noise
-				foreach (['kind', 'lastUpdate', 'error', 'attributes'] as $column) {
-					if (stripos($errorLines[0], $column) !== false) {
+				foreach (['kind', 'lastUpdate', 'error', 'attributes', 'customFavicon'] as $column) {
+					if (strpos($errorLines[0], $column) !== false) {
```
Alkarex:
```suggestion
					if (str_contains($errorLines[0], $column)) {
```

app/Models/FeedDAO.php
```diff
@@ -23,7 +25,7 @@ protected function autoUpdateDb(array $errorInfo): bool {
 		if (isset($errorInfo[0])) {
 			if ($errorInfo[0] === FreshRSS_DatabaseDAO::ER_BAD_FIELD_ERROR || $errorInfo[0] === FreshRSS_DatabaseDAOPGSQL::UNDEFINED_COLUMN) {
 				$errorLines = explode("\n", (string)$errorInfo[2], 2);	// The relevant column name is on the first line, other lines are noise
-				foreach (['kind'] as $column) {
+				foreach (['kind', 'customFavicon'] as $column) {
 					if (stripos($errorLines[0], $column) !== false) {
```
Alkarex:
```suggestion
					if (str_contains($errorLines[0], $column)) {
```

p/themes/base-theme/frss.rtl.css
```diff
@@ -501,7 +506,19 @@ td.numeric {
...
+	display: none;
+}
+
 /*=== Buttons */
+button[disabled] {
+	opacity: 0.7;
+}
+
+button[disabled]:hover {
+	cursor: not-allowed !important;
```
Alkarex: If possible, please try to avoid `!important`

p/themes/base-theme/frss.css
```diff
@@ -501,7 +506,19 @@ td.numeric {
...
+	display: none;
+}
+
 /*=== Buttons */
+button[disabled] {
+	opacity: 0.7;
+}
+
+button[disabled]:hover {
+	cursor: not-allowed !important;
```
Alkarex: If possible, please try to avoid `!important`

app/Controllers/indexController.php
```diff
@@ -49,7 +49,7 @@ public function normalAction(): void {
 		$this->_csp([
 			'default-src' => "'self'",
 			'frame-src' => '*',
-			'img-src' => '* data:',
+			'img-src' => '* data: blob:',
```
Alkarex, line 60: `blob:` is only for the preview, right?

app/SQL/install.sql.pgsql.php
```diff
@@ -27,6 +27,7 @@
 	"httpAuth" VARCHAR(1024) DEFAULT NULL,
 	"error" SMALLINT DEFAULT 0,
 	"ttl" INT NOT NULL DEFAULT 0,
+	"customFavicon" SMALLINT DEFAULT 0,
```
Alkarex: As an alternative, I would suggest to put that as a JSON field in the `attributes`, since there does not seem to be any need for SQL operation on it such as sort or search. It is also much simpler and easier to maintain than a new column.

Alkarex: Something like `$feed->_attribute('customFavicon', true)` and `$feed->attributeBoolean('customFavicon')`

p/themes/base-theme/frss.css
Inverle: I would have to modify the code of other themes in order to make that possible. Maybe should be implemented in #7616 instead

app/Controllers/indexController.php
Inverle: Yes (for the uploaded image before it's submitted)

app/SQL/install.sql.pgsql.php
Inverle: Done

app/i18n/en-us/sub.php
```diff
@@ -50,6 +50,7 @@
 			'password' => 'HTTP password',	// IGNORE
 			'username' => 'HTTP username',	// IGNORE
 		),
+		'change_favicon' => 'Change...',	// TODO
```
Alkarex:
```suggestion
		'change_favicon' => 'Change...',	// IGNORE
```

app/i18n/en-us/sub.php
```diff
@@ -80,6 +81,7 @@
 		),
 		'http_headers' => 'HTTP Headers',	// IGNORE
 		'http_headers_help' => 'Headers are separated by a newline, and the name and value of a header are separated by a colon (e.g: <kbd><code>Accept: application/atom+xml<br />Authorization: Bearer some-token</code></kbd>).',	// IGNORE
+		'icon' => 'Icon',	// TODO
```
Alkarex:
```suggestion
		'icon' => 'Icon',	// IGNORE
```

app/i18n/en-us/sub.php
```diff
@@ -212,6 +214,7 @@
 		),
 		'proxy' => 'Set a proxy for fetching this feed',	// IGNORE
 		'proxy_help' => 'Select a protocol (e.g: SOCKS5) and enter the proxy address (e.g: <kbd>127.0.0.1:1080</kbd> or <kbd>username:password@127.0.0.1:1080</kbd>)',	// IGNORE
+		'reset_favicon' => 'Reset to default',	// TODO
```
Alkarex:
```suggestion
		'reset_favicon' => 'Reset to default',	// IGNORE
```

app/Controllers/subscriptionController.php
```diff
@@ -343,7 +365,22 @@ public function feedAction(): void {
 					$url_redirect = ['c' => 'subscription', 'params' => ['id' => $id]];
 			}
 
-			if ($values['url'] != '' && $feedDAO->updateFeed($id, $values) !== false) {
+			if ($favicon_uploaded && !$resetFavicon) {
+				require_once(LIB_PATH . '/favicons.php');
+				$contents = file_get_contents(is_string($favicon_path) ? $favicon_path : '');
+				if (!isImgMime(is_string($contents) ? $contents : '')) {
```
Alkarex: Should we maybe have some kind of max size check or something like that?

Alkarex (approved):

app/Controllers/subscriptionController.php
Inverle: Yes

Inverle:
There are new i18n entries in [2f1f483](https://github.com/FreshRSS/FreshRSS/pull/7646/commits/2f1f483b102b9e4cd61f1a8b7df6ed96b52cbfe0) (for error messages)

app/Models/Feed.php
```diff
@@ -286,7 +298,8 @@ public static function faviconDelete(string $hash): void {
 		@unlink($path . '.txt');
 	}
 	public function favicon(): string {
-		return Minz_Url::display('/f.php?' . $this->hashFavicon());
+		$hash = $this->hashFavicon();
+		return Minz_Url::display('/f.php?h=' . $hash . '&t=' . @filemtime(DATA_PATH . '/favicons/' . $hash . '.ico'));
```
Inverle: Let me know if I should do this only for custom favicons

app/i18n/fr/feedback.php
```diff
@@ -94,6 +94,10 @@
 			'cache_cleared' => 'Le cache de <em>%s</em> a été vidée.',
 			'deleted' => 'Le flux a été supprimé.',
 			'error' => 'Une erreur est survenue',
+			'favicon' => array(
+				'too_large' => 'Uploaded icon is too large. The maximum file size is <em>%s</em>',	// TODO
+				'unsupported_format' => 'Unsupported image file format',	// TODO
```
Alkarex:
```suggestion
				'too_large' => 'L’icône téléversée est trop grosse. La taille maximale du fichier est de <em>%s</em>.',
				'unsupported_format' => 'Format d’image non pris en charge !',
```

app/i18n/el/feedback.php
```diff
@@ -94,6 +94,10 @@
 			'cache_cleared' => '<em>%s</em> cache has been cleared',	// TODO
 			'deleted' => 'Feed has been deleted',	// TODO
 			'error' => 'Feed cannot be updated',	// TODO
+			'favicon' => array(
+				'too_large' => 'Uploaded icon is too large. The maximum file size is <em>%s</em>',	// TODO
+				'unsupported_format' => 'Unsupported image file format',	// TODO
```
Alkarex:
```suggestion
				'too_large' => 'Uploaded icon is too large. The maximum file size is <em>%s</em>.',	// TODO
				'unsupported_format' => 'Unsupported image file format!',	// TODO
```

Maybe with some punctuation

app/Controllers/subscriptionController.php
```diff
@@ -343,7 +368,27 @@ public function feedAction(): void {
 					$url_redirect = ['c' => 'subscription', 'params' => ['id' => $id]];
 			}
 
-			if ($values['url'] != '' && $feedDAO->updateFeed($id, $values) !== false) {
+			if ($favicon_uploaded && !$resetFavicon) {
+				require_once(LIB_PATH . '/favicons.php');
+				$max_size = 1 * 1024 * 1024; // 1 MB
+				if ($favicon_size > $max_size) {
+					Minz_Request::bad(_t('feedback.sub.feed.favicon.too_large', '1 MB'), $url_redirect);
```
Alkarex: Consider using our `format_bytes()` function instead of hardcoded `1 MB`

app/Models/Feed.php
Alkarex: I think it is good enough with custom favicons, but it could be nice to have the same logic for non-custom favicons

Inverle:
@Alkarex Is a default limit of 1 MB fine?

Inverle:
TODO: delete custom favicons when a user deletes their account (but maybe in a separate PR along with non-custom favicons as well?)

Alkarex:
> @Alkarex Is a default limit of 1 MB fine?

Yes, I think that is just fine

app/views/configure/system.phtml
```diff
@@ -60,6 +60,18 @@
 			</div>
 		</div>
 
+		<div class="form-group">
+			<?php
+				$max_favicon_upload_size = sprintf('%.2f', FreshRSS_Context::systemConf()->limits['max_favicon_upload_size'] / 1024 / 1024);
+			?>
+			<label class="group-name" for="max-favicon-upload-size"><?= _t('admin.system.custom-favicon.max-upload-size') ?></label>
```
Alkarex: Fine, but for such a niche option, I would rather keep only the option in `data/config.php` without Web UI. I am worried of having too many options exposed.

config.default.php
```diff
@@ -125,6 +125,9 @@
 		#   0 for an unlimited number of accounts
 		#   1 is to not allow user registrations (1 is corresponding to the admin account)
 		'max_registrations' => 1,
+
+		# Max amount of bytes that are allowed for upload of custom favicon
+		'max_favicon_upload_size' => 1 * 1024 * 1024, # 1 MB
```
```suggestion
		'max_favicon_upload_size' => 1048576,	# 1 MiB
```

Good enough with the comment to avoid some multiplications in a file that is used in every single request

p/f.php
```diff
@@ -14,7 +14,7 @@ function show_default_favicon(int $cacheSeconds = 3600): void {
 	}
 }
 
-$id = $_SERVER['QUERY_STRING'] ?? '0';
+$id = $_GET['h'] ?? '0';
```
Alkarex, line 17: Any reason for this change?

app/views/configure/system.phtml
Inverle: I was wondering earlier if it's even worth exposing the option via Web UI. Should I remove it from the UI?

Alkarex: Yes, I would prefer not having the Web UI

p/f.php
Inverle: Of course. We only want to extract the hash from the request, without the cache buster.
For example `/f.php?h=aaaaaa&t=123456` would result in: `aaaaaa&t=123456`

Inverle:
The PR is marked as draft because I'm considering an option for allowing extensions to set custom favicons for feeds

app/views/helpers/javascript_vars.phtml
```diff
@@ -83,4 +83,4 @@ echo htmlspecialchars(json_encode([
 		'spinner' => '../themes/icons/spinner.svg',
 	],
 	'extensions' => $extData,
-], JSON_UNESCAPED_UNICODE) ?: '', ENT_NOQUOTES, 'UTF-8');
+], JSON_UNESCAPED_UNICODE | JSON_HEX_TAG);
```
Inverle: This still prevents escaping from the `jsonVars` JSON by using `</script>` [1], but it may introduce HTML injection elsewhere if there is any code that relied on passing the data sanitized with `htmlspecialchars()` into `innerHTML` [2]

I haven't seen any such code though.
(`htmlspecialchars()` sanitization is replaced with [`JSON_HEX_TAG`](https://www.php.net/manual/en/json.constants.php#constant.json-hex-tag) instead here)

[1]: It looks like PHP already escapes slashes in JSON by default, so maybe not even `JSON_HEX_TAG` is needed https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped
[2]: The biggest concern may be extensions that put some of their variables into the context

Alkarex:
> The PR is marked as draft because I'm considering an option for allowing extensions to set custom favicons for feeds

I suggest we merge this one first as it looks almost ready and you can continue the extension aspect in another PR

Inverle:
> I suggest we merge this one first as it looks almost ready and you can continue the extension aspect in another PR

Ok I can do that in another PR. Is there anything left to do in this one?

app/Models/Feed.php
```diff
@@ -286,7 +298,12 @@ public static function faviconDelete(string $hash): void {
 		@unlink($path . '.txt');
 	}
 	public function favicon(): string {
-		return Minz_Url::display('/f.php?' . $this->hashFavicon());
+		$hash = $this->hashFavicon();
+		$url = '/f.php?h=' . $hash;
+		if ($this->attributeBoolean('customFavicon')) {
+			$url .= '&t=' . @filemtime(DATA_PATH . '/favicons/' . $hash . '.ico');
```
Alkarex, line 441: When including the `t` parameter, we could serve the favicon with `Cache-Control: immutable`

app/i18n/fr/feedback.php
```diff
@@ -94,6 +94,10 @@
 			'cache_cleared' => 'Le cache de <em>%s</em> a été vidée.',
 			'deleted' => 'Le flux a été supprimé.',
 			'error' => 'Une erreur est survenue',
+			'favicon' => array(
+				'too_large' => 'L’icône téléversée est trop grosse. La taille maximale du fichier est de <em>%s</em>.',
```
Alkarex:
```suggestion
				'too_large' => 'L’icône téléversée est trop grosse ! La taille maximale du fichier est de <em>%s</em>.',
```

app/views/helpers/javascript_vars.phtml
Alkarex: Looks good to me at a first glance, although `JSON_HEX_AMP` is likely needed to complement `JSON_HEX_TAG`

Alkarex:
> Is there anything left to do in this one?

Besides a minor comment and the fact that I would like to follow up on the CSS `!important` (can wait for another PR), I think it looks good 👍🏻  Would be nice with a little bit testing including the latest changes before merging

Inverle:
> > Is there anything left to do in this one?
> 
> Besides a minor comment and the fact that I would like to follow up on the CSS `!important` (can wait for another PR), I think it looks good 👍🏻 Would be nice with a little bit testing including the latest changes before merging

It turns out that `!important` wasn't needed. I must have been not seeing my changes being applied earlier because of caching.

Inverle:
While testing the button CSS change, I noticed this bug on Ansum:

Manage feed from aside:

image

Manage feed from subscription management:

image

Buttons don't look quite right here.

app/Models/Feed.php
Inverle: Done

app/views/helpers/javascript_vars.phtml
Inverle: I don't think it is. Before when we were using `htmlspecialchars()` the result was that something like `&lt;em&gt;1 MiB&lt;/em&gt;` was being printed as text into the HTML. And that's exactly what was in the JSON.

(added it anyway)

app/Models/Feed.php
Inverle: #1416

math-GH:
I am not sure if it is a good idea to keep the image size of the original image for the favicon. I uploaded a bigger image and it keeps it size. 

Could it be downsized to max. 128x128pixel?

p/themes/base-theme/frss.css
```diff
@@ -171,6 +171,14 @@ p {
 	margin: 1rem 0 0.5rem;
 }
 
+p.error {
+	color: var(--frss-font-color-error);
+}
+
+p.error:empty {
+	display: none;
+}
```
math-GH, line 185: when is the case, that the error text is empty?

p/themes/base-theme/frss.css
```diff
@@ -205,6 +213,11 @@ img.favicon {
 	vertical-align: middle;
 }
 
+img.favicon.upload {
+	transform: scale(1.5);
+	margin-left: 0.3rem;
```
math-GH: `transform: `: why do you think it is useful to have a 1.5 bigger favicon here while it is used in normal size only in the rest of the application?

`margin-left`: I am not a fan of using `rem` outside `1.0rem`, `.75rem`, `.5rem`, `.25rem`, `.125rem`. In this case `0.25rem` would fit well too. But without any margin-left it would be also fine, wouldn't it?

Inverle: > But without any margin-left it would be also fine, wouldn't it?


No, it's wrongly offset, and the size is wrong too. Try removing those properties in DevTools and see how it looks.

Maybe there's a better way to do it but unfortunately I'm not aware of such.

Frenzie:
> Could it be downsized to max. 128x128pixel?

That seems a bit low if you're going through the whole effort of adding a custom one. Though to be fair I think the largest "wild" favicon I have among my feeds at a quick glance is 48 × 48 and that does clearly look significantly better than the more common 16 × 16 ones as well as the 32 × 32 ones.

One a semi-related note, most browsers support SVG favicons now.

p/themes/base-theme/frss.css
```diff
@@ -501,7 +514,20 @@ td.numeric {
 	}
 }
 
+input#favicon-upload {
+	display: none;
+}
```
math-GH, line 528: It would be okay for me, but my 2 cents: I would prefer to keep the default upload input. It is widely known and recognized. And it is great to see the file name before the upload starts. You know.... sometimes you have different image files that looks similar on a very small scale.

p/themes/base-theme/frss.css
Inverle: > when is the case, that the error text is empty?

Before any errors occur the text in the HTML is empty:

https://github.com/FreshRSS/FreshRSS/blob/6ddd8b0f09470307a34b7ceb41c1f8f0b74462b5/app/views/helpers/feed/update.phtml#L104

And being unset in the JS in various places when no longer relevant:

https://github.com/FreshRSS/FreshRSS/blob/6ddd8b0f09470307a34b7ceb41c1f8f0b74462b5/p/scripts/extra.js#L175-L180

p/themes/base-theme/frss.css
```diff
@@ -501,7 +514,20 @@ td.numeric {
...
+
 /*=== Buttons */
+button[disabled] {
+	opacity: 0.5;
+	color: #a6a6a6;
+}
```
math-GH: could you please have a look onto `input:disabled`? It would be great to have it very similar/the same.

Regarding the color: please use variables for colors. Please no plain color anymore. It would be very useful for the other themes or dark mode or user theme.

p/themes/base-theme/frss.css
Frenzie: I think he's saying to remove both the scale and the margin and then it's fine.

p/themes/base-theme/frss.css
Inverle: Personally I'm not interested in seeing the filename after uploading, since I can just look at it before uploading.

I decided to do it this way, since I don't like the styling of the default upload input and haven't investigated if it's possible to display the filename while having it styled nicely.

Displaying the filename definitely is possible with JS though.

p/themes/base-theme/frss.css
```diff
@@ -501,7 +514,20 @@ td.numeric {
...
+
+button[disabled]:hover {
+	cursor: not-allowed;
```
math-GH, line 541: please add this `cursor` into `input:disabled` too for having it stringend

math-GH: please check the other themes too (i.e. North theme). Disabled buttons should not get a border highlight while hovering

p/themes/base-theme/frss.css
Inverle: Maybe I put the margin after the transform. I don't remember.

It looks better when bigger though

image

vs

image

math-GH: regarding the margin-left: Please check other themes too (i.e. North theme)

p/themes/base-theme/frss.css
Inverle: > Disabled buttons should not get a border highlight while hovering

I have checked all of the themes both with light mode and dark mode, and the disabled button looks fine. I wasn't bothered by the border highlight but I'll look into it

p/themes/base-theme/frss.css
Inverle: > regarding the margin-left: Please check other themes too (i.e. North theme)

I checked, the icon looks fine.

Frenzie: > It looks better when bigger though

It seems to me it makes more sense to say width/height: 1.5rem than to do this scale + margin dance in any case.

math-GH:
What do you think of having the favicon above the `title` input?

It feels more natural IMHO.
grafik

p/themes/base-theme/frss.css
Inverle: Indeed, I'll fix that.

Inverle:
> What do you think of having the favicon above the title input?

image

> It feels more natural IMHO.

Kind of. Does that look better?

Frenzie:
> What do you think of having the favicon above the title input?

That seems a bit odd to me. You mean something like that way the order of the settings matches how it's typically displayed?

Frenzie:
> Does that look better?

I think two things are being conflated here. Displaying the icon earlier in some manner feels natural, the most logical position simply being left of the title. But changing it feels about right where it was before.

Alkarex:
> I am not sure if it is a good idea to keep the image size of the original image for the favicon. I uploaded a bigger image and it keeps it size.
> 
> Could it be downsized to max. 128x128pixel?

Do you mean by CSS or by physically downsampling the image? If the latter, I would suggest to postpone that to another PR as it involves other types of dependencies. And we do not do that for favicons provided by feeds, which could face the same problem

math-GH:
> > Does that look better?
> 
> I think two things are being conflated here. Displaying the icon earlier in some manner feels natural, the most logical position simply being left of the title. But changing it feels about right where it was before.

that was my thought too.
IMHO: Having the favicon below the category select list is not the best place.

math-GH:
> > I am not sure if it is a good idea to keep the image size of the original image for the favicon. I uploaded a bigger image and it keeps it size.
> > Could it be downsized to max. 128x128pixel?
> 
> Do you mean by CSS or by physically downsampling the image? If the latter, I would suggest to postpone that to another PR as it involves other types of dependencies. And we do not do that for favicons provided by feeds, which could face the same problem

I mean physically. Most websites care about a correct size of their favicon. But a normal user (I think about my parents XD ) do not care about any image dimensions. They would upload a 50MP image if they like the picture. Does it make sense to load it fullsize to size it down via CSS into 32x32pixl

Alkarex:
> I mean physically

Yes, but then I suggest keeping that to another issue as it is another beast. I am not super keen on adding a dependency to e.g. `gd` (which we do not have at the moment) just for such a niche case

Edit: `gd` would increase our Alpine image by 18.4MB, just to speak about size

lib/Minz/ExtensionManager.php
```diff
@@ -362,7 +366,10 @@ public static function callHook(string $hook_name, ...$args) {
 			return self::callOneToOne($hook_name, $args[0] ?? null);
 		} elseif ($signature === 'PassArguments') {
 			foreach (self::$hook_list[$hook_name]['list'] as $function) {
-				call_user_func($function, ...$args);
+				$result = call_user_func($function, ...$args);
+				if (isset($result)) {
```
Alkarex:
```suggestion
				if ($result !== null) {
```

That variable will always be set, by it may be null

math-GH:
> > I mean physically
> 
> Yes, but then I suggest keeping that to another issue as it is another beast. I am not super keen on adding a dependency to e.g. `gd` (which we do not have at the moment) just for such a niche case
> 
> Edit: `gd` would increase our Alpine image by 18.4MB, just to speak about size

Fair point. In that case I would limit the upload size instead of downsizing it.

Inverle:
> I would limit the upload size instead of downsizing it.

It's already limited to 1MiB but we could reduce it to 512KiB maybe

(biggest icon in my feeds is 0.2mb)

johnsturgeon:
> It's already limited to 1MiB but we could reduce it to 512KiB maybe

IMO 1MiB is enough to keep folks from pushing icons that are orders of magnitude too large.  I think the idea of a 'limit' is really to prevent something bad from happening, and there might be scenarios where a file pushes over 500k

app/i18n/fr/sub.php
```diff
@@ -74,12 +75,15 @@
 			'help' => 'Fichier XML (données partielles. <a href="https://freshrss.github.io/FreshRSS/en/developers/OPML.html" target="_blank">Voir documentation</a>)',
 			'label' => 'Exporter en OPML',
 		),
+		'ext_favicon' => 'Set automatically',	// TODO
+		'favicon_changed_by_ext' => 'The icon has been set by the <b>%s</b> extension.',	// TODO
```
```suggestion
		'ext_favicon' => 'Définie automatiquement',
		'favicon_changed_by_ext' => 'L’icône a été définie par l’extension <b>%s</b>.',
```

Inverle:
#7792

PR #7654, opened 2025-06-05
Handle redirects when scraping feed from HTML

### LINKED ISSUE #6991: [Bug] FreshRSS body scraper does not properly handle relative links
### Describe the bug

FreshRSS tries to save a local HTML version of pages in the feed, but incorrectly handles urls with relative links. Specifically, it treats the referring page as the site domain rather than the actual location of the article.

### To Reproduce

Feed: <https://blog.thea.codes/feed.xml>
Feed entry: <https://blog.thea.codes/typefaces-i-love/> (No body)
Scraped with CSS Selector: `article`

Real HTML:
```html
...
<img alt="Nunito specimen" src="./nunito.png">
...
```

FreshRSS scrape:
```html
...
<img alt="Nunito specimen" src="https://blog.thea.codes/nunito.png">
...
```

### Expected behavior

Correct scrape:
```html
...
<img alt="Nunito specimen" src="https://blog.thea.codes/typefaces-i-love/nunito.png">
...
```

### FreshRSS version

1.24.3

### Environment information

- Database version: [e.g. Mysql 5.7, SQLite]
- PHP version: 8.2.20
- Installation type: Docker

### Additional context

_No response_

Inverle:
Fixes #6991

It wasn't possible to correctly scrape https://blog.thea.codes/feed.xml because the entry URLs didn't have a trailing slash in them.
If you opened one of the entries in a browser, it would automatically redirect you to a URL with a trailing slash in it.

e.g. https://blog.thea.codes/my-2024 -> https://blog.thea.codes/my-2024/

In this PR, redirects are now handled properly, and the entry URLs are replaced with the redirected versions of them.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -882,7 +882,12 @@ HTML;
 			// Follow redirections
 			if ($maxRedirs > 0) {
 				if ($url !== $response['effective_url']) {
-					return $this->getContentByParsing($response['effective_url'], $maxRedirs - $response['redirect_count'], redirected: true, reloadAction: $reloadAction);
+					return $this->getContentByParsing(
+						$response['effective_url'],
+						$maxRedirs - $response['redirect_count'],
+						redirected: true,
+						reloadAction: $reloadAction
+					);
 				}
 				$metas = $xpath->query('//meta[@content]') ?: [];
 				foreach ($metas as $meta) {
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -880,15 +880,12 @@ HTML;
 			$xpath = new DOMXPath($doc);
 
 			// Follow redirections
+			if ($url !== $response['effective_url']) {
+				$maxRedirs -= $response['redirect_count'];
+				$url = $response['effective_url'];
+				$redirected = true;
+			}
 			if ($maxRedirs > 0) {
-				if ($url !== $response['effective_url']) {
-					return $this->getContentByParsing(
-						$response['effective_url'],
-						$maxRedirs - $response['redirect_count'],
-						redirected: true,
-						reloadAction: $reloadAction
-					);
-				}
 				$metas = $xpath->query('//meta[@content]') ?: [];
 				foreach ($metas as $meta) {
 					if ($meta instanceof DOMElement && strtolower(trim($meta->getAttribute('http-equiv'))) === 'refresh') {
@@ -904,7 +901,7 @@ HTML;
 			// Update entry URL after following all redirects
 			if ($redirected && !$response['fail']) {
 				$entryDAO = FreshRSS_Factory::createEntryDao();
-				$this->_link(htmlspecialchars($url, ENT_QUOTES));
+				$this->_link(htmlspecialchars($url, ENT_COMPAT, 'UTF-8'));
 				if (!$reloadAction && $entryDAO->updateEntry($this->toArray()) === false) {
 					return '';
 				}
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -845,7 +845,7 @@ HTML;
 	 * @param string $url Overridden URL. Will default to the entry URL.
 	 * @throws Minz_Exception
 	 */
-	public function getContentByParsing(string $url = '', int $maxRedirs = 4, bool $redirected = false, bool $reloadAction = false): string {
+	public function getContentByParsing(string $url = '', int $maxRedirs = 4): string {
 		$url = $url ?: htmlspecialchars_decode($this->link(), ENT_QUOTES);
 		$feed = $this->feed();
 		if ($url === '' || $feed === null || $feed->pathEntries() === '') {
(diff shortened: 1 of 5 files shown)


app/Models/Entry.php
```diff
@@ -872,26 +872,44 @@ public function getContentByParsing(string $url = '', int $maxRedirs = 3): strin
...
+							return $this->getContentByParsing($refresh, $maxRedirs - 1, redirected: true, reloadAction: $reloadAction);
 						}
 					}
 				}
 			}
 
+			// Update entry URL after following all redirects
+			if ($redirected && !$response['fail']) {
+				$entryDAO = FreshRSS_Factory::createEntryDao();
+				$this->_link(htmlspecialchars($url, ENT_QUOTES));
```
```suggestion
				$this->_link(htmlspecialchars($url, ENT_COMPAT, 'UTF-8'));
```

app/Models/Entry.php
```diff
@@ -872,26 +872,44 @@ public function getContentByParsing(string $url = '', int $maxRedirs = 3): strin
...
+				if (!$reloadAction && $entryDAO->updateEntry($this->toArray()) === false) {
```
Alkarex: Could you please explain this line?
Do we need to commit the changes to DB at that stage? Any change to `entry` should already be committed later

Inverle: We need to update the URL of the entry in the preview as well, which is why we are checking that the request didn't come from `reloadAction()` before making changes to `entry`. (because `reloadAction()` already commits changes to `entry`, as you said)

from `reloadAction()`:
https://github.com/FreshRSS/FreshRSS/blob/08e7933c1e26f22c91905ab52475365880d8a730/app/Controllers/feedController.php#L1120-L1124

`loadCompleteContent()` then calls `getContentByParsing()`:
https://github.com/FreshRSS/FreshRSS/blob/08e7933c1e26f22c91905ab52475365880d8a730/app/Models/Entry.php#L989-L990

meanwhile `contentSelectorPreviewAction()` only calls `getContentByParsing()`:
https://github.com/FreshRSS/FreshRSS/blob/08e7933c1e26f22c91905ab52475365880d8a730/app/Controllers/feedController.php#L1191-L1205

note: even though I called the boolean `reloadAction`  it will also affect `actualizeAction()`, since it calls `loadCompleteContent()`

Alkarex: > We need to update the URL of the entry in the preview as well

Ok, but the preview should not change the DB, right?

Inverle: > Ok, but the preview should not change the DB, right?

When the user accesses the preview once, then there is no problem with not changing the entry URL. But then the content for the previous URL is being cached, and later being accessed when the user changes their preview selector (resulting in broken images, again).

See:
https://github.com/FreshRSS/FreshRSS/blob/08e7933c1e26f22c91905ab52475365880d8a730/app/Models/Entry.php#L874-L876

We only know which URL we should cache after the `httpGet()` call, and `httpGet()` requires a cache path.

(suggestions for a different approach welcome)

Alkarex: I have a new approach coming. The proposed one breaks several things including multi-user cache and article unicity logic

Inverle: I recently realized that it breaks other stuff, so a new approach is still welcome as I haven't come up with a better one myself yet.

Alkarex: See https://github.com/FreshRSS/FreshRSS/pull/7654/commits/b844eb466276fd455be2240298751b0aae3c75af
Relying on HTML base, standard way to save an HTML document with relative references.
Tests welcome

Inverle: Okay that sounds like a better solution than mine.
I'll try testing it.

lib/lib_rss.php
```diff
@@ -507,12 +507,53 @@ function enforceHttpEncoding(string $html, string $contentType = ''): string {
...
+	$doc->loadHTML($html, LIBXML_NONET | LIBXML_NOERROR | LIBXML_NOWARNING);
+	if ($doc->documentElement === null) {
+		return '';
+	}
+	$xpath = new DOMXPath($doc);
+	$bases = $xpath->evaluate('//base');
+	if ($bases instanceof DOMNodeList && $bases->length > 0) {
+		foreach ($bases as $base) {
+			if ($base instanceof DOMElement) {
+				$base->setAttribute('href', $href);
```
Alkarex: This part is wrong. I will fix tomorrow

Alkarex: Done

app/Models/Entry.php
Alkarex: Help welcome in particular to test with pages using an encoding different from UTF-8.

Inverle: Unfortunately I'm not aware of other feeds with this exact problem other than https://blog.thea.codes/feed.xml
(unless you mean feeds in general)

The fix does work for https://blog.thea.codes/feed.xml and hasn't broken any of my other feeds.

Alkarex: Tested with Windows-1252 and is also fine

Inverle: Do you know any feeds that are Windows-1252?

Alkarex: I just made a test myself. But the thing to be tested was not the feed itself, but the targeted HTML pages

lib/lib_rss.php
```diff
@@ -507,12 +507,47 @@ function enforceHttpEncoding(string $html, string $contentType = ''): string {
...
+		$base->setAttribute('href', $href);
+		$head = null;
+		$heads = $xpath->evaluate('//head');
+		if ($heads instanceof DOMNodeList && $heads->length > 0) {
+			$head = $heads->item(0);
+		}
+		if ($head instanceof DOMElement) {
+			$head->insertBefore($base, $head->firstChild);
+		} else {
+			$doc->insertBefore($base, $doc->documentElement->firstChild);
```
dotsam, line 538: There seems to be an error in this code path, changing it to `$doc->documentElement->insertBefore($base, $doc->documentElement->firstChild);` seems to resolve it for me, otherwise it throws a `Fatal error: Uncaught DOMException: Not Found Error`

Alkarex: Thanks @dotsam . Would you feel like sending a PR?

Alkarex: https://github.com/FreshRSS/FreshRSS/pull/7741

PR #7675, opened 2025-06-17
Fix #7648: Normalize feed URLs to prevent duplicates

### LINKED ISSUE #7648: [Feature] Show a warning when subscribing to an already added feed

I accidentally subscribed twice to the same RSS feed (same URL) and placed it in the same category. FreshRSS allowed the subscription without any warning, and as a result, I started seeing all articles duplicated in my feed reader. This can be confusing and creates unnecessary clutter.

### Use case

Users who manage many feeds might unintentionally try to subscribe again to a feed they already follow — for example, while reorganizing categories or using the subscription form in a rush. Currently, there’s no way to know you’re duplicating a subscription unless you manually check your list or notice the duplicated articles later

I’d like FreshRSS to display a clear warning or confirmation message when a user tries to subscribe to a feed that’s already present in their subscriptions (ideally even within another category). For example:

“You are already subscribed to this feed. Do you want to continue?”

At minimum, a passive notification (like a message at the top of the subscription page) would help

I currently have to manually check the list of subscribed feeds before adding a new one. This works, but it’s tedious, especially with a long list or when using a mobile device. I also tried relying on feed deduplication, but that doesn’t prevent double subscriptions.

### Additional context

_No response_

jnarusis:
Closes #7648

    Adds a normalizeUrl() function to Minz/Helper.php for consistent feed URL comparison

    Applies normalization in feedController::subscribeAction()

    Prevents duplicate subscriptions caused by slight URL variations (e.g., trailing slash, scheme)

    Attempt to subscribe to the same feed using different formats:

        http://example.com/feed

        http://example.com/feed/

        https://example.com/feed

        https://example.com/feed/

    Observe that after the first subscription, duplicates are prevented with the “already subscribed” message

    Confirm no regression in normal feed addition behavior

Pull request checklist:

- [x] Clear commit messages
- [x] Code manually tested
- [ ] Unit tests written (optional if too hard)
- [ ] Documentation updated

### DIFF
--- a/app/Models/FeedDAO.php
+++ b/app/Models/FeedDAO.php
@@ -1,6 +1,8 @@
 <?php
 declare(strict_types=1);
 
+require_once __DIR__ . '/FeedHelper.php';
+
 class FreshRSS_FeedDAO extends Minz_ModelPdo {
 
 	protected function addColumn(string $name): bool {
@@ -83,11 +85,12 @@ class FreshRSS_FeedDAO extends Minz_ModelPdo {
 
 	public function addFeedObject(FreshRSS_Feed $feed): int|false {
 		// Add feed only if we don’t find it in DB
-		$feed_search = $this->searchByUrl($feed->url());
+		$normalized_url = FeedHelper::normalizeUrl($feed->url());
+$feed_search = $this->searchByUrl($normalized_url);
 		if ($feed_search === null) {
 			$values = [
 				'id' => $feed->id(),
-				'url' => $feed->url(),
+				'url' => FeedHelper::normalizeUrl($feed->url()),
 				'kind' => $feed->kind(),
 				'category' => $feed->categoryId(),
 				'name' => $feed->name(true),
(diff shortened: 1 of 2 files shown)


Alkarex:
Hello,
I think it would be better to show a warning and a confirmation dialogue in case of potential duplicate, allowing users to proceed anyway if they want

Frenzie:
This is only on adding a feed, correct? Concretely, we wouldn't want it to keep being normalized to `/feed/` while the site keeps redirecting that to `/feed` for example. But as far as that goes, will that get in the way of how it's intended to function?

Alkarex:
> This is only on adding a feed, correct? Concretely, we wouldn't want it to keep being normalized to `/feed/` while the site keeps redirecting that to `/feed` for example. But as far as that goes, will that get in the way of how it's intended to function?

Indeed, that is also why I am proposing to only have a confirmation dialogue in case of likely duplicate, without affecting the real URL in DB, which should not be changed

PR #7682, opened 2025-06-20
Restore user config from backup when it doesn't exist

### LINKED ISSUE #7681: [Bug] User config save seems to be unsafe
### Describe the bug

This isn't the first time this happens but is quite random. Today I restarted my docker containers and upon restarting freshrss, I couldn't log in and got an `Error during context user init!` page.

After investigating a big, I see that the `config.php` file of my users disappeared and the `config.php.bak.php` has the time of the container restart : 
```bash
xxxx@DS920:/volume1/DockerData/FreshRSS/config/users$ ll rakambda
total 1.1M
drwxrwxr-x 1 docker.freshrss 33    70 Jun 20 18:51 .
drwxrwxr-x 1 docker.freshrss 33    84 Mar 28 20:12 ..
-rw-rw-r-- 1              33 33  4.6K Jun 20 18:50 config.php.bak.php
drwxrwxr-x 1 docker.freshrss 33    34 Dec  7  2024 extensions
-rw-rwxr-- 1 docker.freshrss 33 1020K Jun 20 18:50 log.txt
```

The backup of an hour before the restart shows the file being present : 
```bash
xxxx@DS920:/volume1/DockerData/#snapshot/GMT+02-2025.06.20-16.00.02/FreshRSS/config/users$ ll rakambda
total 1.1M
drwxrwxr-x 1 docker.freshrss 33    90 Jun 20 15:40 .
drwxrwxr-x 1 docker.freshrss 33    84 Mar 28 20:12 ..
-rw-r--r-- 1              33 33  4.6K Jun 20 15:40 config.php
-rw-r--r-- 1              33 33  4.6K Jun 20 15:35 config.php.bak.php
drwxrwxr-x 1 docker.freshrss 33    34 Dec  7  2024 extensions
-rw-rwxr-- 1 docker.freshrss 33 1020K Jun 20 15:40 log.txt
```

Also i was using freshrss through the website seconds before the restart and wasn't getting the error page.

I'm not totally sure this is where the issue is, but looking at the save function, it seems to be able to cause the issue (and from the github research seems to be the only place with a `.bak`) : https://github.com/FreshRSS/FreshRSS/blob/a6948218fb1c66fe146c7651555e5a1f791c8112/lib/Minz/Configuration.php#L204-L211

If the `rename` happens but the `file_put_contents` never completes or fails, then we're left with a missing `config.php` file.

I wonder if there is an intentional reason for the rename to be used ?

A simple copy `config.php` -> `config.php.bak.php` could be done and then `file_put_contents` is made atomically ? This would allow the original `config.php` file to still be present if the write never happens.

---

Note: Restoring the `config/users` folder from the backup solved the issue.

Note: I tried to look around the user logs  and `_` logs but didn't seem to find anything related to this issue. No logs were made around the time of the reboot (maybe it didn't get the time to be written?). Maybe I'll try to set the logs in debug mode in case it happens again. But as it has a random aspect don't know when it may happen.

### To Reproduce

1. Trigger a user config save
2. Somehow manage to stop the app while the save is still ongoing

### Expected behavior

User config not disapeparing

### FreshRSS version

1.26.3

### System information

- Database version: MariaDB Docker / 11.4.5-MariaDB-log - Alpine Linux
- PHP version: 8.2.28
- Installation type: Docker `freshrss/freshrss:latest` `f9efa6d86798`
- Web server type: Synology NAS - nginx reverse proxy
- Device: PC
- OS: Win11
- Browser: Vivaldi 7.5.3731.3

### Additional context

Docker compose entry : 

```yaml
  freshrss:
    <<: *default
    image: "freshrss/freshrss:latest"
    container_name: "fresh-rss"
    depends_on:
      mariadb:
        condition: service_healthy
        restart: false
      rss-hub:
        condition: service_started
        restart: false
      loki:
        condition: service_started
        restart: false
    env_file:
      - "/volume1/docker/FreshRSS/container.env"
    volumes:      
      - "/volume1/docker/FreshRSS/constants.local.php:/var/www/constants.local.php"
      - "/volume1/DockerData/FreshRSS/cache:/cache"
      - "/volume1/DockerData/FreshRSS/config:/var/www/FreshRSS/data"
      - "/volume1/DockerData/FreshRSS/extensions:/var/www/FreshRSS/extensions"
    ports:
      - "8218:8080"
    logging: *loggingLoki
```

container.env : 
```
TZ=Europe/Paris
LOG_LEVEL=info

LISTEN=8080
CRON_MIN=*/5
FRESHRSS_ENV=production
```

constants.local.php : 
```php
<?php
define('MAX_LOG_SIZE', 104857600);
```

Inverle:
Fixes #7681

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -17,14 +17,17 @@ class FreshRSS_user_Controller extends FreshRSS_ActionController {
 
 	public static function userExists(string $username): bool {
 		$config_path = USERS_PATH . '/' . $username . '/config.php';
-		if (!@file_exists($config_path) && @file_exists($config_path . '.bak.php')) {
-			Minz_Log::warning('Config for user ' . $username . ' not found. Attempting to restore from backup.', ADMIN_LOG);
+		if (@file_exists($config_path)) {
+			return true;
+		} elseif (@file_exists($config_path . '.bak.php')) {
+			Minz_Log::warning('Config for user “' . $username . '” not found. Attempting to restore from backup.', ADMIN_LOG);
 			if (!copy($config_path . '.bak.php', $config_path)) {
 				@unlink($config_path);
 				return false;
 			}
+			return @file_exists($config_path);
 		}
-		return @file_exists($config_path);
+		return false;
 	}
 
 	/** @param array<string,mixed> $userConfigUpdated */
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -17,7 +17,7 @@ class FreshRSS_user_Controller extends FreshRSS_ActionController {
 
 	public static function userExists(string $username): bool {
 		$config_path = USERS_PATH . '/' . $username . '/config.php';
-		if (!file_exists($config_path) && file_exists($config_path . '.bak.php')) {
+		if (!@file_exists($config_path) && @file_exists($config_path . '.bak.php')) {
 			Minz_Log::warning('Config for user ' . $username . ' not found. Attempting to restore from backup.', ADMIN_LOG);
 			copy($config_path . '.bak.php', $config_path);
 		}
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -19,7 +19,9 @@ class FreshRSS_user_Controller extends FreshRSS_ActionController {
 		$config_path = USERS_PATH . '/' . $username . '/config.php';
 		if (!@file_exists($config_path) && @file_exists($config_path . '.bak.php')) {
 			Minz_Log::warning('Config for user ' . $username . ' not found. Attempting to restore from backup.', ADMIN_LOG);
-			copy($config_path . '.bak.php', $config_path);
+			if (!copy($config_path . '.bak.php', $config_path)) {
+				@unlink($config_path);
+			}
 		}
 		return @file_exists($config_path);
 	}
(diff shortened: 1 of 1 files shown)


app/Controllers/userController.php
```diff
@@ -16,7 +16,14 @@ public static function checkUsername(string $username): bool {
 	}
 
 	public static function userExists(string $username): bool {
-		return @file_exists(USERS_PATH . '/' . $username . '/config.php');
+		$config_path = USERS_PATH . '/' . $username . '/config.php';
+		if (!@file_exists($config_path) && @file_exists($config_path . '.bak.php')) {
+			Minz_Log::warning('Config for user ' . $username . ' not found. Attempting to restore from backup.', ADMIN_LOG);
+			if (!copy($config_path . '.bak.php', $config_path)) {
+				@unlink($config_path);
```
Frenzie, line 25: Maybe followed by an immediate return false instead of relying on the line below?

Frenzie (approved):

app/Controllers/userController.php
Inverle: Sure

Alkarex:
Slight rewrite to avoid reducing the performance of the main case. I hope it is OK
https://github.com/FreshRSS/FreshRSS/pull/7682/commits/d8623291f800c51ec37758e2b6dc965a25dcc26d

Inverle:
Sure

PR #7683, opened 2025-06-21
i18n:ID

the7thNightmare:
The translation for Indonesian Language.

### DIFF
--- a/app/i18n/id/conf.php
+++ b/app/i18n/id/conf.php
@@ -12,7 +12,7 @@
 
 return array(
 	'archiving' => array(
-		'_' => 'Mengarsipkan',
+		'_' => 'Pengarsipan',
 		'exception' => 'Pengecualian Pembersihan',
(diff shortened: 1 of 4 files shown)


the7thNightmare:
Also, what is "Article Unicity Criteria"? I tried that thing but I still don't know what that do, so I can't really translate that.

Frenzie:
It's to decide uniqueness, as in when an article is updated or when it's a brand new one that should be show.

the7thNightmare:
Thanks!

Alkarex:
@the7thNightmare Is it good enough? https://github.com/FreshRSS/FreshRSS/pull/7683/commits/dd2e7ce975eb03795a451cdb7079f8b2d07c85cd

Frenzie:
Feels more like a bug in the check translation script?

Alkarex:
> Feels more like a bug in the check translation script?

Yes, but that workaround is easier :-)

PR #7684, opened 2025-06-22
Use `X-Requested-With` header instead of the `ajax` parameter

Inverle:
Prevents unintended use of `ajax` in requests.


Inverle:
Problem:
`Cache-Control` conflicts with these changes.
When the user opens a slider, and later the page then they get the slider content instead of the page content.
I wasn't able to fix this with [`Vary`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary)
edit: fixed by appending value of `X-Requested-With`  to `Etags`

also TODO: update docs so they say to use the `X-Requested-With` header instead of `ajax` parameter~~
edit: done

Alkarex:
See also
https://github.com/FreshRSS/FreshRSS/blob/51298cd6bc100b1cc6508abb602a59c01a9e2c04/docs/en/users/09_refreshing_feeds.md#L61-L84

Alkarex:
I think we should start by forbidding the `ajax` mode for problematic cases. Moving to an HTTP header might not be necessary

Inverle:
> See also

~~Yes I've seen it but there are too many files to change, so I left it for now~~
edit: done

> I think we should start by forbidding the ajax mode for problematic cases

We can't do that because it's needed for the slider (unless we remove it)
~~Maybe just omit the cache on those requests?~~

~~One other approach is to disable the buttons in HTML and later enable them with JS, but it's not the best one.~~
edit: fixed, see above

Inverle:
@Alkarex Should I generate the .po files too?
It looks like they haven't been in sync for a while now

Frenzie:
Now that po4a has a sufficiently high version on GHA I was hoping to automate something about that but I haven't found the time. A few years ago I'd have needed to add compilation or a custom Docker image which seemed like too much effort but perhaps that was a bad choice because of course everybody forgets or doesn't even know about it. :-)

app/actualize_script.php
```diff
@@ -12,8 +12,8 @@
 // Set the header params ($_GET) to call the FRSS application.
 $_GET['c'] = 'feed';
 $_GET['a'] = 'actualize';
-$_GET['ajax'] = 1;
 $_GET['maxFeeds'] = PHP_INT_MAX;
+$_SERVER['HTTP_X_REQUESTED_WITH'] = 'XMLHttpRequest';
```
Alkarex:
```suggestion
$_SERVER['HTTP_X_AJAX'] = '1';
```

I am guessing this has been forgotten from previous changes

app/Controllers/statsController.php
```diff
@@ -101,9 +101,9 @@ public function indexAction(): void {
 	 */
 	public function feedAction(): void {
 		$id = Minz_Request::paramInt('id');
-		$ajax = Minz_Request::paramBoolean('ajax');
+		$ajax = Minz_Request::isAjax();
 		if ($ajax) {
-			$url_redirect = ['c' => 'subscription', 'a' => 'feed', 'params' => ['id' => (string)$id, 'from' => 'stats', 'ajax' => (string)$ajax]];
+			$url_redirect = ['c' => 'subscription', 'a' => 'feed', 'params' => ['id' => (string)$id, 'from' => 'stats']];
```
Alkarex: The two conditions are now identical

lib/http-conditional.php
```diff
@@ -105,7 +105,8 @@ function httpConditional(int $UnixTimeStamp, int $cacheSeconds = 0, int $cachePr
 		$_sessionMode = $session;
 		$myQuery .= print_r($_SESSION, true) . session_name() . '=' . session_id();
 	}
-	$etagServer = '"' . md5($scriptName . $myQuery . '#' . $dateLastModif) . '"';
+	$isAjax = is_string($_SERVER['HTTP_X_AJAX'] ?? null) ? $_SERVER['HTTP_X_AJAX'] : ''; // FreshRSS
```
Alkarex: I am not super happy about this customisation.
Would it be possible to refactor using the standard HTTP `Vary` behaviour?

app/actualize_script.php
Inverle: Indeed

lib/http-conditional.php
Inverle: `Vary` header doesn't work. See above: https://github.com/FreshRSS/FreshRSS/pull/7684#issuecomment-2993835596
I was also considering the `Accept` header, since XMLHttpRequest/fetch send `*/*` by default, but I'd like an opinion.

app/Controllers/statsController.php
Inverle: What should be done about this?

Alkarex: Since the `if` and the `else` are identical, the whole `if ($ajax)` logic can be removed

lib/http-conditional.php
Frenzie: Regarding the name, isn't `HTTP_X_REQUESTED_WITH` or something like it preferable? Naming it AJAX seems a bit confusing.

I assume all of these AJAX/Fetch requests do send a CSRF token?

Inverle: I thought it would be more confusing if we were to have to send `X-Requested-With` using `fetch()`
Should it also have the value of `XMLHttpRequest` in that case?

Frenzie: I wouldn't put in some irrelevant implementation detail. `FreshRSS`, possibly with some affix, seems like the most obvious thing to put in.

Although I have to say I don't understand what difference it's supposed to make — accidentally put in ajax=1, accidentally put in X_AJAX, still the same mistake, still just as easily spoofed.

Inverle: https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-j66v-hvqx-5vh3

Inverle: > I wouldn't put in some irrelevant implementation detail. `FreshRSS`, possibly with some affix, seems like the most obvious thing to put in.
> 
> Although I have to say I don't understand what difference it's supposed to make — accidentally put in ajax=1, accidentally put in X_AJAX, still the same mistake, still just as easily spoofed.

I changed the header name to `X-Requested-With` with required value of `FreshRSS`

Frenzie (approved):
lgtm, probably needs more testing though ;-)

Alkarex:
Sorry 😓, but I am not so glad of this PR and would like to pursue an alternative instead:
* https://github.com/FreshRSS/FreshRSS/pull/7771 

My reasoning is that:
* A new HTTP header will break a bunch of existing FreshRSS deployments, many of which are behind a proxy that forwards only specific headers (see the numerous tickets we have had related to nginx, to the headers for the GReader API, turn-key solutions such as Yunohost, etc.). While acceptable if very needed, I would like to make efforts to avoid it.
* Buttons marked with `confirm` (at least if loaded from JavaScript) should be marked as `disabled` anyway until the JavaScript event that confirms when clicking on them is fully loaded / active.
* The current solution works (mostly) without JavaScript (using HTTP Basic Auth login), and this PR breaks (some of) that. While this is not a must have, I like to keep essential features working without JavaScript for a little longer (I have a plan for a parallel all-JS UI, for that is for later, and without removing the mostly-HTML UI)
* Not super important, but I am nevertheless not so happy of customising the `http-conditional` library with some non-standard conditions.

I hope that makes sense 🙈

Inverle:
That's fine, I do realize that my PR isn't necessarily the best solution given these circumstances.

> Buttons marked with confirm (at least if loaded from JavaScript) should be marked as disabled anyway until the JavaScript event that confirms when clicking on them is fully loaded / active.

I think it would be a good addition to also require a separate "CSRF token" for AJAX requests, just as @Frenzie suggested [before](https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-j66v-hvqx-5vh3#advisory-comment-133928), to cover other buttons than those marked with `confirm()`

Alkarex:
> I think it would be a good addition to also require a separate "CSRF token" for AJAX requests

Yes, good idea. Help welcome

PR #7697, opened 2025-06-26
Docblock update for getFileUrl

tryallthethings:

PHPStan was complaining about my code, although it should be correct due to my understanding:

```php
Minz_View::appendStyle($this->getFileUrl('style.css', 'css')); 
Minz_View::appendScript($this->getFileUrl('script.js', 'js'), false, true, false);
```

```phpstan
 ------ ----------------------------------------------------------------------- 
  Line   extension.php                                                          
 ------ ----------------------------------------------------------------------- 
  80     Parameter #2 $type of method Minz_Extension::getFileUrl() expects '',  
         'css' given.                                                           
         🪪 argument.type                                                       
  81     Parameter #2 $type of method Minz_Extension::getFileUrl() expects '',  
         'js' given.                                                            
         🪪 argument.type                                                       
  82     Parameter #2 $type of method Minz_Extension::getFileUrl() expects '',  
         'js' given.                                                            
         🪪 argument.type                                                       
 ------ ----------------------------------------------------------------------- 
```

1. Run PHPStan on PHP code containing the above calls

# Pull request checklist:

- [X] clear commit messages
- [ ] code manually tested
 -> No idea how to do that. PHPStan is new to me
- [ ] documentation updated

### DIFF
--- a/lib/Minz/Extension.php
+++ b/lib/Minz/Extension.php
@@ -207,7 +207,7 @@ abstract class Minz_Extension {
 	 *
 	 * @param string $filename The name of the file.
 	 * @param string $type Optional. The type of file (e.g., 'css', 'js'), used as a URL parameter. Defaults to an empty string.
-	 * @param bool $add_version Optional. Add a version parameter to prevent caching. `true` by default.
+	 * @param bool   $isStatic indicates if the file is a static file or a user file. Default is static.
 	 * @return string The URL of the extension file.
 	 */
 	final public function getFileUrl(string $filename, string $type = '', bool $isStatic = true): string {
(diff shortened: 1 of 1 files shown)


Alkarex:
Hello,
The current doc is on purpose, to catch the attention on the fact that the second parameter is deprecated and should not be used anymore.
Best to use named parameters, or if not possible, pass a `''` blank string.

Example:
https://github.com/FreshRSS/FreshRSS/blob/7c57f38008136202ba7b38e3154ac87be4eefb68/lib/core-extensions/UserCSS/extension.php#L14-L15

tryallthethings:
Okay, fair enough. I thought this was an oversight.

PR #7703, opened 2025-06-29
Fix multiple auth headers bug

### LINKED ISSUE #7699: [Bug] Remote-User Header Not Being Passed Correctly
### Describe the bug

Since version [1.26.2](https://github.com/FreshRSS/FreshRSS/releases/tag/1.26.2) and later, I recieve the error:

> Error 403 - Forbidden
> 
> You don’t have permission to access this page [HTTP Remote-User= ; Remote IP address=172.20.0.6]

Note how there is only a space after 'Remote-User=', that's not me removing something for privacy, that's just what it shows.
Authelia OIDC functions with other servicies so I don't believe that the issue lies with that.

I have tested with versions 1.25.0 to 1.26.1 with everything else being the same, and it works fine and showed the remote user header being passed correctly so I believe there is some regression in 1.26.2

The docker logs give this error:

> [warning] --- Multiple HTTP authentication headers!

### To Reproduce

1. Enable FreshRSS version 1.26.2 or later
2. Attemp to login in
3. Authelia will ask if you want to give FreshRSS permission, accept
4. The error should occur

### Expected behavior

Should seemlessly go to the feed page of the authelia user after login.

### FreshRSS version

1.26.2 and later

### System information

- Database version: SQLite
- PHP version: Sorry I don't know how to check or change this - the default for the image I assume
- Installation type: Docker
- Web server type: Whatever is default for the debian image - I don't know how to check this either.
- Device: PC
- OS: Windows 10
- Browser: both Firefox and Chrome

### Additional context

_No response_

Inverle:
Closes #7699

for example when we have `HTTP_REMOTE_USER` and `REMOTE_USER` set to the same value, we don't want to treat them as separate headers

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -857,7 +857,7 @@ function checkTrustedIP(): bool {
 
 function httpAuthUser(bool $onlyTrusted = true): string {
 	$check_array = array_intersect_key($_SERVER, array_flip(['REMOTE_USER', 'REDIRECT_REMOTE_USER', 'HTTP_REMOTE_USER', 'HTTP_X_WEBAUTH_USER']));
-	$auths = array_filter(array_unique($check_array));
+	$auths = array_unique($check_array);
 	if (count($auths) > 1) {
 		Minz_Log::warning('Multiple HTTP authentication headers!');
 		return '';
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -856,7 +856,8 @@ function checkTrustedIP(): bool {
 }
 
 function httpAuthUser(bool $onlyTrusted = true): string {
-	$auths = array_intersect_key(array_unique($_SERVER), ['REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => '', 'HTTP_REMOTE_USER' => '', 'HTTP_X_WEBAUTH_USER' => '']);
+	$check_array = array_intersect_key($_SERVER, array_flip(['REMOTE_USER', 'REDIRECT_REMOTE_USER', 'HTTP_REMOTE_USER', 'HTTP_X_WEBAUTH_USER']));
+	$auths = array_filter(array_unique($check_array));
 	if (count($auths) > 1) {
 		Minz_Log::warning('Multiple HTTP authentication headers!');
 		return '';
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -856,7 +856,7 @@ function checkTrustedIP(): bool {
 }
 
 function httpAuthUser(bool $onlyTrusted = true): string {
-	$auths = array_intersect_key($_SERVER, ['REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => '', 'HTTP_REMOTE_USER' => '', 'HTTP_X_WEBAUTH_USER' => '']);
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Fine by me

Inverle:
I need to improve this approach a bit
edit: done

lib/lib_rss.php
```diff
@@ -856,7 +856,8 @@ function checkTrustedIP(): bool {
 }
 
 function httpAuthUser(bool $onlyTrusted = true): string {
-	$auths = array_intersect_key($_SERVER, ['REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => '', 'HTTP_REMOTE_USER' => '', 'HTTP_X_WEBAUTH_USER' => '']);
+	$check_array = array_intersect_key($_SERVER, array_flip(['REMOTE_USER', 'REDIRECT_REMOTE_USER', 'HTTP_REMOTE_USER', 'HTTP_X_WEBAUTH_USER']));
+	$auths = array_unique($check_array);
```
```suggestion
	$auths = array_intersect_key($_SERVER, ['REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => '', 'HTTP_REMOTE_USER' => '', 'HTTP_X_WEBAUTH_USER' => '']);
	$auths = array_unique($auths);
```

If possible, I would prefer the previous version than with `array_flip()`, to avoid a function call and arguably clearer

Alkarex: Hum, I am not sure this is right. We need to respect priorities. `REMOTE_USER` has precedence over `HTTP_REMOTE_USER`, since `REMOTE_USER` is safer (cannot be set by HTTP headers).

Alkarex: For instance, if `REMOTE_USER` gets removed and `HTTP_REMOTE_USER` remains, the behaviour will be different depending on the trusted IPs.

Inverle: > Hum, I am not sure this is right. We need to respect priorities. `REMOTE_USER` has precedence over `HTTP_REMOTE_USER`, since `REMOTE_USER` is safer (cannot be set by HTTP headers).

I think it's correct, we're now excluding headers with the same value before checking if there are multiple headers i.e. to pass the check if both `REMOTE_USER` and `HTTP_REMOTE_USER` are set, but with the same value

> For instance, if `REMOTE_USER` gets removed and `HTTP_REMOTE_USER` remains, the behaviour will be different depending on the trusted IPs.

The check doesn't affect below code (unless it fails, which it would if they're different username values): https://github.com/FreshRSS/FreshRSS/blob/0bca0d8afcf2509d7b227450450d66c5348e9e1a/lib/lib_rss.php#L860-L880

Alkarex: If I am not mistaken, if `REMOTE_USER` gets removed and `HTTP_REMOTE_USER` remains, then the test will pass only if the trusted IP logic works, as opposed to before when `REMOTE_USER` always worked independently of the trusted IP logic

Inverle: I'm not exactly sure what the issue could be here, since `REMOTE_USER` would be deleted from `$auths` and not `$_SERVER`

What should I be doing instead?

Alkarex: > I'm not exactly sure what the issue could be here, since `REMOTE_USER` would be deleted from `$auths` and not `$_SERVER`

You are right. It should be fine

PR #7717, opened 2025-07-05
Fix custom icons not displaying

Inverle:
Started happening recently, I haven't noticed this behavior before.


p/f.php
```diff
@@ -48,6 +52,8 @@ function show_default_favicon(int $cacheSeconds = 3600): void {
 	}
 }
 
+displayIcon:
```
Alkarex: Could you please refactor, for instance by making that a function?

Inverle: Sure [db05443](https://github.com/FreshRSS/FreshRSS/pull/7717/commits/db05443739728b376a4268e6edd6747665348d94)

Alkarex: Much cleaner, thanks

PR #7719, opened 2025-07-05
Include enclosures in entries hash

### LINKED ISSUE #7718: [Bug] Enclosures are not updated in item attributes
### Describe the bug

When the upstream feed updates the list and properties of enclosures (in my case, specifically, URLs with frequently updated access tokens), the new properties are present in cache, but never updated in the item's database entry.

While enclosures are designed to provide a way to pre-fetch and cache attachments to feed items, this is not done by FreshRSS. This means that, for example, an enclosure can contain a link that will become invalid, and the media will be unreachable.

### To Reproduce

1. Subscribe to a feed
2. Fetch items on the feed and retrieve at least one with an `<enclosure>`
3. Wait until the feed updates properties in the `<enclosure>` of the item (e.g. URL)
4. When displayed or viewed in database, the enclosures show outdated information

### Expected behavior

When an item is fetched, the enclosures attributes are updated, just like the content is.

### FreshRSS version

1.26.3

### System information

- Database version: PostgreSQL 15.13
- PHP version: 8.2.28
- Installation type: Docker
- Web server type: nginx
- Device: all
- OS: Debian 12
- Browser: all (at least Firefox and latest Chromiun, but probably irrelevant)

### Additional context

i checked some [best practices](https://www.rssboard.org/rss-profile#element-channel-item-enclosure) around `<enclosure>` to see if it even made sense to update them, and it does not seem to be a discouraged practice.

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/7718
Negative side-effect: users using the option to automatically mark updated articles as unread will have some articles with enclosures appear as unread


Lymkwi:
Bit of a hack for testing but i manually added the lines on my personal deployment environment where i noticed a feed with outdated enclosure URLs and refreshed all feeds: from what i see, enclosures were correctly updated! i worried older items marked as read could be marked unread suddenly because of hash changes but no such problem happened.

TL;DR works for my specific problem in #7718 .

PR #7728, opened 2025-07-10
Improve sharing via Print

Inverle:
List of changes:
* The temporary document for printing is now in an `<iframe>` instead of a new tab
* The whole `<head>` element is copied to the temporary document, except for `<script>` tags to copy over the `<meta>` tags as well
* URLs that contain the instance base URL are now removed from the printed PDF
* The saved filename (PDF) will now default to the article title
* `<details>` is auto expanded
* Styling:
   * The main document's `<html>` class is copied over to preserve some styling that might use those classes
   * Instead of writing `content_el.innerHTML` to the temporary document, `content_el.outerHTML` is now written instead to apply the styles that select `.content`
   * `.dropdown-menu` is now hidden in the printed document, because it can't be expanded anyway
   * Headers and footers are hidden in the printed document
* The printed document will now display correctly all the time, by waiting for it to load before calling `print()`
   * Before, the stylesheets might've not finished loading and the document was broken
* Better browser support on mobile for this feature
   * Before, the document would fail to print on Chrome Mobile
   
Tested on:
* Firefox - both desktop and mobile, works ✅
* Chrome - both desktop and mobile, works ✅
* Opera - desktop, works (same as Chrome) ✅
* Brave - both desktop and mobile (same as Chrome), works ✅
* Safari - both desktop and mobile, works✅
* Microsoft Edge - both desktop and mobile, works ✅
* GNOME Web - desktop, works ✅
* SeaMonkey - desktop, works ✅

Known issues:
* Images may not finish loading the first time the print dialog is opened

TODO:
* [x] Test on Safari
* [x] Try to fix GNOME Web


p/scripts/main.js
```diff
@@ -1232,24 +1239,79 @@ function init_stream(stream) {
...
+			</head>
+			<body>
+				${content_el.outerHTML}
+			</body>
+			</html>
+			`;
+			document.body.prepend(print_frame);
+
+			const tmp_window = print_frame.contentWindow;
+			tmp_window.onload = () => tmp_window.print();
```
Frenzie: It doesn't work like this?
```suggestion
			tmp_window.onload = tmp_window.print;
```

p/scripts/main.js
```diff
@@ -1232,24 +1239,79 @@ function init_stream(stream) {
...
+
+			tmp_window.onafterprint = () => {
+				const isFirefoxMobile = navigator.userAgent.includes('Firefox') && navigator.userAgent.includes('Mobile');
+				if (isFirefoxMobile) {
+					// Setting location.hash here crashes the tab on Firefox Mobile
+					return;
+				}
+				// Required for `mouseover` to trigger after print dialog is no longer open
+				location.hash = 'close';
+			};
```
Frenzie: > The temporary document for printing is now in an <iframe> instead of a new tab

Perhaps this is one of those personal preference things but I'm not overly fond of that. I wasn't going to say anything because it didn't sound like a big deal, but um, well, these lines kind of speak for themselves. What is the reason for wanting to use an iframe exactly?

p/themes/base-theme/frss.css
```diff
@@ -2779,13 +2779,26 @@ html.slider-active {
...
+	/* This hides the headers and footers in the printed document on Chrome */
+	/* Supported since Chrome 131: https://developer.chrome.com/release-notes/131#page_margin_boxes */
+
+	@page {
+		/* Firefox and Safari do not support those yet */
+		/* See: https://developer.mozilla.org/en-US/docs/Web/CSS/@page#browser_compatibility */
+
+		@top-left { content: ''; }
+
+		@bottom-left { content: ''; }
```
Frenzie: How does this interact with RTL?

Edit: I see, it doesn't at the moment.

Alkarex:
For the record, I like to ensure basic support for [SeaMonkey](https://www.seamonkey-project.org/) (seems to work, at a quick glance)

p/scripts/main.js
Inverle: > well, these lines kind of speak for themselves

Sure having to check for the user agent is a bit ridiculous but it was a quick fix, if that's what you mean.
I'll try something else instead, maybe I don't have to set `location.hash` at all.

> What is the reason for wanting to use an iframe exactly?

- The temporary document for printing is now in an `<iframe>` instead of a __***new tab***__

So, the reason is I don't like that printing causes a new tab to be opened.

p/themes/base-theme/frss.css
Inverle: Could you give an example?

Frenzie: The .rtl stylesheet hasn't swapped left and right.

Alkarex: Maybe just add `@top-right` here as well

p/scripts/main.js
Inverle: Is this any better?
https://github.com/FreshRSS/FreshRSS/blob/7790e649cad18697c0a8bfb62ab2901bd5866c21/p/scripts/main.js#L1294-L1316

p/themes/base-theme/frss.css
Inverle: I don't think this changes anything but I added `@top-right` and `@bottom-right`

p/scripts/main.js
Frenzie: > So, the reason is I don't like that printing causes a new tab to be opened.

Well sure, but I prefer a new tab to be opened for printing and I don't quite understand why you don't. :-) I'd much prefer it didn't automatically close! You don't even notice it opens a new tab and I think that's a bad thing. The print view is perfect to save as HTML too for example. Just a single line removal to get rid of the auto-close would make it much more user-friendly in that sense. The fact that save page as works in print preview in most browsers nowadays feels more like a fluke than something well thought out. And mind, after testing I realized save page doesn't save the print preview from an iframe but sticks to the parent page.

So from my perspective you have a lesser user experience at the cost of lower quality code with unnecessary hacks and complexities.

As a compromise, perhaps it should be separated into "print" and "show print view" or something to that effect. Though I've always thought auto-print but don't auto-close serves both purposes well enough. But given the code above I don't see any role for an iframe there, only a question of whether to auto-print and auto-close.

p/themes/base-theme/frss.css
Frenzie: Even if it doesn't right this moment because of lacking browser `@page` support I imagine they'll add it eventually. In any case I think the main possibilities are the following:

1. The browser UI language is set to an RTL language
2. `<html dir="rtl">` 
3. CSS direction override.

Incidentally, I just realized that instead of setting it to empty wouldn't it be better to set it to the value of the original URL?

p/scripts/main.js
Inverle: > Well sure, but I prefer a new tab to be opened for printing and I don't quite understand why you don't. :-) 

The page that is currently being open for printing looks odd, because you can see a blank page behind the print dialog and the tab doesn't have a title.

> The print view is perfect to save as HTML too for example

I didn't realize it's possible to do that in the print preview. I guess that makes opening a new tab a better option automatically.

> Just a single line removal to get rid of the auto-close would make it much more user-friendly in that sense

I think that this should be configurable (as a matter of preference)

> So from my perspective you have a lesser user experience at the cost of lower quality code with unnecessary hacks and complexities.

That's the worst part, it shouldn't be this hard to print from an iframe.
It's probably also not guaranteed that this will continue working after browser updates.

> As a compromise, perhaps it should be separated into "print" and "show print view" or something to that effect. Though I've always thought auto-print but don't auto-close serves both purposes well enough. But given the code above I don't see any role for an iframe there, only a question of whether to auto-print and auto-close.

That's an option too, but maybe something like `Save as HTML` instead of `show print view`?

Also, for showing a print view, maybe it should just be a separate .phtml view instead of opening it with JS

p/themes/base-theme/frss.css
Inverle: > The browser UI language is set to an RTL language

I think I will have to test to be sure.

> Incidentally, I just realized that instead of setting it to empty wouldn't it be better to set it to the value of the original URL?

Possibly

p/scripts/main.js
Frenzie: > That's the worst part, it shouldn't be this hard to print from an iframe.

It does surprise me.

> That's an option too, but maybe something like Save as HTML instead of show print view?

Possibly. I'd say the point is simply "show single article" (in clean print view) and what the user does is up to them, regardless if I've thought of it. Put another way, I wouldn't really want someone to optimize the saving by using blob for example, because then that's all you can do. Though granted, you could then do whatever you want with the resulting file so maybe it doesn't matter too much.

> Also, for showing a print view, maybe it should just be a separate .phtml view instead of opening it with JS

No objections here, but overall I'm totally fine with JS as in this PR as well. More than that, it seems done to me, besides the minor iframe detail.

p/themes/base-theme/frss.css
Frenzie: > Possibly

I do think it would be good to have it somewhere. But of course that needn't be the footer.

Frenzie: Speaking of which all URLs should be made explicit really. It's really simple with generated content but then it'd have to be immediately behind all links. For something more advanced like footnotes you'd need JS or PHP.

Anyway, that's not for this PR (unless you want it to be, heh).

Inverle: > Speaking of which all URLs should be made explicit really. It's really simple with generated content but then it'd have to be immediately behind all links. For something more advanced like footnotes you'd need JS or PHP.

For the print of the whole page we already have this:
(doesn't apply to the JS print dialog because it doesn't include `.flux_content`)
https://github.com/FreshRSS/FreshRSS/blob/e6540335caa0b6d39c213a9a11ca77ad1eebe504/p/themes/base-theme/frss.css#L2827-L2830

But it looks awful, so it would have to be footnotes if anything.

> Anyway, that's not for this PR (unless you want it to be, heh).

This is yet again something that should be configurable, as footnotes wouldn't be needed in PDFs for example since you can just click/hover on links.
Note that footnotes could conflict with the actual contents of the article.

Frenzie: > But it looks awful, so it would have to be footnotes if anything.

I think it looks reasonable enough for the most part. I prefer parentheses over square brackets though. And, well, not too fond of the italic really. I just mean the principle.

> in PDFs for example since you can just click/hover on links.

Eh, that's debatable. It's one thing for navigational aids within the document but making it less useful when printed strikes me as somewhat besides the point.

Inverle: > I think it looks reasonable enough for the most part. I prefer parentheses over square brackets though

I'm fine about going with this approach, but it needs to be configurable.
(turned off by default?)

PR #7730, opened 2025-07-13
Fix HTML queries with a single feed

### LINKED ISSUE #7652: [Bug] Favicon doesn't display in user query with a single feed
### Describe the bug

If you have a user query that looks like this (displays a single feed):

Image

the favicon will not display correctly on the shared HTML page (sometimes the icon is a globe instead of this one):

Image

note that the url is incorrect too (example.net)

in reality, the feed does have a favicon:

Image

This does not happen when I'm sharing a category, favorites, labels or main stream.

### To Reproduce

N/A

### Expected behavior

 N/A

### FreshRSS version

1.26.4-dev

### System information

- Database version: [e.g. SQLite, PostgreSQL 17] SQLite 3.45.3
- PHP version: [e.g. PHP 8.2] PHP 8.3.14
- Installation type: [e.g. Docker, Docker image source, git, Yunohost] git
- Web server type: [e.g. Apache, nginx] nginx/1.28.0
- Device: [e.g. iPhone13]
- Browser: [e.g. Firefox 134]

### Additional context

_No response_

Inverle:
fix #7652

### DIFF
--- a/app/views/index/html.phtml
+++ b/app/views/index/html.phtml
@@ -20,7 +20,7 @@
 	<?php
 		foreach ($this->entries as $entry):
 			$this->entry = $entry;
-			$this->feed = $this->feeds[$entry->feedId()] ??
+			$this->feed = $this->feeds[0] ??
 				FreshRSS_Category::findFeed($this->categories, $entry->feedId()) ??
 				FreshRSS_Feed::default();
 	?>
--- a/app/views/index/html.phtml
+++ b/app/views/index/html.phtml
@@ -20,7 +20,7 @@
 	<?php
 		foreach ($this->entries as $entry):
 			$this->entry = $entry;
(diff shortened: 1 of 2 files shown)


Inverle:
`$this->feeds[0]` refers to https://github.com/FreshRSS/FreshRSS/blob/8385eadf49ddb5e6f46fc34155aec66005d9350b/p/api/query.php#L145

Alkarex:
I am not in front of a computer at the moment, but I have a suspicion that it is the referred line that should be fixed to: 

```php
$view->feeds = [$feed->id() => $feed];
```

Inverle:
It probably should be
```
app/Controllers/indexController.php:248: $this->view->feeds = [$feed->id() => $feed];
```
(used similarly in another place)

Inverle:
Although I believe it was still working without breaking anything else

PR #7752, opened 2025-07-24
Update chart.js from 3.5.1 to 4.5.0

Inverle:
https://www.chartjs.org/docs/4.5.0/migration/v4-migration.html

note:

> * `dist/chart.min.js` has been renamed to `dist/chart.umd.min.js` (and before 4.5.0 `dist/chart.umd.js`)

---

SHA256:
```
2f27bcf471b2d69dd78494f6e2172fb28470eb843820e2f96bb85d39f9618d30  chart.umd.min.js
```

can be verified with: `curl -s https://unpkg.com/chart.js@4.5.0/dist/chart.umd.min.js | sha256sum`


p/scripts/vendor/chart.umd.min.js
```diff
@@ -0,0 +1,14 @@
...
+//# sourceMappingURL=chart.umd.min.js.map
```
Inverle, line 14: We can leave this line even though we don't include a .map file

Alkarex: We could consider adding a `.map` if it appears desirable at some point

Inverle:
Updating chart.js broke SeaMonkey support for statistics

Alkarex:
> Updating chart.js broke SeaMonkey support for statistics

It is not critical to support SeaMonkey for non-essential features, but it would be good if we could get it back, especially since it is an interesting baseline: if it works in SeaMonkey, it should cover a great percentage of the use-cases for people who cannot update their browser fast.

Frenzie:
If it's just some syntax and not browser features that are the issue, letting Babel loose on it should solve it. Though I'm not entirely sure how or if we should do anything with NPM.

Inverle:
> If it's just some syntax and not browser features that are the issue,

It is a syntax error

Inverle:
https://github.com/chartjs/Chart.js/blob/6372280085625b43ef34d7b70f3e86b063d22a10/rollup.config.js#L17-L34

changing `es2022` to `es2021` fixes the issue

Frenzie:
Excellent. Rollup support for such things is fairly limited so I'd half assumed it wouldn't suffice, but of course it's not like Babel would be difficult either. Unfortunately it doesn't really answer the harder question: maintain a minimal fork? Stick a README-type file next to ChartJS? Or in short, how do we not forget to do it in the future if that's what we want.

Inverle:
Would it be fine to submit a PR with these instructions?

```bash
cd Chart.js && \
git checkout tags/v4.5.0 && \
pnpm install && \
pnpm run build && \
# SHA256 of dist/chart.umd.min.js should be: 311a5a5e0db077a787b782977f359a72f8584f94d800e324dcc6aefe8a006dd7 
```

Frenzie:
Fine by me in any case.

Inverle:
#7816

PR #7712, opened 2025-07-01
Extension installation and updates via GUI

tryallthethings:
Feel free to make changes. The only issue I've found is that hitting the GitHub rate limit is an option when testing 😄

# Closes 
#2900 
FreshRSS/Extensions/issues/323

# Changes proposed in this pull request:

- Added extension installation and update mechanism for individual extensions and bulk updates
- Added helper functions to detect repository type, find version matching https://raw.githubusercontent.com/FreshRSS/Extensions/master/extensions.json, download release, find release notes, sanitize downloaded extension files
- Added translations for update actions
- change in base CSS to display button icon and text without wrapping (Plus icon and install button text will wrap by default since the column is too small)

# How to test the feature manually:

1. Replace files with the changes of this PR
2. Go to Settings → Extensions 
3. Every known extension will have an "Install" or "Update" button in the newly added "Actions" column. Clicking it will install the extension or update it

# Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] unit tests written (optional if too hard)
- [X] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #7753, opened 2025-07-25
Implement sudo mode / reauthentication

Inverle:
* Useful for securing #7712
* Similar to GitHub's [sudo mode](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/sudo-mode)

Currently protected pages:
* Administration
   * Manage users
   * Authentication
   * Update (only `apply` action)
* Extensions
  * UserJS configuration
  
New `config.php` settings:
* `reauth_required`: whether reauthentication prompts should be shown
* `reauth_time`: time after reauthenticating you won't be asked to again in seconds (default: 1200s / 20 min)

In order to ask for reauthentication, developers can use the following code snippet:
```php
if (FreshRSS_Auth::requestReauth()) {
     return;
}
```
Optionally, a redirect parameter can be passed into the function. By default it will redirect to the page that requested the reauth.

<details>
<summary>Screenshot of reauth page</summary>

image

</details>


Inverle:
From GitHub's page:

> GitHub has a two-hour session timeout period before prompting you for authentication again. During this time, any sensitive action that you perform will reset the timer.

Would automatically resetting the timer be a good idea too?

Frenzie:
> Would automatically resetting the timer be a good idea too?

Seems pretty hostile any other way. Although I suppose if it's as long as 2 hours it may not matter much in practice.

app/Controllers/indexController.php
```diff
@@ -354,6 +354,10 @@ public function logsAction(): void {
 			Minz_Error::error(403);
 		}
 
+		if (FreshRSS_Auth::hasAccess('admin') && FreshRSS_Auth::requestReauth()) {
+			return;
```
Alkarex: Isn't it strange to allow other users to see their logs while requiring admins to re-authenticate?
That is maybe superfluous here.

Alkarex:
> > Would automatically resetting the timer be a good idea too?
> 
> Seems pretty hostile any other way. Although I suppose if it's as long as 2 hours it may not matter much in practice.

I also believe that in the case of FreshRSS, it will likely not make any significant difference in practice, so it is fine either way for me, favouring slightly the "keep it simple" approach :-)

app/Controllers/indexController.php
Inverle: I'm making an assumption that admins have access to some logs that normal users don't, if it's a wrong assumption then it probably makes sense to remove these lines.

app/Models/Auth.php
```diff
@@ -230,4 +237,60 @@ public static function isCsrfOk(?string $token = null): bool {
...
+				header('HTTP/1.1 403 Forbidden');
+				exit();
+			}
+
+			$sid = Minz_Session::paramString('sid');
+
+			$payload = Minz_Url::serialize([
+				'returnTo' => $redirect ?? Minz_Request::currentRequest(),
+			]);
+			$signature = hash_hmac('sha256', $payload, $sid);
```
Alkarex: Is the HMAC mechanism needed, or could the redirection information simply be saved in the session?

app/Controllers/indexController.php
Alkarex: I cannot immediately think of especially sensitive information visible in the those logs.
There are additional logs (e.g. API, WebSub) but not exposed through the Web UI.

app/Models/Auth.php
Inverle: Yes, it would mostly work the same way. If that's preferred, I'll rewrite it to be stored in the session instead.

The added benefit in doing it the current way is better handling of multiple tabs. If the user opens a reauth page, then another tab on a different reauth page and returns to the previous tab and enters their password, they'll be redirected to the wrong page.

It's not likely to happen often, though I personally thought of this as a better way because of this.
Technically it could be stored in the URL too without HMAC, but that's just in case.

edit: I was also thinking of another approach with `Minz_Request::forward()` without redirects. I'm not sure yet how I would implement it

Alkarex: > Technically it could be stored in the URL too without HMAC, but that's just in case.

That sounds simpler, maybe supplemented by a check of a CSRF token. I am slightly worried (for long-term maintenance) about adding some dependencies to more complicated crypto functions such as `hash_hmac()` if we do not really need them

Inverle: Done, no CSRF token check though

app/Controllers/indexController.php
Inverle: Not needed for now - removed

Inverle: (although extensions might log sensitive information?)

PR #7670, opened 2025-06-15
SimplePie: Fix propagation of HTTP error codes

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7038
https://github.com/FreshRSS/simplepie/pull/36
upstream https://github.com/simplepie/simplepie/pull/905

PR #7760, opened 2025-07-27
Implement support for HTTP 429 Too Many Requests

### LINKED ISSUE #6973: [Bug] Openrss blocking freshrss for too many simultaneous requests
### Describe the bug

according to [issue 163](https://openrss.org/issue/163) there is a problem because FreshRSS tries to pull all sources at the same time.
Please check [their guidelines](https://openrss.org/guides/developers-guide-to-open-rss-feeds) and see how the fetching could be improved.

### To Reproduce

Subscribe to multiple OpenRSS feeds

### Expected behavior

Requests towards the same source should be [spaced out](https://openrss.org/guides/developers-guide-to-open-rss-feeds#space-out-your-requests).

### FreshRSS version

1.24.3

### LINKED ISSUE #7571: Implement support for HTTP header Retry-After (429, 503)

https://github.com/FreshRSS/FreshRSS/issues/6712#issuecomment-2293772071
But it would be good to first find some examples of feeds actually implementing that

### Use case

_No response_

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After

_No response_

### Additional context

_No response_

Alkarex:
As well as *503 Service Unavailable*.
Will obey the corresponding HTTP `Retry-After` header at domain level.


lib/simplepie/simplepie/src/File.php
```diff
@@ -127,7 +127,7 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
                 curl_setopt($fp, CURLOPT_URL, $url);
                 curl_setopt($fp, CURLOPT_HEADER, 1);
                 curl_setopt($fp, CURLOPT_RETURNTRANSFER, 1);
-                curl_setopt($fp, CURLOPT_FAILONERROR, 1);
+                // curl_setopt($fp, CURLOPT_FAILONERROR, 1); // FreshRSS removed to retrieve headers even on HTTP errors
```
Alkarex, line 130: https://github.com/simplepie/simplepie/commit/dcabc40d90d4807f2d069576f56dbe030defd59d

lib/simplepie/simplepie/src/Misc.php
```diff
@@ -1737,7 +1737,7 @@ public static function strip_comments(string $data)
     }
 
     /**
-     * @return int|bool
+     * @return int|false
```
Alkarex, line 1740: https://github.com/simplepie/simplepie/pull/926

Alkarex:
Related:

* https://github.com/FreshRSS/FreshRSS/issues/7571
* https://github.com/FreshRSS/FreshRSS/issues/6973
* https://github.com/FreshRSS/FreshRSS/discussions/3862
* https://github.com/FreshRSS/FreshRSS/pull/7670
* https://github.com/simplepie/simplepie/pull/905#issuecomment-3123376251
* https://github.com/FreshRSS/simplepie/pull/36

@pe1uca Feedback welcome here

Frenzie:
Even if the value it sets is more than a month? I think there's a point somewhere past let's say a week where it ceases to be meaningful (though granted, perhaps the app intends to communicate never to return again? :-)

Alkarex:
> Even if the value it sets is more than a month? I think there's a point somewhere past let's say a week where it ceases to be meaningful (though granted, perhaps the app intends to communicate never to return again? :-)

Yes indeed. That was my intention, but I forgot :-)

Frenzie (approved):
I haven't tried it ('cause I'd have to set up a test server that sends out these headers or something) but it looks good. :+1:

PR #7762, opened 2025-07-29
Regenerate cookie ID after logging out

Inverle:
To make the session cookie no longer usable if hijacked and put in another browser after user logs out

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -226,6 +226,12 @@ class FreshRSS_auth_Controller extends FreshRSS_ActionController {
 		if (Minz_Request::isPost()) {
 			invalidateHttpCache();
 			FreshRSS_Auth::removeAccess();
+
+			ini_set('session.use_cookies', '1');
+			Minz_Session::lock();
+			Minz_Session::regenerateID();
+			Minz_Session::unlock();
+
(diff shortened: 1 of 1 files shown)


Inverle:
Might need further improvement as multiple cookies are being sent when logging out:

image

I haven't noticed this during the testing (only one cookie was being sent) but it doesn't seem to cause any other issues than this.

See https://www.php.net/manual/en/function.session-regenerate-id.php:

> Warning
>
>Currently, session_regenerate_id does not handle an unstable network well, e.g. Mobile and WiFi network. Therefore, you may experience a lost session by calling session_regenerate_id.

>You should not destroy old session data immediately, but should use destroy time-stamp and control access to old session ID. Otherwise, concurrent access to page may result in inconsistent state, or you may have lost session, or it may cause client (browser) side race condition and may create many session ID needlessly. Immediate session data deletion disables session hijack attack detection and prevention also.

> session_regenerate_id([bool](https://www.php.net/manual/en/language.types.boolean.php) $delete_old_session = [false](https://www.php.net/manual/en/reserved.constants.php#constant.false)): [bool](https://www.php.net/manual/en/language.types.boolean.php)

https://github.com/FreshRSS/FreshRSS/blob/e967b07589f687fcd2f71e2df265fcb7c4f15c07/lib/Minz/Session.php#L203-L205

PR #7763, opened 2025-07-29
Require current password when setting new password

Inverle:
* Prevents bypassing reauthentication in #7753 and other unauthorized password changes
* The user is no longer signed out after changing password, just other sessions

<details>
<summary>Before (screenshot)</summary>

image

</details>

<details>
<summary>After (screenshot)</summary>

image

</details>


app/Controllers/userController.php
```diff
@@ -114,9 +115,41 @@ public function profileAction(): void {
...
+						_t('feedback.profile.passwords_dont_match'),
+						['c' => 'user', 'a' => 'profile']
+					);
+					return;
+				}
+
+				ini_set('session.use_cookies', '1');
+				Minz_Session::lock();
+				Minz_Session::regenerateID();
+				Minz_Session::unlock();
```
Inverle, line 151: https://github.com/FreshRSS/FreshRSS/pull/7762#issuecomment-3134444730

Alkarex: Why is it needed to regenerate the session ID here?

Inverle: A good reason for a user to change their password is if someone had gotten into the account. In such a scenario, the session ID may have also been hijacked.

Inverle:
Even though this change was made, it's still possible to partially backdoor an account without providing the current password by setting API password or master authentication token.
Not sure how to solve this in the UI

first idea is to put profile page behind reauthentication and remove *Current password* field from *Change password* section

Inverle:
New idea:
Divide all authentication fields into separate forms/sections which would require entering the current password (on the profile page, where they are currently)

There is also the problem that it's harder for a user who had their account compromised to even know their API password/token changed, so even if they change the password, the attacker can still access the user's account.

Maybe there should be a date displayed of when the API password/token was last changed next to the fields?

Issue #5518, opened 2023-07-10
[Feature] Use favicon that is included in the feed


Chris3773:
**Is your feature request related to a problem? Please describe.**
I'm using rss bridge for a custom feed and it has the option to return an URL to icon for the feed. However it looks like FreshRSS ignores the field and pulls from the Website URL which in this case doesn't have a valid favicon.

**Describe the solution you’d like**
It would be nice if it would look for an `icon` field before trying to pull it from the website URL.

Alkarex:
Hello,
For reference, could you please paste here a snippet illustrating how this information looks like? If possible both for RSS-Bridge RSS and ATOM formats

Chris3773:
It looks like only the ATOM and JSON formats return back an `icon` field.

ATOM:
```
<?xml version="1.0" encoding="UTF-8"?>
<feed xmlns="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">
  <title type="text">....</title>
  <id>...</id>
  <icon>https://www.example.com/favicon.ico</icon>
  <logo>https://www.example.com/favicon.ico</logo>
  ...
</feed>
```
JSON:
```
{
    "version": "https:\/\/jsonfeed.org\/version\/1",
    "title": "....",
    "home_page_url": "https:\/\/www.example.com",
    "feed_url": "...",
    "icon": "https:\/\/www.example.com\/favicon.ico",
    "favicon": "https:\/\/www.example.com\/favicon.ico",
    "items": []
}
```

aidistan:
I'm also a heavy user of RSS Bridge and other RSS aggregators. This feature is truly needed.

I conducted an investigation and here's my proposed solution:

1. Introduce a new attribute, `imageUrl`, to the FreshRSS_Feed class, and make it persistent..
2. Store the value returned by `$simplePie->get_image_url()` alongside other attributes such as `website` and `descriptions`.
3. Update the `favicon()` function to first check for `imageUrl`. If available, provide it directly; otherwise, fallback to the original method (`Minz_Url::display('/f.php?' . $this->hash());`).

I'm not proficient in handling persistence in PHP. This is the best effort I could offer.

bowencool:
Please also support this format:

<img width="746" height="271" alt="Image" src="https://github.com/user-attachments/assets/33fa324e-9103-461b-ab49-3ca166472619" />

Alkarex:
Help welcome

Alkarex:
[...]

PR #7767, opened 2025-07-31
Rework fetch favicons

Alkarex:
* Use main function `httpGet()` instead of local one;
* Use HTTP cache, for multiple feeds from the same domain, also between users;
* Do not default to feed URL when there is no website URL

TODO for later:
* https://github.com/FreshRSS/FreshRSS/issues/7774

### DIFF
--- a/lib/favicons.php
+++ b/lib/favicons.php
@@ -22,81 +22,14 @@ function isImgMime(string $content): bool {
 	return $isImage;
 }
 
-/** @param array<int,int|bool|string> $curlOptions */
-function downloadHttp(string &$url, array $curlOptions = []): string {
-	if (($retryAfter = FreshRSS_http_Util::getRetryAfter($url)) > 0) {
-		Minz_Log::warning('For that domain, will first retry favicon after ' . date('c', $retryAfter) . '. ' . \SimplePie\Misc::url_remove_credentials($url));
-		return '';
-	}
-
-	syslog(LOG_INFO, 'FreshRSS Favicon GET ' . $url);
-	$url2 = checkUrl($url);
-	if ($url2 == false) {
-		return '';
-	}
-	$url = $url2;
-
-	$ch = curl_init($url);
-	if ($ch === false) {
-		return '';
-	}
-	curl_setopt_array($ch, [
-			CURLOPT_HEADER => true,
-			CURLOPT_RETURNTRANSFER => true,
-			CURLOPT_TIMEOUT => 15,
-			CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
-			CURLOPT_MAXREDIRS => 10,
-			CURLOPT_FOLLOWLOCATION => true,
-			CURLOPT_ENCODING => '',	//Enable all encodings
-			//CURLOPT_VERBOSE => 1,	// To debug sent HTTP headers
-		]);
-
-	FreshRSS_Context::initSystem();
-	if (FreshRSS_Context::hasSystemConf()) {
-		curl_setopt_array($ch, FreshRSS_Context::systemConf()->curl_options);
-	}
-
-	curl_setopt_array($ch, $curlOptions);
-
-	$response = curl_exec($ch);
-	$c_status = curl_getinfo($ch, CURLINFO_HTTP_CODE);
-	$c_effective_url = curl_getinfo($ch, CURLINFO_EFFECTIVE_URL);
-	curl_close($ch);
-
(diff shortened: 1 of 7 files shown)


Alkarex:
Related to:
* https://github.com/FreshRSS/FreshRSS/issues/6973
* https://github.com/FreshRSS/FreshRSS/pull/7646
* https://github.com/FreshRSS/FreshRSS/pull/7760

@Inverle Tests welcome

lib/favicons.php
```diff
@@ -22,81 +22,14 @@ function isImgMime(string $content): bool {
...
+	return CACHE_PATH . '/' . sha1($url) . '.ico';
 }
 
-function searchFavicon(string &$url): string {
+function searchFavicon(string $url): string {
 	$dom = new DOMDocument();
-	$html = downloadHttp($url);
-
-	if ($html == '' || !@$dom->loadHTML($html, LIBXML_NONET | LIBXML_NOERROR | LIBXML_NOWARNING)) {
+	['body' => $html, 'effective_url' => $effective_url, 'fail' => $fail] = httpGet($url, cachePath: CACHE_PATH . '/' . sha1($url) . '.html', type: 'html');
```
Inverle, line 31: I was recently seeing `data:` image URIs trying to be fetched in the logs, and it looks like `checkUrl()` is missing here which may be the reason for that.
Also note the other 2 `httpGet()` below

Unfortunately I cleared the logs before so I don't have a screenshot and I wasn't able to figure out which feed emitted the log.

I was only able to figure out it came from `httpGet()` by the beginning of the log message `Error fetching content: HTTP code `

Inverle: Would be nice to have
```php
syslog(LOG_INFO, 'FreshRSS Favicon GET ' . $url);
```
back again too

Alkarex: > Would be nice to have
> 
> ```
> syslog(LOG_INFO, 'FreshRSS Favicon GET ' . $url);
> ```
> 
> back again too

We have something like:

```
FreshRSS[33]: FreshRSS GET html https://github.com/FreshRSS/FreshRSS/
FreshRSS[33]: FreshRSS GET ico https://github.githubassets.com/favicons/favicon.svg
172.18.0.1 - - [13/Aug/2025:14:17:39 +0200] "GET /f.php?h=00f87409 HTTP/1.1" 200 506 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:141.0) Gecko/20100101 Firefox/141.0"
```

Isn't it good enough?

Inverle: Okay I missed that, sorry

Inverle: I was checking my logs and looks like it happened again a few days ago

image

~~Still have no idea which feed is causing this~~

Inverle: I'm subscribed to multiple feeds that do this but here's one of them: https://poorlydrawnlines.com/feed/

it has `<link rel="icon" href="data:,">` in the HTML of https://poorlydrawnlines.com/
may require clicking `Clear cache` to trigger the behavior

Inverle: The feed does have an icon though and even if it didn't `data:,` wouldn't be any better.

So I wonder:
* should we support icons with `data:` URIs? (and when to prefer them over HTTP)
* if not, should we add a `checkUrl()` before downloading the icon so that an error message doesn't appear or leave it?

Alkarex: It would be nice to support Atom's `<icon>` for favicons:
* https://github.com/FreshRSS/FreshRSS/issues/7774
* https://github.com/FreshRSS/FreshRSS/issues/5518

And supporting `data:` (to be decoded to a normal image) would be a nice addition.

And in any case, yes for `checkUrl()`.

Inverle: > And supporting `data:` (to be decoded to a normal image) would be a nice addition.

Yes, but when should it be preferred over a HTTP icon?
Should it only be used when there is no HTTP or default `favicon.ico` icon?

Or: simply preferred over `favicon.ico` and used if it's valid by checking with `isImgMime()`

Multiple `<link rel="icon">` can be specified too, so not sure how to go about this

Alkarex: If an icon is defined at feed level, it should have priority. (Watch out that feeds may also have some larger logos, which are not appropriate for a favicon).
If multiple icons are defined at HTML level, I do not have a clear priority strategy, but should ideally be based on size information.

PR #7770, opened 2025-07-31
Strip more unsafe attributes e.g. `referrerpolicy`

Inverle:
Especially `referrerpolicy`, feeds can use: `` to collect instance URLs (which are meant to be private)

We should start using a whitelist for attributes/tags at some point


Alkarex:
> We should start using a whitelist for attributes/tags at some point

Indeed. PR welcome (requires adding a new SimplePie method, which we can merge in our fork https://github.com/FreshRSS/simplepie )

Inverle:
> > We should start using a whitelist for attributes/tags at some point
> 
> Indeed. PR welcome (requires adding a new SimplePie method, which we can merge in our fork https://github.com/FreshRSS/simplepie )

There has been some discussion on this: https://github.com/simplepie/simplepie/issues/872#issuecomment-2265906334

Alkarex:
I would suggest a simple method on the same model than `strip_attributes()` but with a white-list, doing nothing if empty.

Inverle:
Yes, that's what I've been thinking too.

PR #7771, opened 2025-07-31
Improve security of form for user details

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/pull/7684
The form buttons requiring confirmation are disabled in HTML in the case of Ajax, and only enabled again if our own JavaScript is running

### DIFF
--- a/app/views/user/details.phtml
+++ b/app/views/user/details.phtml
@@ -69,14 +69,17 @@
 
 		<div class="form-group form-actions">
 			<noscript><b><?= _t('gen.js.should_be_activated'); ?></b></noscript>
+			<?php
+				$disabledIfAjax = Minz_Request::paramBoolean('ajax') ? ' disabled="disabled"' : '';
+			?>
 			<div class="group-controls">
 				<button type="submit" class="btn btn-important" name="action" value="update"><?= _t('gen.action.update') ?></button>
-				<button type="submit" class="btn btn-attention confirm" name="action" value="purge"><?= _t('gen.action.purge') ?></button>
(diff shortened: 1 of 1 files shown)


Inverle:
One problem with this approach is that there are many more forms that need to be protected.
It would be a lot of work to put `<?= $disabledIfAjax ?>` everywhere, and you would also have to remember to put it when adding a new button.

For example:
https://github.com/FreshRSS/FreshRSS/blob/edge/app/views/helpers/extension/configure.phtml
https://github.com/FreshRSS/FreshRSS/blob/edge/app/views/helpers/feed/update.phtml
https://github.com/FreshRSS/FreshRSS/blob/edge/app/views/tag/update.phtml
https://github.com/FreshRSS/FreshRSS/blob/edge/app/views/configure/query.phtml

Alkarex:
Reference to comment:
* https://github.com/FreshRSS/FreshRSS/pull/7684#issuecomment-3148365068

Alkarex:
I cannot work on that for a few days, so let's merge this simple fix already and get back to it. Help welcome in the meantime

Inverle:
This doesn't reenable the buttons with JS though, so they are disabled even in the slider now 🤔

image

Alkarex:
Arg, I thought I had tested it. We already have some JS that enables all `<button class="confirm" disabled="disabled">` but it is apparently not fired in the slider...

Alkarex:
Quick fix https://github.com/FreshRSS/FreshRSS/pull/7786

PR #7785, opened 2025-08-03
New JS attribute: `data-auto-leave-validation`

Inverle:
Instead of a repeating pattern like: `<input type="text" value="something" data-leave-validation="something">`, you can now put a `data-auto-leave-validation="1"` attribute on a `<form>` for example, and it will automatically set the `data-leave-validation` attributes inside the form elements.

`data_auto_leave_validation(parent)`  from `extra.js` is called on slider open and page load.

todo:
* [x] test if I broke anything - probably not
* [x] see if this works correctly with other `<input>` types - seems like it
* [x] if there are any places where leave validation is missing, it should be fixed in this PR - I think I didn't miss any?


Alkarex:
Good idea, much cleaner 👍🏻

app/views/auth/index.phtml
```diff
@@ -11,7 +11,7 @@
 		<div class="form-group">
 			<label class="group-name" for="auth_type"><?= _t('admin.auth.type') ?></label>
 			<div class="group-controls">
-				<select id="auth_type" name="auth_type" required="required" data-leave-validation="<?= FreshRSS_Context::systemConf()->auth_type ?>">
+				<select id="auth_type" name="auth_type" required="required">
```
Alkarex, line 14: I have not checked out the code, but isn't the new `data-auto-leave-validation` missing at `form` level here?

Inverle: Good catch. I haven't tested the changes fully yet.

Inverle: fixed

Alkarex:
I find the text message a bit confusing.
At the moment, it is showing `gen.js.confirm_action` => *Are you sure you want to perform this action? It cannot be cancelled!*
I think we need to make a new one (if a more appropriate one does not already exist) for the case when the user has modified some things and forgotten to save.

Inverle:
> I find the text message a bit confusing.
At the moment, it is showing gen.js.confirm_action => Are you sure you want to perform this action? It cannot be cancelled!

That's not related to this PR, it just affects `onbeforeunload` not confirm buttons.

> I think we need to make a new one (if a more appropriate one does not already exist) for the case when the user has modified some things and forgotten to save.

The message that shows up when you `return false` on `onbeforeunload` cannot be changed though:

image

Frenzie:
> The message that shows up when you return false on onbeforeunload cannot be changed though:

It used to be able to be changed, but I think browsers stopped allowing it.

Alkarex:
> > The message that shows up when you return false on onbeforeunload cannot be changed though:
> 
> It used to be able to be changed, but I think browsers stopped allowing it.

Ah, that might be why I got confused. It seems to differ depending on whether it is called from JS or from a navigation event:

Testing the new PR **with** the slider, it shows:

image

Upon navigation, it shows:

image

Inverle:
Oh I see what you meant now (slider exit with unsaved settings is the same message as in confirm button)

Alkarex:
@Inverle What is the status? Will you try to make a new i18n string with a better message?

Inverle:
See [77b02d8](https://github.com/FreshRSS/FreshRSS/pull/7785/commits/77b02d8d2fa6960145073c837b744d9a004091f9)

image

app/i18n/fr/gen.php
```diff
@@ -145,6 +145,7 @@
 		'category_empty' => 'Catégorie vide',
 		'confirm_action' => 'Êtes-vous sûr(e) de vouloir continuer ? Cette action ne peut être annulée !',
 		'confirm_action_feed_cat' => 'Êtes-vous sûr(e) de vouloir continuer ? Vous perdrez les favoris et les filtres associés. Cette action ne peut être annulée !',
+		'confirm_exit_slider' => 'Are you sure you want to discard unsaved settings?',	// TODO
```
```suggestion
		'confirm_exit_slider' => 'Êtes-vous sûr de vouloir abandonner les paramètres non enregistrés ?',
```

app/i18n/nl/gen.php
```diff
@@ -145,6 +145,7 @@
 		'category_empty' => 'Lege categorie',
 		'confirm_action' => 'Weet u zeker dat u dit wilt doen? Het kan niet ongedaan worden gemaakt!',
 		'confirm_action_feed_cat' => 'Weet u zeker dat u dit wilt doen? U verliest alle gereleteerde favorieten en gebruikers informatie. Het kan niet ongedaan worden gemaakt!',
+		'confirm_exit_slider' => 'Are you sure you want to discard unsaved settings?',	// TODO
```
Frenzie: Weet
```suggestion
		'confirm_exit_slider' => 'Weet u zeker dat u de niet opgeslagen instellingen wilt negeren?',
```

Issue #7570, opened 2025-05-09
[Bug] Wrong iconURL returned in GReader API in v1.26.2


ghost:
### Describe the bug

On a new Docker container running FreshRSS v1.26.2, the wrong iconURL is returned from the GReader API. It appears that an iconUrl is generated, but it only returns the default favicon (globe). I've verified this by running FreshRSS v1.26.1 in a Docker container and the issue doesn't occur there.

### To Reproduce

1. Start a new Docker container (or any fresh install, I'm guessing)
2. Setup accounts and API access
3. Run ```curl -X POST -d "Email=apiuser&Passwd=apipasswd" 'HOST/api/greader.php/accounts/ClientLogin'``` to get API auth
4. Run ```curl -s -H "Authorization:GoogleLogin auth=apiuser/AUTHKEY" 'HOST/api/greader.php/reader/api/0/subscription/list?output=json'``` and find any RSS entry.
5. Go to the iconURL item that is returned and it should return the default favicon

### Expected behavior

The correct iconUrl should be returned from the GReader API.

### FreshRSS version

1.26.2

### System information

- Database version: SQLite
- PHP version: 8.2.28
- Installation type: Official Docker container
- Web server type: Apache/2.4.62
- Device: curl
- OS: Debian 12
- Browser: N/A (This does affect applications that use the GReader API)

Alkarex:
Make sure that you first have updated your feeds. See https://github.com/FreshRSS/FreshRSS/discussions/7551

ghost:
> Make sure that you first have updated your feeds. See [#7551](https://github.com/FreshRSS/FreshRSS/discussions/7551)

The feed icons on v1.26.2 work in the browser, but the correct iconUrl still isn't returned from the GReader API after multiple feed refreshes.

PR #7792, opened 2025-08-05
GReader API: fix incorrect favicon URL

hilariousperson:
Since 7915abd833e1ab7a72ad27b3ec52020ac9ab7051, `f.php?<iconHash>` is no longer correct and should be replaced with `f.php?q=<iconHash>`. This was done in the commit everywhere except in the GReader API as far as I can tell. This pull request fixes this inconsistency.

How to test this fix (thanks to CarelessCaution in #7570 for this useful test case for a different bug):

Obtain GReader API credentials:

    curl -X POST -d "Email=apiuser&Passwd=apipasswd" 'HOST/api/greader.php/accounts/ClientLogin'

Check feed icon URLs:

    curl -s -H "Authorization:GoogleLogin auth=AUTH_TOKEN" 'HOST/api/greader.php/reader/api/0/subscription/list?output=json'

Let me know if anything is wrong, this is a first time contribution. Thanks for the software :)

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -438,7 +438,7 @@ class FreshRSS_Feed extends Minz_Model {
 		@unlink($path . '.ico');
 		@unlink($path . '.txt');
 	}
-	public function favicon(): string {
+	public function favicon(bool $absolute = false): string {
 		$hash = $this->hashFavicon();
 		$url = '/f.php?h=' . $hash;
 		if ($this->customFavicon()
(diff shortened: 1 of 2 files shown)


Alkarex:
Thanks for the but report and the PR. And welcome :-)
Regression from https://github.com/FreshRSS/FreshRSS/pull/7646
The fix was not good enough, but I have just made a new suggestion, untested. Could you please take a look?

Alkarex:
Ping @Inverle

Inverle:
This is probably also the cause of https://github.com/jocmp/capyreader/issues/1329

Inverle:
I can confirm the fix is working (tested with readrops)

PR #7793, opened 2025-08-06
Optimize how much data needs to be `chown`/`chmod`ed on container startup

### LINKED ISSUE #7764: [Bug] Long delay when starting docker image due to recursive `chown`
### Describe the bug

When running the current `freshrss/freshrss:latest` container, startup takes an extremely long time due to the following line: https://github.com/FreshRSS/FreshRSS/blob/697f8f58119cf544be3b46d2209993fbd3d67979/cli/access-permissions.sh#L15

Logs (via `docker logs --follow --timestamps <container id>`):
```
2025-07-31T03:13:27.908365000Z Enabling module auth_openidc.
2025-07-31T03:25:33.676488000Z [Thu Jul 31 03:25:33.676143 2025] [mpm_prefork:notice] [pid 1:tid 1] AH00163: Apache/2.4.62 (Debian) configured -- resuming normal operations
```

Notice the roughly 12 minute gap in the logs between when the entry point runs and when Apache is started. Basically all of that time is spent running the above `chown` command. This is on a fresh instance with only a single user and a single feed.

### To Reproduce

I can reproduce it by running the following docker compose file:

```yaml
services:
  freshrss:
    image: freshrss/freshrss:latest
    container_name: freshrss
    restart: unless-stopped
    volumes:
      - /data/freshrss/data:/var/www/FreshRSS/data
      - /data/freshrss/extensions:/var/www/FreshRSS/extensions
    ports:
      - 8080:80
    environment:
      CRON_MIN: 2,32
      OIDC_ENABLED: 1
      OIDC_PROVIDER_METADATA_URL: https://<REDACTED>/.well-known/openid-configuration
      OIDC_CLIENT_ID: <REDACTED>
      OIDC_CLIENT_SECRET: <REDACTED>
      OIDC_CLIENT_CRYPTO_KEY: <REDACTED>
      OIDC_REMOTE_USER_CLAIM: preferred_username
      OIDC_SCOPES: openid email profile
      OIDC_X_FORWARDED_HEADERS: X-Forwarded-Host X-Forwarded-Port X-Forwarded-Proto
```

I also tested the `alpine`, `edge`, and `edge-alpine` tags with the same result (unsurprising since they all run the same startup script).

### Expected behavior

I expect the container to start without a long delay.

### FreshRSS version

1.26.3

### System information

- Database version: SQLite
- PHP version: \<managed by docker image\>
- Installation type: Docker (`latest`, `alpine`, `edge`, and `edge-alpine`)
- Web server type: \<managed by docker image\>
- Device: N/A
- Host OS: Debian 12.10 container running under Proxmox (kernel `6.8.12-1-pve`) with nesting enabled so it can run containers within it.
- Host filesystem: ZFS backed by a pool of hard drives
- Browser: N/A

Relevant lines from `docker info`:
```
Server:
 Server Version: 28.0.1
 Storage Driver: overlay2
  Backing Filesystem: zfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: true
 Kernel Version: 6.8.12-1-pve
 Operating System: Debian GNU/Linux 12 (bookworm)
 OSType: linux
 Architecture: x86_64
```

### Additional context

`chown` being slow is a known issue with the `overlay2` storage driver when paired with some filesystems/kernels/other. See:
 - https://github.com/docker/for-linux/issues/388
 - https://github.com/linuxserver/docker-bookstack/issues/60

I don't think this issue can be avoided entirely, but there are some potential ways to improve it. The most common recommendation seems to be to use the `--chown` flag when `COPY`-ing files into the image wherever possible so only a few directories have the be `chown`'d at runtime. Here's a random example of doing this that I found linked from the above issue: https://github.com/wizarrrr/wizarr/commit/f68faee52d42de4b27ae385014fad564c74ca719

pR0Ps:
This works around an issue where `chmod`/`chown` operations inside a container can be extremely slow when using the `overlay2` storage driver, resulting in 10min+ container startup times.

It modifies the owner of the webapp when building the container so that only the `data` and `extensions` directories (which are commonly mapped as volumes into the container) have to be modified by the `access-permissions.sh` script at container startup.

When not running via docker the behaviour of the `access-permissions.sh` script is unchanged.

This reduced the container startup time for me from ~15 minutes to a few seconds.

Closes #7764

- Uses the `COPY` command in the dockerfiles to set the group of the copied data as `www-data`
- Changes the docker entrypoint script to call `access-permissions.sh --only-userdirs` 
- Changes `access-permissions.sh` to only `chown`/`chmod` the `data` and `extensions` directories instead of the entire webapp if the `--only-userdirs` flag is provided.

1. Run the container and ensure it starts up quickly and properly.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update && \
 RUN mkdir -p /var/www/FreshRSS/ /run/apache2/
 WORKDIR /var/www/FreshRSS
 
-COPY --chown=:www-data . /var/www/FreshRSS
+COPY --chown=root:www-data . /var/www/FreshRSS
 COPY ./Docker/*.Apache.conf /etc/apache2/sites-available/
 
 ARG FRESHRSS_VERSION
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update && \
 RUN mkdir -p /var/www/FreshRSS/ /run/apache2/
 WORKDIR /var/www/FreshRSS
 
-COPY . /var/www/FreshRSS
+COPY --chown=:www-data . /var/www/FreshRSS
 COPY ./Docker/*.Apache.conf /etc/apache2/sites-available/
 
 ARG FRESHRSS_VERSION
--- a/cli/access-permissions.sh
+++ b/cli/access-permissions.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Apply access permissions
 
 if [ ! -f './constants.php' ] || [ ! -d './cli/' ]; then
@@ -11,19 +11,22 @@ if [ "$(id -u)" -ne 0 ]; then
 	exit 3
 fi
 
-# If specified, only modify the data and extension dirs
+# Always fix permissions on the data and extensions directories
+# Optionally also fix permissions on the entire webapp
+data_path="${DATA_PATH:-./data}"
+to_update=("${data_path}")
 if [ "${1:-}" = "--only-userdirs" ]; then
-	to_update="./data ./extensions"
+	to_update+=("./extensions")
 else
-	to_update="."
+	to_update+=(".")
 fi
 
 # Based on group access
-chown -R :www-data $to_update
+chown -R :www-data "${to_update[@]}"
 
 # Read files, and directory traversal
-chmod -R g+rX $to_update
+chmod -R g+rX "${to_update[@]}"
 
-# Write access
-mkdir -p ./data/users/_/
-chmod -R g+w ./data/
+# Write access to data
+mkdir -p "${data_path}/users/_/"
+chmod -R g+w "${data_path}"
--- a/cli/access-permissions.sh
+++ b/cli/access-permissions.sh
@@ -20,6 +20,8 @@ else
 	to_update="."
 fi
 
+mkdir -p "${data_path}/users/_/"
+
 # Based on group access
 chown -R :www-data "$data_path" "$to_update"
 
@@ -27,5 +29,4 @@ chown -R :www-data "$data_path" "$to_update"
 chmod -R g+rX "$data_path" "$to_update"
 
 # Write access to data
-mkdir -p "${data_path}/users/_/"
 chmod -R g+w "$data_path"
--- a/cli/access-permissions.sh
+++ b/cli/access-permissions.sh
@@ -11,11 +11,18 @@ if [ "$(id -u)" -ne 0 ]; then
 	exit 3
 fi
 
+# If specified, only modify the data and extension dirs
+if [ "${1:-}" = "--only-userdirs" ]; then
+	to_update="./data ./extensions"
+else
+	to_update="."
+fi
+
 # Based on group access
-chown -R :www-data .
+chown -R :www-data $to_update
 
 # Read files, and directory traversal
-chmod -R g+rX .
+chmod -R g+rX $to_update
 
 # Write access
 mkdir -p ./data/users/_/
--- a/cli/access-permissions.sh
+++ b/cli/access-permissions.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 # Apply access permissions
 
 if [ ! -f './constants.php' ] || [ ! -d './cli/' ]; then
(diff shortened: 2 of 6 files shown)


cli/access-permissions.sh
```diff
@@ -11,11 +11,18 @@ if [ "$(id -u)" -ne 0 ]; then
 	exit 3
 fi
 
+# If specified, only modify the data and extension dirs
+if [ "${1:-}" = "--only-userdirs" ]; then
+	to_update="./data ./extensions"
```
Alkarex: It would be nice to include the `DATA_PATH` environment variable as target, if defined:
https://github.com/FreshRSS/FreshRSS/blob/9faf2c1fa3c8c3ed82ff4c95733a018d0f9977a7/constants.php#L63-L69

Edit: relevant for the cases both with `--only-userdirs` and without

Alkarex:
From CI:

```
Running shellcheck on cli/access-permissions.sh

In cli/access-permissions.sh line 22:
chown -R :www-data $to_update
                   ^--------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
chown -R :www-data "$to_update"


In cli/access-permissions.sh line 25:
chmod -R g+rX $to_update
              ^--------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
chmod -R g+rX "$to_update"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

cli/access-permissions.sh
pR0Ps: Added processing of `DATA_PATH` environment variable both with and without `--only-userdirs`.

Docker/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN apt-get update && \
 RUN mkdir -p /var/www/FreshRSS/ /run/apache2/
 WORKDIR /var/www/FreshRSS
 
-COPY . /var/www/FreshRSS
+COPY --chown=:www-data . /var/www/FreshRSS
```
Alkarex: For some reason, this syntax fails on my Docker `Docker version 27.5.1, build 27.5.1-0ubuntu3` with the following error message:
> unable to convert uid/gid chown string to host mapping: can't find uid for user : no such user:

Alkarex:
I have made a [few more improvements](https://github.com/FreshRSS/FreshRSS/pull/7793/commits/ac75ce17d53ef8fc01e173b6cb4bb3016f57b945). Please double-check

Docker/Dockerfile
pR0Ps: Strange, I checked the docs and there's no mention of the user-less syntax that I used but it seemed to work. Your version is definitely more correct though.

pR0Ps:
Your changes look good. I just pushed a very minor change to the `REAPPLY_PERMISSIONS` variable - feel free to discard it if you don't agree.

The `edge` branch is broken for me right now (I think it's OIDC-related, ~~haven't tracked it down fully yet~~ **EDIT:** See #7798) but when I apply all the commits in the PR on top of the current `latest` branch, it works perfectly.

pR0Ps:
Actually, I found an issue with the latest commit while trying it with a fresh install. 

This line: https://github.com/FreshRSS/FreshRSS/blob/1f8273803f632b40c260721cb3257ae4e2251f5a/Docker/entrypoint.sh#L50
will copy some data in the `./data` directory as root and will therefore always need another call to `./cli/access-permissions.sh --only-userdirs` to fix the permissions.

pR0Ps:
I've pushed a fix for the above issue (removing the conditional reapplication of permissions).

In order to test this latest iteration of the PR fully, I ran FreshRSS in a very basic configuration (no OIDC) with both a fresh install (empty data directory) and an install with existing data:
```yaml
services:
  freshrss:
    build:
      context: https://github.com/pR0Ps/FreshRSS.git#edge
      dockerfile: Docker/Dockerfile-Newest
    container_name: freshrss
    restart: unless-stopped
    volumes:
      - /data/freshrss/data:/var/www/FreshRSS/data
      - /data/freshrss/extensions:/var/www/FreshRSS/extensions
    ports:
      - 8080:80
    environment:
      CRON_MIN: 8,38
      OIDC_ENABLED: 0
```

I also tested all combinations of this PR based on the `edge`/`latest` branch , as well as specifying `Dockerfile-Newest`/`Dockerfile` - all of them worked.

Alkarex:
Thanks again! Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #7804, opened 2025-08-09
Show warning when unsafe CSP policy is in use

Inverle:
Also contains some forgotten changes from #7677

Example of why this is useful: https://github.com/YunoHost/yunohost/blob/9d98459b2208e7084a958acc22562b7139276cf5/conf/nginx/security.conf.inc#L31
I found that YunoHost overwrites the CSP header and makes using XSS payloads like `<video onloadstart="alert(1)"><source>` possible.

image

<details>
<summary>Show screenshot of warning</summary>

image

</details>


Alkarex:
Thanks, looks good 👍 
Have you reported the issue to Yunohost?

Inverle:
> Have you reported the issue to Yunohost?

Looks like it's a known issue: https://github.com/YunoHost/issues/issues/2301
(it affects all apps)

Inverle:
The only fix that is possible is adding this line to https://github.com/YunoHost-Apps/freshrss_ynh/blob/master/conf/nginx.conf

```
more_set_headers "Content-Security-Policy: default-src 'self'; frame-src *; img-src * data: blob:; frame-ancestors 'self'; media-src *; upgrade-insecure-requests"
```

Note it is slightly less secure to set this CSP globally on all pages instead of how it's done in FreshRSS, and some extensions may not work correctly.

Inverle:
I made a PR: https://github.com/YunoHost-Apps/freshrss_ynh/pull/209

Alkarex:
Ping @ericgaspar, @lapineige

app/i18n/en/gen.php
```diff
@@ -155,6 +155,7 @@
 		'labels_empty' => 'No labels',
 		'new_article' => 'There are new articles available, click to refresh the page.',
 		'should_be_activated' => 'JavaScript must be enabled',
+		'unsafe_csp_header' => 'The CSP header in use is unsafe and FreshRSS may be vulnerable to XSS attacks. <a href="https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security">See documentation</a>',
```
Alkarex:
```suggestion
		'unsafe_csp_header' => 'The CSP header in use is unsafe and FreshRSS may be vulnerable to XSS attacks. <a href="https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security" target="_blank">See documentation</a>',
```

Maybe?

ericgaspar:
!testme

app/i18n/en/gen.php
Inverle: done

lapineige:
Thank you for the warning. So I guess we should adjust the package CSP policy ?

PR #7810, opened 2025-08-10
Put CSP everywhere

Inverle:
* Puts CSP everywhere in `p/api`
   * including the HTML query page ❗
   * Also in `p/ext.php`
* Puts `X-Content-Type-Options: nosniff` everywhere
* Fixes custom icon configuration not showing `blob:` icon in statsController (idle feeds)
   * Also removes `style-src 'unsafe-inline'` since it doesn't seem to be needed
* Improves CSP of `p/f.php`

### DIFF
--- a/app/Controllers/statsController.php
+++ b/app/Controllers/statsController.php
@@ -30,8 +30,7 @@ class FreshRSS_stats_Controller extends FreshRSS_ActionController {
 		$this->_csp([
 			'default-src' => "'self'",
 			'frame-ancestors' => "'none'",
-			'img-src' => '* data:',
-			'style-src' => "'self' 'unsafe-inline'",
+			'img-src' => '* data: blob:',
 		]);
 
 		$catDAO = FreshRSS_Factory::createCategoryDao();
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -1,6 +1,6 @@
 <?php
 declare(strict_types=1);
-header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'");
+header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'; sandbox");
 header('X-Content-Type-Options: nosniff');
 
 /**
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -1,5 +1,7 @@
 <?php
 declare(strict_types=1);
+header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'");
+header('X-Content-Type-Options: nosniff');
 
 /**
  * Fever API for FreshRSS
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -28,7 +28,7 @@ Server-side API compatible with Google Reader API layer 2
 require(__DIR__ . '/../../constants.php');
 require(LIB_PATH . '/lib_rss.php');	//Includes class autoloader
 
-header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'");
+header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'; sandbox");
 header('X-Content-Type-Options: nosniff');
 
 if (PHP_INT_SIZE < 8) {	//32-bit
--- a/p/api/greader.php
+++ b/p/api/greader.php
@@ -28,6 +28,9 @@ Server-side API compatible with Google Reader API layer 2
 require(__DIR__ . '/../../constants.php');
 require(LIB_PATH . '/lib_rss.php');	//Includes class autoloader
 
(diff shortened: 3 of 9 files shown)


p/api/greader.php
```diff
@@ -28,6 +28,9 @@
 require(__DIR__ . '/../../constants.php');
 require(LIB_PATH . '/lib_rss.php');	//Includes class autoloader
 
+header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'");
```
Alkarex: There is no HTML here. Is it needed?

p/api/pshb.php
```diff
@@ -6,6 +6,7 @@
 const MAX_PAYLOAD = 3_145_728;
 
 header('Content-Type: text/plain; charset=UTF-8');
+header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'");
```
Alkarex: No HTML here either

p/api/fever.php
```diff
@@ -1,5 +1,7 @@
 <?php
 declare(strict_types=1);
+header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'");
```
Alkarex: There is no HTML here. Is it needed?

p/api/greader.php
Inverle: GReader can return `application/xml` (although with content disposition)

try pasting this in the browser address bar:
`data:application/xml,<script src="data:text/javascript,alert(1)" xmlns="http://www.w3.org/1999/xhtml"></script>`

(XML can run JS)

edit: and just because it has `Content-Disposition` doesn't mean it's safe: https://www.mozilla.org/en-US/security/advisories/mfsa2025-51/#CVE-2025-6430 (see recent CVE)

p/api/pshb.php
Inverle: https://github.com/FreshRSS/FreshRSS/pull/7810#discussion_r2265484422
just to be safe (maybe not needed with `text/plain` though)

p/api/fever.php
Inverle: https://github.com/FreshRSS/FreshRSS/pull/7810#discussion_r2265484422
just to be safe

app/Controllers/statsController.php
```diff
@@ -30,8 +30,7 @@ public function firstAction(): void {
 		$this->_csp([
 			'default-src' => "'self'",
 			'frame-ancestors' => "'none'",
-			'img-src' => '* data:',
-			'style-src' => "'self' 'unsafe-inline'",
```
Alkarex, line 34: `unsafe-inline` not needed anymore for stats?

Inverle: I don't see any CSP errors in the console when opening stats page. Haven't tested with an older browser, but I think `unsafe-inline` is not needed since the inline styles are inserted using JS (possibly with something like `el.style.color = 'red'` rather than setting the `style` attribute)

Inverle: image

Inverle: This is really old CSP: https://github.com/FreshRSS/FreshRSS/commit/cb913a3a76daf357ad36ca39c26b4aaf800211d2
FreshRSS used flotr2 instead of chart.js back then

Answer: https://github.com/FreshRSS/FreshRSS/issues/1075#issuecomment-186879067

Alkarex: > image

Do you have a suggestion for that issue?

Inverle: No issue here, chart.js applies the styles in a way that doesn't require `unsafe-inline`
edit: see https://github.com/chartjs/Chart.js/blob/6372280085625b43ef34d7b70f3e86b063d22a10/src/platform/platform.dom.js#L49-L85

Inverle:
Should https://github.com/FreshRSS/FreshRSS/blob/edge/p/api/misc.php also have CSP?

Alkarex:
> Should https://github.com/FreshRSS/FreshRSS/blob/edge/p/api/misc.php also have CSP?

It may be used for any form of API, potentially also some Web-based, so yes, good idea

Inverle:
So if we added
```php
header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'; sandbox");
```
at the top of misc.php, an extension can just overwrite it with another `header()` call?

Inverle:
https://github.com/ravenscroftj/freshrss-flaresolverr-extension/pull/31

Alkarex:
> an extension can just overwrite it with another `header()` call?

Hum, to be tested, and may depend on output buffering, but in general I am not sure that works

PR #7812, opened 2025-08-10
Fix autocomplete issues in change password form

Inverle:
## Screenshots

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

This is an example on Firefox, where the `Master authentication token` field was incorrectly being autofilled.
Red borders are indicating that the fields are required.

## List of changes

* `required="required"` is now being added to the password fields if the section is open
* The `challenge` field is being added if section is open instead of when at least one of the password fields isn't empty due to autocomplete
* Added `autocomplete="new-password"` on fields that shouldn't be autocompleted
   * Unfortunately Chrome requires a workaround with CSS
   * Not tested on Safari yet
* User will be redirected to profile page after successfully changing their password instead of index page

## How to test

Autocomplete related changes should be tested on a HTTPS page with saved credentials for FreshRSS


Alkarex:
Can this wait till 1.28?

Inverle:
It's just a minor annoyance I think, but it could be confusing if someone goes to the profile page and has a password saved

PR #7816, opened 2025-08-11
Fix chart.js support for SeaMonkey

Inverle:
SeaMonkey support was broken by #7752 

Reproduce JS file:
```bash
cd Chart.js && \
git checkout tags/v4.5.0 && \
pnpm install && \
pnpm run build && \
# SHA256 of dist/chart.umd.min.js should be: 311a5a5e0db077a787b782977f359a72f8584f94d800e324dcc6aefe8a006dd7
```


Alkarex:
That is fine, but isn't it simpler to take an official build of an older version, such as 3.9.1 or whichever is the latest branch targetting es2021?

Inverle:
> That is fine, but isn't it simpler to take an official build of an older version, such as 3.9.1 

Isn't it better to have the latest version instead?
Indeed I tested that v3.9.1 is the last version to work on SeaMonkey, but it's still a bit old.

Note it's not specifically es2021, it just happened to be the latest one to work. Probably should be es2017 since SeaMonkey is based on Firefox 60 which released in 2018.

Alkarex:
> Isn't it better to have the latest version instead?

For the limited use we have, only if we need some kind of new feature or fix

PR #7775, opened 2025-08-01
Bump SimplePie with PHPStan Level 8

Alkarex:
* https://github.com/FreshRSS/simplepie/pull/45
* https://github.com/FreshRSS/simplepie/pull/46
* https://github.com/FreshRSS/simplepie/pull/47

SimplePie increased to PHPStan Level 8:
* https://github.com/simplepie/simplepie/pull/857

Including my two PRs:
* https://github.com/simplepie/simplepie/pull/932
* https://github.com/simplepie/simplepie/pull/933

Resolve upstream sync of Expose HTTP status:
* https://github.com/simplepie/simplepie/pull/905#issuecomment-3007605779
* https://github.com/simplepie/simplepie/pull/909
* https://github.com/FreshRSS/FreshRSS/issues/7038
* https://github.com/FreshRSS/FreshRSS/pull/7670

PR #7824, opened 2025-08-14
Fix regression on_http_response

Alkarex:
Fix regression (bad merge) from https://github.com/FreshRSS/FreshRSS/pull/7775
* https://github.com/FreshRSS/simplepie/pull/50

### DIFF
--- a/app/Models/SimplePieResponse.php
+++ b/app/Models/SimplePieResponse.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 final class FreshRSS_SimplePieResponse extends \SimplePie\File
 {
 	#[\Override]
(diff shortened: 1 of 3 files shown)


app/Models/SimplePieResponse.php
```diff
@@ -4,7 +4,7 @@
 final class FreshRSS_SimplePieResponse extends \SimplePie\File
 {
 	#[\Override]
-	protected function on_http_response(string|false $response = ''): void {
+	protected function on_http_response($response = ''): void {
```
Alkarex, line 7: https://github.com/FreshRSS/FreshRSS/pull/7825

PR #7829, opened 2025-08-15
Regenerate session ID on login

Inverle:
Follow-up to #7762

### DIFF
--- a/lib/Minz/Session.php
+++ b/lib/Minz/Session.php
@@ -200,8 +200,11 @@ class Minz_Session {
 	 * Regenerate a session id.
 	 * Useful to call session_set_cookie_params after session_start()
 	 */
-	public static function regenerateID(): void {
-		session_regenerate_id(true);
+	public static function regenerateID(bool $delete_old_session = true): void {
+		ini_set('session.use_cookies', '1');
+		Minz_Session::lock();
+		session_regenerate_id($delete_old_session);
+		Minz_Session::unlock();
 	}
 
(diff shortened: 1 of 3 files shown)


Inverle:
works properly now

lib/Minz/Session.php
```diff
@@ -198,10 +198,18 @@ public static function keepCookie(int $l): void {
...
+		session_start();
 		session_regenerate_id(true);
+		session_write_close();
+		$newId = session_id();
+		if ($newId === false) {
+			Minz_Error::error(500);
+			return;
+		}
+		$lifetime = session_get_cookie_params()['lifetime'];
+		setcookie('FreshRSS', $newId, $lifetime, self::getCookieDir(), '', Minz_Request::isHttps(), true);
```
Alkarex: Please pass `'FreshRSS'` as parameter `$name` to keep the same convention as the related functions just below

Inverle: Made some other changes too: https://github.com/FreshRSS/FreshRSS/pull/7829/commits/a8dcc22f9ad711fffec8280fac4fffd89f1114aa

Alkarex:
Can it wait till after the release of 1.27.0?

Inverle:
Probably

Inverle:
Forgot to also apply this when creating user or deleting account

Not critical but will make a PR later
(because `giveAccess()` checks if `passwordHash` is the same in the session and for bcrypt it's always different so no risk of session fixation here - assumption from earlier testing)

PR #7830, opened 2025-08-16
Improve leave validation

Inverle:
fix: no confirmation popup was shown when exiting slider with shortcut (ESC)
fix: no confirmation popup was shown after exiting slider and having reset the icon to default or set by extension
fix: input for `close_dropdown` wasn't named correctly for label to work
refactor: `close_dropdown` was renamed to `close_menus`

TODO for later PR: move custom favicon code to `feed.js` instead of `extra.js`

### DIFF
--- a/app/Models/Context.php
+++ b/app/Models/Context.php
@@ -151,6 +151,15 @@ final class FreshRSS_Context {
 		if (!in_array(FreshRSS_Context::$user_conf->display_categories, [ 'active', 'remember', 'all', 'none' ], true)) {
 			FreshRSS_Context::$user_conf->display_categories = FreshRSS_Context::$user_conf->display_categories === true ? 'all' : 'active';
 		}
+
+		// FreshRSS 1.27.0
+		if (array_key_exists('close_dropdown', FreshRSS_Context::$user_conf->shortcuts)) {
+			$shortcuts = FreshRSS_Context::$user_conf->shortcuts;
+			$shortcuts['close_menus'] = $shortcuts['close_dropdown'];
+			unset($shortcuts['close_dropdown']);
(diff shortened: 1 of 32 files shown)


Alkarex:
Can it wait till after the release of 1.27.0?

Inverle:
Probably

app/Models/Context.php
```diff
@@ -151,6 +151,15 @@ public static function initUser(string $username = '', bool $userMustExist = tru
 		if (!in_array(FreshRSS_Context::$user_conf->display_categories, [ 'active', 'remember', 'all', 'none' ], true)) {
 			FreshRSS_Context::$user_conf->display_categories = FreshRSS_Context::$user_conf->display_categories === true ? 'all' : 'active';
 		}
+
+		// FreshRSS 1.27.0
+		if (array_key_exists('close_dropdown', FreshRSS_Context::$user_conf->shortcuts)) {
```
Alkarex:
```suggestion
		// FreshRSS 1.27.1+
		if (isset(FreshRSS_Context::$user_conf->shortcuts['close_dropdown'])) {
```

Inverle: Would it be interesting to move this code to https://github.com/FreshRSS/FreshRSS/tree/edge/app/migrations instead?

Alkarex: I am not so found of the migration approach, because people might go back and forth in versions, restore a backup after a migration, etc.
But `config-user.default.php` is supposed to be applied first, therefore ensuring the keys exists. It would be good to double-check the behaviour in the case of nested array though

Inverle: This is for moving the previous value to the new key though

PR #7715, opened 2025-07-04
Show translation status in README.md

Inverle:
Increases size of:
* [`README.md`](https://github.com/FreshRSS/FreshRSS/blob/edge/README.md) `17.5 KiB` -> `21.3 KiB`
* [`README.fr.md`](https://github.com/FreshRSS/FreshRSS/blob/edge/README.fr.md) `23 KiB` -> `26.9 KiB`

This should improve the visibility of which languages need help with translation and hopefully invite more contributors.

See the new [`README.md`](https://github.com/FreshRSS/FreshRSS/blob/db725d9c41f51872cf518536a23788b93c324e52/README.md) with translation status or the new [`README.fr.md`](https://github.com/FreshRSS/FreshRSS/blob/db725d9c41f51872cf518536a23788b93c324e52/README.fr.md)

PR #7842, opened 2025-08-19
Replace translation status flags with text section

Inverle:
See: https://github.com/FreshRSS/FreshRSS/pull/7715#issuecomment-3201783547

Using SVGs with unicode/image flags inside of them was a bad idea because:
* They don't render as intended in Chromium browsers on Windows
* A country flag might not correctly represent a language

This is more in the style of [Weblate](https://github.com/FreshRSS/FreshRSS/pull/7715#issuecomment-3202079748) but in a table instead. (showing language names instead of flags)

---

[View the new translation progress table](https://github.com/Inverle/FreshRSS/tree/translation-status-v2#internationalization-20-languages-translation-progress)


Alkarex:
This now takes a huge amount of space, so I suggest putting it inside a `<details> + <summary>` section.
But open for discussion

README.md
```diff
@@ -47,6 +40,42 @@ To facilitate contributions, the [following option](.devcontainer/README.md) is
 
 [![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://github.com/codespaces/new?hide_repo_select=true&ref=edge&repo=6322699)
 
+## Translation progress
```
math-GH:
```suggestion
## Internationalization: 20+ languages. Translation progress
```

README.md
```diff
@@ -47,6 +40,42 @@ To facilitate contributions, the [following option](.devcontainer/README.md) is
...
+
+| Language | Translated (%) | |
+| - | - | - |
+| Čeština | ⚠️ 89% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fcs+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Deutsch | ✅ 95% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fde+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Ελληνικά | ❌ 23% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fel+%2F%28TODO%7CDIRTY%29%24%2F) |
+| English | ✅ 100% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen+%2F%28TODO%7CDIRTY%29%24%2F) |
+| English (United States) | ✅ 100% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen-us+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Español | ✅ 93% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fes+%2F%28TODO%7CDIRTY%29%24%2F) |
+| فارسی | ✅ 99% | [contribute](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffa+%2F%28TODO%7CDIRTY%29%24%2F) |
```
math-GH: could you fix this ltr and rtl issue? Or is it on purpose? It looks difficult to maintain this line

math-GH:
> This now takes a huge amount of space, so I suggest putting it inside a `<details> + <summary>` section. But open for discussion

It would be fine for me. 

Another ideas:
* First 10 rows + `<details>` for the rest rows.
* having 2 parallel tables / 2 "colums" instead of 1 "colum"

README.md
Inverle: See the rendered [markdown table](https://github.com/Inverle/FreshRSS/blob/dce1776f7bd376943d9c4b257ccc9a5d07cdd4cd/#translation-progress), it doesn't look like this
Also the line is autogenerated

aledeg:
While I understand the idea behind this modification, there are a few thing I think we should address here.
1. It's nice you used the language name in its own language. But I think it is a bit hard for people who do not recognize them. For instance, I do not know what this language is (ex: `فارسی`). To mitigate that, a simple solution would be to include the ISO code somewhere in the table.
2. I found hard to have a visual cue of the completion of a language. The intent is there but I am not a big fan of the use of the Unicode indicators. I think I prefer progress bars like displayed in the Weblate documentation.
3. Speaking of progress, I think the column showing the progress should be named this way instead of `translated`. I think it's more accurate.

Apart from that, I like the way you removed the flags.

My 2¢

Alkarex:
> 2\. I found hard to have a visual cue of the completion of a language. The intent is there but I am not a big fan of the use of the Unicode indicators. I think I prefer progress bars

Sticking to Unicode to avoid images, what about something like: ⣀ ⣄ ⣤ ⣦ ⣶ ⣷ ⣿ ?
There are other possible variants such as ■□□□□□□□□□

Frenzie:
That's not a bad idea. :-)

aledeg:
> > 2. I found hard to have a visual cue of the completion of a language. The intent is there but I am not a big fan of the use of the Unicode indicators. I think I prefer progress bars
> 
> Sticking to Unicode to avoid images, what about something like: ⣀ ⣄ ⣤ ⣦ ⣶ ⣷ ⣿ ? There are other possible variants such as ■□□□□□□□□□

We can use HTML in markdown files. So why not using directly [the Progress Indicator element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress)?

Alkarex:
> We can use HTML in markdown files. So why not using directly [the Progress Indicator element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress)?

Fine for me

Inverle:
GitHub sanitizes `<progress>` out of the HTML
So we're going to have to use ■□□□□□□□□□ instead

Inverle:
See: https://github.com/Inverle/FreshRSS/tree/translation-status-v2#internationalization-20-languages-translation-progress

math-GH:
I am fine with it.

I would suggest to have the screenshot on the readme.md first before that i18n table.

aledeg:
> See: [Inverle/FreshRSS@`translation-status-v2`#internationalization-20-languages-translation-progress](https://github.com/Inverle/FreshRSS/tree/translation-status-v2#internationalization-20-languages-translation-progress)

I like that! Good job!

Just a bit of nitpicking, I think ❰ and ❱ are a visual distraction. See my edit below

Screen Shot 2025-09-30 at 15 40 07

Inverle:
> I would suggest to have the screenshot on the readme.md first before that i18n table.
> Just a bit of nitpicking, I think ❰ and ❱ are a visual distraction. See my edit below

Done

aledeg:
Nice!

Some nitpicking again! I think your header is not the proper one. Currently, it looks like it is part of the extension section.

aledeg (approved):

Alkarex:
To be standard compliant (RFC 5646 / IETF BCP 47 / ISO 3166), the region subtag is supposed to be uppercased. [Fixed](https://github.com/FreshRSS/FreshRSS/pull/7842/commits/f530929886f136af583b3db1e53fd12127cebce8).
https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag

I am actually sad that our folders are not using the proper standard (e.g. [`pt-pt`](https://github.com/FreshRSS/FreshRSS/tree/128c375fc9de5119b1963d92dc9cd3e423053111/app/i18n/pt-pt) with the wrong casing for the country code) and would like to fix that (PR welcome).

app/i18n/fr/gen.php
```diff
@@ -217,6 +216,11 @@
 		'weeks' => 'semaines',
 		'years' => 'années',
 	),
+	'readme' => array(
```
Inverle, line 219: @Alkarex

It would be nice to translate these strings for the French readme.

Alkarex: Done :-)
Maybe the link *contribute* could use the targetted language

README.fr.md
```diff
@@ -226,6 +219,42 @@ Le processus d’import/export à l’aide d’une base de données SQLite est u
 FreshRSS permet l’ajout d’extensions en plus des fonctionnalités natives.
 Voir le [dépôt dédié à ces extensions](https://github.com/FreshRSS/Extensions).
 
+# Internationalization: 20+ languages. Translation progress
```
Inverle: Also here @Alkarex 
(anchor link above needs to be replaced too)

Alkarex: Done

README.fr.md
```diff
@@ -1,20 +1,13 @@
...
-
 * Lire ce document sur [github.com/FreshRSS/FreshRSS/](https://github.com/FreshRSS/FreshRSS/blob/edge/README.md) pour avoir les images et liens corrects.
 * [English version](README.md)
 
 # FreshRSS
 
 FreshRSS est un agrégateur de flux RSS à auto-héberger.
 
-Il se veut léger et facile à prendre en main tout en étant un outil puissant et paramétrable.
+Il se veut léger et facile à prendre en main tout en étant un outil puissant et paramétrable. <!-- todo: put ./#translation-progress link but with french header id here -->
```
Inverle: > (anchor link above needs to be replaced too)

for example:

and has been `[translated](./#internationalisation--20-langues-statut-des-traductions)` into many languages

Alkarex: Ok, done, and links fixed (the `./` should not be used there as it moves away from the current file)

Inverle: Last time I was testing it didn't work without `./` 🤔 

(didn't test correctly: https://github.com/Inverle/FreshRSS/commits/test/)

PR #7843, opened 2025-08-20
Fix fetching OPML url with special characters

Inverle:
Add dynamic OPML with url `http://freshrss2.local/api/query.php?user=X&t=X&f=opml`

Expect to see in server logs:
```
172.18.0.2 - - [20/Aug/2025:02:25:38 +0000] "GET /api/query.php?user=X&t=X&f=opml HTTP/1.1" 200 1591 "-" "FreshRSS/1.27.1-dev (Linux; https://freshrss.org)"
```

Instead this happens:
```
172.18.0.2 - - [20/Aug/2025:02:20:41 +0000] "GET /api/query.php?user=X&amp;t=X&amp;f=opml HTTP/1.1" 422 18 "-" "FreshRSS/1.27.1-dev (Linux; https://freshrss.org)"
```

image

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -182,7 +182,7 @@ class FreshRSS_Category extends Minz_Model {
 	}
 
 	public function refreshDynamicOpml(): bool {
-		$url = htmlspecialchars_decode($this->attributeString('opml_url') ?? '', ENT_QUOTES);
+		$url = $this->attributeString('opml_url');
 		if ($url == null) {
 			return false;
 		}
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -182,7 +182,7 @@ class FreshRSS_Category extends Minz_Model {
 	}
 
 	public function refreshDynamicOpml(): bool {
-		$url = htmlspecialchars_decode($this->attributeString('opml_url'));
+		$url = htmlspecialchars_decode($this->attributeString('opml_url') ?? '', ENT_QUOTES);
 		if ($url == null) {
 			return false;
 		}
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -182,7 +182,7 @@ class FreshRSS_Category extends Minz_Model {
 	}
 
 	public function refreshDynamicOpml(): bool {
-		$url = $this->attributeString('opml_url');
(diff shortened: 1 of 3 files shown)


app/Models/Category.php
```diff
@@ -182,7 +182,7 @@ public function cacheFilename(string $url): string {
 	}
 
 	public function refreshDynamicOpml(): bool {
-		$url = $this->attributeString('opml_url');
+		$url = htmlspecialchars_decode($this->attributeString('opml_url') ?? '', ENT_QUOTES);
```
Alkarex: The attributes are not supposed to be XML encoded, so there is likely a problem in the edit/save function

Inverle: Is there a different way to do it other than assigning a plaintext parameter instead?

https://github.com/FreshRSS/FreshRSS/blob/339dcb1ea324e19aacfd7be02f75133a70a43d7c/app/Controllers/categoryController.php#L62-L66

URLs are also decoded in other places:

https://github.com/FreshRSS/FreshRSS/blob/339dcb1ea324e19aacfd7be02f75133a70a43d7c/app/Models/Entry.php#L856
https://github.com/FreshRSS/FreshRSS/blob/339dcb1ea324e19aacfd7be02f75133a70a43d7c/app/Models/Feed.php#L566
https://github.com/FreshRSS/FreshRSS/blob/339dcb1ea324e19aacfd7be02f75133a70a43d7c/app/views/helpers/feed/update.phtml#L98-L99
https://github.com/FreshRSS/FreshRSS/blob/339dcb1ea324e19aacfd7be02f75133a70a43d7c/app/views/helpers/export/opml.phtml#L18-L19
etc.

Alkarex: The convention (so far) is to have HTML-encoded content in the native columns, and to have non-encoded content in the (JSON) attributes

Inverle: So I should use `$opml_url = checkUrl(Minz_Request::paramString('opml_url', plaintext: true)); ` and adjust to that in other places?

Alkarex: Yes, something like that 👍🏻

Inverle: Done

Alkarex (approved):

PR #7857, opened 2025-08-25
Add option for CSP frame-ancestors

Alkarex:
https://github.com/FreshRSS/FreshRSS/discussions/7856
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors


Alkarex:
A few additional places may be required, ~~such as in `p/f.php`, `p/api/index.php`~~, maybe `app/install.php` (but unsure what to do there)

Edit: `p/f.php`, `p/api/index.php` fixed

PR #7858, opened 2025-08-26
Fix Docker Traefik .yml and SERVER_DNS

ejohnmarlo:
Closes #

- Rename tls.yml to tls.yaml so it will be align to the entry in docker-compose-proxy.yml:

   volumes:
     - ./tls.yaml:/etc/traefik/tls.yaml

   To prevent error:
   error="error reading configuration file: /etc/traefik/tls.yaml - read /etc/traefik/tls.yaml: is a directory"

- Added parameter SERVER_DNS in the example.env

   To prevent warning:
   WARN[0000] The "SERVER_DNS" variable is not set. Defaulting to a blank string.

1. Copy example.env to .env 
2. Run: docker compose -f docker-compose.yml -f docker-compose-proxy.yml -f docker-compose-db.yml up
3. There should be no warning and error for the tls.yaml and SERVER_DNS

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Alkarex:
Thanks for the heads up 👍🏻 
I have taken the other approach to consistently using `.yml`, as it looks like `.yml` is more common for Traefik
(see e.g. https://doc.traefik.io/traefik/getting-started/configuration-overview/ )
Could you please check whether that still works for you?

ejohnmarlo:
> Thanks for the heads up 👍🏻 I have taken the other approach to consistently using `.yml`, as it looks like `.yml` is more common for Traefik (see e.g. https://doc.traefik.io/traefik/getting-started/configuration-overview/ ) Could you please check whether that still works for you?

I tested the approach of consistently using .yml, and it works without any issues on my end.

PR #7861, opened 2025-08-26
Docker CMD compatibility

### LINKED ISSUE #7859: [Bug] Latest docker image results in start loop, extensions folder permissions issue
### Describe the bug

Updating a running docker container to the latest image (sha256:6033114bec2d1722d2e3e809cedd8221188aef84c82f82b12b2cfd7e7b9bbb8a) results in the container being stuck in a start loop when extensions were installed. The containers log suggests it's a permission error with many entries like these:

`chmod: changing permissions of './extensions/xextension-threepanesview/metadata.json': Operation not permitted`

Delteting the fresshrss_extensions volume fails because of the same permission error.
Removing the container and then trying to delete the volume also fails.

I'm now stuck with a zombie volume that can't be removed no matter what I try.

The only solution was to recreate the container with a different (new) extension volume.

### To Reproduce

1. have a running docker container of freshrss and extensions installed
2. recreate the container with the latest image (sha256:6033114bec2d1722d2e3e809cedd8221188aef84c82f82b12b2cfd7e7b9bbb8a)
3. Deploy

### Expected behavior

_No response_

### FreshRSS version

1.27.0 ?

### System information

Docker on Alpine

### Additional context

_No response_

Alkarex:
Some caller systems do not seem to obey the [SHELL instruction](https://docs.docker.com/reference/dockerfile/#shell) (or may be wrongly escaping quotes)
fix https://github.com/FreshRSS/FreshRSS/issues/7859#issuecomment-3225691432
fix https://github.com/FreshRSS/FreshRSS/discussions/5611
fix https://github.com/FreshRSS/FreshRSS/discussions/7267

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -64,6 +64,6 @@ ENTRYPOINT ["./Docker/entrypoint.sh"]
 
 EXPOSE 80
 # hadolint ignore=DL3025
-CMD ([ -z "$CRON_MIN" ] || cron) && \
+CMD [ -z "$CRON_MIN" ] || cron && \
 	. /etc/apache2/envvars && \
 	exec apache2 -D FOREGROUND $([ "${OIDC_ENABLED:-0}" -ne 0 ] && echo "-D OIDC_ENABLED")
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -63,6 +63,7 @@ ENV TRUSTED_PROXY=''
 ENTRYPOINT ["./Docker/entrypoint.sh"]
 
 EXPOSE 80
-
 # hadolint ignore=DL3025
-CMD sh -c '([ -z "$CRON_MIN" ] || cron) && . /etc/apache2/envvars && if [ "${OIDC_ENABLED:-0}" -ne 0 ]; then exec apache2 -D FOREGROUND -D OIDC_ENABLED; else exec apache2 -D FOREGROUND; fi'
+CMD ([ -z "$CRON_MIN" ] || cron) && \
+	. /etc/apache2/envvars && \
+	exec apache2 -D FOREGROUND $([ "${OIDC_ENABLED:-0}" -ne 0 ] && echo "-D OIDC_ENABLED")
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -66,4 +66,4 @@ EXPOSE 80
 # hadolint ignore=DL3025
 CMD ([ -z "$CRON_MIN" ] || cron) && \
 	. /etc/apache2/envvars && \
-	exec apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ]&& echo "-D OIDC_ENABLED")
+	exec apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ] && echo "-D OIDC_ENABLED")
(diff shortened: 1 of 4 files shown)


Docker/Dockerfile
```diff
@@ -63,7 +63,6 @@ ENV TRUSTED_PROXY=''
 ENTRYPOINT ["./Docker/entrypoint.sh"]
 
 EXPOSE 80
+
 # hadolint ignore=DL3025
-CMD ([ -z "$CRON_MIN" ] || cron) && \
-	. /etc/apache2/envvars && \
-	exec apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ] && echo '-D OIDC_ENABLED')
+CMD sh -c '[ -z "$CRON_MIN" ] || cron; . /etc/apache2/envvars; if [ "${OIDC_ENABLED:-0}" -ne 0 ]; then exec apache2 -D FOREGROUND -D OIDC_ENABLED; else exec apache2 -D FOREGROUND; fi'
```
Frenzie: Nothing in that command is Bash syntax so I'd rather not rewrite the command without the && to fail early for example.

Frenzie: Or put another way, you could just change the SHELL command to `/bin/sh`.

Alkarex: Edits welcome

Frenzie: I wouldn't be inclined to edit anything at all.

You might phrase that as:
```suggestion
CMD sh -c '([ -z "$CRON_MIN" ] || cron) && . /etc/apache2/envvars && exec apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ] && echo "-D OIDC_ENABLED")'
```

But in doing so I just realized something. Naturally I had to change `echo '-D OIDC_ENABLED'` to make that work, which might indicate they're not ignoring the `SHELL` instruction but are astoundingly bad at escaping, because it'd need to become either `\'` or `"` to work as part of `-c` (or vice versa if you use `-c ""`).

Which suggests that maybe you just need to change the quotes:
```suggestion
CMD ([ -z "$CRON_MIN" ] || cron) && \
	. /etc/apache2/envvars && \
	exec apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ] && echo "-D OIDC_ENABLED")
```

Because if you don't do that, the error you'll see is of course:

```
Syntax error: end of file unexpected (expecting ")")
```

Which I believe is the exact error in the referenced issue.

Frenzie: In fact on closer inspection, the specific form of the error from https://github.com/FreshRSS/FreshRSS/issues/7859#issuecomment-3225691432 
```
OIDC_ENABLED): -c: line 2: unexpected EOF while looking for matching `)'
```
is how Bash presents that syntax error! The one I quoted above was from Dash.

Alkarex: What about now, with the `&&`s?

Alkarex: Ok, could be good to test with one of the failing systems

Alkarex: What about now?
https://github.com/FreshRSS/FreshRSS/pull/7861/commits/c8f3ec86b6181756a74240130fe8ae4640a2ed21

Frenzie: lgtm in any case

Alkarex:
Ok, now it only boils down to changing the quotes, which should be safe enough to merge

Frenzie (approved):

PR #7866, opened 2025-08-28
Fix curl response parsing

Inverle:
Regression from #7760 

After curl followed any redirects there would incorrectly be HTTP headers from the next response in the parsed response body.

Test feeds:
* https://research.securitum.com/feed/
   * FreshRSS tried to fetch `/favicon.ico` to download the icon, got redirected and the default WordPress icon didn't download correctly due to `isImgMime()` failing when seeing HTTP headers at the beginning
* https://turnoff.us/feed.xml
   * Set `html` as the CSS selector and click preview, see headers:
       
image

* https://blog.thea.codes/feed.xml
     * Same as above

### DIFF
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -539,20 +539,6 @@ function enforceHtmlBase(string $html, string $href): string {
 }
 
 /**
- * @param string $response Result from curl_exec() with CURLOPT_HEADER option enabled
- * @return ?\SimplePie\HTTP\Parser<false>
- */
-function parse_curl_response(string $response): ?\SimplePie\HTTP\Parser {
-	$parser = new \SimplePie\HTTP\Parser($response);
-	if (!$parser->parse()) return null;
-	while ($is_redirect = ($parser->status_code >= 300 && $parser->status_code <= 399)) {
-		$parser = new \SimplePie\HTTP\Parser($parser->body);
-		if (!$parser->parse()) return null;
-	}
-	return $parser;
-}
-
-/**
  * @param string $type {html,ico,json,opml,xml}
  * @param array<string,mixed> $attributes
  * @param array<int,mixed> $curl_options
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
:thinking:

Inverle:
@Frenzie can you check again?

used approach from https://github.com/FreshRSS/FreshRSS/blob/288992d9ad64c52b77983e27f497aa6f4ed5728e/lib/simplepie/simplepie/src/File.php#L164 instead

Frenzie:
Oh yes, with a max number of redirects is obviously better. I'd overlooked that. :-)

lib/lib_rss.php
```diff
@@ -643,13 +643,16 @@ function httpGet(string $url, string $cachePath, string $type = 'html', array $a
...
-	$parser = new \SimplePie\HTTP\Parser(is_string($response) ? $response : '');
-	if ($parser->parse()) {
-		$headers = $parser->headers;
-		$body = $parser->body;
-	} else {
-		$headers = [];
-		$body = false;
+	$body = false;
+	$headers = [];
+	if ($c_error == '') {
```
Alkarex: We need to parse the headers also in case of errors, to cover the cases of e.g. 429, 503

Alkarex: It was probably just the line `prepareHeaders()` that was missing, wasn't it? Without changing the rest

Inverle: Does curl return a status code if there's a curl error? 🤔
Not HTTP error, since [`CUROPT_FAILONERROR`](https://www.php.net/manual/en/curl.constants.php#constant.curlopt-failonerror) is not set

The error check may not be needed but `assert()` is. [1]
Although I don't understand why PHPStan doesn't warn against the fact that `curl_exec()` can return `false`

https://github.com/FreshRSS/FreshRSS/blob/288992d9ad64c52b77983e27f497aa6f4ed5728e/lib/simplepie/simplepie/src/File.php#L161-L166

[1]: specifically because of `non-negative-int` type
https://github.com/FreshRSS/FreshRSS/blob/288992d9ad64c52b77983e27f497aa6f4ed5728e/lib/simplepie/simplepie/src/HTTP/Parser.php#L503

Inverle: By the way, there would be no need for parsing `CURLOPT_HEADER` output at all if we could use [`CURLINFO_RETRY_AFTER`](https://www.php.net/manual/en/curl.constants.php#constant.curlinfo-retry-after) instead, since that's the only header we need but that's only available in PHP 8.2
FreshRSS needs to support PHP 8.1

Alkarex: > Does curl return a status code if there's a curl error? 🤔

Maybe not, indeed

lib/lib_rss.php
```diff
@@ -642,14 +642,18 @@ function httpGet(string $url, string $cachePath, string $type = 'html', array $a
...
-		$headers = $parser->headers;
-		$body = $parser->body;
-	} else {
-		$headers = [];
-		$body = false;
+	$c_errno = curl_errno($ch);
+
+	$body = false;
+	$headers = [];
+	if ($response !== false && $c_errno === 0) {
```
Alkarex: Is the condition `$c_errno === 0` adding anything?
Keeping only the first `$response !== false` seems to make more sense as it is only in the case when there is no response that we should no try to process it

Inverle: > Is the condition `$c_errno === 0` adding anything?

No it isn't

PR #7882, opened 2025-08-30
feed update slider: <label> -> <div>

math-GH:
Ref #7646

`<label>` needs to have an HTML input within or a `for=""` attribute

- fixed 2 `<label>` who are not a 'real' label, so a `<div>` is the better choice

1. open feed configuration
2. click on "Icon" (first lines of config page) or "Export as OPML" (last lines of config page)
3. nothing will happen. Mouse cursor will not change while hovering over the labels.


Inverle:
Indeed

PR #7883, opened 2025-08-30
Fix log CRLF injection

Inverle:
Can inject new lines into `Minz_Request::paramString('challenge')` f.e. by logging in with challenge `x%0D%0A[Sat, 30 Aug 2025 21:39:50 +0000] [warning] --- something`
https://github.com/FreshRSS/FreshRSS/blob/200eafb352f807bd70592b2ccc06745017328a85/app/Controllers/authController.php#L181
image

If it were `Minz_Request::paramString('challenge', plaintext: true)` or somewhere else, HTML could be injected into `level`
https://github.com/FreshRSS/FreshRSS/blob/200eafb352f807bd70592b2ccc06745017328a85/app/views/index/logs.phtml#L27

which is why its values are restricted now

### DIFF
--- a/lib/Minz/Log.php
+++ b/lib/Minz/Log.php
@@ -56,7 +56,7 @@ class Minz_Log {
 					$level_label = 'info';
 			}
 
-			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . str_replace(["\r", "\n"], '', $information) . "\n";
+			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . str_replace(["\r", "\n"], ' ', $information) . "\n";
 
 			if (defined('COPY_LOG_TO_SYSLOG') && COPY_LOG_TO_SYSLOG) {
 				syslog($level, '[' . $username . '] ' . trim($log));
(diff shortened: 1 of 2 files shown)


lib/Minz/Log.php
```diff
@@ -56,7 +56,7 @@ public static function record(string $information, int $level, ?string $file_nam
 					$level_label = 'info';
 			}
 
-			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . $information . "\n";
+			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . str_replace(["\r", "\n"], '', $information) . "\n";
```
Alkarex:
```suggestion
			$log = '[' . date('r') . '] [' . $level_label . '] --- ' . str_replace(["\r", "\n"], ' ', $information) . "\n";
```

Maybe?

Inverle: Sure

Inverle:
Before merging it would be useful to check if there are any other characters that may be used as line delimiter other than `\r` or `\n` (in `^`)

https://github.com/FreshRSS/FreshRSS/blob/87879e83927d66332d2d0794bc47eeb55f8c9b46/app/Models/LogDAO.php#L18

edit: looks like not

PR #7893, opened 2025-09-01
Fix/config dropdown menu mobile view

math-GH:
Before:
grafik

1: menu is not right hand side aligned
2: rounded corner (Origine theme specific)

After:
grafik

Menu is aligned on the right hand side
Sharp corner

- CSS

1. small screen for mobile view
2. open config menu right top corner

Negative check:
other dropdown menus should not be touched (f.e. User Queries menu, search, article labels menu)

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2643,9 +2643,9 @@ html.slider-active {
 		width: 90%;
 		height: 100vh;
 		top: 0;
-		right: 0;
+		right: 0 !important;
 		bottom: 0;
-		left: auto;
+		left: auto !important;
 		position: fixed;
 		padding-top: 0;
 		margin-top: 0;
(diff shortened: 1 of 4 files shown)


p/themes/base-theme/frss.css
```diff
@@ -2643,9 +2643,9 @@ html.slider-active {
 		width: 90%;
 		height: 100vh;
 		top: 0;
-		right: 0;
+		right: 0 !important;
```
Alkarex, line 2646: I would be happier if we could avoid all the `!important` but that can wait

math-GH: yes, me too

Inverle:
Wouldn't it be better to align the menu in the middle or full screen?

math-GH:
> Wouldn't it be better to align the menu in the middle or full screen?

The idea behind is:
The config menu button opens the menu right side aligned.
The "folder" icon button opens the main navigation left side aligned

Frenzie:
It made sense to me that way.

Inverle:
> The idea behind is:
> The config menu button opens the menu right side aligned.

I just got too used to it being on the left side that I got kinda confused, but I like it now

PR #7907, opened 2025-09-03
Fixes for no-cache.txt

### LINKED ISSUE #7885: [Bug] no-cache.txt causes undefined function errors
### Describe the bug

Create `data/no-cache.txt`

Observe following response in requests like `/i/?c=extension&a=serve&x=User CSS&f=style.css`:

<br />
<b>Fatal error</b>:  Uncaught Error: Call to undefined function httpConditional() in /var/www/FreshRSS/app/Controllers/extensionController.php:333
Stack trace:
#0 /var/www/FreshRSS/lib/Minz/Dispatcher.php(117): FreshRSS_extension_Controller-&gt;serveAction()
#1 /var/www/FreshRSS/lib/Minz/Dispatcher.php(48): Minz_Dispatcher-&gt;launchAction()
#2 /var/www/FreshRSS/lib/Minz/FrontController.php(61): Minz_Dispatcher-&gt;run()
#3 /var/www/FreshRSS/p/i/index.php(59): Minz_FrontController-&gt;run()
#4 {main}
  thrown in <b>/var/www/FreshRSS/app/Controllers/extensionController.php</b> on line <b>333</b><br />

### FreshRSS version

1.27.1-dev

Inverle:
* Closes and fixes error from #7885
* `no-cache.txt` is now respected in `f.php`, `ext.php` and `serve` action in `extensionController`

### DIFF
--- a/lib/http-conditional.php
+++ b/lib/http-conditional.php
@@ -80,7 +80,7 @@ $_sessionMode = false;
  * @return bool True if the connection can be closed (e.g.: the client has already the latest version), false if the new content has to be send to the client.
  */
 function httpConditional(int $UnixTimeStamp, int $cacheSeconds = 0, int $cachePrivacy = 0, bool $feedMode = false, bool $compression = false, bool $session = false): bool {
-	if (headers_sent()) return false;
+	if (headers_sent() || (defined('NO_CACHE') && constant('NO_CACHE') === true)) return false;
 
 	if (is_string($_SERVER['SCRIPT_FILENAME'] ?? null)) $scriptName = $_SERVER['SCRIPT_FILENAME'];
 	elseif (is_string($_SERVER['PATH_TRANSLATED'] ?? null)) $scriptName = $_SERVER['PATH_TRANSLATED'];
--- a/lib/http-conditional.php
+++ b/lib/http-conditional.php
@@ -80,7 +80,7 @@ $_sessionMode = false;
  * @return bool True if the connection can be closed (e.g.: the client has already the latest version), false if the new content has to be send to the client.
  */
 function httpConditional(int $UnixTimeStamp, int $cacheSeconds = 0, int $cachePrivacy = 0, bool $feedMode = false, bool $compression = false, bool $session = false): bool {
-	if (headers_sent() || (defined('NO_CACHE') && constant('NO_CACHE') === true)) return false;
+	if (headers_sent()) return false;
 
 	if (is_string($_SERVER['SCRIPT_FILENAME'] ?? null)) $scriptName = $_SERVER['SCRIPT_FILENAME'];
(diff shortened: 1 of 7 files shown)


lib/http-conditional.php
```diff
@@ -80,7 +80,7 @@
  * @return bool True if the connection can be closed (e.g.: the client has already the latest version), false if the new content has to be send to the client.
  */
 function httpConditional(int $UnixTimeStamp, int $cacheSeconds = 0, int $cachePrivacy = 0, bool $feedMode = false, bool $compression = false, bool $session = false): bool {
-	if (headers_sent()) return false;
+	if (headers_sent() || (defined('NO_CACHE') && constant('NO_CACHE') === true)) return false;
```
Alkarex: I would rather like to avoid modifying this file if possible, which is meant to be a standalone one.
I would prefer keeping the `if (!file_exists(DATA_PATH . '/no-cache.txt'))` test (also to make it easier to understand, in my opinion).

Alkarex:
Isn't the fix to simply include `require(LIB_PATH . '/http-conditional.php');` unconditionally?

Inverle:
> Isn't the fix to simply include `require(LIB_PATH . '/http-conditional.php');` unconditionally?

I think we don't want to call `httpConditional()` if `no-cache.txt` exists
I didn't want to put `file_exists` in each condition, but I can do that instead if we have to avoid making changes to http-conditional
(since that can also be forgotten later for such a niche feature)

Alkarex:
Yes please 👍🏻

Inverle:
@Alkarex done

p/f.php
```diff
@@ -8,9 +8,12 @@
 header("Content-Security-Policy: default-src 'none'; frame-ancestors 'none'; sandbox");
 header('X-Content-Type-Options: nosniff');
 
+$no_cache = file_exists(DATA_PATH . '/no-cache.txt');
+
 function show_default_favicon(int $cacheSeconds = 3600): void {
+	global $no_cache;
 	$default_mtime = @filemtime(DEFAULT_FAVICON) ?: 0;
-	if (!httpConditional($default_mtime, $cacheSeconds, 2)) {
+	if ($no_cache || httpConditional($default_mtime, $cacheSeconds, 2)) {
```
Alkarex: Isn't a `!` missing there?

Inverle: Fixed

p/f.php
```diff
@@ -59,11 +62,11 @@ function show_default_favicon(int $cacheSeconds = 3600): void {
 	}
 }
 
-if (!httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
+if ($no_cache || httpConditional($ico_mtime, mt_rand(14, 21) * 86400, 2)) {
```
Alkarex: And here as well with `!`, right?

Alkarex (approved):

PR #7917, opened 2025-09-05
Add new file in i18n

aledeg:
When manipulating I18N files, it is now possible to add a new file to all
languages. This action is available both in the manipulation script and
the makefile.

- add a new translation file from cli

1. create a new i18n file from cli or make (ex: `xxx.php`)
2. validate that the `xxx.php` file is created empty.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,14 @@ node_modules/.bin/rtlcss:
 	npm install
 
 ##@ I18n
+.PHONY: i18n-add-file
+i18n-add-file: ## Add a new translation file to all supported languages
+ifndef key
+	$(error To add a new file, you need to provide one in the "key" variable)
+endif
+	@$(PHP) ./cli/manipulate.translation.php --action add --key $(key)
+	@echo File added.
+
 .PHONY: i18n-add-key
 i18n-add-key: ## Add a translation key to all supported languages
 ifndef key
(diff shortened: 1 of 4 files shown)


Frenzie:
> When supplied with an unknown root key, we expect to create a new i18n file instead of raising an exception.

Should it do both? Otherwise in case of a typo you'd suddenly have a completely new language to delete. So an exception unless run with `-c` (for create), for example.

aledeg:
It does not create a new language. It creates a new translation file alongside `conf.php`, `sub.php`, etc for all languages.

If there is a typo, I think it's easy to revert with a simple command like `find app/i18n -iname \*xxx.php* -delete` if we've created `xxx.php` files.

aledeg:
If you try to update a key with an unknown root key, you still have the exception.

Frenzie:
> If there is a typo, I think it's easy to revert with a simple command like find app/i18n -iname \*xxx.php* -delete if we've created xxx.php files.

I don't mean from that perspective, but from the perspective of not noticing for 10 minutes and suddenly needing to try to merge `conf.php` and `comf.php`

aledeg:
> > If there is a typo, I think it's easy to revert with a simple command like find app/i18n -iname *xxx.php* -delete if we've created xxx.php files.
> 
> I don't mean from that perspective, but from the perspective of not noticing for 10 minutes and suddenly needing to try to merge `conf.php` and `comf.php`

I see what you mean. I'll figure something else.

aledeg:
@Frenzie I've change the whole thing to introduce an `add-file` feature. This way, there is no side-effect on other section of the code.

Frenzie:
Sounds good to me. :+1:

PR #7924, opened 2025-09-06
Implement whitelist for SimplePie sanitizer

Inverle:
ref: https://github.com/FreshRSS/FreshRSS/pull/7770#issuecomment-3140334326

https://github.com/FreshRSS/simplepie/pull/53


Inverle:
Still have to blacklist `javascript:` URIs but maybe in a later PR
(unless it's fine to do it in this one)

Also continue work on topic of https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-f6r4-jrvc-cfmr#advisory-comment-123323

Frenzie:
If the `<blink>`, `<marquee>` (and `<plaintext>`) behavior is changed (and I don't mind, but I think it's a change in behavior worth pointing out) I think the less offensive `<font>` should probably also be whitelisted.

In practice it's all fairly moot since when do you ever see such things anymore anyway.

Inverle:
I thought about allowing `<font>` but it may result in broken content (due to setting dark color on dark background f.e. `<font color="black">some text</font>`)

`<blink>` doesn't seem supported by browsers anymore
I haven't tested `<marquee>`
And why could `<plaintext>` be an issue?
...
Oh okay I thought `<plaintext>` behaved differently:

image

Frenzie:
`marquee` works. I didn't realize `blink` had been phased out in Firefox 23.

> Oh okay I thought `<plaintext>` behaved differently:

I thought it was fine as long as it was closed (which possibly isn't worth the risk but `pre` does have similar issues). But we'd have to check mid-'90s browsers to gauge the actually correct behavior, as opposed to the current vestigial possibly completely broken behavior.

For clarity probably just comment it out instead of removing it?

Inverle:
Yes I know `<marquee>` works, but I'm not sure how it would look in FreshRSS yet ;)

I think `<pre>` is fine to use, unlike `<plaintext>`

> For clarity probably just comment it out instead of removing it?

Sure

Frenzie:
`pre` is closer to an absolute requirement than just "fine." :-) Implicitly I guess I'm trying to say there's more to it than just whitelist/blacklist and I'm not immediately sure how any of that is currently handled. It doesn't directly relate to this PR unless the PR changes possible auto-closing behavior.

Inverle:
I am not familiar with any issues that `<pre>` may cause

Though indeed auto-closing behavior is currently changed due to `<plaintext>`. (once I comment it out, it won't be anymore)
SimplePie does try to close `<plaintext>` if it's not closed, but it's not aware of the fact that it doesn't change anything.

Alkarex:
Looks good at a quick glance, and let's test this code here, but then it cannot be merged directly. The SimpePie changes need to be made as a PR to https://github.com/FreshRSS/simplepie (where we can merge fast) as well as upstream https://github.com/simplepie/simplepie (where it risk taking a long time)

Alkarex:
From memory, `<plaintext>` is not supposed to be closed. Everything from there is just plain text. It should be filtered out

Inverle:
Yes it's filtered out now

lib/simplepie/simplepie/src/Sanitize.php
```diff
@@ -236,6 +240,26 @@ public function strip_htmltags($tags = ['base', 'blink', 'body', 'doctype', 'emb
...
     }
 
+    /**
+     * @param array<string,array> $tags Set array of allowed tags and attributes.
+     * @return void
+     */
+    public function whitelist_tags(array $tags = [])
+    {
+        $this->strip_htmltags = [];
+        $this->strip_attributes = [];
```
Inverle, line 255: Is this a good idea?

Frenzie: As a user of the library I'd be annoyed if it did anything else. That is, if I put `class` or `onsomething` into the whitelist (without which it'll be stripped regardless) and nothing happened because it was still being stripped.

Inverle: Ok, just making sure if there isn't a better way

It's specifically to reset the default values
https://github.com/FreshRSS/FreshRSS/blob/de624dc8ce63ec819c61216d9d44f828841c293e/lib/simplepie/simplepie/src/Sanitize.php#L46-L50

Frenzie: Do you mean you'd rather have a setting for which of the two to use while leaving those two arrays as is?

Inverle: I was wondering about that but probably not a setting

Right now there's the problem that `data-*` is allowed, so I'll have to think about a better solution other than calling `strip_attributes()` after `whitelist_tags()`.

It would be better if the order didn't matter in this case

Inverle:
https://github.com/FreshRSS/simplepie/pull/53

Frenzie:
I can see the case for contenteditable (though for a blogpost saying "look at this contenteditable" it'd be a bit disappointing) but why noscript/noembed/noframes and xmp? With particular emphasis on noscript.

Inverle:
> why noscript/noembed/noframes and xmp?

Potential mXSS risk e.g.
```html
<xmp><a href="</xmp><svg/onload=alert(1)>">
```

This could also be done with `textarea` or `iframe` which we allow instead of `xmp`, and additionally SimplePie encodes tags inside attributes (as it should):

```html
<textarea><a href="&lt;/textarea&gt;&lt;svg/onload=alert(1)&gt;"></a></textarea>
```

Despite that, I find it risky.

> With particular emphasis on noscript.

I would argue `<xmp>` is more useful. It's the only tag I found in use in my feeds out of those[1].
(with that in mind, I may put it back)

[1]: Actually `<noscript>` was already being stripped out which is probably why I can't find it (I'm searching through the SQLite of my main instance, not running code from this PR)

Frenzie:
> [1]: Actually `<noscript>` was already being stripped out which is probably why I can't find it (I'm searching through the SQLite of my main instance, not running code from this PR)

Right, on second thought I was wrong, or at least not completely right. We are operating a noscript environment here from the perspective of the feed content, so we want to see what's inside the `noscript`.

But I think that kind of thing should be made explicit for readers.
```
// 'noscript' => [], // not whitelisted because we want to see the contents
```

The same applies to the other two.

As an aside I do have multiple noscripts but they're messed up:

image


> > why noscript/noembed/noframes and xmp?
> 
> Potential mXSS risk e.g.
> 
> ```
> <xmp><a href="</xmp><svg/onload=alert(1)>">
> ```

It seems that if there's a risk there it comes from stripping `<xmp>`, doesn't it?

Inverle:
> // 'noscript' => [], // not whitelisted because we want to see the contents
> We are operating a noscript environment here from the perspective of the feed content, so we want to see what's inside the noscript

Not necessarily, my idea is that it's fine to allow `<noscript>` after all, so that some unnecessary content isn't visible.

If `<noscript>` would be stripped out then `<noscript><p>x</p></noscript>` would turn into `<p>x</p>` and that could be unwanted.

But I guess that depends on the feed.

> It seems that if there's a risk there it comes from stripping `<xmp>`, doesn't it?

No, why?

Frenzie:
> If `<noscript>` would be stripped out then `<noscript><p>x</p></noscript>` would turn into `<p>x</p>` and that could be unwanted.

On the contrary, that's very wanted and the reason it exists. :-) See the above as a concrete example. It's just that something weird happened somewhere along the chain. But you could turn it into `<noscript-stripped>` or something to allow for a simple CSS selector to hide it for one or all feeds in CustomCSS.

> No, why?

What I mean is any potential risk isn't from `<xmp>` at all but from the stripper not understanding `<xmp>` properly. In any case, my point is that as a reader I want to see that there's a rationale and that they weren't overlooked (give or take some of the most obvious like `script` and `on…`). Otherwise I'll think 'huh, I guess they don't know about `<xmp>`, or "looks like they forgot about `<track>`, etc.

Inverle:
> What I mean is any potential risk isn't from <xmp> at all but from the stripper not understanding <xmp> properly. In any case, my point is that as a reader I want to see that there's a rationale and that they weren't overlooked (give or take some of the most obvious like script and on…). Otherwise I'll think 'huh, I guess they don't know about <xmp>, or "looks like they forgot about <track>, etc.

Indeed I overlooked this, but I assume it would be quite dangerous to just ignore content inside `<xmp>`.

It is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/xmp) after all.

Right now if you have `<xmp><script>alert(1)</script></xmp>` the `<script>` will be stripped out.

Although the same happens with `<textarea><script>alert(1)</script></textarea>`, which is potentially a bug, and it should be turned into `<textarea>&lt;script&gt;alert(1)&lt;/script&gt;</textarea>` instead.

Honestly I'm not sure whether to keep it or remove it.

Frenzie:
> It is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/xmp) after all.

Unless [Google gets ideas](https://www.canistilluse.com/) it's not suddenly going to stop working. All engines have had automated regression tests for something so basic for literal decades. (But this isn't an argument in favor of keeping it since the impact of stripping it is effectively nothing, which doesn't compare favorably to even a fairly low risk.)

Opera (Presto) didn't support `<xmp>` though, iirc? So if that's indeed the case it was potentially quite a dangerous element 20 years ago.

Frenzie:
PS Ideally it wouldn't strip but convert to `pre` with the contents as `<` → `&lt;` and such. But as per above, on account of the exceedingly low impact it's one of those things that's only worth it if you're into it. ;-)

Inverle:
Good idea. I can probably do that.
Maybe not in scope of this PR though?

Frenzie:
Whatever you want I'd say? ;-) It seems on topic for the PR to me.

Alkarex:
Pending: https://github.com/FreshRSS/simplepie/pull/53#issuecomment-3427126682

lib/lib_rss.php
```diff
@@ -366,22 +366,165 @@ function customSimplePie(array $attributes = [], array $curl_options = []): \Sim
...
+		'mtext' => [],
+		'mtr' => ['columnalign', 'rowalign'],
+		'munder' => ['accentunder'],
+		'munderover' => ['accent', 'accentunder'],
+		// TODO: Support SVG after sanitizing and URL rewriting of xlink:href
+	]);
+	$simplePie->strip_attributes([
+		'data-auto-leave-validation', 'data-field', 'data-form', 'data-help', 'data-input', 'data-leave-validation', 'data-method',
+		'data-no-leave-validation', 'data-original', 'data-type', 'data-unread', 'data-url',
+	]);
```
Alkarex, line 526: Are all of them necessary?
~~Or alternatively, do we need to allow data-* attributes at all?~~

Inverle: Well they may be useful for User JS / User CSS or extensions, so probably yes

It would be easier if FreshRSS data attributes had a prefix e.g. `data-frss-leave-validation`

Alkarex: I mean, do we need to strip all those `data-*` attributes such as `data-unread` ?

Inverle: I just picked out all that I could find being used in JS, `data-leave-validation` can be used to annoy the user, `data-original` can cause issues with lazy loading, and the rest I haven't reviewed.

Requires reviewing by searching for each `data-attr` and `dataset.attr` in JS/CSS to see which ones are harmless
(maybe `dataset[attr]` too)

Alkarex: Yes, there are probably only a couple of them that should be removed, I think

Alkarex: Simplified in https://github.com/FreshRSS/FreshRSS/pull/7924/commits/2e06af6228690594dcc663954f84bfdc2c7ada79
Double-checks welcome.
And the remaining `data-original` should be removed when we update to native `loading="lazy"` https://github.com/FreshRSS/FreshRSS/issues/2371

Inverle: I would still keep `data-leave-validation`, `data-auto-leave-validation`, `data-no-leave-validation` as written above

Alkarex: Restored for now, but let's follow-up to remove them in another PR, as I do not find that so clean.
None of this logic is supposed to run on article content, and at a quick glance, this looks like it is indeed correct.
If there is any problematic path, it should be fixed.

Inverle: > None of this logic is supposed to run on article content, and at a quick glance, this looks like it is indeed correct.

Hmm, maybe it is correct that `strip_attributes` isn't needed, but it also might be bad to assume that `data_auto_leave_validation()` or similar won't be called later (even by some UserJS/extension)

edit: `extra.js` is loaded after opening a slider (feed settings), but `data_auto_leave_validation(document.body)` is called only on actions other than `'normal', 'global', 'reader'`

Alkarex: By the way @Inverle , I have sent you an e-mail on your commit e-mail address. Please take a look when you can.

Inverle: @Alkarex I replied

Alkarex: @Inverle One more e-mail :-)

PR #7925, opened 2025-09-06
Update FreshRSS-screenshot.png

math-GH:
Update the FreshRSS screenshot used in the readme.md (English and French)

Old screenshot (from 2024):

FreshRSS-screenshot

New screenshot (from February 2025) - this screenshot is currently used on freshrss.org

grafik


Inverle:
Also was on my TODO list.

Updating the screenshot could be later automated with [Playwright](https://playwright.dev/) maybe (or some other headless browser tool)

math-GH:
Once in ~10~ 5 years would be enough :D

Inverle:
I assume the PNG could be compressed (further?) too

related: https://github.com/FreshRSS/freshrss.org/pull/79

Inverle:
I noticed that the original screenshot from `freshrss.org` is a way smaller WebP (`60 KiB` instead of `216 KiB`, also better quality)
Any reason not to switch to WebP?

Frenzie:
> I assume the PNG could be compressed (further?) too

This type of screenshot can be exposed to pngquant (lossy) with no perceptible loss in quality at about a third of the size.

> I noticed that the original screenshot from freshrss.org is a way smaller WebP (60 KiB instead of 216 KiB, also better quality)

Mind, the blurriness (which I'd rather do without) likely contributes to increasing the filesize in this instance.

math-GH:
I would be fine with WebP too as it is used on the website. Currently I just replaced the exisiting PNG with a new one.

@Alkarex What do you think? WebP is widely supported by all/the most browsers. Any objections?

Frenzie:
As stated the screenshot on the website is already WebP. ;-)

Alkarex:
I would suggest a screenshot with a bit more information like in the old one (e.g., multiple feeds, some unread articles, some labels...)

PR #7934, opened 2025-09-09
Use `Minz_Translate::exists()` for language check in `createUser()`

Inverle:
(consistency)

Related / follow-up: #7878

### DIFF
--- a/app/Controllers/userController.php
+++ b/app/Controllers/userController.php
@@ -305,8 +305,7 @@ class FreshRSS_user_Controller extends FreshRSS_ActionController {
 		$configPath = '';
 
 		if ($ok) {
-			$languages = Minz_Translate::availableLanguages();
-			if (empty($userConfig['language']) || !in_array($userConfig['language'], $languages, true)) {
+			if (!Minz_Translate::exists(is_string($userConfig['language']) ? $userConfig['language'] : '')) {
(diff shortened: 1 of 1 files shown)


app/Controllers/userController.php
```diff
@@ -305,8 +305,7 @@ public static function createUser(string $new_user_name, ?string $email, string
 		$configPath = '';
 
 		if ($ok) {
-			$languages = Minz_Translate::availableLanguages();
-			if (empty($userConfig['language']) || !in_array($userConfig['language'], $languages, true)) {
+			if (!Minz_Translate::exists(is_string($userConfig['language']) ? $userConfig['language'] : '')) {
```
aledeg, line 308: Does the following works? It could be simpler
```suggestion
			if (!Minz_Translate::exists($userConfig['language'] ?? '')) {
```

Inverle: No it doesn't
```
 ------ -------------------------------------------------------------------------------------------
  Line   app/Controllers/userController.php
 ------ -------------------------------------------------------------------------------------------
  308    Parameter #1 $lang of static method Minz_Translate::exists() expects string, mixed given.
         🪪  argument.type
 ------ -------------------------------------------------------------------------------------------
 ```

PR #7937, opened 2025-09-10
add collection of scraping settings

### LINKED ISSUE #7894: [Feature] Docs: Examples of fetch full-text & XPath scraping

The ability to fetch full-text and doing XPath-based scraping to turn any website into a feed is really amazing and powerful. At the same time, both can require some effort to figure out the best configurations to make specific pages work.

I had a look if there's any repo/collection of examples, which doesn't seem to be the case yet. So I made [a small repo to collect examples](https://codeberg.org/gedankenstuecke/freshrss-fulltext-settings), with the idea that it could maybe help avoid duplication of effort, if other folks want to set up any of those feeds. And of course also the hope that others would share their own config for specific pages, slowly building a collection of setups that can be copied by folks (a [longer rational is here](https://tzovar.as/fulltext-freshrss/)). 

What do folks think, would this generally be something useful for other FreshRSS users, and if so: would it be worth linking to it from the scraping docs? If there's an interest I'd be happy to make a PR to add a link, but wanted to first see if you feel that's useful or not!

(I'm not 100% sure if this is the best place to open this, but given that the docs are rolled into this repo I thought so. If not, please point me to a better direction ❤️)

### Use case

_No response_

Just adding a link to the docs to point interested users to potential scraping/full text configs they could copy.

_No response_

### Additional context

_No response_

gedankenstuecke:
Closes #7894

- adds the link to the collection of XPath scraping settings to the appropriate documentation page

Pull request checklist:

- [x] documentation updated

### DIFF
--- a/docs/en/users/11_website_scraping.md
+++ b/docs/en/users/11_website_scraping.md
@@ -59,3 +59,4 @@ Support string concatenation with a syntax like: `meta.title & " some text"` usi
 ## Recommended external manuals
 
 - [XPath Scraping with FreshRSS, by Dan Q](https://danq.me/2022/09/27/freshrss-xpath/) (September 2022)
+- A [crowdsourced collection of XPath/scraping settings](https://codeberg.org/gedankenstuecke/freshrss-xpath-scraping-settings-examples) for various websites (incl. News pages). Also has OPML files to easily import settings. (August 2025)


aledeg (approved):

Alkarex:
Suggestions of improvements:
1. Organise the repo to scale with many examples, meaning a structure probably based on breaking down the DNS (and maybe even more) like:

```
- com
- org
    - freshrss
        - something.opml.xml
```

2. Implement in FreshRSS the ability to import an OPML by URL
3. Implement in FreshRSS the ability to keep track of a remote OPML for settings changes, similarly to what we do for categories with *Dynamic OPML`

gedankenstuecke:
Thanks for the suggestions @Alkarex! I agree those improvements would be great! 

My idea was to use language instead of DNS as the first-level grouping, as I think people would prefer using that for filtering - or at least discovery of feeds (and there's no perfect overlap between TLD & languages, including in the existing examples).

PR #7939, opened 2025-09-10
Rewrite PHP opening tags

aledeg:
This allows to remove the use of `echo` and be consistent through out the file. Some empty PHP tags where removed as well.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/views/index/normal.phtml
+++ b/app/views/index/normal.phtml
@@ -61,26 +61,23 @@ $today = @strtotime('today');
 			$this->entry->feed() ?? FreshRSS_Feed::default();
 
 		if ($display_today && $this->entry->isDay(FreshRSS_Days::TODAY, $today)) {
-			?><div class="day" id="day_today"><?php
-				echo _t('index.feed.received.today');
-				?><span class="date"> — <?= timestamptodate(time(), false) ?></span><?php
-				?><span class="name"><?= FreshRSS_Context::$name ?></span><?php
-			?></div><?php
+			?><div class="day" id="day_today"><?= _t('index.feed.received.today') ?>
+			<span class="date"> — <?= timestamptodate(time(), false) ?></span>
+			<span class="name"><?= FreshRSS_Context::$name ?></span>
+			</div><?php
 			$display_today = false;
 		}
 		if ($display_yesterday && $this->entry->isDay(FreshRSS_Days::YESTERDAY, $today)) {
-			?><div class="day" id="day_yesterday"><?php
-				echo _t('index.feed.received.yesterday');
-				?><span class="date"> — <?= timestamptodate(time() - 86400, false) ?></span><?php
-				?><span class="name"><?= FreshRSS_Context::$name ?></span><?php
-			?></div><?php
+			?><div class="day" id="day_yesterday"><?= _t('index.feed.received.yesterday') ?>
+				<span class="date"> — <?= timestamptodate(time() - 86400, false) ?></span>
+				<span class="name"><?= FreshRSS_Context::$name ?></span>
+			</div><?php
 			$display_yesterday = false;
 		}
 		if ($display_others && $this->entry->isDay(FreshRSS_Days::BEFORE_YESTERDAY, $today)) {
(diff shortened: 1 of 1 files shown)


aledeg:
If I am not mistaken, I think some PHP tags were useless.

Frenzie:
It does seem strange that there's a seemingly random `echo` in between the `<?=`.

> If I am not mistaken, I think some PHP tags were useless.

Slightly mistaken I believe. The line-break and indentation are hidden in the output that way without needing post-processing.

aledeg:
> It does seem strange that there's a seemingly random `echo` in between the `<?=`.

I am not sure to understand what you mean. I replaced `<?php echo` to use `<?=` instead.


> > If I am not mistaken, I think some PHP tags were useless.
> 
> Slightly mistaken I believe. The line-break and indentation are hidden in the output that way without needing post-processing.

Still not sure to understand. No matter how many line breaks, white spaces or tabs you have, the output will be the same in the browser.

Frenzie:
I mean the prior situation with the echo seemed surprising. :-)

> Still not sure to understand.

Whoever wrote it didn't want it to be in the HTML at all. Spaces are collapsed into a single space but that's not the same thing as no space at all.

aledeg:
I see.
But with the modification, there is still a space between HTML tags since they are on different lines.

Alkarex:
The `echo` statements were forgotten when the move to `<?= ... ?>` was done, so good fix.

The other changes are fine as well. In code sections used many times (such as in article headers), I am sometimes stripping the white-space to reduce output size, but here, this code is outputted only once, so it does not matter much.

The new PR produces this kind of white-space:

```html
<div class="day" id="day_before_yesterday">Reçu avant avant-hier				<span class="name"></span>
			</div>
```

Alkarex (approved):

aledeg:
I understand the idea behind the empty tags now. Maybe we could have some kind of post-processor to remove the overhead of white spaces. But then we have an overhead of processing time.

When using template engines (blade, twig, ...), they have options to remove unnecessary white spaces. Maybe we should add that as well.

math-GH:
I am late to the party and want to second @Frenzie

app/views/index/normal.phtml
```diff
@@ -61,26 +61,23 @@ $today = @strtotime('today');
 			$this->entry->feed() ?? FreshRSS_Feed::default();
 
 		if ($display_today && $this->entry->isDay(FreshRSS_Days::TODAY, $today)) {
-			?><div class="day" id="day_today"><?php
-				echo _t('index.feed.received.today');
-				?><span class="date"> — <?= timestamptodate(time(), false) ?></span><?php
-				?><span class="name"><?= FreshRSS_Context::$name ?></span><?php
-			?></div><?php
+			?><div class="day" id="day_today"><?= _t('index.feed.received.today') ?>
+			<span class="date"> — <?= timestamptodate(time(), false) ?></span>
```
math-GH, line 65: now there is more than 1 space before the `—`

math-GH: maybe it is a good point to remove the `—` and replace it with CSS because this character is used for styling here. CSS would be the better solution IMHO

Alkarex:
> Maybe we could have some kind of post-processor to remove the overhead of white spaces. But then we have an overhead of processing time.

Yes, I would not be in favour of using any other layer of processing. There are only a few places where I think we should make an effort with white-space and that can be done with native syntax easily enough.

Alkarex:
A little follow-up https://github.com/FreshRSS/FreshRSS/pull/7941

app/views/index/normal.phtml
Alkarex: PR welcome if CSS is better. Is there any regression here? If so, can be fixed in https://github.com/FreshRSS/FreshRSS/pull/7941

Frenzie: Like I said, no space is different than a space. Any number of tabs, spaces and newlines are collapsed into a space but that's fundamentally distinct from no space.

PR #7942, opened 2025-09-10
configurable notification timeout

### LINKED ISSUE #5466: Pop-up about feeds marked as read
Hello,

any way to disable it? It's pretty annoying to see that every time I mark news read.

math-GH:
Ref #7931
Ref #5466
Ref #6409

Still a draft but feedback already welcome

added configuration in "Display"
grafik


app/Controllers/entryController.php
```diff
@@ -190,7 +190,8 @@ public function readAction(): void {
 					'a' => 'index',
 					'params' => $params,
 				],
-				'readAction'
+				'readAction ',
+				(bool) FreshRSS_Context::userConf()->good_notification_timeout,
```
math-GH: @Alkarex @Frenzie @Inverle : could you please give me feedback?
Is it a good approach to add a bool parameter here? Is it okay how it is done?
I did this commit as example for verification if it is good or not. (back end development is not my best side :D)

Frenzie: I think it'd be much clearer to simply write `good_notification_timeout > 0` than to cast to boolean, which seems more appropriate when dealing with 0 or 1.

Inverle: Yes I thought the same, also preferably add `showNotification: ` (named parameter) to specify what the boolean is

math-GH: thank you very much :)

math-GH:
@Frenzie : Do you have better ideas how to name the configurations?

grafik

Frenzie:
"Confirmation (notification) timeout" and "Warning (notification) timeout"?

math-GH:
Would be "banner" instead of "notification" also good?

What about "Show confirmation banner [_____} seconds"?

aledeg:
Maybe using a range input makes more sense here since you've defined your number input with a min and a max value.

math-GH:
> Maybe using a range input makes more sense here since you've defined your number input with a min and a max value.

The max is 60 seconds. Much more than any user would define. I would expect most users will set 0 till 5. A slider would make the UI worse

Frenzie:
> What about "Show confirmation banner [_____} seconds"?

I don't think that works. I'd go with something like 'confirmation notice' or 'acknowledgement' to prevent making it sound like a confirmation request.

But banner should be fine, or at least Android uses it that way.

math-GH:
Thanks for your input.

The labels are now updated:
grafik

app/i18n/en/conf.php
```diff
@@ -105,6 +105,16 @@
...
 		'small' => 'Small',
 	),
+	'notification_timeout' => array(
+		'bad' => array(
+			'label' => 'Show warning banner',	// TODO
+			'seconds' => 'seconds (minimum 1)',	// TODO
+		),
+		'good' => array(
+			'label' => 'Show acknowledgement banner',	// TODO
+			'seconds' => 'seconds (0 means disabled notifications)',	// TODO
```
Frenzie: Since there are no notifications in the label:
```suggestion
			'seconds' => 'seconds (0 means not shown)',
```

```suggestion
			'seconds' => 'seconds (0 means disabled)',
```

app/i18n/en/conf.php
```diff
@@ -105,6 +105,16 @@
 		'none' => 'None',
 		'small' => 'Small',
 	),
+	'notification_timeout' => array(
+		'bad' => array(
+			'label' => 'Show warning banner',	// TODO
+			'seconds' => 'seconds (minimum 1)',	// TODO
```
Frenzie: "minimum 1" doesn't strike me as proper English (I assume it's a bit of a `mindestens 1` sneaking in)
```suggestion
			'seconds' => 'seconds (at least 1)',
```

```suggestion
			'seconds' => 'seconds (a minimum of 1)',
```

app/Controllers/configureController.php
```diff
@@ -74,14 +74,23 @@ public function displayAction(): void {
 			FreshRSS_Context::userConf()->bottomline_date = Minz_Request::paramBoolean('bottomline_date');
 			FreshRSS_Context::userConf()->bottomline_link = Minz_Request::paramBoolean('bottomline_link');
 			FreshRSS_Context::userConf()->show_nav_buttons = Minz_Request::paramBoolean('show_nav_buttons');
-			FreshRSS_Context::userConf()->html5_notif_timeout = Minz_Request::paramInt('html5_notif_timeout');
+			FreshRSS_Context::userConf()->html5_notif_timeout =
+				Minz_Request::paramInt('html5_notif_timeout') < 0 ? 0 : Minz_Request::paramInt('html5_notif_timeout');
```
Alkarex:
```suggestion
			FreshRSS_Context::userConf()->html5_notif_timeout = max(0, Minz_Request::paramInt('html5_notif_timeout'));
```

Maybe a less verbose syntax, also for the lines below

app/i18n/pl/conf.php
```diff
@@ -105,6 +105,16 @@
...
+	'notification_timeout' => array(
+		'bad' => array(
+			'label' => 'Show warning banner',	// TODO
+			'seconds' => 'seconds (at least 1)',	// TODO
+		),
+		'good' => array(
+			'label' => 'Show acknowledgement banner',	// TODO
+			'seconds' => 'seconds (0 means not shown)',	// TODO
+		),
+	),
```
Inverle, line 117:
```suggestion
	'notification_timeout' => array(
		'bad' => array(
			'label' => 'Pokaż baner ostrzeżenia',
			'seconds' => 'sekundy (przynajmniej 1)',
		),
		'good' => array(
			'label' => 'Pokaż baner potwierdzający',
			'seconds' => 'sekundy (0 oznacza nie pokazuj)',
		),
	),
```

app/Controllers/configureController.php
math-GH: nice trick. It is a bit tricky to read but very smart. Thanks!

Alkarex (approved):

PR #7945, opened 2025-09-10
Docker healthcheck

### LINKED ISSUE #7377: [Bug] Docker automatic Healthcheck seems to be missing
### Describe the bug

From pull requests https://github.com/FreshRSS/FreshRSS/pull/2438 and https://github.com/FreshRSS/FreshRSS/pull/2455 and subsequently https://github.com/FreshRSS/FreshRSS/pull/2451 it appears that an automatic healthcheck was implemented for Docker images a long time ago, but as of release 1.26.0 the FreshRSS container does not report a status other than "Running" even if things are working correctly. Furthermore, running `docker inspect <container_name>` does not show a Healthcheck element anywhere in the JSON.

### To Reproduce

1. Deploy FreshRSS 1.26.0 in Docker.
2. After the container is up and running, run `docker inspect <container_name> | grep Healthcheck`
3. There is no Healthcheck element present in the JSON

### Expected behavior

It appears there should be a built-in Healthcheck in the image from past correspondence.

### FreshRSS version

1.26.0

### System information

- PHP version: PHP 8.2
- Installation type: Dockerhub image

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7377

### DIFF
--- a/Docker/freshrss/docker-compose.yml
+++ b/Docker/freshrss/docker-compose.yml
@@ -23,3 +23,11 @@ services:
       TZ: Europe/Paris
       CRON_MIN: '3,33'
       TRUSTED_PROXY: 172.16.0.1/12 192.168.0.1/16
+    # # Optional healthcheck section:
+    # healthcheck:
+    #   test: ["CMD", "cli/health.php"]
+    #   timeout: 10s
+    #   start_period: 60s
+    #   start_interval: 11s
+    #   interval: 75s
+    #   retries: 3
--- a/cli/health.php
+++ b/cli/health.php
@@ -26,6 +26,6 @@ curl_setopt_array($ch, [
 $content = curl_exec($ch);
 $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
 
-if ($httpCode !== 200 || !is_string($content) || !str_contains($content, 'jsonVars')) {
+if ($httpCode !== 200 || !is_string($content) || !str_contains($content, 'jsonVars') || !str_contains($content, '</html>')) {
 	die(1);
 }
--- a/cli/health.php
+++ b/cli/health.php
@@ -1,0 +1,29 @@
+#!/usr/bin/env php
+<?php
+declare(strict_types=1);
+
+if (php_sapi_name() !== 'cli') {
+	fwrite(STDERR, 'Error: This script may only be invoked from command line!' . PHP_EOL);
+	die(2);
+}
+
+$options = getopt('', ['url::', 'connect_timeout::', 'timeout::']);
+$address = is_string($options['url'] ?? null) ? $options['url'] : 'http://localhost/i/';
+$ch = curl_init($address);
+if ($ch === false) {
+	fwrite(STDERR, 'Error: Failed to initialize cURL!' . PHP_EOL);
+	die(3);
+}
+curl_setopt_array($ch, [
+	CURLOPT_CONNECTTIMEOUT => is_numeric($options['connect_timeout'] ?? null) ? (int)$options['connect_timeout'] : 3,
+	CURLOPT_TIMEOUT => is_numeric($options['timeout'] ?? null) ? (int)$options['timeout'] : 5,
+	CURLOPT_ENCODING => '',	//Enable all encodings
+	CURLOPT_RETURNTRANSFER => true,
+]);
+$content = curl_exec($ch);
+$httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+curl_close($ch);
+
+if ($httpCode !== 200 || !is_string($content) || !str_contains($content, 'jsonVars')) {
+	die(1);
+}
--- a/cli/health.php
+++ b/cli/health.php
@@ -22,7 +22,6 @@ curl_setopt_array($ch, [
 ]);
 $content = curl_exec($ch);
 $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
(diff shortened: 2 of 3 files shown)


Docker/freshrss/docker-compose.yml
```diff
@@ -23,3 +23,11 @@ services:
...
       CRON_MIN: '3,33'
       TRUSTED_PROXY: 172.16.0.1/12 192.168.0.1/16
+    # # Optional healthcheck section:
+    # healthcheck:
+    #   test: ["CMD", "cli/health.php"]
+    #   timeout: 10s
+    #   start_period: 60s
+    #   start_interval: 11s
+    #   interval: 75s
+    #   retries: 3
```
schildbach, line 33: I'd indent this by one more space, because if you remove the `#` the YAML format will be invalid.

schildbach:
I just tested this locally, and the health check works. I didn't test a failure case however.

```
            "Health": {
                "Status": "healthy",
                "FailingStreak": 0,
                "Log": [
                    {
                        "Start": "2025-09-11T00:31:17.938021194+02:00",
                        "End": "2025-09-11T00:31:18.040876228+02:00",
                        "ExitCode": 0,
                        "Output": ""
                    }
                ]
            }
```

Docker/freshrss/docker-compose.yml
schildbach: Or actually I think one space needs to be removed on each of the lines.

cli/health.php
```diff
@@ -0,0 +1,31 @@
...
+	CURLOPT_ENCODING => '',	//Enable all encodings
+	CURLOPT_HTTPHEADER => [
+		'Connection: close',
+	],
+	CURLOPT_RETURNTRANSFER => true,
+]);
+$content = curl_exec($ch);
+$httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+
+if ($httpCode !== 200 || !is_string($content) || !str_contains($content, 'jsonVars')) {
```
schildbach: Maybe it's obvious to PHP experts, but for me a comment would be helpful on why is it expecting the magic string "jsonVars"?

Docker/freshrss/docker-compose.yml
Alkarex: It looks good to me.

Using automatic *Toggle line comment* from an IDE such as VSCode produces the following result (and it works as well in the other direction):

```yml
    environment:
      TZ: Europe/Paris
      CRON_MIN: '3,33'
      TRUSTED_PROXY: 172.16.0.1/12 192.168.0.1/16
    # Optional healthcheck section:
    healthcheck:
      test: ["CMD", "cli/health.php"]
      timeout: 10s
      start_period: 60s
      start_interval: 11s
      interval: 75s
      retries: 3
```

Did you try in an IDE (and in that case which one?) or did you uncomment manually line by line?

Note that the convention for this type of comment is the comment sign itself followed by a space, as hinted in the first line `# # Optional healthcheck section:`

Side note: This is another example where tabs for indenting (which I prefer) would have reduced confusion 🤓

cli/health.php
Alkarex: `jsonVars` is just something relatively unique, which is found in FreshRSS HTML responses:

https://github.com/FreshRSS/FreshRSS/blob/e7c9c3d11e9d5291acd31cd667504fba4e80a0c8/app/layout/layout.phtml#L34

This allows reporting an error if another kind of content is served at that address

Docker/freshrss/docker-compose.yml
schildbach: I'm used to adding/removing just one `#` char by hand. Anyway, if IDEs do it the way you describe then I guess I need to adapt.

cli/health.php
Frenzie: A status code is sent with the headers. If an error should occur along the way after the headers in rare cases you might receive a 200 with what's still a practical failure, in any language. This is fairly decent evidence that the first few hundred bytes should actually be correct in a way that something generic like `</html>` wouldn't be, though @Alkarex it might potentially be worth adding something like that as an additional safety check at the end?

```suggestion
if ($httpCode !== 200 || !is_string($content) || !str_contains($content, 'jsonVars') || !str_contains($content, '</html>')) {
```

I'm not sure if a comment can add much though. It's just a string of an above average uniqueness level that should be present but that's what the code itself already says. I suppose you could add that it's evidence a couple of templating functions already succeeded or something.

schildbach: Ok, thanks. I think I'd add that info as a code comment.

schildbach (commented):
Consider this approved, with or without the code comment.

Docker/freshrss/docker-compose.yml
Frenzie: I misunderstood the initial comment. The diff makes it look aligned with what's above.

> Anyway, if IDEs do it the way you describe then I guess I need to adapt.

Geany (Scintilla) does something clever I wish other editors did too btw: it'll write `//~`, `#~` etc. to distinguish a human comment from a machine-assisted comment. It nearly completely removes the accidental uncommenting that sometimes occurs in other editors.

Specifically, if you didn't add the extra comment in front of `# Optional healthcheck section:` you might accidentally turn it into a syntax error when uncommenting. Scintilla-style commenting makes it so you don't have to be nearly as precise in where you add or remove comments.

Frenzie (approved):
Script's good, didn't check the Docker part (but lgtm)

Issue #6470, opened 2024-05-17
[Bug] There ARE 1 new articles


AbelLykens:
### Describe the bug

![image](https://github.com/FreshRSS/FreshRSS/assets/47113175/f9900267-4fc5-445d-974b-d6a9a8ec8398)

_Are_ should be _is_ in these notifications. Love Freshrss so thanks!!

### To Reproduce

Get 1 new unread article, have notifications enabled.

### Expected behavior

Are should be is

### FreshRSS version

 1.23.1

### Environment information

- Database version: Mariadb
- PHP version: PHP
- Installation type: From source
-Web server type: Apache
- Device: PC
- OS: Hosted on Debian 12, viewing on Win11
- Browser: Firefox 126

math-GH:
I would say that it is worth to fix ;)

But keep in mind that different languages have different issues with numbers and grammar.

Here an example from Russian language: 2,3,4 have another grammar case then 6 upwards. (it repeats with 22,23,24, 32,33,34,42,43 and so on)

```
Есть 1 новая статья

Есть 2 новые статьи
Есть 3 новые статьи
Есть 4 новые статьи

Есть 5 новых статей
Есть 6 новых статей
```

Manitosh-Paul:
Hello team, I am a new contributor and want to contribute to open source.
This issue can be found in file: https://github.com/FreshRSS/FreshRSS/blob/5b197b8130efed7953f42cde0cbd2217b7fb0dbe/app/i18n/el/gen.php#L132C5-L132C22

However, there is one more issue with this bug i.e.,
If there is only one unread article the message should be like this: "There is %%d new article to read on FreshRSS." otherwise, the message should be: "There are %%d new articles to read on FreshRSS."

As, I am new contibutor in PHP too, I think we can use conditional statement something like this in line number 132:
'body_new_articles' => ('%%d'==1 ? 'There is %%d new article to read on FreshRSS.' : 'There are %%d new articles to read on FreshRSS.'),

Alkarex:
@Manitosh-Paul Thanks for trying to help :-)
As @math-GH is writing just above, the grammar of numbers is more complicated than just having two cases for singular and plural (although it would probably cover several languages well enough). So the first step is to list a couple of approaches, for instance by finding out how it is done in other projects (maybe WordPress for instance, although I have not sure they are facing this problem).

Manitosh-Paul:
[...]

Issue #6315, opened 2024-04-17
i18n: Query strings (refactoring)


math-GH:
> It is because they are used by string concatenation (which I would like to get rid of precisely because it makes validation and maintenance quite challenging). And there is another layer of property with another name for the getter, which makes it even more difficult to search:

https://github.com/FreshRSS/FreshRSS/blob/0ffcf41f9371b9321856d2c709cf7bcb4f7c3c60/app/views/configure/queries.phtml#L57

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/pull/6052#discussion_r1568296833_

PR #7947, opened 2025-09-10
POC - i18n modification

aledeg:
This is an attempt to translate without the need to concatenate strings and to pluralize.
This is not meant to be merge as-is, there is a lot of work before being ready.

At the moment, I've made a POC only for the state in the user query page (French only).

Nonetheless, it's worth testing it.

See #6470 & #6315

### DIFF
--- a/app/i18n/fr/conf.php
+++ b/app/i18n/fr/conf.php
@@ -153,7 +153,7 @@ return array(
 		'get_Z' => 'Tous les flux, y compris les archivés',
 		'get_all' => 'Afficher tous les articles',
 		'get_all_labels' => 'Afficher les articles avec une étiquette',
-		'get_category' => 'Afficher la catégorie <em>%s<em>',
+		'get_category' => 'Afficher la catégorie <em>{0}<em>',
 		'get_favorite' => 'Afficher les articles favoris',
 		'get_feed' => 'Afficher le flux <em>%s</em>',
 		'get_important' => 'Afficher les articles des flux importants',
@@ -182,22 +182,24 @@ return array(
 			'opml' => 'Lien partageable de la liste des flux au format OPML',
 			'rss' => 'Lien partageable du flux RSS',
 		),
-		'state_0' => 'Afficher tous les articles',
-		'state_1' => 'Afficher les articles lus',
-		'state_2' => 'Afficher les articles non lus',
-		'state_3' => 'Afficher tous les articles',
-		'state_4' => 'Afficher les articles favoris',
-		'state_5' => 'Afficher les articles lus et favoris',
-		'state_6' => 'Afficher les articles non lus et favoris',
-		'state_7' => 'Afficher les articles favoris',
-		'state_8' => 'Afficher les articles non favoris',
-		'state_9' => 'Afficher les articles lus et non favoris',
-		'state_10' => 'Afficher les articles non lus et non favoris',
-		'state_11' => 'Afficher les articles non favoris',
-		'state_12' => 'Afficher tous les articles',
-		'state_13' => 'Afficher les articles lus',
-		'state_14' => 'Afficher les articles non lus',
-		'state_15' => 'Afficher tous les articles',
+		'state' => 'Afficher {0, plural,
+			=0{tous les articles}
+			=1{les articles lus}
+			=2{les articles non lus}
+			=3{tous les articles}
+			=4{les articles favoris}
+			=5{les articles lus et favoris}
+			=6{les articles non lus et favoris}
+			=7{les articles favoris}
+			=8{les articles non favoris}
+			=9{les articles lus et non favoris}
+			=10{les articles non lus et non favoris}
+			=11{les articles non favoris}
+			=12{tous les articles}
+			=13{les articles lus}
+			=14{les articles non lus}
+			=15{tous les articles}
+			other{}}',
 		'title' => 'Filtres utilisateurs',
 	),
 	'reading' => array(
--- a/lib/Minz/Translate.php
+++ b/lib/Minz/Translate.php
@@ -249,8 +249,7 @@ class Minz_Translate {
 			}
 		}
 
-		// Get the facultative arguments to replace i18n variables.
-		return empty($args) ? $translation_value : vsprintf($translation_value, $args);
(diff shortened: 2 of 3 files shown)


app/i18n/fr/conf.php
```diff
@@ -182,22 +182,24 @@
...
+			=7{les articles favoris}
+			=8{les articles non favoris}
+			=9{les articles lus et non favoris}
+			=10{les articles non lus et non favoris}
+			=11{les articles non favoris}
+			=12{tous les articles}
+			=13{les articles lus}
+			=14{les articles non lus}
+			=15{tous les articles}
+			other{}}',
```
aledeg, line 202: This is pluralized only for the POC. For a real implementation, it's better to use `select` instead of `plural`.

lib/Minz/Translate.php
```diff
@@ -249,8 +249,7 @@ public static function t(string $key, ...$args): string {
 			}
 		}
 
-		// Get the facultative arguments to replace i18n variables.
-		return empty($args) ? $translation_value : vsprintf($translation_value, $args);
+		return msgfmt_format_message(self::language(), $translation_value, $args);
```
Frenzie, line 252: I think you forgot to add the most important piece of this PR? ;-)

app/i18n/fr/conf.php
Frenzie: Shouldn't there be placeholders in there somewhere? Since in most Western European languages you just have singular and plural, but even in languages with 9 plurals they still tend to have the same one for 20-100 and for 100+ for example. But maybe it's just a bit confusing because `msgfmt_format_message` is a bit of a mystery. :-)

Frenzie: Also I'd probably suggest to put the plural formula in the translation metadata itself rather than hardcoding it somewhere separately, though ultimately in this context it doesn't matter too much.

In GetText that looks like:

```
"Plural-Forms: nplurals=2; plural=n != 1;\n"
```
(Most Western European languages.)

```
"Plural-Forms: nplurals=4; plural=(n%10==1 && n%100!=11 ? 0 : n%10>=2 && "
"n%10<=4 && (n%100<12 || n%100>14) ? 1 : n%10==0 || (n%10>=5 && n%10<=9) || "
"(n%100>=11 && n%100<=14)? 2 : 3);\n"
```
(Notoriously difficult Russian plural forms.)

lib/Minz/Translate.php
aledeg: What do you mean?

app/i18n/fr/conf.php
aledeg: I am not sure the icu format works like that. I am pretty sure there is not support for ranges nor calculation. I was just playing around the thing to find if we could use it.

As mentioned in my comment, this is not how we should implement it for the state use-case. We should use a select instead:
```suggestion
		'state' => 'Afficher {0, select,
			0{tous les articles}
			1{les articles lus}
			2{les articles non lus}
			3{tous les articles}
			4{les articles favoris}
			5{les articles lus et favoris}
			6{les articles non lus et favoris}
			7{les articles favoris}
			8{les articles non favoris}
			9{les articles lus et non favoris}
			10{les articles non lus et non favoris}
			11{les articles non favoris}
			12{tous les articles}
			13{les articles lus}
			14{les articles non lus}
			15{tous les articles}
			other{}}',
```

I am not sure if there is a simple way to support complex pluralization (Russian) in a simple way. And yes, the format in that method is pretty complicated to go by.

Frenzie: GetText is quite simple. nplurals defines how many there are, and the formula provides a number pointing at one of them. (So for Dutch/French/English/German, simply 0 or 1.)

lib/Minz/Translate.php
Frenzie: I thought this was a custom GetText-like function. I didn't realize it was just a new PHP 5.3 built-in function.

aledeg: Yes, that's why I was limited to ICU syntax to describe translations. It's not custom code.

app/i18n/fr/conf.php
math-GH: It is not only Russian. As far as I can oversee it it is complicated in other Slavic languages. The rule is in general: 
1: easy
all numbers that are ending on 2, 3, or 4 (except 12,13,14; examples: 22,23,24,52,53,54,102,103,104 etc): handled different
others: simple plural.

aledeg: Oh, I need to make some more tests. Some keyword exists and they are used differently by the function depending on the locale.

See https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html

aledeg: I've made a new test (data is irrelevant) and it works.
```php
		'state' => 'Afficher {0, plural,
			=0{0000}
			one{1111}
			two{2222}
			few{3333}
			many{4444}
			other{5555}}',
```
So it will work for any language as stated in the aforementioned link.

aledeg:
Should I keep going with this?

Alkarex:
> Should I keep going with this?

For me this is a *nice to have* but not a high priority. Some people may be more sensitive, though. As long as it does not introduce too much new code or dependencies, then this is positive.

aledeg:
It requires intl only. Are we good if this becomes a requirement?
If not, then I'll close that PR.

Alkarex:
> It requires intl only. Are we good if this becomes a requirement?

We already have `php-intl`:

https://github.com/FreshRSS/FreshRSS/blob/be49726ebb700aca030004d367c029082cfc6427/Docker/Dockerfile#L12

Although without error if not available:

https://github.com/FreshRSS/FreshRSS/blob/be49726ebb700aca030004d367c029082cfc6427/lib/lib_rss.php#L149-L150

It would be nice to keep it recommended but optional if possible

aledeg:
If optional, it means we have to have 2 versions of the translation process and 2 versions of the translation string. That will become a nightmare to maintain.

As it is not an urgent task, I'll close it and we'll see later.

Alkarex:
Although nice to keep optional, if it is too complicated, I do not think requiring `php-intl` is a big problem, so if you get motivated, feel free to go ahead

PR #7957, opened 2025-09-12
Fix JS error when navigating to last article with keyboard

Inverle:
This error would print in the console if navigating to last article with <kbd>J</kbd> or <kbd>K</kbd> key:

image

---

To reproduce the bug:
<ol>
<li>Select unread + read view, while having all articles marked as read</li>
<li>Mark two as unread and go to unread only view</li>
<li>Navigate with either <kbd>J</kbd> or <kbd>K</kbd> until you go past the last article or before first article</li>
<li>See error in console and no navigation with <code>first_feed()</code> or <code>last_feed()</code></li>
</ol>

note: I'm not sure if the fix is what the expected behavior is supposed to be

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -592,7 +592,7 @@ function prev_feed() {
 			adjacent = feed;
 		}
 	}
-	if (found) {
+	if (found && adjacent) {
 		delayedClick(adjacent.querySelector('a.item-title'));
 	} else {
 		last_feed();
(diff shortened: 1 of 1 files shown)


Frenzie:
I'd think the expected behavior is the next category when applicable and otherwise nothing. (Which I think is broadly what already happens?)

Inverle:
> Which I think is broadly what already happens?

Does not due to the error as far as I can tell

Frenzie:
I meant in all other scenarios, hence why it should conform.

PR #7960, opened 2025-09-14
Minor explicit namespace fix for PHPUnit

Alkarex:


aledeg (approved):
Maybe you could also mark those classes as `final`

Alkarex:
> Maybe you could also mark those classes as `final`

Good idea. Done

PR #7962, opened 2025-09-14
Fix 7307 - Scroll after load

### LINKED ISSUE #7307: [Bug] Shortcut for my labels menu: scrolling behavior
> ... while the share menu seems to make the window move a bit further down so that we see every entry (at least the sole one I have), it would be good to be able to see all the entries on the labels menu too.
> 
> Image 

 _Originally posted by @Alwaysin in [#2087](https://github.com/FreshRSS/FreshRSS/issues/2087#issuecomment-2631308371)_

raspo:
Closes #7307 

- Wrap the body of `loadDynamicTags()` into a Promise
- inside `mylabels()`, wait for the promise to complete, then scroll
- when `loadDynamicTags()` is not being called, return a self-resolving promise

Note: most of the code-changes are due to re-indentation as as result of wrapping in `return new Promise` and `loadMenuPromise.then(() => {`

1. Click on an long article
2. press keyboard shortcut `l` to go to "My labels"
3. page should scroll down
4. the `My labels` popup should be fully visible

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested


Alkarex:
Thanks for the PR :-)
I think we will first release a minor version 1.27.1 (soon) before merging this PR

p/scripts/main.js
```diff
@@ -797,32 +798,37 @@ function mylabels(key) {
 		return;
 	}
 
+	let loadMenuPromise = new Promise(resolve => resolve(true));
```
Frenzie: Is that the same as
```suggestion
	let loadMenuPromise = Promise.resolve(true);
```

raspo: Updated

math-GH (approved):

Alkarex:
Sorry for the delay. I was not super found of introducing some Promise wrappers, so I have rewritten using async/await pattern: https://github.com/FreshRSS/FreshRSS/pull/7962/commits/00593c48d1551456c3c6d865d3dcb416f3ba74e9

Please double-check

PR #7965, opened 2025-09-15
Add namespace on i18n files

aledeg:
This includes classes and their tests.
This is to prepare for PSR-4.

See https://www.php-fig.org/psr/psr-4/

- add namespace on i18n classes
- add namespace on i18n test classes

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


cli/i18n/I18nCompletionValidator.php
```diff
@@ -1,8 +1,12 @@
 <?php
 declare(strict_types=1);
 
+namespace Cli\I18n;
```
Alkarex, line 4: It would probably be better to have `FreshRSS` at top level, no?

Alkarex:
While not opposing, I think it adds a lot of noise (+50 lines that do nothing just for this small section and poorer syntax like `\Exception`) when used at app level, so I am just not super in a hurry to roll out :-)
It is something else for libraries. So maybe it would be better to start with Minz.

ColonelMoutarde:
> While not opposing, I think it adds a lot of noise (+50 lines that do nothing just for this small section and poorer syntax like `\Exception`) when used at app level, so I am just not super in a hurry to roll out :-) It is something else for libraries. So maybe it would be better to start with Minz.

I love this PR, but the top namespace can be "FreshRss"

PR #7970, opened 2025-09-15
Clarify: Visibility hidden vs. archived

### LINKED ISSUE #7887: [Bug] Archived feeds are being updated automatically
### Describe the bug

When a feed is archived, I don't expect it to keep getting updated automatically (that also spams my logs unnecessarily) unless I change the visibility back to display it:

Image

Image

Archived feeds should be improved in some other ways too, but I'll make a PR for that later

### FreshRSS version

1.27.1-dev

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/7887
We have two concepts: how much a feed is shown or not (controlled by priority), and how often a feed is refreshed (or not, in which case it is archived).
This PR removes the wording *Archived* from the *visibility* parameter, since this is not what it does.


Inverle:
So is there currently a way to exclude a feed from being refreshed, including manual refreshes?

Inverle:
Another annoying bug is that you can't directly filter a hidden feed, it just shows a 404 page.

(while the statistics page somewhat works, but shows "All feeds" in the select menu)

Alkarex:
> So is there currently a way to exclude a feed from being refreshed, including manual refreshes?

Yes, with the *Mute* button:
image

The reason it keeps the refresh value is to make it easier to toggle some feeds on and off, but the UI could likely be improved.

Alkarex:
> Another annoying bug is that you can't directly filter a hidden feed, it just shows a 404 page.

Right, that is a bug that should be fixed

Inverle:
> Yes, with the *Mute* button

I've had no idea what it does before, but now I know ;)

Is this mentioned anywhere in the documentation?

Something like this would be helpful (tooltip):

image

Frenzie:
> I've had no idea what it does before, but now I know ;)

Yes, I mentioned *mumble, mumble* years ago that it sounds like it's refreshing and just not telling you but I guess I forgot to do something about it.

aledeg (approved):

Alkarex:
> > Another annoying bug is that you can't directly filter a hidden feed, it just shows a 404 page.
> 
> Right, that is a bug that should be fixed

@Inverle Fixed in https://github.com/FreshRSS/FreshRSS/pull/7972, tests welcome

PR #7971, opened 2025-09-15
Safer calling of `install.php`

Inverle:
To prevent `install.php` from being accessed in unintended places

### DIFF
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -19,6 +19,7 @@ class FreshRSS extends Minz_FrontController {
 	 * - Enable user extensions (need all the other initializations)
 	 */
 	public function init(): void {
+		define('FRESHRSS_INIT', true);
 		if (!isset($_SESSION)) {
 			Minz_Session::init('FreshRSS');
 		}
--- a/app/FreshRSS.php
+++ b/app/FreshRSS.php
@@ -19,7 +19,6 @@ class FreshRSS extends Minz_FrontController {
 	 * - Enable user extensions (need all the other initializations)
 	 */
 	public function init(): void {
-		define('FRESHRSS_INIT', true);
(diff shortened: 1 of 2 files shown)


app/FreshRSS.php
```diff
@@ -19,6 +19,7 @@ class FreshRSS extends Minz_FrontController {
 	 * - Enable user extensions (need all the other initializations)
 	 */
 	public function init(): void {
+		define('FRESHRSS_INIT', true);
```
Alkarex: Instead of this additional constant, would it be sufficient to test whether the session has already been started?

app/install.php
```diff
@@ -1,6 +1,11 @@
 <?php
 declare(strict_types=1);
 
+if (defined('FRESHRSS_INIT') || basename(is_string($_SERVER['SCRIPT_NAME']) ? $_SERVER['SCRIPT_NAME'] : '') !== 'index.php') {
```
```suggestion
if (isset($_SESSION) || basename(is_string($_SERVER['SCRIPT_NAME']) ? $_SERVER['SCRIPT_NAME'] : '') !== 'index.php') {
```

(not tested)

app/FreshRSS.php
Inverle: ~~Hm I thought `Minz_Session::init` was being called twice in some place, but I missed the `isset` check, because of that I didn't do that.~~

~~So yes, but I would have to test what the behavior is when the session isn't setup correctly.~~
~~probably is fine though since there is no `session_start()` before~~

The ideal solution would be to wrap everything inside `install.php` into a function and call the function after including `install.php` somewhere.

Or refactor `install.php` into a controller

Inverle:
Done

app/install.php
Inverle: Looks like it's sufficient

PR #7972, opened 2025-09-15
Add new visibility priority *Show in its feed*

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/pull/7970#issuecomment-3293917428 (you can't directly filter a hidden feed, it just shows a 404 page)
And add a new visibility *Show in its feed* to show the feed in the list but not its articles.
Ensure that visibility *hidden* is not shown to API (unchanged).


Inverle:
Seems to work as expected

app/i18n/pl/sub.php
```diff
@@ -210,6 +210,7 @@
 		'priority' => array(
 			'_' => 'Widoczność',
 			'category' => 'Pokaż w kategorii kanału',
+			'feed' => 'Show in its feed',	// TODO
```
Inverle:
```suggestion
			'feed' => 'Pokaż w kanale',
```

PR #7983, opened 2025-09-18
Fix SimplePie support of HTTP trailer headers

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/discussions/7981
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Trailer
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing
We need to use content-length to know where the body stops, but content-length is wrong if any compression was used.
So let cURL perform the separation of HTTP headers and body instead of using the SimplePie parser.


Alkarex:
Upstream issue:
* https://github.com/simplepie/simplepie/issues/942

PR #7993, opened 2025-09-22
Upgrade to PHP_CodeSniffer 4

Alkarex:
https://github.com/PHPCSStandards/PHP_CodeSniffer/wiki/Version-4.0-User-Upgrade-Guide
A couple of rules have changed name. Exclusion paths have slightly changed behaviour.
Main change is that it was unhappy with our i18n header comments, so minor edit there, since that rule kind of makes sense the rest of the time

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -6,7 +6,7 @@
 	<exclude-pattern>/data/config.php</exclude-pattern>
 	<exclude-pattern>/data/update.php</exclude-pattern>
 	<exclude-pattern>/data/users/*/config.php</exclude-pattern>
-	<exclude-pattern>/(?-i:extensions)/*</exclude-pattern><!-- Case insensitive -->
+	<exclude-pattern>/(?-i:extensions)/*</exclude-pattern><!-- Case sensitive to allow our Extensions repo -->
 	<exclude-pattern>/lib/http-conditional.php</exclude-pattern>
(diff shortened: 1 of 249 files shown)


phpcs.xml
Alkarex, line 1: The interesting configuration changes are here

phpcs.xml
```diff
@@ -2,24 +2,23 @@
...
-	<exclude-pattern>./lib/phpmailer/</exclude-pattern>
-	<exclude-pattern>./lib/simplepie/</exclude-pattern>
-	<exclude-pattern>./node_modules/</exclude-pattern>
-	<exclude-pattern>./p/scripts/vendor/</exclude-pattern>
-	<exclude-pattern>./vendor/</exclude-pattern>
+	<exclude-pattern>/\.git/*</exclude-pattern>
+	<exclude-pattern>/data/config.php</exclude-pattern>
+	<exclude-pattern>/data/update.php</exclude-pattern>
+	<exclude-pattern>/data/users/*/config.php</exclude-pattern>
+	<exclude-pattern>/(?-i:extensions)/*</exclude-pattern><!-- Case insensitive -->
```
Alkarex: This is because the same rules are also used for our `Extensions` repo, and if not case-sensitive, that rule would exclude the full repo due to its name, since it is apparently not possible to match using an absolute path from the root of the repo

aledeg (approved):

PR #7958, opened 2025-09-13
Move unsafe autologin to an extension

Inverle:
Completes the following TODO (#7923):

https://github.com/FreshRSS/FreshRSS/blob/de624dc8ce63ec819c61216d9d44f828841c293e/app/Controllers/authController.php#L105

Extension PR: https://github.com/FreshRSS/Extensions/pull/364

PR #7999, opened 2025-09-23
Prevent logout CSRFs

Inverle:
By avoiding `FreshRSS_Context::initUser()` calls

### DIFF
--- a/app/Controllers/authController.php
+++ b/app/Controllers/authController.php
@@ -70,7 +70,7 @@ class FreshRSS_auth_Controller extends FreshRSS_ActionController {
 	 * the user is already connected.
 	 */
 	public function loginAction(): void {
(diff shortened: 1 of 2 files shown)


app/Controllers/authController.php
```diff
@@ -70,7 +70,7 @@ public function indexAction(): void {
 	 * the user is already connected.
 	 */
 	public function loginAction(): void {
-		if (FreshRSS_Auth::hasAccess() && Minz_Request::paramString('u') === '') {
+		if (FreshRSS_Auth::hasAccess() && !(FreshRSS_Context::systemConf()->unsafe_autologin_enabled && Minz_Request::paramString('u') !== '')) {
```
Inverle, line 73: A better fix is available at #7958 - with a better extension implementation of the unsafe autologin feature as well

PR #7833, opened 2025-08-18
Update German

beerisgood:
German translation added for https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fde+%2F%28TODO%7CDIRTY%29%24%2F&type=code

PR #8028, opened 2025-09-28
Update German

scharmach:
@Anagastes @beerisgood @math-GH Thanks for all the work updating the German translation in https://github.com/FreshRSS/FreshRSS/pull/7833!

I've noticed these two things and wanted to further add upon it:

- The translation "Vorgestern" was technically not correct in `index.php`, in my opinion
- Also corrected a small typo in `sub.php` that I've noticed

### DIFF
--- a/app/i18n/de/sub.php
+++ b/app/i18n/de/sub.php
@@ -237,7 +237,7 @@ return array(
 		'ttl' => 'Aktualisiere automatisch nicht öfter als',
 		'unicityCriteria' => array(
 			'_' => 'Einzigartigkeit eines Artikels',
-			'forced' => '<span title="Einzigartikkeit-Einstellungen blockieren, selbst wenn der Feed Duplikat-Artikel hat">Erzwingen</span>',
+			'forced' => '<span title="Einzigartigkeit-Einstellungen blockieren, selbst wenn der Feed Duplikat-Artikel hat">Erzwingen</span>',
 			'help' => 'Relevant für defekte Feeds.<br />⚠️ Änderungen werden Duplikate erzeugen.',
 			'id' => 'Standard ID (Standardeinstellung)',
 			'link' => 'Link',	// IGNORE
--- a/app/i18n/de/index.php
+++ b/app/i18n/de/index.php
@@ -37,7 +37,7 @@ return array(
 	'feed' => array(
 		'empty' => 'Es gibt keinen Artikel zum Anzeigen.',
 		'received' => array(
-			'before_yesterday' => 'Vorgestern empfangen',
+			'before_yesterday' => 'Vor gestern empfangen',
 			'today' => 'Heute empfangen',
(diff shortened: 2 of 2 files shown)


Anagastes:
> @Anagastes @beerisgood @math-GH Thanks for all the work updating the German translation in #7833!
> 
> I've noticed these two things and wanted to further add upon it:
> 
>     * The translation "Vorgestern" was technically not correct in `index.php`, in my opinion
> 
>     * Also corrected a small typo in `sub.php` that I've noticed

The phrase "Vor gestern" does not exist in the German language.

See here.
https://de.langenscheidt.com/deutsch-englisch/vorgestern

So the word ‘Vorgestern’ is technically correct :)
I would stick with the existing translation.

app/i18n/de/sub.php
```diff
@@ -237,7 +237,7 @@
 		'ttl' => 'Aktualisiere automatisch nicht öfter als',
 		'unicityCriteria' => array(
 			'_' => 'Einzigartigkeit eines Artikels',
-			'forced' => '<span title="Einzigartikkeit-Einstellungen blockieren, selbst wenn der Feed Duplikat-Artikel hat">Erzwingen</span>',
+			'forced' => '<span title="Einzigartigkeit-Einstellungen blockieren, selbst wenn der Feed Duplikat-Artikel hat">Erzwingen</span>',
```
Alkarex, line 240: What about here @Anagastes ? Approved?

Anagastes: Ah, yea it is an typo fix :) 

tikk -> tigk

So its ok ^^

Frenzie:
@Anagastes
> The phrase "Vor gestern" does not exist in the German language.

The OP's point is that it's not the day before yesterday (vorgestern) but all days before yesterday (vor gestern). If there's a better way to phrase that please share. :-)

Anagastes:
> @Anagastes
> 
> > The phrase "Vor gestern" does not exist in the German language.
> 
> The OP's point is that it's not the day before yesterday (vorgestern) but all days before yesterday (vor gestern). If there's a better way to phrase that please share. :-)

ah ok, yeah. your right... 

Okay, any alternative would break the similarity...

In this context, that's perfectly fine.

scharmach:
> > @Anagastes
> > > The phrase "Vor gestern" does not exist in the German language.
> > 
> > 
> > The OP's point is that it's not the day before yesterday (vorgestern) but all days before yesterday (vor gestern). If there's a better way to phrase that please share. :-)
> 
> ah ok, yeah. your right...
> 
> Okay, any alternative would break the similarity...
> 
> In this context, that's perfectly fine.

I was also struggling to find a better wording. And anything I came up with broke consistency, indeed. "Älter als gestern" was my favourite without consistency, if I ever had one.

Thanks for approving now.

PR #8029, opened 2025-09-28
Create separate `Retry-After` files for proxies

Inverle:
Bad proxies are able to send a false `Retry-After` header and affect the availability of feeds (domain-wide) for other users.

This PR starts including the address of the proxy if present in filenames for `Retry-After` to mitigate the issue.


lib/simplepie/simplepie/src/File.php
```diff
@@ -26,6 +26,11 @@ class File implements Response
      */
     public $url;
 
+    /**
+     * @var array<mixed> Array of curl options in use
+     */
+    public $curl_options; // FreshRSS
```
Alkarex: Can you think of any alternative that would not require making a change to SimplePie?
If not, or not elegant enough, we will go for this approach

Inverle: Isn't a change to SimplePie required given there is no other way to access the set proxy in 

https://github.com/FreshRSS/FreshRSS/blob/ffac30cbdfff23415877623a56087e075696d954/app/Models/SimplePieResponse.php#L7

?

Inverle: Potentially it could be possible to use reflection(maybe?) but that wouldn't be an elegant solution

Alkarex: I have made a change to reduce the differences in the SimplePie library: https://github.com/FreshRSS/FreshRSS/pull/8029/commits/010f67dc416224d697c20145ea013569592d49c8
If good enough, please make a PR to our fork https://github.com/FreshRSS/simplepie

Inverle: https://github.com/FreshRSS/simplepie/pull/62

lib/lib_rss.php
```diff
@@ -735,7 +735,24 @@ function httpGet(string $url, string $cachePath, string $type = 'html', array $a
 		cleanCache(CLEANCACHE_HOURS);
 	}
 
-	if (($retryAfter = FreshRSS_http_Util::getRetryAfter($url)) > 0) {
+	$options = [];
+	$accept = '';
+	$proxy = is_string(FreshRSS_Context::systemConf()->curl_options[CURLOPT_PROXY] ?? null) ? FreshRSS_Context::systemConf()->curl_options[CURLOPT_PROXY] : '';
+	if (is_array($attributes['curl_params'] ?? null)) {
+		$options = sanitizeCurlParams($attributes['curl_params']);
+		$proxy = is_string($options[CURLOPT_PROXY]) ? $options[CURLOPT_PROXY] : '';
```
Alkarex, line 743: Fix https://github.com/FreshRSS/FreshRSS/pull/8218

PR #8033, opened 2025-09-29
Reverse arguments to array_merge of cusotm config

larsks:
The installer was merging the initial configuration into the custom
configuration, which meant that any keys set in the system configuration
would take precedence over custom configuration. Practically, this meant it
was not possible to preconfigure the database connection via
`config.custom.php`.

This commit reverses the order of the arguments to the `array_merge`
function so that keys in the custom configuration will override keys in the
initial configuration.

See also: https://github.com/FreshRSS/FreshRSS/discussions/8030


Alkarex:
Thanks 👍🏻 

Issue #7110, opened 2024-12-16
[Feature] Implement priority for hooks


pe1uca:
**Is your feature request related to a problem? Please describe.**
When [AutoTTL](https://github.com/mgnsk/FreshRSS-AutoTTL) runs, it assumes the feed has been updated in the event `feed_before_actualize`  
This could not be true when any other extension runs a hook in the same event and prevents a feed from being updated after AutoTTL has already processed the feed.  
In my set up it's [RateLimiter](https://github.com/pe1uca/xExtension-RateLimiter/).  

This is really bad on feeds which go into AutoTTL's max TTL since they don't get updated until that next max (which I have set for a day) and then once they can update again they most probably will be blocked again by the RateLimit while AutoTTL again thinks they got properly updated and won't let them be updated for another day.   

This could also be used by other hooks which want to preprocess the entities before other extensions.  
Maybe two extensions want to modify an entry with `entry_before_insert` or `entry_before_display`, but the changes of the extension X need to be present for extension Y to actually have an effect, if they run in reverse order Y's changes won't be applied.  

**Describe the solution you’d like**
Add  an optional parameter to the `Minz_Extension::registerHook` and `Minz_ExtensionManager::addHook` functions to give a numbered priority by the extension.  
This way the extension would call something like `$this->registerHook('feed_before_actualize', [$this,'feedBeforeActualizeHook'], 999);`, any any existing extension can just ignore the new parameter.  

**Describe alternatives you’ve considered**
Check how FreshRSS loads extensions and manipulate them to be loaded in certain order (this has the drawback of all hooks the extension uses having the same "priority)  

**Additional context**

Currently hooks are stored like this  
https://github.com/FreshRSS/FreshRSS/blob/6b14a743ccfe5c14e15acac858c0906f4fccdf67/lib/Minz/ExtensionManager.php#L332

I'm thinking this could be changed to `self::$hook_list[$hook_name]['list'][$priority][] = $hook_function;`, ensuring the order of the array is correct to be called by `foreach`.    
Then each call function would require to have a double foreach  
```php
foreach (self::$hook_list[$hook_name]['list'] as $prioBucket) {
	foreach ($prioBuckets as $function) {
		// Code to call the hooks
	}
}
```

Alkarex:
Alternatively, we could sort the `$hook_list` array based on priority (expressed as a new attribute of an extension `metadata.json`) and leave the rest of the code as it is.

PR welcome with either approach

[...]

PR #8036, opened 2025-09-30
Add hook enums

aledeg:

- add an enum to handle hook types (enum are available since PHP 8.1)
- change hook calls from string value to enum value

1. load extensions
2. validate they are working the same way

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -55,7 +55,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 		$url = trim($url);
 
 		/** @var string|null $urlHooked */
-		$urlHooked = Minz_ExtensionManager::callHook('check_url_before_add', $url);
+		$urlHooked = Minz_ExtensionManager::callHook(Minz_HookType::CheckUrlBeforeAdd, $url);
 		if ($urlHooked === null) {
 			throw new FreshRSS_FeedNotAdded_Exception($url);
 		}
@@ -106,7 +106,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 		}
 
 		/** @var FreshRSS_Feed|null $feed */
-		$feed = Minz_ExtensionManager::callHook('feed_before_insert', $feed);
+		$feed = Minz_ExtensionManager::callHook(Minz_HookType::FeedBeforeInsert, $feed);
 		if ($feed === null) {
 			throw new FreshRSS_FeedNotAdded_Exception($url);
 		}
@@ -465,7 +465,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 		$categoriesEntriesTitle = [];
 
 		foreach ($feeds as $feed) {
-			$feed = Minz_ExtensionManager::callHook('feed_before_actualize', $feed);
+			$feed = Minz_ExtensionManager::callHook(Minz_HookType::FeedBeforeActualize, $feed);
 			if (!($feed instanceof FreshRSS_Feed)) {
 				continue;
 			}
@@ -616,10 +616,10 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 							$entry->_isFavorite(null);	// Do not change favourite state
 							$entry->_isRead($mark_updated_article_unread ? false : null);	//Change is_read according to policy.
 							if ($mark_updated_article_unread) {
-								Minz_ExtensionManager::callHook('entry_auto_unread', $entry, 'updated_article');
+								Minz_ExtensionManager::callHook(Minz_HookType::EntryAutoUnread, $entry, 'updated_article');
 							}
 
-							$entry = Minz_ExtensionManager::callHook('entry_before_insert', $entry);
+							$entry = Minz_ExtensionManager::callHook(Minz_HookType::EntryBeforeInsert, $entry);
 							if (!($entry instanceof FreshRSS_Entry)) {
 								// An extension has returned a null value, there is nothing to insert.
 								continue;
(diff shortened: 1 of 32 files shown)


aledeg:
@Alkarex I think this could improve our code by removing manual validation of the hook type strings. Do you think it's worth continuing the work on that?

I was thinking about #7110 and how we could accommodate such changes and I though this might be the first step toward that.

What do you think?

app/Controllers/feedController.php
```diff
@@ -55,7 +55,7 @@ public static function addFeed(string $url, string $title = '', int $cat_id = 0,
 		$url = trim($url);
 
 		/** @var string|null $urlHooked */
-		$urlHooked = Minz_ExtensionManager::callHook('check_url_before_add', $url);
+		$urlHooked = Minz_ExtensionManager::callHook(Minz_HookType::CheckUrlBeforeAdd->value, $url);
```
```suggestion
		$urlHooked = Minz_ExtensionManager::callHook(Minz_HookType::CheckUrlBeforeAdd, $url);
```

Suggestion, and for all similar calls. Requires function signature update. See below

lib/Minz/ExtensionManager.php
```diff
@@ -372,11 +268,10 @@ public static function addHook(string $hook_name, $hook_function): void {
...
-	 * The hook name must be a valid one. For the valid list, see self::$hook_list
-	 * array keys.
+	 * The hook name must be a valid one. For the valid list, see Minz_HookType enum.
 	 *
 	 * @param string $hook_name the hook to call.
-	 * @param mixed ...$args additional parameters (for signature, please see self::$hook_list).
+	 * @param mixed ...$args additional parameters (for signature, please see Minz_HookType enum).
 	 * @return mixed|void|null final result of the called hook.
 	 */
 	public static function callHook(string $hook_name, ...$args) {
```
Alkarex:
```suggestion
	public static function callHook(Minz_HookType|string $hook_name, ...$args) {
		if ($hook_name instanceof Minz_HookType) {
			$hook_name = $hook_name->value;
		}
```

Alkarex:
I think that looks like a good step 👍🏻 
After that, I think we should offer a strong-typed version of the calls and deprecate the `callHook(string, ...mixed): mixed` approach.

app/Controllers/feedController.php
aledeg: That was my ultimate goal. But I did not want to spend too much time if this was not the way to follow.

aledeg:
Thank you for your input.

I'll continue in that direction and I'll polish this PR.

aledeg:
@Alkarex Whenever you have time!

lib/Minz/ExtensionManager.php
```diff
@@ -468,9 +396,14 @@ public static function callHookVoid(string $hook_name): void {
 	 * Call a hook which takes no argument and returns nothing.
 	 * Same as callHookVoid but only calls the first extension.
 	 *
-	 * @param string $hook_name is the hook to call.
+	 * @param string|Minz_HookType $hook is the hook to call.
 	 */
-	public static function callHookUnique(string $hook_name): bool {
+	public static function callHookUnique(string|Minz_HookType $hook): bool {
+		if (null === $hook = self::extractHook($hook)) {
+			throw new \RuntimeException("The \"{$hook}\" does not exist!");
```
Alkarex:
```suggestion
			throw new \RuntimeException("The hook “{$hook}” does not exist!");
```

Minor: typographic quotes

lib/Minz/HookType.php
```diff
@@ -0,0 +1,66 @@
+<?php
+declare(strict_types=1);
+
+enum Minz_HookType: string {
+	case ApiMisc = 'api_misc';	// function(): void
+	case BeforeLoginBtn = 'before_login_btn';	// function(): string
+	case CheckUrlBeforeAdd = 'check_url_before_add';	// function($url) -> Url | null
```
Alkarex:
```suggestion
	case CheckUrlBeforeAdd = 'check_url_before_add';	// function(string $url): string|null
```

Favour PHP native typing syntax when possible

lib/Minz/HookType.php
```diff
@@ -0,0 +1,66 @@
...
+				return Minz_HookSignature::OneToOne;
+			case Minz_HookType::CustomFaviconBtnUrl:
+			case Minz_HookType::CustomFaviconHash:
+			case Minz_HookType::EntriesFavorite:
+			case Minz_HookType::EntryAutoRead:
+			case Minz_HookType::EntryAutoUnread:
+			case Minz_HookType::SimplepieAfterInit:
+			case Minz_HookType::SimplepieBeforeInit:
+				return Minz_HookSignature::PassArguments;
+		}
```
Alkarex, line 66: Please consider adding a default case, otherwise not all paths return a value.
Minor: Maybe the `match` structure could be more compact

Alkarex (changes requested):

lib/Minz/ExtensionManager.php
aledeg: done

lib/Minz/HookType.php
aledeg: done

lib/Minz/HookType.php
aledeg: Done.

I started with the `match` structure but it became less readable. I switched to `switch` to spread the values.

Alkarex (approved):

aledeg:
@Alkarex Should I sign my commits? This will rewrite the history though.

Alkarex:
> Should I sign my commits? This will rewrite the history though.

No need for this one, but would be nice for future ones

PR #8052, opened 2025-10-01
PHPStan 2.1.29

Alkarex:
Replace:
* https://github.com/FreshRSS/FreshRSS/pull/8049
* https://github.com/FreshRSS/FreshRSS/pull/8050

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -1017,7 +1017,7 @@ HTML;
 				return false;
 			}
 			$xpath = new DOMXPath($doc);
-			$filterednodes = $xpath->query((new Gt\CssXPath\Translator($feed->attributeString('path_entries_filter') ?? '', '//'))->asXPath()) ?: [];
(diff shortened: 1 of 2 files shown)


app/Models/Entry.php
```diff
@@ -1017,7 +1017,7 @@ public function loadCompleteContent(bool $force = false): bool {
 				return false;
 			}
 			$xpath = new DOMXPath($doc);
-			$filterednodes = $xpath->query((new Gt\CssXPath\Translator($feed->attributeString('path_entries_filter') ?? '', '//'))->asXPath()) ?: [];
+			$filterednodes = $xpath->query((new Gt\CssXPath\Translator($feed->attributeString('path_entries_filter'), '//'))->asXPath()) ?: [];
```
Alkarex, line 1020: PHPStan has become a bit smarter: this condition is already tested a bit higher up

PR #7913, opened 2025-09-04
Enhance navigation bar

aledeg:
Now, it includes a button to open the current article in a new window and a button to toggle the current article favourite flag.

See #7912

Closes #7912

Changes proposed in this pull request:

- Add 2 buttons in the navbar to ease mobile navigation

How to test the feature manually:

1. use the 2 new buttons and validate that they behave properly

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

# The new navbar
Screen Shot 2025-09-04 at 08 26 15

Issue #7912, opened 2025-09-03
[Feature] Add Fav-Button to Bottom-Navbar on mobile view


SHU-red:
### Is your feature request related to a problem? Please describe.

- Due to the fact that FreshRSS is very configurabe im using no 3rd party app on my phone
- Often i iterate through the main feed by using the bottom arrow buttons to go through every article
- Often im NOT reading the articles, but just add the ones where the headline seems to be interesing to the favourites
- When ive got enough time im reading all articles in my favs
- But to add an article to my favs i have to reach out to the top of the screen and try to hit the very tiny start-button

### Use case

- The same convenience to step through the articles with the arrow-buttons on the bottom-navbar, i would love to have the possiblitity to also have a fav button on the bottom of the screen

### Describe the solution you’d like

- Add option in settings to add fav-/star-button to navbar on small/mobile screens

### Describe alternatives you’ve considered

- 3rd party app
- reach out to the fav button on top of the screen for every interesting article

aledeg:
This! And also a button to open the original article.

math-GH:
You can add the favorite button to the bottom line 

<img width="875" height="142" alt="Image" src="https://github.com/user-attachments/assets/d929c79c-75e0-4686-a68f-9fa267ddc763" />

IMHO: To have this bottom line fixed at the bottom while scrolling within the article would be nice.

aledeg:
The thing is the star is still tiny and you still need to scroll up or down depending of the length of the article. I do have that in my configuration but it feels like a chore every time I am browsing on mobile.

I think we could have other buttons in the navigation buttons. We could also implement gestures in that area on top of buttons.

math-GH:
Could you please try this User CSS?
It makes the bottom line sticky while scrolling.

```
.flux_content > footer {
  position: sticky;
  bottom: 60px;
}
```

(known "bug": the line does not have a background color yet so the content shines through)

SHU-red:
> You can add the favorite button to the bottom line
[...]

PR #8054, opened 2025-10-02
Add a new hook in the UI

aledeg:
The new hook allows extension to add their own tool bar to navigate between entries. For instance, if the user wants less or more buttons that what's available by default.

See #7912
See #7913

1. create an extension using the new hook
2. validate that the hook output is displayed in the DOM right before the navigation buttons

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


Alkarex:
Maybe something to add to the reader view as well?

aledeg:
Where do you think we should add that?

Alkarex:
Around there, I would think:

https://github.com/FreshRSS/FreshRSS/blob/128c375fc9de5119b1963d92dc9cd3e423053111/app/views/helpers/index/article.phtml#L49-L50

app/views/index/normal.phtml
```diff
@@ -87,6 +87,7 @@ $today = @strtotime('today');
 		?>" data-entry="<?= $this->entry->id()
 		?>" data-feed="<?= $this->feed->id()
 		?>" data-priority="<?= $this->feed->priority()
+		?>" data-link="<?= $this->entry->link()
```
Inverle, line 90: https://github.com/FreshRSS/FreshRSS/pull/8397#discussion_r2659220799

PR #8056, opened 2025-10-02
Add default parameter on int parameters

aledeg:
Before, the method had a default value of 0 when retrieving an int parameter on the request. Which is not always what we want.
Now, the default value can be overridden when needed. If no value is provided, the default value is still 0.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -124,11 +124,11 @@ class Minz_Request {
 		return is_numeric(self::$params[$key] ?? null) ? (int)self::$params[$key] : null;
 	}
 
-	public static function paramInt(string $key): int {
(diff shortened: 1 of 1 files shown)


Alkarex:
For those use-cases, the idea was to use a construct like:

https://github.com/FreshRSS/FreshRSS/blob/032316155c03ca1013644476b8f9413ba436b744/app/Controllers/configureController.php#L554


Could that work for you?

Similarly for other types, such as:

https://github.com/FreshRSS/FreshRSS/blob/032316155c03ca1013644476b8f9413ba436b744/app/Controllers/configureController.php#L135

aledeg:
Sure.

PR #8057, opened 2025-10-02
Fix navigating between read feeds using shortcut shift+j/k

Inverle:
Before, the shortcuts marked in red and blue behaved the same way (with shift modifier)
Specifically <kbd>Shift + J</kbd> had the same behavior as <kbd>Shift + K</kbd> which means it only jumped to unread and not read/unread.

image

Now the shift modifier shortcuts match the behavior of the alt and no modifier shortcuts.

<kbd>Shift + K</kbd> was corrected too, but there is no alternative for it yet, since I don't know which  default key should be used for it - <kbd>l</kbd> is already taken by *My labels* shortcut

suggestions welcome

TODO:
* [ ] auto scroll to filtered feed on aside (done by https://github.com/FreshRSS/FreshRSS/pull/8281)

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -569,7 +569,7 @@ function next_unread_entry(skipping) {
 	toggleContent(new_active, old_active, skipping);
 }
 
-function prev_feed() {
+function prev_feed(jump_to_unread) {
 	let found = false;
 	let adjacent = null;
 	const feeds = document.querySelectorAll('#aside_feed .feed');
@@ -585,7 +585,7 @@ function prev_feed() {
 		if (getComputedStyle(feed).display === 'none') {
 			continue;
 		}
-		if (feed.dataset.unread != 0) {
+		if (jump_to_unread && feed.dataset.unread != 0) {
 			delayedClick(feed.querySelector('a.item-title'));
 			return;
 		} else if (adjacent === null) {
@@ -599,7 +599,7 @@ function prev_feed() {
 	}
 }
 
-function next_feed() {
+function next_feed(jump_to_unread) {
 	let found = false;
 	let adjacent = null;
 	const feeds = document.querySelectorAll('#aside_feed .feed');
@@ -615,7 +615,7 @@ function next_feed() {
 		if (getComputedStyle(feed).display === 'none') {
 			continue;
 		}
-		if (feed.dataset.unread != 0) {
+		if (jump_to_unread && feed.dataset.unread != 0) {
 			delayedClick(feed.querySelector('a.item-title'));
 			return;
 		} else if (adjacent === null) {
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -569,7 +569,7 @@ function next_unread_entry(skipping) {
 	toggleContent(new_active, old_active, skipping);
 }
 
-function prev_feed() {
+function prev_feed(jump_to_unread) {
```
Inverle, line 572: > <kbd>Shift + K</kbd> was corrected too, but there is no alternative for it yet, since I don't know which  default key should be used for it - <kbd>l</kbd> is already taken by *My labels* shortcut

Only `false` is being passed here so far.

Alkarex:
Ping @Frenzie

Frenzie:
Could you describe the behavior in more detail? I don't very well understand what the change is but it could just be that the change doesn't make any sense to me.

The use case for j/k not skipping entries is either that you purposefully showed both read and unread, or going back to an entry you just marked as read.

But if you're in unread view, shift+j and shift+k should go to the next feed with unread articles, not first past twenty random invisible feeds that are all read. Perhaps that's different in mixed unread/read view.

Inverle:
> purposefully showed both read and unread, 

That's what I normally use

image

---

> The use case for j/k not skipping entries is either that you purposefully showed both read and unread, or going back to an entry you just marked as read.

What if I want to go back to some feed I read before like 2 weeks ago?
(or any feed that doesn't have unreads at any point)

> But if you're in unread view, shift+j and shift+k should go to the next feed with unread articles, not first past twenty random invisible feeds that are all read. Perhaps that's different in mixed unread/read view.

It does that.

I'll try to record a GIF maybe showing the difference
To be clear, this is fixing a bug with the shortcuts not doing exactly what's described.

Frenzie:
> What if I want to go back to some feed I read before like 2 weeks ago?
(or any feed that doesn't have unreads at any point)

The way I see it J is normal navigation. K is exceptional. What's normal and what's exceptional is situational and could be contradictory on a different level.

> It does that.
> 
> I'll try to record a GIF maybe showing the difference

Indeed, I just tried it in a few scenarios and I couldn't even spot any difference.

Alkarex:
Could one of you help with a status here?

Frenzie:
If the effect is that the behavior matches up with that which is displayed in the feed list on the left, it should be fine.

Alkarex:
Probably related:
* https://github.com/FreshRSS/FreshRSS/issues/8426

p/scripts/main.js
Alkarex, line 561: * https://github.com/FreshRSS/FreshRSS/pull/8466

PR #8061, opened 2025-10-03
Strengthen some crypto

Alkarex:
For login, tokens, nonces

### DIFF
--- a/app/Models/Auth.php
+++ b/app/Models/Auth.php
@@ -217,8 +217,7 @@ class FreshRSS_Auth {
 	public static function csrfToken(): string {
 		$csrf = Minz_Session::paramString('csrf');
 		if ($csrf == '') {
-			$salt = FreshRSS_Context::systemConf()->salt;
-			$csrf = sha1($salt . uniqid('' . random_int(0, mt_getrandmax()), true));
+			$csrf = hash('sha256', FreshRSS_Context::systemConf()->salt . random_bytes(32));
 			Minz_Session::_param('csrf', $csrf);
 		}
 		return $csrf;
--- a/app/Controllers/javascriptController.php
+++ b/app/Controllers/javascriptController.php
@@ -74,12 +74,11 @@ class FreshRSS_javascript_Controller extends FreshRSS_ActionController {
 		$user_conf = get_user_configuration($user);
 		if ($user_conf !== null) {
 			try {
-				$salt = FreshRSS_Context::systemConf()->salt;
 				$s = $user_conf->passwordHash;
 				if (strlen($s) >= 60) {
 					//CRYPT_BLOWFISH Salt: "$2a$", a two digit cost parameter, "$", and 22 characters from the alphabet "./0-9A-Za-z".
 					$this->view->salt1 = substr($s, 0, 29);
-					$this->view->nonce = sha1($salt . uniqid('' . mt_rand(), true));
+					$this->view->nonce = hash('sha256', FreshRSS_Context::systemConf()->salt . $user . random_bytes(32));
 					Minz_Session::_param('nonce', $this->view->nonce);
 					return;	//Success
 				}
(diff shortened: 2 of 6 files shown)


Frenzie (approved):
Sounds good :+1:

app/Models/Auth.php
```diff
@@ -217,8 +217,7 @@ public static function accessNeedsAction(): bool {
 	public static function csrfToken(): string {
 		$csrf = Minz_Session::paramString('csrf');
 		if ($csrf == '') {
-			$salt = FreshRSS_Context::systemConf()->salt;
-			$csrf = sha1($salt . uniqid('' . random_int(0, mt_getrandmax()), true));
+			$csrf = hash('sha256', FreshRSS_Context::systemConf()->salt . random_bytes(32));
```
Alkarex, line 220: I did a micro-benchmark here and this is about twice as fast as before on my platform

Inverle, line 220: What about
```suggestion
			$csrf = bin2hex(random_bytes(20));
```
?

Alkarex: I like the idea of keeping our salt

Inverle:
https://pentesterlab.com/blog/freshrss-bcrypt-truncation-auth-bypass

Some good points:
> Over-engineering can hurt security. The challenge-response protocol with client-side bcrypt is already unusual. Most web applications just send the password over HTTPS and hash server-side. The added complexity created a subtle interaction between nonce length and bcrypt truncation that a simpler design would not have. More layers of crypto does not automatically mean more security.

> Test authentication. The FreshRSS test suite has no tests for the authentication flow at all. The only password-related test checks validation rules (minimum length, non-empty), not actual credential verification. The complexity of the challenge-response protocol likely made it harder to test than a simple password check, but that is exactly the kind of code that needs tests the most. A single test case (log in with the wrong password and assert failure) would have caught this immediately.

Alkarex:
Interesting article, but I strongly disagree about the headline *It is a good example of how over-engineering can hurt the security of an application*.

First, my fault is to have addressed https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-j9wc-gwc6-p786 instead of putting it to our backlog. I was fully aware that any change to the security part is a risk. So when there is a mostly theoretical security advisory submitted to us, it is not worth prioritising it. But I balance that with aspects such as being welcoming, engaging to community members, and I would probably have ended up spending more time explaining why it should be queued for later (e.g. after having some tests!) than making the patch, and it would have led to frustrations. (Now I spend time writing about some noise, and got some frustration, and got a security issue, so lose-lose-lose situation...)

Back to why I think the headline (and the section further down) is misleading: I do not believe it is wise for the article to (indirectly) suggest that it would have been better to send the credentials in plaintext. This is a misunderstanding of several things:
* This method predates offerings such as Let's Encrypt, so when it was developed, it was precisely because the majority of instances were without HTTPS due to price and complexity.
* While it is true that HTTPS is now much more prevalent, even today, many instances are (still) running without HTTPS.
* HTTPS is not common for local networks (such as 192.168.x.x) and if you look at messages from various users even recently, this is a very common use-case.

Side note: Traditional journalistic methodology (before the click‑bait era) recommends communicating with people familiar with a case before publishing conclusions, e.g. to avoid misunderstandings.

Regarding tests, this is stating the obvious, but is missing the point:
* Almost nobody contributes with PR for improving tests. The (sad?) reality is that there are significantly more people sending (often barely relevant) security advisories and writing blog posts than contributing with the corresponding patches, let alone sending test PRs proactively. I cannot blame anyone, because this is not super rewarding, but when there is time for the first, it would be nice to take the time for the second as well.
* Testing security aspects such as our different login methods is *not* as easy as it may sound
* There are plenty of aspects for which unit tests would be very relevant so it is a matter of priorities, and it is typically after the fact that it becomes apparent which easy test could have prevented a regression. But even after some publicity, it is not like we are getting more patches...
* So, famous words: PR welcome :-)

Frenzie:
Yes, I didn't like that phrase much either and I also disagreed that the concept was unusual.

PR #8067, opened 2025-10-04
Do not mark important feeds as read from category

Alkarex:
Important feeds are not supposed to be marked as read from another view. This was already working for the main view, but this PR fixes the case of a category view.
Follow-up of:
* https://github.com/FreshRSS/FreshRSS/pull/5782

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -571,9 +571,9 @@ SQL;
 UPDATE `_entry`
 SET is_read = ?
 WHERE is_read <> ? AND id <= ?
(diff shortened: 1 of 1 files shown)


Inverle:
So I have `Mark an article as read: when the article is viewed` enabled, and whenever I click on a important article in any view, it's still marked as read?

Alkarex:
> So I have `Mark an article as read: when the article is viewed` enabled, and whenever I click on a important article in any view, it's still marked as read?

Yes, that is intentional. Important articles should still be able to be marked as read individually from anywhere

Inverle:
Oh this is about mark all as read in a category

PR #8081, opened 2025-10-10
Add support for extension compatibility

### LINKED ISSUE #5903: Add compatibility range for extensions
I don't think there is a way to mark an extension as only compatible with version <= x => ?

_Originally posted by @nicofrand in https://github.com/FreshRSS/FreshRSS/issues/5537#issuecomment-1640256145_

aledeg:
The compatibility does support only a minimum version of FreshRSS. If we need
something a bit more clever in the future, it is possible to handle a rule
with a bit more complexity.

See #5903

-
-
-

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


aledeg:
@Alkarex This is a first draft for the support of extension compatibility. Is there something else to add for a first version or we'll start with that?

I'll work on the extension list generation after this is completed.

app/Controllers/extensionController.php
```diff
@@ -40,7 +42,7 @@ public function indexAction(): void {
 
 	/**
 	 * Fetch extension list from GitHub
-	 * @return list<array{name:string,author:string,description:string,version:string,entrypoint:string,type:'system'|'user',url:string,method:string,directory:string}>
+	 * @return list<ExtensionFullMetadata>
```
Alkarex:
```suggestion
	 * @phpstan-return list<ExtensionFullMetadata>
```

When it is something only supported by PHPStan

app/i18n/pl/admin.php
```diff
@@ -118,6 +118,7 @@
 		'empty_list' => 'Brak zainstalowanych rozszerzeń',
 		'empty_list_help' => 'Sprawdź dziennik, aby ustalić powód pustej listy rozszerzeń.',
 		'enabled' => 'Włączone',
+		'is_compatible' => 'Is compatible',	// TODO
```
Inverle:
```suggestion
		'is_compatible' => 'Jest kompatybilne',
```

Inverle:
There has to be a `white-space: nowrap` CSS rule on the `<th>` element:

image

aledeg:
> There has to be a `white-space: nowrap` CSS rule on the `<th>` element:
> image

What would be the css selector to apply that rule. I don't think I can apply it on `th` directly, the impact would be to important.
Do you have any idea?

Inverle:
`div.table-wrapper th`?

Frenzie:
You could always just add a class to that column (though I don't necessarily know if I agree with the comment :p)

Alkarex:
Sorry @aledeg for using your PR as a test for implementing a new GitHub Action:
* https://github.com/FreshRSS/FreshRSS/pull/8098

I had to try a few times to get it to work...

aledeg:
@Alkarex I think there is a rounding error when computing the completeness of translation. Maybe not an error but an incertainty.

Alkarex:
> @Alkarex I think there is a rounding error when computing the completeness of translation. Maybe not an error but an incertainty.

When new strings are added as is the case in this PR, then it is expected that the percentage will change

aledeg:
> > @Alkarex I think there is a rounding error when computing the completeness of translation. Maybe not an error but an incertainty.
> 
> When new strings are added as is the case in this PR, then it is expected that the percentage will change

I mean that I ran the command `make readme` multiple times on my machine to fix that but there was not change detected. But on the CI/CD, changes were detected.

Alkarex:
Try `make fix-all` (I have not tried the other ones in a long time, except `make test-all`)

Alkarex:
> I mean that I ran the command `make readme` multiple times on my machine to fix that but there was not change detected. But on the CI/CD, changes were detected.

Ah, but I cannot remember what the status was, but this might also be due to the fact that when updating the branch, it may import new strings that have recently  been merged

aledeg:
I've messed up. I've needed to make a rebase.

PR #8099, opened 2025-10-12
some testing

Inverle:
ignore this PR

PR #8100, opened 2025-10-12
Fix critical GHA vulnerability

Inverle:
See https://github.com/FreshRSS/FreshRSS/pull/8099

The commands action is currently disabled, needs some more improvement.
We need to make sure that the code within the PR cannot access `GITHUB_TOKEN`.

I am not happy with having to grant `contents: write` permission so that commits can be added into the current PR, so some workaround is needed if possible. Otherwise it is safer to restrict permissions to just owner/member, and check PR contents before running `/fix-all`.

Current zizmor output:
```
info[template-injection]: code injection via template expansion
   --> ./commands.yml:140:37
    |
136 |         uses: actions/github-script@v8
    |         ------------------------------ action accepts arbitrary code
...
139 |           script: |
    |           ------ via this input
140 |             const makeFailed = '${{ steps.fix.outputs.make_failed }}' === 'true';
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → Low

info[template-injection]: code injection via template expansion
   --> ./commands.yml:141:36
    |
136 |         uses: actions/github-script@v8
    |         ------------------------------ action accepts arbitrary code
...
139 |           script: |
    |           ------ via this input
140 |             const makeFailed = '${{ steps.fix.outputs.make_failed }}' === 'true';
141 |             const noChanges = '${{ steps.commit.outputs.no_changes || 'false' }}' === 'true';
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → Low

info[template-injection]: code injection via template expansion
   --> ./commands.yml:142:33
    |
136 |         uses: actions/github-script@v8
    |         ------------------------------ action accepts arbitrary code
...
139 |           script: |
    |           ------ via this input
...
142 |             const pushed = '${{ steps.commit.outputs.pushed || 'false' }}' === 'true';
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → Low

info[template-injection]: code injection via template expansion
   --> ./commands.yml:143:36
    |
136 |         uses: actions/github-script@v8
    |         ------------------------------ action accepts arbitrary code
...
139 |           script: |
    |           ------ via this input
...
143 |             const commitSha = '${{ steps.commit.outputs.commit_sha }}';
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → Low

info[template-injection]: code injection via template expansion
   --> ./commands.yml:144:36
    |
136 |         uses: actions/github-script@v8
    |         ------------------------------ action accepts arbitrary code
...
139 |           script: |
    |           ------ via this input
...
144 |             const jobStatus = '${{ job.status }}';
    |                                    ^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → Low

15 findings (10 suppressed): 5 informational, 0 low, 0 medium, 0 high
```

### DIFF
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -16,8 +16,7 @@ jobs:
       github.event.comment.body == '/fix-all' &&
       (
         github.event.comment.author_association == 'OWNER' ||
-        github.event.comment.author_association == 'MEMBER' ||
-        (github.event.comment.user != null && github.event.comment.user.login == github.event.issue.user.login)
+        github.event.comment.author_association == 'MEMBER'
       )
     runs-on: ubuntu-latest
 
@@ -63,6 +62,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           repository: ${{ fromJSON(steps.pr.outputs.result).repo.full_name }}
           ref: ${{ fromJSON(steps.pr.outputs.result).ref }}
+          persist-credentials: false
 
       - name: Use Composer cache
         id: composer-cache


Inverle:
This is still potentially unsafe.
Untrusted code had already run, is it possible to access the `GITHUB_TOKEN` from memory / environment variables from a background process that spawned earlier? is it possible to just make `git` run something else?:

https://github.com/FreshRSS/FreshRSS/blob/b7337153596f1282128e1d8b827f884d87cb0f7a/.github/workflows/commands.yml#L98-L106

Also no idea if this is safe either:

https://github.com/FreshRSS/FreshRSS/blob/b7337153596f1282128e1d8b827f884d87cb0f7a/.github/workflows/commands.yml#L35-L44

(and other `actions/github-script@v8` runs, is it a separate environment?)

Inverle:
> Also no idea if this is safe either:

https://docs.github.com/en/actions/reference/security/secure-use#use-an-action-instead-of-an-inline-script

maybe?

But then this is concerning:

https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/run-jobs-in-a-container#:~:text=If%20you%20do%20not%20set%20a%20container%2C%20all%20steps%20will%20run%20directly%20on%20the%20host%20specified%20by%20runs%2Don%20unless%20a%20step%20refers%20to%20an%20action%20configured%20to%20run%20in%20a%20container.

Frenzie:
> (and other actions/github-script@v8 runs, is it a separate environment?)

I don't believe they can access environment variables from another step.

There isn't some kind of "it can push to the PR but *only* the PR permission"? Or just make the bot post 'hey, please run make fix-all' instead of doing it unsupervised?

Alkarex:
I have made another attempt in https://github.com/FreshRSS/FreshRSS/pull/8106
Keeping write permissions only for `issues` and using GitHub API instead of git.
Not tested at all yet.

Inverle:
> I don't believe they can access environment variables from another step.

All steps run on the same host in the context of the `runner` user as a different process, which means they can access each others environment variables via `/proc/[pid]/environ`

> There isn't some kind of "it can push to the PR but only the PR permission"?

Unfortunately there isn't

Frenzie:
> All steps run on the same host in the context of the runner user as a different process, which means they can access each others environment variables via /proc/[pid]/environ

I'd find that quite surprising. You mean this is what you found after testing? Of course it should be tested and not assumed, but my expectations would be:

1. there is no pid in the first place because it already finished
2. permission denied

Inverle:
replaced by https://github.com/FreshRSS/FreshRSS/pull/8123

PR #8107, opened 2025-10-13
Track plain text file changes as feed kind, diffing against previous content

nykula:
**Note:** Draft, seems to work but the files I need to watch don't update very often so I still have to test it more.

**Goal:** Notify about updates to translatable files in free software using different version control systems and hosting platforms (SourceForge SVN, Google Sheets). There are specialized tools that scrape websites using headless browsers, make screenshots etc, but my need is simpler and mostly matches what FreshRSS is already doing.

Ideally the tool would store a diff output against the most recently existing item, not the entire file contents, but I don't know how to implement that.

- add `KIND_PLAIN_TEXT` and `TYPE_PLAIN_TEXT` constants
- implement `loadPlainText` function in Feed model, call it from feed controller
- add `text` mimetype option to `httpGet`

1. Host a plain text file somewhere
2. Subscribe to a new feed using `Plain text (entire file)` type and file URL
3. Replace feed title and description with something short describing the file
4. Edit the file
5. Get a notification about the edit during next feeds update

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated


Alkarex:
Hello and thanks for this PR.
Quick feedback already, and I will come back to it later:
* It is not a new type of feed, but an attribute, similar to what we do with feeds requiring special cURL options, etc.
* So instead of a new feed type, we could expose a checkbox when adding a feed, to allow wrong Mime types.
* The information is already saved at the moment as an anchor `#force_feed` at the end of an URL when adding a feed, but updating to a checkbox (saved as an attribute) would probably be a nice improvement

nykula:
Hmm, I initially thought of reusing the existing HTML+Xpath scraping function and adding a checkbox to turn off DOM parsing, but that was a more invasive change and decoupling the two functions seemed to make sense to me. The idea is to create and expose the simplest RSS feed possible (using FreshRSS capabilities) from a non-RSS source, which isn't as much about the mime type, but about it being just a text, config or code file.

Frenzie:
I suspect there's a misunderstanding here. XML feeds that are sometimes sent out as `text/plain`.

But what if these plain text files are really, really big?

nykula:
What would be an appropriate `CURLOPT_MAXFILESIZE` in that case, and would it help? The files I want to watch are indeed of varied size and not very small, some are just 5 KB, but the largest one so far is 400 KB.

Frenzie:
My question was merely what would happen. :-) If the answer is unlimited I'd say limit it to whatever the code has decided elsewhere.

Alkarex:
> I suspect there's a misunderstanding here. XML feeds that are sometimes sent out as `text/plain`.

@Frenzie Indeed, that is what I thought it was about.

@nykula Would you have an example of use-case including example of corresponding text document?

nykula:
Sure. I monitor the following files:
 
Faircamp [en.rs](https://codeberg.org/simonrepp/faircamp/raw/branch/main/translations/src/en.rs) at Codeberg. The host provides an RSS feed of all commits but not one of changes to a specific file, as far as I know.

Ghost Commander [strings.xml](https://sourceforge.net/p/ghostcommander/svn/HEAD/tree/src/main/res/values/strings.xml?format=raw) at SourceForge. The host also seems to provide an RSS feed of all commits without more precision.

Helio translations at Google Sheets. The host offers no RSS feeds, so I monitor its [CSV export](https://docs.google.com/spreadsheets/d/1RBDyVUDZcx6Y_OZzicnfwaWSI5uAcPbHgByo6xiFhBQ/export?format=csv&id=1RBDyVUDZcx6Y_OZzicnfwaWSI5uAcPbHgByo6xiFhBQ&gid=0).

(would like to, haven't figured out a setup yet) LineageOS at Crowdin. The host used to offer an RSS feed but removed it with a promise of a JSON API and didn't implement it.

I learned yesterday about a utility called [urlwatch](https://thp.io/2008/urlwatch/), which is probably a cleaner way to observe files than patching FreshRSS, but I rely on a shared server and would rather it didn't keep my SMTP credentials.

nykula:
Added a naive diff using php array functions. It gets stored on the entry as a `diff` attribute, which feed views show instead of the full content when available. This addresses the concern about big files, and lets me see which new lines are there to translate without comparing the files manually.

The diff runs against the latest existing article in the same feed that doesn't have the same guid (content hash), for this I added a `latestExceptGuid` function to the EntryDAO model. If there's nothing yet to diff against, the full text is returned to the views.

Since the PHP standard library doesn't provide a more sophisticated text diffing algorithm like Python's difflib, I want to see the insertions first and the deletions later, so the diff in the resulting feed looks like this (is it acceptable to hardcode Unicode emoji in the model?):

```
➕
'New line for translation' => '',
'Another new line' => '',
'Etc' => '',

➖
'Removed line' => 'Translation that is now unused',
```

Alkarex:
> is it acceptable to hardcode Unicode emoji

Yes

nykula:
Hello, this patch has been working as expected for me for the last few weeks, so I rebased against edge, fixed all linter complaints and unmarked its draft status.

Alkarex:
Thanks, I will try to review shortly. This will target version 1.29.0 and not the upcoming minor 1.28.1 that focusses on bug fixes and which will be released ASAP.

> rebased against edge

Side note: I prefer to just merge edge (or just click the *Update branch* button, which does it well), because rebasing or any other form of git history rewriting break many things: comments, reviews, changes since last review, co-authored-by, permalinks, local branches, etc.

image

We *Squash and merge* so no need to manually squash multiple commits :-)

app/Models/EntryDAO.php
```diff
@@ -835,6 +835,21 @@ public function searchByGuid(int $id_feed, string $guid): ?FreshRSS_Entry {
 		return isset($res[0]) ? FreshRSS_Entry::fromArray($res[0]) : null;
 	}
 
+	public function latestExceptGuid(int $id_feed, string $guid): ?FreshRSS_Entry {
+		$content = static::isCompressed() ? 'UNCOMPRESS(content_bin) AS content' : 'content';
+		$hash = static::sqlHexEncode('hash');
+		$sql = <<<SQL
+SELECT id, guid, title, author, {$content}, link, date, `lastSeen`, `lastUserModified`, {$hash} AS hash, is_read, is_favorite, id_feed, tags, attributes
+FROM `_entry` WHERE id_feed=:id_feed AND guid!=:guid
+ORDER BY date DESC
```
Alkarex:
```suggestion
ORDER BY id DESC
```

We are probably interested in the latest received (date can potentially be anything)

app/Models/Feed.php
```diff
@@ -1150,6 +1151,65 @@ public function loadHtmlXpath(): ?FreshRSS_SimplePieCustom {
...
+				if (!empty($insertions)) {
+					$diff .= '➕<pre class="text-plain">' . htmlspecialchars(implode("\n", $insertions)) . '</pre>';
+				}
+				if (!empty($deletions)) {
+					$diff .= '➖<pre class="text-plain">' . htmlspecialchars(implode("\n", $deletions)) . '</pre>';
+				}
+			}
+		}
+		if ($diff !== '') {
+			$attributesByGuid[$item['guid']] = ['diff' => $diff];
```
Alkarex, line 1262: I have not considered it in depth, but instead of saving the diff in an attribute, could it be an idea to just use the normal article content and adding `<ins>` and `<del>` tags, which can be further styled by CSS, and during the next diff, the `<del>` tags can just be stripped before the comparison.
* https://developer.mozilla.org/docs/Web/HTML/Reference/Elements/ins
* https://developer.mozilla.org/docs/Web/HTML/Reference/Elements/del

nykula: About keeping diffs in content, my case involves long files with just a couple lines being added or removed at a time, for which showing small diffs instead of the full content with highlights seemed a good match. I also used simple `array_diff`, no full-fledged diff algorithm, so the code doesn't currently know at which places new lines appear or disappear, making it difficult to figure out where to put `<ins>` and `<del>` (otherwise, I agree using semantic HTML for that is a good idea).

I did initially try to keep the content and diff together in the content column, but I got lost in things related to loadCompleteContent, withEnclosures, _content, DOMDocument etc and didn't come up with a version of code working reliably. (Maybe someone else reading this is more confident about implementing it that way?) Keeping just the original content in the content column, while treating the diff like an optional view-related thing stored in an attribute, on the other hand, was quite straightforward to implement and worked the first time.

app/Models/EntryDAO.php
nykula: Yes, comparing against the latest received is the idea. I'll try changing that to `id` locally. Where can different dates come from? I saw custom timestamp parsing logic in loadHtmlXpath but in my branch there's currently no logic like that.

nykula: Changing to `id` has worked fine in my local copy.

nykula:
Re-applied my patch over edge. I remember that you asked to do merges instead, but I'm less familiar with that workflow. Just trying to keep my local fork in sync. The code is the same as before, except the `date` -> `id` change in `ORDER BY` as was requested.

PR #8112, opened 2025-10-13
Option to show user labels instead of tags in RSS share

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/discussions/8108#discussioncomment-14668813

image


Alkarex:
Merging to ease testing

PR #7886, opened 2025-08-31
Last user modified

pudymody:
Closes #7862 

Changes proposed in this pull request:

This is an implementation of the proposed feature. It allows entries to have a new field that will be updated whenever an item is marked as read/unerad or bookmark/removed from bookmarks. And a new sort criteria to sort by it.

I couldnt find any documentation about migrations but i think this is the way. I only tested this with sqlite, but i believe the others should work, its a copy-paste from the existing ones.

If you find this ok, i could continue with the tag implementation to also modify this date whenever a new tag is added or removed from an article. In this or another PR as preferred.

How to test the feature manually:

1. Mark items from a feed as read/unread
2. Mark items from a feed as bookmark / remove bookmark
3. Sort by the new criteria

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #8131, opened 2025-10-17
Implement filter on last modified date by server

### LINKED ISSUE #2566: Greader API : get modified items
Hello,

I am writing an [Android RSS client](https://github.com/readrops/Readrops) with FreshRSS support and I encountered a problem when syncing items using Greader API.

I would like to get from the server items which read state has been modified. Let's say I synchronize my items list with my Android client. Then, I mark some items as read in the web client. I would like to make notice the mobile client that several items have been marked as read.

I already know that if an item contains `user/-/state/com.google/read` in the categories field in the returned json, like that :
``` 
"categories": [
    "user/-/state/com.google/reading-list",  
    "user/-/label/IT",
    "user/-/state/com.google/read"
]
```
it is read an I can mark it as read in the mobile client database. 

When fetching items with `reader/api/0/stream/contents/user/-/state/com.google/reading-list` end point and the parameter `ot` (unix timestamp), I can get new items, and, if they have been read before the mobile client fetch them, they will have `user/-/state/com.google/read` in their categories field. But if I mark an item older than the ot parameter as read, I won't get it.

Am I missing something which could solve this problem ? If not, Is there anything doable for this ?

Otherwise, thanks for this awesome project that is FreshRSS !

### LINKED ISSUE #7304: [Feature] API include modified articles when filtering on ot
Allows API clients to update articles, which *content* has been modified since the last sync
https://github.com/jocmp/capyreader/discussions/533#discussioncomment-11341808
https://github.com/FreshRSS/FreshRSS/issues/2566

Related:
* https://freshrss.github.io/FreshRSS/en/developers/06_GoogleReader_API.html
* https://www.inoreader.com/developers/item-ids
* https://feedhq.readthedocs.io/en/latest/api/reference.html#stream-contents

Alkarex:
Especially relevant for API, to get the modified changes: the API will now return the articles that are new or which content has been modified since `ot`:

fix https://github.com/FreshRSS/FreshRSS/issues/7304
fix https://github.com/FreshRSS/FreshRSS/issues/2566
https://github.com/jocmp/capyreader/discussions/533#discussioncomment-11341808

New corresponding search operator `mdate:` and new UI:

image


Inverle:
What's the difference between user modification date and server modification date?

Alkarex:
> What the difference between user modification date and server modification date?

* *Server modification date*: When the article has been modified in the RSS feed, server side, by the author.
* *User modification date*: When the FreshRSS user has marked an article as read/unread favourite/unfavourite, or changed and refreshed the full content retrieval

Alkarex:
For a later PR: Help welcome to add somewhere in the UI this date of last modification

app/i18n/pl/gen.php
```diff
@@ -228,6 +228,7 @@
 		'categories' => 'Kategorie',
 		'content' => 'Zawartość',
 		'date_from' => 'Od',
+		'date_modified' => 'Server Modification Date',	// TODO
```
Inverle:
```suggestion
		'date_modified' => 'Data modyfikacji przez serwer',
```

jocmp:
@Alkarex I tried out this branch in a local dev container with the following cURL

```
curl -X GET 'http://example.com/api/greader.php/reader/api/0/stream/items/ids?s=user/-/state/com.google/reading-list&ot=1760418503&output=json' \
  --header 'Authorization: GoogleLogin auth=alice/<token>'
```

- `user/-/state/com.google/reading-list` to get the reading list
- `1760418503` to specify the articles modified since the time of `2025-10-14T00:08:23-05:00`

I can confirm that it returned a single article that had been modified by the server since that time:

```jsonc
{
  "itemRefs": [
    {
      "id": "1761617402426051"
    }
  ]
}
```

However, I don't think this fits into Capy Reader's model for fetching data ([source code](https://github.com/jocmp/capyreader/blob/16ab8c873c3ededf77a652f857389f137c36655e/capy/src/main/java/com/jocmp/capy/accounts/reader/ReaderAccountDelegate.kt#L328-L353)) as it currently exists. At the moment Capy Reader will always fetches all data from the reading list, label, or feed on refresh.

If I'm following this correctly, for Capy Reader's implementation, this would cover articles that are:

- No longer in the reading list
- Beyond the [10k latest articles](https://github.com/jocmp/capyreader/blob/cbcecc4291a68c9e4e6bc22b8a587028871efede/readerclient/src/main/java/com/jocmp/readerclient/ext/GoogleReaderExt.kt#L11) in a feed
- Have been updated

Let me know if that's the use case you have in mind and/or if I'm misunderstanding the changeset still!

Alkarex:
@jocmp Quick answer already (I have not looked at your source code):

> At the moment Capy Reader will always fetches all data from the reading list, label, or feed on refresh.

This is what I want to avoid. Article content should only be downloaded once.

The idea is to include all new articles since `ot` (as before) as well as articles which text content has changed since `ot` (new behaviour), so a client does not have to re-download article content just in case a few of them have been modified.

> If I'm following this correctly, for Capy Reader's implementation, this would cover articles that are:
> 
>     * No longer in the reading list
> 
>     * Beyond the [10k latest articles](https://github.com/jocmp/capyreader/blob/cbcecc4291a68c9e4e6bc22b8a587028871efede/readerclient/src/main/java/com/jocmp/readerclient/ext/GoogleReaderExt.kt#L11) in a feed
> 
>     * Have been updated

Right. I could modify this PR to return articles that are:
* In the top `n` new articles (according to received date), AND
* New *or modified* since `ot`

Alkarex:
P.S. Now that I think again about it, a similar problem exists already regarding the definition what is returned. `n` is used to limit the length of the response but does not give guarantee about the content, as it varies depending on e.g. the read/unread state. For instance, when filtering for instance on unread articles, the response will contain articles beyond the 10 latest articles.

P.P.S: And what is your more precise definition of the reading list?

jocmp:
My definition of reading list is the `user/-/state/com.google/reading-list` category. So when a user views Capy's "All" item it essentially displays the `reading-list` category.

In contrast, if a user enters a feed on the app, e.g. Feed #7, then the category is `feed/7`.

---

To your earlier mention of "New or modified since ot," I think including "new" based on a filter would be extremely useful since it would avoid the use of `n`, and always be based on a timestamp instead of a quantity.

My favorite existing use of something like this is still [Feedbin's "since"](https://github.com/feedbin/feedbin-api/blob/master/content/entries.md#entries) parameter that works like this:

> `GET /v2/entries.json?since=2013-02-02T14:07:33.000000Z` will get all entries created after the iso 8601 timestamp.

So in short, if `ot` also returned new items since the specified timestamp, then I could rework Capy to just fetch a diff instead of a quantity by way of the current `n` query param. Let me know what you think!

Alkarex:
Good that we talk about it. The Greader API is already (also in its implementation in FreshRSS before this PR) doing exactly what you ask. I believe the `ot` parameter is equivalent to the `since` you mention.

https://github.com/FreshRSS/FreshRSS/blob/6440f9c9a91160103b7c2c369761ada31926c02a/p/api/greader.php#L1150-L1154

The `n` is necessary to avoid overly large transfers, but it is a max, not a target. Indeed, if the last synchronisation was a long time ago, the server could very well have far more than 10k articles to transfer.

https://github.com/FreshRSS/FreshRSS/blob/6440f9c9a91160103b7c2c369761ada31926c02a/p/api/greader.php#L1146-L1147

So the only change offered by this PR is precisely to make the content since last synchronisation even better, by including articles which content has changed, something that is otherwise difficult to get, except be re-downloading everything.

jocmp:
Thanks @Alkarex. In that case I'll investigate how to re-add `/reader/api/0/stream/items/ids` with `ot` specifically for FreshRSS in Capy Reader (https://github.com/jocmp/capyreader/issues/1540). It should make it much faster!

Alkarex:
Please double-check the following approach, and feedback / questions welcome:
* https://github.com/FreshRSS/FreshRSS/issues/2566#issuecomment-541317776
* https://github.com/jocmp/capyreader/discussions/533#discussioncomment-11341808
* From https://freshrss.github.io/FreshRSS/en/developers/06_GoogleReader_API.html#synchronisation-strategy

TijnvandenEijnde:
@Alkarex I have been following your 7-step synchronization strategy (https://github.com/FreshRSS/FreshRSS/issues/2566#issuecomment-541317776), it has been very helpful thank you! 

I am trying to implement step 5 of that strategy to sync recently starred items from the web client with the Your News app, which is why I ended up on this PR. However, based on this explanation:

> > What the difference between user modification date and server modification date?
> 
> * _Server modification date_: When the article has been modified in the RSS feed, server side, by the author.
> * _User modification date_: When the FreshRSS user has marked an article as read/unread favourite/unfavourite, or changed and refreshed the full content retrieval

I am unclear whether the `ot` parameter filters by server modification date or user modification date. For step 5 of the sync strategy, should this API call:

`GET /api/greader.php/reader/api/0/stream/contents/user/-/state/com.google/starred?n=9999&ot=1763804501` 

return articles that were recently starred by the user?

I have verified that I am on the correct branch and the database has been updated with both `lastModified` and `lastUserModified` columns, but the `ot` parameter does not seem to return recently starred items. Is there a way to filter by `lastUserModified`, or should I use a different approach for this use case?

Alkarex:
Hello @TijnvandenEijnde ,
Glad to hear :-) See also https://freshrss.github.io/FreshRSS/en/developers/06_GoogleReader_API.html#synchronisation-strategy

In the current PR, the proposed idea is that `ot` will filter on the *server modification date*
https://github.com/Alkarex/FreshRSS/blob/608e222e6c70e186d58617ec03c7f76748b29152/p/api/greader.php#L660-L669

Before this PR, `ot` was filtering on the *received date*.

The recent *user modification date* (https://github.com/FreshRSS/FreshRSS/pull/7886) is not used by the API so far.

So indeed, if the user stars a very old article, and the number of favourites is higher than what is retrieved by the app, it will not appear in the results.

I will try to think more about it later tonight.

TijnvandenEijnde:
Thank you @Alkarex. I’m almost finished implementing the synchronisation strategy, and I’m very happy with the results so far. I ended up using 6 of the 7 recommended API calls:

1. `/reader/api/0/subscription/list` — Retrieves all feeds and categories.  
2. `/reader/api/0/unread-count` — Provides unread counts for each feed and category, plus the total.  
3. `/reader/api/0/stream/contents/user/-/state/com.google/reading-list` — Returns all articles.  
4. `/reader/api/0/stream/items/ids` — Returns IDs of all unread articles.  
5. `/reader/api/0/stream/contents/user/-/state/com.google/starred` — Returns all starred articles.  
6. `/reader/api/0/stream/items/ids` — Returns IDs of all starred articles.

I think the last piece I need is the change proposed in PR (https://github.com/FreshRSS/FreshRSS/pull/7886). However, I do not see any mentions of starred/unstarred, but I assume they work the same.

At the moment, I noticed that even when a recent article is starred, it does not appear in the results of API call `5` when the `ot` filter is applied. I am using the `updated` value returned by the API and using it as the `ot` filter for subsequent requests, so the timestamps should be correct. Also, the number of retrieved starred items is set to 9999, while the total number of articles in the account is only around 7000–8000, so the limit should not be an issue.

Can you confirm that this is the intended behaviour with the current implementation, or if I might be missing something?

Unstarring an item does get picked up, but that is because API call `6` returns the complete set of starred IDs.

The reason I am using API call `5` is to retrieve both read and unread starred articles. A possible workaround would be to search for recently starred items within the articles returned by API call `3`, but that does not always work because the application settings may prevent `3` from returning both read and unread articles.

Alkarex:
Thanks for the feedback. I will try to think a bit about it and propose an improvement of the current PR.
Meanwhile, please remember to test on a much larger database (like 100k+ articles, 10k+ favourites, 10k+ unreads). Typical shortcomings of new implementations is to scale poorly due to assuming small numbers.

TijnvandenEijnde:
Likewise! I will make sure to test it before releasing a new version of my application and creating a PR to update the README.md. Thank again!

jocmp:
Sorry for leaving this for so long. Time went by too quickly 😅

For Capy Reader, one goal of mine would be to avoid having to use the `n=` parameter to fetch new articles altogether. Ideally, the modification date would cover "new" articles as well

> Server modification date: When the article has been modified in the RSS feed, server side, by the author.

revised to cover

> Server modification date: When the article has been **created** or modified in the RSS feed, server side, by the author.

Alkarex:
> For Capy Reader, one goal of mine would be to avoid having to use the `n=` parameter to fetch new articles altogether. 

Even with the proposed changes, `n=` is needed (otherwise there will be an internal default value) because the number of new articles can be infinitely high.

> Ideally, the modification date would cover "new" articles as well

Yes, for sure new articles are included.

The final question I would like to double check is whether the change from providing only new articles (which is the behaviour today) to providing new articles + server-modified articles (proposed behaviour) is fine for Capy Reader.

Alkarex:
@TijnvandenEijnde Sorry for the delay, and please test again. `ot` now operates on the date when articles have been received as well as when articles have been modified by the server. After pondering pros and cons, the date of user modification is *not* used as an inclusion or exclusion criteria.
The only change is thus that older articles modified by the server may be included among the newly received articles.
The idea is to avoid having to re-download old articles just in case they might have been modified.

jocmp:
> The final question I would like to double check is whether the change from providing only new articles (which is the behaviour today) to providing new articles + server-modified articles (proposed behaviour) is fine for Capy Reader.

I'll try and test this this week!

Maybe to give an example, say a feed has 1 unread article and 1 read article and the client has cached both.

Let's say 2 new articles are fetched on the server and the read article is updated. I would expect that 3 articles are returned with the `ot` flag to the client.

jocmp:
@Alkarex I tested out this branch, and I believe I have something working with the PR linked above. 

Based on a client-side "last refreshed at" timestamp, Capy Reader will now fire a request that looks like the following cURL. The existing `n=10000` may be overkill. Let me know if you have any thoughts!

```
curl 'https://example.com/api/greader.php/reader/api/0/stream/items/ids?s=user%2F-%2Fstate%2Fcom.google%2Freading-list&ot=1772395935&n=10000&output=json' \
  -H 'authorization: GoogleLogin auth=alice/<something secret>'
```

One last question: say a user is on an older FreshRSS instance (e.g. v1.27 and this is released in v1.29) - will the `ot` variable still work even if it doesn't include the modified articles?

Alkarex:
Sound good :-)

> One last question: say a user is on an older FreshRSS instance (e.g. v1.27 and this is released in v1.29) - will the `ot` variable still work even if it doesn't include the modified articles?

For older FreshRSS instances, `ot` will work as it always have, i.e. by returning new articles, but not including articles, which have been modified at their source. So let's say an author has corrected a typo in an article, then it will not be included in that batch. (I do not think it is a big problem).

TijnvandenEijnde:
> @TijnvandenEijnde Sorry for the delay, and please test again. `ot` now operates on the date when articles have been received as well as when articles have been modified by the server. After pondering pros and cons, the date of user modification is _not_ used as an inclusion or exclusion criteria. The only change is thus that older articles modified by the server may be included among the newly received articles. The idea is to avoid having to re-download old articles just in case they might have been modified.

Hi @Alkarex, apologies for the late response. I just tested it and everything is working great. Thank you for your work!

Issue #2342, opened 2019-04-04
create a `latest-release` git tag


hrehfeld:
I update my installation with `git pull && git checkout <latest release tag>`. 

How about we maintain a `latest-release` tag in the git repo that always points to the latest release (e.g. `latest-release -> 1.14.0`), so we can have easy cronjobs like `git pull && git checkout latest-release`?

hrehfeld:
(once the database needs migration, or manual intervention is required, this would most likely break)

Frenzie:
On this repo that's exactly what `git checkout master` would do.

Alkarex:
Like @Frenzie says, this is what the /master branch actually is.
All the development is done in features branches, which are then merged in /dev (which can be seen as a rolling release branch), which is then every 2-3 months merged in /master to form a new release.
Example: https://github.com/FreshRSS/FreshRSS/pull/2338

hrehfeld:
Awesome, sorry for not checking `master` before ;-)

As you added the Documentation label, I'm not gonna close, but feel free to do so!

PR #8140, opened 2025-10-21
dark theme background_color

McFev:

- Added background_color so that the app doesn't blind with white color upon opening.

1. open app on android

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/p/themes/manifest.json
+++ b/p/themes/manifest.json
@@ -25,5 +25,6 @@
   ],
   "categories": ["news"],
   "display": "standalone",
+  "background_color": "#002342",
   "description": "FreshRSS is a self-hosted RSS feed aggregator."
 }


math-GH:
I am not sure if this is a good approach here. The dark background could be nice for the users who prefer the dark background but what about the light background prefered users? They would have a dark background too now.

Currently I cannot reproduce the reason of this PR. I use FreshRSS with Firefox and dark theme on Android. The background is automatically dark.

Could you screen cast your current issue?

Frenzie:
The issue is the brief few hundred millisecond flash when opening the app. No background will otherwise be affected since that's all in CSS. I don't think there's another way to do this and it seems reasonable to assume that a dark flash is generally less bothersome than a light flash.

McFev:
p.s. I looked into what other apps do this, found Telegram - it starts with a dark background on any selected theme.

math-GH:
unfortunately I cannot watch the recording
grafik

> it seems reasonable to assume that a dark flash is generally less bothersome than a light flash.

I agree

Frenzie:
> p.s. I looked into what other apps do this, found Telegram - it starts with a dark background on any selected theme.

That seems very odd for a native Android app btw. Unlike the web manifest they can follow along with light/dark.

p/themes/manifest.json
```diff
@@ -25,5 +25,6 @@
   ],
   "categories": ["news"],
   "display": "standalone",
+  "background_color": "#002342",
```
Frenzie, line 28: I forgot to ask, does this color come from anywhere in particular?

McFev: I used https://realfavicongenerator.net/ image

`#002342` darker than the dark blue from the logo. I am not against any dark color, this color is an example.

Frenzie (approved):

Alkarex:
Thanks @McFev ! Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #8149, opened 2025-10-22
Configuration → Display → Theme: increase button visibility

Frenzie:
Closes #6577.

Maintains a little bit of distinction between hover and non-hover, but the focus is on always making it visible.

### Before
Image

### After
Left not hovered, right left arrow hovered. The distinction is obvious enough to show activation when actually hovering; it doesn't show very well on the screenshot.
image image

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

Issue #6577, opened 2024-06-21
[Feature] Wow, that theme can be changed. I just found out


w316525096:
**Is your feature request related to a problem? Please describe.**
The theme here is difficult to discover, at least for me. It took a week of use before I realized that the theme here could be changed.

**Describe the solution you’d like**
Make the change options more obvious; the white arrow in the red box is not easy to find.

**Describe alternatives you’ve considered**
A clear and concise description of any alternative solutions or features you’ve considered.

**Additional context**
screenshots： 
![image](https://github.com/FreshRSS/FreshRSS/assets/25398881/74303ccc-307e-4d85-af7d-73a9f32ac062)

aledeg:
Maybe we could change how we display that. My thought are:

1. dropping the slide show.
2. extracting the theme selection to its own page.
3. display every theme at one on said page.

Any thoughts?

Frenzie:
That sounds a bit unwieldy to me. You could also just add a dropdown above the preview?

aledeg:
My thoughts were that if we have themes with configuration, it will be easier to introduce it in a separate page.

For instance if you could chose the font of the theme, the color palette, or any other parameter, it will be easier if everything is visible at once.

shadowlocked:
It would be enough to move the arrows outside of the preview box, because they are far too transparent and subtle. As a new user desperate for some way to change theme, I [vibe-coded my own](https://github.com/FreshRSS/FreshRSS/discussions/8147) before someone filled me in!

I say just add a drop-down or lose the transparency on the arrows, or shift the CSS around a bit so that the switching mechanism is very clear. I don't think it should get its own page.

Frenzie:
I just took a look and there's also a super easy change lurking in there.

opacity: 0.5 (now)

<img width="1920" height="960" alt="Image" src="https://github.com/user-attachments/assets/4ebbffa3-e90d-42b2-9cdd-536197eabde3" />

<hr>

opacity: 1

<img width="1920" height="960" alt="Image" src="https://github.com/user-attachments/assets/a8dacfa2-7582-4268-96ea-a9ce0cdd17d4" />

<hr>

and additionally black shadow instead of gray

<img width="1920" height="960" alt="Image" src="https://github.com/user-attachments/assets/8d2464c4-35bb-4a6d-ad8e-abeb6116ce22" />

PR #8152, opened 2025-10-23
Improve theme slider

math-GH:
Ref #8149
Ref #6577

Before
grafik

After
Buttons as navigation elements. Not it is crystal clear that the themes can be changed.
The theme counter is a bit more prominent to show the big amount of shipped themes.
grafik


math-GH:
to discuss:
having a `title` attribute onto the prev/next buttons. Currently there is no generic next/prev i18n String

Frenzie:
I'd prefer my original suggestion fwiw: that is, stick a simple select above it (and hide the other stuff from screenreaders).

math-GH:
a pro argument for the current solution: it is 0% JavaScript, just HTML+CSS

Frenzie:
Perhaps. The CSS tortures radiobuttons into submission while a select would work entirely as it should with or without JS. But it's true that without JS the approaches would be competitive rather than complementary, so you'd have to hide one of the two.

Frenzie:
I should clarify that from my perspective the main or only issues with that theme selector are that you can't browse through the themes with your arrow keys and that you're forced to go through them one by one. Also the focus is lost every time you click, while it should remain on the previous or next button.

In short, if you don't like the arrows I'd remove them completely instead, although I'd rather just keep it as it is and add a select to resolve the remaining issues. (I was planning to look into that this weekend if I found the time.)

Alkarex:
Let's merge then, but additional simplification welcome

PR #8154, opened 2025-10-24
CssXPath: Implement ~ subsequent-sibling

### LINKED ISSUE #8143: [Bug] subsequent-sibling combinator (~) breaks CSS selector of the elements to remove
### Describe the bug

Having a tilde anywhere in the css selector of elements to remove seems to break the entire selector, rather than the desired behavior of removing elements matching the right side that come after an element matching the left side - see https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Combinators#subsequent-sibling_combinator

(What I'd really like is a content-based selector like `:contains("Your ad here!")` because the feed I'm reading sticks a bunch of ads into the text at the end of each post. But I noticed that one of the ads has a button and figured I could use something like `button ~ p` to match all the ads after that one. Alas, it didn't work.)

### To Reproduce

1. Go to Manage for any feed that uses the Content retrieval feature to grab the website's contents
2. Under "Advanced" > "CSS selector of the elements to remove", enter a basic CSS selector like `img` that matches something in the article and observe that it works by clicking the eyeball button in the Content retrial section
3. Add a comma and a new selector with a subsequent sibling contaminator, so that the full field is something like `img, img ~ p` - this should remove all images and all paragraphs that follow an image
4. Click the eyeball button again and observe the no content is being removed now

### Expected behavior

Correct behavior would be to implement the subsequent sibling contaminator, e.g. `img~p` should remove all `<p>` tags that follow an `` tag within the same parent tag. 

Silently skipping over just that comma-separated part of the selector would be less bad

A way to enter multiple separate selectors that are evaluated independently, instead of a comma-separated list that all gets evaluated at once would also be nice.

### FreshRSS version

1.27.1 

### System information

- Database version: SQLite?
- PHP version: 8.4.13
- Installation type: Docker - lscr.io/linuxserver/freshrss
- Web server type: openresty 
- Device: Macbook Pro
- OS: Unraid on the server, MacOS 15.7.1
- Browser: Firefox 143.0.4

It'd be nice if there was a page in the admin settings that could provide most of this info. It has the PHP version, and I thought it had the db info in there somewhere, but I can't seem to find it now.

### Additional context

_No response_

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/8143
Upstream PR https://github.com/phpgt/CssXPath/pull/231


Alkarex:
Ping @nfriedly if you can test

nfriedly:
I'll try to test it out. I have a lot going on in the next few days, so I'm not sure how soon I'll get to it, though. 

I've been using the linuxserver.io docker image, but I think I can figure out how to get it running from source.

PR #8161, opened 2025-10-26
Fix DOM fatal error while filtering retrieved full content

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/8132 was not sufficient.
Fix *PHP Fatal error: Uncaught Error: Couldn't fetch DOMElement. Node no longer exists*


Frenzie (approved):
🤞

Issue #8157, opened 2025-10-25
[Bug] "Mark as read" button unresponsive


selfdocumentingcode:
### Describe the bug

"The Mark as read" button is unresponsive.
No client side errors are logged in the browser console when clicking the button, nor are any requests sent to the backend.  

### To Reproduce

1. Go to 'Any feed page'
2. Click on 'Mark as read''
4. Nothing happens

### Expected behavior

A confirmation pop-up for marking the feed as read. 

### FreshRSS version

1.27.1

### System information

- Database version: SQLite 3.40.1
- PHP version: 8.2.29
- Installation type: Docker
- Web server type: Apache/2.4.65 (Debian)
- Device: Computer
- OS: Linux 6.16.7-200.fc42.x86_64
- Browser: Firefox 144.0

Inverle:
> No client side errors are logged in the browser console when clicking the button, nor are any requests sent to the backend.

Some other things to check:

* Is the button enabled? (Maybe CSS is not showing it's disabled so check inspect element)
* Try using the <kbd>Shift + R</kbd> shortcut which does the same as the button
* Do you have any FreshRSS extensions enabled? If so try disabling them and check again
* Is the button still unresponsive after disabling the *Display a confirmation dialog on “mark all as read” actions* option in *Configuration > Reading*?

Frenzie:
Also potentially relevant (but only after trying to disable any extensions), which theme?

selfdocumentingcode:
It just suddenly started working again after re-opening FreshRSS, now, a few hours later since I reported it. (I did not restart the browser).

The button did not look disabled. (it was colored blue in my theme, as compared to the disabled ones which where gray)
I do not have any FreshRSS extensions installed for what is worth. 

Something I probably should have mentioned in my original post is that I had just subscribed to a new feed which had around 300 entries and when I tried to mark them all as read that's when I experienced the button being unresponsive.
Re-opening FreshRSS did not help, and the mark as read button did not work for any feed.  

So, I can't reproduce it anymore, but I will try to look for the things you mentioned if it happens again. 

@Frenzie I usually use "Origine-compact", but I suspected it might be theme related too so I switched to another (Swage), but the button was still unresponsive even after switching the theme.

math-GH:
Let's close this issue for now. Please reopen it if you can reproduce it. It would be great if you could support it with screen shots

PR #8163, opened 2025-10-27
Mark as read 1 or 7 days based on publication date

Alkarex:
Allow the UI for marking as read articles older than 1 or 7 days to also work when sorting by publication date (in which case, the publication date is used).

image


Frenzie:
Is this related to #8157?

Frenzie (approved):

Alkarex:
> Is this related to #8157?

No, it should not be

PR #8165, opened 2025-10-29
Improve anonymous authentication logic

Inverle:
* Moves token checking logic to a single function called in `FreshRSS_Auth::removeAccess()`: `Minz_Request::tokenIsOk()` (the logic is also corrected)
* The token is no longer displayed in the HTML source for users viewing anonymously without signing in


Alkarex:
I do not get why `tokenIsOk()` is only called from `FreshRSS_Auth::removeAccess()` 🤔

Inverle:
It may be useful somewhere else later, anyway it makes `FreshRSS_Auth::removeAccess()` easier to understand

app/Controllers/feedController.php
```diff
@@ -13,12 +13,6 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 	#[\Override]
 	public function firstAction(): void {
 		if (!FreshRSS_Auth::hasAccess()) {
-			// Token is useful in the case that anonymous refresh is forbidden
-			// and CRON task cannot be used with php command so the user can
-			// set a CRON task to refresh his feeds by using token inside url
-			$token = FreshRSS_Context::userConf()->token;
-			$token_param = Minz_Request::paramString('token');
-			$token_is_ok = ($token != '' && $token == $token_param);
```
Alkarex, line 21: Is this part (for instance) still working, as it is not calling `tokenIsOk()` ?

Inverle: `tokenIsOk()` is checked earlier before the action is executed (yes, it's working)

Alkarex: Do you mean due to calling `removeAccess()` ?

Inverle: Yes

Alkarex: I have not looked at the exact flow, but I think it would nicer and less error prone during future maintenance to keep some explicit logic at such places

Inverle:
@Alkarex Can we merge this now?

Alkarex:
> @Alkarex Can we merge this now?

On my way to bed, but I will try to check tomorrow morning 🥱

Alkarex (approved):

app/Controllers/feedController.php
Inverle: https://github.com/FreshRSS/FreshRSS/pull/8371

PR #8167, opened 2025-10-30
Add support for keyboard navigation in theme switcher

Inverle:
Partially solves issue (navigating with keyboard) described in https://github.com/FreshRSS/FreshRSS/pull/8152#issuecomment-3439257667

The JS part doesn't work in SeaMonkey, I have not investigated why.


Alkarex:
> The JS part doesn't work in SeaMonkey, I have not investigated why.

I am more surprised that it works in other browsers. If I am not mistaken, `<label>` elements are not supposed to be directly focusable but rather depend on their associated `<input>` to be focussed, which happens to be hidden in our case here.

I think the underlying method to change style should be reworked to be simpler and more semantically valid.

We can merge here in the meantime, but it adds to something that is not so nice.

p/scripts/extra.js
```diff
@@ -129,6 +129,32 @@ function init_password_observers(parent) {
 }
 // </show password>
 
+function init_display(parent) {
+	// Keyboard support for theme switcher navigation
+	const wrapper = parent.querySelector('div.theme-preview-list-wrapper');
+	if (wrapper) {
+		wrapper.addEventListener('keypress', function (e) {
```
Alkarex, line 140: `keypress` is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event

Frenzie:
Yes, I'd rather avoid such complications and just simplify it to a `select`, since then you also wouldn't need to move left or right ten times to reach a desired theme. All it'd need is an onchange → update image. Though I suppose if you really want you can also use role=radio and the like to construct whatever you want.

p/scripts/extra.js
Inverle: Interesting

Inverle:
> We can merge here in the meantime, but it adds to something that is not so nice.

I'm gonna make a second PR for the `<select>` approach and close this one maybe

p/scripts/extra.js
Frenzie: Keydown is preferable for most things anyway. Keyup/press just adds a bit of extra delay.

Inverle:
https://github.com/FreshRSS/FreshRSS/pull/8190

PR #8169, opened 2025-10-31
[sqlite export] add username and timestamp

scollovati:

- add username and timestamp to sqlite user export, similar to the ZIP export. Useful for archiving purposes.

1. Go to http://localhost:8080/i/?c=importExport
2. Export Sqlite at the end of the page
3. Verify that the filename is `freshrss_[username]_[date exported]_db.sqlite`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -752,11 +752,13 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 			return;
 		}
 
-		// Format export as for the ZIP file. See Services/ExportService::zip
-		$username = Minz_User::name() ?? '_';
-		$day = date('Y-m-d');
-		$this->view->sqliteName = 'freshrss_' . $username . '_' . $day . '_db.sqlite';
 		$this->view->sqlitePath = $path;
+		$this->view->sqliteName = basename($path);
+		if ($this->view->sqliteName === 'db.sqlite') {
+			$username = Minz_User::name() ?? '_';
+			$date = date('Y-m-d_H-i-s', filemtime($path) ?: time());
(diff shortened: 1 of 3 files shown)


Alkarex:
Hello and thanks for your PR. At a quick glance, this would break existing features. The idea is to be able to download multiple versions:

image

I understand that your PR is for the case of an SQLite setup with only the main file and no back-up, but we need to address the case of back-ups and other databases. I am taking a little look now.

By the way, it would be very nice to also be able to trigger back-ups from the Web UI. The challenge is to account for servers with limited resources and potentially many users. Help welcome:

* PR 1 (would be enough to have the settings in `config.php` for now and without UI):
    * Setting to allow running automatic export to SQLite
    * Setting to limit the number of kept past SQLite exports
    * Setting to chose the max DB size to allow SQLite export from Web UI
* Later PR
    * Mechanism to import an SQLite backup from Web UI (would first require some discussions)
* Later PR
    * Mechanism to ask for an SQLite export, which will be done at the next cron run, useful for larger DBs for which the Web UI could be problematic

Alkarex:
Fixed in https://github.com/FreshRSS/FreshRSS/pull/8169/commits/576053ca95a1cedf4d73775741170bf339a7e44d
Double-check welcome

scollovati:
> Fixed in [576053c](https://github.com/FreshRSS/FreshRSS/commit/576053ca95a1cedf4d73775741170bf339a7e44d) Double-check welcome

Thanks @Alkarex it's working fine. I've never seen more than one file, so I was not aware of that possibility.

PR #8182, opened 2025-11-02
Custom favicons async/await

Alkarex:
Rewrote the last Promise pattern of our code-base with an async/await pattern: [custom feed favicons](https://github.com/FreshRSS/FreshRSS/pull/7646)
Related to:
* https://github.com/FreshRSS/FreshRSS/pull/7962

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -249,26 +249,27 @@ function init_update_feed() {
 	});
 
 	if (faviconExtBtn) {
-		faviconExtBtn.onclick = function (e) {
+		faviconExtBtn.onclick = async function (e) {
 			e.preventDefault();
 			faviconExtBtn.disabled = true;
-			fetch(faviconExtBtn.dataset.extensionUrl, {
-				method: 'POST',
-				headers: {
-					'Content-Type': 'application/json; charset=utf-8'
-				},
-				body: JSON.stringify({
-					'_csrf': context.csrf,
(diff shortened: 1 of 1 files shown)


Alkarex:
Ping @Inverle

Frenzie:
Why is it necessary to await them instead of just acting when it's done?

Alkarex:
> Why is it necessary to await them instead of just acting when it's done?

I have not looked so much at the logic - this PR is merely a rewrite with another syntax.
But acting when it is done is one of the goals of await. Do you have a simpler approach in mind?

Frenzie:
Never mind, I was merely mistaken about what await does then. <del>But mind my comparative lack of familiarity is because await is brand new while the other way is a decade old, which seems like an independently relevant fact.</del>

Edit: never mind, I misread caniuse. Their vertical alignment seems a bit odd.

Frenzie:
> I have not looked so much at the logic - this PR is merely a rewrite with another syntax. 

But in any case, what's the point of that?

Alkarex:
> never mind, I misread caniuse

Indeed newer than Promise, but I think we have waited enough :-) We had a first example of async/await in the same file a while ago already:

image

> But in any case, what's the point of that?

To finalise the migration to the modern syntax, and avoid PRs which risk taking an example from elsewhere in our existing code.

Frenzie (approved):

PR #8193, opened 2025-11-08
Housekeeping lib_rss.php

Alkarex:
`lib_rss.php` had become much too large, especially after https://github.com/FreshRSS/FreshRSS/pull/7924
Moved most functions to other places (reduced by 905 lines).
Mostly no change of code otherwise (see comments).


app/Models/SimplePieCustom.php
```diff
@@ -0,0 +1,295 @@
+<?php
+declare(strict_types=1);
+
+final class FreshRSS_SimplePieCustom extends \SimplePie\SimplePie
```
Alkarex, line 4: Here is one of the main changes instead of the function `customSimplePie()`

app/Models/SimplePieResponse.php
```diff
@@ -5,7 +5,9 @@ final class FreshRSS_SimplePieResponse extends \SimplePie\File
 {
 	#[\Override]
 	protected function on_http_response($response, array $curl_options = []): void {
-		syslog(LOG_INFO, 'FreshRSS SimplePie GET ' . $this->get_status_code() . ' ' . \SimplePie\Misc::url_remove_credentials($this->get_final_requested_uri()));
+		if (FreshRSS_Context::systemConf()->simplepie_syslog_enabled) {
```
Alkarex, line 8: Fixed bug

app/Models/SimplePieCustom.php
```diff
@@ -0,0 +1,295 @@
...
+{
+	/**
+	 * @param array<string,mixed> $attributes
+	 * @param array<int,mixed> $curl_options
+	 * @throws FreshRSS_Context_Exception
+	 */
+	public function __construct(array $attributes = [], array $curl_options = []) {
+		parent::__construct();
+		$limits = FreshRSS_Context::systemConf()->limits;
+		$this->get_registry()->register(\SimplePie\File::class, FreshRSS_SimplePieResponse::class);
```
Alkarex, line 14: Fixed bug: before, this was only called when `simplepie_syslog_enabled` was enabled, preventing the logic for HTTP 429 and HTTP 503 to work.

lib/lib_rss.php
```diff
@@ -922,253 +307,6 @@ function invalidateHttpCache(string $username = ''): bool {
...
-
-	$ip_net_bits = substr($binary_ip, 0, $mask_bits);
-	$subnet_bits = substr($binary_subnet, 0, $mask_bits);
-	return $ip_net_bits === $subnet_bits;
-}
-
-/**
- * Use CONN_REMOTE_ADDR (if available, to be robust even when using Apache mod_remoteip) or REMOTE_ADDR environment variable to determine the connection IP.
- */
-function connectionRemoteAddress(): string {
```
Inverle, line 1031: Breaks xExtension-Captcha https://github.com/FreshRSS/Extensions/blob/869a1d1d5707c8ca490414cc2bff1b18c3b39983/xExtension-Captcha/extension.php#L38

Alkarex: Thanks, [fixed ](https://github.com/FreshRSS/FreshRSS/pull/8193/commits/37688a79d1af80c245dd88380f176518b3412337) and checked by running tests in our Extensions repo: 
```sh
cd ../Extensions/
composer test
composer run-script phpstan-third-party
```

marienfressinaud:
Note that these change can break extensions that rely on methods that have been moved. For instance, I've fixed the demo extension which was using the `max_registrations_reached()` function (https://github.com/FreshRSS/xExtension-Demo/commit/22fcda03c21996649072faf89d3c577adbcdb6d2).

Maybe some communication can be useful before releasing the next version?

Alkarex:
> Note that these change can break extensions that rely on methods that have been moved. For instance, I've fixed the demo extension which was using the `max_registrations_reached()` function ([FreshRSS/xExtension-Demo@22fcda0](https://github.com/FreshRSS/xExtension-Demo/commit/22fcda03c21996649072faf89d3c577adbcdb6d2)).
> 
> Maybe some communication can be useful before releasing the next version?

Yes indeed, although my impression is that none of the communication channels that we have have a very impactful effect.
This is also a reason for running a good period of time in our `edge` rolling release before a slow release.

I also have some TODO to improve e.g. https://github.com/FreshRSS/xExtension-HelloWorld to have a proper CI/CD to detect such things automatically, which I hope can be gradually incorporated to existing extensions.

Alkarex:
At least:
https://github.com/FreshRSS/FreshRSS/blob/190f8ecaf4cd5f2f3df48424466f31a460ae65ea/CHANGELOG.md?plain=1#L111

Issue #7611, opened 2025-05-24
[Feature] Get rid of SCSS - just use plain CSS


math-GH:
### Is your feature request related to a problem? Please describe.

Over the years CSS got a lot of improvments that were bridged before with SCSS and other precompiling frameworks.

### Use case

CSS variables and CSS nesting are now in the widely basline of the most used browser.

Fallback: If someone uses a browser that does not support them: the most standard themes are currently not using CSS nesting

### Describe the solution you’d like

TODO:
* [x] Ansum theme https://github.com/FreshRSS/FreshRSS/pull/8241
* [x] Mapco theme https://github.com/FreshRSS/FreshRSS/pull/8241
* [x] Swage theme https://github.com/FreshRSS/FreshRSS/pull/8200
* [x] delete the dev tools dependencies of SCSS https://github.com/FreshRSS/FreshRSS/pull/8241

math-GH:
Could we delete the SCSS files and keep the css files as they are now? Or should we take something over from SCSS into CSS?

Ping @Inverle

Alkarex:
I think the structure of the existing SCSS files should be kept, using standard CSS nesting

Inverle:
Yes, but I'm not sure if there is any specific SCSS syntax that needs to be removed from the .scss files

Frenzie:
Perhaps SASS has added an option to export to native CSS nesting?

Alkarex:
First step on Swage: https://github.com/FreshRSS/FreshRSS/pull/8200

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/8241
Tests welcome

PR #8200, opened 2025-11-09
Reduce SCSS: rewrite Swage as CSS

Alkarex:
Contribute to https://github.com/FreshRSS/FreshRSS/issues/7611

Convert from SCSS to native CSS.

Taking advantage of:
* https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Nesting
* https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector
* https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Colors/Using_relative_colors#using_math_functions

This was mostly to get started and get a better idea of the difficulties.

I am happy to see that the differences between the original SCSS and the native CSS are limited.
The relevant diff to look at is: https://github.com/FreshRSS/FreshRSS/pull/8200/commits/dd1bc7c663cf7afdde7af3161ddc213d360247ef

The main remaining issue is that the *CSS Color Module Level 4, Relative Color* is not mainstream yet, so a small compatibility section is needed with precomputed colour values.


p/themes/Ansum/_mixins.scss
Alkarex, line 1: Unused

p/themes/Swage/swage.css
```diff
@@ -1,85 +1,84 @@
...
+		--color-text-bad-lighter: #c46178;
+		--color-box-shadow-light: #ded7d7;
+		--color-border-light-darker: #ded7d7;
+		--color-border-nav: #00488b;
+		--color-background-light-darker: #ded7d7;
+		--color-background-nav-darker: #00488b;
+		--color-background-alert-darker: #f95c20;
+		--color-background-bad: #c46178;
+	}
 }
```
Alkarex, line 82: Remaining compatibility section

Frenzie:
Fwiw I assume parsing it will be slightly slower? Though probably negligible compared to even our fairly fast and minimal JS.

PR #8201, opened 2025-11-11
Implement button for toggling sidebar on all views

### LINKED ISSUE #2792: mobile UI in desktop?
is there any option to set the UI view on firefox desktop to appear as mobile view like my tablet?
I like the simple and wide view of mobile view.

### LINKED ISSUE #4136: Show/Hide Subscription panel[Feature]
subscription list(left of screen) in get around 1/3 page that is waste of space and distracting
you have an option to hide subscription list in freshness or even show only when mouse hover that area.

### LINKED ISSUE #4224: [BUG] Reading View: Feed menu in side bar: does not work

### LINKED ISSUE #5338: [Feature] Add show/hide button for subscription list in normal view
When in normal view, the subscription list takes a lot of screen space when using freshrss on a tablet screen in portrait orientation.

Enable a show/hide subscription list pane to the top menu.

Have:
1. Reading view abide by the "Show articles unfolded by default" option in Settings (Reading tab) or,
2. Have separate "Show articles unfolded by default" settings for Normal and Reading views in Settings.

### LINKED ISSUE #6119: [Feature] Hide/Unhide sidebar
When I have a lot of new articles in a particular category, I would like to hide the sidebar so that I have more space to display the title of articles.

Button to Hide/Unhide sidebar

-

**Additional context**
-

### LINKED ISSUE #7100: Option to hide the side menu bar on the left
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/7099

<sup>Originally posted by **sassanix** December 13, 2024</sup>
It would be nice to have the option to hide the sidebar with a click, and then allow it to come back. 

It would make it easier to read in portrait mode. </div>

### LINKED ISSUE #7673: [Feature] Toggle sidebar button in all views

Reader view has a toggle icon to display or add the feeds sidebar. Other views don’t have this option.

### Use case

_No response_

Add this feature to all views, including views created by extensions. It can always be useful to get extra space at times, whichever the view.

When the sidebar is hidden, I’d also suggest to add management/add buttons in the toolbar, since they are also hidden.

### Additional context

By the way this button as it is implemented currently in Reader view isn’t accessible: it doesn’t have a title mostly, so it’s hard to guess what the button is about.

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/7673, https://github.com/FreshRSS/FreshRSS/issues/7100, https://github.com/FreshRSS/FreshRSS/issues/6119, https://github.com/FreshRSS/FreshRSS/issues/5338, https://github.com/FreshRSS/FreshRSS/issues/2792, https://github.com/FreshRSS/FreshRSS/issues/4224, https://github.com/FreshRSS/FreshRSS/issues/4136

https://github.com/user-attachments/assets/0629e465-6450-440e-b38b-430e9ff73ef9

Keyboard shortcut for doing the same: <kbd>t</kbd>

TODO:
* [x] fix sidebar buttons in other views that aren't working anymore now since I missed them (e.g. repartition)


app/i18n/fr/conf.php
```diff
@@ -359,6 +359,7 @@
 		'skip_next_article' => 'Passer au suivant sans ouvrir',
 		'skip_previous_article' => 'Passer au précédent sans ouvrir',
 		'title' => 'Raccourcis',
+		'toggle_aside' => 'Toggle sidebar',	// TODO
```
```suggestion
		'toggle_aside' => 'Afficher/masquer la barre latérale',
```

p/themes/base-theme/frss.css
```diff
@@ -2743,6 +2735,7 @@ html.slider-active {
 	}
 
 	#nav_entries {
+		display: table !important;
```
Alkarex, line 2738: When possible, I am happier when `!important` can be avoided, for instance with a more precise selector

Alkarex:
I like the idea but I am not sure the icon and its location is optimal. This could wait to be improved in a follow-up PR, though.

See also https://framagit.org/nicofrand/xextension-togglablemenu

Inverle:
@Alkarex I have made a slightly different implementation before that puts the sidebar button next to the *Subscription management* button, but decided this is better.

https://github.com/FreshRSS/FreshRSS/issues/6119#issuecomment-2923698582

https://github.com/user-attachments/assets/809b67cb-6eb5-4894-8b02-5b111707541b

And the extension is bugged in some ways, just like my implementation above (not shown on video)

Inverle:
Though I do wonder about the icon

Inverle:
Good to see that SCSS is gone from Swage

math-GH:
I am fine with the folder icon because it is now used system wide for the same purpose.

mobile view on almost every page:
grafik

reading view:
grafik

Alkarex:
I can see @math-GH 's point of uniformity. On the other hand, I think the second option looks better for full screen, so I do not have any clear preference.

Alkarex:
Ok, then I suggest to go back to the folder icon for this PR, and come back for improvements in a follow-up.

p/scripts/main.js
```diff
@@ -922,6 +922,64 @@ function init_posts() {
...
+	const media = window.matchMedia('(max-width: 840px)');
+	media.onchange = sync;
+
+	const state = sessionStorage.getItem(`FreshRSS_aside-toggled_${context.current_view}`);
+	if (state !== null) {
+		const active = toggle_aside.classList.contains('active');
+		if (state != active) toggle_aside_click(false);
+	}
+	if (getComputedStyle(aside).display !== 'none') {
+		if (context.current_view === 'normal') aside.classList.add('visible');
```
Inverle, line 971: I don't remember if this line is needed, could be removed later maybe

Alkarex:
Related extension https://github.com/Brawl345/FreshRSS-Toggle-Sidebar

Alkarex:
Just thinking about that now: I think it would be better to align the new button to the left like so:

image

Help welcome

math-GH:
> Just thinking about that now: I think it would be better to align the new button to the left like so:
> image
> 
> Help welcome

This was also my first thought when the update came in

civilblur:
Is there still space for discussion regarding the button placement and the icon itself?

The reason I'm asking is that many sidebar toggle icons have shifted towards [one that looks like this](https://www.svgrepo.com/svg/513851/sidebar-left):


You'll find this type of style in apps like Slack, Figma, Firefox, and MS Edge, where it's typically positioned on the top left area, above the sidebar itself. We can also see something similar on Github today, with a hamburger (three stripes) icon instead.

A folder icon may not provide the best affordance because users typically associate it with drilling down into content, rather than toggling a view on/off.

Frenzie:
It's not so much "a view" as it is "the folder (or category) list." It's the same principle as the "Show member list" icon in Discord. Said member list currently appears as a sidebar.

image

I don't necessarily think using a sidebar or sidebar expand/collapse icon is a particularly good option in comparison, because then it's all about the shape it currently happens to have by default instead of that which it does, and a lack of consistency between desktop and mobile. In a browser it's different because there you have many different sidebar fill options.

> Slack, […] Firefox […]

That's weird. Neither of those have that for me. Not that I consider something like Firefox or Edge relevant to this program regardless since that's a sidebar, not a folder list.

Alkarex:
Indeed, I also think the visual affordance should be improved somehow, especially since the folder icon is used for at least two different meanings, especially in the mobile view.

I have one in Firefox in the top left (and hovering over it makes a little animation):
image

In Firefox, it is for the list of tabs, but I think what the icon tries to convey is that there is an extra panel, independently of its content.

civilblur:
I agree that an icon should convey its function, which is for this reason I wanted to propose a change to the icon and its placement. I believe most users see FreshRSS' left sidebar as their primary way to navigate, rather than solely a category list. If this is true, it suggests that hierarchically speaking, one would identify the sidebar as "navigation" first, with "category" as a secondary layer.

As for "Show member list" on Discord, since it isn't primarily used as a navigation, the way it works there makes sense for their use case. Discord has also positioned the icon at a higher hierarchical level, which is the approach I'm proposing.

> That's weird. Neither of those have that for me. Not that I consider something like Firefox or Edge relevant to this program regardless since that's a sidebar, not a folder list.

Why the other applications are still relevant, even if they're not RSS readers, is due to the prevalence of the icon and how that icon becomes associated with navigating the application. Whether users understand the "category" concept or not, they are already tuned to understand what the sidebar icon implies.

On Slack, this icon is available only if you have multiple workspaces, which might be why the icon isn't there. You can even find this icon on Apple's own iOS applications these days, so a large portion of the population is getting accustomed to it. It's similar to how people learned to understand the hamburger menu icon, even though it started out with a lot of criticism.

image
[Image source](https://www.idownloadblog.com/2024/07/02/how-to-use-tab-bar-ipad-apps/)

PR #8211, opened 2025-11-15
force correct index for MySQL when listing entries

martinhartmann:
# Issue

I have 700k+ entries in my `_entry` table and the performance is abysmal (10+ seconds) when querying the overview (front page) and having the "unread" filter set. I dug into the code and enabled the slow query log which pointed me at the `\FreshRSS_EntryDAO::sqlListWhere` method taking an absurd amount of time.

# Cause

The query it's producing is
```sql
SELECT
	e0.id
  , e0.guid
  , e0.title
  , e0.author
  , UNCOMPRESS( e0.content_bin ) content
  , e0.link
  , e0.date
  , e0.`lastSeen`
  , e0.`lastUserModified`
  , HEX( e0.hash ) hash
  , e0.is_read
  , e0.is_favorite
  , e0.id_feed
  , e0.tags
  , e0.attributes
FROM
	`user_entry` e0
		INNER JOIN (
			           SELECT
				           e.id
			           FROM
				           `user_entry` e
					           INNER JOIN `user_feed` f
					           ON f.id = e.id_feed
			           WHERE f.priority >= 10 AND ( e.is_read = 0 ) AND e.id <= '1763210336616787'
			           ORDER BY e.id DESC
			           LIMIT 201
		           ) e2
		ON e2.id = e0.id
ORDER BY e0.id DESC;
```

An `EXPLAIN` shows the issue:

| id | select\_type | table | partitions | type | possible\_keys | key | key\_len | ref | rows | filtered | Extra |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 1 | PRIMARY | &lt;derived2&gt; | null | ALL | null | null | null | null | 201 | 100 | Using temporary; Using filesort |
| 1 | PRIMARY | e0 | null | eq\_ref | PRIMARY | PRIMARY | 8 | e2.id | 1 | 100 | null |
| 2 | DERIVED | f | null | range | PRIMARY,priority | priority | 1 | null | 96 | 100 | Using where; Using index; Using temporary; Using filesort |
| 2 | DERIVED | e | null | ref | PRIMARY,id\_feed,is\_read,entry\_feed\_read\_index,user\_entry\_id\_feed\_index | id\_feed | 3 | freshrss.f.id | 476 | 50 | Using index condition; Using where |

Its using the wrong index (on the PK), which means it's basically doing a full table scan.

So having exactly 737_141 entries in my table, the query takes (using PhpStorm)
```
100 rows retrieved starting from 1 in 11 s 258 ms (execution: 10 s 796 ms, fetching: 462 ms)
```

Because MySQL is bad at picking the "right" index, this needs to be fixed.

Forcing the better suiting index manually, results in the query

```sql
SELECT
	e0.id
  , e0.guid
  , e0.title
  , e0.author
  , UNCOMPRESS( e0.content_bin ) content
  , e0.link
  , e0.date
  , e0.`lastSeen`
  , e0.`lastUserModified`
  , HEX( e0.hash ) hash
  , e0.is_read
  , e0.is_favorite
  , e0.id_feed
  , e0.tags
  , e0.attributes
FROM
	`user_entry` e0
		INNER JOIN (
			           SELECT
				           e.id
			           FROM
				           `user_entry` e FORCE INDEX (`entry_feed_read_index`)
					           INNER JOIN `user_feed` f
					           ON f.id = e.id_feed
			           WHERE f.priority >= 10 AND ( e.is_read = 0 ) AND e.id <= '1763210336616787'
			           ORDER BY e.id DESC
			           LIMIT 201
		           ) e2
		ON e2.id = e0.id
ORDER BY e0.id DESC;
```

which now isn't even taking a single second.

```
100 rows retrieved starting from 1 in 440 ms (execution: 17 ms, fetching: 423 ms)
```

The `EXPLAIN` shows the resulting query plan:

| id | select\_type | table | partitions | type | possible\_keys | key | key\_len | ref | rows | filtered | Extra |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 1 | PRIMARY | &lt;derived2&gt; | null | ALL | null | null | null | null | 201 | 100 | Using temporary; Using filesort |
| 1 | PRIMARY | e0 | null | eq\_ref | PRIMARY | PRIMARY | 8 | e2.id | 1 | 100 | null |
| 2 | DERIVED | f | null | range | PRIMARY,priority | priority | 1 | null | 96 | 100 | Using where; Using index; Using temporary; Using filesort |
| 2 | DERIVED | e | null | ref | entry\_feed\_read\_index | entry\_feed\_read\_index | 4 | freshrss.f.id,const | 8677 | 33.33 | Using where; Using index |

# Bugfix

`\FreshRSS_EntryDAO::sqlListWhere` needs an update which forces MySQL to pick the right index

```
...
. ' FROM `_entry` e FORCE INDEX (`entry_feed_read_index`) '
...
```

# Result

The front page with the "unread" filter is now instantly available.

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -1443,11 +1443,14 @@ SQL;
 		[$searchValues, $search] = $this->sqlListEntriesWhere(alias: 'e.', state: $state, filters: $filters, id_min: $id_min, id_max: $id_max,
 			sort: $sort, order: $order, continuation_id: $continuation_id, continuation_values: $continuation_values);
 
+		// Help MySQL/MariaDB's optimizer with the query plan:
+		$useEntryIndex = $this->pdo->dbType() === 'mysql' ? 'USE INDEX (entry_feed_read_index) ' : '';
+
 		return [array_merge($values, $searchValues), 'SELECT '
 			. ($type === 'T' ? 'DISTINCT ' : '')
 			. 'e.id'
 			. ($type === 'T' && $sort !== 'id' ? ', ' . $orderBy : '') // SELECT DISTINCT, ORDER BY expressions must appear in SELECT
-			. ' FROM `_entry` e FORCE INDEX (`entry_feed_read_index`) '
+			. ' FROM `_entry` e ' . $useEntryIndex
 			. 'INNER JOIN `_feed` f ON f.id = e.id_feed '
 			. ($sort === 'c.name' ? 'INNER JOIN `_category` c ON c.id = f.category ' : '')
 			. ($type === 't' || $type === 'T' ? 'INNER JOIN `_entrytag` et ON et.id_entry = e.id ' : '')
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -1447,7 +1447,7 @@ SQL;
 			. ($type === 'T' ? 'DISTINCT ' : '')
 			. 'e.id'
 			. ($type === 'T' && $sort !== 'id' ? ', ' . $orderBy : '') // SELECT DISTINCT, ORDER BY expressions must appear in SELECT
-			. ' FROM `_entry` e '
+			. ' FROM `_entry` e FORCE INDEX (`entry_feed_read_index`) '
 			. 'INNER JOIN `_feed` f ON f.id = e.id_feed '
 			. ($sort === 'c.name' ? 'INNER JOIN `_category` c ON c.id = f.category ' : '')
 			. ($type === 't' || $type === 'T' ? 'INNER JOIN `_entrytag` et ON et.id_entry = e.id ' : '')


Alkarex:
I think that would break most databases, so it will need a special case for setups using MySQL.
(MySQL is our least preferred database: PostgreSQL, SQLite, MariaDB, and MySQL as last resort)

Alkarex:
By the way, there are some related performance problems specific to MySQL /  MariaDB, which would welcome some help:
* https://github.com/FreshRSS/FreshRSS/pull/6957
* https://github.com/FreshRSS/FreshRSS/issues/5707#issuecomment-2446505943

Alkarex:
Could you please check whether it is sufficient with `USE INDEX (entry_feed_read_index)` instead of `FORCE` ?

martinhartmann:
Yes you are right, this is a fix specifically for MySQL and probably also MariaDB. I was under the impression that `EntryDAO` is for MySQL/MariaDB because there is also an `EntryDAOPGSQL` and `EntryDAOSQLite`, but that's obviously not the case if I look into the files.

Postgres and SQLite do not support the `USE|FORCE INDEX` syntax, so your fix seems about correct when these two fall back to the default implementation.

I've done some more testing and based on my production dump I can report the following:

Current query in `edge` branch (without `USE|FORCE INDEX`):

- MySQL 5.7: ~ 13s
- MySQL 8.4: ~ 4s
- MySQL 9.4: ~ 4s
- MariaDB 10.6: ~ 5s
- MariaDB 10.11: ~ 5s
- MariaDB 11.8: ~ 400ms
- MariaDB 12.0: ~ 400ms

with `USE INDEX`: ~ 400 ms for every listed RDBMS

So it seems like this fix would directly target MySQL <= 9 and MariaDB < 11. There is no difference when using `USE INDEX` instead of `FORCE INDEX`.

When looking at your posted links, the second one seems to be this exact issue. I've seen some issues on inserting entries in the past, but these all somehow resolved on their own. So for me it's a non-issue. I think it was causing problems, when I've had multiple feeds publishing "lots of" articles. But i cannot give you exact numbers as I don't have this scenario anymore. Currently with 99 feeds subscribed to, I don't have any problems with updating at all.

I must admit I'm still using MySQL 5.7, because my deployment is from 2018. I probably should upgrade my database in the near future... But as the issue still persists with MySQL 9.4 and even applies to MariaDB < 11, this would probably also help someone else, as FreshRSS currently [targets](https://freshrss.github.io/FreshRSS/en/admins/02_Prerequisites.html) MySQL 8.0+ and MariaDB 10.0.5+.

Alkarex:
Thanks for the additional tests 👍🏻
Tests of https://github.com/FreshRSS/FreshRSS/pull/6957 welcome

Alkarex:
* https://github.com/FreshRSS/FreshRSS/pull/8460

Alkarex:
Help welcome to test with MariaDB the following index change (and whether we could remove the `USE INDEX`):
* https://github.com/FreshRSS/FreshRSS/pull/8523

PR #8220, opened 2025-11-16
Rework saving of configuration files

### LINKED ISSUE #8022: [Bug] config.php gets overwritten with 0 bytes
### Describe the bug

Hi there,

i just updated to [1.27.1](https://github.com/FreshRSS/FreshRSS/releases/tag/1.27.1) and i now see:

Image

### To Reproduce

1. Go to 'https://fresh.mydomain.eu/i/?c=auth&a=login'

### Expected behavior

_No response_

### FreshRSS version

0

### System information

- Database version: SQLite
- Installation type: Docker ( `freshrss/freshrss` )
- OS: Ubuntu 24.04.3 LTS

### Additional context

in config.php:
- `'auth_type' => 'form',`

Please advise, thank you.

Inverle:
fix https://github.com/FreshRSS/FreshRSS/issues/8022

1. Execute `dd if=/dev/urandom of=space bs=1M status=progress` and wait until disk is full
2. Try to save the config
3. Instead of it getting overwritten with 0 bytes, the file won't be touched.

### DIFF
--- a/lib/Minz/Configuration.php
+++ b/lib/Minz/Configuration.php
@@ -210,18 +210,21 @@ class Minz_Configuration {
 	 * Save the current configuration in the configuration file.
 	 */
 	public function save(): bool {
+		$tmp_filename = $this->config_filename . '.tmp';
 		$back_filename = $this->config_filename . '.bak.php';
-		@rename($this->config_filename, $back_filename);
 
-		if (file_put_contents($this->config_filename,
+		if (file_put_contents($tmp_filename,
 			"<?php\nreturn " . var_export($this->data, true) . ';', LOCK_EX) === false) {
-			// file_put_contents failed, attempt to restore the backup config
-			if (@copy($back_filename, $this->config_filename) === false) {
-				// to help with second restore process during user context init
-				@unlink($this->config_filename);
-			}
+			@unlink($tmp_filename);
 			return false;
 		}
+
+		if (copy($this->config_filename, $back_filename) === false) {
+			@unlink($tmp_filename);
+			return false;
+		}
+
+		@rename($tmp_filename, $this->config_filename);
 
 		// Clear PHP cache for include
 		if (function_exists('opcache_invalidate')) {
--- a/lib/Minz/Configuration.php
+++ b/lib/Minz/Configuration.php
@@ -213,13 +213,13 @@ class Minz_Configuration {
 		$tmp_filename = $this->config_filename . '.tmp.php';
 		$back_filename = $this->config_filename . '.bak.php';
 
-		if (file_put_contents($tmp_filename,
-			"<?php\nreturn " . var_export($this->data, true) . ';', LOCK_EX) === false) {
+		if (!file_put_contents($tmp_filename,
+"<?php\nreturn " . var_export($this->data, true) . ';', LOCK_EX)) {
 			@unlink($tmp_filename);
 			return false;
 		}
 
-		if (copy($this->config_filename, $back_filename) === false) {
+		if (!copy($this->config_filename, $back_filename)) {
 			@unlink($tmp_filename);
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


lib/Minz/Configuration.php
```diff
@@ -215,6 +215,11 @@ public function save(): bool {
 
 		if (file_put_contents($this->config_filename,
 			"<?php\nreturn " . var_export($this->data, true) . ';', LOCK_EX) === false) {
+			// file_put_contents failed, attempt to restore the backup config
+			if (@copy($back_filename, $this->config_filename) === false) {
```
Inverle: What should be done if the backup is corrupted too?
https://github.com/FreshRSS/FreshRSS/issues/8022#issuecomment-3341920404

(the same happened for me as well, but I don't understand how)

Alkarex:
I think the logic should be reworked. Something like:
1. Save the new configuration in a temporary file within the same folder
2. only if it worked, copy the current configuration as a backup in the same folder
3. only if it worked, atomic rename of the new configuration to the current configuration

Inverle:
@Alkarex It seems to work better: 750802ccfd3feca5617f4c642891b2eec93a13ab

lib/Minz/Configuration.php
```diff
@@ -210,14 +210,22 @@ public function __set(string $key, mixed $value): void {
 	 * Save the current configuration in the configuration file.
 	 */
 	public function save(): bool {
+		$tmp_filename = $this->config_filename . '.tmp';
```
Alkarex:
```suggestion
		$tmp_filename = $this->config_filename . '.tmp.php';
```

Minor preference

lib/Minz/Configuration.php
```diff
@@ -210,14 +210,22 @@ public function __set(string $key, mixed $value): void {
...
 		$back_filename = $this->config_filename . '.bak.php';
-		@rename($this->config_filename, $back_filename);
 
-		if (file_put_contents($this->config_filename,
+		if (file_put_contents($tmp_filename,
```
Alkarex: Missing a negation maybe?

Inverle: There is a `=== false` at the end

Alkarex: Ah, sorry, at the end

Inverle:
By the way there are many places which call `FreshRSS_Context::userConf()->save()` and don't check the return value to show an error message instead of a successful update message.

lib/Minz/Configuration.php
```diff
@@ -210,14 +210,22 @@ public function __set(string $key, mixed $value): void {
...
 			"<?php\nreturn " . var_export($this->data, true) . ';', LOCK_EX) === false) {
```
Alkarex:
```suggestion
		if (!file_put_contents($tmp_filename,
"<?php\nreturn " . var_export($this->data, true) . ';', LOCK_EX)) {
```

I was confused by the previous syntax, and a return of (int)0 should also be accounted as an error

Inverle: Does that ever happen though?

lib/Minz/Configuration.php
```diff
@@ -210,14 +210,22 @@ public function __set(string $key, mixed $value): void {
...
+			@unlink($tmp_filename);
 			return false;
 		}
 
+		if (copy($this->config_filename, $back_filename) === false) {
```
Alkarex:
```suggestion
		if (!copy($this->config_filename, $back_filename)) {
```

At least for natively boolean expressions, let's stick to the boolean syntax

lib/Minz/Configuration.php
Inverle: > This function returns the number of bytes that were written to the file, or [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) on failure.

never mind

Alkarex (approved):

PR #7959, opened 2025-09-14
Add option to apply filter actions to existing articles

raspo:
This PR adds a new checkbox to the `Filter actions` settings that lets the user apply filters retroactively; in other words: mark as read existing articles, not just new incoming ones.

CleanShot 2025-09-13 at 18 15 34@2x

**Background:**
I was trying to use the `Filter actions` to filter away certain type of articles that I never want to see in my feed (for example: sponsors or articles reposting the new podcast episode, etc).. but I was very confused by the fact that no matter what filter I'd try, existing articles would never get marked as read automatically. Eventually I discovered that this is intended behavior (as mentioned [here](https://github.com/FreshRSS/FreshRSS/issues/3152#issuecomment-769884867) and [here](https://github.com/FreshRSS/FreshRSS/issues/2941#issuecomment-623097799)).

Changes proposed in this pull request:

- Add new `Also apply filters to existing articles` checkbox
- Call `applyFiltersRetroactively()` when the checkbox has been checked
- `applyFiltersRetroactively` automatically marks as read each entry that is matching the filter

How to test the feature manually:

1. Select a feed from the sidebar, click the ⚙️ icon, then `Manage`
2. In the `Filter actions` section, add a filtering rule inside the `Mark an article as read…` field
3. Hit submit
4. Settings should be saved correctly
5. Open the feed settings again, but this time check the `Also apply filters to existing articles` checkbox
6. Hit submit again
7. Settings should be saved correctly and the filters should have been applied to existing articles (marking them read)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

PS:
Thank you so much for your work on FreshRSS, I have been using it for many months and it has never needed any maintenance. I haven't written PHP in over 10 years, so I probably made a bunch of mistakes. For the translations I only added one string into the english one and then ran `make i18n-format`, I hope that's ok.

PR #8222, opened 2025-11-16
Fix HTML encoding breaking search in preview filters

Inverle:
Follow-up of #7959

---

image

wrongly resulted in

image


Alkarex:
This PR does not work at the moment. Quick test with `intitle:" & "` which expects to filter ` & ` (with space on each side) in the title, and which does not work anymore. Since the title is saved in database as HTML, the search input must also be as HTML, i.e. ` &amp; `

Alkarex:
I think the place to fix instead is:
https://github.com/FreshRSS/FreshRSS/blob/a5c4f99b807b2f61c51ea25d4288257d0488fe7c/app/Controllers/categoryController.php#L183-L188

By building a BooleanSearch instead of only manual strings:
https://github.com/FreshRSS/FreshRSS/blob/a5c4f99b807b2f61c51ea25d4288257d0488fe7c/app/Models/BooleanSearch.php#L429-L432

Alternatively, inspired by the following (which could likely be improved):
https://github.com/FreshRSS/FreshRSS/blob/8fdc574827c668f098995891eac41f0688b842b6/app/Models/BooleanSearch.php#L29-L38

Alkarex:
Alternative in https://github.com/FreshRSS/FreshRSS/pull/8259
Tests welcome

PR #8223, opened 2025-11-16
Fix MySQL commitNewEntries

Alkarex:
I just realised that `commitNewEntries()` was not sorting articles properly before insertion in database when using MySQL: Articles were not sorted by publication date as expected from the temporary table before insertion in the final table. MySQL was not picking the correct field, so fixed with an explicit alias.
Discovered because I did some tests with MySQL in https://github.com/FreshRSS/FreshRSS/pull/6957
At the same time, I did the same change for PostgreSQL and SQLite although those were not affected.

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -248,18 +248,18 @@ SQL;
 
 	public function commitNewEntries(): bool {
 		$sql = <<<'SQL'
-SET @rank=(SELECT MAX(id) - COUNT(*) FROM `_entrytmp`);
+			SET @rank=(SELECT MAX(id) - COUNT(*) FROM `_entrytmp`);
 
-INSERT IGNORE INTO `_entry` (
-	id, guid, title, author, content_bin, link, date, `lastSeen`,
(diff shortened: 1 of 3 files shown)


app/Models/EntryDAO.php
```diff
@@ -248,18 +248,18 @@ public function addEntry(array $valuesTmp, bool $useTmpTable = true): bool {
...
-SQL;
+			SET @rank=(SELECT MAX(id) - COUNT(*) FROM `_entrytmp`);
+
+			INSERT IGNORE INTO `_entry` (
+				id, guid, title, author, content_bin, link, date, `lastSeen`,
+				hash, is_read, is_favorite, id_feed, tags, attributes
+			)
+			SELECT @rank:=@rank+1 AS id, guid, title, author, content_bin, link, date, `lastSeen`, hash, is_read, is_favorite, id_feed, tags, attributes
+			FROM `_entrytmp` etmp
+			ORDER BY etmp.date, etmp.id;
```
Alkarex, line 259: Here is the fix, with explicit alias

PR #8230, opened 2025-11-18
More visible selected article

Alkarex:
There was not enough difference between selected and not-selected articles, especially not due to hovering using the same style.
Already relevant as is, and more relevant for a future feature depending on selected article.
Additional help welcome to further improve the styling, in this PR or a future one, also for the other themes.

In this example, the third article is the one selected:

image


Alkarex:
Ping @math-GH if you are around

math-GH:
Would be a better visible background (higher contrast/other color) another option?

Playing around with the borders shifts the content and would lead to jumping elements.

Alkarex:
> Would be a better visible background (higher contrast/other color) another option?

Yes for sure, other options could be fine. I am not sure more contrast or colour change would be enough though.

> Playing around with the borders shifts the content and would lead to jumping elements.

I quite like the visual indicator on the side. It can surely be made not to modify the alignments. Although I have to admit that even if not intended, and even if it is likely better to remove them, I find the 3 pixels of offset to provide an additional visual clue for the selected article, which I do not dislike.

Alkarex:
Now there is no offset anymore:

image

Alkarex:
Let's merge https://github.com/FreshRSS/FreshRSS/pull/8241 first

PR #8241, opened 2025-11-22
Use native CSS instead of SCSS

### LINKED ISSUE #7611: [Feature] Get rid of SCSS - just use plain CSS

Alkarex:
I was tired of having to handle the SCSS themes differently.
Ansum/Mapco SCSS -> CSS & remove sass dependencies
fix https://github.com/FreshRSS/FreshRSS/issues/7611
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8200
The changes are actually quite limited (mostly changing the syntax of the variables and a few other things)
Require support for:
* https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Nesting
* https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector
* https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix

Tests welcome.
The styles are left as individual CSS files as designed originally in SCSS. I expect proper HTTP caching and HTTP/2 to be sufficient instead of aggregating back to a single CSS, but might be done in another PR if needed.

At the same time, fixed the fact that we had various whitespace styles in our JSON files, in particular in those themes, sometimes with 2 spaces, or 4 spaces, or tabs. Use tabs everywhere as defined in editorconfig.

I would like to merge this one first to complete https://github.com/FreshRSS/FreshRSS/pull/8230


package.json
```diff
@@ -1,50 +1,52 @@
...
+	},
+	"scripts": {
+		"eslint": "eslint .",
+		"eslint_fix": "eslint --fix .",
+		"markdownlint": "markdownlint '**/*.md'",
+		"markdownlint_fix": "markdownlint --fix '**/*.md'",
+		"rtlcss": "npm run rtlcss:run && npm run rtlcss:clean && npm run rtlcss:imports",
+		"rtlcss:run": "rtlcss -d p/themes/",
+		"rtlcss:clean": "find p/themes/ -type f -name '*.rtl.rtl.css' -delete",
+		"rtlcss:imports": "find p/themes -type f -name '*.rtl.css' -print0 | xargs -0 sed -i -E '/\\.rtl\\.css/! s|^(@import \"[^\"]+)(\\.css\";)$|\\1.rtl\\2|'",
```
Alkarex: This part is modified to rewrite the CSS `@import` URLs to RTL.

Inverle:
> The styles are left as individual CSS files as designed originally in SCSS. I expect proper HTTP caching and HTTP/2 to be sufficient instead of aggregating back to a single CSS, but might be done in another PR if needed.

I'm not sure if I even like the CSS being split between files in the first place. I prefer sections split with comments, like in frss.css

> At the same time, fixed the fact that we had various whitespace styles in our JSON files, in particular in those themes, sometimes with 2 spaces, or 4 spaces, or tabs. Use tabs everywhere as defined in editorconfig.

I was annoyed by the spaces instead of tabs in composer.json, so this is a nice change

package.json
Inverle: Is it possible to do this differently? 🤔

I found https://github.com/MohammadYounes/rtlcss/issues/108#issuecomment-347033134 but not sure how relevant it is

Alkarex: Ah, I tried a bit `stringMap` without success but not the approach you found.
Could be better indeed. Help welcome, or I will come back to it later tonight.

p/themes/Ansum/metadata.json
```diff
@@ -1,8 +1,8 @@
...
-  "name": "Ansum",
-  "author": "Thomas Guesnon",
-  "description": "Thème sablonneux",
-  "version": 0.1,
-  "files": ["_frss.css", "ansum.css"],
-  "theme-color": "#fbf9f6"
+	"name": "Ansum",
+	"author": "Thomas Guesnon",
+	"description": "Thème sablonneux",
+	"version": 0.2,
```
Inverle, line 5: I forget that themes have a version assigned to them, there were probably many theme changes without a version bump.

The versioning itself is also inconsistent, for example this has **0**.2 set as the version, but Swage is **1**.0 (why is the major version not 1?)
Even Origine is 0.2

Alkarex:
> I'm not sure if I even like the CSS being split between files in the first place

Yes, I do not have a strong stance on the topic, but I like the idea of making as few changes as possible in this PR to focus on the migration from SCSS to CSS and stay close to the original author's intention.

p/themes/Ansum/metadata.json
Inverle: This is a nitpick of course, but also note that some themes have a French description and some an English description.

Alkarex: Indeed, and not all themes have a version though. This should be cleaned. But best in another PR and help welcome.

Frenzie:
> I'm not sure if I even like the CSS being split between files in the first place. I prefer sections split with comments, like in frss.css

It can be nice for some purposes, but agreed, this isn't one of them.

package.json
Alkarex: Seems to work fine https://github.com/FreshRSS/FreshRSS/pull/8241/commits/854d9894d2b810a25e6075811e7785de674a3fe1 👍🏻

Alkarex: We now lack a more elegant approach to avoid `find p/themes/ -type f -name '*.rtl.rtl.css' -delete`

p/themes/Ansum/metadata.json
Inverle: After this one is merged though because there would be merge conflicts otherwise

Inverle (approved):

PR #8246, opened 2025-11-24
SimplePie fix merge HTTP headers

Alkarex:
HTTP headers were not merged properly, and were also not accounting for possible casing difference.
See https://github.com/FreshRSS/FreshRSS/issues/8189#issuecomment-3569434305

Upstream PR:
* https://github.com/simplepie/simplepie/pull/956

### DIFF
--- a/lib/simplepie/simplepie/src/File.php
+++ b/lib/simplepie/simplepie/src/File.php
@@ -113,7 +113,7 @@ class File implements Response
                 $fp = curl_init();
                 if (is_array($curl_options[CURLOPT_HTTPHEADER] ?? null)) {
                     // Save the original casing of HTTP headers
-                    $headers_keys = array_combine(array_map('strtolower', array_keys($headers)), array_keys($headers));
+                    $headers_keys = array_combine(array_map('strtolower', array_keys($headers)), array_keys($headers)) ?: [];
                     // Merge HTTP headers case-insensitively from cURL options with the ones from the dedicated parameter
                     $headers = array_change_key_case($headers, CASE_LOWER);
                     foreach ($curl_options[CURLOPT_HTTPHEADER] as $header) {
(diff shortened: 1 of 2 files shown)


lib/simplepie/simplepie/src/File.php
```diff
@@ -111,15 +111,33 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
             if (!$force_fsockopen && function_exists('curl_exec')) {
                 $this->method = \SimplePie\SimplePie::FILE_SOURCE_REMOTE | \SimplePie\SimplePie::FILE_SOURCE_CURL;
                 $fp = curl_init();
-                $headers2 = [];
-                foreach ($headers as $key => $value) {
-                    $headers2[] = "$key: $value";
-                }
-                if (isset($curl_options[CURLOPT_HTTPHEADER])) {
-                    if (is_array($curl_options[CURLOPT_HTTPHEADER])) {
-                        $headers2 = array_merge($headers2, $curl_options[CURLOPT_HTTPHEADER]);
```
Alkarex, line 120: Could not work because those are arrays of strings and not arrays of key-values.

lib/simplepie/simplepie/src/File.php
```diff
@@ -111,15 +111,33 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
             if (!$force_fsockopen && function_exists('curl_exec')) {
                 $this->method = \SimplePie\SimplePie::FILE_SOURCE_REMOTE | \SimplePie\SimplePie::FILE_SOURCE_CURL;
                 $fp = curl_init();
-                $headers2 = [];
```
Alkarex, line 114: Renamed as `$headers_inlined` below

lib/simplepie/simplepie/src/File.php
```diff
@@ -111,15 +111,33 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
...
+                            $parts = explode(':', $header, 2);
+                            if (count($parts) >= 2) {
+                                $key = trim($parts[0]);
+                                $headers_keys[strtolower($key)] = $key;
+                                $headers[strtolower($key)] = trim($parts[1]);
+                            }
+                        }
                     }
-                    unset($curl_options[CURLOPT_HTTPHEADER]);
+                    // Restore original casing of HTTP headers
```
Alkarex, line 129: We could send everything lowercased as required by HTTP/2, but better to keep original casing to avoid being discriminated on that when using HTTP/1.1.

Alkarex:
I need to rethink that. There are valid reasons for multiple identical HTTP headers such as:
```
Set-Cookie: a=1
Set-Cookie: b=2
```

PR #8247, opened 2025-11-25
Improve layout of documentation page and add search feature

### LINKED ISSUE #5325: [BUG] Documentation center: Jekyll warning
`Deprecation: The 'include' configuration option must now be specified as an array, but you specified a string. For now, we've treated the string you provided as a list of comma-separated values`

grafik

Who could fix it?

### LINKED ISSUE #7915: [Feature] Better documentation navigation

I wish we would have a better navigation in our documentation tool/help center.
Something like a navigation bar would help to navigation back and forth and would make it simpler do walk around.

### Use case

_No response_

left hand side: having a navigation through the chapters and pages

_No response_

### Additional context

_No response_

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/7915, https://github.com/FreshRSS/FreshRSS/issues/5325

Demo: https://inverle.github.io/FreshRSS/


docs/Gemfile
Inverle, line 1: Needed for proper working of local build

docs/_config.yml
```diff
@@ -1,9 +1,19 @@
...
 description: Documentation center
 
 logo: /img/FreshRSS-logo.png
-show_downloads: true
 
 include: [contributing.md]
-exclude: [CHANGELOG*.md]
+exclude: [CHANGELOG*.md, README.md]
+
+plugins:
```
Inverle, line 40: Needed for proper working of local build

Frenzie:
You're probably aware, but the mobile styles seem to kick in too late (should be at something like 1200 already, also compare [Bootstrap](https://getbootstrap.com/docs/5.3/getting-started/introduction/)).

<kbd>image</kbd>

docs/assets/css/style.css
```diff
@@ -0,0 +1,137 @@
+@import url('https://fonts.googleapis.com/css2?family=Inconsolata:wght@200..900&display=swap');
```
Inverle: Still need to improve, because a different font is used before this one is loaded and that makes page loads look weird

Inverle:
Please review again

Alkarex:
Nice work, looks good to me 👍🏻 
My main feedback would be on internationalisation: the root seems to auto-redirect to `/en/`. It would be nice to be able to see that there is both English and French.
And a similar comment about search, which seems to search in both languages, which is not necessary a problem (it might even be good). If possible, just adding a horizontal separator in the search results between the languages could be enough:

image

Alkarex (approved):

docs/index.md
```diff
@@ -4,3 +4,6 @@ If you want to contribute, you can [find us on GitHub](https://github.com/FreshR
 
 - [English documentation](./en/index.md)
 - [Documentation française](./fr/index.md)
+
+<!-- todo: french docs -->
+<meta http-equiv="Refresh" content="0; url=./en/">
```
Frenzie, line 12: You're probably aware, but you can match against navigator.languages. Something in this general direction.

```js
const supportedLanguages = ['en', 'fr'];
// supported for a decade, probably not worth checking anything besides a try/catch just in case
for (const lang of navigator.languages) {
 const normalizedLang = lang.toLowerCase();
 const matched = supportedLanguages.find(supported =>
  supported.toLowerCase() === normalizedLang ||
  normalizedLang.startsWith(supported + '-')
 );
 location.href = ...
 break; // script can sometimes keep running for a couple hundred ms more; it'd be weird to redirect to the second choice
}
```

I wonder if the refresh `<meta>` works in a `<noscript>`?

Inverle:
I have a WIP language dropdown

docs/index.md
Inverle: This will be replaced by a language dropdown instead, it is probably not a good idea to auto redirect to French, since it's not as up to date as the English documentation

Inverle:
@Alkarex See https://inverle.github.io/FreshRSS/

docs/_config.yml
```diff
@@ -1,9 +1,53 @@
...
+    choose_language: Choose language
+    copy_to_clipboard: Copy to clipboard
+  fr:
+    back_to_freshrss: Retour à freshrss.org
+    search_docs: Rechercher dans la documentation…
+    search: Rechercher
+    toggle_aside: Afficher/masquer le panneau
+    close: Fermer
+    choose_language: Choisir la langue
+    copy_to_clipboard: Copier dans le presse-papiers
```
Inverle, line 38: This is translated with ChatGPT so correct if needed

docs/_includes/docs_nav.html
```diff
@@ -0,0 +1,77 @@
...
+		</li>
+		<li>
+			<a href="{{ '/en/contributing.html' | relative_url}}">Contributor guidelines</a>
+		</li>
+	</ul>
+{% elsif page.lang == 'fr' %}
+	<ul>
+		<li><a href="{{ '/fr/' | relative_url }}">Accueil</a></li>
+		<li>
+			<a href="{{ '/fr/users/02_First_steps.html' | relative_url }}">Documentation utilisateur</a>
```
Inverle, line 42: Same here and below

Inverle:
What should be done here? (typos CI fail)
```
error: `contributeurs` should be `contributes`
  --> ./docs/_includes/docs_nav.html:73:78
   |
73 |    <a href="{{ '/fr/contributing.html' | relative_url}}">Directives pour les contributeurs</a>
   |                                                                              ^^^^^^^^^^^^^
   |
```

docs/_config.yml
Frenzie: It's correct according to https://fr.techdico.com/traduction/anglais-francais/copy+to+clipboard.html

Frenzie:
> (typos CI fail)

Just yesterday I named something `anc_whatever` (for ancestor_whatever) and a similar system complained it should be `and`. I caved and just called it `ancestor_whatever` since there was an easy solution. :-)

As for the question I'd say just exclude all of French? Some kind of English spellcheck seems extremely unlikely to be useful there. ;-)

math-GH:
I like it to have the navigation on the left hand side.

But I am not a fan of this general layout. It looks very nerdy and difficult to read. Is there any chance of a more conservative style?

Inverle:
> But I am not a fan of this general layout. It looks very nerdy and difficult to read. Is there any chance of a more conservative style?

Difficult to read because of the font? I thought about changing it, but not sure to which one.
As for the layout, I'm also open to suggestions.

math-GH:
That are the fonts of Origine theme: "OpenSans", "Cantarell", "Helvetica", "Arial"

docs/_config.yml
Alkarex: Looks good

.github/workflows/jekyll-gh-pages.yml
```diff
@@ -31,10 +31,11 @@ jobs:
       - name: Setup Pages
         uses: actions/configure-pages@v5
       - name: Build with Jekyll
-        uses: actions/jekyll-build-pages@v1
-        with:
-          source: ./docs/
-          destination: ./_site
+        run: |
+          cd docs
+          sudo gem install bundler
```
Alkarex: I have not looked in details, but `sudo` looks strange to me there. Might be installable locally instead of globally?

docs/README.md
```diff
@@ -0,0 +1,13 @@
+This is the documentation deployed by GitHub Pages.
+
+To build and serve locally, first install the necessary packages:
+```console
+bundle install
+```
+
+Then serve:
+```console
+bundle exec jekyll serve -H 0.0.0.0 --watch --incremental
```
Alkarex:
```suggestion
bundle exec jekyll serve -H 127.0.0.1 --watch --incremental
```

A localhost IP is likely sufficient (and safer) for a local test?

docs/README.md
```diff
@@ -0,0 +1,13 @@
+This is the documentation deployed by GitHub Pages.
+
+To build and serve locally, first install the necessary packages:
+```console
```
Alkarex:
```suggestion
```sh
```

I suggest using `sh` for such commands. `console` is meant to be used with a console prefix such as `$` to distinguish between commands and outputs.

Example with `console` (and without the expected prefix) in VSCode:
Image

Example with `sh` in VSCode:
Image

docs/README.md
```diff
@@ -0,0 +1,13 @@
...
+bundle install
+```
+
+Then serve:
+```console
+bundle exec jekyll serve -H 0.0.0.0 --watch --incremental
+```
+
+The documentation should be reachable at `http://0.0.0.0:4000/FreshRSS/`.
```
Alkarex:
```suggestion
The documentation should be reachable at <http://127.0.0.1:4000/FreshRSS/>.
```

Suggest standard markdown syntax for links

.github/workflows/jekyll-gh-pages.yml
Inverle: Yes, there is an option for that I think, but doesn't make a difference here.

docs/README.md
Inverle: Good to know, thanks.

docs/README.md
Inverle: I prefer 0.0.0.0, but sure.

.github/workflows/jekyll-gh-pages.yml
Inverle: Though I probably should mention that this step increases the build time in CI from 19s to 59s.

https://github.com/FreshRSS/FreshRSS/actions/runs/19749114713/job/56588969757
vs
https://github.com/Inverle/FreshRSS/actions/runs/19775210787/job/56666596731

Alkarex: Any chance of using caching some more?
`gem install bundler` should not run every time

Inverle: I just found https://github.com/actions/jekyll-build-pages/issues/46#issuecomment-1215505246, will look into improving this. (you're correct about that)

Inverle: Okay it now takes 35s to build when gems are already cached:

https://github.com/Inverle/FreshRSS/actions/runs/19804796794/job/56737317095
https://github.com/Inverle/FreshRSS/actions/caches

Inverle: It's still a long build time, but I think the build got slower after I added either the `jekyll-i18n_tags` plugin or

https://github.com/FreshRSS/FreshRSS/blob/6e87494b4da378287d169e550b9211eeb0f2284c/docs/_config.yml#L10-L20

which it required

Alkarex: It would be good to skip the build altogether when the content of `./docs/` has not changed (e.g. based on a hash), which is the case for most PRs / commits

Inverle: Yes, there is a way to do that too. And I thought about it even before this PR.

Inverle: Done: https://github.com/FreshRSS/FreshRSS/pull/8247/commits/df30aa20d9db02cc40acfd321323b34e6c49e4e9

Inverle: Some changes may be required to the repository settings after this PR is merged, but I'm not sure. (needs testing after merge)

Inverle:
I changed the font to Open Sans

Alkarex:
Minor comment: I think line height should be slightly increased:

image

Inverle:
@Alkarex done

math-GH:
Kudos for this PR!

Inverle:
Looks like there was already a sidebar before 👀 
https://web.archive.org/web/20170714232519/http://doc.freshrss.org/en/

PR #8254, opened 2025-11-26
Fix MariaDB for updates

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/8252
regression from https://github.com/FreshRSS/FreshRSS/pull/6957
The current code works with MySQL 8+ (as well as PostgreSQL and SQLite), but not with MariaDB, because MariaDB does not support CTE (Common Table Expression) fully (which I was not aware of)...

PR #8255, opened 2025-11-26
Fix MariaDB updateCachedValues

Alkarex:
Changed syntax for an even wider compatibility.
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8254
Regression from https://github.com/FreshRSS/FreshRSS/pull/6957
MariaDB images on Docker Hub go back to 10.6, so changed documentation to indicate support from that version, as I cannot easily test even older versions.


WillPresley:
Testing the branch now, will follow up shortly.

Edit: Manually refreshing seems to still work as expected and no errors in the logs. Will let the automatic updates run a few times in the background before checking back in.

WillPresley (commented):
As far as I can tell, this is working great. I don't have a way to test MariaDB or specific versions, however.

Alwaysin:
I can confirm that I have no error anymore with this patch, and the banner came back! Thank you!

PR #8263, opened 2025-11-29
SimplePie: Disallow `javascript:` URI protocol

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/7924

See https://portswigger.net/web-security/cross-site-scripting/cheat-sheet#protocols


lib/simplepie/simplepie/src/SimplePie.php
```diff
@@ -692,6 +692,13 @@ class SimplePie
      */
     public $allow_aria_attr = true;
 
+    /**
+     * @var string[] Stores array of disallowed URI protocols
+     * @see SimplePie::disallow_uri_protocols()
+     * @access private
+     */
+    public $disallowed_uri_protocols = ['javascript'];
```
Alkarex: I have not though about it much yet, but should it maybe be the opposite, with a list of allowed protocols?

lib/simplepie/simplepie/src/Sanitize.php
```diff
@@ -541,6 +551,12 @@ public function sanitize(string $data, int $type, string $base = '')
                     $this->replace_urls($document, $element, $attributes);
                 }
 
+                if ($this->disallowed_uri_protocols) {
+                    foreach ($this->disallowed_uri_protocols as $protocol) {
+                        $this->strip_uri_protocol($xpath, $protocol);
```
Alkarex: I have not looked into it yet, but I hope there is a better place to call this logic from, instead of having to search via XPath again. Probably in the existing URL processing e.g. for relative URLs, sanitising URLs, HTTPS-enforcing.

lib/simplepie/simplepie/src/SimplePie.php
Frenzie: That sounds inconvenient. Wikipedia contains a very incomplete list of [some custom ones](https://en.wikipedia.org/wiki/List_of_URI_schemes#Unofficial_but_common_URI_schemes) (otoh most notably it's missing gemini, irc and magnet) but countless apps depend on locally register custom protocol schemes locally on the fly.

Random example: https://wails.io/docs/guides/custom-protocol-schemes/

Edit: A quick `grep -r "x-scheme-handler" /usr/share/applications ~/.local/share/applications` also shows `itch:`, `steam:`, `steamlink:`, `apt:`, `goldendict:`, `telnet:`, `ssh:`, etc. And what I have installed and what's used in my feeds will be completely different for another person. :-)

lib/simplepie/simplepie/src/Sanitize.php
Alkarex: Maybe:
https://github.com/FreshRSS/FreshRSS/blob/78e40c6fe3afe7f815ef9d32646610e2d5436ba3/lib/simplepie/simplepie/src/Sanitize.php#L647

Alkarex: Refactored in https://github.com/FreshRSS/simplepie/pull/65/commits/dc32e351120484c653fc77b7e1da3de52575b99f
Feedback welcome

Inverle:
Test feed:
```xml
<rss version="2.0" xmlns:content="http://purl.org/rss/1.0/modules/content/">
  <channel>
    <title>Malicious Feed</title>
    <link>http://example.com</link>
    <item>
      <title>Click to read article</title>
      <link>javascript:alert(1)</link>
      <description>Interesting article content</description>
      <content:encoded><![CDATA[
          <a href="javascript:alert(2)">test1</a>
          <iframe src="javascript:alert(3)">test2</iframe>
          <math href="javascript:alert(4)">test3</math>
          <math><maction href="javascript:alert(5)">test4</maction></math>
        ]]></content:encoded>
    </item>
  </channel>
</rss>
```
HTML is sanitized properly, but the URL inside `<link>` is not.

Inverle:
https://github.com/FreshRSS/simplepie/pull/80

Alkarex (approved):

PR #8266, opened 2025-11-30
Web scraping support date format epoch milliseconds

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/discussions/8264
Auto detect whether a `U` date format should be in seconds or milliseconds.


Frenzie (approved):
Nice idea

PR #8277, opened 2025-12-01
Scaling of user statistics

### LINKED ISSUE #8268: [Bug] User Management is too slow
### Describe the bug

Hello. I am the administrator of the instance https://rss.cheredeprince.net. I am proud to say that we are more than 1K active users. 
My problem is that the user management is too slow. For instance, the page "Manage Users" reaches my reverse proxy time out and the command line `cli/user-info.php --json` takes ages (10 minutes) to run. 

Is there something you can do about that ?

### To Reproduce

1. create a large instance with 1K users 
2. run the script cli/user-info.php --json 

### Expected behavior

_No response_

### FreshRSS version

1.27.2-dev

### System information

- Database version: Postgresql
- PHP version: 8.1
- Installation type: git
- Web server type: nginx
- OS: Debian 12

### Additional context

_No response_

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/8268
To better support user management on FreshRSS instance with many users.

SQL speed improved. On a reduced test with 5 users, including some large accounts (PostgreSQL on a very tiny and slow server), improving from ~2.3s to ~1.8s, which gives ~20% speed improvement.

Then tested with 1000 users, with only the default feed (on my old desktop computer):

```sh
for i in {1..1000}; do ./cli/create-user.php --user=freshrss$i --password=freshrss; done
app/actualize_script.php
cli/access-permissions.sh
```

SQLite:

```console
$ time cli/user-info.php | wc -l
1001

real    0m1.366s
user    0m0.908s
sys     0m0.475s
```

PostgreSQL:

```console
$ time cli/user-info.php | wc -l
1001

real    0m28.498s
user    0m12.137s
sys     0m2.217s
```

MariaDB:

```console
# time ./cli/user-info.php | wc -l
1001

real    0m49.485s
user    0m1.276s
sys     0m2.258s
```

Yes, SQLite is much faster - not a surprise for such use-cases, where the TCP connection is not re-used.

I have added some CLI options to disable some statistics:

```sh
cli/user-info.php --no-db-size --no-db-counts
```

For the Web UI, I have disabled detailed user statistics if it takes too long, and retrieve missing user statistics asynchronously via JavaScript. Lazy loading of the user details based on IntersectionObserver, with maximum 10 requests in parallel.
Web UI tested on 1000 users as well. Checked with SeaMonkey.


Alkarex:
I remember that @marienfressinaud also mentioned having faced a similar need

Alkarex:
Alright, for 1000 users, this seems to give a massive improvement for all databases, in particular in the Web UI.

Alkarex:
I have added performance statistics higher up

marienfressinaud:
> I remember that @marienfressinaud also mentioned having faced a similar need

Yes indeed! I removed a bunch of inactive users since then so it wasn't a big issue anymore. I'll try to test your patch in the week, thanks :)

marienfressinaud:
I've just tested it and it works fine. It now takes ~2.5 seconds to load 60 accounts, compared to > 5 seconds previously. The user statistics are loaded synchronously as the request doesn't take too much time.

Alkarex:
> I've just tested it and it works fine. It now takes ~2.5 seconds to load 60 accounts, compared to > 5 seconds previously. The user statistics are loaded synchronously as the request doesn't take too much time.

Thanks for the feedback :-)

Issue #6831, opened 2024-09-24
Take advantage of PHP 8.4+ native HTML5 support


Alkarex:
https://wiki.php.net/rfc/domdocument_html5_parser
https://wiki.php.net/rfc/opt_in_dom_spec_compliance

Test for availability of the new HTML functions such as `\Dom\HTMLDocument::createFromString()` and use them if possible, keeping the old code while we still support PHP 8.3-

Needed for fixing issues such as https://github.com/FreshRSS/FreshRSS/issues/5201

PR #8279, opened 2025-12-02
Fix encoding being corrupted in non-ASCII HTML feeds

ostrolucky:
Even though when fetching the feed all characters are correcly encoded, they are being maligned when passing it to DOMDocument, since it incorrectly detects the content as CP-1252 encoding.

Can be tested on eg. https://www.npc.sk/aktuality

1. Add https://www.npc.sk/aktuality feed
2. Observe if content loaded has any maligned characters

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -1035,6 +1035,8 @@ class FreshRSS_Feed extends Minz_Model {
 		$view->rss_url = htmlspecialchars($feedSourceUrl, ENT_COMPAT, 'UTF-8');
 		$view->html_url = $view->rss_url;
 		$view->entries = [];
+		// As DOMDocument is designed for XML rather than HTML, not having this header makes it detect wrong encoding
+		$html = '<?xml version="1.0" encoding="UTF-8" ?>' . $html;
 
 		try {
 			$doc = new DOMDocument();


Alkarex:
Hello, 
The encoding problem is when you retrieve additional content, right? 
Could you please share the OPML settings of the feed given as example?

ostrolucky:
I don't know what you mean by additional content. Any content loaded from the feed is saved with maligned characters, including title of the feed.

[f_2 Aktuality _ NPC - Národné podnikateľské centrum.opml.xml](https://github.com/user-attachments/files/23893131/f_2.Aktuality._.NPC.-.Narodne.podnikatelske.centrum.opml.xml)

Alkarex:
Thanks for the OPML. Would you happen to have one more example of problematic encoding?

On my side, without your patch, it looks good apparently, and I would be pleased to be able to reproduce a problem:

image

Could you tell more about your PHP version and platform?

Frenzie:
Extensions could also play a role, depending.

Edit: by which I mean, see what happens when you disable them, if applicable. :-)

Alkarex:
As far as I can see, `https://www.npc.sk/aktuality` returns `Content-Type: text/html; charset=utf-8` and UTF-8 is already passed to our function in charge of enforcing encoding, so all looks good without this PR:
https://github.com/FreshRSS/FreshRSS/blob/26bb2aa975de30fbc9d7146128533637ac0fb26a/app/Utils/httpUtil.php#L394-L396

which already applies the same as your PR, but after checking that is is appropriate (UTF-8 is not the only encoding we must support):
https://github.com/FreshRSS/FreshRSS/blob/26bb2aa975de30fbc9d7146128533637ac0fb26a/app/Utils/httpUtil.php#L205-L208

ostrolucky:
Yes like I explained in description, when fetching, proper encoding is used. But when saving, CP-1252 is wrongly detected. I'm checking what could be the reason for it to be working for you. For me it doesn't work on my linux nor macos machine. I'll try docker too.

Alkarex:
What does a `curl -v 'https://www.npc.sk/aktuality'` return from your server, for Content-Type? For me it is UTF-8, not CP-1252

ostrolucky:
Yes it's utf-8 for me too. Again, fetching is correctly detected in UTF-8. Code path I modified kicks in shortly before saving this data though and that's where encoding is corrupted for me. As soon as `$doc->loadHTML` below loads this UTF-8 content, it mangles it because it detects cp-1252 instead.

ostrolucky:
I traced this to a buggy PHP build of the PHP distribution I'm using :( closing

ostrolucky:
Update: It's libxml issue https://gitlab.gnome.org/GNOME/libxml2/-/issues/1021. My system was shipping newer version.

Alkarex:
Thanks for the info.
In think we have older versions in our Docker images:
* https://packages.debian.org/trixie/libxml2
* https://pkgs.alpinelinux.org/packages?name=libxml2&branch=v3.23&arch=x86_64

Please test https://github.com/FreshRSS/FreshRSS/pull/8285 if you can

Alkarex:
Although I still believe our current code should work even with a slightly buggy libxml2:

https://github.com/FreshRSS/FreshRSS/pull/8279#issuecomment-3606987520

ostrolucky:
> Please test https://github.com/FreshRSS/FreshRSS/pull/8285 if you can

This gives libxml 2.13.9, which is not high enough to reproduce the issue. I can prepare docker image which has PHP with high enough libxml version so you can see the issue yourself. Would you like?

> Although I still believe our current code should work even with a slightly buggy libxml2:

So actually, libxml closed the issue sayings it's not a bug. Why bug is not solved by code you mentioned I explained: It's not triggered there, but later. Issue is specifically that libxml is used to decode content twice. This snippet reproduces the issue:
```php
$doc = new DOMDocument();
$doc->loadHTML(file_get_contents('https://www.npc.sk/aktuality'));
$doc->loadHTML($doc->saveHTML(), LIBXML_NOWARNING);

echo $doc->encoding;
```

This doesn't reproduce the issue:
```php
$doc = new DOMDocument();
$doc->loadHTML(file_get_contents('https://www.npc.sk/aktuality'));

echo $doc->encoding;
```

So either we find a way to not feed the content to that parsing library out and again, or we go with solution in this PR which gives libxml hint to use utf-8, or we use better html parsing solution in this project and not something that was designed for xml, or something else

Inverle:
Related: https://github.com/FreshRSS/FreshRSS/issues/6831

Alkarex:
> I can prepare docker image which has PHP with high enough libxml version so you can see the issue yourself. Would you like?

Yes, I would be interested

ostrolucky:
Here is a reproducer with archlinux base https://github.com/FreshRSS/FreshRSS/compare/edge...ostrolucky:libxml2.15-image?expand=1

Alkarex:
> Here is a reproducer with archlinux base https://github.com/FreshRSS/FreshRSS/compare/edge...ostrolucky:libxml2.15-image?expand=1

Thanks @ostrolucky , very useful 👍🏻 
Here is a fix:
* https://github.com/FreshRSS/FreshRSS/pull/8296

Tests welcome with various encodings and various PHP / libxml2 versions.

Alkarex:
Taking the opportunity to test with Arch Linux some more:
* https://github.com/FreshRSS/FreshRSS/pull/8299

Tests welcome

PR #8283, opened 2025-12-03
Fix CLI user creation warning when language is not given

Alkarex:
Discovered during https://github.com/FreshRSS/FreshRSS/pull/8277


Inverle:
Yes I found this too before when using a bash for loop and creating 1000 users, didn't keep note of it though

Inverle (approved):

PR #8294, opened 2025-12-07
New links in transitions and jump to next transition

Alkarex:
Easier to explain graphically:

image

The jump to next section ⏭ works when the sorting criterion is a date.

Edit: https://github.com/FreshRSS/FreshRSS/pull/8293 is actually needed after all


Frenzie:
Maybe make it something like 0.8 opacity?

Alkarex:
> Maybe make it something like 0.8 opacity?

I went for `opacity: 0.5`

image

Frenzie (approved):
Seems like a lot of code but if it's something you like. ;-)

Alkarex:
> Seems like a lot of code but if it's something you like. ;-)

It should though be noted that that major part is https://github.com/FreshRSS/FreshRSS/pull/8293 (with general improvements to work with searches), so this PR will actually be quite small.

Kamaris521:
Is there an option to hide / suppress those transition and jump to headers?

Alkarex:
> Is there an option to hide / suppress those transition and jump to headers?

@Kamaris521 See:
* https://github.com/FreshRSS/FreshRSS/discussions/8355#discussioncomment-15346265
* https://github.com/FreshRSS/FreshRSS/discussions/8358#discussioncomment-15353373

PR #8296, opened 2025-12-08
Fix saveHTML() scrambling encoding in newer libxml2

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/pull/8279#issuecomment-3620674818

### DIFF
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -203,8 +203,8 @@ final class FreshRSS_http_Util {
 			}
 		}
 		if ($httpCharsetNormalized === 'UTF-8') {
-			// Save encoding information as XML declaration
-			return '<' . '?xml version="1.0" encoding="' . $httpCharsetNormalized . '" ?' . ">\n" . $html;
+			// Save encoding information as Unicode BOM
+			return "\xEF\xBB\xBF" . $html;
 		}
 		// Give up
 		return $html;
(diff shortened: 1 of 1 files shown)


app/Utils/httpUtil.php
```diff
@@ -203,8 +203,8 @@ private static function enforceHttpEncoding(string $html, string $contentType =
 			}
 		}
 		if ($httpCharsetNormalized === 'UTF-8') {
-			// Save encoding information as XML declaration
-			return '<' . '?xml version="1.0" encoding="' . $httpCharsetNormalized . '" ?' . ">\n" . $html;
```
Frenzie, line 207: And they claim it's not a bug to ignore the XML declaration? O.o

Alkarex: It is during the `saveHTML()` that the issue arises, because it modifies the start of the HTML document, removing important clues about encoding later needed by `loadHTML()`.
So it is not that bad, but still...

Alkarex: We could have kept the XML preamble, but it looks like the Unicode BOM is a bit more friendly.
During save, libxml2 modifies the start like so:

```html
<!DOCTYPE html>
<!--?xml version=\"1.0\" encoding=\"UTF-8\" ?--><html
```

Frenzie (approved):

PR #8298, opened 2025-12-08
Fix chart resize regression due to chart.js v4 update

Inverle:
* Fixes regression from https://github.com/FreshRSS/FreshRSS/pull/7752 (present since 1.27.0 release) 
Clicking the enlarge button resulted in:

instead of:

a

* Add workaround for chart.js zoom bug
* Update to chart.js v4.5.1
Reproduce JS file:
```sh
cd Chart.js && \
git checkout tags/v4.5.1 && \
pnpm install && \
pnpm run build && \
# SHA256 of dist/chart.umd.min.js should be: a0e5c5870e66487338f2d8e412e8c154c1d784e278e9054d5800162fb9541344
```

After this PR, charts continue to work in SeaMonkey however the enlarge function has only been fixed on Firefox and Chrome (~~Safari untested~~ tested working), even though it was working before.
I played around with the options at https://www.chartjs.org/docs/latest/configuration/responsive.html but couldn't get chart.js to work the same way it did before, so I override the behavior manually instead.

For testing, compare behavior of stats controller in this branch, current edge branch and 1.26.3 (`git checkout tags/1.26.3`)


p/scripts/statsWithChartjs.js
```diff
@@ -36,6 +36,36 @@ function initCharts() {
...
+
+	window.addEventListener('resize', () => {
+		for (const id in Chart.instances) {
+			const instance = Chart.instances[id];
+
+			// Workaround for a chart.js bug which makes the canvas blurry if
+			// page was loaded with very low zoom level and later zoomed in
+			instance.update();
+			instance.resize();
+		}
```
Inverle, line 66: I'm calling this a bug since it can be reproduced on the chart.js documentation page itself: https://www.chartjs.org/docs/latest/charts/doughnut.html (in older versions than v4 too)

image

Alkarex (approved):

Inverle:
> Fixes regression from https://github.com/FreshRSS/FreshRSS/pull/7752 (present since 1.27.0 release)

Another approach is to downgrade to chart.js v3.9.1 but I'm not too happy about this option.

PR #8306, opened 2025-12-11
Fix htmlspecialchars for search

Alkarex:
Regression from https://github.com/FreshRSS/FreshRSS/pull/8293

PR #8307, opened 2025-12-11
Improve scrolling into filter in sidebar

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8281

todo:
* [x] Include labels (prefix `t_`) too
* [x] Keep sidebar scrollTop when using the nav menu
* [ ] ~~Make this work in the reader view's sidebar too~~ for separate PR
* [x] Prevent whole page from scrolling on `scrollIntoView()` call, just scroll in the sidebar (probably related: https://github.com/FreshRSS/FreshRSS/pull/8306#issuecomment-3647414618)

### DIFF
--- a/app/layout/nav_menu.phtml
+++ b/app/layout/nav_menu.phtml
@@ -98,7 +98,7 @@
 					$url_query['c'] = 'configure';
 					$url_query['a'] = 'bookmarkQuery';
 				?>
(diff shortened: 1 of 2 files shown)


Inverle:
> Make this work in the reader view's sidebar too

This TODO will be done in a separate PR since it requires optimizing the sidebar toggle code.
edit: it does work on Chrome already though, but only if `#stream` isn't too large / breaks randomly (Firefox is slower it seems)

Alkarex (approved):

PR #8311, opened 2025-12-12
Fix search encoding and quoting

Alkarex:
Revised the encoding approach for searches: the HTML encoding is done just before its use for DB search.
Fix also some cases with wrong quoting.
Fix https://github.com/FreshRSS/FreshRSS/pull/8306#issuecomment-3643865439
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8293


Alkarex:
Should already be significantly better. Merging for wider tests.

PR #8312, opened 2025-12-12
Bump to API level 4 for Fever; add with_ids so we can mass-change rea…

### LINKED ISSUE #8305: [Feature] Fever mark function should accept item list using with_ids (API enhancement)

I've added the Fever API to the FeedMonkey RSS reader. The one catch is the way it looks at articles, it wants to check off lists of item ID's as read/saved and such.  Not "before X" or "in feed Y".  Just lists of item id's.  Funneling 50 or so through the mark= API one at a time is slow--many seconds slow.

So I've bumped the Fever API level to 4 and added support fork mark= to have with_ids supported as well as id=.  It's just a list of item ID"s, and most of the backend of FreshRSS is set up to use such a thing.  Let me know if there's interest, and I can push a PR so you can look at how I did things, but there's nothing surprising.

### Use case

_No response_

Bump API lev to 4 (so clients can know this is available) and support with_ids

Mark w. "before", mark "group", and such do not fit the model of reading content, at least on FeedMonkey.  Maybe Google Reader does this, but Fever is a very tight, tidy API and it feels like overkill to switch to Google's big API.

### Additional context

_No response_

vandys:
As discussed in https://github.com/FreshRSS/FreshRSS/issues/8305

A small enhancement to the Fever API to vastly increase efficiency when bulk marking items.
-

The client at:

https://sources.vsta.org:7100/FeedMonkey/file?name=js/Fever.js&ci=tip

has code to talk to this API enhancement (when the API level says it's supported).  In particular walkArticles().

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -79,8 +79,10 @@ final class FeverDAO extends Minz_ModelPdo
 	}
 
 	/**
-	 * @param array<string|int> $feed_ids
-	 * @param array<string> $entry_ids
+	 * @param array<numeric-string|int> $feed_ids
+	 * @param array<numeric-string> $entry_ids
+	 * @param numeric-string|'' $max_id
+	 * @param numeric-string|'' $since_id
 	 * @return FreshRSS_Entry[]
 	 */
 	public function findEntries(array $feed_ids, array $entry_ids, string $max_id, string $since_id): array {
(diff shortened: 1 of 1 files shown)


Alkarex:
Try to run `make fix-all`

Tip: https://editorconfig.org

p/api/fever.php
```diff
@@ -227,8 +229,15 @@ public function process(): array {
...
 		}
 
-		if (is_string($_REQUEST['mark'] ?? null) && is_string($_REQUEST['as'] ?? null) && is_string($_REQUEST['id'] ?? null) && ctype_digit($_REQUEST['id'])) {
-			$id = $_REQUEST['id'];
+		if (is_string($_REQUEST['mark'] ?? null) && is_string($_REQUEST['as'] ?? null)) {
+			if (is_string($_REQUEST['id'] ?? null) && ctype_digit($_REQUEST['id'])) {
+				$id = $_REQUEST['id'];
+			} elseif (is_string($_REQUEST['with_ids'] ?? null)) {
+				$id = array_values(array_filter(explode(',', $_REQUEST['with_ids']), 'ctype_digit'));
+				// N.B., this will make tagging of feeds or groups unhappy.
```
Alkarex: Should be addressed?

vandys: Seems like there are two ways to go.  Performance of mass-marking feeds/groups is much less important than individual items.  So either document the level 4 API (I believe there's an existing ticket for making the Fever API doc a part of FreshRSS--since the Fever website & email server is missing).  Or I can get those code paths working with arrays of ID's as well.  Happy to do the work, just let me know what you think.

Alkarex: Ok, then I think raising an error or continuing only with the first ID should be sufficient

vandys: As coded, they'll get an error when the non-integer (i.e., the array of strings) reaches, say, setFeedAsRead().  I could add code to check and header()/die() if you want the API to be more explicit in rejecting this?

Alkarex (approved):

Alkarex:
Good enough for me with my last changes. How is it from your side?

vandys:
Hi, sorry for the delay--recovering from a busy weekend.  I've pulled in your work & it's all good on my side.  I also forced my client to do the "mark item" the old way, and that code path still works correctly.  Thank you, it's been a pleasure working with you!

PR #8329, opened 2025-12-17
More uniform SQL search and PHP search

Alkarex:
The behaviour depends though on the database.
MySQL is particularly annoying.
Improve https://github.com/FreshRSS/FreshRSS/discussions/8265#discussioncomment-15278980


Alkarex:
Let's merge here for more tests, and we need another PR to add `intl` as a recommended extension: it was already the case for several other needs, but we do not test it.

PR #8330, opened 2025-12-17
Allow negative category sort numbers

### LINKED ISSUE #8304: [Feature] Allow negative category sort order

There is no easy way to force a category to the end of the list.

### Use case

I'd like to push the "Uncategorized" category to the bottom of the list.

The category sort number only allows numbers >=1, with higher numbers being placed first. I would like to set negative numbers to place categories later.

I imagine it's possible to fix this by giving all categories a sort number, and giving "Uncategorized" a smaller one, but that seems like a lot of work, since I have ~20 categories and there isn't any good way to bulk update categories.

### Additional context

_No response_

stag-enterprises:
Closes #8304 

- Remove the min=1 attribute so negative numbers can be used to force categories to the bottom

### DIFF
--- a/app/views/helpers/category/update.phtml
+++ b/app/views/helpers/category/update.phtml
@@ -31,7 +31,7 @@
 			<div class="form-group">
 				<label class="group-name" for="position"><?= _t('sub.category.position') ?></label>
 				<div class="group-controls">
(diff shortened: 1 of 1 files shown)


Alkarex:
Thanks!

PR #8331, opened 2025-12-18
Add myself to credits

stag-enterprises:
https://github.com/FreshRSS/FreshRSS/pull/8330#issuecomment-3667491619

Sorry if this isn't the right way to do this, but that PR is closed and I can't edit it anymore.


Alkarex:
Thanks! And that is how I was expecting :-)

PR #8335, opened 2025-12-21
Display sidebar dropdowns above if no space below

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/7801

PR #8336, opened 2025-12-21
Fix dropdown menus in sidebar going under the navigation bar

Inverle:
Missed in #8335 because I was testing without having the navigation bar enabled.

image


Alkarex:
Same with me :-)

PR #8340, opened 2025-12-23
Handle fetch of text/plain as <pre>

### LINKED ISSUE #8328: [Feature/Bug] Poor handling of pre-formatted plain text articles

Some feeds (official and web-scraped) do not return content well suited for FreshRSS or rss reader apps pulling from FreshRSS. In my case, these feeds return pre-formatted plain text intended to be displayed with monospace.

### Use case

The FreeBSD project provides two easily scrapable pages listing announced security notices and security vulnerabilities. The "articles" are PGP-signed files that are formatted to be viewable in `less` or `vi`

- https://www.freebsd.org/security/notices/
- https://www.freebsd.org/security/advisories/

A simple checkbox to mark the feed articles as pre-formatted plain-text would work fine. A more well-thought solution could allow the user to transform the article content in some arbitrary way (i.e: to entrap the escaped content in a `<pre><code>` block). Maybe a list of sed-like expressions such as `s/.*/<pre><code>\(1)</code></pre>/g`?

- An extension. This would work similarly to the official YouTube-to-embed extension, however as extensions can't have per-feed settings (afiak), it wouldn't be very flexable.
- Using a custom stylesheet. This works, however only *in* FreshRSS. External RSS readers, even if they were sent the custom CSS, would be unlikely to respect it.

### Additional context

**An example article displayed as-is:**
Image

**An example article with `white-space: pre; font-family: monospace;` applied to `.content .text`:**
Image

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/8328

Wrap `text/plain` Web scraping in a `<pre>...</pre>`

image

### DIFF
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -407,7 +407,12 @@ final class FreshRSS_http_Util {
 				$body = self::enforceHttpEncoding($body, $c_content_type);
 			}
 			if (in_array($type, ['html'], true)) {
-				$body = self::enforceHtmlBase($body, $c_effective_url);
+				if (stripos($c_content_type, 'text/plain') !== false) {
+					// Plain text to be displayed as preformatted text. Prefixed with UTF-8 BOM
+					$body = "\xEF\xBB\xBF" . '<pre>' . htmlspecialchars($body, ENT_NOQUOTES, 'UTF-8') . '</pre>';
+				} else {
+					$body = self::enforceHtmlBase($body, $c_effective_url);
+				}
 			}
 		}
 
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -409,7 +409,7 @@ final class FreshRSS_http_Util {
 			if (in_array($type, ['html'], true)) {
 				if (stripos($c_content_type, 'text/plain') !== false) {
 					// Plain text to be displayed as preformatted text. Prefixed with UTF-8 BOM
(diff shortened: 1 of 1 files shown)


app/Utils/httpUtil.php
```diff
@@ -407,7 +407,12 @@ public static function httpGet(string $url, string $cachePath, string $type = 'h
 				$body = self::enforceHttpEncoding($body, $c_content_type);
 			}
 			if (in_array($type, ['html'], true)) {
-				$body = self::enforceHtmlBase($body, $c_effective_url);
+				if (stripos($c_content_type, 'text/plain') !== false) {
+					// Plain text to be displayed as preformatted text. Prefixed with UTF-8 BOM
+					$body = "\xEF\xBB\xBF" . '<pre>' . htmlspecialchars($body, ENT_NOQUOTES, 'UTF-8') . '</pre>';
```
Frenzie: I'd probably add a class (not sure what exactly, something relating to extracted text perhaps?) so people can easily apply custom styles without affecting all PRE elements.

Inverle: `<pre class="plaintext">` maybe

Frenzie: I didn't like that so I didn't suggest it, but it might do. I was thinking something like `extracted-text` but I'm not fond of that either.

Inverle (approved):

app/Utils/httpUtil.php
Alkarex: Could `<pre class="text-plain">` do?

Frenzie: Good enough for me.

PR #8341, opened 2025-12-24
Fix serialisation of dates in searches

Alkarex:
Dates should not be expanded/resolved in string serialisations of search expressions, as it otherwise break relative user queries such as `P30D`.

Fix of https://github.com/FreshRSS/FreshRSS/pull/8293


Alkarex:
Merging this bug-fix fast to hopefully release 1.28.0 shortly 🎄

PR #8343, opened 2025-12-24
Prevent enumeration of users on instance

Inverle:
+redirect to index in formLogin if logged in
+try to simplify formLogin by not using internal redirects

~~This is a WIP, have not reviewed yet properly if timing attacks are possible for user enumeration.~~

Following user enumeration techniques have been fixed:
* If you entered a valid username with a wrong password, you'd get a fail notification - but if you entered an invalid username then you'd get redirected to a 403 "Login is invalid" page which allowed for trivial user enumeration
   * Solution: always display a notification
   * Related: https://github.com/FreshRSS/FreshRSS/issues/4095
* Multiple reloads of `?c=javascript&a=nonce&user=<user>` displayed a different `salt1` value each time for non-existent users, now a value is displayed based on hash of username if the user doesn't exist
    * Still possible to check if a user exists unauthenticated if they happen to change their password (salt1 for existing users is first 29 characters of password hash, which I'm not sure is a great idea - maybe something should be done with that)
    * Tried to simplify the `nonce` endpoint
    * Overall I think it would be best to get rid of bcrypt.js challenge authentication
    
One user enumeration technique that is not possible to fix is when the registration form is enabled (can't register existing usernames)


Alkarex:
Is this still a draft?

Inverle:
Aside from

> This is a WIP, have not reviewed yet properly if timing attacks are possible for user enumeration.

I think it was ready

app/Controllers/authController.php
```diff
@@ -113,20 +117,16 @@ public function formLoginAction(): void {
 		$limits = FreshRSS_Context::systemConf()->limits;
 		$this->view->cookie_days = (int)round($limits['cookie_duration'] / 86400, 1);
 
-		$isPOST = Minz_Request::isPost() && !Minz_Session::paramBoolean('POST_to_GET');
-		Minz_Session::_param('POST_to_GET');
-
-		if ($isPOST) {
```
Inverle, line 119: Similar changes to these from #8535

app/Controllers/javascriptController.php
```diff
@@ -88,29 +85,20 @@ public function nonceAction(): void {
...
-					$this->view->nonce = hash('sha256', FreshRSS_Context::systemConf()->salt . $user . random_bytes(32));
-					Minz_Session::_param('nonce', $this->view->nonce);
-					return;	//Success
-				}
-			} catch (Minz_Exception $me) {
-				Minz_Log::warning('Nonce failure: ' . $me->getMessage());
+			$s = $user_conf->passwordHash;
+			if (strlen($s) >= 60) {
+				// CRYPT_BLOWFISH Salt: "$2a$", a two digit cost parameter, "$", and 22 characters from the alphabet "./0-9A-Za-z".
+				$this->view->salt1 = substr($s, 0, 29);
```
Inverle, line 94: Does it have to be done this way?

Alkarex: I cannot remember the exact details, but this is not the only solution, if you have something else in mind.
This was a way to extract the salt.

SamCyber001:
The timing thing from the description is still there though. Absent users hit crypt() and real ones just substr(), so absent comes back a lot slower: I got ~6ms for a real user vs ~100ms for a missing one and the ranges didn't overlap, so you can still tell them apart by response time alone.
Simplest fix is probably to skip crypt() for the fake and just use '$2y$09$' . $salt. Only thing to watch is the last salt char, since crypt() normalizes it and a raw base64 one might stand out. Otherwise run the same crypt() on both, or a usleep like the login action.
Can suggest a small patch if useful.

Inverle:
Sure, you can send the patch here

SamCyber001:
Patch below. It skips crypt() on the fake path and fixes the last salt char by hand, so you
get the same salt crypt() produced (checked a bunch of usernames) but without the bcrypt cost.
Missing vs real users take the same time now, so the timing signal's gone.

```diff
-		// Failure: Return static random data for given username.
-		$salt = substr(str_replace('+', '.', base64_encode(hash('sha256', FreshRSS_Context::systemConf()->salt . $user))), 0, 22);
-		$this->view->salt1 = substr(crypt('failure', '$2y$09$' . $salt), 0, 29);
+		// Failure: static per-user salt, indistinguishable from a real one. Not crypt() here: it
+		// ran only on this branch, so a missing user was slower and enumerable by timing.
+		$alphabet = './ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+		$salt = substr(str_replace('+', '.', base64_encode(hash('sha256', FreshRSS_Context::systemConf()->salt . $user))), 0, 22);
+		$salt[21] = '.Oeu'[strpos($alphabet, $salt[21]) >> 4]; // last char carries 2 bits; keep it canonical like crypt() did
+		$this->view->salt1 = '$2y$09$' . $salt;
```

Alkarex:
> Patch below

Thanks! Could you please add it using the *Add a suggestion* feature, so it can be merged with proper credits?

image

Issue #7835, opened 2025-08-18
[Bug] 1.27.0 removes all CSP Header


sleeksorrow:
### Describe the bug

It seems that version 1.27.0 actively removes all CSP headers.

I updated from 1.26.3 to 1.27.0 using git. Afterwards I got the warning in FreshRSS: "The CSP header in use is unsafe and FreshRSS may be vulnerable to XSS attacks." I read the documentation and tested myself and I found outr two things:
1: There are no CSP headers sent in FreshRSS.
2: I have no CSP headers set myself. Other paths on my webserver also have no CSP headers.
3: If I enable CSP headers in my webserver config, then they are sent in all other locations but not on /FreshRSS path.

Using Apache2 2.4.58 on Ubuntu 24.04.

I also downgraded back to version 1.26.3 and the warning is gone. But when I was searching respone headers I also do not see any. Also if I enable them in server config, then they are in other paths but not in /FreshRSS, just like with 1.27.0, just without warning. So maybe the problem is something else and 1.27.0 has this new warning feature? I see some CSP changes in release information. But then I'm quite clueless where I messed up. I tend to leave as much in default as possible.

### To Reproduce

1. Go to my private server URL /FreshRSS
2. Log in
3. See warning message
4. Cannot find CSP Headers in response

### Expected behavior

1. Go to my private server URL /FreshRSS
2. Log in
3. See no warning message
4. Find secure CSP Headers in response

### FreshRSS version

1.27.0

### System information

- Database version: mariadb 10.11.13
- PHP version: 8.3
- Installation type: git
- Web server type: Apache 2.4.58
- OS: Ubuntu 24.04
- Browser: Firefox 141.0.3

Alkarex:
Hello,
Do you have some kind of proxy in addition to your Apache?

sleeksorrow:
No, just vanilla apache with mod_php and SSL. I checked my site config and did not find any special stuff. I had some ProxyPass directives that are obsolete and removed them to make it even simpler. If you think it can be some server config setting I will try more. I set some other headers, I can try to comment that out for a test.
EDIT: No change by that

Vagenthar:
It has also stop working in iFrame. refused connections.

vk2r:
From what I've been reading, in the new version of FreshRSS you have to send a specific CSP policy, otherwise it will display an error on the page, defined as a security issue. See [CSP Header](https://github.com/FreshRSS/FreshRSS/pull/7804).

[...]

PR #8344, opened 2025-12-24
Change `Content-Disposition: inline` to `attachment` in `f.php`

Inverle:
Some [misconfigured instances](https://github.com/FreshRSS/FreshRSS/issues/7835) may be stripping out the CSP header that `f.php` sends, which can be mitigated by forcing the browser to download the image instead of displaying it and executing JS code from unsanitized SVGs for example.

Contributes to https://github.com/FreshRSS/FreshRSS/pull/8263 and https://github.com/FreshRSS/FreshRSS/pull/7924
(improving security when CSP is not present)


Inverle:
> which can be mitigated by forcing the browser to download the image

or at least that's my assumption currently, need to double check

Issue #8349, opened 2025-12-25
[Feature] Support `h2c` (HTTP/2) protocol


schildbach:
### Is your feature request related to a problem? Please describe.

I'm using the Docker image of FreshRSS 1.28.0. Currently, it only supports HTTP/1.1 on port 80. I propose to also support HTTP/2 on that port (aka `h2c`).

HTTP/2 has many advantages, one of them would be that an SSL-terminating reverse proxy would not need to convert between HTTP/2 and HTTP/1.1.

I assume the file `Docker/FreshRSS.Apache.conf` would beed to be adapted according to these [Apache2 instructions](https://httpd.apache.org/docs/2.4/howto/http2.html).

If you're interested, I could try my luck at a PR.

### Use case

Seamless integration with Traefik h2c forwarder.

### Describe the solution you’d like

see above

Alkarex:
Hello,
General intro: The Docker image is supposed to be served by a proper proxy with TLS termination such as Traefik or Caddy (or by something more generic but less good such as nginx).
Back to your proposition: Regarding internal communication with the proxy, supporting `h2c` could indeed be relevant, although the benefits are probably not as important as for the communication with remote clients. PR welcome, yes, thanks.

schildbach:
Darn, I just learned that the `http2` module requires `mpm_event` (rather than the current `mpm_prefork`) and that in turn isn't compatible with `php8.4` (CGI, I assume). So a switch to `php8.4-fpm` would be required first. What are your experiences with FPM, and how do you think about switching?

Alkarex:
> What are your experiences with FPM, and how do you think about switching?

There are a couple of notable differences to be aware of, in particular the control over buffering.
For now (until a completely new UI at some later point in the future), it is a must to have a more-or-less working `flush()`.

An easy test is to add a `sleep(30);` just after this line (for instance like so `<?php flush(); sleep(30); ?>`):
https://github.com/FreshRSS/FreshRSS/blob/6fb5263633ce318432cf3d177539cbd574e0fce2/app/layout/nav_menu.phtml#L287

And the expectation is to see a spinner:

<img width="908" height="275" alt="Image" src="https://github.com/user-attachments/assets/819000cc-1a85-4c3f-bade-512b56cecc45" />

This tests the case of a slow DB search, for instance.

Alkarex:
And I would also be interested in performance+resource test comparisons (I will help with that), with a priority on being light for single-user setups on tiny server.

[...]

PR #8353, opened 2025-12-25
Dockerfile: switch from CGI to FastCGI for PHP

schildbach:
Disclaimer:

This PR requires extra review scrutiny, as I've got no experience with PHP. I'm already running this PR plus an HTTP/2 (see #8349) enablement on my private server and will look into any problems that might arise. So far, it feels very performant.

Switch from CGI to FastCGI for PHP when running inside Docker

1. build and run the Debian docker image
2. test UI via web browser
3. test API via reader apps (Newsflash etc)

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -7,7 +7,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && \
 	apt-get install --no-install-recommends -y \
 	ca-certificates cron \
-	apache2 libapache2-mod-php \
+	apache2 libapache2-mod-fcgid php-fpm \
 	libapache2-mod-auth-openidc \
 	php-curl php-gmp php-intl php-mbstring php-xml php-zip \
 	php-sqlite3 php-mysql php-pgsql && \
@@ -42,7 +42,9 @@ RUN \
 	sed -r -i "/^\s*(CustomLog|ErrorLog|Listen) /s/^/#/" /etc/apache2/apache2.conf && \
 	sed -r -i "/^\s*Listen /s/^/#/" /etc/apache2/ports.conf && \
 	# Enable required Apache modules
-	a2enmod -q deflate expires filter headers mime remoteip setenvif && \
+	a2enmod -q fcgid proxy_fcgi deflate expires filter headers mime remoteip setenvif && \
+	# Enable php-fpm
+	a2enconf -q php8.4-fpm && \
 	# Enable FreshRSS configuration for Apache
 	a2ensite -q 'FreshRSS*' && \
 	# Disable unwanted PHP modules
@@ -68,6 +70,9 @@ ENTRYPOINT ["./Docker/entrypoint.sh"]
 
 EXPOSE 80
 # hadolint ignore=DL3025
-CMD ([ -z "$CRON_MIN" ] || cron) && \
+CMD bash -c "\
+	/usr/sbin/php-fpm8.4 -F & \
+	([ -z "$CRON_MIN" ] || cron) && \
 	. /etc/apache2/envvars && \
-	exec apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ] && echo "-D OIDC_ENABLED")
+	apache2 -D FOREGROUND $([ -n "$OIDC_ENABLED" ] && [ "$OIDC_ENABLED" -ne 0 ] && echo "-D OIDC_ENABLED") & \
+	wait -n"


Docker/Dockerfile
```diff
@@ -69,5 +71,6 @@ ENTRYPOINT ["./Docker/entrypoint.sh"]
 EXPOSE 80
 # hadolint ignore=DL3025
 CMD ([ -z "$CRON_MIN" ] || cron) && \
+	/usr/sbin/php-fpm8.4 -F & \
```
Frenzie, line 74: Is that a typo or meant to be a background process? It looks odd in this location either way.

schildbach: Yes, it's meant to be run in the background. Otherwise, Apache (down below) would never be started.

schildbach: The idea of the construct is that both Apache and FPM need to run in parallel, and both shall send their stdout/stderr to the console so it's sent to the Docker log.

Frenzie: I'll attempt to clarify what I mean.

1. Compare the difference between these two commands:
   ```
   false && true
   false && { false; } & true
   ```
   The behavior is quite different. The former won't execute the second/third (Apache), while the latter will. This change in behavior won't occur if you put the background process first.
   ```
   { false; } & false && true
   ```

2. The entire `-F &`. Keep it in foreground, then put it in the background. That sounds a bit weird. How is it different from a plain `/usr/sbin/php-fpm8.4 -D;`? Something related to logging? Doesn't it start in the background by default regardless?
3. How dependable is it in case of errors? How much harder is it to debug? Shouldn't it be managed by supervisord or by separate Docker containers instead?

schildbach: ad 1: So you suggest to put the FPM line before the cron line?

ad 2: Yes, the idea is that FPM sends its output to stdout/stderr so it appears in the Docker logs. Try `echo hello &` in a shell and you'll see the "hello". Deamonizing the process won't do this. But if you prefer we can use `-D` rather than the shell backgrounding (`&`).

ad 3: FPM needs to run, otherwise the app won't work (apart from some staticly served resources maybe). Debugging I can't tell, but I tried to preserve log output (see 2). For development the old configuration can stilll be used if desired. A separate container for FPM would probably be possible, though it's a more complicated setup that will require some kind of orchestrator (e.g. compose).

schildbach:
I just tried how a PHP error would look like and it's logged by the FastCGI proxy anyway:

```
[Fri Dec 26 12:35:44.066206 2025] [proxy_fcgi:error] [pid 38:tid 43] [client fd00::1:47838] AH01071: Got error 'PHP message: PHP Fatal error:  Uncaught Error: Undefined constant "blablablbala" in /var/www/FreshRSS/p/i/index.php:31\nStack trace:\n#0 {main}\n  thrown in /var/www/FreshRSS/p/i/index.php on line 31'
```

So I pushed a fix on top that changes the shell backgrounding to daemonizing (`-D` switch).

schildbach:
And another patch adds a symlink from the FPM error log to stderr. This causes FPM startup messages be visible in the Docker log, and I assume also serious errors would show up.

Docker/Dockerfile
schildbach: Did my last two patches address your points?

Frenzie: > ad 1: So you suggest to put the FPM line before the cron line?

Unless the change in behavior is on purpose. I'm just saying it looks weird and suspicious to change the behavior. These things jump out. ;-)

> ad 2: Yes, the idea is that FPM sends its output to stdout/stderr so it appears in the Docker logs. Try echo hello & in a shell and you'll see the "hello". Deamonizing the process won't do this. But if you prefer we can use -D rather than the shell backgrounding (&).

It's less about any particular choice as it is about documenting why the choice was made. Even you might not know anymore a month from now and then you just have some mystery flags that require extensive research and testing to adjust.

> ad 3: FPM needs to run, otherwise the app won't work (apart from some staticly served resources maybe). 

That's my point. Something like supervisord would ensure it actually runs. Right now it doesn't matter because Docker will always make sure Apache runs (that is, the container will restart in the unlikely event Apache should crash), but it won't for this background process. Presumably and hopefully this is an almost entirely academic concern, but it seems to be a potential point of failure that doesn't currently exist.

schildbach: I looked at supervisord but it seems quite involved. I'd rather not add so much complexity, if it's only academic.

Frenzie: It's a very real risk. What I mean is that PHP is probably fairly battle tested by now.

schildbach: Apparently we could use this pattern:

```
process1 &   # fpm
process2 &   # apache
wait -nf         # wait for any of the two processes to finish
```

However this seems to work only with shell backgrounding, not with deamonizing.

Frenzie: That pattern makes a lot of sense to me (with the -F then).

schildbach: Ok, I'll try that. Will take until first week of Jan though.

Alkarex:
Thanks for this PR @schildbach 
This is a nice way to start testing.

Apparently, `flush()` is not working https://github.com/FreshRSS/FreshRSS/issues/8349#issuecomment-3691649306
So that will require a bit more research

Docker/Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && \
 	apt-get install --no-install-recommends -y \
 	ca-certificates cron \
-	apache2 libapache2-mod-php \
+	apache2 libapache2-mod-fcgid php-fpm \
```
Alkarex, line 10: I am not sure of the exact pros and cons for the FreshRSS main use-cases, but https://cwiki.apache.org/confluence/display/HTTPD/Php-fcgid tells:

> On what conditions not to run php on fcgid
> If you run httpd 2.4, you should consider [PHP-FPM](http://wiki.apache.org/httpd/PHP-FPM)

And https://cwiki.apache.org/confluence/display/httpd/PHP-FPM tells:

> using mod_proxy_fcgi and php-fpm

schildbach:
I just implemented the "process1 & ; process2 & ; wait -n" pattern described above, and successfully tested that it works as expected by killing either of the processes.

I also squashed and rebased to edge. I also added a more detailed commit description.

So I think what's left is the `flush()` issue. Can some PHP dev help with this? I have zero PHP experience. Where in the app does the issue surface?

Docker/Dockerfile
schildbach: Fwiw, the Docker image is using Apache/2.4.65 (Debian).

Docker/Dockerfile
schildbach: Done.

Frenzie:
It can be a big speed difference if you set the network to emulate some slow speed (3G or less), especially if the server and/or DB is also being a bit slow.

Frenzie:
Basically it means output should be sent ASAP and not at some later point when the full page is done.

Edit: you can test really easily by inserting a sleep(2) or some such in a basic for loop.

Alkarex:
> So I think what's left is the `flush()` issue. Can some PHP dev help with this? I have zero PHP experience. Where in the app does the issue surface?

This is not specific to PHP. It is the same problem with any other language. The flush cannot be controlled from us anymore because there are suddenly several layers, each with their own buffer.

Explanation + how to reproduce: https://github.com/FreshRSS/FreshRSS/issues/8349#issuecomment-3691649306

The main issue is that users are faced with a white page for potentially a long time when the database request is slow, which is the case for instance when making a search. Before this PR, since we could control the buffering, a partial page is sent so that at least the menus and a waiting animation is shown.

Another issue is that time to first byte gets poorer, leading to reduced performance especially on slow networks.

I think the options include reducing the buffering size at some of the new layers introduced by this PR, but this will reduce the performances in some other situations. I am not sure there is an easy solution for that.

A first step in any case would be to add a performance test, which can be used to compare various approaches (not only for this PR). I have some elements of that on my own TODO list.

schildbach:
Ok, thanks, now I understand the problem. Let me ask two questions:

- How does the common case of a reverse proxy relate to the issue? Is `flush()` relayed e.g. via Traefik? My understanding is if it doesn't "arrive" at the user's web browser, `flush()` is ineffective.
- Isn't this kind of reliance on long-running requests a bad design anyway? I'd expect to receive a quick response with the page layout and a loading indicator (in place of the DB results), and then a second (background, long-running) request queries the DB?

Alkarex:
> How does the common case of a reverse proxy relate to the issue? Is `flush()` relayed e.g. via Traefik? My understanding is if it doesn't "arrive" at the user's web browser, `flush()` is ineffective.

I have not tested in depth, but I run with Traefik myself and it works fine. Additional investigation welcome.

> Isn't this kind of reliance on long-running requests a bad design anyway?

That requires a rewrite using JavaScript to load the data, which has pros and cons. I do plan to propose an alternative pure-JS UI at some point in the future, but I like the idea of having a true HTML response.

> I'd expect to receive a quick response with the page layout and a loading indicator (in place of the DB results), and then a second (background, long-running) request queries the DB?

This is the behaviour today, but with a single query.

Frenzie:
> Isn't this kind of reliance on long-running requests a bad design anyway? I'd expect to receive a quick response with the page layout and a loading indicator (in place of the DB results), and then a second (background, long-running) request queries the DB?

No, not in the slightest. The exact opposite is true. This is how you make fast, *dependable* websites that always work no matter what happens.

PR #8365, opened 2025-12-28
Running FreshRSS in a rootless Docker container

### LINKED ISSUE #8362: [Feature] Run Docker rootless to improve security

To prevent root access to the Docker host-server it it better to run the Docker container in rootless mode.

### Use case

Improve security running FreshRSS as a Docker container.

Provide an updated Dockerfile to enable FreshRSS in a rootless Docker container.
This means adding a run-as user + fixing some permissions in the Docker container when building it.
This also means running a replacement for cron (which requires root privileges) like supercronic.
I will provide both the updated Dockerfile + some additional instructions in a PR.

_No response_

### Additional context

_No response_

tmvtmv:
Closes #8362 

  - Create a non-root user `freshrss` (uid 1000) to run Apache
  - Replace system cron with `supercronic` (works without root privileges)
  - Change default listening port from 80 to 8080 (non-privileged port)
  - Add new `entrypoint-nonroot.sh` script adapted for rootless operation
  - Pre-create data directories with proper permissions at build time
  - Update documentation to reflect new port mappings

  ### Benefits
  - Improved container security by not running as root
  - Compatible with rootless Docker and Podman
  - Follows container security best practices

1. `cd Docker`
2. `docker build -f Dockerfile-Newest .. -t freshrss:nonroot`
3. `docker run -d -p8080:8080 -e 'CRON_MIN=15'  --name freshrss freshrss:nonroot`
4. `docker exec -it freshrss sh -c "netstat -an | grep LISTEN"`
     Expected output:
     `tcp        0      0 :::8080                 :::*                    LISTEN`
5. `docker exec -it freshrss sh -c "ps -a | grep [c]rontab"`
     Expected output:
     `11 freshrss  0:00 supercronic /var/www/FreshRSS/Docker/crontab.freshrss`
6. Surf to http://localhost:8080 and FreshRSS should work as normal.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [X] documentation updated

### DIFF
--- a/Docker/Dockerfile-Newest
+++ b/Docker/Dockerfile-Newest
@@ -3,20 +3,34 @@ FROM alpine:edge
 
 ENV TZ=UTC
 SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
+ARG TARGETARCH
 RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories && \
 	apk add --no-cache \
-	tzdata \
+	tzdata sed \
 	apache2 php85-apache2 \
 	apache-mod-auth-openidc \
 	php85 php85-curl php85-gmp php85-intl php85-mbstring php85-xml php85-zip \
 	php85-ctype php85-dom php85-fileinfo php85-iconv php85-json php85-openssl php85-phar php85-session php85-simplexml php85-xmlreader php85-xmlwriter php85-xml php85-tokenizer php85-zlib \
-	php85-pdo_sqlite php85-pdo_mysql php85-pdo_pgsql
+	php85-pdo_sqlite php85-pdo_mysql php85-pdo_pgsql && \
+	# Install supercronic for rootless cron
+	wget -q -O /usr/local/bin/supercronic https://github.com/aptible/supercronic/releases/download/v0.2.33/supercronic-linux-${TARGETARCH} && \
+	chmod +x /usr/local/bin/supercronic
 
-RUN mkdir -p /var/www/FreshRSS /run/apache2/
+# Create non-root user
+RUN adduser -D -u 1000 -G apache freshrss && \
+	mkdir -p /var/www/FreshRSS /run/apache2/ && \
+	chown -R freshrss:apache /run/apache2/
 WORKDIR /var/www/FreshRSS
 
-COPY --chown=root:www-data . /var/www/FreshRSS
-COPY ./Docker/*.Apache.conf /etc/apache2/conf.d/
+COPY --chown=freshrss:apache . /var/www/FreshRSS
+COPY --chown=freshrss:apache ./Docker/*.Apache.conf /etc/apache2/conf.d/
+
+# Create data directories with proper permissions
+RUN mkdir -p ./data/cache ./data/extensions-data ./data/favicons ./data/fever \
+	./data/PubSubHubbub/feeds ./data/PubSubHubbub/keys ./data/Retry-After \
+	./data/tokens ./data/users/_ && \
+	chown -R freshrss:apache ./data ./extensions && \
+	chmod -R 775 ./data ./extensions
 
 ARG FRESHRSS_VERSION
 ARG SOURCE_COMMIT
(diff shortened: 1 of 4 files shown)


Docker/Dockerfile-Newest
```diff
@@ -3,20 +3,34 @@ FROM alpine:edge
...
-	tzdata \
+	tzdata sed \
 	apache2 php85-apache2 \
 	apache-mod-auth-openidc \
 	php85 php85-curl php85-gmp php85-intl php85-mbstring php85-xml php85-zip \
 	php85-ctype php85-dom php85-fileinfo php85-iconv php85-json php85-openssl php85-phar php85-session php85-simplexml php85-xmlreader php85-xmlwriter php85-xml php85-tokenizer php85-zlib \
-	php85-pdo_sqlite php85-pdo_mysql php85-pdo_pgsql
+	php85-pdo_sqlite php85-pdo_mysql php85-pdo_pgsql && \
+	# Install supercronic for rootless cron
+	wget -q -O /usr/local/bin/supercronic https://github.com/aptible/supercronic/releases/download/v0.2.33/supercronic-linux-${TARGETARCH} && \
```
Alkarex, line 16: I am not so keen on downloading third-party binaries like that. Isn't there another option from the existing repos?

Docker/Dockerfile-Newest
```diff
@@ -3,20 +3,34 @@ FROM alpine:edge
...
-COPY ./Docker/*.Apache.conf /etc/apache2/conf.d/
+COPY --chown=freshrss:apache . /var/www/FreshRSS
+COPY --chown=freshrss:apache ./Docker/*.Apache.conf /etc/apache2/conf.d/
+
+# Create data directories with proper permissions
+RUN mkdir -p ./data/cache ./data/extensions-data ./data/favicons ./data/fever \
+	./data/PubSubHubbub/feeds ./data/PubSubHubbub/keys ./data/Retry-After \
+	./data/tokens ./data/users/_ && \
+	chown -R freshrss:apache ./data ./extensions && \
+	chmod -R 775 ./data ./extensions
```
Alkarex, line 33: Would be good to use the existing `cli/access-permissions.sh` if possible instead of all those manual changes, which will require maintenance

But I am not sure this code makes much sense since the whole `./data/` folder is supposed to be mounted from a volume and the data structure re-created at runtime

Docker/entrypoint-nonroot.sh
```diff
@@ -0,0 +1,101 @@
...
+fi
+
+if [ -n "$CRON_MIN" ]; then
+	# Generate crontab with custom minute value for supercronic
+	sed -r "s#^[^ ]+ #$CRON_MIN #" /var/www/FreshRSS/Docker/crontab.freshrss.default > /var/www/FreshRSS/Docker/crontab.freshrss
+fi
+
+# Ensure data directories exist (non-root safe)
+data_path="${DATA_PATH:-./data}"
+mkdir -p "${data_path}/users/_/" 2>/dev/null || true
```
Alkarex, line 63: I do not believe this is sufficient. The whole structure must potentially be re-created at runtime, best with `./cli/access-permissions.sh` so there is a single place to maintain

Docker/entrypoint-nonroot.sh
```diff
@@ -0,0 +1,101 @@
...
+	if [ -n "$OIDC_SCOPES" ]; then
+		# Compatibility with : as separator instead of space
+		OIDC_SCOPES=$(echo "$OIDC_SCOPES" | tr ':' ' ')
+		export OIDC_SCOPES
+	fi
+fi
+
+if [ -n "$CRON_MIN" ]; then
+	# Generate crontab with custom minute value for supercronic
+	sed -r "s#^[^ ]+ #$CRON_MIN #" /var/www/FreshRSS/Docker/crontab.freshrss.default > /var/www/FreshRSS/Docker/crontab.freshrss
```
Alkarex, line 58: The logic for passing environment information has been removed, which looks wrong

Docker/Dockerfile-Newest
Alkarex, line 1: This is probably not the right file to modify.
This one is a copy of the Alpine one with the newest possible versions of e.g. PHP and is used only for development.
It should probably be a new Dockerfile to get started, before potentially promoting it to the main Alpine Dockerfile

Docker/entrypoint-nonroot.sh
Alkarex, line 1: This is mostly a copy of the existing `Docker/entrypoint.sh` with only a few changes (and what appears to be a few bugs).
It would be nice to be able to use the same `Docker/entrypoint.sh` with a test when needed (as little as possible). It is for instance easy to test if this is run as root or not.

Docker/Dockerfile-Newest
Alkarex: See e.g. https://pkgs.alpinelinux.org/packages?name=supercronic

Docker/FreshRSS.Apache.conf
```diff
@@ -1,5 +1,5 @@
 ServerName freshrss.localhost
-Listen 80
+Listen 8080
```
Alkarex, line 2: This is also controlled by the `LISTEN` environment variable.
If the default is changed here, all our Docker images should be changed, and not only this new one

Docker/README.md
```diff
@@ -507,7 +507,7 @@ upstream freshrss {
 }
 
 server {
-	listen 80;
+	listen 8080;
```
Alkarex, line 510: This one is unrelated

Docker/README.md
```diff
@@ -557,7 +557,7 @@ upstream freshrss {
...
```
Alkarex, line 560: This one is also unrelated

Docker/FreshRSS.Apache.conf
tmvtmv: I changed this because a non-root user is not allowed to run services listening on port 1-1024.

Docker/README.md
tmvtmv: I changed this because a non-root user is not allowed to run services listening on port 1-1024.

Docker/README.md
tmvtmv: I changed this because a non-root user is not allowed to run services listening on port 1-1024.

tmvtmv:
Thanks for your feedback! I’ll come up with a better proposal :-)

Docker/Dockerfile-Newest
tmvtmv: Good catch! I totally forgot to check the existing repo for supercronic.

Docker/Dockerfile-Newest
tmvtmv: You are right. By setting the correct ownership to /var/www/FreshRSS earlier in the Dockerfile, this section is not required anymore.

```
RUN adduser -D -u 1000 -G apache freshrss && \
	mkdir -p /var/www/FreshRSS /run/apache2/ && \
	chown -R freshrss:apache /var/www/FreshRSS /run/apache2/
```

Docker/entrypoint-nonroot.sh
tmvtmv: Whoops. I was too much focused on getting supercronic to work … fixed.

Docker/README.md
Alkarex: But this is unrelated to this non-root Docker PR.
This line defines the nginx listening port in case of reverse proxy

Docker/README.md
Alkarex: Still unrelated to this PR, as this is an example of reverse proxy configuration.

Docker/FreshRSS.Apache.conf
Alkarex: My comment still stands. An approach is to define a default `LISTEN` to 8080 for the new Dockerfile and leave this line unchanged, otherwise the other Dockerfiles will fail

Docker/Dockerfile-Newest
tmvtmv: OK, I'll create a new file: Dockerfile-Alpine-nonroot

PR #8376, opened 2025-12-31
Switch to using CURLOPT_ACCEPT_ENCODING instead of the deprecated CURLOPT_ENCODING

### LINKED ISSUE #8374: [Bug] Handling of feeds with invalid Content-Encoding is broken
### Describe the bug

I recently switched to FreshRSS and found one of my feeds not working. It returned the following error in the logs:
```
A feed could not be found at `https://blog.example.com/index.xml`; the status code is `200` and content-type is `` [https://blog.example.com/index.xml]
```
After googling a bit, I came across the advise to try adding `#force_feed` to the end of the feed URL. This lead to the following error message:
```
A feed could not be found at `https://blog.example.com/index.xml`. Empty body. [https://blog.example.com/index.xml#force_feed]
```

Fetching the `index.xml` manually via cURL worked and returned the expected result. But it had this header set:
```
Content-Encoding: aws-chunked
```
This seems to be an issue of how Ceph S3 works with [Hugo's deploy functionality](https://gohugo.io/configuration/deployment/), so it will be fixed in the blog's setup.

But because the W3C Feed Validator thought the feed was fine, I decided to dig a bit further, and I believe I found a bug in FreshRSS. I added a number of debug prints throughout the code in the 1.28.0 version and ended up here:
https://github.com/FreshRSS/FreshRSS/blob/fdd82820f16733b6e07def5b590fd94879e5a520/lib/simplepie/simplepie/src/SimplePie.php#L2163

When outputting `$file->get_body_content()`, it is entirely empty. At the same time, the `$file->get_status_code` is still `200`. This then leads to the rest of the function to be executed, trying to find a feed, but not being able to.

I then dug a bit into the actual fetching code, and finally found the issue here:
https://github.com/FreshRSS/FreshRSS/blob/19bdfc026700087bc1c910fbf8920aa0a77f4f5c/lib/simplepie/simplepie/src/File.php#L146-L154

Outputting the curl error:
```
cURL error 61: Unrecognized content encoding type. libcurl understands deflate, gzip, br, zstd content encodings
```
And the `responseHeaders`:
```
HTTP/2 200 \r\naccept-ranges: bytes\r\ncontent-encoding: aws-chunked\r\ncontent-type: application/rss+xml\r\ndate: Tue, 30 Dec 2025 22:50:37 GMT\r\netag: "xxxxx"\r\nlast-modified: Sat, 13 Dec 2025 21:23:42 GMT\r\nserver: Ceph Object Gateway (squid)\r\nx-amz-meta-md5chksum: yyyyy\r\ncontent-length: 11242\r\n\r\n\r\n
```
Meaning the error does make sense. Looking further, I found that the `CURLOPT_ENCODING` option is deprecated, and setting it to `none` likely doesn't have any effect anymore. As a consequence, the retry of the fetch in line 152 also fails. This failure then seems to be uncaught, and execution continues as if the fetch succeeded, but the body is left empty.

See the deprecation info [in the PHP docs](https://www.php.net/manual/en/curl.constants.php#constant.curlopt-encoding). The replacement seems to be the [CURLOPT_ACCEPT_ENCODING](https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html) option.

I was able to fix the issue by changing line 150 to this:
```php
curl_setopt($fp, CURLOPT_ACCEPT_ENCODING, null);
```
Then, using the original feed worked. But it still needed the `#force_feed` appended, otherwise no articles are loaded, likely due to the same content-encoding issue.

### To Reproduce

1. Go to a page which sends a `content-encoding: aws-chunked` header for its feed file
2. Try to add that file as a feed in FreshRSS, either with or without `#force_feed` appended
3. Observe that FreshRSS says there's no feed to be found
4. Check the W3C feed validator to see that it thinks the feed is perfectly fine

### Expected behavior

The feed is correctly fetched and added to FreshRSS.

### FreshRSS version

1.28.0

### System information

- Database version: SQLite
- PHP version: From Docker container for v1.28.0
- Installation type: Docker
- Web server type: Apache from Docker container, fronted by Traefik as a proxy
- Device: Desktop
- OS: Gentoo Linux
- Browser: Firefox 146.0

### Additional context

_No response_

mmeier86:
Closes #8374 

- Replacing deprecated `CURLOPT_ENCODING` with `CURLOPT_ACCEPT_ENCODING`

As per the [CURLOPT_ACCEPT_ENCODING docs](https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html), `CURLOPT_ENCODING` is obsolete since cURL 7.21.6.

1. Set up a test instance and add any feet
2. Observe that adding feeds still works

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -1415,7 +1415,7 @@ class FreshRSS_Feed extends Minz_Model {
 				CURLOPT_USERAGENT => FRESHRSS_USERAGENT,
 				CURLOPT_MAXREDIRS => 10,
 				CURLOPT_FOLLOWLOCATION => true,
-				CURLOPT_ENCODING => '',	//Enable all encodings
+				CURLOPT_ACCEPT_ENCODING => '',	//Enable all encodings
 				//CURLOPT_VERBOSE => 1,	// To debug sent HTTP headers
 			]);
 			$response = curl_exec($ch);
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -339,7 +339,7 @@ final class FreshRSS_http_Util {
 			CURLOPT_MAXREDIRS => 4,
 			CURLOPT_RETURNTRANSFER => true,
 			CURLOPT_FOLLOWLOCATION => true,
-			CURLOPT_ENCODING => '',	//Enable all encodings
+			CURLOPT_ACCEPT_ENCODING => '',	//Enable all encodings
 			//CURLOPT_VERBOSE => 1,	// To debug sent HTTP headers
 		]);
 
--- a/cli/health.php
+++ b/cli/health.php
@@ -17,7 +17,7 @@ if ($ch === false) {
 curl_setopt_array($ch, [
 	CURLOPT_CONNECTTIMEOUT => is_numeric($options['connect_timeout'] ?? null) ? (int)$options['connect_timeout'] : 3,
 	CURLOPT_TIMEOUT => is_numeric($options['timeout'] ?? null) ? (int)$options['timeout'] : 5,
-	CURLOPT_ENCODING => '',	//Enable all encodings
+	CURLOPT_ACCEPT_ENCODING => '',	//Enable all encodings
 	CURLOPT_HTTPHEADER => [
 		'Connection: close',
 	],
--- a/lib/composer.json
+++ b/lib/composer.json
@@ -14,7 +14,7 @@
 		"marienfressinaud/lib_opml": "0.5.1",
 		"phpgt/cssxpath": "v1.4.0",
 		"phpmailer/phpmailer": "7.0.1",
-		"simplepie/simplepie": "dev-freshrss#e7b26b4f01d377dc8174d5d4aee961604534d065"
+		"simplepie/simplepie": "dev-merge-mmeier86"
 	},
 	"config": {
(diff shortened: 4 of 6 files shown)


Alkarex:
Thanks!
Let's focus this PR on updating to `CURLOPT_ACCEPT_ENCODING` and follow up in another one.

The same renaming should be proposed upstream to https://github.com/simplepie/simplepie if you feel like it

mmeier86:
I've adapted the description to focus on replacing `CURLOPT_ENCODING` only.

In contrast to the SimplePie PR, I did not introduce additional `if` conditions checking the cURL version, as FreshRSS already requires PHP 8, and PHP 8+ [requires cURL 7.29 already](https://www.php.net/manual/en/curl.requirements.php).

Alkarex:
Very good. Please check my suggestion https://github.com/simplepie/simplepie/pull/960#pullrequestreview-3621961424 and I will make some merging

Alkarex:
Merging in progress:
* https://github.com/FreshRSS/simplepie/pull/66

Alkarex:
* https://github.com/FreshRSS/simplepie/pull/67

Alkarex:
Merged!

Issue #6643, opened 2024-07-20
[Feature] Add more options in 'Mark as read' drop-down menu


arseru:
**Is your feature request related to a problem? Please describe.**
The drop-down menu has 2 options for marking items as read older than: one day, or one week.
![freshrss](https://github.com/user-attachments/assets/ca30e3a0-7d89-4b33-8ac9-87e55a43020d)

**Describe the solution you’d like**
I'd like to request if it would be possible for the developers to add more options to that menu, e.g. similar to what I can find in inoreader:
![inoreader](https://github.com/user-attachments/assets/d8947b5c-a8f8-4e94-9730-3ebea6c271a5)

**Describe alternatives you’ve considered**
Maybe an alternative could be to only add a new menu item where you can enter the number of custom days (`Older than N days`), but I'm more in favor of the approach above.

**Additional context**
I just wanted to use this section to thank the developers of FreshRSS for all their work - I've tried many different self-hosted news aggregators and FreshRSS is by far my favorite! I'd like to contribute code in the future, but I'm no PHP expert yet :see_no_evil:

Alkarex:
Hello and thanks for your message.
Regarding your question, PR welcome. In the meantime, you can save a *user query* to search for articles older than some days and then mark them as read. Example: `!date:P30d` (articles not in the last 30 days)
https://freshrss.github.io/FreshRSS/en/users/10_filter.html#with-the-search-field

arseru:
Nice, I didn't know I could do this and it does the trick, thanks a lot!

I'm closing the issue to avoid adding extra backlog to the project, but once I find the time I'll try to create a PR for it myself :)

Alkarex:
* https://github.com/FreshRSS/FreshRSS/pull/8383

PR #8383, opened 2026-01-01
Add more time options to Mark as Read dropdown

Raineer:
Adds options for: Older than 2 days, Older than 3 days, Older than 1 month
Implements #6643 (Already closed, but not implemented)

- Added strings, menu options, logic to the Mark as Read dropdown
- I chose the options based on my own use cases of trying to quickly clean up an unwieldy RSS queue, without resetting it too much

1. Accumulate RSS entries over 1+ months, ideally with entries each day
2. Select Mark as Read with one of the new entries
3. Mark as Unread to quickly redo the test with the other entries

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated - Did not find this Mark as Read dropdown mentioned in the documentation.

### DIFF
--- a/app/i18n/en/index.php
+++ b/app/i18n/en/index.php
@@ -66,7 +66,10 @@ return array(
 	'menu' => array(
 		'about' => 'About FreshRSS',
 		'before_one_day' => 'Older than one day',
+		'before_one_month' => 'Older than one month',	// TODO
 		'before_one_week' => 'Older than one week',
+		'before_three_days' => 'Older than three days',	// TODO
+		'before_two_days' => 'Older than two days',	// TODO
 		'bookmark_query' => 'Bookmark current query',
 		'favorites' => 'Favourites (%s)',
 		'global_view' => 'Global view',
--- a/app/i18n/en/index.php
+++ b/app/i18n/en/index.php
@@ -66,10 +66,10 @@ return array(
 	'menu' => array(
 		'about' => 'About FreshRSS',
 		'before_one_day' => 'Older than one day',
-		'before_one_month' => 'Older than one month',	// TODO
-		'before_one_week' => 'Older than one week',
-		'before_three_days' => 'Older than three days',	// TODO
-		'before_two_days' => 'Older than two days',	// TODO
+		'before_one_month' => 'Older than one month',
+'before_one_week' => 'Older than one week',
+		'before_three_days' => 'Older than three days',
+		'before_two_days' => 'Older than two days',
 		'bookmark_query' => 'Bookmark current query',
 		'favorites' => 'Favourites (%s)',
 		'global_view' => 'Global view',
--- a/app/i18n/en/index.php
+++ b/app/i18n/en/index.php
@@ -67,7 +67,7 @@ return array(
 		'about' => 'About FreshRSS',
 		'before_one_day' => 'Older than one day',
 		'before_one_month' => 'Older than one month',
-'before_one_week' => 'Older than one week',
+		'before_one_week' => 'Older than one week',
 		'before_three_days' => 'Older than three days',
 		'before_two_days' => 'Older than two days',
 		'bookmark_query' => 'Bookmark current query',
--- a/app/i18n/fr/index.php
+++ b/app/i18n/fr/index.php
@@ -66,7 +66,10 @@ return array(
 	'menu' => array(
 		'about' => 'À propos de FreshRSS',
 		'before_one_day' => 'Antérieurs à 1 jour',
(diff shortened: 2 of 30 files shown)


app/i18n/en/index.php
```diff
@@ -66,7 +66,10 @@
 	'menu' => array(
 		'about' => 'About FreshRSS',
 		'before_one_day' => 'Older than one day',
+		'before_one_month' => 'Older than one month',	// TODO
 		'before_one_week' => 'Older than one week',
+		'before_three_days' => 'Older than three days',	// TODO
+		'before_two_days' => 'Older than two days',	// TODO
```
Alkarex:
```suggestion
		'before_one_month' => 'Older than one month',
'before_one_week' => 'Older than one week',
		'before_three_days' => 'Older than three days',
		'before_two_days' => 'Older than two days',
```

app/i18n/fr/index.php
```diff
@@ -66,7 +66,10 @@
 	'menu' => array(
 		'about' => 'À propos de FreshRSS',
 		'before_one_day' => 'Antérieurs à 1 jour',
+		'before_one_month' => 'Older than one month',	// TODO
 		'before_one_week' => 'Antérieurs à 1 semaine',
+		'before_three_days' => 'Older than three days',	// TODO
+		'before_two_days' => 'Older than two days',	// TODO
```
Alkarex:
```suggestion
		'before_one_month' => 'Antérieurs à 1 mois',
		'before_one_week' => 'Antérieurs à 1 semaine',
		'before_three_days' => 'Antérieurs à 3 jours',
		'before_two_days' => ''Antérieurs à 2 jours',
```

Alkarex:
Thanks! Those are fine, but let's not add more than those new ones, since more custom actions can be achieved on demand:

https://github.com/FreshRSS/FreshRSS/issues/6643#issuecomment-2241778321
> Hello and thanks for your message. Regarding your question, PR welcome. In the meantime, you can save a _user query_ to search for articles older than some days and then mark them as read. Example: `!date:P30d` (articles not in the last 30 days) https://freshrss.github.io/FreshRSS/en/users/10_filter.html#with-the-search-field

Alkarex (approved):

app/layout/nav_menu.phtml
```diff
@@ -157,15 +157,24 @@
...
+	$mark_before_three_days = $mark_read_url;
 	$mark_before_one_week = $mark_read_url;
+	$mark_before_one_month = $mark_read_url;
 	switch (FreshRSS_Context::$sort) {
 		case 'id':
 			$mark_before_today['params']['idMax'] = $today . '000000';
+			$mark_before_two_days['params']['idMax'] = ($today - 172800) . '000000';
+			$mark_before_three_days['params']['idMax'] = ($today - 259200) . '000000';
 			$mark_before_one_week['params']['idMax'] = ($today - 604800) . '000000';
+			$mark_before_one_month['params']['idMax'] = ($today - 2592000) . '000000';
```
Alkarex, line 170: We should discuss the values and their i18n translations.
*before today* has targetted articles older than midnight of the current day.
We should be clearer whether we keep this behaviour or change to target articles older than 24 hours.

This has an effect on the other values as well.

Frenzie: I wouldn't like making that 24 hours at all. Then it'd be random instead of stable.

Alkarex: If so, then we should update the translations. We could go for something like:
* Older than today
* Older than yesterday
* Older than this week
* Older than last week
* Older than this month
* Older than last month
* Older than this year

But this relies on a calendar concept, which will complexify i18n, in particular regarding definitions of a week, a month, a year...

The simpler alternative would be based on fixed durations, such as:
* Older than 24h
* Older than 2 days
* Older than 3 days
* Older than 7 days
* Older than 30 days

The simpler approach based on fixed durations is the closest to the current PR, with only minor time adjustments and i18n translations.

An hybrid approach could be to keep *Older than today* and then use the fixed durations from *Older than 24h*.

Alkarex: @Raineer Would you feel like making the following update? In any case, we need to change something:

* Older than today
* Older than 24h
* Older than 2 days
* Older than 3 days
* Older than 7 days
* Older than 30 days

Raineer: Yes, I can do this. I'm plenty good with your suggestions. The goal was to just get a little more refined with the options without going the user query route. The balance of options here is important. Thank you for the consideration and feedback.

PR #8400, opened 2026-01-04
Add SSRF mitigations using `filter_var` and `CURLOPT_RESOLVE`

Inverle:
The idea is to prevent FreshRSS from sending any HTTP requests to internal services, except for the ones that are explicitly allowed in the config.

Based on https://github.com/moodle/moodle/blob/6e82b46a480826d1a85394d9e5087f7d82d1dd52/lib/filelib.php#L3818 and https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php


app/Utils/httpUtil.php
```diff
@@ -256,6 +256,42 @@ private static function enforceHtmlBase(string $html, string $href): string {
...
+				'http' => 80,
+				'https' => 443,
+				default => 0,
+			};
+		}
+		$ip = '';
+		if (filter_var($host, FILTER_VALIDATE_IP) !== false) {
+			$ip = $host;
+		} else {
+			$ip = gethostbyname($host);
```
Inverle: This is IPv4 only

app/Utils/httpUtil.php
```diff
@@ -325,102 +361,121 @@ public static function httpGet(string $url, string $cachePath, string $type = 'h
...
-					$errorMessage .= ' We may retry after ' . date('c', $retryAfter);
+			if (in_array($c_status, [301, 302, 303, 307, 308], true)) {
+				// Handle the redirect by making another request
+				$location = \SimplePie\Misc::absolutize_url($headers['location'] ?? $url, $url);
+				if ($location === false) {
+					$location = $url;
 				}
+				$url = $location;
+				$redirs++;
+				continue;
```
Inverle, line 620: Should probably return an error if hit max redirects.

Alkarex:
Since this will be a breaking change, let's target 1.29.0

app/Utils/httpUtil.php
```diff
@@ -325,102 +365,121 @@ public static function httpGet(string $url, string $cachePath, string $type = 'h
...
+				CURLOPT_CONNECTTIMEOUT => $feed_timeout > 0 ? $feed_timeout : $limits['timeout'],
+				CURLOPT_TIMEOUT => $feed_timeout > 0 ? $feed_timeout : $limits['timeout'],
+				CURLOPT_RETURNTRANSFER => true,
+				CURLOPT_FOLLOWLOCATION => false,
+				CURLOPT_ACCEPT_ENCODING => '',	//Enable all encodings
+				//CURLOPT_VERBOSE => 1,	// To debug sent HTTP headers
+				CURLOPT_RESOLVE => [$resolve], // Prevent DNS rebinding
+			]);
+
+			curl_setopt_array($ch, $options);
```
Inverle, line 524: TODO: this may contain `CURLOPT_COOKIE` which should not be sent if redirected to a different host

app/i18n/fr/admin.php
```diff
@@ -89,6 +89,7 @@
 		),
 		'force_email_validation' => 'Forcer la validation des adresses email',
 		'instance-name' => 'Nom de l’instance',
+		'internal-host-allowlist' => 'Internal host allowlist',	// TODO
```
```suggestion
		'internal-host-allowlist' => 'Liste d’adresses internes autorisées',
```

Inverle:
Related: https://github.com/simplepie/simplepie/issues/968

Inverle:
Another implementation worth looking at: https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php

Inverle:
~~TODO: i need to reimplement the same redirection logic in SimplePie's File.php as in httpGet() right now..~~

never mind, my test script was broken it works fine already

Inverle:
and also POST needs to change to GET on redirect

phpcs.xml
```diff
@@ -54,6 +54,7 @@
 		<exclude name="Squiz.Functions.MultiLineFunctionDeclaration.Indent"/>
 		<exclude name="Squiz.Functions.MultiLineFunctionDeclaration.OneParamPerLine"/>
 		<exclude name="Squiz.WhiteSpace.ScopeClosingBrace.ContentBefore"/>
+		<exclude name="Generic.CodeAnalysis.EmptyStatement.DetectedIf"/>
```
Alkarex, line 59: That does not seem needed, is it?

config.default.php
```diff
@@ -228,5 +228,15 @@
...
+
+	# Requests to internal hosts such as 127.0.0.1 are blocked by default
+	# Blocked ranges include:
+	# - 10.0.0.0/8
+	# - 172.16.0.0/12
+	# - 192.168.0.0/16
+	#
+	# Here you can add overrides for particular ip/domain:port combinations
+	# Examples: 127.0.0.1:8080, rssbridge.local:443 etc.
+	'internal_host_allowlist' => []
```
```suggestion
	'internal_host_allowlist' => [],
```

Minor preference

app/Utils/httpUtil.php
```diff
@@ -256,6 +270,101 @@ private static function enforceHtmlBase(string $html, string $href): string {
...
+				return null;
+			}
+			// Extra check because the above one might not be enough: https://github.com/php/php-src/issues/16944
+			// Workaround is available by using `FILTER_FLAG_GLOBAL_RANGE` instead, but that was only added in PHP 8.2, and we need to support PHP 8.1+
+			foreach (self::PRIVATE_SUBNETS as $cidr) {
+				if (self::checkCIDR($ip, $cidr)) {
+					return null;
+				}
+			}
+
```
Alkarex, line 385: It would be nice to include the version for PHP 8.2+ already, so that we can just remove the fallback later

app/Utils/httpUtil.php
```diff
@@ -256,6 +270,101 @@ private static function enforceHtmlBase(string $html, string $href): string {
...
+			};
+		}
+		$resolve_str = "$host:$port:";
+		$ips_ok = [];
+		$ips = [];
+		$records = [];
+		if (filter_var($host, FILTER_VALIDATE_IP) !== false) {
+			$ips[] = $host;
+		} else {
+			$records = dns_get_record($host, DNS_A + DNS_AAAA);
```
Alkarex: I am a bit worried of introducing such low-level networking functions.
Might be necessary, but we will need to make sure that the function works even when the call fails, and to test on a few different platforms.

Alkarex: @math-GH If I remember correctly, you have some deployments on Windows, don't you? It would be nice if you could test this PR on your Windows setup.

phpcs.xml
Inverle: Not anymore, but I would prefer to keep it

app/Utils/httpUtil.php
Inverle: See https://github.com/php/php-src/issues/16944#issuecomment-2501186795

The original bug is now fixed in recent PHP patch versions, but even on the latest Ubuntu 24.04 LTS, there is only 8.3.6: https://repology.org/project/php/versions#ubuntu_24_04, while the bug was fixed in 8.3.16

If we drop PHP 8.1 at some point, then the fallback might not be needed, depending on what version range we'll be supporting. (8.4.x had it fixed in 8.4.3, 8.5.x didn't have the bug at any point)

Inverle: Though maybe it is better to only use this check

app/Utils/httpUtil.php
Inverle: Well, this PR in its current state may cause issues with using SOCKS5H proxies since `dns_get_record()` won't go through the proxy.
The SSRF check can probably be skipped altogether if a proxy is being used.

Inverle: > The SSRF check can probably be skipped altogether if a proxy is being used.

00bd6323a7aa26154ca7a06924650b972322f766
(was meant to be `empty()` not `!empty()` - oops)

lib/simplepie/simplepie/src/File.php
```diff
@@ -109,6 +109,17 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
                 $headers = [];
             }
             if (!$force_fsockopen && function_exists('curl_exec')) {
+                $resolve = false; // FreshRSS
+                if (empty($curl_options[CURLOPT_PROXY] ?? null)) { // FreshRSS
+                    $resolve = \FreshRSS_http_Util::getCurlResolveInfo($url);
```
Alkarex: We should try to avoid backward dependencies from the SimplePie library to FreshRSS.
In other situations, we have used a hook or inheritance

Inverle: Makes sense, but what kind of hook are you suggesting?

Alkarex: I have made a suggestion (not tested):
https://github.com/FreshRSS/FreshRSS/pull/8400/commits/bd59b50e768678cffdea52259eb613dc5939af49

app/Models/SimplePieFetch.php
```diff
@@ -25,9 +25,27 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
...
 	}
 
+	#[\Override]
+	protected function is_url_resolve_allowed(string $url): bool {
+		$resolve = \FreshRSS_http_Util::getCurlResolveInfo($url);
+		if ($resolve === null) {
+			Minz_Log::warning('Fetching that URL is not allowed, because the host’s IP is not on the allowlist: ' . \SimplePie\Misc::url_remove_credentials($url));
+			return false;
+		} elseif ($resolve === false) {
+			return false;
```
Alkarex: I am worried about this line. Might be better true by default

app/Utils/httpUtil.php
```diff
@@ -256,6 +269,101 @@ private static function enforceHtmlBase(string $html, string $href): string {
...
+		}
+		$port = parse_url($url)['port'] ?? null;
+		if ($port === null) {
+			$port = match ($scheme) {
+				'http' => 80,
+				'https' => 443,
+				default => 0,
+			};
+		}
+		$resolve_str = "$host:$port:";
```
Alkarex, line 333: I think it is necessary to cache the results between calls to this function (e.g. using `static $resolve_ok = []` or something like that) to limit the performance degradation, and to avoid potentially expensive DNS calls depending on system configurations

lib/simplepie/simplepie/src/File.php
Inverle: See https://github.com/FreshRSS/FreshRSS/pull/8400/commits/14e7cf31eb0939d3682375fae31d8a72a2d19a82

app/Utils/httpUtil.php
Inverle: I was wondering about that myself, but I did not observe any performance regression on my setup.

But if there were a performance impact, then I'm not sure if storing the cache in an array would be enough, since it would reset for every feed update? (since every feed update is one request when updating from Web UI)

app/Models/SimplePieFetch.php
Inverle: Why `true`?

Alkarex: Like in not breaking everything if for some reason our DNS resolution is not working

app/Utils/httpUtil.php
Alkarex: There can be thousands of calls to the same host:port during a given refresh session

Inverle: During a refresh session yes, but during one update request not necessarily unless there are a lot of redirects or HTML pages being fetched and scraped (more likely).

So I agree it would be good to have a DNS cache even if not persistent across multiple update requests.

app/Models/SimplePieFetch.php
Inverle: I'm not sure that would be entirely safe. The `dns_get_record()` call could fail randomly (on some configurations, maybe?), let's say, perhaps, and then cURL could potentially resolve a domain to 127.0.0.1, and the check would be bypassed.

also `[]` not `true` since `CURLOPT_RESOLVE` doesn't accept booleans, a mistake I made earlier

app/Utils/httpUtil.php
Alkarex: (The scenario I am thinking about is a cron job to refresh many feeds, potentially many from a the same domain, and with many additional scraping)

app/Models/SimplePieFetch.php
Alkarex: Right, then I think a configuration option to disable those checks would be good, as I am afraid it will break some setups.
Even better, it could also be a part of the existing syntax: for instance something like allowing `::/0` or similar would skip the DNS lookups.

app/Utils/httpUtil.php
Inverle: Done and the other suggestion too

Alkarex:
Conflict due to https://github.com/FreshRSS/FreshRSS/pull/8628

lib/simplepie/simplepie/src/File.php
```diff
@@ -179,7 +194,8 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
                     if ($parser->parse()) {
                         $this->set_headers($parser->headers);
                         $this->body = $responseBody;
-                        if ((in_array($this->status_code, [300, 301, 302, 303, 307]) || $this->status_code > 307 && $this->status_code < 400) && ($locationHeader = $this->get_header_line('location')) !== '' && $this->redirects < $redirects) {
+                        if ((in_array($this->status_code, [300, 301, 302, 303, 307]) || $this->status_code > 307 && $this->status_code < 400) &&
+                            ($locationHeader = $this->get_header_line('location')) !== '' && ($this->redirects < $redirects || $redirects === -1)) { // FreshRSS: added infinite redirects for -1
```
Alkarex: What is the case for the infinite redirects?

Alkarex: Note: cURL internal default seems to be 30 https://curl.se/libcurl/c/CURLOPT_MAXREDIRS.html

Inverle: > What is the case for the infinite redirects?

This is something that is currently supported, and I want to keep the previous behavior:

image

Inverle:
Fixed the conflicts

config.default.php
```diff
@@ -228,5 +228,18 @@
...
+	# - 10.0.0.0/8
+	# - 172.16.0.0/12
+	# - 192.168.0.0/16
+	#
+	# Here you can add overrides for particular ip/domain:port combinations
+	# Examples: 127.0.0.1:8080, rssbridge.local:443 etc.
+	#
+	# CIDR is permitted too
+	# Examples: 0.0.0.0/0, ::/0 (adding both disables SSRF checks entirely)
+	'internal_host_allowlist' => [],
```
Alkarex, line 260: It would be nice to add a pointer to those options in the documentation
https://freshrss.github.io/FreshRSS/en/admins/09_AccessControl.html

lib/simplepie/simplepie/src/File.php
```diff
@@ -320,6 +335,16 @@ protected function on_http_response($response, array $curl_options = []): void
...
     }
 
+    /**
+     * Event to allow inheriting classes to control fetching certain URLs.
+     * @param string $url
+     * @return array<string>|null|false A value for CURLOPT_RESOLVE as an array, null if a disallowed IP address was found in DNS records, false if the domain failed to resolve
+     */
+    protected function get_curl_resolve_info(string $url): array|null|false
+    {
+       return [];
```
Alkarex: One of the several reasons I do not like spaces: here there are 7 instead of 8.
Patch on its way along with more substantial things than this detail :-)

Alkarex:
@Inverle I have made some proofreading and corresponding changes. Please double-check

app/Utils/httpUtil.php
```diff
@@ -449,17 +451,17 @@ public static function httpGet(string $url, string $cachePath, string $type = 'h
...
 		}
 		while (true) {
 			$url = is_string($url) ? $url : '';
 			$resolve = [];
-			if (empty($options[CURLOPT_PROXY] ?? null)) {
+			if ($proxy === '') {
 				$resolve = self::getCurlResolveInfo($url);
 				if ($resolve === null) {
-					Minz_Log::warning("Fetching $url is not allowed, because the host's IP is not on the allowlist.");
+					Minz_Log::warning("Fetching $url is not allowed, because the host’s IP is not in the allowlist.");
```
Inverle: Shouldn't I have done this with `\SimplePie\Misc::url_remove_credentials($url)`?

app/Utils/httpUtil.php
```diff
@@ -449,17 +451,17 @@ public static function httpGet(string $url, string $cachePath, string $type = 'h
...
 		$redirs = 0;
-		$max_redirs = $options[CURLOPT_MAXREDIRS] ?? 4;
+		$max_redirs = $curl_options[CURLOPT_MAXREDIRS] ?? $options[CURLOPT_MAXREDIRS] ?? FreshRSS_Context::systemConf()->curl_options[CURLOPT_MAXREDIRS] ?? null;
 		if (!is_int($max_redirs)) {
 			$max_redirs = 4;
 		}
 		while (true) {
 			$url = is_string($url) ? $url : '';
 			$resolve = [];
-			if (empty($options[CURLOPT_PROXY] ?? null)) {
```
Inverle: I guess this is fine (`$proxy === ''`) since it is validated that `$proxy` is a string above, so it can't be `false` or `null`

app/Utils/httpUtil.php
```diff
@@ -461,7 +461,7 @@ public static function httpGet(string $url, string $cachePath, string $type = 'h
 			if ($proxy === '') {
 				$resolve = self::getCurlResolveInfo($url);
 				if ($resolve === null) {
-					Minz_Log::warning("Fetching $url is not allowed, because the host’s IP is not in the allowlist.");
+					Minz_Log::warning('Fetching ' . \SimplePie\Misc::url_remove_credentials($url) . ' is not allowed, because the host’s IP is not in the allowlist.');
```
Alkarex: I have a slight preference to put to URL at the end, since it can potentially be very long

Inverle: Done

Inverle:
Not directly relevant to the PR, but why is it done this way here?

https://github.com/FreshRSS/FreshRSS/blob/3fac1bdf201f0831db083ece6bf4c8c56a0ad91a/app/Utils/httpUtil.php#L388-L397

It doesn't seem that `$errorMessage` is being printed anywhere.
The lines were added in https://github.com/FreshRSS/FreshRSS/pull/7760

Alkarex:
Before I forget, we need to implement and document a way to change the settings from environment variable, in particular from Docker Compose, otherwise this PR will break many setups (including several of my own :-)).

Inverle:
So like `TRUSTED_PROXY` but `INTERNAL_HOST_ALLOWLIST`?

Alkarex:
> So like `TRUSTED_PROXY` but `INTERNAL_HOST_ALLOWLIST`?

Yes, something like that indeed. But I am wondering whether the IP is enough or whether we should also allow some hostnames. In container setups in particular, the IPs are not known because they are automatic, so ideally we should allow something like `rss-bridge`.

Inverle:
Not sure why yet, but I'm getting a lot of empty log messages with this PR

image

Inverle:
> Yes, something like that indeed. But I am wondering whether the IP is enough or whether we should also allow some hostnames. In container setups in particular, the IPs are not known because they are automatic, so ideally we should allow something like `rss-bridge`.

Is this not covered already by adding `rss-bridge:443` and `rss-bridge:80` into the allowlist?
or do you mean that the port shouldn't be required?

Alkarex:
> It doesn't seem that `$errorMessage` is being printed anywhere. The lines were added in #7760

Well spotted. It was probably inspired by the following code, but lacking the FreshRSS_Feed_Exception or similar logging...

https://github.com/FreshRSS/FreshRSS/blob/3fac1bdf201f0831db083ece6bf4c8c56a0ad91a/app/Models/Feed.php#L591-L609

Alkarex:
> > Yes, something like that indeed. But I am wondering whether the IP is enough or whether we should also allow some hostnames. In container setups in particular, the IPs are not known because they are automatic, so ideally we should allow something like `rss-bridge`.
> 
> Is this not covered already by adding `rss-bridge:443` and `rss-bridge:80` into the allowlist? or do you mean that the port shouldn't be required?

Fine if providing a list of DNS/hostnames works. I mostly checked the IP parts, and I did not realise it was already there

Inverle:
I have added the environment variable support and documentation.

app/views/configure/system.phtml
```diff
@@ -64,6 +64,31 @@
 			</div>
 		</div>
 
+		<?php
+			$internal_host_allowlist = getenv('INTERNAL_HOST_ALLOWLIST');
+			if (is_string($internal_host_allowlist) && $internal_host_allowlist !== '') {
```
Alkarex, line 69: Changed from `!= 0`. Is this new syntax still covering your intention @Inverle ?

Inverle: Probably yes, I borrowed the code from

https://github.com/FreshRSS/FreshRSS/blob/3fac1bdf201f0831db083ece6bf4c8c56a0ad91a/app/Utils/httpUtil.php#L483-L489

app/i18n/fr/admin.php
```diff
@@ -94,7 +94,7 @@
 		'internal-host-allowlist' => 'Liste d’adresses internes autorisées',
 		'max-categories' => 'Limite de catégories par utilisateur',
 		'max-feeds' => 'Limite de flux par utilisateur',
-		'override-by-env-var' => 'This setting is set by the environment variable <kbd>%s</kbd>.',	// TODO
+		'override-by-env-var' => ' Cette option est définie par la variable d’environnement <kbd>%s</kbd>.',
```
Inverle: There is some whitespace at the beginning of the string, probably not intentional?

app/i18n/fr/admin.php
```diff
@@ -91,8 +91,10 @@
 		'default_closed_registration_message' => 'Ce serveur n’accepte plus de nouvelles inscriptions pour le moment.',
 		'force_email_validation' => 'Forcer la validation des adresses email',
 		'instance-name' => 'Nom de l’instance',
+		'internal-host-allowlist' => 'Liste d’adresses internes autorisées',
 		'max-categories' => 'Limite de catégories par utilisateur',
 		'max-feeds' => 'Limite de flux par utilisateur',
+		'override-by-env-var' => ' Cette option est définie par la variable d’environnement <kbd>%s</kbd>.',
```
```suggestion
		'override-by-env-var' => 'Cette option est définie par la variable d’environnement <kbd>%s</kbd>.',
```

Alkarex:
I have made a Docker image of this PR: `freshrss/freshrss:8400`
Help welcome to test.
@math-GH If you still have a Windows setup, a quick test of this PR would be appreciated

Alkarex:
Testing in Docker (Debian image) on an Ubuntu server:

I observe a few feeds apparently failing to follow 302 redirects, which seems to be reproducible:

```
Retrieved unsupported status code "302" [https://www.technologyreview.com/stories.rss] 
Retrieved unsupported status code "302" [http://www.luc-damas.fr/hop/feed/]
```

I also see several DNS resolution bugs when refreshing by cron (seems better manually, but would need more tests):
```
Failed to resolve domain: https://feeds.feedburner.com/hackaday/LgoM [https://feeds.feedburner.com/hackaday/LgoM] 
Failed to resolve domain: https://www.lemonde.fr/les-decodeurs/rss_full.xml [https://www.lemonde.fr/les-decodeurs/rss_full.xml]
Failed to resolve domain: https://www.dr.dk/nyheder/service/feeds/udland [https://www.dr.dk/nyheder/service/feeds/udland] 
Failed to resolve domain: https://github.com/Alkarex.atom [https://github.com/Alkarex.atom]
Failed to resolve domain: https://blogs.windows.com/msedgedev/feed/ [https://blogs.windows.com/msedgedev/feed/] 
Failed to resolve domain: https://www.60millions-mag.com/rss.xml [https://www.60millions-mag.com/rss.xml]
```

All those domains are working fine when reverting to `edge` branch.

Frenzie:
The CNAME points to another domain. Maybe that's related? Though it doesn't seem to be an immediate problem.

```
$ php -r '$h="feeds.feedburner.com"; $r=@dns_get_record($h, DNS_A+DNS_AAAA); var_dump($r); var_dump($r===false);'
array(2) {
  [0]=>
  array(5) {
    ["host"]=>
    string(17) "www4.l.google.com"
    ["class"]=>
    string(2) "IN"
    ["ttl"]=>
    int(300)
    ["type"]=>
    string(1) "A"
    ["ip"]=>
    string(14) "64.233.184.118"
  }
  [1]=>
  array(5) {
    ["host"]=>
    string(17) "www4.l.google.com"
    ["class"]=>
    string(2) "IN"
    ["ttl"]=>
    int(300)
    ["type"]=>
    string(4) "AAAA"
    ["ipv6"]=>
    string(22) "2a00:1450:400c:c0b::76"
  }
}
bool(false)
```

```
$ host -a feeds.feedburner.com
Trying "feeds.feedburner.com"
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 12873
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;feeds.feedburner.com.          IN      ANY

;; ANSWER SECTION:
feeds.feedburner.com.   256     IN      CNAME   www4.l.google.com.

Received 66 bytes from 127.0.0.1#53 in 0 ms
```

Except for blogs.windows.com, which does its own special thing:

```
$ host -a blogs.windows.com
Trying "blogs.windows.com"
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 33789
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;blogs.windows.com.             IN      ANY

;; ANSWER SECTION:
blogs.windows.com.      300     IN      HINFO   "RFC8482" ""

Received 56 bytes from 127.0.0.1#53 in 68 ms
```

Alkarex:
@Inverle Are you able to reproduce the issues on your side?

Inverle:
@Alkarex I may have fixed the issues with a9c17ecfcdaac10c756f0dbe60fadfd0d675e5ed, please retest (at least the 302 ones)

Inverle:
I'm still awaiting feedback, since I need to confirm that the DNS bugs were resolved too.
On my side, they appear to be fixed now. Maybe we can fix the conflicts and have this merged into edge for more testing?

Inverle:
I have also found that the empty log message issue https://github.com/FreshRSS/FreshRSS/pull/8400#issuecomment-4179851083 is due to this line (so it's unrelated to this PR):
https://github.com/FreshRSS/FreshRSS/blob/0376d402b033d14a34661b968b29d165bacc8fbc/app/Models/Entry.php#L1102

Alkarex:
I have just started testing again with one of my instances 🤞🏻
To avoid further delays, I would like to release 1.29.0 in the coming days (plenty of changes already and long time since last release) but without this PR, and merge this PR into edge shortly thereafter to give it a chance to be tested more largely (I am still worried of potential regressions).

Alkarex:
Found one issue already (would require a bit more debugging to be sure of the details):

Requests to the same server are failing even when using the public address.

From inside the Docker container running FreshRSS, a manual cURL returns a public IP, but those URLs are failing with the current PR.

Alkarex:
> Found one issue already (would require a bit more debugging to be sure of the details):
> 
> Requests to the same server are failing even when using the public address.
> 
> From inside the Docker container running FreshRSS, a manual cURL returns a public IP, but those URLs are failing with the current PR.

There is visibly something wrong with the DNS resolution. From inside the container:

```sh
# curl -v https://alexandre.alapetite.fr/blog/rss.rdf.fr.xml
* Host alexandre.alapetite.fr:443 was resolved.
* IPv6: 2001:41d0:a:67e6::1
* IPv4: 37.187.114.230
```

Manually running `dns_get_record` in CLI with PHP inside the container correctly returns (although IPv6 seems to be missing):
```json
[{"host":"srv4.alapetite.fr","class":"IN","ttl":3352,"type":"A","ip":"37.187.114.230"}]
```

But in this PR, the `$ip` variable gets set to `127.0.1.1` (which is not correct) when it arrives at:

```
if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE) === false) {
```

Will need more debugging later.

Alkarex:
I am not always getting the same responses from `dns_get_record` apparently, so sometimes it works, sometimes not.

I am not sure when the `127.0.1.1` case happens:

```json
[{"host":"srv4.alapetite.fr","class":"IN","ttl":0,"type":"A","ip":"127.0.1.1"},
{"host":"srv4.alapetite.fr","class":"IN","ttl":1759,"type":"AAAA","ipv6":"2001:41d0:a:67e6::1"}]
```

When using cURL and multiple IPs are returned, cURL will try one and then the next one if the first fails. I do not think this PR does that, which will likely break several cases.
For instance in the above response, we should have tried the IPv6.

Inverle:
> When using cURL and multiple IPs are returned, cURL will try one and then the next one if the first fails. I do not think this PR does that, which will likely break several cases.

See the comment above the `getCurlResolveInfo()` function:
https://github.com/Inverle/FreshRSS/blob/46d5a2773893ba86d7371584c33543e67e8d44ef/app/Utils/httpUtil.php#L286

> * Returns a value for CURLOPT_RESOLVE as an array, null if a disallowed IP address was found in DNS records, false if the domain failed to resolve

Basically in this case, a potential fix would be to just exclude the disallowed IPs from the final result, instead of erroring out instantly when a disallowed IP is found.
An error would only be returned if there were no allowed IPs found in the DNS records.

However, it seems strange that `127.0.1.1` is randomly received from DNS like that.

Frenzie:
I'd say it's quite common. (Except not random obviously, but very much on purpose.)

> Returns a value for CURLOPT_RESOLVE as an array, null if a disallowed IP address was found in DNS records, false if the domain failed to resolve

Yes, that behavior is clearly against expectations. The entire point  of multiple records is that there are multiple records. In order to spread the load for example you'll return a couple of options from a larger pool. If the client can't connect with one  of the given options or it doesn't like it for some other reason it should try the other option.

Alkarex:
Thanks @Inverle , the logic seems improved already

Alkarex:
@Inverle I have added some more help messages, and also added a `*` syntax to disable the `CURLOPT_RESOLVE` part (I am still worried this part of the code may lead to regressions, so I think it is good to have a mechanism to disable it). Could you please double-check?

image

Inverle:
> Could you please double-check?

Looks good

Alkarex:
* https://github.com/FreshRSS/simplepie/pull/76

Alkarex:
Ah, we lack updating `Docker/README.md` and `Docker/freshrss/docker-compose.yml`
I will get back to it tomorrow if you do not do it before

app/Utils/httpUtil.php
```diff
@@ -332,115 +477,170 @@ public static function httpGet(string $url, ?string $cachePath = null, string $t
...
+					unset($options[CURLOPT_POSTFIELDS]);
+					if (is_array($options[CURLOPT_HTTPHEADER] ?? null)) {
+						$options[CURLOPT_HTTPHEADER] = array_filter($options[CURLOPT_HTTPHEADER], fn(mixed $header): bool =>
+							!str_starts_with(
+								strtolower(trim(is_string($header) ? $header : '')),
+								'content-type:'
+							));
+					}
+				}
+				$url = $location;
```
Inverle, line 619: Something interesting: I noticed that SimplePie does not implement this POST state/cookie unsetting as this code (the SimplePie redirect code instead of `CURLOPT_FOLLOWLOCATION`), which means the same does not happen for normal RSS feed types currently. It does apply if you set the feed type to *JSON* for example, since it goes through the `httpGet()` function.

The exact same behavior is present on edge branch without this PR merged.

I have to manually import a RSS OPML into the edge version with `frss:CURLOPT_FOLLOWLOCATION` set to `1` to stop the bug from happening (it seems that SimplePie redirects are used instead of the intended curl ones in this case)
I can also force the `frss:CURLOPT_FOLLOWLOCATION` setting by changing *Max HTTP redirects* to something other than 0.

On this PR with the imported OPML, it will continue sending the redirects through SimplePie (which is currently broken) as intended.

Inverle: But even this code has issues though, if `Cookie` is set in the headers instead of `CURLOPT_COOKIE`, it won't be unset.
The `Authorization` header won't be either.

More reference:
https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html
> By default, libcurl only sends Authorization: or explicitly set Cookie: headers to the initial host given in the original URL, to avoid leaking username + password to other sites. [CURLOPT_UNRESTRICTED_AUTH](https://curl.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html) is provided to change that behavior.

> Due to the way HTTP works, almost any header can be made to contain data a client may not want to pass on to other servers than the initially intended host and for all other headers than the two mentioned above, there is no protection from this happening when libcurl is told to follow redirects. 

edit: fixed in `httpGet()`, now SimplePie needs to be fixed
edit2: SimplePie fixed

Inverle:
After merging https://github.com/FreshRSS/simplepie/pull/78, we can sync SimplePie and merge here, if you consider the PR ready now.

Inverle:
Probably this part still needs a SSRF check?

https://github.com/FreshRSS/FreshRSS/blob/7cbf597585158c89a51f01730e37ed8eb9b712e0/app/Models/Feed.php#L1475-L1489

Inverle:
My untested attempt at adding a SSRF check: https://github.com/FreshRSS/FreshRSS/pull/8400/commits/83758e6cf922603db7abc91e5358ea09f12f3016

If this is fine, then I can do the same for https://github.com/FreshRSS/FreshRSS/blob/7cbf597585158c89a51f01730e37ed8eb9b712e0/app/Controllers/updateController.php#L216-L218 later, just so we use `httpGet()` everywhere, not for preventing SSRF.

app/Models/SimplePieFetch.php
```diff
@@ -25,9 +25,22 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
 			}
 		}
 
+		$redirects = $curl_options[CURLOPT_MAXREDIRS] ?? null;
+		if (!is_int($redirects)) {
+			$redirects = 4;
+		} elseif ($redirects < 0) {
+			$redirects = -1; // infinite redirects
+		}
+		unset($curl_options[CURLOPT_FOLLOWLOCATION]); // Always use the custom SimplePie redirects for security
```
Alkarex: We used to allow `CURLOPT_FOLLOWLOCATION` (also in OPML export/import).
It would be nice to preserve some of it:
* falsy values should be ok and would probably better be preserved
* null/unset are obviously ok
* truthy is what we want to avoid, but open to discussion maybe acceptable for admin users?

Alkarex: Suggestion https://github.com/FreshRSS/FreshRSS/pull/8400/commits/cfbab689ff5410589cf123c218aed7d6a3eabb5a

Alkarex:
We need to make sure we do not re-introduce a similar bug than https://github.com/FreshRSS/FreshRSS/pull/1657
I am not sure this still even applies to PHP 8.1+.

Edit: should be fine at a first quick glance

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ See also [the FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases).
 
 ## 2026-XX-XX FreshRSS 1.29.2-dev
 
+* Security
+	* 💥 Disallow access to local networks such as `127.0.0.1` by default, for security reasons (breaking change) [#8400](https://github.com/FreshRSS/FreshRSS/pull/8400)
+		* Selected local networks can be allowed under *System configuration* or using the `INTERNAL_HOST_ALLOWLIST` environment variable
+		* Passing `*` allows all networks like before (unsafe)
```
Inverle, line 10: Is it intentional that a second "Security" section is being added into the changelog?

Alkarex: Fixed

Alkarex (approved):

Inverle:
I wanted to mention that `CURLOPT_PROXY` needs to be restricted too, in addition to how `CURLOPT_URL` is restricted. But I'll do that in a new PR then.

Alkarex:
Communication: https://framapiaf.org/@freshrss/116828770254213489

app/Utils/httpUtil.php
```diff
@@ -256,12 +272,145 @@ private static function enforceHtmlBase(string $html, string $href): string {
...
+				$add_ip = "[$ip]";
+			}
+			foreach ($cidr_allowlist as $cidr) {
+				if (self::checkCIDR($ip, $cidr)) {
+					$ips_ok[] = $add_ip;
+					continue 2;
+				}
+			}
+			if (in_array($allowlist_str, $internal_host_allowlist, true) ||
+				in_array("$host:$port", $internal_host_allowlist, true)) {
```
Inverle, line 370: If `$host` is IPv6, then this is not correct because the squared brackets are cut off earlier, but they're supposed to be entered in the settings.

Inverle: I took a closer look, that part was for domain allowlist, so it's fine
(`$allowlist_str` covers `[$ip]:$port`)

Inverle:
https://github.com/FreshRSS/FreshRSS/pull/8950

PR #8427, opened 2026-01-10
Fix MySQL transliterator_transliterate fallback

Alkarex:
The string syntax of `strtr()` cannot handle mutibytes characters, so need to be rewritten using an array map approach.
Extend the fallback replacements to include the Windows/ISO charsets of the latin languages for which we have a translation.
Found during debugging of https://github.com/FreshRSS/FreshRSS/pull/8425


Frenzie (approved):
Ouch :-)

PR #8436, opened 2026-01-15
fix validator url update

### LINKED ISSUE #8435: [Bug] Check the validity of the feed button uses stale param
### Describe the bug

Image

### To Reproduce

1. Go to subscription management (/i/?c=subscription)
2. Click on any feed's setting
3. Change the feed url (in the above screenshot, from rss.xml to index.xml)
4. Hover on the Check the validity of the feed button
5. See it uses old feed url (rss.xml) as the parameter

### Expected behavior

It should use the new feed url as parameter.

### FreshRSS version

1.28.0

### System information

- Installation type: docker
- Device: macbook air m1
- OS: macOS 15.6.1 (24G90)
- Browser: Dia v1.13.1 (73101) based on Chromium version 143.0.7499.194 

### Additional context

_No response_

johan456789:
Closes #8435

  - update validator links to use the same open-url handler with prefix + encoding
  - ensure the validator link reflects the current #url field value before opening
  - keep existing open-url behavior for other links unchanged

  1. Open feed edit (or add feed) form.
  2. Change the feed URL in the URL field.
  3. Click “Check the validity of the feed” and verify it opens the validator with the updated URL.

  Pull request checklist:

  - [x] clear commit messages
  - [x] code manually tested
  - [ ] documentation updated

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -398,29 +398,15 @@ function close_slider_listener(ev) {
 // overwrites the href attribute from the url input
 function updateHref(ev) {
 	const urlField = document.getElementById(this.getAttribute('data-input'));
-	const url = urlField.value;
-	if (url.length > 0) {
+	const rawUrl = urlField.value;
+	const prefix = this.getAttribute('data-prefix') || '';
+	const shouldEncode = this.getAttribute('data-encode') === '1';
+	const url = prefix + (shouldEncode ? encodeURIComponent(rawUrl) : rawUrl);
+	if (rawUrl.length > 0) {
 		this.href = url;
 		return true;
 	} else {
 		urlField.focus();
-		this.removeAttribute('href');
-		ev.preventDefault();
-		return false;
-	}
-}
-
-function updateValidatorHref(ev) {
-	const urlField = document.getElementById(this.getAttribute('data-input'));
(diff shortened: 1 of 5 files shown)


p/scripts/extra.js
```diff
@@ -395,11 +395,14 @@ function close_slider_listener(ev) {
...
 
-// overwrites the href attribute from the url input
+// updates href from the input value, with optional prefix/encoding
 function updateHref(ev) {
 	const urlField = document.getElementById(this.getAttribute('data-input'));
-	const url = urlField.value;
-	if (url.length > 0) {
+	const rawUrl = urlField.value;
+	const prefix = this.getAttribute('data-prefix') || '';
+	const shouldEncode = this.getAttribute('data-encode') === '1';
```
Frenzie, line 403: When should it not be encoded?

johan456789: buttons like this should open the unencoded one

https://github.com/FreshRSS/FreshRSS/blob/307b527279aa3d3911a5bf411433df4be2093022/app/views/feed/add.phtml#L50

Frenzie: Alright.

Alkarex:
Thanks 👍🏻

PR #8439, opened 2026-01-15
Update documentation in regards to CSP warnings

### LINKED ISSUE #8438: [Bug] Potentially outdated Content-Security-Policy rules
### Describe the bug

It seems like FreshRSS version 1.28.0 might need some updates to its CSP rules.

When opening a FreshRSS instance you can see the follow **warning** in a chromium console:
> Note that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.
>   init_csp_alert	@	main.js?1768484807:2327
>   init_main_afterDOM	@	main.js?1768484807:2352

While firefox might even be blocking some elements due to CSP rules:
> Content-Security-Policy: The page’s settings blocked a JavaScript eval (script-src) from being executed because it violates the following directive: “default-src 'self'” (Missing 'unsafe-eval') main.js:2327:3
> The Notification permission may only be requested from inside a short running user-generated event handler.

Whatever might be using eval based javascript execution I'm not sure though, but it seems to come to FreshRSS core as it seems to be reproducable on https://demo.freshrss.org as well.

### To Reproduce

1. Go to https://demo.freshrss.org
2. Open dev console (eg. CTRL + SHIFT + I)
3. Navigate to browser console
4. See warning

### Expected behavior

No warnings

### FreshRSS version

1.28.0

### System information

See demo instance https://demo.freshrss.org 

### Additional context

_No response_

Hufschmidt:
Closes #8438

- Add a (documentation) reference to the CSP warning caused by the CSP correctness check to the relevant documentation
- Update the documentation to reflect that this browser warning is triggered also by correct CSP configuration

1. Login as admin into FreshRSS instance
2. View developer console (CTRL + SHIFT + I) and look at console warnings
~~3. Clicking on the code that tirggered the warning should now also have the documentation link~~
3. Additional console information should be printed (on valid installs) towards the documentation

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/docs/en/admins/10_ServerConfig.md
+++ b/docs/en/admins/10_ServerConfig.md
@@ -125,7 +125,7 @@ Debug your own CSP header:
 * With DevTools network tab: press F12
 * [CSP Evaluator](https://csp-evaluator.withgoogle.com/)
 
-If you’re aware of the risks or haven’t overwritten the CSP header and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.
+If you’re aware of the risks and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.
 
 Note that FreshRSS already ships with a secure CSP configuration, therefore it’s not necessary to make any adjustments to CSP unless you’re writing an extension.
 
--- a/docs/en/admins/10_ServerConfig.md
+++ b/docs/en/admins/10_ServerConfig.md
@@ -115,17 +115,17 @@ server {
 
 ## Security
 
-Avoid overwriting the [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) header with directives such as `more_set_headers "Content-Security-Policy: ..."`
+Avoid overwriting the [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) header with directives such as `more_set_headers "Content-Security-Policy: ..."` or `add_header 'Content-Security-Policy' '...'`.
 
 ✅ Example of good CSP: `default-src 'self'; frame-ancestors 'self'`
 
 ❌ Bad CSP: `upgrade-insecure-requests`
 
-Debug CSP header:
+Debug your own CSP header:
 * With DevTools network tab: press F12
 * [CSP Evaluator](https://csp-evaluator.withgoogle.com/)
 
-If you're aware of the risks and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`
+If you're aware of the risks or haven't overwritten the CSP header and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.
 
 Note that FreshRSS already ships with a secure CSP configuration, therefore it's not necessary to make any adjustments to CSP unless you're writing an extension.
 
--- a/docs/en/admins/10_ServerConfig.md
+++ b/docs/en/admins/10_ServerConfig.md
@@ -125,8 +125,8 @@ Debug your own CSP header:
 * With DevTools network tab: press F12
 * [CSP Evaluator](https://csp-evaluator.withgoogle.com/)
 
-If you're aware of the risks or haven't overwritten the CSP header and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.
+If you’re aware of the risks or haven’t overwritten the CSP header and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.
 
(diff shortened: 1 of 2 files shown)


p/scripts/main.js
```diff
@@ -2325,6 +2325,7 @@ function init_csp_alert() {
 	try {
 		// eslint-disable-next-line no-new-func
 		Function();
+		// See https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/admins/10_ServerConfig.md#security
 	} catch (_) {
 		// Exit if 'script-src' is set and 'unsafe-eval' isn't set in CSP
```
Frenzie, line 2329: Possibly stick that here as a console.info('If you see an unsafe eval warning everything is working correctly. See https://…')?

Hufschmidt: Good idea! :+1:

Hufschmidt: Done.
Moved to catch() block so those that have seemingly working system can see the message.

This isn't shown to those where the check succeeds, but then FreshRSS will display an alert header instead it seems. :)

Frenzie (approved):
Fine by me (besides that missing semicolon)

p/scripts/main.js
```diff
@@ -2327,6 +2327,7 @@ function init_csp_alert() {
 		Function();
 	} catch (_) {
 		// Exit if 'script-src' is set and 'unsafe-eval' isn't set in CSP
+		console.info("If you see a 'unsafe-eval' warning, everything is working as intended.\nSee https://github.com/FreshRSS/FreshRSS/blob/edge/docs/en/admins/10_ServerConfig.md#security");
```
Inverle:
```suggestion
		console.info("If you see a 'unsafe-eval' warning, everything is working as intended.\nSee https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security");
```

Hufschmidt: Ah yeah. I was a bit iffy on pointing to a document in git, totally forgot about the static website.

Inverle (approved):

docs/en/admins/10_ServerConfig.md
```diff
@@ -115,18 +115,18 @@ server {
...
 
 ❌ Bad CSP: `upgrade-insecure-requests`
 
-Debug CSP header:
+Debug your own CSP header:
 * With DevTools network tab: press F12
 * [CSP Evaluator](https://csp-evaluator.withgoogle.com/)
 
-If you're aware of the risks and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`
+If you’re aware of the risks or haven’t overwritten the CSP header and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.
```
Alkarex: I am not sure to understand the *or haven’t overwritten the CSP header*  part.
That does not seem relevant for the choice, is it?

Hufschmidt: To my understanding the check was meant to ensure that the CSP header send by FreshRSS is not changed/overwritten by the frontend server (eg. Nginx, Apache) in a way that potentially opens a security hole, eg. by enabling `unsafe-eval` for the `script-src` policy.

I'm mostly only familiar with Nginx, so potentially Apache might be different, but to my knowledge you cannot "extend" an upstream header (like CSP header send by FreshRSS) in Nginx, you can only supply your own via the `add_header` directive. This effectively **overwrites** the header send by FreshRSS with your own.

Since I stumbled upon the initial warning and wasn't doing any custom CSP header in my Nginx configuration I wanted to add this part to the documentation to make clear when its safe to ignore the warning when you don't do custom CSP policies.

Maybe the wording could be changed for a bit more clarity to this?
> If you're aware of the risks or _haven't configured any custom CSP policy elsewhere_ and want to ignore the warning shown to admin users, change the `suppress_csp_warning` setting to `true` in `./data/config.php`.

The reasoning being trying to cover these two cases for when its ok to disable the warning [1]:
- You understand CSP header and (mostly) understand the security they provide and the risk accociated.
- Not everyone knows what the CSP header is, but then they likely haven't customized them anyway.

---
[1]: Important: I mean safe to ignore the CSP warning generated by the browser that is simply triggered by the check being blocked by the browser, not the alert shown by/in FreshRSS on a instance with problematic/wrong CSP policy. Maybe that could also the cleared up in the text. Hmmm...

Alkarex: But being overwritten or not is not what is important. What matters is whether the CSP is working or not.
Furthermore, the local configuration might change in the future, so it is good to have a warning when this occurs.
I would be more inclined to just remove the *or haven’t overwritten the CSP header* part, which I find both a bit confusing and not quite accurate.

Alkarex:
Thanks @Hufschmidt 

docs/en/admins/10_ServerConfig.md
Hufschmidt: Good point, thanks for updating and merging. 
Didn't look at my mails on the weekend. :D

PR #8447, opened 2026-01-18
Rewrite setcookie using new syntax

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/2630 now that we have PHP 7.3+ (even PHP 8.1+).
* The new syntax natively supports `samesite`, and also avoids the need of re-setting all parameters.
* Use automatic path instead of own function `getCookieDir()`.

Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8446
* Sanitize lifetime of session cookies from PHP ini to avoid likely invalid/misunderstood values


lib/Minz/Session.php
```diff
@@ -43,19 +43,24 @@ public static function init(string $name, bool $volatile = false): void {
 			return;
 		}
 
-		$cookie = session_get_cookie_params();
-		self::keepCookie($cookie['lifetime']);
+		$params = session_get_cookie_params();
+		// Sanitize lifetime of session cookies from PHP ini `session.cookie_lifetime` (default 0)
+		$params['lifetime'] = ($params['lifetime'] <= 0 || $params['lifetime'] > 86400) ? 0 : $params['lifetime'];
```
Alkarex, line 48: I am wondering whether we should just use `0` here instead of trying to use `session.cookie_lifetime` from `php.ini` (default 0).

Alkarex: Let's come back to that in another PR if desired

Alkarex:
Delete legacy cookies in https://github.com/FreshRSS/FreshRSS/pull/8778

Alkarex:
Follow-up in https://github.com/FreshRSS/FreshRSS/pull/8867

Issue #8442, opened 2026-01-17
Is the a api or end point to connect to push my own feeds?


rupakbajgain:
I am using FreshRSS on daily basis, i find it very useful,
Currently on my setup
I am creating my own setup , i am creating my own feeds on python and giving it to FreshRSS. so i can get anything i want as feed and use less of those media sent one,{fb, insta} as i can filter my own feed.)

It works good on most situation,
Except, (It may seem wierd )
The min update time, most of time it is enough but if i have data i would hope it was at instant(even 30s as i cache in my own machine if it is from external resource). And some infos may slip between frames. I dont want to handle those things.
I hope there was some easy way i can push and update the feeds,

 Or can i manage my own urls from external source.
I mean creating my end-points and adding here can be trouble when i try to add it in bulks. Hope there was easy way to do it.

I dont know where to ask,(since these may be not feature everyone needs),
Any code helps, or any idea, or help me by showing the files i need to tinker with(some hints).

May be these things are already possible, just tell me how can i work with it,
FreshRSS does collect, and manage those feeds , so i was hoping i could just write minimal scripts and get anything i need in one place.

Thank you.

Alkarex:
Hello,
There is already a standard API for instant push updates, which is [WebSub](https://en.wikipedia.org/wiki/WebSub) (formerly [PubSubHubbub](https://github.com/pubsubhubbub/)).
This protocol is supported by e.g. WordPress.
It is basically just sending an HTTP POST with an Atom content.
The endpoint in FreshRSS is `https://freshrss.example.net/api/pshb.php`
In a FreshRSS instance where the API is accessible from Internet (needed also for e.g. mobile apps), this mechanism can be tested with e.g. http://push-tester.cweiske.de/feed.php

The only difficulty is the initial handshake:
1. FreshRSS pulls a feeds, which happens to declare a hub like `<link rel="hub" href="http://pshb.example.net"/>`
2. FreshRSS calls the hub to register for the updates of that feed, which creates a local key at FreshRSS level, valid for all local users.
3. Every time the feed has a new update, the hub pushes by HTTP POST that update to all subscribers

The WebSub protocol is quite simple, so you could have a look already.
And for the specific case of pushing some updates from a local feed, we could try to simplify even further

rupakbajgain:
Ok, In simple i think it would be easier if there were apis to manage the feeds directly.
[...]

PR #8450, opened 2026-01-20
pubSub, if hub have same host it can be reached without being public

rupakbajgain:
Compares host of hub and baseUrl, to detect they can connect each other ofr pubSubHub

Helps #8442

- if host of hub and baseUrl are same then they hub can access freshrss even in localhost

1. Create pubsub server on localhost and connect to RSS instance on localhost.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested (works for me on localhost to localhost communication)


rupakbajgain:
image
Something seems wrong, i am getting 410. feed key not found.

Edit: it feels like old cache issue, if i clear all, it works fine now.

Alkarex:
I have made some fixes. Please test that it still works for you

rupakbajgain:
ok i manually tested it. it is ok.
testing it with docket was hard(maintaining same host, need to be hosted by same system). So i copied that file to my working freshrss.
{isSameHost was minimal test i thought for can the host server connect to me. for self hosted system.}

PR #8456, opened 2026-01-22
Add support for podman

rupakbajgain:

- Detects podman instead of docker if it is available in make

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,15 +7,18 @@ endif
 PORT ?= 8080
 NETWORK ?= freshrss-network
 
+# detect docker or podman
+DOCKER ?= $(shell command -v podman 2>/dev/null || command -v docker)
+
 ifdef NO_DOCKER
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -7,15 +7,18 @@ endif
 PORT ?= 8080
 NETWORK ?= freshrss-network
 
+# detect docker or podman
+DOCKER ?= $(shell command -v podman 2>/dev/null || command -v docker)
```
Alkarex: * Could we put `docker` first?
* Fine as it is, but only if not too inconvenient, it would be nice to have a construct based on `which` to be more similar to detections we do elsewhere such as:
https://github.com/FreshRSS/FreshRSS/blob/bb579956cb38258b1264adaba94e53e14b2b1004/Docker/entrypoint.sh#L48

Alkarex:
Thanks, and sorry for the delay...

PR #8458, opened 2026-01-22
Add option to enable notifications

### LINKED ISSUE #7330: PWA request for notifications
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/7327

<sup>Originally posted by **roughnecks** February 11, 2025</sup>
Hello,

is there a way in FreshRSS PWA to make a request to enable push notifications? Have looked around the app without finding anything relevant.

Thanks</div>

rupakbajgain:
Closes #7330

- Default behavior is same
- Added FreshRSS_Context::userConf()->html5_disable_notif so that, it determines weather user wants notification. (will not show any even it has permission) (i want default false so disable, so old configs get proper default values)
- Added button such that checking it makes it request permission too 
image

- [x] clear commit messages
- [x] code manually tested

- [ ] test notification actually happening (how can i trigger it, do i have to wait it), this code fixes permissions.


app/Controllers/configureController.php
```diff
@@ -75,6 +75,8 @@ public function displayAction(): void {
 			FreshRSS_Context::userConf()->bottomline_link = Minz_Request::paramBoolean('bottomline_link');
 			FreshRSS_Context::userConf()->show_nav_buttons = Minz_Request::paramBoolean('show_nav_buttons');
 			FreshRSS_Context::userConf()->html5_notif_timeout = max(0, Minz_Request::paramInt('html5_notif_timeout'));
+
+			FreshRSS_Context::userConf()->html5_disable_notif = !Minz_Request::paramBoolean('html5_enable_notif');
```
Alkarex: I would prefer a positive parameter such as `html5_notif_enabled` if possible

rupakbajgain: @Alkarex i started with default positive.
But there was small issue, 
Old user config would have their notification turned off, because this variable is false.
I can revert it to old code.

https://github.com/FreshRSS/FreshRSS/pull/8457

Alkarex: Even if set to true in `config-user.default.php`?

Alkarex:
I have made several fixes. Please test again

rupakbajgain:
ok. manually tested it. same behavior.

arseru:
Thanks for adding this! It doesn't work for me though: if I click on "Enable notification" in Configuration => Display menu, it doesn't do anything. If I click it while having the browser Console opened, an error is shown:

> The Notification permission may only be requested in a secure context.

I guess that's because I always access my FreshRSS instance from a local IP, e.g. http://192.168.1.152:8080/i/?c=configure&a=display

Is that expected? Here's my system information:

```
FreshRSS version
    1.29.1
PHP version
    8.4.21
cURL version
    8.14.1
    OpenSSL/3.5.6
Database
    sqlite
    3.46.1
Server software
    Apache/2.4.67 (Debian)
    Linux 6.8.12-20-pve #1 SMP PREEMPT_DYNAMIC PMX 6.8.12-20 (2026-03-13T08:15Z) x86_64 
Browser
    Mozilla/5.0 (X11; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0
```

Inverle:
Yes, you need to put your instance behind HTTPS.

Frenzie:
localhost is also considered a secure context, so another option could be a simple port forward or proxy of some sort.

At its simplest:

```
socat TCP-LISTEN:8080,fork TCP:192.168.1.152:8080
```

Issue #8175, opened 2025-11-01
[Bug] emmbed youtube video error 153


fernandeusto:
### Describe the bug

The emmbed videos with youtube video extension show a 153 error and link to youtube web page

<img width="581" height="184" alt="Image" src="https://github.com/user-attachments/assets/b1054683-9bd6-4f7b-8167-2be00744fb9d" />

 

### To Reproduce

1. Install YouTube Video Feed extension
2. Add youtube chanell to rss
3. Open youtube video post
4. Error 153 is Show

### Expected behavior

View on emmbed frame

### FreshRSS version

1.27.1

### System information

- Installation type: Docker
- OS: MacOS14
- Browser: Firefox, Safari

Inverle:
Please update the extension

See: https://github.com/FreshRSS/Extensions/pull/382

fernandeusto:
I had already updated the extension.

In fact, in the extension file within Docker I can see this code:
<img width="885" height="234" alt="Image" src="https://github.com/user-attachments/assets/65131f5b-bcf7-4dbc-be8b-d316b81c3124" />

But when I open the video from the browser, it still shows this on F12, and I've already cleared the browser cache and tried it on computers where I've never opened FreshRSS before: 

`<iframe class="youlag-iframe" src="https://www.youtube.com/embed/XXXXXX" frameborder="0" allowfullscreen=""></iframe>`

Inverle:
It looks like [youlag](https://github.com/civilblur/youlag) is interfering with the YT extension
Opened a PR to fix this: https://github.com/civilblur/youlag/pull/12

fernandeusto:
That's right, it's been fixed by modifying this line in the User.js extension 

[https://github.com/civilblur/youlag/pull/12/commits/0f1140774662762ac04993676fe8ec64abee8451](url)

Thanks.

jeffxjh:
i don't install youlag,and i see the iframe exist referrerpolicy="strict-origin-when-cross-origin", i also restart service,clear chrome appdata,hard refresh page.but still have problem.

olivluca:
I have the same problem (error 153) but on embedded youtube videos on other feeds (e.g. hackaday or engadget).
I have installed the youtube extension but it doesn't seem to change the iframe for feeds that are not youtube feeds.
I just switched from inoreader where I could see the embedded videos.

This is the iframe from inoreader:
[...]

PR #8467, opened 2026-01-26
add allowfullscreen to iframe

olivluca:
With the "allowfullscreen" attribute it's possible to view an embedded video in full screen.

When a feed has an embedded video that's rendered in an iframe, without the allowfullscreen attribute it's not possible to toggle the video to full screen, adding the attribute makes it possible.

I'm not sure it's necessary to add it to both places, but I tested manually with a couple of feeds and now I can see youtoube videos in full screen without having to go to youtube.

See also [here](https://github.com/FreshRSS/FreshRSS/issues/8175#issuecomment-3790152826)


Alkarex:
Thanks 👍🏻 

olivluca:
No need to for one liner. Thank you for merging it.

PR #8474, opened 2026-01-30
Update Chinese user messages for username validation

xtmd:
Closes #

- Improve Chinese translations

### DIFF
--- a/app/i18n/zh-CN/user.php
+++ b/app/i18n/zh-CN/user.php
@@ -48,7 +48,7 @@ return array(
 		),
 	),
 	'username' => array(
-		'invalid' => '无效用户名',
-		'taken' => '已存在此用户名',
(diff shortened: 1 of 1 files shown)


app/i18n/zh-CN/user.php
```diff
@@ -48,7 +48,7 @@
 		),
 	),
 	'username' => array(
-		'invalid' => '无效用户名',
-		'taken' => '已存在此用户名',
+		'invalid' => '此用户名无效',
+		'taken' => '此用户名已存在',
```
Alkarex, line 52: Ping @yzqzss if you are around, and if you have any feedback on this change

yzqzss: This PR moves 'username'(用户名) to the front aligns better with Chinese grammar. No change in meaning. LGTM.

PR #8478, opened 2026-01-31
Initial conventions for AI agents and humans: AGENTS.md, SKILLS.md, instructions.md

Alkarex:
Not stable yet, but initial structure based on:
* https://agents.md
* https://agentskills.io/specification
* https://code.visualstudio.com/docs/copilot/customization/custom-instructions
* https://code.visualstudio.com/docs/copilot/customization/agent-skills

P.S. This is not meant to favour AI coding, but rather to make sure that when AI tools are used, they are at least used in an efficient way. Note that this is also relevant for humans (both as direct reading or through questions to an AI), and those conventions may be an occasion to clean and restructure our documentation.


Alkarex:
Example of automatic discovery and selection of those instructions:

image

.github/instructions/css.instructions.md
```diff
@@ -0,0 +1,31 @@
...
+	npm run stylelint_fix
+	# or (targeting more than just CSS)
+	make fix-all
+	```
+* Validation can be done with:
+	```sh
+	npm run stylelint
+	# or (targeting more than just CSS)
+	make test-all
+	```
```
Frenzie, line 20: In VS Code at least I believe all of this linting type stuff is done automatically by the tooling around the agents based on config files like `.stylelintrc.json`. Well, the linting anyway, not the fix-all.

Alkarex: If you know where this is defined or how the discovery is done, I am interested, especially if it is based on some kind of standard

Frenzie: I think dev containers may be able to automate the extensions, but indeed my thinking was incomplete. It's simply when you have the relevant extensions installed, that it automatically does it based on whether the config is present.

But in any case I believe the tooling then feeds that back to the agent stuff without it having to run commands and read output.

So your idea is probably best, because it's at least theoretically editor agnostic.

PR #8496, opened 2026-02-08
Fix slider behavior when using navigate back button

### LINKED ISSUE #8488: [Bug] Page scroll locked when navigating back to close slider
### Describe the bug

When triggering a slider to show and then closing it via the browser's back navigation button, the page scroll ends up locked. This is due to the `slider-active` css class remaining on the `html` element.

### To Reproduce

1. Expand a category on the sidebar.
2. Click on the settings icon of any of the listed feeds.
3. In the settings dropdown, click "Manage" to open up the slider.
4. Hit the browser's back navigation button. The URI fragment changes from `#slider` back to `#dropdown-{n}`.
5. The slider is collapsed, but the page scroll is still locked.

### Expected behavior

Allow page scroll when the slider is no longer in view.

### FreshRSS version

1.28.1

### System information

- Browser: Firefox, Chromium (desktop and mobile)

### Additional context

It's possible to reproduce this wherever the slider is used, e.g. via the extensions settings as well.

The use case of using the back navigation button to close the slider might be more prominent on mobile, rather than desktop, but it still affect both cases.

Inverle:
Closes #8488 

* Ensure leave confirmation is shown by preventing navigation using the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API)
* Prevent scroll lock by removing it on `hashchange` instead of `onclick` on the close buttons

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -362,6 +362,18 @@ function open_slider_listener(ev) {
 					f.insertAdjacentHTML('afterbegin', '<input type="hidden" name="slider" value="1" />');
 				});
 				context.ajax_loading = false;
+
+				window.addEventListener('hashchange', () => {
+					if (location.hash.substr(1) === 'slider') {
+						// Triggers when slider is closed
+						window.addEventListener('hashchange', () => {
+							location.hash = 'close';
+							slider.querySelectorAll('form').forEach(function (f) { f.reset(); });
+							document.documentElement.classList.remove('slider-active');
+						}, { once: true });
+					}
+				}, { once: true });
+
 				slider.dispatchEvent(freshrssSliderLoadEvent);
 			};
 			req.send();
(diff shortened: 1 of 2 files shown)


Frenzie (approved):

p/scripts/extra.js
```diff
@@ -362,6 +362,18 @@ function open_slider_listener(ev) {
 					f.insertAdjacentHTML('afterbegin', '<input type="hidden" name="slider" value="1" />');
 				});
 				context.ajax_loading = false;
+
+				window.addEventListener('hashchange', () => {
+					if (location.hash.substr(1) === 'slider') {
```
Frenzie, line 367: Sidenote, ESLint says that `substr()` is deprecated in favor of `substring()`. (Well, actually it annoyingly doesn't add that last part.) Interestingly, `substring()` was already supported by ancient Opera 3 / IE6 / Firefox 1 (likely dating back further to Netscape or at least Mozilla) so I'm not sure what's up there exactly. Maybe a minor distinction between original IE vs NS implementations or something…

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

PR #8509, opened 2026-02-16
Shuffle Sort 2

mdemoss:
This is a re-work of my previous PR https://github.com/FreshRSS/FreshRSS/pull/2463

image

## Notes:

The new `utime` parameter is used to ensure that the sort is stable, but the rules for when it or the similar `rid` parameter get set aren't completely clear.

Only tested on mariaDB so far, and mariaDB after 10.3 has a performance regression.

The random ordering can probably be re-worked to function similarly.

The difference between Shuffle and Random right now is that Shuffle alters the distribution of items and supports scrolling. 

Adjustments were made to how loading in more entries on scroll works.

Added icons for Random and Shuffle options.

The intended function of `sqlListWhere` and `listWhereRaw` could use some clarification. It may make sense to switch over to named parameters for clarity.

## Pull request checklist:

- [x ] clear commit messages
- [x ] code manually tested
- [ ] documentation updated


mdemoss:
Might remove changes I made to docker files for my own testing with mariadb.

mdemoss:
Further note: Switching over to named parameters in the SQL query is probably **mandatory** because in this version I just used string interpolation.

mdemoss:
I see the edge branch has some in-progress work related to a secondary sort order. Possibly it would be easier to merge shuffle after that has settled.

app/layout/nav_menu.phtml
```diff
@@ -91,7 +91,7 @@
 				<?php
 					$classSeparator = '';
 					if (count(FreshRSS_Context::userConf()->queries) > 0) {
-						$classSeparator = ' separator';
+					$classSeparator = ' separator';
```
Alkarex: Please avoid unrelated whitespace changes. Several other cases below

Docker/freshrss/docker-compose-db-maria.yml
Alkarex, line 1: Should probably be removed or moved to another PR.
(Noting that MySQL/MariaDB is the least preferred database, i.e. after PostgreSQL and SQLite)

Docker/freshrss/docker-compose-development.yml
```diff
@@ -2,3 +2,7 @@ services:
   freshrss:
     volumes:
       - ../..:/var/www/FreshRSS
+  freshrss-db:
+    ports:
+      - 3306:3306
```
Alkarex: Unrelated, and works only with MySQL/MariaDB

constants.php
```diff
@@ -6,7 +6,7 @@
 /** @var string */
 const FRESHRSS_MIN_PHP_VERSION = '8.1.0';
 /** @var string */
-const FRESHRSS_VERSION = '1.28.2-dev';
+const FRESHRSS_VERSION = '1.28.1-shuffle';
```
Alkarex: We do not typically change that in PRs, but the previous value comes from the wrong branch. It should be the value from the edge branch `1.28.2-dev`

app/Models/EntryDAO.php
```diff
@@ -1470,119 +1461,106 @@ private function sqlListWhere(string $type = 'a', int $id = 0, int $state = Fres
...
 				in_array($type, ['a', 'A', 'Z', 'i', 's', 'f'], true) &&	// for some of the queries using the feed table,
 				preg_match('/is_read\\s*=\\s*[01]\\b/', $search))	// where is_read is a criteria
 			? 'USE INDEX (entry_feed_read_index) ' : '';
 
-		return [array_merge($values, $searchValues), 'SELECT '
+		return [array_merge($values, $searchValues, $outerSearchValues), 'SELECT '
 			. ($type === 'T' ? 'DISTINCT ' : '')
 			. 'e.id'
 			. ($type === 'T' && $sort !== 'id' ? ', ' . $orderBy : '') // SELECT DISTINCT, ORDER BY expressions must appear in SELECT
+			. ($order === 'SHUF' ? ', (((RANK() over (partition by e.id_feed order by e.id DESC) -1 ) DIV 3) * 8693) + ((709*(e.id+ ' . Minz_Request::serverTimestamp() . ' )) MOD 509) as shuffleOrderKey ' : ' ' )
```
Alkarex: Where do those hardcoded values come from?

Alkarex:
> The difference between Shuffle and Random right now is that Shuffle alters the distribution of items and supports scrolling.

What do you mean by *alters the distribution of items* ?

app/Models/EntryDAO.php
Alkarex: That line does not work with PostgreSQL nor SQLite (e.g., `DIV`, `MOD`)
See e.g. `sqlRandom()` for inspiration

app/layout/nav_menu.phtml
mdemoss: This was done by `make composer-fix`. Possibly something went wrong if it's making unrelated changes.

Docker/freshrss/docker-compose-db-maria.yml
mdemoss: I need this for my local testing, but I agree I should separate it out.

constants.php
mdemoss: That was just for my own local use. It's accurate to the extent it's only been merged that far.

app/layout/nav_menu.phtml
Inverle: Probably something that phpcbf is doing, I've had this happen too before.

app/Models/EntryDAO.php
mdemoss: The values 709 and 509 are meant to be "large enough" primes. If `Minz_Request::serverTimestamp()` changes then the order should change in a way that looks unpredictable. Newer items should come first, so that part is multiplied by a larger number 8693. It might actually be better if that were a power of 2 instead of a prime.

mdemoss:
> > The difference between Shuffle and Random right now is that Shuffle alters the distribution of items and supports scrolling.
> 
> What do you mean by _alters the distribution of items_ ?

If I understand the random sort correctly, it should pick entries uniformly without respect to the feed id. Shuffle is intended to draw from recent entries in each feed, so the distribution will be changed. Feeds with many items will appear relatively less frequently and feeds with few items will appear relatively more frequently.

p/themes/icons/dice-solid.svg
Inverle, line 1: SVGs need to be optimized with https://jakearchibald.github.io/svgomg/

p/themes/icons/shuffle.svg
Inverle, line 1: This one too

mdemoss:
The shuffle feature is now working with PostgreSQL and SQLite. Performance in PostgreSQL seems good. Using `cli/db-restore.php` and `cli/reconfigure.php` I was able to test on my large (77k unread) collection of entries.

app/Models/EntryDAO.php
mdemoss: I changed things to follow Knuth's multiplicative hash method more closely and to use smaller numbers where possible.

Further improvements might be possible by XOR-ing the timestamp (which acts like a seed), but SQLite may not have a straightforward XOR operator.

mdemoss:
Next, I want to integrate it with the normal way of doing pagination instead of making it a special case, if possible.

Issue #678, opened 2014-10-24
Call for testing /dev before 0.9.2 beta


marienfressinaud:
I had planned a new version for the next week (31th) so it's time to test the `dev` branch! With my last commits there are a lot of features which can be broken (but not so much I think :))

I18n fixes (see #655 and #648) will be delayed for the next beta since there are a lot of strings to change. i18n.freshrss.org will need some improvements to improve security and features.

Hope all will be ok and don't hesitate to open new tickets and mentioning this one if you think your problem should be corrected in the coming beta.

Issue #8508, opened 2026-02-16
[Feature] Show time since when a feed has problems


cweiske:
### Is your feature request related to a problem? Please describe.

When a feed has an error, I get the message
> Blast! This feed has encountered a problem. If this situation persists, please verify that it is still reachable.

in the feed details.

### Use case

I am interested how long the feed error persisted to see if it's a temporary problem or a longer one. (Feeds with long running problems can be removed)

### Describe the solution you’d like

The error message should show the age of the issue:
> Blast! This feed has encountered a problem. If this situation persists, please verify that it is still reachable.
> The last successful update was on 2025-02-25 12:41 (11 months ago)

### Additional context

This is with FreshRSS  1.28.0

cweiske:
The `feed.error` column contains a timestamp since #8646. We can use it to determine if we have errors and when the last successful update was:
- `feed.error`: Date of last feed update error
- `feed.lastUpdate`: Date of last successful feed update

Alkarex:
Good idea @cweiske . PR welcome if you can

cweiske:
I'm currently working on it.

Issue #2185, opened 2018-12-16
Improve WebSub subscriber


Alkarex:
Pass all tests of https://websub.rocks/subscriber , especially when the page is not ATOM/RSS.

math-GH:
Any news here?

Alkarex:
Not yet, except reporting a bug in the test suite https://github.com/aaronpk/websub.rocks/issues/25#issuecomment-1076847556

Alkarex:
https://github.com/aaronpk/websub.rocks/issues/38

Issue #2124, opened 2018-11-13
FreshRSS as a WebSub server


Alkarex:
FreshRSS is already a PubSubHubbub / WebSub client. It could also be a WebSub server.

FreshRSS exposes some RSS feeds for its various views / searches. So other clients can subscribe to feeds generated by FreshRSS. It would be great to also expose WebSub for those FreshRSS-generated feeds, so that the clients would receive push notifications instead of having to rely on pull.

It would allow more efficient registration between FreshRSS instances, or from FreshRSS to WebSub-compatible clients.

Test:
* https://websub.rocks/publisher

Triggered by https://h.kher.nl/@AntoineD/101063442972700071

Alkarex:
Related to https://github.com/FreshRSS/FreshRSS/issues/2051

pattems:
As a part of this, It might be a good time to switch terminology from PubSubHubbub to WebSub.

PR #8510, opened 2026-02-17
Re-add database status in installation check

cweiske:

- The "installation check" page has a check for database validity, which is not shown. It was disabled in 2014 to get a beta out (see https://github.com/FreshRSS/FreshRSS/issues/678). This PR re-enables it and adjusts the `DatabaseDAO::*isCorrect()` methods.

1. Open the gear settings menu on the top right
2. Click "Installation check"
3. Scroll down to see the "Database" check results

I tested it on SQLite.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Related: https://github.com/FreshRSS/FreshRSS/discussions/3873

----

If this functionality is not desired, I can make a patch that completely removes the "DatabaseDAO::*Correct()"  methods, since they are not used elsewhere.


Frenzie:
I think it says it was disabled for the scenario where you have too many columns (let's say because an extension added an `xx_my_fun_extension` column). What would happen in that scenario?

cweiske:
As soon as the number of columns in the database is different from the expected number, it will fail and show a "database table xxx is incomplete".

Frenzie:
Pinging @Alkarex
I'd say extra columns aren't an issue; only missing base columns are.

cweiske:
I've modified the code so that `DatabaseDAO::checkTable()` now only checks if at least the expected columns exist, and allows additional columns provided by extensions.

Frenzie:
Sounds good! 👍

cweiske:
(I only rebased)

cweiske:
I've resolved the conflicts against the edge branch.

Alkarex:
A bit more complex than the current PR, but it would be interesting to have a more precise check, including indices, column types and lengths

cweiske:
My original desire was to implement a solution for #8508, and while I was hacking I saw that this old feature here was still in the code but hidden. I decided to invest some time to get that working again, and did so.

If you'd like to expand the scope to an exact database schema comparison - and not only if a column exists or not - then a totally different approach is needed:
- Provide a way for extensions to communicate which columns and indices they require in which tables
- Merge extension configuration + FreshRss table schemas into one aggregate schema
- Compare that aggregated schema with the actual database schema

Having database schema in two versions - as .sql file and php-readable - is not desirable, so we'd either have to parse the .sql files ourselves (as the TYPO3 CMS does), or build a new schema migrator that only works on php-defined table schemas, and we could drop the .sql files completely. Doctrine might have something for that, but I don't know yet.

But this all is quite some effort (not to mention the problem that column `content_bin` in MySQL is called `content` in SQLite + PostgreSQL, which makes proper schema handling a hassle) and is not just "a bit more complex".

Please decide how you'd like to see this going forward:
1. Merge it as it is
2. Request the complex approach, with all the tasks I've listed here.
3. Request to remove this feature completely

I'll happily make a patch for 3, but I won't do 2.

cweiske:
I've fixed the phpstan errors.

Alkarex:
Thanks again :-)

And by the way, are you the person behind http://push-tester.cweiske.de ? We refer to that service in our documentation and I use it myself once in a while

cweiske:
Yep, that's me :)

Alkarex:
> Yep, that's me :)

Very nice, thanks for that! Please tell if you observe anything wrong regarding WebSub coming from FreshRSS.

And I would be interested in adding WebSub publisher (+ maybe hub) capabilities in FreshRSS, to supplement our existing RSS re-sharing features
https://freshrss.github.io/FreshRSS/en/users/user_queries.html#share-your-user-queries

* https://github.com/FreshRSS/FreshRSS/issues/2124
* https://github.com/FreshRSS/FreshRSS/issues/2185

Maybe you could have some feedback when we get that far

PR #8234, opened 2025-11-19
Implement sort order per feed

PeterVavercak:
Adresses [#4761](https://github.com/FreshRSS/FreshRSS/issues/4761)

Changes proposed in this pull request:

- Added number of sorted feeds and associative array for feed sorting option in Context.
- Number of sorted feeds and local sorting option by its index saved into Minz Request Parameters.
- Number of sorted feeds and local sorting options deleted when choosing another Option Of Global Sorting.
- Added option of allowing sorting by feed in configuration.
- Added variable for allowing local sorting in userConf.
- Added function to get feeds by current get in context.
- Added menu button for all individual feed sorting.
- New database options for individual feed sorting in EntryDAO.
- Considered choosing new entries based on chosen load limit.
- Local sorting parameter saved into continuation value in Index Controller.

How to test the feature manually:

1. At the bottom of Reading Configuration menu turn on individual sorting option menu 
2. Choose Sorting by feed option
3. Choose feed at next sorting menu and choose sorting option for that feed

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #8536, opened 2026-02-28
i18n: pl

Inverle:
99% -> 100%

### DIFF
--- a/app/i18n/pl/index.php
+++ b/app/i18n/pl/index.php
@@ -85,14 +85,14 @@ return array(
 		'rss_view' => 'Kanał RSS',
 		'search_short' => 'Szukaj',
 		'sort' => array(
-			'asc' => 'Ascending',	// TODO
+			'asc' => 'Rosnąco',
 			'c' => array(
 				'name_asc' => 'Tytuł kategorii i kanału A→Z',
 				'name_desc' => 'Tytuł kategorii i kanału Z→A',
 			),
 			'date_asc' => 'Data publikacji 1→9',
 			'date_desc' => 'Data publikacji 9→1',
-			'desc' => 'Descending',	// TODO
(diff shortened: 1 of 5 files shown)


app/i18n/pl/index.php
```diff
@@ -85,14 +85,14 @@
 		'rss_view' => 'Kanał RSS',
 		'search_short' => 'Szukaj',
 		'sort' => array(
-			'asc' => 'Ascending',	// TODO
+			'asc' => 'Rosnąco',
```
Inverle, line 88: Where is this used in the UI?

app/i18n/pl/index.php
```diff
@@ -85,14 +85,14 @@
...
 			'c' => array(
 				'name_asc' => 'Tytuł kategorii i kanału A→Z',
 				'name_desc' => 'Tytuł kategorii i kanału Z→A',
 			),
 			'date_asc' => 'Data publikacji 1→9',
 			'date_desc' => 'Data publikacji 9→1',
-			'desc' => 'Descending',	// TODO
+			'desc' => 'Malejąco',
```
Inverle, line 95: This also

Inverle: Only reference I was able to find:

https://github.com/FreshRSS/FreshRSS/blob/41031fc6d3dc1bb43d88566cafb53181c01af234/app/layout/nav_menu.phtml#L227-L237

but... where is `$title` used?

Frenzie: Pinging @PeterVavercak (#8234).

Alkarex: Probably forgotten during some refactoring

Alkarex: Let's merge here first

PR #8538, opened 2026-02-28
Fix only the title being highlighted after hovering a bit on an article

Inverle:
Before:

before

After:

after

Affected themes were:

* Alternative Dark
* Origine
* Pafat

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -777,10 +777,6 @@ kbd {
 	background: var(--background-color-hover);
 }
 
-.flux:not(.current):hover .item .title {
-	background: var(--background-color-hover);
-}
-
 .flux.current:not(.active) {
 	background: var(--background-color-hover);
 }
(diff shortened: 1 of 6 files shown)


Frenzie:
What are the steps to reproduce this issue? I can't confirm.

Inverle:
Hover slightly over the article by moving the mouse cursor slowly from the sidebar towards an article. (see *before* gif)

Frenzie:
No, that highlights the entire row. But perhaps you mean it's only in Origine dark mode or something?

Inverle:
Happens regardless of dark mode in Origine. Both on Firefox and Chromium.

Inverle:
Try maximizing your browser, hiding the sidebar and dragging the cursor all the way to the left and moving it up and down (near the articles).

Frenzie:
https://github.com/user-attachments/assets/f43408ef-7f93-4f74-9c88-b9932ac2cebb

Frenzie:
Are you sure it's not an issue with some extension?

Inverle:
That's because you are hovering unread articles, if you mark one as read then you will be able to reproduce the issue.

Frenzie:
Yes, indeed. Unread articles have a single pixel where they don't highlight the full row.

PR #8541, opened 2026-03-01
Improve pagination styling in logs page

Inverle:
Before:

After:

---

* Made button text actually centered
* Removed italic font
* Added `overflow-x: auto` so page doesn't overflow if there's too many pagination buttons displayed on a small screen

### DIFF
--- a/app/views/helpers/logs_pagination.phtml
+++ b/app/views/helpers/logs_pagination.phtml
@@ -7,59 +7,61 @@
 ?>
 <?php if ($this->nbPage > 1) { ?>
 <nav class="nav-pagination nav-list">
-	<ul class="pagination">
-		<?php
-			/** @var string $getteur from Minz_Paginator::render() */
-			$params[$getteur] = 1;
-		?>
-		<li class="item pager-first">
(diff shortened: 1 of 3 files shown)


app/views/helpers/logs_pagination.phtml
```diff
@@ -7,59 +7,61 @@
...
 <nav class="nav-pagination nav-list">
-	<ul class="pagination">
-		<?php
-			/** @var string $getteur from Minz_Paginator::render() */
-			$params[$getteur] = 1;
-		?>
-		<li class="item pager-first">
-			<a href="<?= Minz_Url::display(['c' => $c, 'a' => $a, 'params' => $params]) ?>">« <?= _t('conf.logs.pagination.first') ?></a>
-		</li>
+	<div class="pagination-wrapper">
```
Frenzie, line 10: That nav-pagination can't do the same things?

Inverle: Has a little `padding-bottom`:

image

Frenzie: Hm, that looks a bit weird on that screenshot really. Shouldn't that be made margin-bottom then?

Alkarex: Let's take remaining optimisations in another PR, as this is already a nice improvement

PR #8542, opened 2026-03-01
Only display scrollbar everywhere if there's an overflow

Inverle:
Replaces all instances of `overflow-{x,y}: scroll` with `overflow-{x,y}: auto`

Empty scrollbars may appear due to usage of `scroll` instead of `auto` on Chromium, even when there is no overflow. As far as I know this does not happen with Firefox.

### DIFF
--- a/docs/_includes/docs.css
+++ b/docs/_includes/docs.css
@@ -35,7 +35,7 @@ aside {
 }
 
 aside > nav.docs {
-	overflow-y: scroll;
+	overflow-y: auto;
 	max-height: 90vh;
 }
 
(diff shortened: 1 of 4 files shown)


Frenzie:
Note that without setting scroll you can easily achieve the Chromium resizing loop of doom.

Edit: but it doesn't look like that should apply to these instances... hopefully.

Frenzie:
To clarify, in Chrome there are many edge cases where no explicitly setting scroll will result in a loop that goes something like:

1. Window has a certain width or height.
2. Some condition triggers (usually some kind of media query but not necessarily)
3. This results in the scrollbar appearing, which cancels the condition.
4. Rinse and repeat ad infinitum.

Setting overflow scroll seemingly unnecessarily is a common way to avoid this Chromium problem.

Supposedly there's a new CSS property to tell Chromium to always reserve the space, but of course that doesn't actually work.

However, all of the above should only apply to media queries and container queries.

PR #8552, opened 2026-03-02
Fix redirect to wrong view after mark as read

### LINKED ISSUE #6509: [Bug] Mark as read redirects to wrong view
### Describe the bug

see "to reproduce"

### To Reproduce

1. Go to global view
2. Click on a feed
3. Scroll down to the big "mark as read" button
4. the normal view will be opened

### Expected behavior

instead of normal view opened I would expect to see the global view

### FreshRSS version

1.24.1-dev

### Environment information

- Database version: SQLite
- PHP version: PHP 8.0

### Additional context

_No response_

Inverle:
Closes #6509

A wrong redirect (to normal view) happened in case of both the big mark as read button on the bottom and the navbar one, for reader and global views.


Frenzie (approved):
lgtm

PR #8553, opened 2026-03-02
Don't mark as read if middle click is outside of article link

### LINKED ISSUE #6451: [Bug] Clicking on mouse wheel out of article's title then on marks as read without opening the article
### Describe the bug

Not sure you can do anything about it but sometimes I misclick and am out of the article's title. When I click back directly on the title correctly, it marks as read but does not open the article at all (you would expect it to be opened on a new tab in the foreground).
chrome_keF1USKIvX

### To Reproduce

1. Click with the mouse wheel on an article's title, but a little off
2. Click again but on it
3. See article being marked as read but not opened

### Expected behavior

Either article still being marked as read but opened on new tab, or article staying unread

EDIT : maybe the solution is rather to make the mouse wheel click work on the whole entry, as it is currently the case when clicking with left mouse click to expand/collapse.

### FreshRSS version

edge

### Environment information

Database version: MariaDB 10.11
PHP version: PHP 8.2
Installation type: git
Web server type: Apache
Device: Laptop
OS: Win11
Browser: Chrome 124

### Additional context

_No response_

Inverle:
Closes #6451

https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event

> The auxclick event is fired at an [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) when a non-primary pointing device button (any mouse button other than the primary—usually leftmost—button) has been pressed and released both within the same element.

> auxclick is fired after the [mousedown](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousedown_event) and [mouseup](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseup_event) events have been fired, in that order.


p/scripts/main.js
```diff
@@ -1515,7 +1515,7 @@ function init_stream(stream) {
 				} else {
 					el.parentElement.click();	// Normal click, just toggle article.
 				}
-			} else if (ev.which == 2 && !ev.ctrlKey) {	// Simple middle click: same behaviour as CTRL+click
+			} else if (ev.type === 'auxclick' && ev.which == 2 && !ev.ctrlKey) {	// Simple middle click: same behaviour as CTRL+click
```
Frenzie: ev.which is deprecated in favor of ev.button

But what concretely is auxclick supposed to do here besides prefilter out the first button?

Inverle: > ev.which is deprecated in favor of ev.button

Yes, I didn't change it for now though

> But what concretely is auxclick supposed to do here besides prefilter out the first button?

So that this doesn't happen:

Peek 2026-03-03 16-53

If you activate auto scroll outside of the article link and then deactivate it on the article link, it will mark the article as read even though the article wasn't opened by the browser.

Here is how auxclick works instead:

Peek 2026-03-03 16-53-2

Frenzie: Sure, that's mouseup vs click. But I meant why auxclick. :-)

MDN suggests there may be some advantages to it regarding opening popups.
> Starting in Firefox 68, the auxclick event is used to trigger the new tab on middle-click action; previously, this had been done with the click event. Apps can prevent middle-click from opening new tabs (or middle-click to paste, if that feature is enabled) by intercepting auxclick on links, and auxclick event handlers can now open popups without triggering the popup blocker.

But I'm not overly fond of "new in 2024" if that's the only reason to use auxclick instead of click.

image

Inverle: `click` doesn't work on middle click, only `auxclick` does.

Chrome and Firefox had support for it for a long time already, Safari added `auxclick` support in 2024: https://caniuse.com/auxclick

Frenzie: I see. I suppose that'll do for now, but in that case I find this whole `stream.onmouseup = stream.onauxclick = …` concept confusing to read due to the inapplicability of some if branches and I'd rather repeat two or three lines of code for clarity.

More important, I think the lines below might trigger twice for the same reason you added the `if (ev.type === 'auxclick'`? (This ties in with the clarity thing. ;-)

Alkarex: All good here, or what is the status?

Frenzie: As I wrote previously, the `if (context.auto_mark_site) {` on line 1527 would fire first on `mouseup` and once again on `auxclick` (directly after `mouseup`), wouldn't it? So it should include if `ev.type === 'auxclick'` check (or the whole function should be refactored into two separate functions but never mind that).

Additionally I imagine it's cheaper to start with `if (context.auto_mark_site && ev.type === 'auxclick' && ev.which == 3) {` thus skipping the `.closest()` stuff, but that aside.

Alkarex: Would one of you two be motivated to implement those last suggestions?

Frenzie: For me it doesn't "mark as read if middle click is outside of article link", so no, as far as I'm concerned nothing should be changed in the first place.

Frenzie: That is, it either opens in the background and marks as read or it doesn't do anything.

Frenzie: Anyway, assuming it's indeed an issue for different configurations, I'd split it up like this for clarity:
```js
	stream.onmouseup = function (ev) {	// Mouseup enables us to catch control+click in IE/Edge
		if (ev.altKey || ev.metaKey || ev.shiftKey || ev.which != 1) {
			return;
		}

		let el = ev.target.closest('.item a.title');
		if (el) {
			if (ev.ctrlKey) {	// Control+click
				if (context.auto_mark_site) {
					mark_read(el.closest('.flux'), true, false);
				}
			} else {
				el.parentElement.click();	// Normal click, just toggle article.
			}
			return;
		}
	};

	stream.onauxclick = function (ev) {	// Auxclick enables us to catch middle click without affecting left click handling
// Whether you want ctrlKey here or not is a bit unclear.
// Currently ctrl+middle click allows you to open without marking as read but maybe that's a bug and not a feature?
		if (ev.altKey || ev.metaKey || ev.shiftKey || ev.which != 2 || ev.ctrlKey) {
			return;
		}

		let el = ev.target.closest('.item a.title');
		if (el) {
			if (context.auto_mark_article) {
				const new_active = el.closest('.flux');
				mark_read(new_active, true, false);
			}
			return;
		}

		if (context.auto_mark_site) {
			el = ev.target.closest('.flux .link > a');
			if (el) {
				mark_read(el.closest('.flux'), true, false);
			}
		}
	};
```

Plain diff that equals this PR, but with changed ctrl+middle click behavior (also marks as read):

<details>

```diff
diff --git a/p/scripts/main.js b/p/scripts/main.js
index 354a0a090..799c2b934 100644
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1504,38 +1504,40 @@ function init_stream(stream) {
                }
        };
 
-       stream.onmouseup = function (ev) {      // Mouseup enables us to catch middle click, and control+click in IE/Edge
-               if (ev.altKey || ev.metaKey || ev.shiftKey) {
+       stream.onmouseup = function (ev) {      // Mouseup enables us to catch control+click in IE/Edge
+               if (ev.which != 1 || ev.altKey || ev.metaKey || ev.shiftKey) {
                        return;
                }
 
                let el = ev.target.closest('.item a.title');
                if (el) {
-                       if (ev.which == 1) {
-                               if (ev.ctrlKey) {       // Control+click
-                                       if (context.auto_mark_site) {
-                                               mark_read(el.closest('.flux'), true, false);
-                                       }
-                               } else {
-                                       el.parentElement.click();       // Normal click, just toggle article.
-                               }
-                       } else if (ev.which == 2 && !ev.ctrlKey) {      // Simple middle click: same behaviour as CTRL+click
-                               if (context.auto_mark_article) {
-                                       const new_active = el.closest('.flux');
-                                       mark_read(new_active, true, false);
+                       if (ev.ctrlKey) {       // Control+click
+                               if (context.auto_mark_site) {
+                                       mark_read(el.closest('.flux'), true, false);
                                }
+                       } else {
+                               el.parentElement.click();       // Normal click, just toggle article.
+                       }
+               }
+       };
+
+       stream.onauxclick = function (ev) {     // Auxclick enables us to catch middle click without affecting left click handling
+               if (ev.which != 2 || ev.altKey || ev.metaKey || ev.shiftKey) {
+                       return;
+               }
+
+               let el = ev.target.closest('.item a.title');
+               if (el) {
+                       if (context.auto_mark_article) {
+                               const new_active = el.closest('.flux');
+                               mark_read(new_active, true, false);
                        }
                        return;
                }
 
                if (context.auto_mark_site) {
-                       // catch mouseup instead of click so we can have the correct behaviour
-                       // with middle button click (scroll button).
                        el = ev.target.closest('.flux .link > a');
                        if (el) {
-                               if (ev.which == 3) {
-                                       return;
-                               }
                                mark_read(el.closest('.flux'), true, false);
                        }
                }
```

</details>

Diff that avoids marking as read on ctrl+middle click:

<details>

```diff
diff --git a/p/scripts/main.js b/p/scripts/main.js
index 354a0a090..3438441b4 100644
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1504,38 +1504,40 @@ function init_stream(stream) {
                }
        };
 
-       stream.onmouseup = function (ev) {      // Mouseup enables us to catch middle click, and control+click in IE/Edge
-               if (ev.altKey || ev.metaKey || ev.shiftKey) {
+       stream.onmouseup = function (ev) {      // Mouseup enables us to catch control+click in IE/Edge
+               if (ev.which != 1 || ev.altKey || ev.metaKey || ev.shiftKey) {
                        return;
                }
 
                let el = ev.target.closest('.item a.title');
                if (el) {
-                       if (ev.which == 1) {
-                               if (ev.ctrlKey) {       // Control+click
-                                       if (context.auto_mark_site) {
-                                               mark_read(el.closest('.flux'), true, false);
-                                       }
-                               } else {
-                                       el.parentElement.click();       // Normal click, just toggle article.
-                               }
-                       } else if (ev.which == 2 && !ev.ctrlKey) {      // Simple middle click: same behaviour as CTRL+click
-                               if (context.auto_mark_article) {
-                                       const new_active = el.closest('.flux');
-                                       mark_read(new_active, true, false);
+                       if (ev.ctrlKey) {       // Control+click
+                               if (context.auto_mark_site) {
+                                       mark_read(el.closest('.flux'), true, false);
                                }
+                       } else {
+                               el.parentElement.click();       // Normal click, just toggle article.
+                       }
+               }
+       };
+
+       stream.onauxclick = function (ev) {     // Auxclick enables us to catch middle click without affecting left click handling
+               if (ev.which != 2 || ev.altKey || ev.ctrlKey || ev.metaKey || ev.shiftKey) {
+                       return;
+               }
+
+               let el = ev.target.closest('.item a.title');
+               if (el) {
+                       if (context.auto_mark_article) {
+                               const new_active = el.closest('.flux');
+                               mark_read(new_active, true, false);
                        }
                        return;
                }
 
                if (context.auto_mark_site) {
-                       // catch mouseup instead of click so we can have the correct behaviour
-                       // with middle button click (scroll button).
                        el = ev.target.closest('.flux .link > a');
                        if (el) {
-                               if (ev.which == 3) {
-                                       return;
-                               }
                                mark_read(el.closest('.flux'), true, false);
                        }
                }
```

</details>

Inverle: > For me it doesn't "mark as read if middle click is outside of article link", so no, as far as I'm concerned nothing should be changed in the first place.

I mean the second middle click while auto-scrolling specifically. Can you not reproduce it?

Frenzie: It looks like I can reproduce it, but in that case the description is wrong.

With autoscroll enabled the first middle click starts autoscroll, the second middle click (or left click) ends it. Doing so somewhere on the right where the article link isn't present doesn't do anything. But ending autoscroll by middle clicking *on* a link marks it as read without opening it.

But either way, the second diff I gave should behave identically to this PR, I hope you agree that it's much clearer, and it's nicer to read than the current code prior to this PR to boot. I've decided I consider the ability to press <kbd>Ctrl</kbd> to open in background while not marking as read a feature, not a bug.

Frenzie (approved):

Alkarex:
Probable regression:
* https://github.com/FreshRSS/FreshRSS/issues/8806

PR #8577, opened 2026-03-07
Create dynamic favicons from SVG instead of PNG canvas

### LINKED ISSUE #4091: [BUG] Favicon broken in Firefox with arkenfox
**Describe the bug**
With Firefox, the FreshRSS Favicon seems to be broken. At least this is what i get in Firefox currently:
grafik

Maybe this hase something todo, with the current profile. In this profile, i use a modified version of arkenfox user.js to prevent fingerprinting. It seems that Firefox has a problem to create the favicon, when some fingerprinting prevent actions are taken.

**To Reproduce**
1. Install FreshRSS on a new instance
2. Open this instance in Firefox
3. Look at the Favion

**Expected behavior**
The instance should use a correct favicon.

**Screenshots**
grafik

- Device: Laptop
- OS: Windows / Linux Mint
- Browser: Firefox 96 / arkenfox-user,js
- FreshRSS version: 1.18.1
- Database version: SQLite
- PHP version: PHP 7.4
- Installation type: manual installation

**Additional context**
I think the reason is the usage ofen arkenfox-user.js. Maybe there could be another way to create the favicion. You can find more about arkenfox at https://github.com/arkenfox

Inverle:
Closes #4091

The dynamic favicon didn't render properly on browsers that have canvas fingerprinting defenses (see before favicon):

This is the reason for using SVG instead of canvas for rendering the dynamic favicon.

---

Here are larger versions of both canvas and SVG implementations so that it can be compared the 
favicon will still look the same as before:

Old canvas rendering (resolution is low because this is a small PNG):

New SVG rendering:

---

More comparisons

Tab on the left is canvas, tab on the right is SVG

image

image

^ If you're wondering why there is spacing between the digits on the first tab - it's because I marked an article as unread, this needs to be fixed too

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -2234,39 +2234,45 @@ function faviconNbUnread(n) {
 		const t = document.querySelector('.category.all .title');
 		n = t ? str2int(t.getAttribute('data-unread')) : 0;
 	}
-	// http://remysharp.com/2010/08/24/dynamic-favicons/
-	const canvas = document.createElement('canvas');
+	const svgBase = document.querySelector('template#dynamic_favicon_base').innerHTML;
 	const link = document.getElementById('favicon').cloneNode(true);
-	const ratio = window.devicePixelRatio;
-	if (canvas.getContext && link) {
-		canvas.height = canvas.width = 16 * ratio;
-		const img = document.createElement('img');
-		img.onload = function () {
-			const ctx = canvas.getContext('2d');
-			ctx.drawImage(this, 0, 0, canvas.width, canvas.height);
-			if (n > 0) {
-				let text = '';
-				if (n < 1000) {
-					text = n;
-				} else if (n < 100000) {
-					text = Math.floor(n / 1000) + 'k';
-				} else {
-					text = 'E' + Math.floor(Math.log10(n));
-				}
-				ctx.font = 'bold ' + 9 * ratio + 'px "Arial", sans-serif';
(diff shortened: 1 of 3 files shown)


polybjorn:
Tested on LibreWolf with privacy.resistFingerprinting enabled — the exact scenario from #4091.

favicon screenshot

And the world is in order again :)

FreshRSS edge, Raspberry Pi / nginx

Frenzie:
> If you're wondering why there is spacing between the digits on the first tab - it's because I marked an article as unread, this needs to be fixed too

That's nicer to read though. ;-)

Frenzie (approved):
I'd written almost the exact same thing as a quick test. Lucky it didn't take long.

Alkarex:
Tested with:
* Firefox/148.0 ✅
* SeaMonkey/2.53.23 ✅
* Edge/145.0.3800.97 ✅

p/scripts/main.js
```diff
@@ -2234,39 +2234,45 @@ function faviconNbUnread(n) {
 		const t = document.querySelector('.category.all .title');
 		n = t ? str2int(t.getAttribute('data-unread')) : 0;
 	}
-	// http://remysharp.com/2010/08/24/dynamic-favicons/
-	const canvas = document.createElement('canvas');
+	const svgBase = document.querySelector('template#dynamic_favicon_base').innerHTML;
```
Alkarex, line 2237: https://github.com/FreshRSS/FreshRSS/issues/8586

PR #8600, opened 2026-03-14
Allow passing `Minz_HookType` as hook name in `registerHook()`

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8036

It's already used as an example here: https://freshrss.github.io/FreshRSS/en/developers/03_Backend/05_Extensions.html#the-hooks-system 
and here:
https://github.com/aledeg/xExtension-WhiteList/blob/ea17f2304c8ecc28c36e70988c27a233093e8f26/extension.php#L7-L11
but it did not work due to PHP throwing an error.


Alkarex (approved):

Alkarex:
Ok, but merge conflict

Inverle:
Fixed

Inverle:
Also corrects `Minz_Controller` as `Minz_ActionController` in the documentation

PR #8606, opened 2026-03-14
Fix wrong navigation buttons layout on Chromium

### LINKED ISSUE #8605: [Bug] bottom left nav buttons issue on recent Chrome browser
### Describe the bug

I am not certain, but it seems like there is an issue with the navigation buttons in recent versions of Chrome. I am not seeing a similar issue in MSEdge, or Firefox when connecting to the same freshrss instance. The 'previous article', 'go up', 'next article' navigation buttons are vertically stacked instead of side-by-side.

### To Reproduce

Open and login to a completely new freshrss container instance using Chrome guest profile so that no browser extensions are loaded and should have a clean and default browser profile.

Image

Using `Chrome Version 146.0.7680.80 (Official Build) (64-bit)`

Started with a completely fresh container instance, and logged in. There were no freshrss extensions installed.

### Expected behavior

_No response_

### FreshRSS version

1.28.1

### System information

- Database version: SQLItte
- PHP version: N/A
- Installation type: Docker using `freshrss/freshrss:latest` aka `freshrss/freshrss@sha256:9100f649f5c946f589f54cdb9be7a65996528f48f691ef90eb262a0e06e5a522`
- Device: Desktop Computer
- OS: Windows 
- Browser: Chrome Version 146.0.7680.80 (Official Build) (64-bit)

### Additional context

_No response_

Inverle:
Closes #8605

Since Chromium Canary 146.0.7669.0 (and Stable 146.0.7680.71) there is some sort of bug that broke this layout:
image

So instead of using `display: table`, we now use `display: flex`, to also modernize the CSS aside from fixing this.

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1938,7 +1938,7 @@ a.website:hover .favicon {
 
 #nav_entries button {
 	background-color: transparent;
-	width: 33.3%;
+	flex: 1;
 	height: 3rem;
 	border: 0;
 	cursor: pointer;
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1924,11 +1924,11 @@ a.website:hover .favicon {
 
 /*=== Navigation menu (for articles) */
 #nav_entries {
-	display: table;
+	display: flex;
+	justify-content: space-evenly;
 	position: fixed;
 	bottom: 0; left: 0;
 	width: 300px;
-	table-layout: fixed;
 	padding-bottom: env(safe-area-inset-bottom);
 	z-index: 50;
 }
@@ -1939,7 +1939,6 @@ a.website:hover .favicon {
 
 #nav_entries button {
 	background-color: transparent;
-	display: table-cell;
 	width: 33.3%;
 	height: 3rem;
 	border: 0;
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1925,7 +1925,6 @@ a.website:hover .favicon {
 /*=== Navigation menu (for articles) */
 #nav_entries {
 	display: flex;
-	justify-content: space-evenly;
 	position: fixed;
 	bottom: 0; left: 0;
 	width: 300px;
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2768,7 +2768,7 @@ html.slider-active {
 	}
 
 	#nav_entries {
-		display: table !important;
+		display: flex !important;
 		width: 100%;
 	}
 
--- a/p/themes/base-theme/frss.rtl.css
+++ b/p/themes/base-theme/frss.rtl.css
@@ -1938,7 +1938,7 @@ a.website:hover .favicon {
 
 #nav_entries button {
 	background-color: transparent;
-	width: 33.3%;
+	flex: 1;
 	height: 3rem;
 	border: 0;
 	cursor: pointer;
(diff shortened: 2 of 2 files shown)


Alkarex:
Buggy when reducing the size of the window (tested with Firefox):
image

Inverle:
> Buggy when reducing the size of the window (tested with Firefox):

Fixed

p/themes/base-theme/frss.css
```diff
@@ -1939,7 +1938,6 @@ a.website:hover .favicon {
 
 #nav_entries button {
 	background-color: transparent;
-	display: table-cell;
 	width: 33.3%;
```
Frenzie: This doesn't seem like a scenario where widths should be hardcoded. The primary conceptual use case here for flex is that we have a bunch of equal buttons, and the layout remains equally divided if we should add or remove one (`flex: 1`). The conceptual use case for table/grid is that we have exactly three buttons/columns.
```suggestion
	flex: 1;
```

Frenzie (approved):
While the change is fine, I'm less enthused about the provided reasoning. Chromium Canary is unstable and full of bugs. Something is broken in there? Well, so what? Let's report it there instead of immediately catering to random bugs in a development build. And the same goes for whether flex is newer than table. Newer doesn't mean better.

As an aside, to be taken together with my other comment, here's the grid equivalent:

```diff
diff --git a/p/themes/base-theme/frss.css b/p/themes/base-theme/frss.css
index ace60c5cf..b181b58e3 100644
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1924,7 +1924,8 @@ a.website:hover .favicon {
 
 /*=== Navigation menu (for articles) */
 #nav_entries {
-       display: flex;
+       display: grid;
+       grid-template-columns: repeat(3, 1fr);
        position: fixed;
        bottom: 0; left: 0;
        width: 300px;
@@ -1938,7 +1939,6 @@ a.website:hover .favicon {
 
 #nav_entries button {
        background-color: transparent;
-       width: 33.3%;
        height: 3rem;
        border: 0;
        cursor: pointer;
@@ -2767,7 +2767,7 @@ html.slider-active {
        }
 
        #nav_entries {
-               display: flex !important;
+               display: grid !important;
                width: 100%;
        }
```

p/themes/base-theme/frss.css
Inverle: > Let's report it there instead of immediately catering to random bugs in a development build. And the same goes for whether flex is newer than table. Newer doesn't mean better.

I wouldn't care about a development build, the same thing happens since stable 146. (that Canary version is when the issue started happening)
Thanks for both of the suggestions, `flex: 1` is indeed better.

Inverle: Something interesting I've just found is that it's enough to remove the `display: table-cell` property from `#nav_entries button` to make the buttons aligned properly again.
This only works well on Chromium though

Inverle:
```html
<!DOCTYPE html><html><head>
<style>
.table {
    display: table;
    width: 300px;
}
.table button {
    display: table-cell;
    border: 0; /* this can also be background-color: transparent */
}
</style>
</head><body>
<div class="table">
<button>a</button>
<button>b</button>
<button>c</button>
</div>
</body></html>
```
If `border: 0` or `background-color: transparent` get added alongside `display: table-cell` then the buttons align vertically in Chromium :thinking: 

image

Even if you set any kind of background color or border color
(this does not happen in Firefox or earlier Chromium)

Frenzie: > I wouldn't care about a development build, the same thing happens since stable 146. (that Canary version is when the issue started happening)

My bad. If it's already in a stable then indeed that's a different matter. Curious that it wasn't caught by automated regression tests.

Issue #7720, opened 2025-07-06
[Bug] Error 403 Forbidden when using Authentik with Nginx


Ay1tsMe:
### Describe the bug

I have tried following the guide to setup authentik with freshrss [here](https://integrations.goauthentik.io/integrations/services/freshrss/) but I get a Error 403 - Forbidden when I change the authentication method to HTTP.

I am running freshrss in a docker-compose behind nginx reverse proxy.

Here are the freshrss logs:
```
freshrss    | fde4:1d15:8139:0:5ed4:2b1d:a9c4:154 - - [06/Jul/2025:15:54:37 +0800] "GET /i/?rid=686a28ab5e008 HTTP/1.0" 403 4333 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"
freshrss    | [Sun Jul 06 15:54:37.091344 2025] [authz_core:error] [pid 39:tid 39] [client fde4:1d15:8139:0:5ed4:2b1d:a9c4:154:0] AH01630: client denied by server configuration: /var/www/FreshRSS/p/themes/manifest.json
freshrss    | fde4:1d15:8139:0:5ed4:2b1d:a9c4:154 - - [06/Jul/2025:15:54:37 +0800] "GET /themes/manifest.json HTTP/1.0" 403 199 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"
```

Here is the docker-compose.yml
```
version: "2.4"

volumes:
  data:
  extensions:

services:

  freshrss:
    image: freshrss/freshrss:latest
    container_name: freshrss
    hostname: freshrss
    restart: unless-stopped
    logging:
      options:
        max-size: 10m
    volumes:
      - /usr/local/freshrss/data:/var/www/FreshRSS/data
      - /usr/local/freshrss/extensions:/var/www/FreshRSS/extensions
    environment:
      TZ: Australia/Perth
      CRON_MIN: '3,33'
    ports:
      - 8080:80
 ```

Here is the .env for the docker-compose:
```
OIDC_ENABLED=1
OIDC_PROVIDER_METADATA_URL=https://sso.tuxtower.net/application/o/freshrss/.well-known/openid-configuration
OIDC_CLIENT_ID=client_id
[...]

PR #8610, opened 2026-03-15
Do not count environment variables as HTTP headers during HTTP auth

Inverle:
The main purpose of this check was to avoid SSRF related issues, and checking just for the headers is likely enough.

### DIFF
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -497,7 +497,7 @@ final class FreshRSS_http_Util {
 
 	public static function httpAuthUser(bool $onlyTrusted = true): string {
 		$auths = array_unique(
-			array_intersect_key($_SERVER, ['REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => '', 'HTTP_REMOTE_USER' => '', 'HTTP_X_WEBAUTH_USER' => ''])
+			array_intersect_key($_SERVER, ['HTTP_REMOTE_USER' => '', 'HTTP_X_WEBAUTH_USER' => ''])
 		);
 		if (count($auths) > 1) {
 			Minz_Log::warning('Multiple HTTP authentication headers!');


Alkarex:
I am a bit confused there. I believe `REMOTE_USER` is the most standard way to retrieve HTTP Basic Auth information, not the `HTTP_*` versions, which are only used by proxies

Inverle:
`REMOTE_USER` is something that [would be set by OIDC](https://github.com/OpenIDC/mod_auth_openidc#:~:text=SSO%29%2E-,By,server) and should be a trusted value (it is also being checked and returned first), but if you are not using OIDC and are using an authentication proxy instead, then there could be confusion between `HTTP_REMOTE_USER` and `HTTP_X_WEBAUTH_USER` - given a scenario where `HTTP_X_WEBAUTH_USER` is the real value sent by the proxy, and `HTTP_REMOTE_USER` is controlled by an attacker (`HTTP_X_WEBAUTH_USER` is checked after `HTTP_REMOTE_USER`)

> checking just for the headers is likely enough

I could be wrong though

Alkarex:
OpenID Connect is just one application of using HTTP Basic Auth.
The more common way is when HTTP Basic is defined by the Web server itself (e.g. Apache), then only the more secure `REMOTE_USER` is defined, and none of the less secure `HTTP_*` ones. By less secure I mean that they can be defined by the incoming query itself, while `REMOTE_USER` has built-in guarantees.

Inverle:
This change should still be fine though, shouldn't it?

Alkarex:
> This change should still be fine though, shouldn't it?

Is it still needed after https://github.com/FreshRSS/FreshRSS/pull/8627/changes ?

Inverle:
It does solve the issue described in https://github.com/FreshRSS/FreshRSS/issues/7720#issuecomment-4063956651, but I am not sure why we should keep `'REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => ''`.

Alkarex:
> It does solve the issue described in [#7720 (comment)](https://github.com/FreshRSS/FreshRSS/issues/7720#issuecomment-4063956651), but I am not sure why we should keep `'REMOTE_USER' => '', 'REDIRECT_REMOTE_USER' => ''`.

I am also a bit confused, but in the other direction. It is still to address:
* https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-w3m8-wcf4-h8vm
* https://github.com/FreshRSS/FreshRSS/pull/7528

Inverle:
I don't have any issues with this myself, so I'll close the PR for now, but I thought the check is a bit too strict. (maybe I'm not thinking about this correctly though, and it is necessary)

PR #8624, opened 2026-03-21
do not show dot, if no unread number is shown

math-GH:
Closes #8616

Open for discussion: Is the dot really needed to separate the unread number (f.e. `(12)`) from the `FreshRSS` string? A simple `(12) FreshRSS` would do the same job, wouldn't?

### DIFF
--- a/app/Controllers/indexController.php
+++ b/app/Controllers/indexController.php
@@ -149,7 +149,7 @@ class FreshRSS_index_Controller extends FreshRSS_ActionController {
 		if (FreshRSS_Context::$get_unread > 0) {
 			$title = '(' . FreshRSS_Context::$get_unread . ') ' . $title;
 		}
-		if (strlen($title)) {
+		if (strlen($title) > 0) {
 			FreshRSS_View::prependTitle($title . ' · ');
 		}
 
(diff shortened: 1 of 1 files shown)


app/Controllers/indexController.php
```diff
@@ -149,7 +149,9 @@ public function normalAction(): void {
 		if (FreshRSS_Context::$get_unread > 0) {
 			$title = '(' . FreshRSS_Context::$get_unread . ') ' . $title;
 		}
-		FreshRSS_View::prependTitle($title . ' · ');
+		if (strlen($title)) {
```
Alkarex: I would prefer

```suggestion
		if ($title !== '') {
```

or

```suggestion
		if (strlen($title) > 0) {
```

Alkarex:
> Is the dot really needed to separate the unread number (f.e. `(12)`) from the `FreshRSS` string? A simple `(12) FreshRSS` would do the same job, wouldn't?

I do not really mind either way

Frenzie:
I like it better with, but of course it's not "needed".

Also it's good for consistency with other pages that have regular titles. See any of the settings for example; there it is needed.

PR #8625, opened 2026-03-22
Fix wrong error redirect in subscription management

Inverle:
Was supposed to show feeds with errors on redirect, only if they were being shown before, not when you modify a feed with an error.


Alkarex (approved):

Alkarex:
Good catch!

PR #8630, opened 2026-03-23
Docs/improve layout

math-GH:
beautify the docs a bit.

Preview: https://math-gh.github.io/FreshRSS/en/

### DIFF
--- a/docs/_includes/docs.css
+++ b/docs/_includes/docs.css
@@ -24,7 +24,6 @@ body {
 }
 
 aside {
-	padding: 1rem;
 	width: var(--aside-width);
 	height: 100vh;
 	border-right: 1px solid black;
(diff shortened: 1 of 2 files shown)


math-GH:
How can I try out the new layout? How can I trigger the building process and see the result before merging?

Alkarex:
The easiest might be to expose your fork of the doc, like @Inverle did in https://github.com/FreshRSS/FreshRSS/pull/8247

Inverle:
I think what I was doing is triggering the workflow manually via https://github.com/math-GH/FreshRSS/actions/workflows/jekyll-gh-pages.yml (on docs branch) and then it would be deployed at https://math-gh.github.io/FreshRSS/

(although you need to enable GitHub Actions deploying in https://github.com/math-GH/FreshRSS/settings/pages first)

Alternatively, here are brief instructions to run Jekyll locally: https://github.com/FreshRSS/FreshRSS/blob/edge/docs/README.md

math-GH:
> I think what I was doing is triggering the workflow manually via https://github.com/math-GH/FreshRSS/actions/workflows/jekyll-gh-pages.yml (on docs branch) 

great! That was the trick I missed before :)

I will link my page in the description above

PR #8631, opened 2026-03-24
Convert feed view into grid where appropriate

nanos:
This makes the feed view prettier on mobile, if thumbnails and summary are shown, as discussed in https://github.com/FreshRSS/FreshRSS/discussions/8629

- Converts Feed Item list into multiple lines
- But only if both thumbnails and summaries are shown

The code is quite different from what I had done in my own hack: There I had used `display: flex` and then counted items, but that only works for my own hack: here we don't know how many items a given user may have, so we use `display:grid` instead with name grid areas.

1. Ensure you enable both thumbnails and summaries
2. Check out the feed view on mobile
3. Play around with enabling and disabling bookmarks, share button, etc.

**Pull request checklist:**

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

**Screenshots:**

SCR-20260324-jnte
SCR-20260324-jnwf

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2549,7 +2549,6 @@ html.slider-active {
 			"website   website manage-read  manage-bookmark  manage-share  manage-labels  link"
 			"thumbnail content content      content          content       content        content";
 		align-items: start;
-		gap: 8px;
 		list-style: none;
 	}
 	/** Show feed name */
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2541,7 +2541,7 @@ html.slider-active {
 		padding-right: 1rem;
 	}
 
-	.flux .flux_header:has(.thumbnail):has(.summary) {
+	.flux .flux_header.has-thumbnail.has-summary {
 		display: grid;
 		grid-template-columns: auto 1fr auto auto auto;
 		grid-template-rows: auto auto;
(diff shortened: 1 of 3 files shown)


nanos:
I will fix those formatting issues

nanos:
lint complaints have been addressed

Alkarex:
> lint complaints have been addressed

Running `make fix-all` should fix all those small details automatically

nanos:
> Running make fix-all should fix all those small details automatically

Good to know. But not worth installing make for just these few lines.

Alkarex:
> > Running make fix-all should fix all those small details automatically
> 
> Good to know. But not worth installing make for just these few lines.

Alternatively, when limited to CSS files: `npm run fix`

p/themes/base-theme/frss.css
```diff
@@ -2541,6 +2541,55 @@ html.slider-active {
 		padding-right: 1rem;
 	}
 
+	.flux .flux_header:has(.thumbnail):has(.summary) {
+		display: grid;
+		grid-template-columns: auto 1fr auto auto auto;
+		grid-template-rows: auto auto;
+		grid-template-areas: "website   website manage-read  manage-bookmark  manage-share  manage-labels  link"
+			"thumbnail content content      content          content       content        content";
```
Frenzie: The spacing seems odd. If it's meant for alignment or something, shouldn't it align? ;-)

Frenzie (commented):
lgtm besides the minor nit, grid would've been my first instinct for this one too.

p/themes/base-theme/frss.css
nanos: Haha, yeah, it was meant for alignment. And it was aligned. Then the linter complained about the whitespace, so I removed the whitespace, which broke alignment. 

I have removed the white spaces now.

Alkarex:
I was worried of the performance reduction due to the many `:has()` CSS selectors (the few instances of that type of selector that we have are already responsible for most of the time spent for CSS evaluations), so I have made some changes.
Could you please double-check?

Frenzie:
You mean to put that in PHP instead? That probably makes sense, yes.

nanos:
yeah, looks good!

Alkarex:
Thanks again. Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md
If you happen to make a post about it on Mastodon, ping me and we will boost it

PR #8633, opened 2026-03-25
feat(favicon): Use feed-provided icon URL (<image><url>, Atom icon/logo, JSON Feed icon)

### LINKED ISSUE #5518: [Feature] Use favicon that is included in the feed

### LINKED ISSUE #6374: [Bug] How to set the favicon
### Describe the bug

How does Fresh set every feed's favicon?
I use a custom feed generated by `feedgen` and set it's icon url, but it doesn't work and show blank favicon. But other feed are just fine.
So could you tell me how to custom the favicon, thx!

### To Reproduce

just add the feed as normal.
and my costom feed xml look like
```
  <id>https://www.xiaohongshu.com</id>
  <title>小红书关注流</title>
  <updated>2024-04-24T10:16:44.946056+00:00</updated>
  <link href="https://www.xiaohongshu.com/"/>
  <generator uri="https://lkiesow.github.io/python-feedgen" version="1.0.0">python-feedgen</generator>
  <logo>https://img.ixintu.com/upload/jpg/20210522/b466594a956f6023b62f5672c59a07dd_48198_800_800.jpg</logo>
  <subtitle>小红书关注作者的笔记</subtitle>
  <entry>
```

### Expected behavior

_No response_

### FreshRSS version

1.23.1

### Environment information

- Database version: [e.g. Mysql 5.7, SQLite]
- PHP version: [e.g. PHP 8.1]
-Web server type: [e.g. Apache, nginx]
- Device: [e.g. iPhone13]
- OS: [e.g. Ubuntu 22.04, Win10, MacOS14]
- Browser: [e.g. Firefox 124]

### Additional context

_No response_

bowencool:
Closes #5518

- When a feed provides an icon URL (<image><url> in RSS 2.0/1.0, <atom:icon>/<atom:logo> in Atom, icon/favicon fields in JSON Feed), that URL is stored as a feedIconUrl attribute on the feed and used as the primary source for favicon downloads, instead of scraping the feed's website for <link rel="icon"> tags.
- If the feed-provided icon URL fails to return a valid image, the existing fallback chain (website HTML favicon search → /favicon.ico) is preserved.
Custom favicons uploaded by users always take priority and are never overridden.

1. Add an RSS feed that includes a <image><url> element (e.g. an RSSHub feed: `https://rsshub.app/youtube/channel/UC2cRwTuSWxxEtrRnT4lrlQA`). After actualization, confirm the feed's favicon matches the avatar image from the feed, not the Bilibili site favicon.
2. Add an Atom feed containing <atom:icon> or <atom:logo> Confirm the feed icon is used.
3. Add a JSON Feed (spec: icon field). Confirm icon is preferred over favicon when both are present.
4. Temporarily point a feed's <image><url> to a broken URL. Confirm FreshRSS falls back to the website favicon silently.
5. Upload a custom favicon for a feed, then actualize it. Confirm the custom favicon is not replaced.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

image

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -744,7 +744,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			}
 
 			if ($simplePie != null) {
-				$feedImageUrl = html_only_entity_decode($simplePie->get_image_url() ?? '');
+				$feedImageUrl = htmlspecialchars_decode($simplePie->get_image_url() ?? '', ENT_COMPAT);
 				$feedImageUrl = $feedImageUrl !== '' ? (FreshRSS_http_Util::checkUrl($feedImageUrl) ?: '') : '';
 				if ($feedImageUrl !== ($feed->attributeString('feedIconUrl') ?? '')) {
 					$feed->_attribute('feedIconUrl', $feedImageUrl !== '' ? $feedImageUrl : null);
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -744,7 +744,9 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			}
 
 			if ($simplePie != null) {
-				$feedImageUrl = htmlspecialchars_decode($simplePie->get_icon_url() ?? '', ENT_QUOTES);
+				// get_icon_url() is preferred (square icons, Atom <icon>), but falls back to get_image_url()
+				// for feeds that provide <image><url> without explicit dimensions (e.g. RSSHub, Atom <logo>).
+				$feedImageUrl = htmlspecialchars_decode($simplePie->get_icon_url() ?? $simplePie->get_image_url() ?? '', ENT_QUOTES);
 				$feedImageUrl = $feedImageUrl !== '' ? (FreshRSS_http_Util::checkUrl($feedImageUrl) ?: '') : '';
 				if ($feedImageUrl !== ($feed->attributeString('feedIconUrl') ?? '')) {
 					$feed->_attribute('feedIconUrl', $feedImageUrl !== '' ? $feedImageUrl : null);
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -744,7 +744,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			}
 
 			if ($simplePie != null) {
-				$feedImageUrl = htmlspecialchars_decode($simplePie->get_image_url() ?? '', ENT_QUOTES);
+				$feedImageUrl = htmlspecialchars_decode($simplePie->get_icon_url() ?? '', ENT_QUOTES);
 				$feedImageUrl = $feedImageUrl !== '' ? (FreshRSS_http_Util::checkUrl($feedImageUrl) ?: '') : '';
 				if ($feedImageUrl !== ($feed->attributeString('feedIconUrl') ?? '')) {
 					$feed->_attribute('feedIconUrl', $feedImageUrl !== '' ? $feedImageUrl : null);
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -744,7 +744,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			}
 
 			if ($simplePie != null) {
-				$feedImageUrl = htmlspecialchars_decode($simplePie->get_image_url() ?? '', ENT_COMPAT);
+				$feedImageUrl = htmlspecialchars_decode($simplePie->get_image_url() ?? '', ENT_QUOTES);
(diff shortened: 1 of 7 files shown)


app/Controllers/feedController.php
```diff
@@ -744,6 +744,14 @@ public static function actualizeFeeds(?int $feed_id = null, ?string $feed_url =
 			}
 
 			if ($simplePie != null) {
+				$feedImageUrl = html_only_entity_decode($simplePie->get_image_url() ?? '');
```
Alkarex: Note tested, but I think `htmlspecialchars_decode()` would be needed here instead of `html_only_entity_decode()`. To be tested with an URL with special characters, at least the common `&`.
But we first need to check what is returned by `get_image_url()` in terms of encoding.

Alkarex:
Thanks! Not tested yet, but looks good already

app/Controllers/feedController.php
bowencool: Fixed in the latest commit: replaced html_only_entity_decode() with htmlspecialchars_decode(..., ENT_COMPAT), which correctly decodes &amp; back to & in URLs returned by get_image_url().

Alkarex:
It is picking the logo instead of the favicon on a feed such as https://www.lesnumeriques.com/rss.xml

I am looking into it. May require some reordering of priorities in SimplePie

Alkarex:
* https://github.com/simplepie/simplepie/pull/974
* https://github.com/FreshRSS/simplepie/pull/73

Alkarex:
@bowencool I have made some changes. Please test again

bowencool:
> @bowencool I have made some changes. Please test again

Hi @Alkarex, thanks for introducing `get_icon_url()` — the square-only constraint is a nice idea for proper favicon-sized icons.

However, it breaks feeds that provide a valid avatar/icon in `<image><url>` **without** explicit `<width>`/`<height>` tags (e.g. most RSSHub feeds). When those attributes are absent, SimplePie defaults to 88×31, which is not square, so `get_icon_url()` returns `null` — and the code falls back to scraping the website for `<link rel="icon">` instead of using the feed-provided image.

Example feed that exhibits the issue:
```xml
<image>
  <url>https://i0.hdslb.com/bfs/face/142245e4a51501b459f6c0d665688a866d782443.jpg</url>
  <title>小约翰可汗 的 bilibili 空间</title>
  <link>https://space.bilibili.com/23947287</link>
  <!-- no <width> or <height> -->
</image>
```

My fix uses `get_icon_url() ?? get_image_url()`: square-declared icons (and `<atom:icon>`) are still preferred, but `get_image_url()` acts as a broader fallback covering undimensioned `<image><url>`, `<atom:logo>`, and `<itunes:image>`. This way the strict square preference is preserved while still honouring any feed-provided image URL.

Would you consider this approach acceptable, or would you prefer a different strategy (e.g. extending `get_icon_url()` to also accept images with no declared dimensions)?

Alkarex:
Sorry, `get_icon_url() ?? get_image_url()` is not good, as it will return some logos/banners that are not at all appropriate for a favicon.
It is better to return `null` to fall back to the existing favicon discovery.
88×31 is the standard ratio when no width/height is defined, so the RSSHub issue needs to be fixed upstream in RSSHub by either using Atom, an `atom:icon`, or define the width/height. Another option is to make a FreshRSS extension with a special case for RSSHub while waiting for a proper fix.

bowencool (commented):
> Sorry, `get_icon_url() ?? get_image_url()` is not good, as it will return some logos/banners that are not at all appropriate for a favicon.
> It is better to return `null` to fall back to the existing favicon discovery.
> 88×31 is the standard ratio when no width/height is defined, so the RSSHub issue needs to be fixed upstream in RSSHub by either using Atom, an `atom:icon`, or define the width/height. Another option is to make a FreshRSS extension with a special case for RSSHub while waiting for a proper fix.
>  

I agree, thanks for the clarification. I've reverted last commit.

Alkarex:
Thanks again. Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

lib/composer.json
```diff
@@ -14,7 +14,7 @@
 		"marienfressinaud/lib_opml": "0.5.1",
 		"phpgt/cssxpath": "v1.5.0",
 		"phpmailer/phpmailer": "7.0.2",
-		"simplepie/simplepie": "dev-freshrss#6405099830e5383fc2cb9aa1be7a8f42a18cb21c"
+		"simplepie/simplepie": "dev-merge-get_icon_url#23119820c414c6117e98c870a13727d2d41f3992"
```
Alkarex, line 17: * https://github.com/FreshRSS/FreshRSS/pull/8659

PR #8646, opened 2026-03-29
Fix entry->lastSeen in case of feed errors

### LINKED ISSUE #8643: [Bug] The entries' "lastSeen" value might not be correct.
### Describe the bug

This is caused by the entry's lastSeen and feed's lastUpdate going out of sync.  

The function `FreshRSS_EntryDAO::updateLastSeenUnchanged()` is only used in this section below. It assumes the `_entry.lastSeen` and `_feed.lastUpdate` are always in sync.  
https://github.com/FreshRSS/FreshRSS/blob/71c26f5512e0e3f98fd829d571070b76c35d2bd9/app/Controllers/feedController.php#L703-L716  

However, the function `FreshRSS_FeedDAO::updateLastUpdate()` besides the above flow, is also called here, which skips updating the most recent entries.  
https://github.com/FreshRSS/FreshRSS/blob/71c26f5512e0e3f98fd829d571070b76c35d2bd9/app/Controllers/feedController.php#L561-L571  

The next time the feed is updated with no new entries, the "lastSeen" value won't be updated due to how the query is written, seen below.  
https://github.com/FreshRSS/FreshRSS/blob/71c26f5512e0e3f98fd829d571070b76c35d2bd9/app/Models/EntryDAO.php#L1951-L1958  

### To Reproduce

1. Add a feed.  
2. The feed throws an error when tried to be fetched (for example a 500 or a 429).  
3. The next fetch succeeds with no new entries.  
4. See in the DB "lastUpdate" and "lastSeen" are no longer the same value.  

### Expected behavior

"_entry.lastSeen" to always have the time when it was last returned in the feed.  

### FreshRSS version

1.28.2-dev

### System information

- Installation type: Docker compose, from edge branch.

### Additional context

_No response_

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/issues/8643
Fix lastSeen and feed's lastUpdate going out of sync.
Now the feed error field is used to store the timestamp of the error instead of just a boolean.

Previous related PRs:
* https://github.com/FreshRSS/FreshRSS/pull/5404
* https://github.com/FreshRSS/FreshRSS/pull/5382
* https://github.com/FreshRSS/FreshRSS/pull/5315

### DIFF
--- a/app/Controllers/entryController.php
+++ b/app/Controllers/entryController.php
@@ -268,6 +268,7 @@ class FreshRSS_entry_Controller extends FreshRSS_ActionController {
 		}
 
 		$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
+		$databaseDAO->minorDbMaintenance();
 		$databaseDAO->optimize();
 
 		$feedDAO = FreshRSS_Factory::createFeedDao();
@@ -295,6 +296,9 @@ class FreshRSS_entry_Controller extends FreshRSS_ActionController {
 			@set_time_limit(300);
 		}
 
+		$databaseDAO = FreshRSS_Factory::createDatabaseDAO();
+		$databaseDAO->minorDbMaintenance();
+
 		$feedDAO = FreshRSS_Factory::createFeedDao();
 		$feeds = $feedDAO->listFeeds();
 		$nb_total = 0;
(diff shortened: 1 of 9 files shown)


Alkarex:
@pe1uca if you can test, please report whether that fixes your observed problem

rezad1393:
this feature is great.
btw can this be shown as absolute date instead of relative date?

also after update to 1.29 which has this feature, the feeds that have errors from before all say that last sussesfull update was the last sync (for me two hours ago) even though they had error for many days.

for example 
" Last successful update 2 ساعت پیش. "
 it says it was 2 hours ago.

and if I force update this feed this becomes:
"Last successful update 2 ساعت پیش.
Last erroneous update همین الان"
which added that error was "right now" (cause I just force update an offline feed.

I hope this is just because of transient state of metadata added for this feature and upper message gets cleared and updated with the next future successful feed update.

Alkarex:
> btw can this be shown as absolute date instead of relative date?

If you hover over the date, a tooltip with the absolute date is supposed to be shown

PR #8655, opened 2026-03-31
Hook to modify the list of feeds to actualize

### LINKED ISSUE #8650: [Feature] Have a hook to modify the order in which feeds are updated

When the feeds are updated, the ones to a given site can be clumped together. I'm seeing this mostly due to rate limiting.  
On a given run the feeds of YT, for example, try to be updated, some succeed, then YT rate limits us and the rest are skipped/error out. Other feeds to other sites get updated properly.  
On the next run the YT feeds are first to be updated since their "lastUpdate" value is older (as queried by `FreshRSS_FeedDAO::listFeedsOrderUpdate()`), which makes reaching the rate limit easier.  

### Use case

I have several feeds on each of a few sites (e.g. YT and reddit), which have low rate limits on a short fixed window (I think it's between 5 and 10 minutes).  
Due to the amount of feeds (plus some slow sites) I have, the actualize script can take more than 10 minutes.  
This makes the refreshed rate limit not be used at the end of the run and have to wait until the next run of the script.    

I'd like to be able to control those clumps of feeds, which can vary per site due to different configurations on the rate limit window.  
So, an extension that processed and reorders the list of feeds to be updated is what I'm thinking.  
I have in mind two reorder strategies to test:
- Alternate to other sites after X number of feeds to a single site.  
- Move some fraction of the feeds to the same site to the end of the queue.  

The extensions [AutoTTL](https://github.com/mgnsk/FreshRSS-AutoTTL) and [RateLimiter](https://github.com/pe1uca/xExtension-RateLimiter) help mitigate this issue.  
Still, AutoTTL only makes it happen less often, since there a MaxTTL in which all feeds must be updated at once. And RateLimiter mostly helps on not wasting time on http request we know will fail, and not putting the feeds on an error state.  

### Additional context

The hook I have in mind would update the variable `$feeds` to modify the behavior of this foreach    
https://github.com/FreshRSS/FreshRSS/blob/31711076286a54e202e069c71762bdd7822e184a/app/Controllers/feedController.php#L470-L474

pe1uca:
Closes #8650

- Implement new hook to allow extensions to modify the list of feeds to actualize.  

1. Add several feeds from a single site (e.g. reddit and YT).  
2. Add feeds to other sites.  
3. Add and enable the extension https://github.com/pe1uca/xExtension-Declumping  
4. Call `php app/actualize_script.php` to update feeds with a different order.  

A log like this one is needed to properly see the behavior.  
```php
foreach ($feeds as $key => $value) {
    syslog(LOG_INFO, "$key: {$value->name()} ({$value->url()})");
}
```

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated


Alkarex:
I have made a little change. Please double-check

pe1uca:
I was thinking about that sanitization :D  
Although, was probably going to add it inside the foreach, a duplicate of the guard for the `Minz_HookType::FeedBeforeActualize` hook.  

Give it a test and it's working great!

PR #8670, opened 2026-04-02
Show time since when a feed has problems

### LINKED ISSUE #8508: [Feature] Show time since when a feed has problems

cweiske:
Closes #8508

- Use an integer for `Feed::error` everywhere (follow up to #8646)
- Extract `Entry::machineReadableDate()` into function for use in HTML templates
- Add `timeago()` function that converts a unix timestamp into a "4 weeks ago" string
- Show the last successful feed update, and the last erroneous update

1. Update a feed
2. Modify the feed URL in the database and set it to a non-existing URL
3. Update the feed again
4. Open the "Manage feed" and see the expanded error message:

>  Blast! This feed has encountered a problem. If this situation persists, please verify that it is still reachable.
> Last successful update 3 hours ago, last erroneous update 1 hour ago. 

You can hover the relative dates to see the timestamp.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard) - this was not possible. I wanted to test timeago(), but the translator `_t()` is not available in the tests.
- [ ] documentation updated

# Screenshot
2026-04-02 freshrss feed update error


lib/lib_date.php
```diff
@@ -141,3 +141,39 @@ function parseDateInterval(string $dateInterval): array {
 	}
 	return [$min, $max];
 }
+
+/**
+ * Human readable string how long this timestamp is ago ("5 years ago").
+ */
+function timeago(int $t, string $base = 'now'): string {
```
Alkarex: I do not know how well that works with non-European languages.
Maybe @Frenzie has feedback

Frenzie: It might not even work very well for many European languages, mainly the Slavic ones. :-)

But PHP comes with this kind of thing built-in since 8.0: https://www.php.net/manual/en/class.intldateformatter.php

cweiske: `IntlDateFormatter` indeed has relative dates since PHP 8.0, but only for yesterday/today and tomorrow:
```
<?php
$fmt = datefmt_create(
    'en_US',
    IntlDateFormatter::RELATIVE_SHORT,
    IntlDateFormatter::SHORT
);
echo datefmt_format($fmt, time()) . "\n";
echo datefmt_format($fmt, time() + 86400) . "\n";
echo datefmt_format($fmt, time() - 86400) . "\n";
echo datefmt_format($fmt, time() - 2*86400) . "\n";
echo datefmt_format($fmt, time() - 3*86400) . "\n";
``` 
Output:
```
today, 6:47 AM
tomorrow, 6:47 AM
yesterday, 6:47 AM
4/1/26, 6:47 AM
3/31/26, 6:47 AM
```

This is not enough.

Frenzie: I think not hiding such pertinent information is preferable even without localization concerns, or in other words I'd rather see `today, 6:47` than `3 hours ago`, but fwiw it seems there's https://carbon.nesbot.com/ which presumably does localize it properly.

cweiske: My idea was that it's easier for people to understand "last feed update was 6 months ago" than seeing "last feed update was 2025-11-02 11:23:42". The feature is about seeing/understanding which feeds are dead and can be deleted. Relative times make this much easier in my eyes.

I also saw that Carbon has this feature, as well as e.g. https://github.com/jimmiw/php-time-ago - but the composer.json require list is empty, and a select list of dependencies are hard-copied into the lib folder.
This made me assume that FreshRSS does not want to have external dependencies, and I figured that a feature used in a single place does not warrant a new dependency.

But in the end it's your decision.

Alkarex: Yes, I would rather not include another dependency, especially not for something that minor.
@Frenzie Do you have an example or two from our currently supported languages, where the proposed function does not work? I am hoping to find an approach allowing translators to have some kind of graceful fallback.

Alkarex: (But if that feature is something that suddenly is desired for more important use-cases, including a tiny library would be an option)

Frenzie: I'm pretty sure it doesn't even work in German! (Which I suspect to be @cweiske's native language)

3 + months + ago = *VOR* + 3 + Monate*N* (while this function would give `vor 3 Monate` without taking the dative into account)

It doesn't work in languages like Russian.

1 is one form, but also shared by anything (?) ending in 1 like 11, 151, etc.
2-4 months is one plural
5-9 is another plural

GetText has a straightforward solution for that with plural formulas.

So what you need is an original:

Singular: 1 month ago
Plural: %d months ago

Which for e.g. Russian then has

[] %d month1 ago
[] %d month2 ago
[] %d month3 ago
[] %d month4 ago

With the correct entry selected based on the plural formula.

Alkarex: I am looking into it. It seems like some minor changes such as offering a couple of plurals (two, or more) would cover several more cases.
We should also have the cases of the future, which may happen for instance during time changes, and should at least not fail

Frenzie: It's something I have considered adding in the past, but at the time I decided against it since it's a small number of minor details here and there.

But in short, it should use the full strings rather than machine-made composites. The plural situation is less pressing in the sense that it can be added later, but it may be preferable to bite through the sour apple (Dutch idiom, bite the bullet in English ;-).

cweiske: > 3 + months + ago = VOR + 3 + MonateN

You're right. It's a bug that I missed, but it can be fixed for German. And yes, I'm German.

Frenzie: Regarding GetText, its plural formulas look like this:

Dutch / English:
```
nplurals=2; plural=n != 1
```

French:
```
nplurals=2; plural=n > 1
```

Russian:
```
nplurals=4; plural=(n%10==1 && n%100!=11 ? 0 : n%10>=2 && "
"n%10<=4 && (n%100<12 || n%100>14) ? 1 : n%10==0 || (n%10>=5 && n%10<=9) || "
"(n%100>=11 && n%100<=14)? 2 : 3)
```

Other approaches are possible, but it's straightforward to implement and it works very well.

Frenzie: > You're right. It's a bug that I missed, but it can be fixed for German. And yes, I'm German.

True, for German it's not really an issue but in this case more of a typo, but I still think the mistake is indicative of the conceptual problem with translating a disconnected `months` rather than a full `%d months ago`. :-)

Alkarex: Alright, I spent much more time than anticipated, got inspired by ICU, learned several things on several languages along the way...
https://cldr.unicode.org/index/cldr-spec/plural-rules

Alkarex: Thoughts @Frenzie ?

Alkarex:
Little bug spotted, for feeds in error without known error date:
image

Help welcome, or I will come back to it later.
Edit: done

lib/lib_date.php
Frenzie: I'm not fond of the one/few/many adaptation. It's a hardcoded value, in this case supporting a maximum of merely five plural forms, with what will inevitably sometimes be misleading built-in assumptions associated with words like `two` or `few` that wouldn't exist with a plain plural form number 3 (and nobody has to care whether it's few or many or hundreds).

Regarding the five plural forms, Arabic has 6 different plural forms, and I'm not sure if that's the limit:

```
nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 ? 4 : 5);
```

It's more flexible and straightforward to just say plural 2 = 2 based on such a formula. As a human you can always refer to that as "many" if you like, but I prefer the code and what's presented to the translator to be neutral. And I'd say ideally the formula should just be part of the translation data rather than being hardcoded in an imho much harder to find `pluralType` function.

I can take a stab at it in a couple of hours if you like.

[1] You could say `plural_0`, `plural_1`, etc. if you prefer that over plain integers.

Alkarex: The current proposition is based on https://cldr.unicode.org/index/cldr-spec/plural-rules

Alkarex: If I am not mistaken, Arabic should be covered by zero, one, two, few, many, other

Alkarex: But other approaches welcome. I just tried to base it on some kind of existing convention

Alkarex: > ideally the formula should just be part of the translation data

Yes, that could be nice, but I am not sure how to fit it in our existing approach without breaking too many things

Frenzie: Quoting from there:

> _These categories are only mnemonics -- the names don’t necessarily imply the exact contents of the category._ For example, for both English and French the number 1 has the category one (singular). In English, every other number has a plural form, and is given the category other. French is similar, except that the number 0 also has the category one and not other or zero, because the form of units qualified by 0 is also singular.
> 
> *This is worth emphasizing:* A common mistake is to think that “one” is only for only the number 1. Instead, “one” is a category for any number that behaves like 1. So in some languages, for example, one → numbers that end in “1” (like 1, 21, 151) but that don’t end in 11 (like “11, 111, 10311).

Their names are unnecessarily confusing if not straight up misleading, and this kind of "common mistake" would never have happened had they used a simple array-like structure like GetText does, which doesn't have any implications beyond there being _n_ distinct plural forms.

Frenzie: > > ideally the formula should just be part of the translation data
> 
> Yes, that could be nice, but I am not sure how to fit it in our existing approach without breaking too many things

I'll perform some experiments later, but as far as that goes I just figured something like `plural(_func).php` in the language folder. It'd be very similar to what you already wrote, just distributed instead of centralized.

In the simplest sense avoiding word-names like `one` and `two` is extremely trivial. You could just replace it with names like `plural_0`. But I think I might like to try using a real `[0] = 'singular', [1] = 'plural'` type structure.

app/i18n/en/plurals.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+return array(
+	'plural-forms' => 'nplurals=2; plural=(n != 1);',
```
Alkarex: Could we envision a PHP expression (lambda) instead of a string?
It leads to a very very long code and I am worried both of the maintenance and the performance

Frenzie: How do you feel about both?

Copy-paste from pre-existing gettext plurals ⇒ generated PHP code (as part of make fix-all for example)

Alkarex: I am not sure I understand the details of your suggestion, but sounds potentially interesting, especially if it can reduce the amount of code (would be nice to avoid all the parser-related code).

Frenzie: In that case I mean with parser-related code as a cli helper file, so not at runtime.

I'm not attached to the idea of keeping gettext formulas or anything, but I like the ease of finding existing ones as well as their legibility.

Alkarex: Ok, I will try to make a suggestion tomorrow without parsing at all, and if that does not turn out good enough, then your CLI suggestion sounds good

Frenzie: I think that's the same thing really; deleting the script after generation would come down to the same thing except it'd make adding future languages more labor-intense.

Frenzie: The overcomplicated parser can likely be simplified to replacing `n` with `$n` instead… tomorrow.

Frenzie: @Alkarex Alright, I've adapted it to what I consider the best behavior.

* No attempt to be smart. Programmers love concatenating (or in this case nested string formatting) as a type of DRY, but it makes things harder for the translator, and you never know in advance if in some language it might need to be `8 ago days`. Regardless whether it's actually an issue here, it's a bad example to have in the code, and examples will be followed or copy-pasted as starting points.
* The formulas are provided as:
  ```php
  return array(
  	'nplurals' => 3,
  	'plural' => static fn (int $n): int => ($n % 10 == 1 && $n % 100 != 11 ? 0 : ($n != 0 ? 1 : 2)),
  );
  ```
  You can write them differently if you prefer, but as far as the input and output goes I think that's the way to go as long as you don't go full [Fluent](https://projectfluent.org/).
* The GetText formula to PHP formula conversion script can easily be left out, but I like the idea of being able to go to [some overview of GetText formulas](https://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html) in order to copy-paste it for some new language. Of course manual conversion is quite trivial.
  [Edit: the complexity in the GetText conversion script stems from the fact that PHP nested ternary operators are deprecated, and that phpcs demands certain whitespace.]

Frenzie: Sample behavior of `make i18-format` when changing the number of plural forms (which won't or shouldn't happen, but relevant for new languages):

<details>

```diff
diff --git a/app/i18n/cs/gen.php b/app/i18n/cs/gen.php
index 62f6dc913..771157a04 100644
--- a/app/i18n/cs/gen.php
+++ b/app/i18n/cs/gen.php
@@ -145,32 +145,68 @@ return array(
                        0 => 'před %d den',
                        1 => 'před %d dny',
                        2 => 'před %d dní',
+                       3 => '%d days ago',     // TODO
+                       4 => '%d days ago',     // TODO
+                       5 => '%d days ago',     // TODO
+                       6 => '%d days ago',     // TODO
+                       7 => '%d days ago',     // TODO
+                       8 => '%d days ago',     // TODO
                ),
                'hour' => array(
                        0 => 'před %d hodina',
                        1 => 'před %d hodiny',
                        2 => 'před %d hodin',
+                       3 => '%d hours ago',    // TODO
+                       4 => '%d hours ago',    // TODO
+                       5 => '%d hours ago',    // TODO
+                       6 => '%d hours ago',    // TODO
+                       7 => '%d hours ago',    // TODO
+                       8 => '%d hours ago',    // TODO
                ),
                'justnow' => 'právě teď',
                'minute' => array(
                        0 => 'před %d minuta',
                        1 => 'před %d minuty',
                        2 => 'před %d minut',
+                       3 => '%d minutes ago',  // TODO
+                       4 => '%d minutes ago',  // TODO
+                       5 => '%d minutes ago',  // TODO
+                       6 => '%d minutes ago',  // TODO
+                       7 => '%d minutes ago',  // TODO
+                       8 => '%d minutes ago',  // TODO
                ),
                'month' => array(
                        0 => 'před %d měsíc',
                        1 => 'před %d měsíce',
                        2 => 'před %d měsíců',
+                       3 => '%d months ago',   // TODO
+                       4 => '%d months ago',   // TODO
+                       5 => '%d months ago',   // TODO
+                       6 => '%d months ago',   // TODO
+                       7 => '%d months ago',   // TODO
+                       8 => '%d months ago',   // TODO
                ),
                'second' => array(
                        0 => 'před %d sekunda',
                        1 => 'před %d sekundy',
                        2 => 'před %d sekund',
+                       3 => '%d seconds ago',  // TODO
+                       4 => '%d seconds ago',  // TODO
+                       5 => '%d seconds ago',  // TODO
+                       6 => '%d seconds ago',  // TODO
+                       7 => '%d seconds ago',  // TODO
+                       8 => '%d seconds ago',  // TODO
                ),
                'year' => array(
                        0 => 'před %d rok',
                        1 => 'před %d roky',
                        2 => 'před %d let',
+                       3 => '%d years ago',    // TODO
+                       4 => '%d years ago',    // TODO
+                       5 => '%d years ago',    // TODO
+                       6 => '%d years ago',    // TODO
+                       7 => '%d years ago',    // TODO
+                       8 => '%d years ago',    // TODO
                ),
        ),
        'js' => array(
diff --git a/app/i18n/cs/plurals.php b/app/i18n/cs/plurals.php
index dfbad3b9c..960aaef19 100644
--- a/app/i18n/cs/plurals.php
+++ b/app/i18n/cs/plurals.php
@@ -1,10 +1,10 @@
 <?php
 
-// Plural-Forms: nplurals=3; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2;
+// Plural-Forms: nplurals=9; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2;
 // This file is generated by cli/compile.plurals.php.
 // Edit the formula comment and run `make fix-all`.
 
 return array(
-       'nplurals' => 3,
+       'nplurals' => 9,
        'plural' => static fn (int $n): int => ($n == 1 ? 0 : ($n >= 2 && $n <= 4 ? 1 : 2)),
 );
```

</details>

Alkarex:
Looks much better, @Frenzie 👍🏻 
I have made a few changes. Please double-check. Otherwise looks good to me

Frenzie:
The changes look good provided that commit https://github.com/FreshRSS/FreshRSS/pull/8670/changes/3f3e16f47ac7d75e23bf7f1fb8f8b9b48481b43d doesn't introduce huge indentation in the final output, but I saw a few commits later that you regenerated the formulas so I guess it must be, which surprises me a little.

Alkarex:
> The changes look good provided that commit [3f3e16f](https://github.com/FreshRSS/FreshRSS/commit/3f3e16f47ac7d75e23bf7f1fb8f8b9b48481b43d) doesn't introduce huge indentation in the final output, but I saw a few commits later that you regenerated the formulas so I guess it must be, which surprises me a little.

The indentation is allowed since PHP 7.3+
https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc

Frenzie:
I see, it depends on the closing indentation before the closing identifier. Interesting concept. I didn't know they'd added that.

Frenzie (approved):
All fine by me.

cweiske:
Wow! Thanks for caring.

Alkarex:
It would be nice to take advantage of the new `timeago()` on the page https://freshrss.example.net/i/?c=stats&a=idle instead of the manual translations, if someone is motivated

Alkarex:
Follow-up in https://github.com/FreshRSS/FreshRSS/pull/8799

PR #8672, opened 2026-04-03
Implement setting for iframe referrer allowlist

### LINKED ISSUE #8289: [Bug] YouTube iframes don't load correctly due to stripped `referrerpolicy` attribute
Example feed: https://www.troyhunt.com/rss/
It has the necessary `referrerpolicy` attribute required for YT iframes to work, but it's stripped out by FreshRSS for security reasons.

This could be fixed either by adding the attribute in the YouTube Video Feed extension (for all feeds, not just YT feeds) or by having a setting to allow a feed to use this attribute for specific sources.

Related: https://github.com/FreshRSS/Extensions/pull/382

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/8289

Put `www.youtube.com` and `www.youtube-nocookie.com` into `Configuration > Privacy > Sites allowed to see your origin` to make youtube iframes work.

This is a client-side solution inside the Web UI, I have not tested how mobile/desktop apps that use the API behave, and what they would even send inside the `Referer` header..


app/i18n/en/conf.php
```diff
@@ -122,6 +122,7 @@
 	'privacy' => array(
 		'_' => 'Privacy',
 		'retrieve_extension_list' => 'Retrieve extension list',
+		'send_referrer_allowlist' => 'Sites allowed to see your origin (%s)',
```
Frenzie: Maybe something like this? I'm not sure if origin is very clear unless you already know what it is.
```suggestion
		'send_referrer_allowlist' => 'Sites allowed to see your address in iframes (%s)',
```

Frenzie (approved):
Fine by me, although I kind of expected it more on the PHP side. :-)

app/i18n/en/conf.php
Inverle: Would `Sites allowed to see your server address (%s)` be better?
I'm not sure if this will be used only for iframes in the future.

Frenzie: Sure.

Inverle:
> This is a client-side solution inside the Web UI, I have not tested how mobile/desktop apps that use the API behave, and what they would even send inside the `Referer` header..

Changes are probably still needed for API then, but I will do that in another PR:
https://github.com/jocmp/capyreader/pull/1573/changes

Also https://github.com/jocmp/capyreader/issues/1384#issuecomment-3770436700

PR #8675, opened 2026-04-05
Feed list hook traversable

pe1uca:
Improvements to changes in #8655  

I found some cases where returning an array doesn't provide the desired result using the extension to solve the use case mentioned in #8650 (I'll update this issue with more information).  

- Accept a Traversable object as the response from `Minz_HookType::FeedsListBeforeActualize` to provide more control on the order the feeds are updated.  
- Update the `$feed` object with some changes that were being made directly to the DB via `$feedDAO`.  

1. Have several feeds from a couple of sites (e.g. YT, reddit)  
2. Have an extension that registers to the hook.  
  https://github.com/pe1uca/xExtension-Declumping works as an example on the branch `main` for an array, and `traversable` for the Traversable object.  
3. Call `php app/actualize_script.php` to update feeds with a different order.  

A log like this one is needed to properly see the behavior.  
```php
foreach ($feeds as $key => $value) {
    syslog(LOG_INFO, "$key: {$value->name()} ({$value->url()})");
}
```

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -474,9 +474,13 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			$feeds = [];
 		}
 
+		$firstFeed = null;
 		foreach ($feeds as $feed) {
 			if (!($feed instanceof FreshRSS_Feed)) {
 				continue;
+			}
+			if (null === $firstFeed) {
+				$firstFeed = $feed;
 			}
 			$feed = Minz_ExtensionManager::callHook(Minz_HookType::FeedBeforeActualize, $feed);
 			if (!($feed instanceof FreshRSS_Feed)) {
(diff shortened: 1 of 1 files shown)


app/Controllers/feedController.php
```diff
@@ -472,13 +472,18 @@ public static function actualizeFeeds(?int $feed_id = null, ?string $feed_url =
 		$categoriesEntriesGuid = [];
 
 		$feeds = Minz_ExtensionManager::callHook(Minz_HookType::FeedsListBeforeActualize, $feeds);
-		if (is_array($feeds)) {
-			$feeds = array_filter($feeds, static fn($feed): bool => $feed instanceof FreshRSS_Feed);
-		} else {
+		if (!is_array($feeds) && !($feeds instanceof Traversable)) {
```
```suggestion
		if (!is_iterable($feeds)) {
```

Issue #8514, opened 2026-02-18
[Bug] Slow query from EntryDAO::sqlListWhere() with large amount of entries


pe1uca:
### Describe the bug

When scrolling through the UI the last entries take a long time to load.  

### To Reproduce

1. Set up an instance.  
2. Have millions of entries.  
3. Select a feed with few unread entries.  
4. Scroll through the "Normal view" and see a long wait time when fetching the last unread entries of a feed (either on first load, or when scrolling and the spinner at the bottom appears again).  

### Expected behavior

Not have to wait 1 or 2 minutes to read the last entries of a feed.  

### FreshRSS version

1.27.2-dev

### System information

- Database version: PostgreSQL 17
- PHP version: 8.4.16
- Installation type: Docker
- Device: PC
- OS: Fedora 43
- Browser: Firefox 147.0.3

### Additional context

The query generated by this function takes around 1.5 minutes when the entry table has millions of rows.  
https://github.com/FreshRSS/FreshRSS/blob/17be48fe1c4abcb4f1e1004ba38add14d820163d/app/Models/EntryDAO.php#L1421-L1424  

I think in my tests it was when used from `EntryDAO::listWhereRaw()`.  

Here's the query
```sql
SELECT e.id FROM "pe1uca_entry" e INNER JOIN "pe1uca_feed" f ON f.id = e.id_feed WHERE e.id_feed=$1  AND (e.is_read=0) AND e.id <= $2 ORDER BY e.id DESC LIMIT 21
```

And here's where it's being used
```sql
SELECT e0.id, e0.guid, e0.title, e0.author, e0.content, e0.link,
e0.date, e0."lastSeen", e0."lastUserModified", encode(e0.hash, 'hex') AS hash, e0.is_read, e0.is_favorite, e0.id_feed, e0.tags, e0.attributes
FROM "pe1uca_entry" e0 INNER JOIN (SELECT e.id FROM "pe1uca_entry" e INNER JOIN "pe1uca_feed" f ON f.id = e.id_feed WHERE e.id_feed=$1  AND (e.is_read=0) AND e.id <= $2 ORDER BY e.id DESC LIMIT 21) e2 ON e2.id=e0.id ORDER BY e0.id DESC
```
(I tried this last query by replacing the second query with the direct list of ids, and it took 5ms)

Alkarex:
Do you happen to have found a suggestion of improvement?

Alkarex:
And how fast is that query?

```sql
[...]

Issue #8455, opened 2026-01-22
[Bug] Favorites extremely slow to load


alexklapheke:
### Describe the bug

After updating to 1.28, loading favorites is extremely slow. As you can see below, the FreshRSS interface (the requests up through `favicon.ico`) loaded very quickly, but took over a minute to load the list of posts.

<img width="1677" height="496" alt="Image" src="https://github.com/user-attachments/assets/7dc3dd69-4950-4dc6-8be7-1d7e1e3e3b9e" />

### To Reproduce

Click `Favourites` in the sidebar or add `get=s` to the URL

### FreshRSS version

1.28.0

### System information

- Database version: MariaDB 10.6
- PHP version: 8.1.33
- Installation type: Git
- Web server type: Apache 2.4
- OS: FreeBSD 14.3
- Browser: Firefox 147.0.1

### Additional context

This seems to be related to #8211: running `SELECT link FROM _entry WHERE is_favorite = 1` is instantaneous, but `SELECT link FROM freshrssalex_entry USE INDEX (entry_feed_read_index) WHERE is_favorite = 1` takes a full minute, about the same latency as from the web interface.

```
rss@rss.db [rss]> EXPLAIN SELECT link FROM freshrssalex_entry WHERE is_favorite = 1;
+------+-------------+--------------------+------+---------------+-------------+---------+-------+------+-------+
| id   | select_type | table              | type | possible_keys | key         | key_len | ref   | rows | Extra |
+------+-------------+--------------------+------+---------------+-------------+---------+-------+------+-------+
|    1 | SIMPLE      | freshrssalex_entry | ref  | is_favorite   | is_favorite | 1       | const | 175  |       |
+------+-------------+--------------------+------+---------------+-------------+---------+-------+------+-------+
1 row in set (0.005 sec)

rss@rss.db [rss]> EXPLAIN SELECT link FROM freshrssalex_entry USE INDEX (entry_feed_read_index) WHERE is_favorite = 1;
+------+-------------+--------------------+------+---------------+------+---------+------+--------+-------------+
| id   | select_type | table              | type | possible_keys | key  | key_len | ref  | rows   | Extra       |
+------+-------------+--------------------+------+---------------+------+---------+------+--------+-------------+
|    1 | SIMPLE      | freshrssalex_entry | ALL  | NULL          | NULL | NULL    | NULL | 360025 | Using where |
+------+-------------+--------------------+------+---------------+------+---------+------+--------+-------------+
[...]

PR #8678, opened 2026-04-06
feat: Allow feeds to belong to multiple categories (fixes #1989)

### LINKED ISSUE #1989: [Project creation] Categories
Create a project for :

Categories Management
- Feeds in multiple catégories
- Select and move multiple feeds at one time to another category
- Share a category amongst all the users

TAG management 
- add tag to feeds
- add tag to posts

nkcfan:
Closes #1989

- Add `feed_category` join table to all DB schemas (SQLite, MySQL, PostgreSQL) to store many-to-many feed-category relationships
- Add `categoryIds` field to `Feed` model with `categoryIds()` getter and `_categoryIds()` setter
- Add `setFeedCategories()` and `getFeedCategoryIds()` to `FeedDAO`; update `listByCategory()` to use the join table
- Update `CategoryDAO::listCategories()` to join via `feed_category` (falls back to legacy `feed.category` for feeds not yet in the join table)
- Fix OPML import (`ImportService`) to assign feeds to multiple categories when the same feed URL appears under multiple category outlines
- Fix Fever API (`getFeedsGroup`) to return all category memberships for feeds in multiple categories

- Import an OPML file where the same feed URL appears under multiple `<outline>` categories — verify the feed appears under all categories
- Export OPML and verify the feed is listed under each of its categories
- In Fever API client (e.g. Reeder), verify feeds appear in all their assigned categories

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested (OPML import/export round-trip, Fever API)
- [ ] documentation updated

### DIFF
--- a/app/Models/CategoryDAO.php
+++ b/app/Models/CategoryDAO.php
@@ -332,7 +332,13 @@ class FreshRSS_CategoryDAO extends Minz_ModelPdo {
 				SELECT c.id AS c_id, c.name AS c_name, c.kind AS c_kind, c.`lastUpdate` AS c_last_update, c.error AS c_error, c.attributes AS c_attributes,
 					{$feedFields}
 				FROM `_category` c
-				LEFT OUTER JOIN `_feed` f ON f.category=c.id
+				LEFT OUTER JOIN (
+					SELECT f.*, fc.category_id AS _joined_category_id FROM `_feed` f
+					JOIN `_feed_category` fc ON fc.feed_id = f.id
+					UNION
+					SELECT f.*, f.category AS _joined_category_id FROM `_feed` f
+					WHERE NOT EXISTS (SELECT 1 FROM `_feed_category` fc2 WHERE fc2.feed_id = f.id)
+				) f ON f._joined_category_id = c.id
 				GROUP BY f.id, c_id
 				ORDER BY c.name, f.name
 				SQL;
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -53,6 +53,7 @@ class FreshRSS_Feed extends Minz_Model {
 	private int $kind = 0;
 	private int $categoryId = 0;
 	private ?FreshRSS_Category $category = null;
+	private array $categoryIds = [];
 	private int $nbEntries = -1;
 	private int $nbNotRead = -1;
 	private string $name = '';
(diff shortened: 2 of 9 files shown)


app/Models/CategoryDAO.php
```diff
@@ -332,7 +332,13 @@ public function listCategories(bool $prePopulateFeeds = true, bool $details = fa
 				SELECT c.id AS c_id, c.name AS c_name, c.kind AS c_kind, c.`lastUpdate` AS c_last_update, c.error AS c_error, c.attributes AS c_attributes,
 					{$feedFields}
 				FROM `_category` c
-				LEFT OUTER JOIN `_feed` f ON f.category=c.id
+				LEFT OUTER JOIN (
+					SELECT f.*, fc.category_id AS _joined_category_id FROM `_feed` f
+					JOIN `_feed_category` fc ON fc.feed_id = f.id
+					UNION
+					SELECT f.*, f.category AS _joined_category_id FROM `_feed` f
+					WHERE NOT EXISTS (SELECT 1 FROM `_feed_category` fc2 WHERE fc2.feed_id = f.id)
```
Alkarex, line 340: The part specifically is a performance problem, which will need careful analysis and mitigation, in particular on small hardware setups with many feeds and many articles

Alkarex:
Hello and thanks for the PR.
This is something we have discussed before (see e.g. https://github.com/FreshRSS/FreshRSS/issues/1989#issuecomment-416687034), and which will require more analysis before potential inclusion, in particular in terms of performance, but also in terms of API compatibility with other clients, and UX complexification.

The performance part is my main concern, especially when balanced with the fact that this is a relatively niche feature, so regression tests need to be done with performance comparison.

At the moment, we do not have a ready-to-use performance benchmark, but manually testing the SQL queries affected by this change would be enough, in particular the main SQL query returning the articles on the main page (also when combined with filters and search). (I have such a benchmark on my TODO list)

An example of reproducible hardware platform would be a specific Rasperry Pi model, which is nice for benchmarking, but another modest platform would be a start. There needs to be several hundred thousands articles (we have cases with millions), at least 10k+ unread, at least a few hundred feeds (we have cases with thousands) and a a few dozen categories. And this needs to be done for PostgreSQL, SQLite, MariaDB and MySQL.

We have users for whom performance is essential and who could be involved, if they are willing, in testing on their own setups, e.g.:
* https://github.com/FreshRSS/FreshRSS/issues/8514
* https://github.com/FreshRSS/FreshRSS/issues/8455
* https://github.com/FreshRSS/FreshRSS/pull/8211

PR #8217, opened 2025-11-15
Show search query in the page title

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/8133

image

PR #8684, opened 2026-04-08
Decode HTML in article titles

FireFingers21:
I subscribe to the [Malwarebytes RSS feed](https://www.malwarebytes.com/blog/feed/index.xml) which uses encoded punctuation in its article titles. This makes it really hard to read when the title shows up as `Why we＆#8217;re still not doing April Fools＆#8217; Day`.

- Modify the `title()` function to return `$title` with decoded HTML

1. Subscribe to https://www.malwarebytes.com/blog
2. Titles like the one mentioned above now display as `Why we're still not doing April Fools' Day`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -167,7 +167,7 @@ class FreshRSS_Entry extends Minz_Model {
 				}
 			}
 		}
-		return $title;
+		return html_entity_decode($title, ENT_QUOTES | ENT_HTML5, 'UTF-8');
 	}
 
 	#[Deprecated('Use authors() instead')]


Frenzie:
That's not correct. Perhaps it could be an option, but this would break every feed that does things correctly.

FireFingers21:
I had a feeling this feed might not be doing something right. Since this seems to be an edge case, would it be better to close this PR and re-open it as an issue to figure out a better way to handle it?

Frenzie:
Pinging @Alkarex, but probably.

Alkarex:
Indeed, as @Frenzie wrote, it would not be correct to perform on every feed.

See also this related PR draft (help welcome):
* https://github.com/FreshRSS/FreshRSS/pull/4728

Otherwise, I would like to offer in the core the option to apply some regex search&replace on the title and body of articles, which could also help with this issue. Help welcome for that as well.

While waiting for one of those better options, custom cases can be solved with a little FreshRSS extension

FireFingers21:
I'll take a look at that then. Thanks for the feedback!

PR #8687, opened 2026-04-08
Custom feeds order in a category

### LINKED ISSUE #7659: [Feature] Manual ordering of feeds within a category

Feeds in FreshRSS seems ordered alphabetically in each category. This isn’t always what the user might want.

### Use case

_No response_

In “Subscription management” add the ability to drag and drop feeds not only between categories, but also within a category. This would allow users to have their feed in a custom order, reflected in all other parts of the app.

_No response_

### Additional context

_No response_

pe1uca:
Closes #7659

- Allow the user to set a custom order in the list of feeds of a category.  

[sort-example.webm](https://github.com/user-attachments/assets/35b85718-2714-4c27-a550-5d9152389600)

1. Create a category.  
2. Add several feeds.  
3. Click on the up and down arrows button on the header of the category.  
4. Drag and drop feeds within the category.  
5. Click the same button to save the order.  
6. Refresh the page to see the new order loaded.  

I'm creating the MR to show the feature and start getting feedback, as well as requesting some help on a couple of things I tried but wasn't able to solve.  

Help needed:

- [x] How to change the button from the arrows to a save icon.  
  I tried using the CSS class `hidden`, but the class `btn` overrides the ``display: none;``
- [x] How to add a second button.  
  The new button is set with absolute position, so doing the same will be too rigid. Maybe being inside a div?  

Things to do:

- [x] Change gear icon in each feed to an icon to symbolize the reordering state.  
- [x] ~~Handle when feeds are moved between categories.~~ Not much of an issue since the code ignores extra ids, and appends missing ones.    
- [x] Update other parts of the code that might list the feeds (e.g. the main page).  
- [x] Revert to the default order (the second button mentioned above).  
- [x] Handle response from saving the custom order.


pe1uca:
Also something to note, the floppy icon was taken from [here](https://commons.wikimedia.org/wiki/File:Emojione_BW_1F4BE.svg) which states attribution is required.  
The other icons were shared under CC0.  

If required, also would need help on finding appropriated icons to include.  

ETA: The icon for A-Z sorting was taken from [here](https://commons.wikimedia.org/wiki/File:UI_Icon_A-Z_sort.svg)

app/Controllers/categoryController.php
```diff
@@ -194,6 +194,33 @@ public function updateAction(): void {
...
+		$order = Minz_Request::paramArrayInt('feedsOrder');
+
+		$categoryDAO = FreshRSS_Factory::createCategoryDao();
+		$category = $categoryDAO->searchById($id);
+		if ($id === 0 || null === $category) {
+			Minz_Error::error(404);
+			return;
+		}
+
+		$category->_attribute('feedsOrder', json_encode($order));
```
Alkarex: It should not be necessary to apply `json_encode()` here

app/Models/Category.php
```diff
@@ -280,7 +284,20 @@ private function sortFeeds(): void {
 		if ($this->feeds === null) {
 			return;
 		}
-		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b) => strnatcasecmp($a->name(), $b->name()));
+		$order = $this->attributeString('feedsOrder') ?? '[]';
```
Alkarex: Try `attributeArray()` instead, without `json_encode()`

Alkarex:
It only affects the feed order in the feed management view, and not in the aside bar, right?

pe1uca:
> It only affects the feed order in the feed management view, and not in the aside bar, right?

It is also for the sidebar, and should be for all screens where the feeds are listed per category.

Alkarex:
In my tests (SQLite), it works in the feed management view:

image

but appears buggy in the side bar:

image

Anyway, I would be interested in trying to use the same approach than for the custom ordering of categories.
I am taking a look at it

Inverle:
Should also make sure that the feed ordering works with OPML import/export.

Related fix for categories currently: https://github.com/FreshRSS/FreshRSS/pull/9203

PR #8710, opened 2026-04-21
simplepie: fix potential `XML error: No memory` when parsing huge feed

### LINKED ISSUE #8516: [Feed problem] projectzero.google XML error: No memory at line ...
### Feed URL

https://projectzero.google/feed.xml

### Describe the problem

Unable to add the feed likely because the feed xml is 18MB

### Log messages

https://projectzero.google/feed.xml is invalid XML, likely due to invalid characters. XML error: No memory at line 885, column 2167 [https://projectzero.google/feed.xml] 

### W3C Validator

Unable to validate, due to hardcoded resource limits (feed length > 5000000 bytes)

### FreshRSS version

1.28.1

### System information

FreshRSS version
    1.28.1
PHP version
    8.4.16
cURL version
    8.14.1
    OpenSSL/3.5.4
Database
    sqlite
    3.46.1
Server software
    Apache/2.4.66 (Debian)
    Linux 6.17.2-2-pve #1 SMP PREEMPT_DYNAMIC PMX 6.17.2-2 (2025-11-26T12:33Z) x86_64 
Browser
    Mozilla/5.0 (X11; Linux x86_64; rv:147.0) Gecko/20100101 Firefox/147.0

### Additional context

_No response_

### Curl commands

_No response_

### User agents

_No response_

LeeXiaolan:
After upgrade debian-based image from `freshrss:1.27.0` to `freshrss:1.28.1`, feed projectzero `https://projectzero.google/feed.xml` start failing to update, the log says `https://projectzero.google/feed.xml is invalid XML, likely due to invalid characters. XML error: No memory at line 1250, column 2167 [https://projectzero.google/feed.xml]`.

According to the PHP doc[1]:
>  XML_OPTION_PARSE_HUGE (int)
>     Available as of PHP 8.4.0. When libxml2 < 2.7.0 is used (e.g. on PHP 7.x), this option is enabled by default and cannot be disabled.

There is a new option `XML_OPTION_PARSE_HUGE` after PHP 8.4.0. And indeed there is an upgrade to PHP 8.4 between 1.27.0 and 1.28.1 in debian-based docker image. And the new default value is `false`[2].

Set this option to `1` fixed the problem for me.

1. https://www.php.net/manual/en/xml.constants.php#constant.xml-option-parse-huge
2. https://github.com/php/php-src/blob/90744ec52a50e0eacab47d5a51d82c1b26607b76/ext/xml/xml.c#L1116

This should Closes #8516
why alpine-based docker image is not affected, my wild guess is it do not use `libxml2`.

### DIFF
--- a/lib/simplepie/simplepie/src/Parser.php
+++ b/lib/simplepie/simplepie/src/Parser.php
@@ -139,6 +139,9 @@ class Parser implements RegistryAware
             $xml = xml_parser_create_ns($this->encoding, $this->separator);
             xml_parser_set_option($xml, XML_OPTION_SKIP_WHITE, 1);
             xml_parser_set_option($xml, XML_OPTION_CASE_FOLDING, 0);
+            if (80400 <= \PHP_VERSION_ID) {
+                xml_parser_set_option($xml, XML_OPTION_PARSE_HUGE, 1);
+            }
             xml_set_character_data_handler($xml, [$this, 'cdata']);
             xml_set_element_handler($xml, [$this, 'tag_open'], [$this, 'tag_close']);
 


Frenzie (approved):
Sounds good to me.

There is the aside that it should presumably go into https://github.com/FreshRSS/simplepie and preferably upstream as well, but @Alkarex normally deals with that.

Alkarex:
Reference: https://wiki.php.net/rfc/xml_option_parse_huge

> There is the aside that it should presumably go into https://github.com/FreshRSS/simplepie and preferably upstream as well, but @Alkarex normally deals with that.

@LeeXiaolan Could you please send the same PR to https://github.com/simplepie/simplepie ?
I will then take care of the rest.

Personally, I would slightly prefer a `if defined('XML_OPTION_PARSE_HUGE') {}` but that is a detail.

We also have to think of the drawbacks of activating this option (by default). We risk hitting more easily an out-of-memory condition (although it was the same before PHP 8.4), so it would be nice to improve how we handle this condition, in particular in a cron job of a system with multiple users. Another approach would be to make it a feed option.

Frenzie:
It's a pity that it's either a rather small limit or no limit at all.

LeeXiaolan:
> Could you please send the same PR to https://github.com/simplepie/simplepie ?
> I will then take care of the rest.

I just started a two-week-vacation, I will do it after back.

> so it would be nice to improve how we handle this condition

It seems `Parser` does not take any configurations so far, and I am not familiar with its design pattern, so any suggestion?

Alkarex:
* https://github.com/FreshRSS/FreshRSS/pull/8894

Alkarex:
@LeeXiaolan Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #8711, opened 2026-04-21
Nord: fix nav_menu sidebar toggle overlap and mark-read dropdown edge

### LINKED ISSUE #8707: Nord theme: two layout issues in the nav_menu button row
I ran into two small layout bugs in the Nord theme's top button row. Both only show up in Nord. Dark theme and base-theme are fine.

### 1. Sidebar toggle overlaps the first button

When I narrow the browser window, the folder/sidebar icon (`#nav_menu_toggle_aside`) ends up touching or overlapping the first button in `#nav_menu_actions` ("Show read"), while every other pair of buttons keeps a consistent ~4px gap.

The cause is in Nord's `.nav_menu` rule:

```css
/* Nord/nord.css, line 985 */
.nav_menu {
    padding: 5px 0;
    text-align: center;
}
```

`base-theme/frss.css` positions the sidebar toggle with `position: absolute; left: 0.5rem` and relies on `.nav_menu { padding-left: 3rem }` to reserve space for it. Nord's override replaces that padding entirely, so the absolute toggle sits on top of the first in-flow button.

Two possible fixes:

**A. Let the toggle flow inline.**

```css
.nav_menu #nav_menu_toggle_aside {
    position: static;
}
```

The toggle becomes part of the centered button row and picks up the same spacing as the other buttons. Side effect: it no longer anchors to the far-left edge of `.nav_menu`. This is the fix I've been running.

**B. Restore `padding-left` on `.nav_menu` in Nord.**

Keeps the left-anchored layout, but the right value depends on the toggle's rendered width. I tried several values (2.5rem through 5rem) before switching to option A; there doesn't seem to be a single value that looks right across all viewport widths.

Open to whichever you prefer.

### 2. "Mark as read" dropdown has a missing left border

At viewport widths where upstream hides the "Mark as read" text button (`base-theme/frss.css` does this inside `@media (max-width: 840px)`), the remaining dropdown toggle renders with no left border and flat left corners. It looks like half a button.

Nord's `.stick` rule removes the left border and radius from the dropdown toggle, so when the text button was visible they read as one connected control:

```css
/* Nord/nord.css, lines 326-332 */
.nav_menu .stick #mark-read-menu .dropdown-toggle.btn,
.nav_menu .dropdown-menu .stick.search .btn {
    margin-left: 0;
    border-left: 0;
    border-top-left-radius: 0;
    border-bottom-left-radius: 0;
}
```

Once the text button is hidden, the toggle stands alone with no left edge. The fix is to restore the border and radius inside the same media query:

```css
@media (max-width: 840px) {
    .nav_menu .stick #mark-read-menu .dropdown-toggle.btn {
        border-left: 1px solid var(--border-elements);
        border-top-left-radius: 6px;
        border-bottom-left-radius: 6px;
    }
}
```

Feel free to swap `var(--border-elements)` and the `6px` radius for whatever matches Nord's conventions. A note on placement: this selector has the same specificity as the rule it overrides at line 326, so it needs to come after that rule in source order (end of `nord.css` works, or directly after the original block).

### Environment

- FreshRSS 1.28.2-dev
- LibreWolf

### Screenshots

Issue 1, sidebar toggle collision at narrow widths:
- Before: Image
- After: Image

Issue 2, mark-read dropdown edge when the text button is hidden:
- Before: Image
- After: Image

Both fixes are already running on my instance via a small idempotent patch script at https://github.com/polybjorn/freshrss-patches, but they're small enough to drop straight into `p/themes/Nord/nord.css` if you want to take them as-is.

---

This issue was identified and written with the help of Claude Code. The fix was tested and confirmed manually.

polybjorn:
Fixes #8707

Two small layout issues in the Nord theme's top button row, both only visible in Nord.

### 1. Sidebar toggle overlapping the first button

`#nav_menu_toggle_aside` is absolute-positioned by `base-theme/frss.css`, which expects `.nav_menu` to reserve space via `padding-left: 3rem`. Nord's `.nav_menu` rule replaces that padding with `padding: 5px 0`, so at narrow widths the toggle overlaps the first button in the row.

Fix: set `position: static` on the toggle inside Nord so it flows inline with the other buttons.

I also considered restoring `padding-left` on `.nav_menu` instead, but couldn't find a single value that looked right across all viewport widths. Happy to switch approach if preferred.

### 2. Missing left border on the mark-read dropdown

At `max-width: 840px`, `base-theme` hides the "Mark as read" text button and only the dropdown toggle remains. Nord strips that toggle's left border and radius so it visually joins the text button (lines 326-332), but once the text button is hidden the toggle renders with a flat, borderless left edge.

Fix: restore the border and radius on the toggle inside the same media query.

### Notes

- `nord.rtl.css` regenerated with `make rtl`.
- `npm run stylelint` passes.
- Tested locally against FreshRSS 1.28.2-dev in LibreWolf.

Screenshots are on the linked issue.

### DIFF
--- a/p/themes/Nord/nord.css
+++ b/p/themes/Nord/nord.css
@@ -331,6 +331,14 @@ svg:hover {
 	border-bottom-left-radius: 0;
 }
 
+@media (max-width: 840px) {
+	.nav_menu .stick #mark-read-menu .dropdown-toggle.btn {
+		border-left: 1px solid var(--border-elements);
+		border-top-left-radius: 6px;
+		border-bottom-left-radius: 6px;
+	}
+}
+
 
 .stick .btn:last-child {
 	margin-right: 0;
@@ -985,6 +993,10 @@ li.item.active {
 .nav_menu {
 	padding: 5px 0;
 	text-align: center;
+}
+
+.nav_menu #nav_menu_toggle_aside {
+	position: static;
 }
 
 .nav_menu .btn {
--- a/p/themes/Nord/nord.css
+++ b/p/themes/Nord/nord.css
@@ -331,14 +331,6 @@ svg:hover {
 	border-bottom-left-radius: 0;
 }
 
-@media (max-width: 840px) {
-	.nav_menu .stick #mark-read-menu .dropdown-toggle.btn {
-		border-left: 1px solid var(--border-elements);
-		border-top-left-radius: 6px;
-		border-bottom-left-radius: 6px;
-	}
-}
-
 
 .stick .btn:last-child {
 	margin-right: 0;
@@ -1297,6 +1289,12 @@ optgroup::before {
 
 	.nav_menu .btn {
 		margin-bottom: .5rem;
+	}
+
+	.nav_menu .stick #mark-read-menu .dropdown-toggle.btn {
+		border-left: 1px solid var(--border-elements);
+		border-top-left-radius: 6px;
+		border-bottom-left-radius: 6px;
 	}
 
 	.nav_menu .search input {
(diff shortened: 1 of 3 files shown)


p/themes/Nord/nord.css
```diff
@@ -331,6 +331,14 @@ svg:hover {
 	border-bottom-left-radius: 0;
 }
 
+@media (max-width: 840px) {
```
Frenzie: That should go around line 1200 in the existing block, shouldn't it? :-)

Frenzie (approved):

polybjorn:
Force-pushed: signed both commits and rebased onto `edge`. No diff changes.

Inverle:
Regression: https://github.com/FreshRSS/FreshRSS/issues/8898

PR #8714, opened 2026-04-22
Add checkUrl() to favicon functions

Alkarex:


lib/favicons.php
```diff
@@ -68,7 +68,11 @@ function searchFavicon(string $url): string {
 
 		$iri = $href->get_iri();
 		if ($iri == false) {
-			return '';
+			continue;
+		}
+		$iri = FreshRSS_http_Util::checkUrl($iri, false);
```
Inverle: To be consistent with the usage in `download_favicon()` below
```suggestion
		$iri = FreshRSS_http_Util::checkUrl($iri, fixScheme: false);
```

PR #8715, opened 2026-04-23
fix: do not report PRIORITY_HIDDEN feeds in nbUnreadsPerFeed

### LINKED ISSUE #8694: "New articles available" banner loops every 2 minutes for hidden feeds with unreads
When a feed with priority "hidden" (-10) has unread articles, the "There are new articles available" banner shows up every poll cycle (2 min) and never stops. Clicking it reloads the page, but the banner comes back 2 minutes later. The articles themselves never appear since the feed is hidden.

**What happens:**

The `nbUnreadsPerFeed` endpoint returns unread counts for all feeds, including hidden ones. But the sidebar (`aside_feed.phtml:114`) doesn't render feeds with `priority < PRIORITY_FEED`, so there's no DOM element for them.

In `refreshUnreads()` (main.js), when there's no DOM element for a feed, `feed_unreads` falls back to 0. Since the server still reports unreads > 0, the difference is always > 0. And because `incUnreadsFeed` can't update a non-existent element, the JS-side count is stuck at 0. When viewing "All articles" (`isAll` is truthy), the banner triggers every cycle.

**To reproduce:**

1. Set a feed to "Do not show (hidden)"
2. Have or leave some unread articles in that feed
3. View "All articles"
4. Wait ~2 minutes - the banner appears
5. Click it, wait another 2 minutes - it appears again, indefinitely

**Suggested fix:**

Filter hidden feeds from the JSON response in `nbUnreadsPerFeed.phtml`, matching what the sidebar already does:

```php
foreach ($cat->feeds() as $feed) {
    if ($feed->priority() >= FreshRSS_Feed::PRIORITY_FEED) {
        $result['feeds'][$feed->id()] = $feed->nbNotRead();
    }
}
```

Alternatively, `refreshUnreads()` could skip feeds that have no DOM element.

I've tested the `nbUnreadsPerFeed.phtml` fix locally and it resolves the issue.

**Version:** 1.28.1 (also confirmed on 1.28.2-dev edge, code is identical)

polybjorn:
Fixes #8694

## Problem
When a feed has PRIORITY_HIDDEN, its unread count is still reported by
nbUnreadsPerFeed, but the sidebar does not render a DOM element for it
(aside_feed.phtml:114). In refreshUnreads(), the per-feed tracked count
for the missing element stays at 0, while the server keeps reporting
unreads > 0. On "All articles" (isAll), this triggers the "new articles
available" banner every 2 minutes indefinitely.

## Fix
Filter PRIORITY_HIDDEN feeds out of the nbUnreadsPerFeed response. This
matches the existing convention in Category.php:42, which uses
`$feed->priority() > FreshRSS_Feed::PRIORITY_HIDDEN` for the same
"don't count hidden unreads" decision. It also aligns with the SQL for
"All articles" (entryController.php:112-114), which already excludes
hidden feeds from that view.

## Alternatives considered
- Frontend guard in refreshUnreads() to skip feeds with no DOM element:
  treats the symptom, not the cause; any future consumer of the endpoint
  would re-inherit the bug.
- Sending priority in JSON and filtering client-side: duplicates a policy
  that already lives in the model layer.

## Notes
- Only consumer of nbUnreadsPerFeed is p/scripts/main.js:2044. No API,
  CLI, or extension impact.
- Edge case: when viewing a hidden feed directly, aside_feed.phtml
  renders it in the sidebar (the !$f_active exception). With this fix
  the sidebar count for that feed will not auto-refresh on that view.
  isAll is false there, so the banner does not loop in that case.
- Independent confirmation by @WillPresley on the issue.
- Tested locally against 1.28.2-dev.

### DIFF
--- a/app/views/javascript/nbUnreadsPerFeed.phtml
+++ b/app/views/javascript/nbUnreadsPerFeed.phtml
@@ -8,7 +8,9 @@ $result = [
 ];
 foreach ($this->categories as $cat) {
 	foreach ($cat->feeds() as $feed) {
-		$result['feeds'][$feed->id()] = $feed->nbNotRead();
+		if ($feed->priority() > FreshRSS_Feed::PRIORITY_HIDDEN) {
+			$result['feeds'][$feed->id()] = $feed->nbNotRead();
+		}
 	}
 }
 foreach ($this->tags as $tag) {


Alkarex:
Thanks 👍🏻

PR #8725, opened 2026-04-28
fix: use spec-compliant `no-referrer` meta value

### LINKED ISSUE #8718: [Bug] FreshRSS uses invalid meta referrer property
### Describe the bug

Default installations use the following meta tag to prevent leaking referrer info:
```
<meta name="referrer" content="never" />
```
But, according to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/referrer), spec only permits the following 8 content values for referrer:

1. no-referrer
2. origin
3. no-referrer-when-downgrade
4. origin-when-cross-origin
5. same-origin
6. strict-origin
7. strict-origin-when-cross-origin
8. unsafe-URL

### To Reproduce

1. Install FreshRSS
2. View source on any page

### Expected behavior

Comply as closely as possible to spec to elicit the expected behaviour from browsers; safeguard user privacy.

### FreshRSS version

1.28.0

### System information

- Database version: sqlite
- PHP version: 8.4 (lsphp)
- Installation type: git
- Web server type: OpenLiteSpeed 1.8.5 as a Docker container
- Device: n/a
- OS: n/a
- Browser: n/a

### Additional context

The extent to which "never" is respected by various browsers is unknown to me, but it seems like the best long term solution would be to replace the "meta referrer never" by "meta referrer no-referrer"

polybjorn:
`never` was removed from the Referrer Policy spec in 2016 and is only honoured by current browsers as a legacy alias. Replace with the spec-compliant `no-referrer` token.

## How tested

Each browser was pointed at a minimal local page that mirrors how FreshRSS uses the meta tag, then a link to a request inspector was clicked. Pass = no `Referer` header in the resulting request.

**Test page** (served via `python3 -m http.server 8000`):

~~~html
<!DOCTYPE html>
<html>
<head>
  <meta name="referrer" content="no-referrer">
</head>
<body>
  <a href="https://httpbin.org/headers">inspect headers</a>
</body>
</html>
~~~

**Procedure:** open `http://localhost:8000/test.html`, click the link, inspect the JSON `headers` object returned by httpbin.org.

**Results:**

| Browser | Engine | Version | `Referer` sent? |
|---|---|---|---|
| Safari | WebKit | 26 | No (pass) |
| LibreWolf | Gecko | 150 | No (pass) |
| Chromium | Blink | 149 | No (pass) |
| SeaMonkey | Gecko (legacy) | 2.53.23 | Yes (pre-existing, also leaks with `never`) |

The SeaMonkey result is unchanged behaviour: it ignores `<meta name="referrer">` for both `never` and `no-referrer`, so this PR neither fixes nor regresses it. Filed for awareness only.

Closes #8718

### DIFF
--- a/app/layout/layout.phtml
+++ b/app/layout/layout.phtml
@@ -45,7 +45,7 @@
 		<meta name="msapplication-TileColor" content="#FFF" />
 		<meta name="theme-color" content="#FFF" />
 <?php if (!FreshRSS_Context::systemConf()->allow_referrer) { ?>
-		<meta name="referrer" content="never" />
+		<meta name="referrer" content="no-referrer" />
 <?php } ?>
 		<?= FreshRSS_View::headTitle() ?>
 <?php
--- a/app/layout/simple.phtml
+++ b/app/layout/simple.phtml
@@ -33,7 +33,7 @@
 		<meta name="msapplication-TileColor" content="#FFF" />
 		<meta name="theme-color" content="#FFF" />
 <?php if (!FreshRSS_Context::systemConf()->allow_referrer) { ?>
-		<meta name="referrer" content="never" />
+		<meta name="referrer" content="no-referrer" />
 <?php } ?>
 		<?= FreshRSS_View::headTitle() ?>
 		<?php if ($this->rss_url != ''): ?>


zapovit:
Thanks for this, I am still new to github :smile_cat:

PR #8731, opened 2026-04-29
fix(themes): restore radius on lone .group buttons in nav_menu

### LINKED ISSUE #7405: [Bug] The border is different on the buttons
### Describe the bug

In desktop and mobile view the border radius is missing from sorting criteria button, and border radius missing from the Update feeds button left side.

In the mobile view left border is missing from the sorting criteria button and the toggle a side button. If I enabled the border the border radius is missing too from sorting criteria button.

Desktop

Image

Mobile

Image

### To Reproduce

1. Go to the FreshRSS home page and see the buttons.

### Expected behavior

Something like this?

Desktop

Image

Mobile

Image

### FreshRSS version

1.26.0

### System information

- PHP version: 8.4.4
- Installation type: zip
- Web server type: nginx
- Device: Desktop PC
- OS: Windows 11
- Browser: Google Chrome

### Additional context

_No response_

polybjorn:
Fixes #7405

Several themes inherit the same button-radius cascade bug: a single button inside a `.group` element ends up right-rounded only because the `.btn` matches both `:first-child` and `:last-child`, and the right-radius rule wins by source order. Visible on:

- The **Update feeds** button (lone `.btn` in `#nav_menu_actualize`)
- The **Toggle sidebar** button at narrow width (`#nav_menu_toggle_aside`)
- The **Sorting** button (`.btn` inside a lone `.dropdown` in `#nav_menu_sort`)

The fix uses targeted `:only-child` rules that give full radius to lone `.group` children, with selector form `.group > .btn:only-child, .group > .dropdown:only-child > .btn`.

In **Dark** and **Origine**, the Sorting button also needs an explicit `border-left` restored on the dropdown branch. Origine's `.group .dropdown:last-child > .btn { border-left: none }` strips the left border from a lone dropdown's button. Dark inherits this strip rule via its `../Origine/origine.css` import, so the same fix is applied in both themes, each re-asserting the border in its own variable (`--border-color` for Origine, `--dark-border-color0` for Dark). The radius rule and the border-left re-assertion are kept as separate blocks so the border-left only applies to the dropdown case, not to lone `.btn` children.

Themes affected directly: Alternative-Dark, Ansum, Dark, Flat, Mapco, Origine, Pafat. Dark-pink inherits via its `../Alternative-Dark/adark.css` import; Origine-compact inherits via its `../Origine/origine.css` import.

The same restore-radius / restore-border approach was approved in #8711 for Nord's mark-read toggle.

### Out of scope

The mark-read dropdown (`#nav_menu_read_all`) at narrow widths looks similar but is not fixed here. The dropdown is not a CSS `:only-child` of `.group` (its sibling `.read_all` button is still in the DOM, even when it visually wraps to a new row), so the rules in this PR do not match it. That case is handled in a separate upcoming PR with `@media` rules per theme.

#### Sorting button (visible-border theme)

Before:
origine visible-border original

After:
origine visible-border fixed

#### Update feeds button (lone .btn case)

Before:
mapco lone btn original

After:
mapco lone btn fixed

### Notes

- `npm run stylelint` passes.
- Tested locally on FreshRSS 1.28.2-dev across all affected themes.

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -178,6 +178,11 @@ th {
 	border-radius: 0 3px 3px 0;
 }
 
+.group > .btn:only-child,
+.group > .dropdown:only-child > .btn {
+	border-radius: 3px;
+}
+
 .stick .btn + .btn,
 .group .btn + .btn,
 .stick .btn + input,
--- a/p/themes/Alternative-Dark/adark.rtl.css
+++ b/p/themes/Alternative-Dark/adark.rtl.css
@@ -178,6 +178,11 @@ th {
 	border-radius: 3px 0 0 3px;
 }
 
+.group > .btn:only-child,
+.group > .dropdown:only-child > .btn {
+	border-radius: 3px;
+}
+
 .stick .btn + .btn,
 .group .btn + .btn,
 .stick .btn + input,
--- a/p/themes/Ansum/_sidebar.css
+++ b/p/themes/Ansum/_sidebar.css
@@ -126,6 +126,11 @@
 
 }
 
+.group > .btn:only-child,
+.group > .dropdown:only-child > .btn {
+	border-radius: 5px;
+}
+
 .aside {
 	background: var(--sid-bg);
 
--- a/p/themes/Ansum/_sidebar.rtl.css
+++ b/p/themes/Ansum/_sidebar.rtl.css
@@ -126,6 +126,11 @@
 
 }
 
+.group > .btn:only-child,
+.group > .dropdown:only-child > .btn {
+	border-radius: 5px;
+}
+
 .aside {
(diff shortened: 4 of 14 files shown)


Alkarex:
Just to be sure: Is it on purpose that your PRs are marked as draft?

polybjorn:
> Just to be sure: Is it on purpose that your PRs are marked as draft?

Yes - finishing screenshots and a final cross-theme check. Will start marking them now.
Should probably keep it in fork in the future..

Alkarex:
> > Just to be sure: Is it on purpose that your PRs are marked as draft?
> 
> Yes - finishing screenshots and a final cross-theme check. Will start marking them now. Should probably keep it in fork in the future..

All good with a draft PR. I just wanted to be sure it was on purpose

PR #8735, opened 2026-04-29
fix(themes): match hover border on adjacent .group/.stick buttons

polybjorn:
In Alternative-Dark and Dark-pink, hovering a button that isn't first in its `.stick` or `.group` row produces a 3-sided highlight (top, right, bottom) instead of all four. The strip rule that visually joins adjacent buttons (`.stick .btn + .btn, ... { border-left: none }`) wins by specificity over `.btn:hover { border: 1px solid <hover-color> }`, so the left edge stays missing through the hover state.

The fix adds `:not(:hover)` to the `.btn` targets in the strip rules, so the join lifts on hover and `.btn:hover` paints the full border. The second strip rule's `border-radius` was hoisted into its own block to keep the radius applying in all states.

Dark-pink inherits the modified rules and needs no changes of its own. The other themes don't change border on hover, so they're unaffected.

References #7405

#### Before: hovered button missing left border
hover border before

#### After: hover border on all four sides
hover border after

### Notes

- `npm run stylelint` passes.
- Verified on Alternative-Dark and Dark-pink.

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -178,25 +178,24 @@ th {
 	border-radius: 0 3px 3px 0;
 }
 
-.stick .btn + .btn,
-.group .btn + .btn,
+.stick .btn + .btn:not(:hover),
+.group .btn + .btn:not(:hover),
 .stick .btn + input,
-.stick .btn + .dropdown > .btn,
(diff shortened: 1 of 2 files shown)


Alkarex:
I have merged your other PRs. Could you please address the conflicts here?

PR #8747, opened 2026-05-01
fix(themes): restore sidebar slide animation at narrow viewports

polybjorn:
## Summary

The narrow-viewport sidebar slide animation has been dormant since #8201. The `transition: width 200ms linear` rule reaches `.aside` correctly (verified via DevTools computed values), but the inline `display: none` toggle in `toggle_aside_click()` blocks it: width can't interpolate to/from `display: none` without `transition-behavior: allow-discrete`.

## Fix

Replace the inline `display` manipulation with a CSS class `is-hidden`, applied only when `!isNarrow`. At narrow viewports `.aside` is already `position: fixed; width: 0; overflow: hidden`, so width-collapse alone hides it. At wide viewports the class is kept so the sidebar is fully removed from layout flow when toggled off.

Wide viewports remain animation-free by design: `.aside` is `display: table-cell` there, so animating width would reflow the article list every frame.

## Test plan

- [x] Narrow viewport (≤840px): sidebar slides on open and close
- [x] Wide viewport: toggle hides and shows instantly
- [x] Resize across breakpoint while open: no stuck state

## Follow-up

`prefers-reduced-motion: reduce` opt-out would be a nice add but is cleaner once the `.aside { transition }` rule is consolidated; currently lives in many theme files on `edge`.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -932,14 +932,15 @@ function toggle_aside_click(manual = true) {
 	}
 
 	const active = toggle_aside.classList.contains('active');
+	const isNarrow = window.matchMedia('(max-width: 840px)').matches;
 	if (active) {
 		toggle_aside.classList.remove('active');
 		aside.classList.remove('visible');
-		aside.style.display = 'none';
+		aside.classList.toggle('is-hidden', !isNarrow);
 	} else {
 		toggle_aside.classList.add('active');
 		aside.classList.add('visible');
-		aside.style.display = '';
+		aside.classList.remove('is-hidden');
 	}
 
 	if (manual && ['normal', 'reader'].includes(context.current_view)) {
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1330,6 +1330,10 @@ input[type="search"] {
 	vertical-align: top;
 }
 
+.aside.is-hidden {
+	display: none;
+}
+
 .aside + .close-aside {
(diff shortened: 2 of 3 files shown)


p/scripts/main.js
```diff
@@ -932,14 +932,15 @@ function toggle_aside_click(manual = true) {
 	}
 
 	const active = toggle_aside.classList.contains('active');
+	const isNarrow = window.matchMedia('(max-width: 840px)').matches;
```
Alkarex, line 935: Does this work with slightly older engines such as SeaMonkey?

polybjorn: Tested in SeaMonkey 2.53.x: matchMedia works, the new code path runs without errors.

The slide doesn't animate there, but that's pre-existing and unrelated: .aside.visible uses min(var(--width-aside, 320px), calc(100vw - 56px)), and SeaMonkey computes that statically but won't interpolate to a min()-derived endpoint (swapping it for a plain px value animates fine). This PR just exposes it by removing the display: none flip that was hiding the same snap everywhere.

Two options if it's worth fixing: drop the min() on .aside.visible (loses the viewport-minus-56px cap on tiny screens), or compute the width in JS and set it inline. The JS path duplicates the CSS sizing, runs on every toggle for every browser, and adds a second source of truth that can drift.

Alkarex: > if it's worth fixing

No need to fix as long as the main features are working

PR #8748, opened 2026-05-01
i18n: Improve Traditional Chinese translations

IEEE-754:
**Change(s)**
- Add more translations
- Small fixes

**File(s) changed**
`app/i18n/zh-TW/{admin,gen,index,install,sub}.php`


Alkarex:
Would you be able to run `make fix-all` ?

IEEE-754:
> Would you be able to run `make fix-all` ?

Sorry, I didn’t notice that README is updated by scripts.

PR #8764, opened 2026-05-02
feat: Add hash-based article navigation for back/forward button support

### LINKED ISSUE #8759: [Feature] close article on back button

I lost count of how many times I pressed the back button to close an article, and it sends me back to my browser start page instead of the overview of my feed.

### Use case

Maybe I am too used to Gmail where clicking a row opens the content in its own page, and maybe that's not the philosophy behind FreshRSS, but my brain can't stop hitting the back button to go back to the overview after unfolding an article.

Collapse the open article when the back button is pressed, then stop navigation.

Or even better, open articles in their own page, with bookmarkable URLs that create their own history entry.

### Additional context

Basically, just do what Gmail does. It's the same category of UI, and you can just copy it.

(But don't do what Google did to Google Reader 🪦.)

Thank you!

sliekens:
Closes #8759

## Changes proposed in this pull request:

This PR add hash-based article navigation using  `#article-{id}` hash fragments.

Benefits:

1. Support back/forward button functionality
    * Back button to collapse current article
    * Forward button to expand the previous article
2. Remember the expanded article after a full page refresh
3. Enable bookmarkable URLs for individual articles

Implementation changes:

- Add new `init_article_history()` function to handle hash change events and initial page load with hash
- Modify `toggleContent()` in `p/scripts/main.js` to update URL hash when articles are expanded or collapsed

1. Click on any article header to expand it (notice URL changes to `#article-{id}`), then press your browser's back button - the article should collapse instead of leaving FreshRSS
2. After collapsing via back button, press forward - the article should re-expand
3. Expand an article, do a full page refresh - the article should be automatically expanded on page load
4. Expand an article, copy the URL with the hash fragment, then paste it in a new tab - same result
5. Keyboard shortcuts like 'j' / 'k' / 'n' / 'p' update or remove the hash for the current article as expected

## Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

Unit tests: FreshRSS uses PHPUnit for backend tests; there are no JavaScript unit tests in this codebase, so this is N/A for JS changes

Documentation: no user attention required for this change.


p/scripts/main.js
```diff
@@ -2398,6 +2421,70 @@ see https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security`
...
+			};
+			// Try immediately, but also listen for DOMContentLoaded if needed
+			if (document.readyState === 'loading') {
+				document.addEventListener('DOMContentLoaded', expandArticle);
+			} else {
+				expandArticle();
+			}
+			// Prevent scrolling to the element
+			window.scrollTo(0, 0);
+		}
```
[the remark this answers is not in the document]
sliekens: Removing unnecessary parts

p/scripts/main.js
```diff
@@ -2398,6 +2421,70 @@ see https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security`
...
+		const oldHash = ev.oldURL ? ev.oldURL.split('#')[1] || '' : '';
+		const newHash = location.hash.substring(1); // Remove leading #
+
+		// If new hash is an article ID (starts with 'article-')
+		if (newHash.startsWith('article-')) {
+			const articleId = newHash.substring(8); // Remove 'article-' prefix
+			const article = document.getElementById('flux_' + articleId);
+			if (article && !article.classList.contains('active')) {
+				articleHistoryIgnoreToggle = true;
+				toggleContent(article, article, false);
```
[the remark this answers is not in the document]
sliekens: Fixing to pass the current article as old_active when expanding via hash navigation, ensuring the .current marker is updated consistently.

p/scripts/main.js
```diff
@@ -2398,6 +2421,70 @@ see https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security`
...
+				articleHistoryIgnoreToggle = false;
+			}
+			// Prevent scrolling to the element
+			ev.preventDefault();
```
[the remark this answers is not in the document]
sliekens: Removing

p/scripts/main.js
```diff
@@ -2398,6 +2421,70 @@ see https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security`
...
+		} else if (oldHash.startsWith('article-') && !newHash.startsWith('article-')) {
+			// Hash removed - collapse any expanded article
+			const oldArticleId = oldHash.substring(8);
+			const article = document.getElementById('flux_' + oldArticleId);
+			if (article && article.classList.contains('active')) {
+				articleHistoryIgnoreToggle = true;
+				toggleContent(article, article, false);
+				articleHistoryIgnoreToggle = false;
+			}
+		}
```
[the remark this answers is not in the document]
sliekens: Changed to collapse only when the hash is truly cleared

Frenzie:
What concretely does it do when navigating through multiple articles?

sliekens:
**What concretely happens when navigating through multiple articles?**
Each article you open is added as a new history entry. So if you open several articles in a row, pressing “back” will take you through them in reverse order, one at a time.

If you’d rather, I can change this behavior so that only a single history entry is kept. In that case, navigating between articles wouldn’t build up a stack, but that would require a bit more implementation work.

Frenzie:
I think I'd prefer the latter, but I suppose we'll see what the others have to say.

sliekens:
In fact I would also prefer the single history entry, which aligns better with the intended design of getting back to the overview quickly.

Current behavior:

https://github.com/user-attachments/assets/0d9c01e3-0903-4da2-b660-3bbaa569e71a

sliekens:
Updated to the single history entry, I believe this is closer to what most people expect when pressing back:

https://github.com/user-attachments/assets/ee6fb9fb-c831-472e-ae17-a34724465cb4

---

The difference is especially noticeable on mobile 🙂.

Frenzie:
The interaction's a bit odd if you go back to a page where the article ID doesn't exist because it was from further down, then open some article, scroll down, and press back. Instead of closing the opened article it'll now open the previously opened article.

sliekens:
@Frenzie not completely sure I understood, but I identified a problem where article IDs could still be stacked in some circumstances, like when the article is not in view. I think that might explain what you saw, and is fixed in https://github.com/FreshRSS/FreshRSS/commit/3dc1cb3ffbc1aae22f81d1f331581076554b08a3.

sliekens:
Quick update: this PR is now rebased on latest `edge`, I fixed another bug where history grows by 2 items, and I untangled the new code from existing code.

In my first attempt, history management was wired into `toggleContent`, touching existing code (risky) and making it harder to understand, which I guess is why this PR did not receive any code reviews :).

Now, history management is fully encapsulated in `init_article_history`. It is activated by `freshrss:openArticle` and `freshrss:closeArticle` events raised by `toggleContent`.

p/scripts/main.js
```diff
@@ -2442,6 +2451,81 @@ see https://freshrss.github.io/FreshRSS/en/admins/10_ServerConfig.html#security`
...
+	});
+	document.addEventListener('freshrss:closeArticle', function (ev) {
+		if (!ev.detail.fromHistory) {
+			history.replaceState({}, '', location.href.split('#')[0]);
+			// entryPushed stays true — next expand replaces this entry
+		}
+	});
+
+	// Handle browser back/forward
+	window.addEventListener('hashchange', function (ev) {
```
Inverle, line 2507: Wouldn't it be better to handle this with <https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event>?

See:
https://github.com/FreshRSS/FreshRSS/blob/339d5c7ef128e7cf4e50395c6e8c671179ac8acc/p/scripts/main.js#L2461-L2468

The current approach with just `hashchange` seems error-prone.

sliekens: @Inverle it would leave some older browser behind, is that acceptable?

The referenced slider fix appears to be an optional enhancement, it's a no-op when the navigation API is not supported.

```js
if (!('navigation' in window)) { 
    return; 
} 
```

I would probably have to do the same check, given the table below.


| API | Chrome/Edge | Firefox | Firefox ESR / IceCat | Safari | Global support |
|---|---|---|---|---|---|
| Navigation API (`navigate` event) | 102+ (2022) | 147+ (Nov 2025) | ❌ not in ESR 140 | 26.2+ (2025) | ~88% |
| `hashchange` (current) | 5+ (2010) | 3.6+ (2010) | ✅ all ESRs | 5+ (2010) | ~97% |

Inverle: Though I may be misunderstanding this PR, is it supposed to store only one article in the history, or multiple?
As in, is it used only used for expanding/collapsing a single article with the back/forward button, or also for navigating between articles?

Inverle: Oh right, it used to navigate between multiple articles, but doesn't anymore.
I think `hashchange` is probably fine then.

Inverle:
New eslint errors:
```
Error:   557:20  error  Object properties must go on a new line if they aren't all on the same line  @stylistic/object-property-newline
Error:   562:20  error  Object properties must go on a new line if they aren't all on the same line  @stylistic/object-property-newline
```

sliekens:
@Inverle thanks for sniffing that out, fixed and made sure that eslint and all other checks pass on my machine, next workflow run should be green.

p/scripts/main.js
```diff
@@ -549,11 +547,24 @@ function toggleContent(new_active, old_active, skipping) {
 		}
 	}
 
-	if (new_active.classList.contains('active') && !skipping) {
-		if (context.auto_mark_article) {
-			mark_read(new_active, true, true);
+	// URL history management listens to these events — see init_article_history()
```
Inverle:
```suggestion
	// URL history management listens to these events - see init_article_history()
```

p/scripts/main.js
```diff
@@ -549,11 +547,24 @@ function toggleContent(new_active, old_active, skipping) {
...
+	if (!skipping) {
+		if (new_active.classList.contains('active')) {
+			if (context.auto_mark_article) {
+				mark_read(new_active, true, true);
+			}
+			new_active.dispatchEvent(new CustomEvent('freshrss:openArticle', {
+				bubbles: true,
+				cancelable: true,
```
Inverle: If you intend to make the event cancelable, then the return value of `dispatchEvent` should probably be checked. Otherwise I would remove this.

p/scripts/main.js
```diff
@@ -549,11 +547,24 @@ function toggleContent(new_active, old_active, skipping) {
...
+				detail: { articleId: new_active.id.replace(/^flux_/, ''), fromHistory },
+			}));
+		} else if (old_active === new_active) {
+			new_active.dispatchEvent(new CustomEvent('freshrss:closeArticle', {
+				bubbles: true,
+				cancelable: true,
```
Inverle: Same here

Inverle:
There is a bug when you use n/p shortcuts (Focus next/previous without opening), even though the current article was closed, the hash remains.
Also called: `skip_next_entry` and `skip_prev_entry`

Inverle (changes requested):

sliekens:
> There is a bug when you use n/p shortcuts (Focus next/previous without opening), even though the current article was closed, the hash remains.
> Also called: `skip_next_entry` and `skip_prev_entry`

Confirmed, that path needs to raise the same `freshrr:closeArticle` event, preparing a fix now.

p/scripts/main.js
sliekens: I'll remove `cancelable` because the article state has already changed at this point and canceling the event wouldn't help. One nuance: `freshrss:openArticle` was already cancelable before this PR, but the return value was never checked so removing it should be harmless.

Inverle:
> > There is a bug when you use n/p shortcuts (Focus next/previous without opening), even though the current article was closed, the hash remains.
> > Also called: `skip_next_entry` and `skip_prev_entry`
> 
> Confirmed, that path needs to raise the same `freshrr:closeArticle` event, preparing a fix now.

It seems that the fix doesn't work correctly, the hash is only removed if I uncollapse an article twice, and only then use n or p. (instead of it working right away)

sliekens:
@Inverle what browser has the issue? I tried Edge and Firefox but I can't reproduce the problem. I also tried 'j'/'k' and found those also update the article as expected in those browsers.

Inverle:
Also Firefox, will take a closer look.

Issue #8771, opened 2026-05-03
UI regression: clicking logo opens sidebar


Alkarex:
A recent merge introduced a regression: in some cases, clicking the FreshRSS logo opens the sidebar.
More frequent on narrow view.

I cannot dig more into it just now, so help welcome. Maybe for @polybjorn

Inverle:
Caused by https://github.com/FreshRSS/FreshRSS/pull/8747

polybjorn:
Looking into it now.

PR #8774, opened 2026-05-03
fix(layout): reserve logo image box via intrinsic dimensions

polybjorn:
## Summary

Found while investigating #8771: spam-clicking the FreshRSS logo would briefly flash a giant logo on screen between navigations. Couldn't reproduce it deterministically afterwards, but the underlying cause is real and well-known: the default `` has no intrinsic `width`/`height`, so the browser has nothing to allocate a correctly-proportioned box from until either the CSS applies or the image is decoded. On a fast back-to-back navigation a cached image can paint at native size for a frame.

## Fix

Add the SVG's intrinsic dimensions (`1280 × 256`) as `width` and `height` attributes on the default logo `` in `header.phtml` and `simple.phtml`. CSS still controls rendered height (2rem wide, 24px narrow); width auto-derives from the intrinsic ratio. Custom `logo_html` is untouched.

This is documented best practice for `` to prevent [Cumulative Layout Shift](https://web.dev/articles/optimize-cls#images_without_dimensions) regardless of whether the flash is currently observable.

### DIFF
--- a/app/layout/header.phtml
+++ b/app/layout/header.phtml
@@ -6,7 +6,7 @@
 	<div class="item title">
 		<a href="<?= Minz_Url::display(['c' => 'index', 'a' => 'index'], 'html', 'root') ?>">
 			<?php if (FreshRSS_Context::systemConf()->logo_html == '') { ?>
-				<img class="logo" src="<?= _i('FreshRSS-logo', FreshRSS_Themes::ICON_URL) ?>" alt="FreshRSS" loading="lazy" />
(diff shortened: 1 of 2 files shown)


Frenzie:
You should set something like 160 × 32 on it. 1280 × 256 is simply the giant logo unconstrained by CSS.

The layout shift from an unloaded image without `width` and `height` set is the reverse. That's from 0 × 0 or similar to whatever the actual size is.

polybjorn:
> 1280 × 256 is simply the giant logo unconstrained by CSS.

Thanks, you're right. I anchored on "use the image's intrinsic dimensions" from the CLS guidance without thinking through that those *are* the unstyled-frame size.

Will push 160 × 32, matches the rendered size.

Frenzie (approved):

PR #8780, opened 2026-05-03
feat(http): apply Force HTTPS to all fetches; block 30x downgrades

### LINKED ISSUE #7260: Use the Force HTTPS mechanism also for favicons, Web scraping, etc.
At the moment, *the Force HTTPS* mechanism is only enforced in flows going through SimplePie.

https://github.com/FreshRSS/FreshRSS/discussions/7252#discussioncomment-11951305
https://github.com/FreshRSS/FreshRSS/pull/7259

polybjorn:
Closes #7260. Also fixes the favicon symptom in #7252.

## Changes

- Apply `force-https.default.txt` and `data/force-https.txt` to every HTTP fetch, not just feed content. Rewrite happens at the `FreshRSS_http_Util::httpGet` entry point and at the two `curl_init` sites that bypass it: `Feed::pubSubHubbubSubscribe` and the auto-update check in `updateController`.
- `lib/favicons.php` rewrites before deriving `sha1($url)`, so a cached failed-HTTP entry no longer shadows a successful HTTPS fetch.
- For listed hosts, restrict `CURLOPT_REDIR_PROTOCOLS_STR` to `https` so libcurl cannot follow a 30x to plain HTTP. This is the one default-on behaviour change.
- New `force_https_list_enabled` config flag (default `true`) disables both rewrite and redirect protection.
- Adds `tests/app/Utils/httpUtilTest.php` with 11 cases covering rewrite, subdomain match, host case-normalisation, the #7252 cache-key regression, off-switch fail-closed, and immediate reload on `data/force-https.txt` edit.

## How to test

1. Add a domain to `data/force-https.txt`, refresh a feed whose advertised icon is `http://that-domain/...`. With `simplepie_syslog_enabled`, syslog shows `FreshRSS GET ico https://...`.
2. Reproducer for #7252: pre-populate `data/cache/<sha1(http URL)>.html` with a failed marker, then fetch through the favicon path. Without this PR the marker is served back; with this PR the cache key derives from the rewritten HTTPS URL, the poisoned file is bypassed, and a fresh fetch happens.
3. Redirect downgrade block: add `httpbingo.org` to `data/force-https.txt`, fetch `https://httpbingo.org/redirect-to?url=http%3A%2F%2Fhttpbingo.org%2Fget`.
4. Off-switch: set `'force_https_list_enabled' => false` in `data/config.php`, confirm rewriting and redirect protection stop.

## Verified on a live install

`#7252` cache bypass:

```
poisoned http-keyed cache:  data/cache/8674ce5a...html  ("<!-- POISONED -->")
caller-derived cache path:  data/cache/d7b3438d...html
status:                     200
effective_url:              https://github.com/
body length:                551303
poisoned file unchanged?    yes
https-keyed file written?   yes
```

Redirect downgrade block (`httpbingo.org` on the list, https -> http hop):

```
status:         302
redirect_count: 1
fail:           true
error:          Protocol "http" not supported or disabled in libcurl
```

Off-switch (`'force_https_list_enabled' => false`):

```
flag value:             false
forceHttps(github.com): http://github.com/x   (unchanged)
isForceHttpsHost:       false
```

Auto-update and WebSub `curl_init` sites (same `forceHttps()` pattern):

```
http://github.com/foo/?v=1.28.2-dev      ->  https://github.com/foo/?v=1.28.2-dev
http://websub.example.gnu.org/           ->  https://websub.example.gnu.org/
http://pubsubhubbub.appspot.com/         ->  http://pubsubhubbub.appspot.com/  (not on list)
```

## Pull request checklist

- [x] clear commit messages
- [x] code manually tested
- [x] unit tests written
- [ ] documentation updated


Frenzie:
That setting name seems a bit misleading without reading the accompanying text. force_https_list_enabled or file or something?

lib/favicons.php
```diff
@@ -74,6 +75,7 @@ function searchFavicon(string $url): string {
 		if (!is_string($iri) || $iri === '') {
 			continue;
 		}
+		$iri = FreshRSS_http_Util::forceHttps($iri);
```
Alkarex, line 78: Is it maybe sufficient with the `httpGet()` call just after, which already calls `forceHttps()` ?

lib/favicons.php
```diff
@@ -113,7 +117,7 @@ function download_favicon(string $url, string $dest): bool {
 			$favicon = searchFavicon($url);
 		}
 		if ($favicon == '') {
-			$link = FreshRSS_http_Util::checkUrl($rootUrl . 'favicon.ico', fixScheme: false) ?: '';
+			$link = FreshRSS_http_Util::forceHttps(FreshRSS_http_Util::checkUrl($rootUrl . 'favicon.ico', fixScheme: false) ?: '');
```
Alkarex, line 120: I have not checked the details, but could it be an idea to have the `forceHttps()` inside `checkUrl()` ?

lib/favicons.php
polybjorn: The rewrite here is for the cache key, not the fetch. `faviconCachePath($iri)` is `sha1($iri) . '.ico'`, computed by the caller and passed into `httpGet()` as-is. `httpGet()` rewrites the URL it fetches but doesn't recompute the cache path, so without the rewrite at this call site the key stays `sha1(http URL)` and old failed-HTTP cache files keep shadowing successful https fetches. Same reason for the calls on 98, 111 and 120.

A cleaner long-term shape would be for `httpGet()` to derive the cache path itself from the rewritten URL (callers pass a cache directory + type instead of a full path), so the rewrite-then-hash ordering can't be wrong by accident. Maybe out of scope here, but happy to do that as a follow-up if it sounds useful.

lib/favicons.php
polybjorn: That'd work for the cache-key issue, since the favicon call sites all do `checkUrl()` before `faviconCachePath()`. One thing I'm not sure about is that `checkUrl()` is also used outside fetches: in `subscriptionController` and `Feed::_url` / `Feed::_website` it validates URLs that get persisted to the DB, and `pshb.php` uses it for a self-URL identity comparison. Putting the rewrite inside `checkUrl()` would apply it there too, so stored URLs end up as https even if a host later comes off the list, and the pshb comparison could change. It'd also pull `FreshRSS_Context` and the host list into `checkUrl()`, which is currently just a syntactic check.

If you'd rather have it centralised in `checkUrl()`, makes sense as its own PR so the persistence change is visible on its own.

polybjorn:
@Frenzie After a second pass, I find no better renaming alternative than what you suggest. Replaced. Went with `force_https_list_enabled` since "list" covers both the default and `data/` files merged into one runtime list, and both behaviours (rewrite + redirect block) key off that list.

PR #5638, opened 2023-09-06
Import TT-RSS categories

robertdahlem:
Extension to JSON imports: interpret
```
	  "origin": {
		"category": "..."
	  },
```
as category for FreshRSS. When importing a new feed configure the feed into this category. Creates missing categories on the fly.

This is needed by my upcoming TT-RSS exporter. Sorry that this is not relative to `edge` but relative to PR #5629 because I need both for testing.

PR #8786, opened 2026-05-04
feat(import): support category field in JSON feed import

polybjorn:
Supersedes #5638 (open since Sept 2023).

Picks up @robertdahlem's original work, cherry-picked verbatim so authorship is preserved (along with @Alkarex's syntax fix). Rebased onto current `edge` with one polish commit on top.

## What it does

JSON feed import now reads `origin.category` and places the feed in that category, creating it on-the-fly if missing. Without the field, behaviour is unchanged.

```json
{
  "items": [
    {
      "guid": "…",
      "origin": {
        "title": "…",
        "feedUrl": "…",
        "category": "Tech News"
      }
    }
  ]
}
```

Mainly useful for tools producing JSON imports (the TT-RSS exporter that motivated the original PR, hand-edited JSON, scripted provisioning). FreshRSS's own JSON export does not emit `origin.category` today; that would be a separate change.

## What the polish commit changes

- `!== false` after `addCategory()` (per @ColonelMoutarde's earlier review).
- Typed property syntax for `$categoryDAO`, matching `$entryDAO`/`$feedDAO`.
- Pass `$username` to `createCategoryDao()` in `importFile()` so `cli/import-for-user.php --user X` creates categories on the right account.
- Trim the category name; skip if empty after trim. Stops a whitespace-only `"category"` value from creating an empty-named row, and lets `"  Tech News  "` reuse an existing `"Tech News"`.

The `$username` fix could be split out if a tighter PR is preferred.

## Verification

CLI smoke tests via `cli/import-for-user.php`:

- New category created on first occurrence
- Existing category reused (no duplicates)
- `"  Tech News  "` reuses `"Tech News"` after trim
- `"   "` falls back to the default category
- Missing field falls back to the default category
- `--user` propagation works
- 191-char truncation in `addCategory` behaves as expected

`composer test` clean: 624 PHPUnit / 1160 assertions, PHPStan 0 errors, phpcs and lint clean.

A PHPUnit test could be added, but there are no existing tests for `importExportController` to mirror, so placement isn't obvious.

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -581,10 +581,7 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 		$cat_name = trim($origin['category'] ?? '');
 		if ($cat_name !== '') {
 			$new_cat = $this->categoryDAO->searchByName($cat_name);
-			$new_cat_id = $new_cat === null ? $this->categoryDAO->addCategory(['name' => $cat_name]) : $new_cat->id();
-			if ($new_cat_id !== false) {
-				$cat_id = $new_cat_id;
-			}
+			$cat_id = $new_cat?->id() ?: $this->categoryDAO->addCategory(['name' => $cat_name]) ?: FreshRSS_CategoryDAO::DEFAULTCATEGORYID;
 		}
 
 		try {
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -10,8 +10,7 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 
 	private FreshRSS_FeedDAO $feedDAO;
 
-	/** @var FreshRSS_CategoryDAO */
-	private $categoryDAO;
+	private FreshRSS_CategoryDAO $categoryDAO;
 
 	/**
 	 * This action is called before every other action in that class. It is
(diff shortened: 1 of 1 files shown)


ColonelMoutarde:
Good for me :D

app/Controllers/importExportController.php
```diff
@@ -573,10 +577,20 @@ private function addFeedJson(array $origin): ?FreshRSS_Feed {
...
 		$name = empty($origin['title']) ? $website : $origin['title'];
 
+		$cat_id = FreshRSS_CategoryDAO::DEFAULTCATEGORYID;
+		$cat_name = trim($origin['category'] ?? '');
+		if ($cat_name !== '') {
+			$new_cat = $this->categoryDAO->searchByName($cat_name);
+			$new_cat_id = $new_cat === null ? $this->categoryDAO->addCategory(['name' => $cat_name]) : $new_cat->id();
+			if ($new_cat_id !== false) {
+				$cat_id = $new_cat_id;
+			}
```
Alkarex: The proposed syntax is just fine. Alternatively, we can use null-safe approach now that we only support PHP 8.0+, which is a bit shorter, and maybe conveying a more direct intent:


```suggestion
			$cat_id = $new_cat?->id() ?: $this->categoryDAO->addCategory(['name' => $cat_name]) ?: FreshRSS_CategoryDAO::DEFAULTCATEGORYID;
```

If you like it, update, otherwise fine to stick with the existing one.
Note that it also handles the case when category ID may be unset (0).

Alkarex (approved):

app/Controllers/importExportController.php
polybjorn: Worth noting it also tightens behaviour: auto-increment starts at 1 and DEFAULTCATEGORYID = 1, so id 0 means "unset/invalid", and falling back to default for that case is more correct than the solution it replaces. Applied.

PR #8789, opened 2026-05-04
Fix search in shared user queries

Alkarex:
The internal search was shown in the UI as the user search. This was due to the lazy nature of the Generator. Improve the try/catch behaviour at the same time.

How to test:
* Make a user query with a search parameter
* Share the user query as HTML
* Observe the search field (should be empty with this PR,  while it contained the internal search before)

### DIFF
--- a/p/api/query.php
+++ b/p/api/query.php
@@ -117,8 +117,10 @@ $view = new FreshRSS_View();
 
 try {
 	FreshRSS_Context::updateUsingRequest(false);
-	Minz_Request::_param('search', $userSearch->toString());	// Restore user search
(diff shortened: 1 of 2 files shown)


p/api/query.php
```diff
@@ -117,8 +117,10 @@
 
 try {
 	FreshRSS_Context::updateUsingRequest(false);
-	Minz_Request::_param('search', $userSearch->toString());	// Restore user search
 	$view->entries = FreshRSS_index_Controller::listEntriesByContext();
+	$view->entries->current();	// Init the generator to consume the aggregated search and catch potential exceptions
```
Alkarex, line 121: https://github.com/FreshRSS/FreshRSS/pull/8863

PR #8791, opened 2026-05-05
feat(display): expose show_favicons toggle in Display settings

polybjorn:
The `show_favicons` user config flag has been a hidden setting in `data/users/<user>/config.php` with no UI exposure. This PR surfaces it as a checkbox on the Display settings page (below the Thumbnail dropdown), and resolves an overlap with the existing `topline_website` (Website) dropdown.

## What changes

1. New checkbox "Show feed icons in lists" on Display settings, bound to `show_favicons`. Default remains `true`, so existing users see no behaviour change.
2. `entry_header.phtml`: removes the `show_favicons` gate from the normal-view article topline. The `topline_website` dropdown becomes the sole authority for icons in that surface.

image

## Why the entry_header change

Before this PR, both flags gated the topline icon:

```php
if (show_favicons && 'name' !== topline_website) { /* icon */ }
if ('icon' !== topline_website) { /* name */ }
```

With the checkbox newly visible to users, the combination Website='Icon only' + Show feed icons=off would produce an empty website slot in the topline (no icon, no name). Dropping the `show_favicons` gate makes scope cleaner: Website dropdown controls the topline; the checkbox controls every other surface (sidebar, global view, reader view, opened-article header, subscription page, stats page).

The behaviour change only affects users who manually set `show_favicons=false` in config AND have `topline_website` set to `icon` or `full`. Since the flag was never UI-exposed, that population is likely small.

## Out of scope

- Active-feed sidebar icon exception in `aside_feed.phtml` (`||$f_active`) is unchanged.
- `simplify_over_n_feeds` sidebar gating is unchanged.

## Test plan

- Display settings page renders the new checkbox below the Thumbnail dropdown.
- Toggling it persists to `data/users/<user>/config.php`.
- Off: icons hidden in sidebar, normal view collapsed rows (when `topline_website` already excludes them), global view, reader view, opened article inner header.
- On: icons return in all the above.
- Conflict scenario: Website='Icon only' + checkbox off shows the icon in the topline (proving topline is now solely controlled by the dropdown).
- Active-feed sidebar exception still shows that one feed's icon (unchanged behaviour).

`make test-all` clean: 624 PHPUnit / 1160 assertions, PHPStan 0 errors, phpcs/eslint/stylelint/markdownlint clean.

### DIFF
--- a/app/i18n/en/conf.php
+++ b/app/i18n/en/conf.php
@@ -53,6 +53,7 @@ return array(
 			'seconds' => 'seconds (0 means no timeout)',
 			'timeout' => 'HTML5 notification timeout',
 		),
+		'show_feed_icons_in_lists' => 'Show feed icons in lists',
 		'show_nav_buttons' => 'Show the navigation buttons',
 		'show_title_unread' => 'Show number of unread articles in the title',
 		'sidebar_hidden_by_default' => 'Hide sidebar by default',
--- a/app/i18n/fr/conf.php
+++ b/app/i18n/fr/conf.php
@@ -53,7 +53,7 @@ return array(
 			'seconds' => 'secondes (0 signifie aucun timeout)',
 			'timeout' => 'Temps d’affichage de la notification HTML5',
 		),
-		'show_feed_icons_in_lists' => 'Show feed icons in lists',	// TODO
+		'show_feed_icons_in_lists' => 'Afficher les favicônes des flux dans les listes',
 		'show_nav_buttons' => 'Afficher les boutons de navigation',
 		'show_title_unread' => 'Afficher le nombre d’articles non lus dans le titre',
 		'sidebar_hidden_by_default' => 'Masquer la barre latérale par défaut',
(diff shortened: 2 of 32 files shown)


app/i18n/en/conf.php
```diff
@@ -53,6 +53,7 @@
 			'seconds' => 'seconds (0 means no timeout)',
 			'timeout' => 'HTML5 notification timeout',
 		),
+		'show_feed_icons_in_lists' => 'Show feed icons in lists',
```
Alkarex, line 56: Why the *in lists* part?

Alkarex: Ok, I get it:

>  Website dropdown controls the topline; the checkbox controls every other surface (sidebar, global view, reader view, opened-article header, subscription page, stats page).

app/i18n/fr/conf.php
```diff
@@ -53,6 +53,7 @@
 			'seconds' => 'secondes (0 signifie aucun timeout)',
 			'timeout' => 'Temps d’affichage de la notification HTML5',
 		),
+		'show_feed_icons_in_lists' => 'Show feed icons in lists',	// TODO
```
```suggestion
		'show_feed_icons_in_lists' => 'Afficher les favicônes des flux dans les listes',
```

Alkarex:
Is it maybe an error that all those PRs are closed? @polybjorn

polybjorn:
It is - since it wasn't directly salvageable, I've had to reconsider my process first..

I will use the opportunity to be more conservative with what I push for the future.

PR #8793, opened 2026-05-05
Add international calendar support and enhance date formatting using JS Intl API

qasedak:
Adds an optional display setting to reformat all dates into an alternative calendar system (Persian/Jalali, Islamic, Hebrew, Chinese, Japanese, etc.) entirely client-side using the browser's built-in `Intl.DateTimeFormat` API.

## What changes

- **New user setting** — "Calendar" dropdown in Display preferences (between Timezone and Theme). Defaults to blank (disabled), so existing users are unaffected.
- **`p/scripts/dates.js`** (new) — IIFE-scoped script that reformats every `<time datetime="...">` element on the page. Loaded only when the setting is non-blank, so there is zero overhead for users who don't use it.
  - Uses `IntersectionObserver` to convert dates lazily as they scroll into view (performance for long article lists).
  - `MutationObserver` handles dates injected after load (infinite scroll, AJAX).
  - 3-candidate locale fallback chain so unsupported locale/calendar combos degrade gracefully rather than silently failing.
  - Persists settings to `localStorage` so bfcache-restored pages reformat in-place without a full reload.
  - Pauses `MutationObserver` when the tab is hidden.
- **Semantic HTML improvement** (independent of the feature) — several views that rendered bare date strings now use `<time datetime="...">`, which also benefits screen readers and other tooling:
  - `app/views/index/normal.phtml` (two occurrences)
  - `app/views/user/manage.phtml` — last user activity column
  - `app/views/user/details.phtml` — last user activity field
  - `app/Controllers/updateController.php` — last update time
  - `app/Controllers/indexController.php` — entry transition header

## Calendars supported

Gregorian · Persian (Jalali) · Islamic Civil · Islamic Umm al-Qura · Hebrew · Chinese · Japanese · Buddhist (Thai) · Ethiopic · Republic of China · Indian National


Alkarex:
Thanks for the PR.
Open question: Is it sufficient to perform this conversion client side? There are also for instance server-side searches, filters, etc. Is there a risk of conflict?

qasedak:
> Thanks for the PR. Open question: Is it sufficient to perform this conversion client side? There are also for instance server-side searches, filters, etc. Is there a risk of conflict?

There was indeed a conflict with `date:`, `pubdate:`, `mdate:`, and `userdate:` search tokens — I've addressed it in 946a2ab.

The fix works in two directions:

- **On submit**: a `submit` listener rewrites calendar date tokens to Gregorian ISO before the request hits the server (e.g. `date:1402-10` (Jalali) → `date:2023-12-22/2024-01-20`), so the server never sees non-Gregorian dates.
- **On page load**: Gregorian ISO tokens in the search input are converted back to the user's calendar for display (e.g. `date:2023-12-22/2024-01-20` → `date:1402-10`), collapsing whole-year/month ranges to compact form to avoid confusion or double-conversion on result pages.

As for client-side being sufficient in general: the feature is purely presentational — it reformats `<time datetime="...">` elements, where the `datetime` attribute always holds a machine-readable ISO timestamp regardless of what's displayed. The server stores and queries Unix timestamps, so there's no structural conflict. I've been using the `Intl` API for calendar display in production for years and haven't run into issues beyond an occasional brief rendering flash on slow connections, which the `IntersectionObserver` approach here mitigates.

Alkarex:
If you can, please try to run `make fix-all`

qasedak (commented):
> If you can, please try to run `make fix-all` 

done

PR #8799, opened 2026-05-09
More feed info: last received date, publication date

Alkarex:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8670

image

image


rezad1393:
beautiful.

BTW, when user exports the feed (not the sqlite database just the opml) can all these data (error dates, user set update rate) be added to feeds as metadata that compatible apps can import and incompatible apps ignore?

Alkarex:
> beautiful.
> 
> BTW, when user exports the feed (not the sqlite database just the opml) can all these data (error dates, user set update rate) be added to feeds as metadata that compatible apps can import and incompatible apps ignore?

This information indeed comes along when exporting/importing through the SQLite method, but not through OPML. I am not sure it would make sense at all to include such local information in OPML.

rezad1393:
> > beautiful.
> > BTW, when user exports the feed (not the sqlite database just the opml) can all these data (error dates, user set update rate) be added to feeds as metadata that compatible apps can import and incompatible apps ignore?
> 
>  I am not sure it would make sense at all to include such local information in OPML.

"as metadata that compatible apps can import"

Alkarex:
> "as metadata that compatible apps can import"

Obviously, but even then. Let's wait for a concrete use-case to reevaluate

rezad1393:
> > "as metadata that compatible apps can import"
> 
> Obviously, but even then. Let's wait for a concrete use-case to reevaluate

of course.

I was just suggesting adding this data to opml as metadata that can be ignored (if it doesn't mess with regular import on other apps) but would help other apps know for example what a good update interval for a specific feed is .

kinda "if you build it they come".

PR #8804, opened 2026-05-10
fix(themes): replace #load_more loading-state GIF with a CSS spinner

polybjorn:
In its `.loading` state, the "Load more" button keeps its `.btn` frame on every theme, and on most themes the spinner inside is hard to see or invisible. Nord, Flat, Mapco, Ansum and Swage had per-theme overrides that papered over the spinner symptom with coloured fills instead of fixing the cause.

Replace the GIF with a CSS spinner in base-theme that:

- inherits `currentColor` from the button, so it reads correctly on every theme,
- drops the button frame and background fill in the loading state,
- respects `prefers-reduced-motion`,
- renders crisply at any DPI.

The per-theme `#load_more.loading` overrides become redundant and are removed. `loader.gif` is still used by a few other rules in base-theme, so the asset stays.

## Screenshots

| Before (Nord) | Before (Alt-Dark) | After (v2) |
|---|---|---|
| Untitled | Screenshot 2026-05-10 at 22 04 32 | Screenshot 2026-05-11 at 11 05 43 |

## Reproduction

1. Configuration > Reading: turn off "Load more articles at the bottom of the page".
2. DevTools > Network: set throttling to the slowest preset (e.g. GPRS, ~50 kbps) so the loading state is observable.
3. Click "Load more" and watch the loading state. The button frame stays visible on every theme; on most themes the spinner inside is also hard to see or invisible.

## Notes

- RTL stylesheets regenerated via `npm run rtlcss`.
- Per-theme `loader.gif` copies (`Nord/loader.gif`, `Dark/loader.gif`, etc.) have never been referenced: CSS relative URLs in `--frss-loading-image` resolve against `base-theme/frss.css` regardless of active theme. They were already orphaned before this change; cleanup of those copies is a separate change.
- The same spinner pattern would work for the remaining `loader.gif` consumers; after that the asset is fully removable.

### DIFF
--- a/p/themes/Ansum/_components.css
+++ b/p/themes/Ansum/_components.css
@@ -201,11 +201,6 @@
 	}
 }
 
-#load_more.loading,
-#load_more.loading:hover {
-	background: url("loader.gif") center center no-repeat #34495e;
-}
-
 /*=== Boxes */
 .box {
 	background: var(--white);
--- a/p/themes/Ansum/_components.rtl.css
+++ b/p/themes/Ansum/_components.rtl.css
@@ -201,11 +201,6 @@
 	}
 }
 
-#load_more.loading,
-#load_more.loading:hover {
-	background: url("loader.gif") center center no-repeat #34495e;
-}
-
 /*=== Boxes */
 .box {
 	background: var(--white);
--- a/p/themes/Flat/flat.css
+++ b/p/themes/Flat/flat.css
@@ -417,11 +417,6 @@ th {
 	color: #000;
 }
 
-#load_more.loading,
-#load_more.loading:hover {
-	background: url("loader.gif") center center no-repeat #34495e;
-}
-
 /*=== Boxes */
 .box {
 	border: 1px solid #ddd;
--- a/p/themes/Flat/flat.rtl.css
+++ b/p/themes/Flat/flat.rtl.css
@@ -417,11 +417,6 @@ th {
 	color: #000;
 }
 
-#load_more.loading,
-#load_more.loading:hover {
-	background: url("loader.gif") center center no-repeat #34495e;
-}
-
 /*=== Boxes */
 .box {
 	border: 1px solid #ddd;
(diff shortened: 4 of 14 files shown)


Alkarex:
Nice, good idea. Let me test it a bit before merging

Alkarex:
Please write if already tested with some specific browsers

polybjorn:
> Please write if already tested with some specific browsers

I've tested this on LibreWolf where every theme is now working as intended.

After opening the PR I tested with SeaMonkey where the CSS icons doesn't appear, but the loading function doesn't work at all there - after reloading the page, I get logged out for some reason. Have not troubleshooted the cause for that.. If I can assume changing the icons from .gif to CSS code isn't the reason I'm logged out, this is just as broken as it was before (at least on some themes).

Upon further investigation, there are 4 features that start breaking at various times for Safari/Chrome/Firefox, all around Sep-Jan 2021. There are multiple things I can do to move this floor further away if you want.

If this was to get merged, [I was planning on following up](https://github.com/polybjorn/FreshRSS/commit/60453a8e03d875061f086fa4afc03f0e31c85186) by getting rid of the .gifs entirely, so I don't mind spending extra time on this.

Alkarex:
That sounds good, and yes, fixing compatibility with some browsers would be much welcome (although it is ok not to have all details working on older engines, as long as it looks fine). More compatibility attention should be given to the default theme.
* https://github.com/FreshRSS/FreshRSS/issues/6776

polybjorn:
Reworked the spinner with border + rotate instead of conic-gradient + mask + aspect-ratio. 

Screenshot 2026-05-11 at 11 05 43

Oldest browser versions still supported: Firefox 16 (2012), SeaMonkey 2.13 (2012), Chrome 43 (2015), Safari 9 (2015), Edge 12 (2015). Below those, `@keyframes` and `animation` would need `-webkit-` fallbacks.

Verified now working in SeaMonkey 2.53.23. Couldn't test feed clicks or page reload there because login itself hits a separate cookie issue.

PR #8805, opened 2026-05-10
chore(themes): rename "Nord theme" to "Nord"

polybjorn:
Rename the Nord theme display name from "Nord theme" to "Nord". The upstream Nord palette project at https://www.nordtheme.com refers to itself consistently as "Nord" throughout its branding; the "theme" suffix is redundant. The rename also matches the other entries in the theme picker (Ansum, Dark, Mapco, Pafat, Swage), which all use a single short name.

The theme directory, file paths, and CSS class names are unchanged. Only the user-facing display name in the theme picker (and the corresponding rows in the user docs) is affected.

### DIFF
--- a/docs/en/users/05_Configuration.md
+++ b/docs/en/users/05_Configuration.md
@@ -58,7 +58,7 @@ There’s no accounting for tastes, which is why FreshRSS offers 13 official the
 | Dark pink | Miicat_47 | |
 | Flat design | Marien Fressinaud | |
 | Mapco | Thomas Guesnon  | |
-| Nord theme | joelchrono12 | |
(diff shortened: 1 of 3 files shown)


Alkarex:
Does it break the user preference for those who have selected that theme?

polybjorn:
> Does it break the user preference for those who have selected that theme?

The selection is stored as the theme's directory name (`p/themes/Nord/`), which this PR doesn't touch; `FreshRSS_Themes::load()` looks it up by that, and the `name` field in `metadata.json` is display-only. Renaming it can't invalidate an existing selection. Verified locally too: Nord stayed selected after the edit.

PR #5268, opened 2023-04-06
add-feeds.php

NobShen:
Add feeds from a file containing feeds URLs.

Closes #

Changes proposed in this pull request:

- Add a cli named add-feeds.php to import feeds from a text file containing feeds URLs

How to test the feature manually:

1. Create a text file containing feeds URLs, one line per feed i.e. feeds.txt.
3. From a command line, issue:  sudo ./cli/add-feeds.php --user username --filename feeds.txt.

Pull request checklist:

- [ ] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard)
- [x] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #8818, opened 2026-05-11
feat(import): accept .txt URL lists alongside OPML/JSON/ZIP

polybjorn:
Adds plain-text URL lists as a recognised import format. Files with a `.txt` extension are wrapped into an OPML document and handed to the existing OPML import path, so dedup, category assignment and feed limits work the same as they do for OPML. Available through both `cli/import-for-user.php` and the web import form.

Reworks #5268 along the lines suggested in that thread: no new CLI file, integrated into `import-for-user.php` and the web form, detected by extension.

## Format

One URL per line. Blank lines and `#` comments are skipped. A leading UTF-8 BOM is stripped. CRLF and LF both work. Lines that don't parse as a URL are logged and skipped; the rest are imported.

## Test plan

Tested on a live instance (SQLite, PHP 8.5):

- [x] CLI with a mixed file (valid URLs, blanks, comments, malformed URL, URL with a space): valid feeds inserted into the default category, invalid lines on stderr, exit 0
- [x] CLI rerun on the same file: no duplicate rows
- [x] CLI on an empty file: exit 0, no error
- [x] Web path (`$isCli = false`): warnings go to the user log file instead of stderr
- [x] Generated OPML round-trips through `LibOpml::parseString` with `&`, `"`, `<`, `>` in URLs intact
- [x] `composer run-script phpcs`, `phpstan` and `phpunit` (624 tests) pass

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -249,7 +249,8 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 	 * so it can be imported through the existing OPML pipeline.
 	 */
 	private static function txtToOpml(string $contents): string {
-		$contents = preg_replace('/^\xEF\xBB\xBF/', '', $contents) ?? $contents;
+		$utf8BOM = "\xEF\xBB\xBF";
+		$contents = preg_replace('/^' . $utf8BOM . '/', '', $contents) ?? $contents;
 		$outlines = '';
 		foreach (preg_split('/\R/', $contents) ?: [] as $line) {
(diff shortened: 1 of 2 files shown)


Alkarex:
I have made a couple of minor changes. Please double-check

polybjorn:
Pulled both commits and re-tested locally - BOM stripping, comments, blanks, invalid URLs, and `&`/`"`/`'` escaping all still produce valid XML. Had a look for similar nits in the rest of the PR and didn't spot any.

Issue #8183, opened 2025-11-02
Export to SQLite from Web UI


Alkarex:
It would be very nice to also be able to trigger back-ups from the Web UI. The challenge is to account for servers with limited resources and potentially many users. Help welcome:

* [ ] PR 1 (would be enough to have the settings in `config.php` for now and without UI):
    * [x] Setting to allow running automatic export to SQLite #8819
    * [x] Setting to limit the number of kept past SQLite exports #8819
    * [ ] Setting to chose the max DB size to allow SQLite export from Web UI
* [ ] Later PR
    * [ ] Mechanism to import an SQLite backup from Web UI (would first require some discussions)
* [ ] Later PR
    * [ ] Mechanism to ask for an SQLite export, which will be done at the next cron run, useful for larger DBs for which the Web UI could be problematic

_Originally posted by @Alkarex in https://github.com/FreshRSS/FreshRSS/issues/8169#issuecomment-3476079108_

PR #8819, opened 2026-05-11
feat(cli): automatic periodic SQLite export with retention

polybjorn:
First step of the multi-PR plan in #8183: settings and CLI for periodic on-server SQLite export. No Web UI yet.

Kept separate from `cli/db-backup.php` / `cli/db-restore.php`. Those are paired around a fixed `backup.sqlite` filename for one-shot migration; periodic exports want timestamped files and rotation, so they live under their own path `data/users/<user>/sqlite-backups/`.

## Settings

Added to `config.default.php`:

```php
'auto_sqlite_export' => [
    'enabled' => false,
    'retention' => 7,
],
```

`enabled` defaults to `false`; the CLI exits without writing when off. `retention` is how many past backups to keep per user; older ones are deleted after each successful export.

The third setting from #8183 (max DB size for a Web UI trigger) is deferred to the PR that adds the Web UI, since it has no consumer here.

## Usage

```sh
./cli/export-sqlite-auto.php
# Writes: data/users/<username>/sqlite-backups/<YYYYMMDDTHHMMSSZ>.sqlite
```

Timestamps are UTC, basic ISO 8601 format, filename-safe and lexically sortable. Meant to be scheduled by the admin (cron, systemd timer). Supports `-q` / `--quiet` like the other CLI scripts.

## Test plan

- `make test-all` passes locally.
- With `enabled => false`: CLI exits 0, writes nothing.
- With `enabled => true`: each user gets a `.sqlite` file under `sqlite-backups/` containing all FreshRSS tables.
- After 5 runs with `retention => 3`: 3 newest files remain per user.
- `--quiet` suppresses non-error output.

### DIFF
--- a/app/Models/SystemConfiguration.php
+++ b/app/Models/SystemConfiguration.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
  * @property bool $api_enabled
  * @property string $archiving
  * @property 'form'|'http_auth'|'none' $auth_type
+ * @property array{enabled:bool,retention:int} $auto_sqlite_export
  * @property-read bool $reauth_required
  * @property-read int $reauth_time
  * @property-read string $auto_update_url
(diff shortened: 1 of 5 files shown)


Alkarex:
Good idea 👍🏻

PR #8787, opened 2026-05-04
docs: clean up APIs & native apps comparison table

polybjorn:
Cleans up the "APIs & native apps" comparison table and adds clients per open requests.

The diff is noisy: rows re-sorted alphabetically, table reformatted, symbols normalized throughout. 

**Easiest review**: [this gist](https://gist.github.com/polybjorn/284527762bcd99ad0b15fb456663ca1a) has the rendered table plus per-change rationale. 
**For the README in context**: see [the file on this branch](https://github.com/polybjorn/FreshRSS/blob/docs/cleanup-api-clients-table/README.md).

## Collapsibility (open for debate)

The wrap cuts scroll length on a 13-column 20-row table; one click expands it. Counter-argument: the table was added in #2942 for newcomer discoverability. Collapsing it works against that.

`.markdownlint.json` previously forbade inline HTML; this PR adds `details` and `summary` to `allowed_elements` so the wrap can land (both are GitHub-flavored markdown features). If the wrap goes, the lint change goes with it.

## New rows

- **[FeedFlow](https://github.com/prof18/feed-flow)**: Apache-2.0, GReader, cross-platform (Android, iOS, Linux, macOS, Windows). Closes #7301.
- **[FluentFlame Reader](https://github.com/FluentFlame/fluentflame-reader)**: BSD-3-Clause Electron fork of Fluent Reader, marked `WIP` (no releases yet). Closes #8608.

## Status downgrades

The `Status` column maps from upstream as `✔️✔️` -> `Active`, single `✔️` -> `Maintenance`, `Work in progress` -> `WIP`. Two apps were set to Maintenance instead of the default Active:

- **Readrops**: last release `v2.1.1` on 2025-07-07, last commit on default branch 2025-07-20. ~10 months idle.
- **Reeder Classic**: [reederapp.com/classic](https://reederapp.com/classic/) calls it the legacy product: *"this version will remain available as Reeder Classic for those who prefer a more traditional RSS reader experience."*

## Coordination

Open PR #8745 wants to bump Your News Sync efficiency from ⭐ to ⭐⭐⭐. Current value matches the live README; only matters if both land around the same time.

PR #8821, opened 2026-05-12
docs: normalize APIs & native apps table glyphs to monochrome

polybjorn:
Replace variation-selector emoji in the APIs & native apps table with monochrome glyphs that inherit text color in both light and dark mode.

- `✔️` -> `✓`
- `➖` -> `–`
- `❔` -> `?`
- `💲` -> `$`

Stars are kept colored. No data changes, no row reordering, no column renames.

[Related gist](https://gist.github.com/polybjorn/519b54a270f83a3d626944dffa07e287) with examples.


Alkarex:
By the way, when we started using those emojis, they were more colourful.
Now, even before this PR, they are more commonly represented in grey, which I think makes it more difficult to quickly spot differences at a quick glance.

Some other variants do seem to be more commonly coloured, such as:
✅  ❌ ❔

Thoughts?

polybjorn:
The end-goal I'm trying to work at here is a curated [selection of recommended apps](https://github.com/FreshRSS/FreshRSS/pull/8787#issuecomment-4373694058) in README, with unabridged table(s) in docs.

My problem with emojis is that they are not consistent (⭐ ✅ ❌ ❔ $).
- Check (Yes) is colorless in green border, while Cross (No) is the opposite
- Some symbols have no color
- They are rendered differently depending on OS/browser font

An alternative to this is using SVG icons. Personally I'm biased toward [this hand-drawn style](https://icons8.com/icons/set/cross--style-plasticine), and could make a consistent collection of icons if you want to go this route. Some of the icons need to be edited to make a consistent collection. Not really sure if it's relevant I have purchased commercial license for all related icons if used here? Or if license is relevant. Would look horrible in code, but still an improvement visually.

Imagining the future trimmed-down table of recommended apps however, I think it would look clean with simple monochrome symbols as proposed by this PR. I agree with the principle that green/red check/cross next to star rating would look appealing.

I planned on trying to migrate the sync column from ⭐ / ⭐⭐ / ⭐⭐⭐ to ★☆☆ / ★★☆ / ★★★ as well for consistent width, and to limit the focus this column take over all other features, especially if no other symbols are colored.

Alkarex:
Would you mind sharing a couple of screenshots with this PR compared with ✅ ❌ ❔, to better decide?

Frenzie:
View file in the top right of the diff does the trick.

For me it's before:

image

After:

image

I think I somewhat prefer the before but it's a bit skewed by not going with an en-dash (–) or some such but with a visually overwhelming cross instead. I think for that reason it's much easier to see which program supports offline for example in the old one.

polybjorn:
[Here's a gist](https://gist.github.com/polybjorn/519b54a270f83a3d626944dffa07e287) with 4 alternatives in one.

I agree original looks much better in light mode - dark mode is a different story though.

For the emoji version, a drawback looking at it now, is that it's hard to tell where the links are. But I guess that's not too different from the original..

I guess the links kind of break the whole monochrome argument, but the links are much more obvious.

Frenzie:
@polybjorn Could you also add non-emoji with – or — or similar? I don't like that the less interesting information draws more attention to itself.

polybjorn:
> Could you also add non-emoji with – or — or similar?

@Frenzie Now the [gist](https://gist.github.com/polybjorn/519b54a270f83a3d626944dffa07e287) has 6 alternatives, including one for monochrome No = "-" and one for monochrome Unknown = blank.

Frenzie:
I think the en-dash looks much better, but I prefer even the regular dash over the giant cross.

polybjorn:
> the en-dash looks much better

Replaced regular dash (-) with en-dash (–) now.

Alkarex:
I do not like the *Colored circles* option (in particular because I do not like relying *only* on colour) but I do not really mind between any of the others. The *en-dash* option would be fine, although I would have liked to keep a bit of colour

polybjorn:
> I do not like relying _only_ on colour .. although I would have liked to keep a bit of colour

[Added 2 more options](https://gist.github.com/polybjorn/519b54a270f83a3d626944dffa07e287) using draft PNG icons. Check sign would have to be edited so it's not doubled, question mark probably edited blue/white/orange, and there isn't a good dollar symbol available without proper editing. Also the PNGs won't be vertically centered, which is noticeable next to the stars, which have to remain emoji if we want more than 1 symbol per cell.

I guess I could create 3 star icons to keep horizontal rating.

Going this route, all icons would be SVG in the end.

Alkarex:
I would rather avoid using images if possible. Is the *en-dash option* good enough for now?

Alkarex:
Could you please do the same for https://github.com/FreshRSS/FreshRSS/blob/edge/README.fr.md ?

polybjorn:
Done 👍

FR edition is missing the two new apps, and has one inconsistency on API column compared to EN, so I will create a separate PR for that.

PR #8824, opened 2026-05-12
refactor(themes): convert article-row inner layout to CSS Grid

polybjorn:
Replace absolute positioning inside `.titleAuthorSummaryDate` with CSS Grid. Title, date, and summary use `grid-area` instead of `position: absolute` + `margin-top: 2.25rem`.

Drops magic numbers that had to stay in sync:
- Three uses of `155px` (date width, title min, title padding) all describing the same column
- `2.25rem` margin-top on summary that had to clear the absolute title
- `90%` summary max-width to keep it off the date

Truncated-title hover-extend now uses `grid-column: 1 / -1` instead of `padding-right: 0.5rem`.

Depends on #8823: grid inside `display: table` mis-sizes in older Gecko, causing summary to misbehave. With the flex base from that PR, the grid sizes correctly in SeaMonkey 2.53.

## Test plan
- `make test-all` passes
- Verified in LibreWolf and SeaMonkey 2.53 with Nord and Origine themes
- Wide, narrow, and mobile viewports
- Toggle combinations: date on/off, summary on/off, hover-extend on truncated titles

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1506,7 +1506,7 @@ a.website:hover .favicon {
 }
 
 .flux:not(.current):hover .flux_header.has-date .item .title {
-	padding-right: 0.5rem;
+	grid-column: 1 / -1;
 	z-index: 2;
 }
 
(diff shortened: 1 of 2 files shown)


Alkarex:
Conflict handling welcome

polybjorn:
Conflicts cleared 👍

PR #8825, opened 2026-05-12
feat(favicons): use PHP 8.4+ native HTML5 parser

polybjorn:
For #6831.

Switch `searchFavicon()` to `\Dom\HTMLDocument::createFromString()` on PHP 8.4+, with the existing `DOMDocument` path as the fallback for older versions.

The new parser places elements in the XHTML namespace by default, so `//link` and `//base` would match nothing. `Dom\HTML_NO_DEFAULT_NS` opts out of that and keeps the existing XPath queries working.

The favicon-fetch loop moved into `tryFaviconLinks()`, shared between both branches.

PHPStan hides symbols newer than the `phpVersion` min (8.1), so the `\Dom\*` classes are unknown to it regardless of stubs. Rather than ignore rules, `tests/stubs/php84-dom.php` declares the symbols this file uses (via `scanFiles`, excluded from analysis), so the new branch is type-checked like the rest of the code. `Dom\Element::getAttribute()` returns `?string` (unlike `DOMElement`), hence the `?? ''`. The stub can be deleted along with the legacy branch once the minimum PHP version reaches 8.4.

Scope is `lib/favicons.php` only. The other `DOMDocument::loadHTML` sites in the tree have site-specific complications (`saveHTML($node)` round-trips, `recover`/`strictErrorChecking` flags, the `loadXML` branch in `Feed.php` needing `\Dom\XMLDocument`) and are out of scope here.

## Test plan
- [x] `php-lint`, `phpcs`, `phpstan` (level 10), `phpunit` (652/652) all pass
- [x] `phpstan-next` (level max): no errors in changed files
- [x] PHP 8.1, 8.2, 8.3 (Docker): legacy branch extracts identical `<link>` and `<base>` hrefs vs new branch on the same fixture
- [x] PHP 8.5 (live): `searchFavicon()` returns non-empty favicons for news.ycombinator.com, en.wikipedia.org, github.com


polybjorn:
I've replaced the PHPStan ignores with a small stub file (`tests/stubs/php84-dom.php`) that declares the `Dom\*` symbols via `scanFiles`, so the new branch gets properly type-checked. I was wrong that the ignores would clear up on their own: PHPStan hides symbols newer than the `phpVersion` min, so they would have stayed around forever. While checking the stub against PHP 8.5 I also noticed `Dom\Element::getAttribute()` returns `?string` (unlike `DOMElement`), so those call sites now fall back to `''`. Re-ran `composer test` and a live favicon check on PHP 8.5.

PR #8833, opened 2026-05-14
Keep mobile article navigation at the bottom

### LINKED ISSUE #8829: [Bug] bottom bar detaches from bottom, wanders into middle of screen
### Describe the bug

The bottom bar with the left/right buttons sometimes detaches from the bottom and wanders into the center. 

I use freshrss on an iPhone and added it to the Home Screen as described here https://support.apple.com/guide/iphone/iph42ab2f3a7/ios

Here is a video since GitHub does not accept the direct upload  https://youtube.com/shorts/CQNMw8pBD2g

### To Reproduce

1. Add freshrss to home screen
2. use freshrss
3. click left / right buttons on bottom bar. 

### Expected behavior

The bar stays where it is. 

### FreshRSS version

1.28.1

### System information

Installed through truenas https://apps.truenas.com/catalog/freshrss/

- Browser: mobile safari 26.5

### Additional context

_No response_

mzl2233:
This keeps the mobile article navigation fixed to the bottom of the viewport by clearing any inherited top offset in the small-screen rule. It updates both LTR and RTL base theme CSS so the previous/next bar remains anchored while browsing on mobile Safari.

Verification: targeted CSS assertion for both base theme files; `git diff --check`.

Closes #8829

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2869,6 +2869,7 @@ html.slider-active {
 
 	#nav_entries {
 		display: flex !important;
+		top: auto;
 		width: 100%;
 	}
 
--- a/p/themes/base-theme/frss.rtl.css
+++ b/p/themes/base-theme/frss.rtl.css
@@ -2869,6 +2869,7 @@ html.slider-active {
 
 	#nav_entries {
 		display: flex !important;
+		top: auto;
 		width: 100%;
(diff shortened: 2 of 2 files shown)


Frenzie:
So this is an issue specifically for iOS Safari?

Alkarex:
Thanks @mzl2233

PR #8838, opened 2026-05-15
revert(themes): hide feed name at narrow viewports

### LINKED ISSUE #8814: Regression in UI: feed title is too large in mobile view
A recent commit has made the mobile view poorer:

Image

Probably for @polybjorn if you can please take a look.

polybjorn:
Reverts #8801 (keeping its incidental blank-line cleanups) and #8815, restoring icon-only feed column on mobile (#8814).

I filed the original PR thinking the icon-only behavior was an inconsistency bug versus the configurable `topline_website` setting, not realizing it was a deliberate mobile choice. The `clamp()` cap helped, but the column still takes space the title could use, and on phones icon-only is the better tradeoff.

A distinct narrow-viewport setting for `topline_website` was raised as a future improvement.

## Test plan

- [x] Narrow viewport, all three `topline_website` modes: name hidden, favicon shown, title fills the row
- [x] Wide viewport: setting respected, unchanged from edge
- [x] `has-thumbnail.has-summary` grid at narrow: name reappears in its dedicated grid area
- [x] `npm run stylelint` + `npm run rtlcss` clean

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2833,12 +2833,6 @@ html.slider-active {
 		width: 40px;
 	}
 
-	/** Cap website column when feed name is shown, to leave room for the title */
-	.flux_header.websitename .item.website,
-	.flux_header.websitefull .item.website {
-		width: clamp(120px, 18vw, 150px);
-	}
-
 	.flux:not(.current):hover .item.title {
 		position: relative;
 		width: auto;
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
I knew something looked off but I figured it was just my local CSS not playing nice with some more innocuous change and I hadn't investigated yet.

PR #8843, opened 2026-05-15
fix: rewrite img src from srcset when src is a data: URI placeholder

### LINKED ISSUE #5016: [BUG] Fetching article by CSS-selectors: <picture> element is wrong stripped

### LINKED ISSUE #8699: [Bug] Image src not loading from srcset with responsive-img and src stripped by data-sanitized-class due to transparent gif image placeholder
## Issue

I'm having issues retrieving **images** from articles using a placeholder base64 for "hidden image" (see image/gif in `src`), and the actual image(s) is within a `srcset` (with different sizes/qualities available).

The image is part of `responsive-img` class. I don't know if this is the "default way" of working with `reponsive-img` class(es)?

Within FreshRSS web viewer, if I look in the browser "elements" after fetching an article (see example below), it seems to get "caught" by the `data-sanitized-class` somehow:

```html
<div data-sanitized-class="responsive-img F169"><img
    src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7" alt="Tips-bilde av lysfenomen"
    title="Foto: Tipsar"></div>
```

## Example

Example from an article [[source](https://www.nrk.no/mr/lysfenomen-pa-himmelen-observert-fleire-stadar-i-landet-1.17850489)](https://www.nrk.no/mr/lysfenomen-pa-himmelen-observert-fleire-stadar-i-landet-1.17850489) (Norwegian Broadcasting Corporation). Note that this is just as an example, but most images from nrk.no is using `srcset` for the `reponsive-img` and the "hidden gif image" as `src`.

```html
<figure class="image widget rich">
  <div class="responsive-img F169"><img sizes="(min-width:1180px) 767px, (min-width:720px) 67vw, 100vw"
      srcset="https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWow_NoqD_NA7FveKgLdT4dyLQ.jpg 80w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowDIukd5Br-y3eKgLdT4dyLQ.jpg 160w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowuoB3OQjTpifeKgLdT4dyLQ.jpg 350w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowPQYbPh9VutLeKgLdT4dyLQ.jpg 450w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowBdZXm6zDbxbeKgLdT4dyLQ.jpg 650w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWoweJpKUQhX-5_eKgLdT4dyLQ.jpg 1000w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowJu73I85lPhneKgLdT4dyLQ.jpg 1200w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowb_xDZEVCkdPeKgLdT4dyLQ.jpg 1600w, https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowYK12pxH4h9veKgLdT4dyLQ.jpg 2000w"
      src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7" loading="eager"
      alt="Tips-bilde av lysfenomen" title="Foto: Tipsar"></div>
  <figcaption>
    <p class="">LYS: Dette bilde blei tatt i Kristiansund natt til fredag. </p>
    <small>Foto: Tipsar</small>
  </figcaption>
</figure>

```

Feed source: https://www.nrk.no/toppsaker.rss

## What I want

Ideally, I'd like to extract the larget image from the srcset (best image quality) and use that as image src somehow. I guess people (possibly) could have different preferences here. From the `srcset` above, that would be image https://gfx.nrk.no/qjzJQnvlMt-VoCXf0miWowYK12pxH4h9veKgLdT4dyLQ.jpg (which is 2000w, and usually the largest quality from nrk.no).

## Related

I also use Obsidian and Obsidian Web Clipper (Chrome Extension) to download some articles in Markdown-format. I've written a custom filter using regex (for Obsidian Clipper template), which I'm using on (site) nrk.no, which basically extracts the "last image url" from the `srcset`, and uses that as `src` value:

```text
replace:"/srcset=.* ([^ ]+) \d+w .* /g":"src="
```

For reference, my full "selectorHtml" (in Obsidian Template-style)

```text
{{selectorHtml:#innhold|remove_html:(".lp_related")|replace:"/srcset=.* ([^ ]+) \d+w .* /g":"src="|remove_html:(".authors,.article-dateline,.link__container")|markdown}}
```

I know the above is not directly translated into usable code, but hopefully gives an idea of what I'd like to achieve.

I also noticed RSS-Bridge has a `parseSrcsetLargestImageUrl` function (https://github.com/RSS-Bridge/rss-bridge/blob/952857e4b62b632ac2d8a682b271706813448d4e/lib/html.php#L274); maybe we could have something similar, to basically extract the "largest image", and use that as "img src"?

**I'm no developer**, so please correct me if I misunderstood something here, or if there is a better way of retrieving the images from the srcset. I'm unsure on "how" this is best solved, e.g. within data-sanitized, or by e.g. using a custom JS for this specific feed (if possible)? I did a quick grep for `srcset` in the FreshRSS codebase, but did not find anything, other than something from a changelog back in 2016, which does not seem to exist anymore.

Some related issues I've found.
https://github.com/FreshRSS/FreshRSS/discussions/6563
https://github.com/FreshRSS/FreshRSS/issues/1077
https://github.com/FreshRSS/FreshRSS/issues/3823
https://github.com/FreshRSS/FreshRSS/issues/5016
https://github.com/FreshRSS/FreshRSS/issues/6875
https://github.com/FreshRSS/FreshRSS/issues/8317
https://github.com/FreshRSS/FreshRSS/pull/1086
https://github.com/FreshRSS/FreshRSS/pull/5154

### To Reproduce

1. Add RSS feed https://www.nrk.no/toppsaker.rss
2. Load any article
3. All `responsive-img` is caught by `data-sanitized-class`, hence no article images (within body) is loading

### Expected behavior

Images to load successfully, by using the srcset

### FreshRSS version

1.28.1

### System information

FreshRSS version
1.28.1
PHP version
8.4.16
cURL version
8.14.1
OpenSSL/3.5.4
Database
sqlite
3.46.1
Server software
Apache/2.4.66 (Debian)
Linux 6.12.55 #1-NixOS SMP PREEMPT_DYNAMIC Thu Oct 23 14:20:48 UTC 2025 x86_64
Browser
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36

### Additional context

_No response_

polybjorn:
Fixes #8699, fixes #5016.

Lazy-load patterns put a transparent gif in `src` and the real images in `srcset`. FreshRSS stripped `srcset` because it wasn't in the allowed `` (or `<source>`) attributes, so only the placeholder rendered.

## Problem

For markup like

    

just allowing `srcset` through isn't enough: SimplePie's `set_url_replacements()` only handles single-URL attributes, so relative URLs inside `srcset` would still leak.

## Fix

A `\SimplePie\Sanitize` subclass overrides `enforce_allowed_html_nodes` and runs before the parent strips disallowed attributes. For each `` and `<source srcset>` it absolutises every URL against the document base. For `` only, if `src` is empty or a recognised placeholder (the `R0lGODlh` 1x1 transparent GIF or any other `data:` URI under 128 chars), it writes the smallest `Nw` entry as fallback. `src` only loads when `srcset` can't be honoured, so smallest minimises bandwidth and is never larger than what the browser would have picked.

`srcset` and `sizes` are added to the allowed-attribute list on `` and `<source>`. Mixed `Nw`+density srcsets keep both kinds; bare entries (no descriptor) stay bare.

This runs in every path that goes through SimplePie's sanitiser, so it covers both web scraping and SimplePie's own item parsing of `<content:encoded>`.

## Test plan

- [x] 13 integration tests in `tests/app/Models/SimplePieCustomTest.php`.
- [x] `make test-all`: PHPUnit (639), PHPCS, PHPStan level 10, ESLint, Stylelint, markdownlint all clean.
- [x] Live `nrk.no/toppsaker.rss` with web-scraping: 33/33 scraped entries get absolute `srcset`, smallest URL in `src`, no placeholders.
- [x] Synthetic feeds for `<content:encoded>` (lazy-load and WordPress-style) and `<picture><source>` siblings through SimplePie's native parse.


app/Models/SimplePieCustom.php
```diff
@@ -281,13 +281,73 @@ public function __construct(array $attributes = [], array $curl_options = []) {
...
+	 * density-only descriptors (`1x`, `2x`) are not rewritten since such
+	 * images normally ship a working `src` already.
+	 * `srcset` itself is still stripped by the sanitizer afterwards.
+	 * @internal exposed for unit testing only.
+	 */
+	public static function rewriteImgSrcFromSrcset(string $html): string {
+		if (stripos($html, 'srcset') === false) {
+			return $html;
+		}
+		return preg_replace_callback(
```
Alkarex: It would be interesting to investigate whether this could be done in SimplePie when the DOM is parsed, instead of a regex like here, and without having to reparse the DOM

Frenzie:
> Adding srcset to the allowed attributes does not help on its own

Maybe not, but I'd say doing so is extremely desirable.

> Pick the largest entry from srcset by Nw width descriptor and write it into src. 

Fwiw, generally it's said that the fallback or default should be small or medium.

polybjorn:
The DOM subclass covers `<content:encoded>` srcset as well as the web-scraping path, verified via a synthetic feed through SimplePie's native `get_content()`. `srcset` and `sizes` are now in the allowed attributes on `` and `<source>` so the browser handles selection; the fallback `src` is the smallest `Nw` entry, since `src` only loads when `srcset` can't be honoured and should stay conservative on bandwidth.

Same-area audit additions: `<source srcset>` for parity with ``, density entries preserved when mixed with `Nw`, bare-descriptor entries kept as-is, explicit match for the `R0lGODlh` 1x1 GIF placeholder alongside the threshold.

Verified end-to-end: web-scraping against `nrk.no/toppsaker.rss`, native SimplePie parsing through synthetic feeds for `<content:encoded>` and `<picture>/<source>`. `make test-all` clean.

Alkarex:
Sorry for the delay. I wanted to find an approach with a better performance, in particular avoiding another XPath.
I have made a proposition in that direction.
Feedback welcome @polybjorn , @Inverle

lib/simplepie/simplepie/src/Sanitize.php
```diff
@@ -753,6 +754,10 @@ protected function enforce_allowed_html_nodes(\DOMNode $element, bool $allow_dat
                     $element->removeAttributeNode($element->attributes[$i]);
                 }
             }
+
+            if (in_array($tag, ['img', 'source'], true) && $element->hasAttribute('srcset')) {
+                $this->rewriteImgSrcset($element);
```
Alkarex: We may want to rename the parent function though, since its role is now a bit more general

polybjorn: Renamed to `sanitize_html_nodes`. Open for even better alternatives

polybjorn:
The single DOM walk is nicer, thanks. Moving it after `replace_urls` also means `$this->base` is set by the time the rewrite runs, so the `sanitize()` override I had is no longer needed.

While retesting I found three problems in the srcset handling and pushed fixes with tests:

- Splitting on every comma breaks URLs that contain unencoded commas (Cloudinary and similar CDNs). Now parsed per the WHATWG algorithm.
- The rewrite runs after `replace_urls`, so a `javascript:` entry in srcset survived and could even end up in `src`. Entries with disallowed schemes are now dropped.
- When srcset only has density descriptors and `src` is a placeholder, clients that only read `src` still got the placeholder. The lowest-density entry is now used as fallback (width entries still take priority).

The image caching handler only rewrites `src`, so srcset URLs bypass it. Off by default and feels like a separate change if it matters.

Alkarex:
If you can, please make a PR targetting our https://github.com/FreshRSS/simplepie containing the changes from this PR for the file lib/simplepie/simplepie/src/Sanitize.php, and potentially tests/app/Models/SimplePieCustomTest.php or a part of it

polybjorn:
Opened FreshRSS/simplepie#87 with the `Sanitize.php` change and unit tests against `Sanitize` directly. The copy here is provisional and identical to it, so that file does not need reviewing twice.

Marking this a draft until #87 merges. After that it becomes the re-vendored file plus the FreshRSS-only bits: the `srcset`/`sizes` allowlist and a bump of the pinned SimplePie commit in `lib/composer.json`.

One behaviour change since your last look: an empty `src` was resolved against the document base before the placeholder check ran, leaving `src` pointing at the article page. Both PRs now run the allowed-node pass ahead of `replace_urls()`, with a test covering it.

PR #8844, opened 2026-05-15
New option to hide badges showing number of unread articles

Alkarex:
Inspired by @Dan-Q's  *Reducing Phantom Obligation in FreshRSS* https://danq.me/2026/02/06/phantom-obligation-vs-freshrss/

image

Results in:

image

Can be overridden at category and feed levels.


Frenzie:
Funny, I don't experience it that way at all. There's no obligation because I don't have to reply to it or anything. I definitely dislike that about newsletters though.

image

app/i18n/es/conf.php
```diff
@@ -55,6 +55,13 @@
 		),
 		'show_nav_buttons' => 'Mostrar los botones de navegación',
 		'show_title_unread' => 'Mostrar el número de artículos no leídos en el título',
+		'show_unread_count' => array(
+			'_' => 'Show unread counts in sidebar',	// TODO
+			'all' => 'For all categories and feeds',	// TODO
+			'important' => 'For Important feeds only',	// TODO
```
Frenzie:
```suggestion
			'important' => 'For important feeds only',	// TODO
```

Frenzie (approved):

Dan-Q:
Woah; this is super cool!

app/i18n/es/conf.php
Dan-Q: Same applies to the other files too.

Issue #8834, opened 2026-05-14
API Password in plain text


RickFr410:
Hello,
Today, I discover that gReader API calls transmit the API password in clear text. I don't know why it has been made as it is, but this can cause the password to be leaked in logs (I found it in my reverse proxy logs, it can also be the case in some security equipements). It has already been issued in https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-8vvv-jxg6-8578. 
Otherwise, there is no explicit reference about these risks in the documentation or in the API configuration web page.

Alkarex:
An improvement is to make sure that your client transmits the credentials as POST and not GET.
FreshRSS supports both, but it is up to the client to use the safer variant.
For instance, a client such as [Capy Reader](https://github.com/jocmp/capyreader) does it correctly.

RickFr410:
the client can of course send the auth datas by post method, but they are always transmited in clear text. The client cannot change that. I think it lack informations about that when we enable the API.

P.S. : reported to the app developer to add the POST method.

Frenzie:
You are strongly advised to set it up with HTTPS. POST vs GET is just about it potentially showing up in logs.

RickFr410:
It is already the case for me, but all hosters doesn't offer HTTPS. There is more secure methods for API authentication.

Frenzie:
I'm not sure if I agree exactly since a separate API key for API actions is the same way every platform I know does it, but I'll toss in a quick PR to make things clearer.

Inverle:
> P.S. : reported to the app developer to add the POST method.

Linking the issue here: https://github.com/seazon/FeedMe/issues/168

Alkarex:
We could consider adding an option to disallow GET parameters for that API method.
That would not prevent clients from trying, though.

Inverle:
It is now fixed in FeedMe 4.10.0: https://github.com/seazon/FeedMe/releases/tag/4.10.0#:~:text=%5Boptimize%5D%20Enhanced%20security%20for%20sending%20FreshRSS%20authentication%20credentials%2E%20%23168

PR #8845, opened 2026-05-15
Add a warning message to the API password section, add a log warning when a client uses GET, use POST in sample curl API usage

Frenzie:
References #8834.

Looks like this:
image

Alternative phrasing suggestions welcome.

PR #8855, opened 2026-05-17
Change examples in API docs to authenticate via POST instead of GET

Inverle:
Related to https://github.com/FreshRSS/FreshRSS/issues/8834
We should avoid bad examples like this, someone could copy the example blindly and send the credentials in an unsafe way.

### DIFF
--- a/docs/fr/users/06_Mobile_access.md
+++ b/docs/fr/users/06_Mobile_access.md
@@ -76,8 +76,8 @@ possibilité.
 Exemples de requêtes simples :
 
 ```sh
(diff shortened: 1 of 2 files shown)


docs/fr/users/06_Mobile_access.md
Inverle, line 1: This page doesn't appear on the sidebar or in search.
Missed in https://github.com/FreshRSS/FreshRSS/pull/8766 (surface hidden pages)

Inverle:
I didn't notice it was already done in https://github.com/FreshRSS/FreshRSS/pull/8845

PR #8870, opened 2026-05-20
Always jump article to top when header is offscreen, also when 'Stick the article to the top when opened' is disabled

### LINKED ISSUE #4069: [Feature] [Question] UI issues: strange navigation on article click & forced autoload

Frenzie:
Closes #4069.

1. Disable 'Stick the article to the top when opened'
2. Open a very long article
3. No scrolling will occur if you close it again while the header is still onscreen
4. If the header left the screen, opening another article will scroll to it to top 
5. Same for collapsing the current article, but then it'll put the current header at the top

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

---

@rg-software please test to see if this behaves more like how you expect.


polybjorn:
Testing while turning off:
1. Show articles unfolded by default
2. Stick the article to the top when opened
3. Clicking outside of article text area closes the article

Going back and forth from edge and this PR, and I do see the genuine improvement. If I understand it correctly, it only applies to articles below the currently open article, whereas opening the articles above doesn't have this issue?

I hope it answers the original issue.

Frenzie:
Articles above (location 1) have a stable position on the page. The issue is that everything below the open article's header (location 2) will move up as the open article is closed.

image

That needn't mean there aren't any issues with articles above, but that's why I asked for feedback.

rg-software:
Sorry it took me a while to test it. I confirm it works as described, but effectively "4. If the header left the screen, opening another article will scroll to it to top" means you DO stick article to the top, so this behavior feels weird given that "stick to the top" checkbox is actually unchecked. What is the effect of this checkbox then?

Like I mentioned before, my mental model of the correct behavior for the unchecked "stick to the top" box is similar to HTML spoiler tag. click to unfold, click to fold again, no scroll happens. Unfolding one spoiler does not fold other spoilers back, however. Personally, I am fine with both fold or no fold -- neither option feels broken.

Frenzie:
> If the header left the screen, opening another article will scroll to it to top" means you DO stick article to the top, so this behavior feels weird given that "stick to the top" checkbox is actually unchecked. What is the effect of this checkbox then?

The fact is that the intended destination is offscreen, so you need to scroll to the destination to see it. I understand you never wish to auto-close anything, but whether you close the article or not doesn't affect the behavior in this PR. Unless you mean there's an edge case in which it does; in that case there might be an issue that needs to be fixed.

The effect of the checkbox is that it applies when the header of the destination article is onscreen.

With the setting enabled, opening any of these articles will put them to the top of the screen. With the setting disabled, they'll open and otherwise nothing happens.
<kbd>image</kbd>

But when all you see on the screen is this, opening the previous or next article necessitates scrolling to it:
<kbd>image</kbd>

And by corollary, *closing* the article necessitates scrolling to its header lest you only see articles from much further down the list than where you were.

rg-software:
Let me confirm. Consider your screenshot. 

1) Suppose I am clicking on "Pizza met voordeeg..." -- I see it open under mouse cursor. So far so good, this is exactly as expected.

2) Suppose that in the same case there is some long open article above the screenshot. It is open, but I don't see it open -- I am far below, so it is not relevant, I don't care about it. Now I am clicking on "Pizza met voordeeg...", and I see the whole screen scrolling up, which is very unexpected.

IMO, from the user perspective, there is no difference between (1) and (2). Effectively, to avoid jumps, the user has to close open articles manually.

Frenzie:
Thanks, updated. I also noticed a remaining issue where next would open articles offscreen at the bottom. I've currently updated it to scroll up just enough for it not to be hidden, which feels a bit odd because then you have to scroll manually to actually read it but it seemed more in line with the "What is the effect of this checkbox then?" objection that way.

rg-software:
Well, as far as I can tell, it works perfectly! :) Takes a bit to get used to this collapse/uncollapse, but that's it. The only case where I see the screen jumps is when it is expected to do, such as:

- open the next article via a keyboard shortcut if the next article is off the screen;
- open the next article when lost screen space cannot be compensated (e.g., the first article in the list is open, then I click on the next article, but the first article collapses and it is not possible to scroll the screen down).

I'll try this config for a while, but congratulations, I thought making this kind of behavior is quite difficult.

PR #8899, opened 2026-05-31
Update security policy due to many duplicated AI reports

Alkarex:
Inspiration from <https://lkml.org/lkml/2026/5/17/896>


Frenzie:
Corrected in #8900.

PR #8903, opened 2026-06-02
Implement setting for visibility on the category level

### LINKED ISSUE #4913: [Feature] Visibility options for the category
I'd like to be able to not display a category in the main stream.

The same options that a feed has for "visibility" (well, the first and the last: show in main stream / do not show)
chrome_2022-12-02_18-18-41

**Additional context**
I've just used the YT Takeout extension and it created lots of feeds in a category, and all articles are stacking on top of my main stream.

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/4913
Before only feeds had the visibility option, now both categories and feeds have the option.

Currently tested only on SQLite


app/Models/EntryDAO.php
```diff
@@ -1624,11 +1629,11 @@ private function sqlListWhere(string $type = 'a', int $id = 0, int $state = Fres
 
 		return [array_merge($values, $searchValues), 'SELECT '
 			. ($type === 'T' && $sort !== 'rand' ? 'DISTINCT ' : '')
-			. 'e.id'
+			. 'e.id, CASE WHEN f.priority = 10 THEN c.priority ELSE f.priority END AS priority2'
 			. ($type === 'T' && !in_array($sort, ['id', 'rand'], true) ? ', ' . $orderBy : '')	// SELECT DISTINCT, ORDER BY expressions must appear in SELECT
 			. ' FROM `_entry` e ' . $useEntryIndex
 			. 'INNER JOIN `_feed` f ON f.id = e.id_feed '
-			. ($sort === 'c.name' ? 'INNER JOIN `_category` c ON c.id = f.category ' : '')
+			. 'INNER JOIN `_category` c ON c.id = f.category '
```
Alkarex, line 1643: This is a function which has had a lot of performance tuning, on the different databases, and with different situations including high numbers of articles / feeds. Any change here requires careful performance regression assessment.
At a first glance, adding a join in all situations will likely be expensive

Alkarex:
I have not read all the code yet. If not already the case, I think it is sufficient for feeds to be able to be *less* visible than their category, but not *more* visible than their category

app/Models/EntryDAO.php
Inverle: I was considering a different approach too, which would be to have a second `priority2` column on the `feed` table instead of the `category` table, which would be updated for all feeds when the parent category has its priority changed and that column would be used for priority if feed's setting is set to *Use category setting*.

Something like this:
`CASE WHEN f.priority = 10 THEN f.priority2 ELSE f.priority END AS priorityNew`

instead of the current
`CASE WHEN f.priority = 10 THEN c.priority ELSE f.priority END AS priority2`

This approach still requires a join on the feed table, although not the category table anymore.

`priority2` would also need to be updated when a feed is moved to a different category.

Using joins is cleaner, but yes, it could be bad for performance. I have not checked yet, or for compatibility with other databases.

Inverle: I haven't put an index for `category`'s priority column like here, for example:

https://github.com/FreshRSS/FreshRSS/blob/7d7e50260347292db9912ec5f0daee5d393425b7/app/SQL/install.sql.sqlite.php#L37

Inverle:
> I have not read all the code yet. If not already the case, I think it is sufficient for feeds to be able to be less visible than their category, but not more visible than their category

This is what the default option for feed visibility looks like now:

image

The visibility is simply inherited from the category setting if *Use category setting* is used (where the default is *Main stream*)

Which means feed visibility can be both higher or lower than the category visibility, same for the category visibility, which can also be higher or lower than the feed visibility

PR #8913, opened 2026-06-05
Set limits for regex during search

Inverle:
Prevents regex searches from running for too long
(currently only for SQLite)


app/Models/EntryDAOSQLite.php
```diff
@@ -54,7 +54,16 @@ protected function registerSqlFunctions(string $sql): void {
 		// https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators
 		$this->pdo->sqliteCreateFunction('regexp',
 			function (string $pattern, string $text): bool {
-				return preg_match($pattern, $text) === 1;
+				$oldBacktrackLimit = ini_set('pcre.backtrack_limit', '10000');
+				$oldRecursionLimit = ini_set('pcre.recursion_limit', '100');
```
Alkarex: Would it be an idea to set those limit globally and not only there?

Inverle: I suppose it would be fine to do that, currently doing it this way makes sure that nothing overrides the limits.

Alkarex (approved):

PR #8916, opened 2026-06-10
Docker Alpine 3.24.0

Alkarex:
https://alpinelinux.org/posts/Alpine-3.24.0-released.html

Update to PHP 8.5 from PHP 8.4:
revert to using Alpine's default PHP version, which is now PHP 8.5.7 instead of the previous default of PHP 8.3.x.

We will want to first merge the PHP 8.5 bug I have just found in SimplePie https://github.com/simplepie/simplepie/pull/979

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.23
+FROM alpine:3.24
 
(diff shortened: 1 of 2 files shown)


Alkarex:
Done in:
* https://github.com/FreshRSS/FreshRSS/pull/8918

PR #8922, opened 2026-06-13
Halt further regex searching after a limit is hit once

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8913

More aggressive limiting as an additional mitigation, helps to further reduce the time of a malicious request in case there are a lot of articles. (on my main instance the time has fallen from 4s to 0.2s)


app/Models/EntryDAOSQLite.php
```diff
@@ -54,7 +54,16 @@ protected function registerSqlFunctions(string $sql): void {
 		// https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators
 		$this->pdo->sqliteCreateFunction('regexp',
 			function (string $pattern, string $text): bool {
-				return preg_match($pattern, $text) === 1;
```
Alkarex, line 57: What happens when just raising an exception here?

Inverle: Debug backtrace in the function looks like this:
```
#0 [internal function]: FreshRSS_EntryDAOSQLite->{closure:FreshRSS_EntryDAOSQLite::registerSqlFunctions():56}('...', '')
#1 /var/www/FreshRSS/app/Models/EntryDAO.php(1736): PDOStatement->execute()
#2 /var/www/FreshRSS/app/Models/EntryDAO.php(1776): FreshRSS_EntryDAO->listWhereRaw('...', 1, 3, Object(FreshRSS_BooleanSearch), '...', '...', '...', '...', '...', Array, 21, 0, '...', '...')
#3 /var/www/FreshRSS/app/Controllers/indexController.php(412): FreshRSS_EntryDAO->listWhere('...', 1, 3, Object(FreshRSS_BooleanSearch), '...', '...', '...', '...', '...', Array, 21, 0, '...', '...')
#4 [internal function]: FreshRSS_index_Controller::listEntriesByContext(21)
#5 /var/www/FreshRSS/app/Controllers/indexController.php(176): Generator->valid()
#6 /var/www/FreshRSS/app/views/index/normal.phtml(11): FreshRSS_index_Controller::{closure:FreshRSS_index_Controller::normalAction():172}(Object(FreshRSS_View))
#7 /var/www/FreshRSS/lib/Minz/View.php(95): include('...')
#8 /var/www/FreshRSS/lib/Minz/View.php(117): Minz_View->includeFile('...')
#9 /var/www/FreshRSS/app/layout/layout.phtml(96): Minz_View->render()
#10 /var/www/FreshRSS/lib/Minz/View.php(95): include('...')
#11 /var/www/FreshRSS/lib/Minz/View.php(108): Minz_View->includeFile('...')
#12 /var/www/FreshRSS/lib/Minz/View.php(75): Minz_View->buildLayout()
#13 /var/www/FreshRSS/lib/Minz/Dispatcher.php(57): Minz_View->build()
#14 /var/www/FreshRSS/lib/Minz/FrontController.php(61): Minz_Dispatcher->run()
#15 /var/www/FreshRSS/p/i/index.php(58): Minz_FrontController->run()
```
So, if an exception gets thrown, `$stm->execute()` stops executing completely. (we would need to start catching exceptions there though)
However, we may want to still keep searching for the other filters, while silently never matching any regex after the failure, aside from the log message.
Or is the preferred behavior to stop the query?

Alkarex: I have not carefully considered the details, but my first choice would be to stop the SQL request, return an HTTP 4xx or 5xx, and avoid global changes.

Inverle: Which HTTP code exactly?

Inverle: Problem is that currently I'm unable to figure out a way to use `Minz_Error::error()` from `listWhereRaw()` without getting `Warning: Cannot modify header information - headers already sent by (output started at /var/www/FreshRSS/app/layout/layout.phtml:48) in /var/www/FreshRSS/lib/Minz/Request.php on line 500` or nothing happening, if I use it without a redirect.

Inverle: I guess the entries are supposed to be buffered, so that's why.

Inverle: This commit stops the query now, but doesn't return a HTTP error, due to the reason above: https://github.com/FreshRSS/FreshRSS/pull/8922/commits/9594b2f3062269b3b280fb17f8282540c827c7b7

Alkarex:
* https://dev.mysql.com/doc/refman/9.7/en/server-system-variables.html#sysvar_regexp_stack_limit
* (old) https://jira.mariadb.org/browse/MDEV-8637

PR #8936, opened 2026-06-22
Document PHP-FPM write restrictions for FreshRSS data path

### LINKED ISSUE #8882: [Bug] HTTP 500 Permission denied for .../data/user/...
### Describe the bug

This happened to me since a few days, together with the fact, that I had to relogin every time I visited my news page. I triple checked the owner and permissions of all folders and files in /usr/share/Freshrss/... and everything was unchanged and as It should be: everything belonged to www-data:www-data and had write permissions for user and group. After doing some searches I found this helpful thread, explaining to me, that some policy change in php was the cause for this problem. In short: PHP assumes, that user data and configs do not belong into /usr/share/*, so there is no writable space expected. I had to search quite long for finding this, so here's the thread discussing this matter: 

[https://github.com/FreshRSS/FreshRSS/discussions/8828](https://github.com/FreshRSS/FreshRSS/discussions/8828)

This means, that maybe a rethink of the installation routines and of the FAQ might be necessary and helpful.
My working approach to remedy this was, btw., I moved the subfolder data recursively into a user home and placed a link to that folder in /usr/shared/Freshrss/.

### To Reproduce

1. visit your installation as usual
2. try to see user log, to change user data or to perform an update from the web form
3. See error

### Expected behavior

Working as usual

### FreshRSS version

1.26.2, 1.29.1.dev (irrelevant)

### System information

- Database version: irrelevant
- PHP version: PHP 8.3
- Installation type: git checkout, and sometimes internal update
- Web server type: Apache 2.4
- Device: every Browser
- OS: Ubuntu 22.04 (VM on proxmox-host with newer kernel)
- Browser: every browser

The rest is irrelevant for this problem, except: my system uses autoupdate for all apt-installations, so including PHP

nathanpixodeo:
Closes #8882

## Summary
- avoid recommending `/usr/share/` for writable FreshRSS data in the general install docs
- update the Debian/Ubuntu git install path and symlink example to `/var/www/FreshRSS`
- add an FAQ entry for PHP-FPM hardening that can make `/usr/` read-only despite correct Unix permissions

## Testing
- `git diff --check`
- `npx --yes markdownlint-cli@0.49.0 docs/en/admins/03_Installation.md docs/en/admins/06_LinuxInstall.md docs/en/users/07_Frequently_Asked_Questions.md`

### DIFF
--- a/docs/en/admins/03_Installation.md
+++ b/docs/en/admins/03_Installation.md
@@ -6,7 +6,7 @@ Before you begin, make sure that you’ve read the [prerequisites](02_Prerequisi
 
 1. If the computer you’re running on is not currently running a web server, you’ll first need to install and configure a web server, a version of PHP, and an appropriate database, as listed in the prerequisites. Specially, if you are using NGINX, please keep in mind that FreshRSS needs some specified variables to run properly. So make sure you’ve checked our configuration file. [Example Apache and Nginx configuration files can be found here](10_ServerConfig.md).
 
-2. Download your chosen version of FreshRSS, or fetch it via git. It’s advisable that you put FreshRSS in `/usr/share/`, and symlink the `./p/` folder to the root of your web server.[^1]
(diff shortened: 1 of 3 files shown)


Alkarex:
Thanks :-)

Issue #8168, opened 2025-10-30
[Feature] Allow Grouping "by day"


colinstu12:
### Is your feature request related to a problem? Please describe.

Presently posts/articles in feeds will be grouped by "Today", "Yesterday", and "Before Yesterday". 

Is there any way to add the feature of continuing this and simply allow grouping of all posts "By Day"? 

I find it very hard to motivate myself to read feeds when there's way too much stuff grouped / lumped together. 
It's a lot easier and more successful to have them chunked out more. Finishing off these bite-sized chunks just works a lot better for me, and feels like I'm making greater progress, so I continue and read more more-easily. 

### Describe the solution you’d like

Add an option in settings to allow posts in feeds to be grouped "By Day" (and potentially other lengths of time too, "Weekly" for sites that post less frequently for example or if desired by the reader, etc).

Alkarex:
Hello,
Maybe duplicate of the following ticket?
* https://github.com/FreshRSS/FreshRSS/issues/7520

Contributions welcome.

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/8174
Tests welcome

colinstu12:
@Alkarex super appreciate your efforts!! Thank you!!

abodnar:
How do you turn this off and why is it enabled by default?

This has made the view terrible.

<img width="932" height="736" alt="Image" src="https://github.com/user-attachments/assets/93e97e2e-49f5-44db-9528-22222ce3602d" />

Alkarex:
* https://github.com/FreshRSS/FreshRSS/discussions/8355#discussioncomment-15346265
* https://github.com/FreshRSS/FreshRSS/discussions/8358#discussioncomment-15398476

shadowlocked:
This has been the worst thing about finally updating to a later version of FreshRSS. It doesn't seem fair that one user can come in with a request and that request gets implemented, affecting other people, with no consultation and no way back.

Alkarex:
> This has been the worst thing about finally updating to a later version of FreshRSS. It doesn't seem fair that one user can come in with a request and that request gets implemented, affecting other people, with no consultation and no way back.

Way back: Transitions can be hidden if you dislike them: https://github.com/FreshRSS/FreshRSS/discussions/8358

Consultation: After this ticket, there was a PR open for feedback, and then 2 months in our rolling release before making it to a slow/stable release.

[...]

Issue #7520, opened 2025-04-24
[Feature] More date dividers/headers in normal view


pe1uca:
### Is your feature request related to a problem? Please describe.

When reading feeds with very old articles there's only "Before yesterday".  

### Use case

With the default sort, newer first, we get three dividers: "Today", "Yesterday", "Before yesterday". Which work great to focus on the new articles.  
If we order by oldest first we only get "Before yesterday"  

### Describe the solution you’d like

It'll be nice to have a few more date dividers.  
It'll be great to have "X days ago" then "X months ago", but maybe at least "This ago", "This month", "Last month", "This year".

Alkarex:
Implemented in https://github.com/FreshRSS/FreshRSS/pull/8174
Tests welcome

sans-c:
> Implemented in [#8174](https://github.com/FreshRSS/FreshRSS/pull/8174) Tests welcome

Is there a way to turn off this new way of displaying things  or turn off the "grouping dividers" in general?

I have some low activty feeds  with only one article per date for example. The result is, that the dividers now consume a lot of screen estate whereas before they felt a lot less intrusive.

Inverle:
https://github.com/FreshRSS/FreshRSS/discussions/8355#discussioncomment-15346265

PR #8938, opened 2026-06-22
Reintroduce today / yesterday / before yesterday

Frenzie:
References #8358.

Reintroduces the old display method as well as one without any date dividers whatsoever.

Draft partially because it's a poc and partially because I don't really understand how those transitions things are supposed to behave yet.

Switch between the settings in Settings → Display → Date transitions

image

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

### DIFF
--- a/README.fr.md
+++ b/README.fr.md
@@ -227,33 +227,33 @@ Voir le [dépôt dédié à ces extensions](https://github.com/FreshRSS/Extensio
 
 | Langage | Progression | |
 | - | - | - |
-| Čeština (cs) | ￭￭￭￭￭￭￭￭･･ 82% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fcs+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Čeština (cs) | ￭￭￭￭￭￭￭￭･･ 81% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fcs+%2F%28TODO%7CDIRTY%29%24%2F) |
 | Deutsch (de) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fde+%2F%28TODO%7CDIRTY%29%24%2F) |
-| Ελληνικά (el) | ￭￭￭･･･････ 38% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fel+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Ελληνικά (el) | ￭￭￭･･･････ 37% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fel+%2F%28TODO%7CDIRTY%29%24%2F) |
 | English (en) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen+%2F%28TODO%7CDIRTY%29%24%2F) |
 | English (United States) (en-US) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen-US+%2F%28TODO%7CDIRTY%29%24%2F) |
-| Español (es) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fes+%2F%28TODO%7CDIRTY%29%24%2F) |
-| فارسی (fa) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffa+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Español (es) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fes+%2F%28TODO%7CDIRTY%29%24%2F) |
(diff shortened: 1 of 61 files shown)


Alkarex:
It is easy to have *None* (which we alternatively already can with a simple CSS), and *Transitions*.
But the *Today* / *Yesterday* / *Older* comes with its own problems so I am not sure it is worth it (except if you like it).
The current transitions also have other features such as to jump between days.
And there are other sorting criteria than dates, so for instance sorting by category or by feed has their own transitions.

Frenzie:
> But the Today / Yesterday / Older comes with its own problems so I am not sure it is worth it (except if you like it).

I join the complainers in thinking it was quite preferable over both every day and none, and I'd rather revert the changes in question altogether to return to the situation as it was. :-)

These articles are (not) from today or yesterday is an extremely important fact. These articles are from 4 January 2026 is not, and on any single feed view it completely removes the visual clarity that comes from the today / yesterday / before yesterday division. Hiding the dividers is not a solution because then you don't have that clarity either.

What problem are the dividers trying to solve exactly?

Alkarex:
My impression is that it was mostly positive, both in real-life and online, e.g.:
https://framapiaf.org/@freshrss/115711956560451423

The cases of complain I have seen are mostly asking about removing the dividers altogether. Not so much about asking about an "older" divider (which I am not opposing - I just doubt there is much demand for it). While this can already be done by CSS, an option would be fine.

> What problem are the dividers trying to solve exactly?

* That "Today / Yesterday / Older" does not work for other sorting criteria than dates (which is something that has been asked a lot)
* That "Older" becomes enormous, without any division.
* To be consistent: why make a special case for yesterday and not for the day before, or the week before, for instance
* To allow easy filtering by dates and jump between dates

Note that with the current logic, the divider approach could relatively easily offer more options, such as aggregation window (e.g. day, week), or make an option to only show the first *n* dividers to reproduce something similar than *Today / Yesterday / Older*. And that could even be at global, category, and feed levels.

Frenzie:
> The cases of complain I have seen are mostly asking about removing the dividers altogether.

I don't believe that's true. Everyone in <https://github.com/FreshRSS/FreshRSS/discussions/8355> wants this behavior, as does everyone in <https://github.com/FreshRSS/FreshRSS/discussions/8358>, as well as in all other complaints I've seen.

The thing is, if you look at the main feed most of the time, particularly on a "larger" (non-mobile) screen, the distinction is not very noticeable. It feels extremely similar because you mostly deal with what coincidentally happens to be today / yesterday / older (technically day before yesterday). It's only once you go further into the past, or to individual feeds or categories, that it becomes clear how substantially different and worse it is to put a header above every single day.

> That "Today / Yesterday / Older" does not work for other sorting criteria than dates (which is something that has been asked a lot)

Correct, the dividers don't serve any purpose in other scenarios.

> That "Older" becomes enormous, without any division.

The dividers have to be meaningful, and when they're not they merely take up some space in a good case (with a bunch of articles underneath), while unfortunately they range between distracting noise and a waste of space in a bad case (on days with only a single or a handful of articles), especially on mobile.

> To be consistent: why make a special case for yesterday and not for the day before, or the week before, for instance

The week / month / year before would be acceptable but unnecessary. But the underlying reason is simple: because we're humans, not machines. Compare how you deal with physical newspapers. The most recent couple of days are out on the table; the remainder of the archive is nearby until it's recycled.

With or without the day before yesterday is one of those choices that simply isn't wrong. You just have to pick one. Perhaps there's a way to make it simple for the user to customize the pattern.

> To allow easy filtering by dates and jump between dates

Hm, I don't find that easy at all, since if I'm already there what do I need to filter for in the first place? Unless you mean something like click on a random date and then edit the URL to the date you actually want, which I do actually find convenient but it somewhat goes against modern UI conventions. ;-)

> make an option to only show the first n dividers to reproduce something similar than Today / Yesterday / Older.

That's explicitly what we don't want. :-) (But maybe someone does.) Then you get something like:

```
24 August 2025
…
5 May 2023
…
12 December 2018
```

Which is simply restating the problem description. Those are just some random, meaningless dates.

Alkarex:
Yes, it is common to mostly read a few recently received articles, but there are more use-cases, including working with archives, searches, filters, and that potentially on large amounts of articles.

I personally use the navigation between dates a lot. See also https://github.com/FreshRSS/FreshRSS/issues/8168

The re-implementation of the transitions was among other things triggered by the new sorting options. See e.g. https://github.com/FreshRSS/FreshRSS/discussions/8172
For instance, in addition to what already written above, the previous implementation *today / yesterday / before yesterday* was also buggy when sorting by publication dates due to some dates potentially being in the future.

And yes, your example is exactly what I want, with a separation between a group of articles at a date in 2023 and another group of articles at a date in 2018.

But back to solutions: Now as mentioned, I think it would be nice to implement improvements and/or introduce more options (which could be different at global, category, and feed levels) - and your PR goes in that direction.

It might be possible to find a more powerful approach or option satisfying more cases including yours.

My suggestion to make an option to only show the first *n* date transitions was to be understood like so (with an example of n=3):

>Received today — 23 June 2026
>...
>Received yesterday — 22 June 2026
>...
>Received — 21 June 2026 __or before__ (or any variation of that: *Older*, *Until*...)

An alternative, maybe better idea is to have an option with incremental granularity (see also https://github.com/FreshRSS/FreshRSS/issues/7520)

> Received today — 23 June 2026
> ...
> Received yesterday — 22 June 2026
> ...
> Received earlier this week
> ...
> Received earlier this month
> ...
> Received earlier this year
> ...
> Received last year or before

Frenzie:
> Yes, it is common to mostly read a few recently received articles, but there are more use-cases, including working with archives, searches, filters, and that potentially on large amounts of articles.

Exactly, and that scenario is where those headers get in the way, removing clarity. Putting display: none on them for that reason throws out the baby out with the bathwater because they're desired for the well thought out user interface design in the today / yesterday / older dynamic.

> And yes, your example is exactly what I want, with a separation between a group of articles at a date in 2023 and another group of articles at a date in 2018.

A group is the optimistic scenario that occurs some of the time, and when it's a proper group it ranges from a bit silly at worst to actively useful. But it falls apart for groups smaller than 5 or maybe even 10-20 articles, resulting in a page completely taken over by headers instead of actual content.

> For instance, in addition to what already written above, the previous implementation today / yesterday / before yesterday was also buggy when sorting by publication dates due to some dates potentially being in the future.

The bug would likely be showing it in sort by publication date at all. The concept doesn't make much sense outside of sort by reception date.

> An alternative, maybe better idea is to have an option with incremental granularity

It's fine, although I'd consider it a waste of computation time.

> My suggestion to make an option to only show the first n date transitions was to be understood like so (with an example of n=3):

Oh right, so just the old mechanism but with a dynamic number. n as in "days in the past from now", not n as in "days in the results".

> Received today — 23 June 2026
> ...
> Received yesterday — 22 June 2026

Definitely not with the received. The former today / yesterday / before yesterday is the way to go. In fact you just made me realize the "received" spam might play some role here. It's less visually overwhelming without.

Frenzie:
For example, you could envision something subtler, in this general direction.

image

Alkarex:
Alternative proposition (draft):
* https://github.com/FreshRSS/FreshRSS/pull/8967

PR #8945, opened 2026-06-27
Use timing safe comparison for tokens everywhere

Inverle:
Note: I already changed some places that were using normal string comparison to `hash_equals` before, but didn't check other places.

Now the checks should be consistent.


app/Models/Auth.php
```diff
@@ -216,9 +216,9 @@ public static function csrfToken(): string {
 	public static function isCsrfOk(?string $token = null): bool {
 		$csrf = Minz_Session::paramString('csrf');
 		if ($token === null) {
-			$token = $_POST['_csrf'] ?? '';
+			$token = Minz_Request::paramString('_csrf');
```
Alkarex: I would rather make sure this comes from POST parameter and not query parameter

PR #8948, opened 2026-06-28
Report new articles count per feed in actualize_script output

### LINKED ISSUE #8291: [Feature] Fetched articles count on actualize_script output

It would be nice if `actualize_script.php` could display the fetched articles per feed on the output.

### Use case

As I run FreshRSS on a hosted server and update the feeds with with a systemd timer triggered 2 times per week, I could then setup a script in `ExecStartPost` to mail me a report with the contents of `journalctl -t FreshRSS --since=today` and have a nice report of activities

Something like
```
FreshRSS starting feeds actualization at YYYY-MM-DDTHH:MM:SS+00:00
FreshRSS actualize joeuser…
FreshRSS SimplePie GET 200 https://example.com/feed
FreshRSS fetched 50 articles with 25 new from FriendlyFeedName (https://example.com/feed)
…
FreshRSS actualization done [...]
```
or any other output layout you see fit 😄 

None as only FreshRSS can provide the needed info

### Additional context

_No response_

JamBalaya56562:
Closes #8291

`app/actualize_script.php` now reports, for each feed that received new articles, how many new articles were fetched — in addition to the existing per-user summary. This makes batch refreshes (cron / systemd timers) much more informative.

Example output:

```
FreshRSS actualize admin…
	10 new article(s) from feed: FreshRSS releases
	3 new article(s) from feed: Some other blog
```

The feature is implemented **entirely within the CLI script**, reusing the existing extension-hook mechanism that the script already uses for the actualization mutex — no change to `feedController` or to any other behaviour:

- `FeedBeforeActualize` records each feed's name.
- `EntryBeforeAdd` counts new entries per feed. This hook fires only in the "genuinely new GUID" branch of the actualization loop (updated articles go through `EntryBeforeUpdate`), so the count matches the controller's own `$nbNewArticles` counter. Because the counting hook is registered after `$app->init()`, it runs last in the chain and is skipped for entries dropped by another extension — i.e. it counts only articles that are actually added.

Counts are printed through the script's existing `notice()` helper, so they reach STDOUT regardless of the `environment` setting (unlike `Minz_Log::notice`, which is suppressed in `production`).

1. `make start` (or any FreshRSS install) with at least one feed that has new articles.
2. Run `php ./app/actualize_script.php`.
3. Observe a `N new article(s) from feed: …` line for each feed that fetched new articles.

Verified locally with the `freshrss/freshrss:alpine` image (SQLite, default feeds): the default *FreshRSS releases* feed correctly reported `10 new article(s) from feed: FreshRSS releases`. `php -l`, `phpcs`, and `phpstan` all pass.

## Pull request checklist

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard) — the change is integration-level CLI output
- [x] documentation updated (no user-facing documentation change needed)


app/actualize_script.php
```diff
@@ -96,15 +96,32 @@ function notice(string $message): void {
...
+	Minz_ExtensionManager::addHook(Minz_HookType::EntryBeforeAdd, static function (FreshRSS_Entry $entry) use (&$nbNewArticlesByFeed) {
+		$nbNewArticlesByFeed[$entry->feedId()] = ($nbNewArticlesByFeed[$entry->feedId()] ?? 0) + 1;
+		return $entry;
+	});
 
 	notice('FreshRSS actualize ' . $user . '…');
 	echo $user, ' ';	//Buffered
 	$app->run();
 
+	foreach ($nbNewArticlesByFeed as $feedId => $nbNew) {
```
Alkarex: Should the list be sorted? For instance alphabetically, and/or by descending number of articles?

app/actualize_script.php
```diff
@@ -96,15 +96,32 @@ function notice(string $message): void {
...
+	/** @var array<int,string> $feedNamesById */
+	$feedNamesById = [];
+
+	Minz_ExtensionManager::addHook(Minz_HookType::FeedBeforeActualize, static function (FreshRSS_Feed $feed) use ($mutexFile, &$feedNamesById) {
 		touch($mutexFile);
+		$feedNamesById[$feed->id()] = $feed->name() ?: $feed->url(false);
 		return $feed;
 	});
+	Minz_ExtensionManager::addHook(Minz_HookType::EntryBeforeAdd, static function (FreshRSS_Entry $entry) use (&$nbNewArticlesByFeed) {
+		$nbNewArticlesByFeed[$entry->feedId()] = ($nbNewArticlesByFeed[$entry->feedId()] ?? 0) + 1;
```
Alkarex: Note that this covers both new articles, and updated articles

app/actualize_script.php
JamBalaya56562: Done — the per-feed report is now sorted by descending article count, then alphabetically by feed name as a tie-breaker (covers the "and/or": most-active feeds first, with a stable alphabetical order within the same count).

app/actualize_script.php
JamBalaya56562: Good catch, thanks. I traced the actualization loop: `EntryBeforeAdd` (in `feedController.php`, the new-GUID branch, `_isUpdated(false)`) fires only for genuinely new articles, while updated articles (existing GUID, changed hash) go through `EntryBeforeUpdate`. So the original single count was new-only.

I've now added a separate `EntryBeforeUpdate` hook and report the two distinctly, e.g. `10 new, 2 updated article(s) from feed: X` (feeds with only new — or only updated — show just that part). Verified locally: after tampering one existing entry's hash to force re-processing, the run reports `1 updated article(s)` for that feed while the unchanged feeds report nothing.

Alkarex:
I have reduced the code a bit. Double-check welcome

JamBalaya56562:
Thanks! The consolidation into `$feedStatistics` reads well, but I think it introduced one regression: `FeedBeforeActualize` fires *before* the TTL/mute skips in `feedController`, so feeds that are then skipped (or fetched with no change) stay in `$feedStatistics` with `new = 0, updated = 0`, and the output loop has no empty-guard — they now print as a bare `" article(s) from feed: X"`.

Reproduced locally with 2 feeds, running the actualization twice in a row (so TTL skips both on the 2nd run):

```
 article(s) from feed: FreshRSS releases
 article(s) from feed: FreshRSS tags
```

A small guard at the top of the loop restores the "only report feeds that fetched something" behavior:

```php
if ($row['new'] === 0 && $row['updated'] === 0) {
    continue;
}
```

Happy to push that if you'd like.

Alkarex:
Could catch and yes, please do

JamBalaya56562:
Pushed. Added the guard at the top of the output loop. Verified locally: a second immediate run (feeds skipped by TTL) now prints nothing, while feeds with new/updated articles still report correctly. `php -l`, `phpcs` and `phpstan` all pass.

Alkarex (approved):

Alkarex:

PR #8950, opened 2026-06-28
Improve SSRF mitigations by restricting `CURLOPT_PROXY`

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8400
SimplePie PR: https://github.com/FreshRSS/simplepie/pull/83


lib/simplepie/simplepie/src/File.php
```diff
@@ -114,6 +114,8 @@ public function __construct(string $url, int $timeout = 10, int $redirects = 5,
             }
             if (!$force_fsockopen && function_exists('curl_exec')) {
                 $resolve = false; // FreshRSS
+                $proxy = $curl_options[CURLOPT_PROXY] ?? null; // FreshRSS
+                $proxy_type = $curl_options[CURLOPT_PROXYTYPE] ?? null; // FreshRSS
                 if (empty($curl_options[CURLOPT_PROXY] ?? null)) { // FreshRSS
```
Inverle: https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-8p8w-9h98-44j5

reports:

> FreshRSS `edge` protects direct feed fetches with `getCurlResolveInfo()`, which rejects private and reserved destination addresses unless they are explicitly allowlisted. `FreshRSS_http_Util::httpGet()` skips that control whenever `CURLOPT_PROXY` is a non-empty string. A regular user can provide a non-empty proxy address while choosing the UI’s `NONE` proxy type (`CURLOPT_PROXYTYPE=-1`). The application then treats the request as proxied for the allowlist decision, while the effective request is made directly to the feed URL.


I need to check if any extra hardening is needed

Alkarex:
I have made a few fixes. Double-check welcome

Alkarex (approved):

lib/simplepie/simplepie/src/File.php
Inverle: I don't think I can reproduce it (at least after your changes, haven't tested before)
I suppose this is what this PR is supposed to prevent anyway.

Inverle:
Looks good to me. Thanks for fixing up the code!

PR #8954, opened 2026-06-30
Support choosing the sort type when creating or editing a user query

iatman:
Changes
- User queries can now store a sort field (sort type), not only the sort direction. The create and edit forms expose a single combined dropdown (field + direction) that reuses the existing `index.menu.sort.*` options, matching the global reading-settings sort UI.
- The "Bookmark current query" action now also remembers the current sort field, where it previously kept only the order direction.
- Relabel the query sort control from "Sort by date" to "Sort by", and remove the now-unused `conf.query.order_asc` / `conf.query.order_desc` translation keys.

Implementation
- `FreshRSS_UserQuery`: add an optional `sort` property mirroring `order` (constructor parsing, `toArray()`, `getSort()` falling back to the user default sort, and a `hasParameters()` check).
- `configureController`: split the combined dropdown value into `sort` and `order`, validating the field against the implemented sort types (`id`, `c.name`, `date`, `f.name`, `length`, `link`, `title`, `rand`); separate `sort`/`order` request parameters are still accepted so the "bookmark current query" flow keeps working.
- Views render the combined dropdown and persist `sort` as a hidden field.
- `p/api/query.php`: back-fill `sort` from the saved query when the request does not provide one, mirroring the existing `order` handling.
- Add unit tests covering the sort round-trip and `toArray()` inclusion.

Compatibility
- No database or schema change: user queries are stored in the per-user PHP config array, which keeps arbitrary keys verbatim, so no migration is required.
- Upgrade: existing queries without `sort` fall back to the user default sort; behaviour is unchanged until a sort is explicitly chosen.
- Downgrade: older versions ignore the extra `sort` key and keep working; re-saving under an old version simply drops the key with no data loss.
- Shared query URLs without a `sort` parameter remain valid.

## Tests
- `make`-equivalent run in Docker (PHP 8.4): PHPUnit (638 tests), PHP_CodeSniffer, PHPStan — all green.

Closes #

-
-
-

1. In Create a query, set a Filter type (e.g. a category or All).
2. In the Sort by dropdown, pick a non-default option, e.g. Title A→Z (title_asc).
3. Submit.
image

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


iatman:
Hi @Alkarex 👋 — this PR is ready for review whenever you have a moment.

It lets user queries choose a sort criterion (field + direction), reusing the existing sort options, with no DB/schema change (fully backward-compatible). The branch is rebased on the latest edge, conflicts are resolved, and PHPUnit / PHPCS / translations all pass locally. Happy to adjust anything on request — thanks!

PR #8964, opened 2026-07-02
Fix bad header filtering bypass

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/7496

Removes the filtering code from the view and `httpGet()`, while only including it in `sanitizeCurlParams()`, which is called both while making SimplePie requests, and while making other requests using `httpGet()`:

https://github.com/FreshRSS/FreshRSS/blob/c90c86a5c0beb28ca82d36a6a4cf18d3c539fed9/app/Models/SimplePieCustom.php#L32
https://github.com/FreshRSS/FreshRSS/blob/c90c86a5c0beb28ca82d36a6a4cf18d3c539fed9/app/Utils/httpUtil.php#L441-L443

Every read and save of the feed attribute `curl_params` is now sanitized with `sanitizeCurlParams()` as well.

The filtering was also improved by matching different separator characters (other than `-`), since they are allowed by some web servers such as PHP dev server, and turned into `HTTP_X_WEBAUTH_USER` in `$_SERVER` from `X_WebAuth_User`, for instance.

### DIFF
--- a/app/Controllers/feedController.php
+++ b/app/Controllers/feedController.php
@@ -221,7 +221,7 @@ class FreshRSS_feed_Controller extends FreshRSS_ActionController {
 			}
 
 			$attributes = [
-				'curl_params' => empty($opts) ? null : $opts,
+				'curl_params' => empty($opts) ? null : FreshRSS_http_Util::sanitizeCurlParams($opts),
 			];
 			$attributes['ssl_verify'] = Minz_Request::paramTernary('ssl_verify');
 			$timeout = Minz_Request::paramInt('timeout');
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -129,6 +129,11 @@ final class FreshRSS_http_Util {
 			if ($k === CURLOPT_COOKIEFILE) {
 				$curl_params[$k] = '';
 			}
+			// Remove HTTP authentication headers problematic for security
+			if ($k === CURLOPT_HTTPHEADER && is_array($curl_params[$k])) {
+				$curl_params[$k] = array_filter($curl_params[$k],
+					fn($header) => is_string($header) && !preg_match('/^(Remote[-_\s]*User|X[-_\s]*WebAuth[-_\s]*User)\\s*:/i', $header));
+			}
 		}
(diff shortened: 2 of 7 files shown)


app/views/helpers/feed/update.phtml
```diff
@@ -914,9 +914,6 @@
 							$httpHeaders = [];
 						}
 						$httpHeaders = array_filter($httpHeaders, 'is_string');
-						// Remove headers problematic for security
```
Alkarex, line 917: I think we need to call `sanitizeCurlParams()` in `feedAction()` (and potentially here before display)

Inverle: Done

app/Controllers/feedController.php
```diff
@@ -221,7 +221,7 @@ public function addAction(): void {
 			}
 
 			$attributes = [
-				'curl_params' => empty($opts) ? null : $opts,
+				'curl_params' => empty($opts) ? null : FreshRSS_http_Util::sanitizeCurlParams($opts),
 			];
```
Inverle, line 225: In this case it's saved with `_attributes()` later, not `_attribute()`.

Alkarex (approved):

app/Utils/httpUtil.php
```diff
@@ -129,6 +129,11 @@ public static function sanitizeCurlParams(array $curl_params): array {
 			if ($k === CURLOPT_COOKIEFILE) {
 				$curl_params[$k] = '';
 			}
+			// Remove HTTP authentication headers problematic for security
+			if ($k === CURLOPT_HTTPHEADER && is_array($curl_params[$k])) {
+				$curl_params[$k] = array_filter($curl_params[$k],
+					fn($header) => is_string($header) && !preg_match('/^(Remote[-_\s]*User|X[-_\s]*WebAuth[-_\s]*User)\\s*:/i', $header));
```
Inverle, line 135: I could've double escaped the `\s` here, but it's not needed.

Alkarex: Yes, I can slightly better like double escaping, but fine

PR #8968, opened 2026-07-02
Implement fragment navigation via links in article

Inverle:
The idea is if you click on a link that has a fragment at the end of it, e.g. `https://example.com/article#xyz` and it is the same URL as the article URL, but with a fragment (`#xyz`), it should scroll to the element with the ID `xyz` in the article.
Most commonly those links can be found in a table of contents.

It had always annoyed me, that when I clicked on such a link, it would get opened in a new tab, instead of just scrolling to the header.

Since this is in JS it won't work in external apps using the API, so I'll have to test some apps, to see if it's possible to make it work there too. Probably by editing SimplePie code to allow for raw fragment links like `#xyz` instead of turning them into `https://example.com/article#xyz`


Frenzie:
You added some prefix to ensure it's not `#xyz` but `#id-xyz` I imagine?

Inverle:
> You added some prefix to ensure it's not `#xyz` but `#id-xyz` I imagine?

`[entry id]-` is used as prefix

Frenzie:
Is each entry ID unique or does it also need to be combined with a feed ID? :-)

Inverle:
Yes, they are unique.

Alkarex:
I did a quick test with https://linuxfr.org/news/agenda-du-libre-pour-la-semaine-34-de-l-annee-2026 (from https://linuxfr.org/news.atom), which does not seem to work. I have not done any debugging, but I have a suspicion that there is an issue with special characters

Inverle:
> I did a quick test with https://linuxfr.org/news/agenda-du-libre-pour-la-semaine-34-de-l-annee-2026 (from https://linuxfr.org/news.atom), which does not seem to work. I have not done any debugging, but I have a suspicion that there is an issue with special characters

It should be fixed in 20846aa

PR #8969, opened 2026-07-03
Add option to keep or reset article sort order when navigating

### LINKED ISSUE #8953: [Bug]  Sort/order parameters from one category bleed into navigation links for all other categories
### Describe the bug

When a user manually selects a sort order (e.g., `sort=date&order=ASC`) while viewing a feed category, the chosen sort parameters are appended to the sidebar navigation links for **all** other categories. Clicking any other category then carries over the previous category's sort settings, overriding that category's own configured default sort.

**Root Cause**

In `app/layout/aside_feed.phtml`, the sidebar link builder reads the current request's `sort` and `order` parameters and appends them to every category/feed link via `$state_filter_manual`:

```php
// app/layout/aside_feed.phtml, lines 16–20
if (($s = Minz_Request::paramString('sort', plaintext: true)) !== '' && preg_match('/^[a-z.]+$/', $s)) {
    $state_filter_manual .= '&sort=' . $s;
}
if (($s = Minz_Request::paramString('order', plaintext: true)) !== '' && ctype_alpha($s)) {
    $state_filter_manual .= '&order=' . $s;
}
```

In `app/Models/Context.php`, the category's `defaultSort`/`defaultOrder` is only applied when `sort`/`order` are **absent** from the URL:

```php
// app/Models/Context.php, lines 268–294
if (Minz_Request::paramString('order', plaintext: true) === '' || Minz_Request::paramString('sort', plaintext: true) === '') {
    // reads $category->defaultSort() / $category->defaultOrder()
}
$sort  = Minz_Request::paramString('sort',  plaintext: true) ?: $default_sort  ?: ...;
$order = Minz_Request::paramString('order', plaintext: true) ?: $default_order ?: ...;
```

Because the sidebar always injects `sort` and `order` into every link once they appear in the URL, the per-category defaults are never reached when navigating from a category where the sort was manually changed.

**Possible Fix**

When generating sidebar links for a target category/feed, instead of blindly forwarding the current `sort`/`order` parameters, check whether the target category/feed has its own `defaultSort`/`defaultOrder` configured. If it does, omit the `sort`/`order` query parameters from that link so that `Context.php` can apply the target's own defaults. Only forward the parameters if the target has no configured default.

Alternatively, the `state` parameter (read/unread filter) could be treated separately from `sort`/`order`, since preserving the read-state filter across navigation is arguably more useful than preserving a manually chosen sort order.

**Environment**

- Repository: FreshRSS/FreshRSS
- Affected files:
  - `app/layout/aside_feed.phtml` (lines 12–21, 100)
  - `app/Models/Context.php` (lines 266–294)

---

Feel free to adjust the "Possible Fix" section based on the intended design direction before filing.

### To Reproduce

1. Navigate to any feed category (e.g., Category A).
2. Manually change the sort order (e.g., sort by date, ascending).
3. Without changing anything else, click on a different category (Category B) in the sidebar.
4. Observe that Category B is now sorted by date/ascending instead of its own default sort order.

### Expected behavior

Each category should be opened with its own configured `defaultSort`/`defaultOrder`. A manually selected sort in one category should not carry over to a different category.

### FreshRSS version

1.29.1

### System information

- Docker images( freshrss + Sqlite )

### Additional context

The sort and order parameters from the current URL are appended to all sidebar category links. When navigating to another category, those parameters are present in the URL, causing `Context.php` to skip reading the target category's `defaultSort`/`defaultOrder` and use the carried-over values instead.

iatman:
Adds a new per-user reading option, "Keep the current sort order when navigating between categories and feeds" (config key: sticky_sort), under Settings > Reading > "Left navigation: Categories". It defaults to enabled, which reproduces the current behaviour exactly.

Background (addresses #8953)
----------------------------
Once a sort/order was active in one view, it was propagated to every sidebar link and to the "mark all as read" redirect. Opening another category then ignored that category's own defaultSort/defaultOrder, because FreshRSS_Context only applies those defaults when sort/order are absent from the request. As a result the current view's sort (whether chosen manually or resolved from that view's default) "bled" into the next category/feed.

Behaviour of the new option
---------------------------
- ON (default): unchanged. The current sort/order stays sticky while navigating, exactly as before.
- OFF: sidebar links and the "mark all as read" redirect no longer carry sort/order, so each target resolves its own order in FreshRSS_Context: feed/category defaultSort/defaultOrder, falling back to the global Reading sort criterion when none is configured.

order/sort propagation across the UI (for reference) 
----------------------------------------------------
- Categories & feeds (sidebar): now gated by sticky_sort (app/layout/aside_feed.phtml).
- "Mark all as read" -> next category, both the top navigation button and the stream-footer button: now gated by sticky_sort (app/Controllers/entryController.php::readAction, non-AJAX redirect).
- Custom user queries: unaffected. A query's link is a self-contained URL built from that query's own saved parameters, so switching query A -> B never carried A's sort; no option is needed there.
- Stream transition links when sorting by category/feed name: intentionally left unchanged (out of scope; the sort is intrinsic to that grouping).

Compatibility
-------------
- Not a breaking change: the default (true) reproduces the previous behaviour.
- No database migration. The setting is a standard user configuration value with its default declared in config-user.default.php, so existing users keep the previous behaviour transparently.
- The read/unread state filter still persists across navigation regardless of this option.

i18n
----
English strings authored; en-US marked as IGNORE. Translations for the other 25 languages were produced with an automated (AI) translation service and should be reviewed by native speakers.

Closes #8953

-
-
-

1.
2.
3.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated


Alkarex:
I am a bit worried of the AI production of translations, which I cannot review, but we can maybe give it a try for this time

app/i18n/pl/conf.php
```diff
@@ -311,6 +311,8 @@
 			'when' => 'Oznacz wiadomość jako ulubioną…',
 		),
 		'sticky_post' => 'Przesuń wiadomość na górę strony po otworzeniu',
+		'sticky_sort' => 'Zachowaj bieżącą kolejność sortowania podczas przełączania między kategoriami i kanałami',
+		'sticky_sort_help' => 'Gdy odznaczone, każda kategoria i kanał otwiera się z własną domyślną kolejnością sortowania (lub z globalnym kryterium sortowania powyżej, jeśli żadne nie jest ustawione), zamiast zachowywać ostatnio wybraną kolejność sortowania.',
```
Inverle: Looks good

app/i18n/nl/conf.php
```diff
@@ -311,6 +311,8 @@
 			'when' => 'Markeer een artikel als favoriet…',
 		),
 		'sticky_post' => 'Koppel artikel aan de bovenkant als het geopend wordt',
+		'sticky_sort' => 'Huidige sorteervolgorde behouden bij het navigeren tussen categorieën en feeds',
+		'sticky_sort_help' => 'Indien uitgeschakeld, wordt elke categorie en feed geopend met zijn eigen standaardsortering (of met het algemene sorteercriterium hierboven als er geen is ingesteld), in plaats van de laatst geselecteerde sorteervolgorde te behouden.',
```
Frenzie: It's acceptable, but a bit awkward. I'd mark 'em as dirty (if at all, my preference would be a regular todo).

app/i18n/en/conf.php
```diff
@@ -311,6 +311,8 @@
 			'when' => 'Mark an article as favourite…',
 		),
 		'sticky_post' => 'Stick the article to the top when opened',
+		'sticky_sort' => 'Keep the current sort order when navigating between categories and feeds',
```
Frenzie:
```suggestion
		'sticky_sort' => 'Keep sort order across categories and feeds',
```

app/i18n/en/conf.php
```diff
@@ -311,6 +311,8 @@
...
+		'sticky_sort_help' => 'When unchecked, each category and feed opens with its own default sort order (or the global sort criterion above when none is configured), instead of keeping the sort order you last selected.',
```
Frenzie:
```suggestion
		'sticky_sort_help' => 'Determines whether the last sort order is kept active or whether each category or feed always uses its own default or global setting.',
```

Frenzie (commented):
I realized the blame was at least in part with the original. Something like this perhaps.

iatman:
Thanks to Alkarex, Inverle, and Frenzie.
Using AI for i18n translation is an experiment; 
feel free to review the functional code first, and I can revert the i18n translations to English-only if necessary.

Alkarex:
Please re-apply your translations now that the English version (and French version) have been updated

iatman:
Thanks to Alkarex
Update complete.

Alkarex:
@Frenzie Is it better in Dutch now?

app/i18n/nl/conf.php
```diff
@@ -311,6 +311,8 @@
 			'when' => 'Markeer een artikel als favoriet…',
 		),
 		'sticky_post' => 'Koppel artikel aan de bovenkant als het geopend wordt',
+		'sticky_sort' => 'Handmatige sorteervolgorde behouden tijdens navigeren',
+		'sticky_sort_help' => 'Bepaalt of de laatste handmatige sorteervolgorde actief blijft of dat elke categorie of feed altijd zijn eigen standaard- of algemene instelling gebruikt.',
```
```suggestion
		'sticky_sort' => 'Aangepaste sorteervolgorde behouden bij het bladeren',
		'sticky_sort_help' => 'Bepaalt of de laatste aangepaste sorteervolgorde actief blijft of dat elke categorie of feed altijd zijn eigen standaard- of algemene instelling gebruikt.',
```

Frenzie:
> @Frenzie Is it better in Dutch now?

Meh, manual as _handmatig_ applies to a completely different context. Here the meaning is "custom," not "by hand and not automated by a machine." I'd definitely mark 'em all as `DIRTY` like I also did in #8988.

Frenzie:
Though come to think of it, "custom sort order" is more idiomatic English anyway.

Alkarex (approved):

PR #8966, opened 2026-07-02
Worked on Turkish translation

utkutibet:
Changes proposed in this pull request:

- Completed all the remaining translations for Turkish, located in admin.php, api.php, conf.php. 

How to test the feature manually:

1. Set the FreshRSS interface language to Turkish (Settings → Profile → Language)
2. Navigate through the pages affected by the changed strings (listed above)
3. Confirm the Turkish text displays correctly with no missing or broken characters

Pull request checklist:

- [X] clear commit messages
- [ ] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #8971, opened 2026-07-03
Remove `.DS_Store` files and update `.gitignore`

Inverle:
The `.DS_Store` files were accidentally added in #8966


.gitignore
```diff
@@ -12,6 +12,11 @@
 
 /docs/_site/
 /docs/.sass-cache/
+/docs/.bundle/
+/docs/.jekyll-metadata
```
Inverle, line 16: Some other paths worth adding I found in my `git status`.

PR #8972, opened 2026-07-03
Move custom favicons into their own directory

Inverle:
The backup documentation states:

https://github.com/FreshRSS/FreshRSS/blob/8a5f7d24b6f73e529bec042e22eeb3148e0166fc/docs/en/admins/05_Backup.md?plain=1#L3-L5

It says to skip `favicons/`, since it's rebuilt, but this is not true for custom favicons, which are also located there.
Those are currently recognized in `favicons/` by a hash, having a corresponding `.ico` file, but not a `.txt` file.

This change makes custom favicons easier to backup by having a separate directory for them (`custom-favicons`)


Inverle:
Is it intended that some directories are not included here for a check?

https://github.com/FreshRSS/FreshRSS/blob/8a5f7d24b6f73e529bec042e22eeb3148e0166fc/lib/lib_install.php#L46-L51

For example `Retry-After`, `PubSubHubbub` and `extensions-data`?

app/migrations/2026_07_03_MoveCustomFavicons.php
Inverle, line 1: Note: This PR is introducing the first migration into FreshRSS.
If someone needs to rollback their version, they should run `mv data/custom-favicons/*.ico data/favicons/ && rm -rf data/custom-favicons/` and remove the migration manually from `data/applied_migrations.txt`.

Inverle: It would be nice for migrations to be able to have a `rollback()` function, so that you could copy the migration file from a newer version into the older version, and run the rollback with a CLI script.

Or the migration could copy itself into some directory that doesn't get touched after downgrading, and rollback automatically. (but such a mechanism needs to be implemented in the older version first)

Alkarex:
Wouldn’t it be easier and more efficient to just update the documentation to not skip favicons?
I am not super found of the migration mechanism with concerns ranging from performance, to multi-user handling, to back-up restore situations

Inverle:
The migration runs only once and is applied to every user during that run, moving files is pretty fast, so I don't think there should be concerns about performance. (unless you mean that migrations are checked on every request https://github.com/FreshRSS/FreshRSS/pull/2760#issuecomment-619356885)

Only annoyance is if you want to go back to an older version, or if the migration fails for some reason, and you have to fix the server.

Maybe it would be fine to update the documentation, but it would result in that you have to make bigger backups unnecessarily (including lots of standard favicons, which can simply be redownloaded), and if you want to run a command which copies all of the .ico files which don't have a .txt file next to them, then that command won't be simpler than the migration script, I imagine.

It's up to you, though, since this is a kind of "breaking change".

Inverle:
> The migration runs only once and is applied to every user during that run

Since custom favicons aren't attached to users directly, and only are recognized by a hash, it makes sense to use a global migration here.

Inverle:
https://github.com/FreshRSS/FreshRSS/pull/8992

PR #8612, opened 2026-03-16
Improve consistency of slider behavior after submitting form

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/8529

TODO:
* [x] fix the mistakenly replaced indents by phpcbf
* [x] make sure `error` parameter gets preserved after submitting form in subscription management

PR #8974, opened 2026-07-03
Keep scroll position of slider after form submit

Inverle:
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8612#issuecomment-4078560064

Depends on https://github.com/FreshRSS/FreshRSS/pull/8973

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -389,6 +389,13 @@ function init_slider(slider) {
 	document.querySelector('#slider .toggle_aside').addEventListener('click', close_slider_listener);
 
 	if (slider.children.length > 0) {
+		const slider_scrollTop = sessionStorage.getItem('FreshRSS_slider_scrollTop');
+		if (slider_scrollTop) {
+			slider.scrollTop = +slider_scrollTop;
+			sessionStorage.removeItem('FreshRSS_slider_scrollTop');
+		}
+		slider.submitListener = () => sessionStorage.setItem('FreshRSS_slider_scrollTop', slider.scrollTop);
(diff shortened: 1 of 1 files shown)


p/scripts/extra.js
```diff
@@ -389,13 +389,21 @@ function init_slider(slider) {
 	document.querySelector('#slider .toggle_aside').addEventListener('click', close_slider_listener);
 
 	if (slider.children.length > 0) {
+		const slider_scrollTop = sessionStorage.getItem('FreshRSS_slider_scrollTop');
+		if (slider_scrollTop) {
+			slider.scrollTop = +slider_scrollTop;
+			sessionStorage.removeItem('FreshRSS_slider_scrollTop');
+		}
+		slider.submitListener = () => sessionStorage.setItem('FreshRSS_slider_scrollTop', slider.scrollTop);
```
Alkarex, line 397: Do we always have access and the right to write in sessionStorage, or should there be a test/catch?

Inverle: Try/catch is not needed in our case: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage#exceptions

Also, we already use the same code for `FreshRSS_sidebar_scrollTop`
https://github.com/FreshRSS/FreshRSS/blob/44b5c132b82c5558f6193681944d820012c902b1/p/scripts/main.js#L2397-L2401
https://github.com/FreshRSS/FreshRSS/blob/44b5c132b82c5558f6193681944d820012c902b1/p/scripts/main.js#L1071-L1076

PR #8981, opened 2026-07-04
Fever API: log the client IP address on authentication failure

### LINKED ISSUE #6814: [Feature] Fever API: Log authentication failures for e.g. fail2ban

On authentication failures to the fever API, `{"api_version":3,"auth":0}` is returned with a HTTP `200` status code, in nginx logs e.g.:

```
$IP - - $DATE "POST /api/fever.php?api HTTP/2.0" 200 26 "-" "$UA"
```

This makes it impossible for tools like fail2ban to detect and block repeated login failures from the webserver logfile, posing a security risk.

The greader API in turn already returns a `401`:

```
$IP - - $DATE "POST /api/greader.php/accounts/ClientLogin HTTP/3.0" 401 15 "-" "$UA"
```

If possible, return a `401` HTTP status code on fever API authentication failures, which can be detected from webserver logfiles.
The [original fever API spec](https://web.archive.org/https://feedafever.com/api) does not mentions HTTP codes.

For me, it would also be fine to be able to specifically disable the fever API via the config file, as only the greader API is used on my instance.
From my understanding, the current config variable `api_enabled` affects _both_ APIs. (Btw, the [description](https://github.com/FreshRSS/FreshRSS/blob/edge/config.default.php#L70) in the sample config file currently implies only the greader API exists.)

As a quick and dirty solution, I currently delete `p/api/fever.php` on every upgrade.

JamBalaya56562:
Closes #6814

A failed Fever API authentication returns `{"api_version":4,"auth":0}` with HTTP **200**, so from the web-server log a failure is indistinguishable from a success and fail2ban cannot detect/ban repeated attempts (#6814).

The failure was already logged to the API log (`data/users/_/log_api.txt`) in `p/api/fever.php`, but **without the client IP**. This appends the remote IP to that warning, using the existing `Minz_Request::connectionRemoteAddress()` helper (same format as `app/Controllers/authController.php`), so a fail2ban filter watching the FreshRSS API log can extract and ban the address.

Example log line now:

```
[warning] --- Fever API: wrong credentials! deadbeef… ; Remote IP address=203.0.113.5
```

1. Install FreshRSS with the API enabled and create a user with an API password.
2. Send a wrong Fever key: `curl -X POST 'https://your-instance/api/fever.php?api' -d 'api_key=deadbeefdeadbeefdeadbeefdeadbeef'` → response `{"auth":0}`.
3. Check `data/users/_/log_api.txt` — the "wrong credentials" warning now includes `Remote IP address=<your IP>`.

Verified locally with the `freshrss/freshrss:alpine` image (Apache): the log line correctly includes the client IP (`::1`). `php -l`, `phpcs` and `phpstan` all pass.

## Pull request checklist

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated (N/A — no docs change; log message only)

### DIFF
--- a/p/api/fever.php
+++ b/p/api/fever.php
@@ -188,7 +188,8 @@ final class FeverAPI
 				Minz_Log::error('Fever API: Please reset your API password!');
 				Minz_User::change();
 			}
-			Minz_Log::warning('Fever API: wrong credentials! ' . $feverKey, API_LOG);
+			Minz_Log::warning('Fever API: wrong credentials! ' . $feverKey .
+				' ; Remote IP address=' . Minz_Request::connectionRemoteAddress(), API_LOG);
 		}
 		return false;
 	}


Frenzie:
> A failed Fever API authentication returns {"api_version":4,"auth":0} with HTTP 200

Mind, I'm not offering an opinion on the logging (should be fine?), but is it supposed to be 200 as opposed to e.g. 401 or is that a bug?

JamBalaya56562:
Good question — the HTTP 200 is intentional and spec-compliant, not a bug. The Fever API ([feedafever.com/api](https://feedafever.com/api)) always responds with HTTP 200 and signals authentication via the JSON `"auth"` field (`0`/`1`); Fever clients check that field, not the HTTP status. `fever.php` does use proper codes where appropriate (503, 501), and our other API (GReader) returns 401 on auth failure — so this is Fever-protocol-specific by design, and returning 401 here would break Fever clients.

And that 200-on-failure is exactly why this PR is useful: the failure is indistinguishable from success in the web-server log, so fail2ban can't act on it — logging it (with the IP) server-side is the workaround.

Frenzie (approved):
Alright.

PR #8982, opened 2026-07-04
OPML: include each feed's refresh interval (TTL) in export/import

### LINKED ISSUE #5914: [Feature]refresh time included in export
I have a lot of feeds (hundreds).
If I want to move my feeds to another instance of freshrss, then I have to manually set the refresh interval of every feed.
It would be herculean effort to do it. 

add a attribute to every feed portion of export for update interval of that feed to be imported to the freshrss.

nothing really comes to mind.

**Additional context**

JamBalaya56562:
Closes #5914

Moving feeds between FreshRSS instances via OPML lost each feed's **refresh interval (TTL)**, forcing users to reconfigure it manually per feed. This adds a `frss:ttl` attribute to the OPML export (under the existing `https://freshrss.org/opml` namespace) and reads it back on import, so the interval survives a round-trip.

- **Export** (`app/views/helpers/export/opml.phtml`): write `frss:ttl` when the feed's TTL is not the default. The value is the *signed* TTL, so a negative value also carries the muted state.
- **Import** (`app/Services/ImportService.php`): read `frss:ttl` and apply it via `FreshRSS_Feed::_ttl()`, which decodes the sign (restoring both the interval and the muted state).
- **Docs** (`docs/en/developers/OPML.md`): document the new attribute.

A default (non-muted) feed omits the attribute, so it keeps the importing instance's own default.

1. Set a custom refresh interval on a feed (and optionally mute another), then export your subscriptions — the OPML now contains `frss:ttl="…"` for those feeds.
2. Import that OPML on another instance — the feeds keep their refresh interval (and muted state).

Verified locally with `freshrss/freshrss:alpine`: importing an OPML with `frss:ttl="3600"` and `frss:ttl="-1800"` stores `ttl = 3600` and `-1800` in the DB, and re-exporting produces `frss:ttl="3600"` / `frss:ttl="-1800"` (full round-trip). `php -l`, `phpcs` and `phpstan` all pass.

## Pull request checklist

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard) — verified via an import→DB→export round-trip
- [x] documentation updated

### DIFF
--- a/app/views/helpers/export/opml.phtml
+++ b/app/views/helpers/export/opml.phtml
@@ -55,6 +55,11 @@ function feedsToOutlines(array $feeds, bool $excludeMutedFeeds = false): array {
 			$outline['frss:unicityCriteriaForced'] = 'true';
 		}
 
+		// Refresh interval (TTL), signed: a negative value indicates a muted feed
+		if ($feed->ttl(raw: true) !== FreshRSS_Feed::TTL_DEFAULT) {
+			$outline['frss:ttl'] = (string)$feed->ttl(raw: true);
+		}
+
 		if ($feed->kind() === FreshRSS_Feed::KIND_HTML_XPATH || $feed->kind() === FreshRSS_Feed::KIND_XML_XPATH) {
 			/** @var array<string,string> */
 			$xPathSettings = $feed->attributeArray('xpath') ?? [];
--- a/app/Services/ImportService.php
+++ b/app/Services/ImportService.php
@@ -195,6 +195,11 @@ class FreshRSS_Import_Service {
 				$feed->_attribute('unicityCriteriaForced', true);
 			}
 
+			if (isset($feed_elt['frss:ttl'])) {
+				// Signed refresh interval (TTL); a negative value indicates a muted feed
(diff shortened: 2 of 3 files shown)


app/views/helpers/export/opml.phtml
```diff
@@ -55,6 +55,11 @@ function feedsToOutlines(array $feeds, bool $excludeMutedFeeds = false): array {
 			$outline['frss:unicityCriteriaForced'] = 'true';
 		}
 
+		// Refresh interval (TTL), signed: a negative value indicates a muted feed
+		if ($feed->ttl(true) !== FreshRSS_Feed::TTL_DEFAULT) {
+			$outline['frss:ttl'] = (string)$feed->ttl(true);
```
Alkarex:
```suggestion
			$outline['frss:ttl'] = (string)$feed->ttl(raw: true);
```

Let's use the parameter name when the meaning is not obvious

JamBalaya56562: Done, thanks — switched to the named `raw: true` parameter (both the condition and the assignment).

Alkarex (approved):

PR #8983, opened 2026-07-04
Dynamic OPML: optionally reproduce the remote OPML's categories

### LINKED ISSUE #6049: [Feature] Allow Dynamic OPML to create categories
I had this weird approach - that I'm going to use a single .opml file as the "single source of truth" for all my feeds. That way, if I ever want to switch my rss feed app, I can just take my .opml with me and never have to worry about exporting my feeds.

I installed freshrss after learning that it has the "dynamic OPML" feature. From the surface, it looked like exactly what I wanted - I'll just have my .opml file in my dotfiles, I'll edit **the .opml file by hand whenever I want to add a new rss feed** (ie. I won't be using freshrss' webui to add feeds), and then freshrss will sync the newly changed .opml to the freshrss rss feed database and add the necessary feeds.

However, I noticed that there are two ways of using opml in FreshRSS:
1. **Import** the .opml **file** in `i/?c=importExport`
2. Set up dynamic OPML to periodically sync with an OPML hosted on HTTP (this is what I wanted to use)

The problem is that point 1 will create categories, if the .opml is like
```xml
<?xml version="1.0" encoding="UTF-8"?>
<opml version="2.0">
    <head>
        <title>OPML</title>
    </head>
<body>
    <outline text="my favorite german wheel companies">
        <outline text="wheelkommen" type="rss" xmlUrl="https://wheelkoommen.de/rss/"></outline>
    </outline>
</body>
</opml>
```
however if you pass the very same .opml file (hosted on some website) to dynamic OPML, it will lump all categories together into whatever category you set in the "add dynamic opml" form, in the "Category" input

Add a "Create categories that are contained in this .opml file ("Category" will be used as prefix)" checkbox in the "Add dynamic OPML" form. If it's checked, it should also create prefixed categories for the categories contained in the .opml file. If it's unchecked, lump all feeds into one huge category (current behavior)

If the category in the .opml file is called e.g. "mycategory", then whatever was passed in the "Category" field in the "Add dynamic OPML" form should be used as prefix, e.g. if you wrote "my-friends-opml" then it'll be "my-friends-opml-mycategory"

I could just have multiple .opml files for each category in my dotfiles, and sync each one. Would be more comfortable to have just one, though.

**Additional context**
N/A

JamBalaya56562:
Closes #6049

A **Dynamic OPML** category previously placed every feed from the remote OPML into a single category, ignoring the remote OPML's own category structure. This adds an opt-in per-category option, **“Reproduce the categories from the remote OPML”**, so a remote OPML can be used as a source of truth including its categories.

When enabled:
- Categories are created from the remote OPML, named as in the OPML (existing categories with the same name are reused); feeds that are not in any category go into the Dynamic OPML (anchor) category.
- Feeds added by this Dynamic OPML are tagged with a `dynamic_opml` attribute, so the add / mute-on-removal / unmute-on-return sync only ever touches the feeds this Dynamic OPML manages — feeds you added manually to the same categories are never affected.

When disabled (the default), the behavior is unchanged (every feed goes into the single category).

Category creation reuses the existing OPML importer (`FreshRSS_Import_Service::importOpml()`), which already builds categories from the OPML structure; a new `importedCategories()` getter exposes that structure to the Dynamic OPML sync.

1. Add a Dynamic OPML category, tick “Reproduce the categories from the remote OPML”, and point it at a remote OPML that groups feeds into categories.
2. Refresh — the OPML's categories are reproduced and populated; feeds outside any category land in the Dynamic OPML category.
3. Remove a feed from the remote OPML and refresh — it is muted; feeds you added by hand to those categories are left alone.

Verified locally with `freshrss/freshrss:alpine`: an OPML with a `Tech` category plus a root feed creates the `Tech` category and places feeds correctly (tagged), removing a feed mutes it, and a manually-added feed in `Tech` is never muted. `php -l`, `phpcs` and `phpstan` all pass. The new i18n key was propagated to all languages via `cli/manipulate.translation.php --action format`.

## Pull request checklist

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard) — verified via an end-to-end Docker test of the create/sync/mute paths
- [x] documentation updated

### DIFF
--- a/README.fr.md
+++ b/README.fr.md
@@ -233,21 +233,21 @@ Voir le [dépôt dédié à ces extensions](https://github.com/FreshRSS/Extensio
 | Deutsch (de) | ￭￭￭￭￭￭￭￭￭･ 98% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fde+%2F%28TODO%7CDIRTY%29%24%2F) |
 | Ελληνικά (el) | ￭￭￭￭￭￭￭･･･ 75% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fel+%2F%28TODO%7CDIRTY%29%24%2F) |
 | English (en) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen+%2F%28TODO%7CDIRTY%29%24%2F) |
-| English (United States) (en-US) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen-US+%2F%28TODO%7CDIRTY%29%24%2F) |
+| English (United States) (en-US) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen-US+%2F%28TODO%7CDIRTY%29%24%2F) |
 | Español (es) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fes+%2F%28TODO%7CDIRTY%29%24%2F) |
 | فارسی (fa) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffa+%2F%28TODO%7CDIRTY%29%24%2F) |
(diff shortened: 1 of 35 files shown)


Alkarex:
I am a bit unsure about that one. We need to carefully ponder the consequences

JamBalaya56562:
@Alkarex To make this easier to weigh, I went through the behaviour and listed every consequence I could identify. The option is opt-in per Dynamic OPML category and defaults to off, so nothing changes for existing users unless it is ticked.

| Behaviour | Safety property |
|---|---|
| The remote OPML can create categories | Bounded by the existing `max_categories` limit; a warning is logged and the sync stops when it is reached |
| A category with the same name is reused, not duplicated | Only feeds tagged `dynamic_opml = <category id>` are ever muted/unmuted, so feeds you added by hand to that category are never touched |
| Feeds removed from the remote OPML | Muted, never deleted (unchanged from the current behaviour) |
| Categories that become empty | Never deleted |
| Feeds that already exist in FreshRSS | Never moved between categories (`addFeedObject()` does not update `category` for an existing feed) |

Two points where I would rather have your call than guess:

1. **Enabling the option on an existing Dynamic OPML.** Because existing feeds are never moved, the first sync tags them and starts managing them, but they stay in the anchor category; only feeds that appear later are filed into the OPML's categories. I kept it conservative on purpose — silently relocating feeds a user may have organised themselves felt worse than doing nothing — but it does mean the feature only fully applies to new feeds. Would you prefer the sync to re-file the feeds it manages?

2. **Turning the option back off.** The managed set reverts to "the feeds in this category", so feeds previously filed into other categories keep their tag but are no longer muted/unmuted by this Dynamic OPML. I can make the off-switch symmetric if you prefer.

If the category-creation part is the uncomfortable one, I am happy to reduce the scope — e.g. only reuse categories that already exist and never create new ones — or to drop it entirely if you conclude it does not belong in core.

(The branch is rebased on current `edge` and CI is green.)

PR #8984, opened 2026-07-05
Notifications: pluralize the "new articles" count message

### LINKED ISSUE #6470: [Bug] There ARE 1 new articles

JamBalaya56562:
## Problem

The desktop notification used the plural form even for a single article ("There are 1 new articles…"). Closes #6470.

## Approach

After discussion this now uses **@Frenzie's approach from #8988**: instead of any client-side plural logic, the client sends the previous unread count to the server, PHP computes the new-article count and builds the properly pluralized notification body via `Minz_Translate::plural()`, and the client just displays the returned string.

- `app/views/javascript/nbUnreadsPerFeed.phtml` returns a ready-made `notifBody` (using the `%d`-plural `body_new_articles` / `body_unread_articles`).
- `javascript_vars.phtml` no longer needs to expose those two strings to JS.
- `p/scripts/main.js` sends `previous_unread` and shows `json.notifBody`.

Result: correct per-language plurals with no JS plural machinery, and the existing translations are kept (marked `// DIRTY` for translators to refine the forms).

Credit to @Frenzie — this mirrors #8988.

## Checklist

- [x] `composer run-script translations` is a no-op (translation syntax check is green)
- [x] `php -l` passes
- [ ] Tests — n/a


Frenzie:
> The count (nb_new) is computed client-side in p/scripts/main.js (nb_unreads - nb_unreads_before) and the notification is rendered there, so the PHP Minz_Translate::plural() helper cannot be used (PHP never sees the count), and there is no JS plural mechanism.

Tell the AI to skip all of the enormous complexity and just to pass the message string.

JamBalaya56562:
Good call, thanks — simplified. The follow-up commit drops the client-side plural machinery entirely and just uses a single, count-neutral message string (`New articles to read on FreshRSS: %d`), so "There are 1 new articles" no longer happens for any count. The net diff is now a one-line string change, and reverting the rest also restores the original per-language translations. Happy to adjust the wording if you'd prefer something different.

Frenzie:
That could be an option, but it's not quite what I meant. I figured it'd be easier to just show it: #8988.

JamBalaya56562:
Thanks a lot — much cleaner, and now I see exactly what you meant: send the previous count to the server and build the string in PHP. I have adopted your #8988 approach here (credit to you) as a ready, non-draft version in case it helps move it forward.

One tiny thing I noticed while going through it: the new `en-US` entries are marked `// TODO`, but since they are identical to `en`, running `cli/manipulate.translation.php --action ignore_unmodified --language en-US` would mark them `// IGNORE` per the i18n instructions.

Since #8988 is your own and essentially the same, I am happy to close this in favour of #8988 — just let me know which you would prefer to proceed with.

Frenzie:
Alright, we'll continue with my take on it. 👍

PR #8985, opened 2026-07-05
Feeds tree: sort feed names with locale-aware collation

### LINKED ISSUE #8300: [Bug] Order of feeds, both tree and by title
### Describe the bug

Hi:

I've noticed something a bit strange. As this is a “dev” version, I'd like to point out that I'm seeing an inconsistency between the order of the feeds, both in the tree where the feeds are located and in the order of the feed titles.

I will put the screenshot on what I indicate, because what I collect is in Spanish, and therefore, I understand that one to another they come out in different order (only that, for me, the order in the tree is incorrect.

Image

As we can see, in the tree, the two Spanish words appear above the “Ú”, which, as we can see, is placed last. Just this, by having it as “Feeds title A+Z”.

I understand that it's a matter of time and detecting this little bug that I'm giving away.

### To Reproduce

1. And only one category.
2. Go to the category.
3. Not necessarily.
4. It's not a mistake. I understand that it's a problem with the order of the feed titles.

### Expected behavior

I was hoping that the feed title would match, but this happens to me in many feeds, not just this one.

### FreshRSS version

1.27.2-dev

### System information

- Database version: PostgreSQL 18
- PHP version: 8.4
- Installation type: git
- Web server type: nginx
- Device: PC (Linux)
- OS: Debian 14 (Unstable)
- Browser: Firefox 146.0.

### Additional context

I expected that it would have the preferred order, not only for accents, but also for symbols. This is not the case.

JamBalaya56562:
## Problem

The feed **tree** (left sidebar) and the **feed-title view** (main body) order feeds differently for names with accents or symbols. As reported in #8300, a Spanish feed starting with `Ú` appears **last** in the tree, after every ASCII-named feed.

The tree is sorted in PHP by `FreshRSS_Category::sortFeeds()` with `strnatcasecmp()`, which compares raw UTF-8 **bytes** — so any name starting with a non-ASCII character (`0xC3…` for `Á`, `Ñ`, `Ú`, …) sorts after `A–Z`. The main view is sorted by the database according to its collation, which orders accented letters near their base letter — hence the mismatch (as @Alkarex noted in the issue).

Closes #8300.

## Change

`sortFeeds()` now compares with a `Collator` (`ext-intl`, already a hard requirement) for the user's language, falling back to the previous `strnatcasecmp()` if a collator cannot be created:

```php
$collator = \Collator::create(FreshRSS_Context::hasUserConf() ? FreshRSS_Context::userConf()->language : '');
```

Sorting a Spanish set, before / after:

```
old: banana, Manzana, nube, Zorro, Ábaco, Ñandú, Úlcera, árbol
new: Ábaco, árbol, banana, Manzana, nube, Ñandú, Úlcera, Zorro
```

This makes the tree order locale-aware and much closer to the main view. Exact parity with the main view still depends on the database collation (which is admin/DB-specific and cannot be reproduced generically in PHP), but the "accented names sort last" problem is gone.

## Checklist

- [x] Code follows the project style (`php -l`, `phpcs`, `phpstan` pass)
- [ ] Tests added — n/a: a one-line comparator change with no server-side branch worth a fixture (verified manually — see the before/after above)

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -299,7 +299,18 @@ class FreshRSS_Category extends Minz_Model {
 		if ($this->feeds === null) {
 			return;
 		}
-		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b) => strnatcasecmp($a->name(), $b->name()));
+		// Sort locale-aware so accented and non-ASCII feed names sort near their base letter
+		// instead of byte-wise: strnatcasecmp() compares raw UTF-8 bytes, which pushes names
+		// starting with a non-ASCII character (e.g. an accented capital) to the very end.
+		// Without a user language (e.g. CLI), keep the previous byte-wise sort and avoid
+		// Collator::create(''), whose ordering depends on the environment default locale.
+		$language = FreshRSS_Context::hasUserConf() ? FreshRSS_Context::userConf()->language : '';
+		$collator = $language === '' ? null : \Collator::create($language);
+		if ($collator === null) {
+			uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => strnatcasecmp($a->name(), $b->name()));
+			return;
+		}
+		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => (int)$collator->compare($a->name(), $b->name()));
 	}
 
 	/**
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -299,17 +299,16 @@ class FreshRSS_Category extends Minz_Model {
 		if ($this->feeds === null) {
 			return;
 		}
-		// Sort locale-aware with Collator so accented/non-ASCII names sort near their base
-		// letter instead of byte-wise (strnatcasecmp pushes them to the end). Fall back to
-		// strnatcasecmp without a user language (e.g. CLI) or when the intl extension is
-		// missing — class_exists() avoids a fatal "class not found" (intl is not required).
+
+		// Sort locale-aware with Collator if available
 		$language = FreshRSS_Context::hasUserConf() ? FreshRSS_Context::userConf()->language : '';
 		$collator = ($language === '' || !class_exists('Collator')) ? null : \Collator::create($language);
-		if ($collator === null) {
-			uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => strnatcasecmp($a->name(), $b->name()));
+		if ($collator !== null) {
+			uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => (int)$collator->compare($a->name(), $b->name()));
 			return;
 		}
-		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => (int)$collator->compare($a->name(), $b->name()));
+
+		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => strnatcasecmp($a->name(), $b->name()));
 	}
 
 	/**
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -299,16 +299,7 @@ class FreshRSS_Category extends Minz_Model {
 		if ($this->feeds === null) {
 			return;
 		}
-
-		// Sort locale-aware with Collator if available
-		$language = FreshRSS_Context::hasUserConf() ? FreshRSS_Context::userConf()->language : '';
-		$collator = ($language === '' || !class_exists('Collator')) ? null : \Collator::create($language);
-		if ($collator !== null) {
-			uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => (int)$collator->compare($a->name(), $b->name()));
-			return;
-		}
-
-		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => strnatcasecmp($a->name(), $b->name()));
+		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b): int => FreshRSS_Context::localeCompare($a->name(), $b->name()));
 	}
 
 	/**
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -299,14 +299,10 @@ class FreshRSS_Category extends Minz_Model {
 		if ($this->feeds === null) {
 			return;
 		}
-		// Sort locale-aware so accented and non-ASCII feed names sort near their base letter
-		// instead of byte-wise: strnatcasecmp() compares raw UTF-8 bytes, which pushes names
-		// starting with a non-ASCII character (e.g. an accented capital) to the very end.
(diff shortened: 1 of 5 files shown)


app/Models/Category.php
```diff
@@ -299,7 +299,18 @@ private function sortFeeds(): void {
...
 			return;
 		}
-		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b) => strnatcasecmp($a->name(), $b->name()));
+		// Sort locale-aware so accented and non-ASCII feed names sort near their base letter
+		// instead of byte-wise: strnatcasecmp() compares raw UTF-8 bytes, which pushes names
+		// starting with a non-ASCII character (e.g. an accented capital) to the very end.
+		// Without a user language (e.g. CLI), keep the previous byte-wise sort and avoid
+		// Collator::create(''), whose ordering depends on the environment default locale.
+		$language = FreshRSS_Context::hasUserConf() ? FreshRSS_Context::userConf()->language : '';
+		$collator = $language === '' ? null : \Collator::create($language);
```
Frenzie: Is it possible for that `Collator::create()` call to fail if you don't have the rules for that language installed or some such? I mean like how mbstring is currently considered preferable but optional.

Edit: partially never mind, also null:

> Return new instance of [Collator](https://www.php.net/manual/en/class.collator.php) object, or [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) on error.

https://www.php.net/manual/en/collator.create.php

But the question remains the same, since it'd potentially affect the listed requirements.

JamBalaya56562: Good catch. `\Collator::create()` itself returns `null` on a bad/unknown locale, and that case is already handled by the fallback to `strnatcasecmp()`. But you're right that it touches the requirements: if the `intl` extension isn't installed at all, the `\Collator` class doesn't exist and calling `::create()` throws a fatal "Class not found" — which would effectively promote `intl` from *recommended* to *required*.

Rather than change the requirements, I've guarded the call with `class_exists('Collator')`, so installs without `intl` simply keep the previous byte-wise sort. `intl` stays optional (same as `mbstring`), and the locale-aware ordering is a progressive enhancement when it's available — so the listed requirements are unaffected.

app/Models/Category.php
```diff
@@ -299,7 +299,21 @@ private function sortFeeds(): void {
...
 		}
-		uasort($this->feeds, static fn(FreshRSS_Feed $a, FreshRSS_Feed $b) => strnatcasecmp($a->name(), $b->name()));
+		// Sort locale-aware so accented and non-ASCII feed names sort near their base letter
+		// instead of byte-wise: strnatcasecmp() compares raw UTF-8 bytes, which pushes names
+		// starting with a non-ASCII character (e.g. an accented capital) to the very end.
+		// Without a user language (e.g. CLI), keep the previous byte-wise sort and avoid
+		// Collator::create(''), whose ordering depends on the environment default locale.
+		// The intl extension (\Collator) is only recommended, not required, so guard for it
+		// with class_exists(): without it, keep the previous byte-wise sort instead of a
+		// fatal "class not found", so the locale-aware ordering stays a progressive enhancement.
```
Frenzie: The comment is a bit excessive, otherwise lgtm.

Something like this perhaps:
```suggestion
		// Sort locale-aware with Collator::create('') when intl is available 
		// instead of byte-wise with strnatcasecmp(). Without class_exists()
		// you'd end up with a fatal "class not found" error.
```

JamBalaya56562: Agreed, thanks — trimmed it to four lines. I kept the wording slightly different from the suggestion: it mentions `class_exists()` as you proposed, but avoids saying `Collator::create('')`, since the code deliberately does *not* call `Collator` with an empty locale (an empty language falls back to `strnatcasecmp()` to keep ordering deterministic). Pushed as a separate commit.

Frenzie (approved):

Alkarex:
Please search `uasort` in the codebase. I believe there are a few other related instances

JamBalaya56562:
Thanks, I searched for `uasort` usages in the codebase. I found a few related instances:

- `app/Models/FeedDAO.php`: sorts `FreshRSS_Feed` names with `strnatcasecmp()`. This looks directly related to the feed-name collation issue addressed here.
- `app/Models/CategoryDAO.php`: sorts category names. This may be related if we want category names in the feed tree to follow the same locale-aware ordering.
- `app/Models/Share.php`: sorts sharing service names with `strcasecmp()`. This is similar in principle, but less directly related to the feed tree.

Would you like me to update the related feed/category sorting in this PR as well, or keep this PR limited to `Category::sortFeeds()` and handle the other instances separately?

Alkarex:
I have made a proposition. Please double-check

JamBalaya56562:
Thanks, I double-checked the proposition and pushed an extra commit: 925686776b31.

That follow-up addresses two concerns that came up during the re-check:

- enable `Collator::NUMERIC_COLLATION`, so names such as `Feed 2` / `Feed 10` keep the natural ordering behaviour we previously got from `strnatcasecmp()`;
- rebuild the cached `Collator` when the user language changes, so long-lived or multi-user CLI runs do not keep using the first user's locale.

It also falls back to `strnatcasecmp()` if `Collator::compare()` itself returns `false`.

app/Models/Context.php
```diff
@@ -667,4 +667,30 @@ public static function defaultTimeZone(): string {
 		$timezone = ini_get('date.timezone');
 		return $timezone != false ? $timezone : 'UTC';
 	}
+
+	/**
+	 * Sort locale-aware with Collator if available
+	 */
+	public static function localeCompare(string $a, string $b): int {
+		static $collatorLanguage = null;
```
Alkarex: No change needed, but if this information is actually needed, an alternative could be `$collator->getLocale(Locale::VALID_LOCALE)`

app/Models/Context.php
```diff
@@ -667,4 +667,30 @@ public static function defaultTimeZone(): string {
...
+		static $collator = null;
+		$language = FreshRSS_Context::hasUserConf() ? FreshRSS_Context::userConf()->language : '';
```
Alkarex: Although small, I am worried of the performance reduction with this test, since this code is a callback function used a lot.
Could you check whether there is actually a case when the user language change during a request, i.e. when it makes sense to have this test here?

app/Models/Context.php
JamBalaya56562: Thanks. I addressed this in e506bd0bddba by removing the separate cached `$collatorLanguage` state altogether. `localeCompare()` now initialises the `Collator` once and no longer keeps a parallel language variable to compare against, so there is no extra locale information to track here.

app/Models/Context.php
JamBalaya56562: I checked the core request flow and I do not see a normal Web request where the user language changes and then the same request needs this comparator again. The display configuration action can update `FreshRSS_Context::userConf()->language`, but it redirects immediately after saving, so the next request starts with the new language.

The language-change guard was therefore mostly defensive, while this comparator can be called many times by `uasort()`. In e506bd0bddba I changed it so the language is read only when the cached `Collator` is first initialised, avoiding the per-comparison lookup/check while keeping numeric collation and the `strnatcasecmp()` fallback.

Alkarex (approved):

PR #8990, opened 2026-07-05
Add open link behavior options for articles and links

### LINKED ISSUE #1256: Open article in new background tab shortcut
Hello,
first thing - I have moved to FreshRSS from ttrss and never looked back, it fantastic piece of software.

My workflow in FRSS consist usually of browsing new articles, opening the more interesting ones in separate tab, and then closing FRSS and reading all of them. Right now, I can use j/k to browse, but space opens selected article in foregroud and loses focus on FRSS - I have to use the mouse to open article in the background, being able to further browse the feed.

I would love to see the possibility to have keyboard shortcut to open selected article in the background, so I could use FRSS totally without the mouse.

Inverle:
Closes https://github.com/FreshRSS/FreshRSS/issues/1256 (Safari not tested/implemented)

Solves it by:
* Allowing use of <kbd>Ctrl/Shift</kbd> keys while using the open website shortcut, this will cause the article to open in a new background tab, only works on Firefox though.
* Adding an option for default link open behavior, `Open in new background tab`, in this case, which can be applied either only to the shortcut or both shortcut and button. Also works only on Firefox.

I tried to make the option work on Chromium too, but the method I found for doing that was patched recently in v150:

 
image

https://gist.github.com/Inverle/d79be2c30ed9b5b2aba8e09c507a4ea9
new: https://gist.github.com/Inverle/27dee6e393e2d165e429f0464cccf83f

(and when I found it, I was running an outdated version of Chromium, so this is quite unlucky - the first stable build with the fix was released 2026-06-30, so basically 5 days ago..)

edit: I found a new way, I'm not sure how easy it will be to integrate though

In addition:
* Options `Open in new tab` (already default) / `Open in current tab`
* Setting for if the option should be also applied to external links in articles


Inverle:
> I tried to make the option work on Chromium too, but the method I found for doing that was patched recently in v150:

This could be one of the commits which fixed it:
https://chromium.googlesource.com/chromium/src.git/+/aa69447f800b764e509b2cb1cce83f006df4db0b%5E%21/

The bug is privated though: https://issues.chromium.org/issues/503371806

edit: I found a new way, I'm not sure how easy it will be to integrate though

Frenzie:
> Inverle requested review

This kind of behavior isn't something browsers are very fond of due to abuse in the past, but I imagine you already considered that before writing the code. Are you asking for an opinion on something specific? The JS looks good to me as an implementation of what's described in the  OP.

Alkarex:
As this is a bit non-standard, any chance this could be in an extension?

Inverle:
> As this is a bit non-standard, any chance this could be in an extension?

The `Open in new background tab` option is indeed a bit fragile, or more precisely, Chrome has a problem with it. But: 

> Allowing use of Ctrl/Shift keys while using the open website shortcut

seems like a fair change, as well as the *Open in current tab* option, so what exactly should be in an extension?

Frenzie:
It seems somewhat unlikely that the synthetic event will keep working that way in Firefox in the long term.

Alkarex:
I am afraid those mechanisms have been a cat/mouse game for some time and will likely continue with some instability. In particular, if possible, I would prefer code requiring user-agent detection to be in an extension.

Inverle:
> I am afraid those mechanisms have been a cat/mouse game for some time and will likely continue with some instability. In particular, if possible, I would prefer code requiring user-agent detection to be in an extension.

It doesn't require user agent detection, it's just for clarity. I can remove it, and the code will continue to fallback to opening in a new tab (not in background if not supported)

Inverle:
> It seems somewhat unlikely that the synthetic event will keep working that way in Firefox in the long term.

So it is a bug in Firefox, but maybe an intended one.
I was trying to see if this worked in older Firefox, and it only started working in Firefox 96.
In the changelog (https://www.firefox.com/en-US/firefox/96.0/releasenotes/) you can see this note:
> On macOS, command-clicking links in Gmail now opens them in a new tab as expected.

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1739929

The code change that was made: https://phabricator.services.mozilla.com/D131868#differential-review-stage

Potentially, the unintended part is that now untrusted events can trigger such clicks (with Ctrl key), but it is not checked if Ctrl is actually being pressed, unlike in Chrome.

This is working for 4 years now though.

Alkarex:
> command-clicking links in Gmail now opens them in a new tab as expected

It looks like the problem was about opening in a new tab, not the background part, was it? In our case, isn't it solely the background part that does not exist?

Frenzie:
Apparently they're no rush (16 years and counting), but they didn't reject it… https://bugzilla.mozilla.org/show_bug.cgi?id=565104

Inverle:
I made a minimal PR only for the modifiers change: https://github.com/FreshRSS/FreshRSS/pull/9186

I will probably turn the rest of this PR into an extension, as suggested: https://github.com/FreshRSS/FreshRSS/pull/8990#issuecomment-4902109852

PR #8997, opened 2026-07-07
Fix article hover date visibility

### LINKED ISSUE #8994: [Bug] Article's date is hidden when hovering over the title
### Describe the bug

Looks like it appeared since a fairly new commit : hovering over an article's title hides its date. Happens on origine-compact and verified it also happens on another theme (Ansum).

### To Reproduce

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

### Expected behavior

_No response_

### FreshRSS version

up-to-date edge

### System information

- Database version: MariaDB 10.11.14
- PHP version: PHP 8.3
- Installation type: git
- Web server type: Apache
- Device: Computer
- OS: Win 11
- Browser: Chrome 150

### Additional context

_No response_

JamBalaya56562:
## Summary

- Keep the article title in its own grid area on hover when the article row also shows a date.
- Restore a small spacing between the hovered title and the date.
- Regenerate the RTL stylesheet counterpart.

Fixes #8994

## Screenshots

image

The comparison uses a synthetic hover fixture with FreshRSS article-row DOM and theme CSS, so no local feed data is included.

## Testing

- `npm run rtlcss`
- `npx stylelint p/themes/base-theme/frss.css p/themes/base-theme/frss.rtl.css`
- Visual verification with a synthetic hover fixture using the FreshRSS article-row DOM and theme CSS for Origine-compact and Ansum.

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -1511,7 +1511,7 @@ a.website:hover .favicon {
 }
 
 .flux:not(.current):hover .flux_header.has-date .item .title {
-	grid-column: 1 / -1;
+	padding-right: 0.5rem;
 	z-index: 2;
 }
 
--- a/p/themes/base-theme/frss.rtl.css
+++ b/p/themes/base-theme/frss.rtl.css
@@ -1511,7 +1511,7 @@ a.website:hover .favicon {
 }
 
 .flux:not(.current):hover .flux_header.has-date .item .title {
-	grid-column: 1 / -1;
+	padding-left: 0.5rem;
 	z-index: 2;
 }
 


Inverle:
Maybe a regression from https://github.com/FreshRSS/FreshRSS/pull/8824?

JamBalaya56562:
Thanks, yes, I checked #8824 and this does look like the regression source.

That PR converted the article-row title/date/summary layout to CSS Grid and changed the truncated-title hover behavior from the previous small directional padding to `grid-column: 1 / -1`. In rows with `.has-date`, that makes the title span across the date column on hover, which is why the date can be covered by the long title.

This PR keeps the grid layout from #8824, but only removes that hover-time column spanning and restores the small padding so the title remains in its own grid area while the date stays visible.

Alkarex:
Thanks! This has annoyed me a bit for some time :-)

Alkarex (approved):

Alwaysin:
Thanks 🙏

Frenzie:
@Alkarex Could you clarify what the issue is supposed to be here? I don't see any.

JamBalaya56562:
@Frenzie The screenshot in the description had gone missing — it pointed at a fork branch that has since been deleted — which is probably why the PR reads as being about nothing. I have replaced it with a fresh before/after.

Both captures are the same `edge` tree with the same articles; the only difference is the one line this PR changes. Before it, `grid-column: 1 / -1` makes the hovered title span every grid column, so it runs across the date column and hides it (#8994). After it, the title stays inside its own grid area and the date keeps its column.

On your point in #9140: I tested that too, and you are right that the rule is inert on `edge` today. Applying #9140 to `edge` renders pixel-identical (same md5) to `edge` without it — with the grid in place, the title is already confined to `grid-area: title`, so the declaration has nothing left to do. I sanity-checked the harness by exaggerating the new `column-gap` to `6rem`, which does change the output, so the identical result is not a caching artefact.

In other words this PR was a minimal revert of that one line rather than a real fix, and removing the block outright as you propose looks like the better end state.

Method, for reproducibility: a FreshRSS instance with a few deliberately long titles, captured with headless Chromium at 1280 px. The hover state is applied statically by removing `:hover` from the `.flux` rules, so the exact same declarations render without needing a live pointer.

PR #9000, opened 2026-07-07
Minz: Skip loading JS/CSS asset if it's already in the list

Inverle:
Fixes `Uncaught SyntaxError: redeclaration of const freshrssSliderLoadEvent` error shown in console on e.g. `/i/?c=subscription&id=1` due to https://github.com/FreshRSS/FreshRSS/pull/8973

### DIFF
--- a/lib/Minz/View.php
+++ b/lib/Minz/View.php
@@ -225,6 +225,11 @@ class Minz_View {
 		if ($url === '') {
 			return;
 		}
+		foreach (self::$styles as $style) {
+			if (strtolower($style['url']) === strtolower($url)) {
+				return;
+			}
+		}
 		array_unshift(self::$styles, [
 			'url' => $url,
 			'media' => $media,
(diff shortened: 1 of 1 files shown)


lib/Minz/View.php
```diff
@@ -225,6 +225,11 @@ public static function prependStyle(string $url, string $media = 'all', bool $co
 		if ($url === '') {
 			return;
 		}
+		foreach (self::$styles as $style) {
+			if (strtolower($style['url']) === strtolower($url)) {
```
Alkarex: Is the `strtolower` needed?

Inverle: I think there is no point in loading a JS file, for example, with a different casing twice. (that is also exactly what we don't want)

Alkarex: Sure, but I would like to get rid of the overhead if not required, and since we control the call to those functions, it should be fair to leave this concern to the caller

Frenzie: While case sensitive files aren't a thing I'm particularly fond of, the fact is a different case refers to two or more different files which is standard presumed functionality. Blocking that sounds like an antifeature.

Frenzie (changes requested):
Blocking standard behavior is not a good thing, imho.

Frenzie (approved):

PR #9003, opened 2026-07-08
Determine update release channel from git branch, not version string

### LINKED ISSUE #5850: Web update git wrong version message
I also just switched back to latest, but the update still says that it wants to switch to edge:

image

_Originally posted by @Tealk in https://github.com/FreshRSS/FreshRSS/discussions/5729#discussioncomment-7533440_

JamBalaya56562:
Fixes #5850

## Problem

On the admin **Update** page, the "Release channel" row keeps showing **Rolling release ("edge")** — with a link inviting the user to switch to the `edge` branch — even after the user has switched their git checkout back to a stable/`latest` release.

## Root cause

The release channel was derived **only from the `FRESHRSS_VERSION` string**, never from the actual git state:

```php
// before
private function is_release_channel_stable(string $currentVersion): bool {
    return !str_contains($currentVersion, 'dev') && !str_contains($currentVersion, 'edge');
}
```

After a manual `git checkout` to a release, `FRESHRSS_VERSION` (from `constants.php`) can lag the checkout — a between-releases `-dev` string, or PHP opcache still holding the previous `constants.php`. In those cases the check stays `false`, so the UI keeps advertising `edge`. `is_release_channel_stable()` is the single input to the view branch at `app/views/update/index.phtml`.

## Fix

For **git installs**, determine the channel from the checked-out git state. A stable release is checked out **directly on a release tag**, detected with `git tag --points-at HEAD`. Any branch — `edge`/`dev` or a custom development branch — is the rolling channel. Non-git (archive) installs keep the existing version-string heuristic unchanged.

```php
private function is_release_channel_stable(string $currentVersion): bool {
    if (self::isGit()) {
        // Stable = checked out directly on a release tag.
        return self::isOnGitTag();
    }
    return !str_contains($currentVersion, 'dev') && !str_contains($currentVersion, 'edge');
}
```

`isOnGitTag()` runs `git tag --points-at HEAD` — a **live git query**, so it is not affected by a stale `FRESHRSS_VERSION` / opcache (the failure mode #5850 was about). A `gitBranchName(): ?string` helper is kept for the "Release channel" display (`getCurrentGitBranch()`).

> **Review notes (thanks @Frenzie / @Inverle):** an earlier revision used `$branch !== 'edge' && $branch !== 'dev'` (which would treat a custom development branch as stable), then `gitBranchName() === null` (a detached `HEAD` as an indirect proxy for a tag). It now checks the tag directly. I did weigh going back to a `FRESHRSS_VERSION` check extended with `dev`/`alpha`/`beta`/`rc` markers — simpler, and we only ship stable tags — but that reintroduces the reliance on the version constant that was stale in #5850. Happy to switch to that if preferred.

## Behaviour before / after

| State | Actual git state | Before (pre-PR) | After |
|---|---|---|---|
| Release/`latest` (on a tag) | HEAD on a release tag | Rolling release (edge) ❌ | Stable release (latest) ✅ |
| Development checkout | `edge` branch (not a tag) | Rolling release (edge) | Rolling release (edge) (unchanged) |
| Custom branch / raw commit | any non-tag state | — | Rolling channel (**not** treated as stable) |
| Archive (non-git) install | — | version-string based | version-string based (unchanged) |

## Testing

- `php -l app/Controllers/updateController.php` — no syntax errors; `phpcs` / `phpstan` on the changed file — no errors.
- Verified against a real git repo: HEAD on a release tag → stable; on `edge` / a custom branch / a raw detached commit → rolling; non-git → version-string fallback.

Only `app/Controllers/updateController.php` changes; the view and i18n strings are untouched (`admin.update.releaseChannel.edge` / `.latest` already exist).

### DIFF
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -60,14 +60,31 @@ class FreshRSS_update_Controller extends FreshRSS_ActionController {
 		return true;
 	}
 
-	public static function getCurrentGitBranch(): string {
-		$output = [];
-		exec('git branch --show-current', $output, $return);
-		if ($return === 0) {
-			return 'git branch: ' . $output[0];
-		} else {
-			return 'git';
+	/** @return non-empty-string|null */
+	public static function gitBranchName(): ?string {
+		if (!self::isGit() || !function_exists('exec')) {
+			return null;
 		}
+		$cwd = getcwd();
+		if ($cwd !== false) {
+			chdir(FRESHRSS_PATH);
+		}
+		$output = [];
+		$return = 1;
+		exec('git branch --show-current', $output, $return);
+		if ($cwd !== false) {
+			chdir($cwd);
+		}
+		if ($return !== 0) {
+			return null;
+		}
+		$branch = trim(implode('', $output));
+		return $branch === '' ? null : $branch;
+	}
+
+	public static function getCurrentGitBranch(): string {
+		$branch = self::gitBranchName();
+		return $branch !== null ? 'git branch: ' . $branch : 'git';
 	}
 
 	public static function hasGitUpdate(): bool {
@@ -179,6 +196,10 @@ class FreshRSS_update_Controller extends FreshRSS_ActionController {
 	}
 
 	private function is_release_channel_stable(string $currentVersion): bool {
+		if (self::isGit()) {
+			$branch = self::gitBranchName();
+			return $branch !== 'edge' && $branch !== 'dev';
+		}
 		return !str_contains($currentVersion, 'dev') && !str_contains($currentVersion, 'edge');
 	}
 
--- a/app/Controllers/updateController.php
+++ b/app/Controllers/updateController.php
@@ -197,8 +197,10 @@ class FreshRSS_update_Controller extends FreshRSS_ActionController {
 
 	private function is_release_channel_stable(string $currentVersion): bool {
 		if (self::isGit()) {
-			$branch = self::gitBranchName();
-			return $branch !== 'edge' && $branch !== 'dev';
+			// The stable channel is a release checked out on a tag, i.e. a detached HEAD
+			// with no branch name. Any branch — `edge`/`dev` or a custom development
+			// branch — is the rolling channel, so only a detached HEAD is stable.
+			return self::gitBranchName() === null;
 		}
 		return !str_contains($currentVersion, 'dev') && !str_contains($currentVersion, 'edge');
 	}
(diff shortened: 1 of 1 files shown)


app/Controllers/updateController.php
```diff
@@ -179,6 +196,10 @@ public function indexAction(): void {
 	}
 
 	private function is_release_channel_stable(string $currentVersion): bool {
+		if (self::isGit()) {
+			$branch = self::gitBranchName();
+			return $branch !== 'edge' && $branch !== 'dev';
```
Frenzie: Just because you use a custom development branch hardly means it's stable. ;-)

JamBalaya56562: Good point, thanks! Fixed in the follow-up commit: the check is now positive — a git install is on the stable channel only when it's a detached `HEAD` (a release tag, so no branch name). `edge`/`dev` and any custom development branch are therefore all correctly the rolling channel. `git branch --show-current` returns empty on a detached `HEAD`, which `gitBranchName()` maps to `null`.

app/Controllers/updateController.php
```diff
@@ -179,6 +196,12 @@ public function indexAction(): void {
 	}
 
 	private function is_release_channel_stable(string $currentVersion): bool {
+		if (self::isGit()) {
+			// The stable channel is a release checked out on a tag, i.e. a detached HEAD
+			// with no branch name. Any branch — `edge`/`dev` or a custom development
+			// branch — is the rolling channel, so only a detached HEAD is stable.
+			return self::gitBranchName() === null;
```
Frenzie: That sounds like a less dependable way of checking if you're on a tag, doesn't it? ;-) On a tag the contains dev/alpha/beta/rc concept from earlier would work (and in practice I think we only have stable tags).

JamBalaya56562: Fair point — detached `HEAD` was too indirect (it also covers a raw commit checkout / mid-rebase). Changed it to check the tag directly with `git tag --points-at HEAD` in the follow-up commit.

I did weigh going back to the `FRESHRSS_VERSION` string with `dev`/`alpha`/`beta`/`rc` markers as you suggested — it's simpler (no `exec`) and, as you say, we only ship stable tags. I leaned towards the direct git check for one reason: it's a live query, so it isn't affected by a stale `FRESHRSS_VERSION` (opcache / a between-releases `-dev` string), which was the original #5850 symptom. But if you'd rather keep it to the version string, I'm happy to switch — just say the word.

Frenzie:
I'll wait to see what Alkarex thinks.

Issue #5151, opened 2023-03-01
[BUG] Duplicate posts appearing from a WordPress blog


gingerbeardman:
**Describe the bug**
Duplicate posts appearing from a WordPress blog

**To Reproduce**
1. Add [this feed](https://ephemeralenigmascom.wordpress.com/feed/)
2. Wait a while
3. A new post will appear
4. A day later a duplicate of the latest post will appear!

**Expected behavior**
- one of each post
- no duplicate posts

**Screenshots**
<img width="363" alt="FqKVNASXwAAlMLP" src="https://user-images.githubusercontent.com/49612/222266037-6def857d-b1de-4ebe-939b-b600cccd11b7.png">

**Environment information (please complete the following information):**
- Device:  iPhone, MacBook Pro
- OS: iOS 16.3.1, macOS 12.6.3
- Browser: ReadKit for iOS/macOS, FreshRSS on Safari/Chrome
- FreshRSS version: 1.20.2 & 1.21.0
- Database version: PDO
- PHP version: 8.1.16
- Installation type: Docker on Synology

**Additional context**
- checking the feed manually there is only one entry and the updated date of the feed is the same as the post is the same as the original (and duplicate) posts in FreshRSS
- duplicates are visible in FreshRSS web interface, in ReadKit and Reeder
- I've been seeing this for a while now, I do believe since the most recent FreshRSS update
    - lscr.io/linuxserver/freshrss:latest, 2023-02-23 12:51:47

math-GH:
I cannot reproduce it with 1.21.0-dev

gingerbeardman:
There hasn't been an opportunity to reproduce this as yet. 

As there hasn't been a new post to that blog since I opened this issue. 

The blog is one post per week. Every Tuesday.

gingerbeardman:
Since opening I've upgraded to 1.21 and didn't see this issue this week. 

I'll watch for one more week, to confirm, before closing.

gingerbeardman:
So it did happen after all. 

- not sure exactly when

Was anybody else subscribed and watching for it?

![image](https://user-images.githubusercontent.com/49612/224163255-9e6cd318-d521-45d0-ac91-d4dd07a0b64b.png)

math-GH:
oh wait. It could be an issue with the app and not with FreshRSS itself.
Could you please check your FreshRSS webapp?

I watched this feed and do not have this issue in the webapp.

gingerbeardman:
[...]

PR #9005, opened 2026-07-08
WebSub: ignore http/https scheme difference in Self URL comparison

### LINKED ISSUE #3087: Ignore http/https mismatch in feed URL
```
2020-06-28T12:00:54.674808423Z FreshRSS[53]: [_] [Sun, 28 Jun 2020 14:00:54 +0200] [warning] --- Warning: Self URL [http://www.blogger.com/feeds/2471378914199150966/posts/default] does not match registered
canonical URL!: https://www.blogger.com/feeds/2471378914199150966/posts/default
```
The http URL is a redirect to the https. I think this should be ignored to not spam the log with this.

JamBalaya56562:
Closes #3087

## Problem

The WebSub (PubSubHubbub) callback endpoint `p/api/pshb.php` compares the feed's declared **self URL** against the **registered canonical URL** and logs a warning when they differ:

```
Warning: Self URL [http://www.blogger.com/feeds/…] does not match registered canonical URL!: https://www.blogger.com/feeds/…
```

Many feeds serve an `http://` self URL that simply redirects to `https://` (a common security setup). In that case the two URLs differ only by scheme, yet the endpoint logs the warning on every push — spamming the PubSubHubbub log with a non-issue (#3087).

## Fix

Compare the two URLs while treating `http://` and `https://` as equivalent, so a scheme-only difference no longer triggers the warning. Everything else (host, path, query) must still match exactly, so genuine mismatches are still reported.

A small helper is added next to the existing `compareURLOrigins()` in `FreshRSS_http_Util` (which `pshb.php` already uses):

```php
public static function compareURLsIgnoringScheme(string $url1, string $url2): bool {
    $stripScheme = static fn(string $url): string => preg_replace('#^https?://#i', '//', trim($url)) ?? $url;
    return $stripScheme($url1) === $stripScheme($url2);
}
```

`compareURLOrigins()` was not reused because it ignores the path, whereas here the path/query must still be compared.

Both self-URL comparisons in `pshb.php` now go through the helper:
- the XML self vs registered canonical check (the one from the issue), and
- the XML self vs HTTP `Link` header self check just above it, which suffers from the same false positive.

## Changes

- **`app/Utils/httpUtil.php`** — add `compareURLsIgnoringScheme()`.
- **`p/api/pshb.php`** — use it for the two `Self URL … mismatch` comparisons.
- **`tests/app/Utils/httpUtilTest.php`** (new) — unit tests for the helper: scheme-only difference is equal; identical URLs equal; differing path / trailing slash / host / query are not equal; non-http(s) schemes compared as-is.

No user-facing strings change, so no i18n updates.

## Testing

- `vendor/bin/phpunit tests/app/Utils` → **22 tests OK** (11 new).
- `php -l` on both changed files → OK.
- `phpstan` on the changed files → no errors.

### DIFF
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -295,6 +295,11 @@ final class FreshRSS_http_Util {
 			($url1['port'] ?? '') === ($url2['port'] ?? '');
 	}
 
+	public static function compareURLsIgnoringScheme(string $url1, string $url2): bool {
+		$stripScheme = static fn(string $url): string => preg_replace('#^https?://#i', '//', trim($url)) ?? $url;
+		return $stripScheme($url1) === $stripScheme($url2);
+	}
+
 	/**
 	 * Returns a value for CURLOPT_RESOLVE as an array, null if no allowed IPs were found, false if the domain failed to resolve.
 	 *
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -295,9 +295,12 @@ final class FreshRSS_http_Util {
 			($url1['port'] ?? '') === ($url2['port'] ?? '');
 	}
 
-	public static function compareURLsIgnoringScheme(string $url1, string $url2): bool {
-		$stripScheme = static fn(string $url): string => preg_replace('#^https?://#i', '//', trim($url)) ?? $url;
-		return $stripScheme($url1) === $stripScheme($url2);
+	/**
+	 * Return 0 if values on either side are equal ignoring the HTTP vs HTTPS differences, or 1/-1 if they differ.
+	 */
+	public static function compareUrlIgnoringHttps(string $url1, string $url2): int {
+		$normalizeScheme = static fn(string $url): string => preg_replace('#^https?://#i', '//', trim($url)) ?? $url;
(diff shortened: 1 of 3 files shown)


Alkarex:
Related:
* https://github.com/simplepie/simplepie/pull/951
* https://github.com/Automattic/pushpress/pull/16
* https://github.com/FreshRSS/FreshRSS/issues/5151

Alkarex:
I have made some minor preference adjustments. Please double-check

JamBalaya56562:
Thanks a lot for the adjustments @Alkarex! I've double-checked the `Stylistic preferences` commit and it looks good to me — no behaviour change:

- The scheme-normalisation itself is untouched, so the comparison result is identical. Switching the return from a `bool` (`===`) to an `int` comparator (`<=>`) is equivalent for both call sites, since they now test `!== 0` where they previously negated the boolean (equal → no warning, differ → warning).
- The rename to `compareUrlIgnoringHttps` is applied consistently across the definition, both `pshb.php` call sites, and the test.
- The extra `HTTPS://…` vs `http://…` test case is a nice addition, and all the existing equal/not-equal cases still hold.

Happy with it as-is. Thanks again for the review and the polish!

PR #7092, opened 2024-12-12
Integrating Readeck as a Sharing Method

Abdallah-Abdaljalil:
Related to #6998 

Changes proposed in this pull request:

- Added Readeck as a sharing method in shares.php
- Added Readeck to every language folder under app/i18n to make it appear in the interface for any language
- Added Readeck to the documentation of available sharing methods in 08_sharing_services.md

How to test the feature manually:

1. Add Readeck as a sharing method in your FreshRSS
2. Try to share anything using Readeck
3. It should appear in your Readeck locally

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://github.com/FreshRSS/FreshRSS/tree/edge/docs/en/developers/04_Pull_requests.md).

PR #9008, opened 2026-07-10
Add Readeck as a sharing method

### LINKED ISSUE #6998: [Feature] Please add Readeck as sharing method
No!

It would be great to have [Readeck](https://readeck.org/) added to the share-menu in FreshRSS.

Manually adding links to Readeck.

**Additional context**
Readeck is an Open Source  "read-later-application" (like Wallabag).

JamBalaya56562:
Closes #6998

Supersedes the stalled #7092, whose author became unresponsive and whose change did not work: it only added a `method => POST` entry to `app/shares.php`, but FreshRSS had no way to authenticate the request (as noted in that PR's review).

Readeck can only add a bookmark from another site through its REST API — `POST {baseURL}/api/bookmarks` with an `Authorization: Bearer <token>` header and a JSON body. There is no browser-openable "add bookmark" URL, so the usual "open a share URL" mechanism used by every other service cannot work for it. See the measurements below for why the web endpoint is not an alternative.

- Introduce a new `'token'` share form type (name + base URL + API token) in `app/shares.php` and `FreshRSS_Share`, reusable by any future token-authenticated service.
- Add a generic click handler in `p/scripts/main.js` that posts the article link with `Authorization: Bearer` ([RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750)) and a JSON body. It is keyed on the presence of `data-token` and takes the JSON field name from the existing `field` share option, so **no service name appears in core JS** and a future token-authenticated service is just an entry in `shares.php`.
- Extend the `connect-src` Content-Security-Policy directive with the origins of the configured token shares, otherwise the browser would block the request before it is even sent.
- Add **Readeck** using this form type, the `conf.sharing.share_token` label and the `gen.share.readeck` name to all locales, and document the service.

## Why not a plain POST form to Readeck's web endpoint

@Inverle helpfully pointed at [a Readeck comment](https://codeberg.org/readeck/readeck/issues/709#issuecomment-10735763) describing `POST {baseURL}/bookmarks` with a `url` field and no API token, which would have needed no new core code at all. I ran a Readeck container to check, and it cannot work from FreshRSS:

| Request | Result |
|---|---|
| Session cookie, **no `Origin` header** | `303` → `/bookmarks/<id>` — created, no CSRF token needed |
| No session cookie | Not created |
| Session cookie, **`Origin: https://freshrss.example.com`** | **`403 Forbidden`** |

Readeck rejects cross-origin form posts, and its session cookie is `SameSite=Lax`, so it would not be sent on a cross-site POST anyway. The linked use case works because a Firefox search shortcut is an address-bar navigation, which sends no `Origin` header at all; a form submitted from a FreshRSS page always sends one. FreshRSS and Readeck are always different origins, so this applies to same-site deployments too.

1. Go to Settings → Sharing, add **Readeck**, fill in the Readeck base URL and an API token, then save.
2. Open an article's share menu and click **Readeck**.
3. The article is saved to Readeck; the button turns green on success and red on failure.

Note: the request is made from the browser, so the Readeck instance must allow cross-origin requests (CORS) from the FreshRSS origin. This is documented in `docs/en/users/08_sharing_services.md`. The API token is stored in the user configuration and rendered into the share menu (`data-token`), which is inherent to this browser-side approach.

## Pull request checklist

- [x] clear commit messages
- [x] code manually tested
- [x] documentation updated

### DIFF
--- a/app/i18n/fr/conf.php
+++ b/app/i18n/fr/conf.php
@@ -335,6 +335,7 @@ return array(
 		'remove' => 'Supprimer la méthode de partage',
 		'shaarli' => 'Shaarli',	// IGNORE
 		'share_name' => 'Nom du partage à afficher',
+		'share_token' => 'API token',	// TODO
 		'share_url' => 'URL du partage à utiliser',
 		'title' => 'Partage',
 		'twitter' => 'X (Twitter)',	// IGNORE
--- a/app/i18n/fr/conf.php
+++ b/app/i18n/fr/conf.php
@@ -335,7 +335,7 @@ return array(
 		'remove' => 'Supprimer la méthode de partage',
 		'shaarli' => 'Shaarli',	// IGNORE
 		'share_name' => 'Nom du partage à afficher',
-		'share_token' => 'API token',	// TODO
+		'share_token' => 'Jeton d’API',
 		'share_url' => 'URL du partage à utiliser',
 		'title' => 'Partage',
 		'twitter' => 'X (Twitter)',	// IGNORE
--- a/README.fr.md
+++ b/README.fr.md
@@ -234,7 +234,7 @@ Voir le [dépôt dédié à ces extensions](https://github.com/FreshRSS/Extensio
 | English (United States) (en-US) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fen-US+%2F%28TODO%7CDIRTY%29%24%2F) |
 | Español (es) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fes+%2F%28TODO%7CDIRTY%29%24%2F) |
 | فارسی (fa) | ￭￭￭￭￭￭￭￭￭･ 99% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffa+%2F%28TODO%7CDIRTY%29%24%2F) |
-| Suomi (fi) | ￭￭￭￭￭￭￭￭￭･ 92% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffi+%2F%28TODO%7CDIRTY%29%24%2F) |
+| Suomi (fi) | ￭￭￭￭￭￭￭￭￭･ 91% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffi+%2F%28TODO%7CDIRTY%29%24%2F) |
 | Français (fr) | ￭￭￭￭￭￭￭￭￭￭ 100% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Ffr+%2F%28TODO%7CDIRTY%29%24%2F) |
 | עברית (he) | ￭￭￭￭･･････ 41% | [contribuer](https://github.com/search?q=repo%3AFreshRSS%2FFreshRSS+path%3Aapp%2Fi18n%2Fhe+%2F%28TODO%7CDIRTY%29%24%2F) |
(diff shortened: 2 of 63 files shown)


app/i18n/fr/conf.php
```diff
@@ -335,6 +335,7 @@
 		'remove' => 'Supprimer la méthode de partage',
 		'shaarli' => 'Shaarli',	// IGNORE
 		'share_name' => 'Nom du partage à afficher',
+		'share_token' => 'API token',	// TODO
```
```suggestion
		'share_token' => 'Jeton d’API',
```

JamBalaya56562: Thanks @Alkarex! Applied in a755accd — the French locale now reaches 100%.

Alkarex:
I am a bit worried of having third-party specific code in our core code base (e.g. `main.js`).
If possible, help welcome to think about an extension mechanism to allow such a use-case without changing too much the core code base.

Frenzie:
I think the idea of `Authorization: Bearer …` is sufficiently standard not to regard it as specific to Readeck (though if none of the other services require anything like it maybe it's not standard in our context).

But don't we do the rest via PHP rather than JS? That seems like a bit of an unexpected exception.

JamBalaya56562:
Thanks @Alkarex and @Frenzie — both points are fair. @Alkarex, you are right: `data-type="readeck"` would be the first hard-coded third-party service name in core JS. I would like to propose two changes that address this together.

**A. Make the mechanism generic.** Key the behaviour on the `token` form type rather than on the service name (the template already emits it generically via `formType() === 'token'`). Readeck then disappears from core JS, and any future service using `Authorization: Bearer` — standard per [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750), as @Frenzie notes — works with no core JS change, in the same way `'form' => 'advanced'` does today.

**B. Move the request server-side.** Perform the POST from PHP instead of `fetch()`. This removes the JS exception @Frenzie pointed at, and drops three costs of the current design: users no longer need CORS on their Readeck instance, the CSP `connect-src` extension (`shareConnectSrc()`) can be reverted entirely, and the API token no longer reaches the browser.

Trade-off for B: it requires the FreshRSS *server* to reach Readeck, so setups where Readeck is only reachable from the browser (LAN-only, `localhost`) would stop working. If you consider that case important, A alone still removes the service-specific code from core, at the cost of keeping the CORS requirement.

<details><summary>Background: how I checked this</summary>

**Precedent in `main.js`** — every existing `data-type` handler is either a browser capability (`print`, `clipboard`, `web-sharing-api`), a browser workaround (`email-webmail-firefox-fix`), or a generic mechanism (`a[href="POST"]`). None names a third-party service.

**PHP vs JS** — a small clarification: no share performs a server-side request today either. GET shares are `<a target="_blank">` navigations and POST shares are browser `<form>` submissions; PHP only builds the markup. But the substance of the point stands: everything else is declarative (an entry in `shares.php` plus the generic template) and needs zero bespoke JS, while this does.

**Why it ended up in JS** — a browser form POST cannot set an `Authorization` header or send a JSON body, so the existing generic POST path could not carry the token.

</details>

<details><summary>On an extension mechanism (@Alkarex)</summary>

Happy to help think about it, but it looks like separate work: there is currently no share-related hook in `Minz_HookType`, and `FreshRSS_Share::load()` is only ever called with the core `app/shares.php`, so extensions cannot register a share service at all today. Adding that API seems worth its own PR rather than being bundled here.

</details>

Which direction would you prefer — A+B, A only, or should this wait until a share extension point exists?

Frenzie:
Sounds to me like it's broadly alright as is then. ("option A")

Inverle:
It seems there is an option without the API via a POST form? https://codeberg.org/readeck/readeck/issues/709#issuecomment-10735763

JamBalaya56562:
Thanks @Frenzie — done in cca8653. The handler is now keyed on the presence of `data-token` rather than a service name, and the JSON field name comes from the existing `field` share option, so a service authenticating with `Authorization: Bearer` is just an entry in `shares.php` with no core JS change. No third-party name is left in core JS.

JamBalaya56562:
Thanks @Inverle, that was a good lead and I wanted it to work — it would have removed the need for this PR's core changes entirely. I ran a Readeck container to check, and unfortunately it cannot work from FreshRSS:

| Request | Result |
|---|---|
| Session cookie, no `Origin` header | `303` → `/bookmarks/<id>` — created, no CSRF token needed |
| No session cookie | Not created |
| Session cookie, `Origin: https://freshrss.example.com` | **`403 Forbidden`** |

Readeck rejects cross-origin form posts, and its session cookie is `SameSite=Lax`, so it would not be sent on a cross-site POST anyway. The case in the linked comment works because a Firefox search shortcut is an address-bar navigation, which sends no `Origin` header at all — a form submitted from a FreshRSS page always sends one. Since FreshRSS and Readeck are always different origins, this applies to same-site deployments too.

I have added these measurements to the PR description so the API-token approach is justified by evidence rather than assumption.

Inverle:
https://codeberg.org/readeck/readeck/pulls/1241 might improve the CORS situation at least

Issue #3078, opened 2020-06-24
RSS parser not filtering media:group/media:content where isDefault = true


will-davies:
RSS specification for media:group with multiple media:content elements has the ability to define which element is the default.

https://www.rssboard.org/media-rss#media-content
isDefault determines if this is the default object that should be used for the <media:group>. There should only be one default object per <media:group>

Currently if a feed has an item with multiple versions of an image for example, all versions are rendered in the view.

i.e.
`<media:group>
 <media:description>…</media:description>
  <media:content url="http://example.com/image-16x9-2150x1210.jpg" medium="image" type="image/jpeg" width="2150" height="1210"/>
  <media:content url="http://example.com/image-4x3-940x705.jpg" medium="image" type="image/jpeg" width="940" height="705"/>
  <media:content url="http://example.com/image-3x2-940x627.jpg" medium="image" type="image/jpeg" width="940" height="627" isDefault="true"/>
  <media:content url="http://example.com/image-3x4-940x1253.jpg" medium="image" type="image/jpeg" width="940" height="1253"/>
  <media:content url="http://example.com/image-1x1-1400x1400.jpg" medium="image" type="image/jpeg" width="1400" height="1400"/>
  <media:thumbnail url="http://example.com/image-4x3-140x105.jpg" width="140" height="105"/>
</media:group>`

All images are then shown in the reader.

Alkarex:
Could you please provide an example of such a feed, to ease testing?

will-davies:
Absolutely, apologies for not including previously.
http://www.abc.net.au/news/feed/45910/rss.xml

samwilson:
I came to see if this had already been reported, because I've been putting up with seeing all images for ages. I thought "oh good, I've got another example of a feed that's got this", but no! It's just the ABC again: https://www.abc.net.au/radionational/programs/blueprintforliving/feed/6052176/podcast.xml :-)

Still, it does sound like something that FreshRSS could handle (assuming this is a standard form for feeds).

math-GH:
How should we handle the other media enclosures?

Here an example from https://www.abc.net.au/news/feed/45910/rss.xml
5 x media:content
Should we trust the `isDefault` and use this image, regardless if it is to small/big?
```
<media:group>
        <media:description>Tammy is one of the 47 per cent of Australians who live with a chronic condition.</media:description>
        <media:content
[...]

PR #9009, opened 2026-07-10
Show only one representation per media:group

### LINKED ISSUE #8569: Better handling of <media:group>
### Describe the bug

The following feed has a <media:group> with several `media:peerLink` and `media:content`.

The media will be shown one by one

Image

From the RSS 2.0 standard:

> 4.1. [media:group](https://www.rssboard.org/media-rss#media-group)
> 
> <media:group> is a sub-element of <item>. It allows grouping of <media:content> elements that are effectively the same content, yet different representations. For instance: the same song recorded in both the WAV and MP3 format. It's an optional element that must only be used for this purpose.

https://tube.funfacts.de/feeds/videos.xml?videoChannelId=3

Image

[videos.xml](https://github.com/user-attachments/files/25745668/videos.xml)

### To Reproduce

https://tube.funfacts.de/feeds/videos.xml?videoChannelId=3

### Expected behavior

only one video is shown

### FreshRSS version

1.28.2-dev 

### System information

FreshRSS version
    1.28.2-dev
PHP version
    8.2.29
cURL version
    8.9.1
    OpenSSL/3.3.2
Database
    mysql
    10.2.36-MariaDB-log
    mysqlnd 8.2.29

### Additional context

_No response_

TowyTowy:
Fixes #8569 (see also the older #3078)

- A `<media:group>` gathers `<media:content>` elements that are "effectively the same content, yet different representations" ([Media RSS specification](https://www.rssboard.org/media-rss#media-group)), but every representation was stored and rendered as its own enclosure. Articles from PeerTube feeds, for instance, showed the same video five times (1080p/720p/480p/360p + audio-only).
- When loading entries, keep only one `<media:content>` per `<media:group>`: the one marked with `isDefault="true"`, or the first one otherwise (the rule discussed in #3078).
- Elements outside a `<media:group>` are not affected: plain RSS `<enclosure>`, item-level `<media:content>`, and groups with a single `<media:content>` (e.g. YouTube) behave as before. Note: PeerTube items also carry a plain `<enclosure>` (download link for the same video) outside the group, so those articles now show two players instead of six; happy to extend the handling if you would like that covered too.

1. Subscribe to a PeerTube feed, e.g. the videos.xml attached to #8569.
2. Open an article.
3. The video appears once (the `isDefault` 1080p representation) instead of five times.

Verified on a local instance (edge, Docker) with the videos.xml from #8569: each article rendered five `<video>` players plus one `<audio>` player from the `<media:group>` before this change, and only the `isDefault` representation with it.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested (live before/after comparison above; `php -l`, phpcs PSR-12 and PHPStan clean)
- [x] unit tests written (`tests/app/Models/FeedTest.php`, 4 cases incl. no-`isDefault` fallback and non-group enclosures)
- [x] documentation updated (n/a)


app/Models/Feed.php
```diff
@@ -847,9 +899,11 @@ public function loadEntries(FreshRSS_SimplePieCustom $simplePie): Traversable {
 
 			$attributeEnclosures = [];
 			if (!empty($item->get_enclosures())) {
+				// Skip the redundant representations of the same content in a `<media:group>`
+				$redundantEnclosureUrls = self::mediaGroupsRedundantUrls($item);
 				foreach ($item->get_enclosures() as $enclosure) {
```
Alkarex: I am wondering whether this logic might be better placed in the `get_enclosures()` function, with a new parameter to perform the filtrering.
Any thoughts?

TowyTowy:
Agreed — that's the cleaner home for it, and it fits how FreshRSS already treats SimplePie. You already carry local SimplePie patches where it makes sense (e.g. `1349c7c` "SimplePie: Disallow `javascript:` URI protocol" in `lib/simplepie/simplepie/src/`), so handling media:group redundancy at the parsing layer is consistent with that rather than a new kind of divergence — and it's genuinely a Media RSS concern, so `get_enclosures()` is the right place.

Concretely, I'd propose: add an opt-in parameter to `\SimplePie\Item::get_enclosures()` (default off, so the many other callers are unaffected) that, when set, keeps only one `media:content` per `<media:group>` — the `isDefault="true"` one, else the first — and have `Feed.php` pass the flag. That removes the post-filtering from `Feed.php` entirely.

Two small things worth your call before I rework it:
1. Parameter name/shape — e.g. `get_enclosures(bool $dedupe_media_groups = false)`, or would you prefer an options-array style?
2. You periodically bump SimplePie wholesale (#8947), so this patch would need re-applying on each bump — same as your existing SimplePie patches. If you'd rather it eventually come back "for free" in a bump, I'm happy to also send it upstream to simplepie/simplepie (I have a PR open there already) and keep this as the interim.

Happy to move it into `get_enclosures()` as soon as you confirm the shape you'd like.

TowyTowy:
Good call — that's a much cleaner home for it. Moved the logic into `\SimplePie\Item::get_enclosures()` behind a new `bool $excludeMediaGroupAlternatives = false` parameter (default keeps the current behaviour for every existing caller); `Feed::loadEntries()` now just calls `get_enclosures(excludeMediaGroupAlternatives: true)`.

A nice side effect of it living in `Item`: the helper can use the private `get_own_base()` that `get_enclosures()` already uses when building the URLs, so relative `<media:content>` URLs now match exactly — the previous version in `Feed.php` had to fall back to the public `get_base()`. The existing `FeedTest` cases are unchanged and still pass.

Since this now touches the bundled SimplePie, happy to mirror it in the SimplePie fork too if you'd prefer that path.

Alkarex:
Could you please open a PR to our https://github.com/FreshRSS/simplepie repo with the changes you have made to lib/simplepie/simplepie/src/Item.php and potentially tests/app/Models/FeedTest.php ?

TowyTowy:
Done: FreshRSS/simplepie#86 ports the Item.php change (with PHP 7.2-compatible syntax and a regression test in the library's own EnclosureTest). Once that lands I can update this PR's vendored copy to match, so the two stay in sync.

Alkarex:
Thanks! See https://github.com/FreshRSS/FreshRSS/blob/edge/lib/README.md#updating-libraries

TowyTowy:
Done - rebased on edge and regenerated `lib/simplepie/` per lib/README.md (`composer update --no-autoloader`), bumping the pin to `7acfcf7` so the vendored Item.php matches the merged FreshRSS/simplepie#86 exactly. Tests remain green.

Alkarex (approved):

PR #9010, opened 2026-07-10
Warn during install if document root does not point to ./p

### LINKED ISSUE #3679: [BUG] Install routine does not check the correct expose of ./p folder
**Describe the bug**

If the domain does not point directly to ./p folder it is possible to install FreshRSS without any notice in step1 in the install routine.

grafik

In the manual it is described:
grafik

If the ./p folder is not correct exposed the other folders (f.e. ./data folder) are online available, where imporant private information are stored.  They are (mostly) protected with .htaccess (only valid for Apache server)

**To Reproduce**
1. fresh installation
2l set domain to root, not to /p
3. follow the installation

**Expected behavior**
red/yellow notification in step 1, that the domain is not correct set.

 - FreshRSS version: 1.18.1

**Additional context**
that was the reason for #3657

Elgeryy1:
## Description

If the Web server's document root is misconfigured to serve the FreshRSS repository root instead of the `./p` public folder, other folders such as `./data` (credentials, tokens, database) can become publicly accessible. The installer didn't warn about this.

This adds a non-blocking check to installation step 1 (alongside the existing file/permission checks) that compares the detected `DOCUMENT_ROOT` against the `./p` folder and shows a warning when they don't match.

The check is skipped whenever `DOCUMENT_ROOT` is unknown (e.g. CLI/cron context) — including when it's an empty string, since `realpath('')` would otherwise misleadingly resolve to the current working directory and produce a false warning (I caught this while testing against the real cron job in the Docker dev setup).

Fixes #3679

## Testing

- Verified in the Docker development setup (`docker-compose-development.yml`): step 1 of the Web installer now shows "Okay! Your web server document root correctly points to the ./p folder." in a standard install.
- Verified no false warning is raised when running the cron feed-refresh command (`app/actualize_script.php`) directly, where `DOCUMENT_ROOT` is set but empty.
- `composer run-script phpcs` and `composer run-script phpstan` pass with no errors.

### DIFF
--- a/lib/lib_install.php
+++ b/lib/lib_install.php
@@ -49,6 +49,12 @@ function checkRequirements(string $dbType = '', bool $checkPhp = true, bool $che
 	$users = is_dir(USERS_PATH) && touch(USERS_PATH . '/index.html');
 	$favicons = is_dir(DATA_PATH) && touch(DATA_PATH . '/favicons/index.html');
 	$tokens = is_dir(DATA_PATH) && touch(DATA_PATH . '/tokens/index.html');
+	// If the document root is known and does not match the `./p` public folder, other folders such as `./data` are likely web-exposed.
+	// Note: `DOCUMENT_ROOT` may be an empty string outside of a Web server context (e.g. CLI, cron); `realpath('')` would otherwise
+	// misleadingly resolve to the current working directory, so it must be treated the same as "unknown".
+	$docRootPath = isset($_SERVER['DOCUMENT_ROOT']) && is_string($_SERVER['DOCUMENT_ROOT']) && $_SERVER['DOCUMENT_ROOT'] !== ''
+		? realpath($_SERVER['DOCUMENT_ROOT']) : false;
+	$docRootOk = $docRootPath === false || $docRootPath === realpath(PUBLIC_PATH);
 
 	$result = [];
 	if ($checkPhp) {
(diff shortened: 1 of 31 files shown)


lib/lib_install.php
```diff
@@ -49,6 +49,12 @@ function checkRequirements(string $dbType = '', bool $checkPhp = true, bool $che
 	$users = is_dir(USERS_PATH) && touch(USERS_PATH . '/index.html');
 	$favicons = is_dir(DATA_PATH) && touch(DATA_PATH . '/favicons/index.html');
 	$tokens = is_dir(DATA_PATH) && touch(DATA_PATH . '/tokens/index.html');
+	// If the document root is known and does not match the `./p` public folder, other folders such as `./data` are likely web-exposed.
+	// Note: `DOCUMENT_ROOT` may be an empty string outside of a Web server context (e.g. CLI, cron); `realpath('')` would otherwise
+	// misleadingly resolve to the current working directory, so it must be treated the same as "unknown".
+	$docRootPath = isset($_SERVER['DOCUMENT_ROOT']) && is_string($_SERVER['DOCUMENT_ROOT']) && $_SERVER['DOCUMENT_ROOT'] !== ''
+		? realpath($_SERVER['DOCUMENT_ROOT']) : false;
+	$docRootOk = $docRootPath === false || $docRootPath === realpath(PUBLIC_PATH);
```
Alkarex: Maybe the case when `DOCUMENT_ROOT` is not defined should trigger the warning as well

Elgeryy1: Implemented: an unavailable `DOCUMENT_ROOT` now triggers the warning, while an explicitly empty value remains valid for CLI/cron installs. I verified the missing, empty, matching, and invalid document-root cases.

Alkarex (approved):

PR #9011, opened 2026-07-10
Docker: use double quotes for CRON_MIN examples

### LINKED ISSUE #4818: Documentation: Quick Run -e 'CRON_MIN=1,31'  documentation not working,  CRON_MIN=*/20 works
In windows using quotes when creating the docker container using the Quick Run command, does not update the CRON_MIN parameter, so cron and the auto-refresh will not work: -e 'CRON_MIN=1,31' 

Removing the quotes -e CRON_MIN=*/20 works

Elgeryy1:
## Description

The Docker Quick Run / cron examples use single-quoted `-e 'CRON_MIN=1,31'`. This works on Linux/macOS shells and PowerShell, but not on Windows `cmd.exe`, which does not strip single quotes from arguments — they end up included literally in the variable, so the cron job never gets scheduled correctly. Using `-e CRON_MIN=*/20` (no quotes) works, which was the original report.

This switches all the affected examples in `Docker/README.md` and `docs/en/admins/08_FeedUpdates.md` to double quotes, which work consistently across Linux/macOS shells, PowerShell, and cmd.exe, and adds a short note explaining why.

Fixes #4818

## Testing

Verified directly on Windows: `docker run --rm -e 'CRON_MIN=1,31' alpine env` via `cmd.exe` yields the broken literal `'CRON_MIN=1,31'` as the variable, while the double-quoted form yields a clean `CRON_MIN=1,31` in cmd.exe, PowerShell, and bash alike.

`npm run markdownlint`: 0 errors.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -32,12 +32,14 @@ Example running FreshRSS (or scroll down to the [Docker Compose](#docker-compose
 docker run -d --restart unless-stopped --log-opt max-size=10m \
   -p 8080:80 \
   -e TZ=Europe/Paris \
-  -e 'CRON_MIN=1,31' \
+  -e "CRON_MIN=1,31" \
   -v freshrss_data:/var/www/FreshRSS/data \
   -v freshrss_extensions:/var/www/FreshRSS/extensions \
   --name freshrss \
   freshrss/freshrss
 ```
+
+> ℹ️ Use double quotes (`"..."`), not single quotes, around `CRON_MIN` values such as `"CRON_MIN=1,31"`. Single quotes are not stripped by Windows `cmd.exe` and end up included in the variable, whereas double quotes work correctly on Linux/macOS shells, PowerShell, and `cmd.exe` alike.
 
 * Exposing on port 8080
 * With a [server timezone](http://php.net/timezones) (default is `UTC`)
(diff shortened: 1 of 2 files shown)


Docker/README.md
```diff
@@ -32,13 +32,15 @@ Example running FreshRSS (or scroll down to the [Docker Compose](#docker-compose
...
-  -e 'CRON_MIN=1,31' \
+  -e "CRON_MIN=1,31" \
   -v freshrss_data:/var/www/FreshRSS/data \
   -v freshrss_extensions:/var/www/FreshRSS/extensions \
   --name freshrss \
   freshrss/freshrss
 ```
 
+> ℹ️ Use double quotes (`"..."`), not single quotes, around `CRON_MIN` values such as `"CRON_MIN=1,31"`. Single quotes are not stripped by Windows `cmd.exe` and end up included in the variable, whereas double quotes work correctly on Linux/macOS shells, PowerShell, and `cmd.exe` alike.
+
```
Alkarex: I am not sure this up to our documentation to include such Docker/platform specific information.
While the changes are fine, at a first glance I think it would be better to remove the comment to avoid adding too much noise to an already information-heavy documentation

Alkarex:
I am actually wondering whether there is any relevant case remaining for which the quotes are relevant at all.
If you have access to a Windows setup, would you be able to test whether e.g. `CRON_MIN=1,31` and `CRON_MIN=*/20` work without quotes, in docker run and Docker Compose?

Frenzie:
At a glance it seems unnecessary, at least in non-Windows (bash, zsh and such).

Docker/README.md
Elgeryy1: Addressed: I removed the platform-specific explanatory note, and the examples now use bare `CRON_MIN` values.

Alkarex (approved):

PR #9012, opened 2026-07-10
Add per-feed option to force IPv4 or IPv6

### LINKED ISSUE #4560: [Feature] Force ipv4 for individual feeds

Elgeryy1:
## Description

Some feeds are only reachable over one IP version, or resolve to a broken address on the other (cURL error 35 on some hosts). This exposes `CURLOPT_IPRESOLVE` as a new per-feed tri-state option (By default / IPv4 / IPv6), following the existing `ssl_verify` pattern:

- Read the `ipresolve` parameter in both the quick "add feed" flow (`feedController::addAction`) and the feed edit flow (`subscriptionController::feedAction`)
- Apply it in both cURL fetch paths (`SimplePieCustom` and `httpUtil`) using the correct `CURL_IPRESOLVE_V4`/`V6` integer constants
- Add the matching select to the add and edit feed forms

Fixes #4560. Related to the earlier attempt in #4573, which stalled and was closed — that PR set `CURLOPT_IPRESOLVE` to a string instead of the integer constant, which wouldn't have worked; this one uses the real PHP constants.

## Testing

`vendor/bin/phpcs` and `vendor/bin/phpstan analyse` (whole project): no errors.

Known gap, matching existing `ssl_verify` scope: favicon fetching does not currently pass per-feed curl attributes through, so this option (like `ssl_verify`) does not yet apply there. Happy to extend if that's wanted.


JamBalaya56562:
@Elgeryy1 Could you resolve merge conflicts?

Alkarex (approved):

Alkarex:
I have updated the code to use native `CURL_IPRESOLVE_*` constants.
Tests welcome.

And ideas welcome to allow setting more cURL parameters without having to list them all in our code base.

JamBalaya56562:
Thanks @Alkarex for jumping on this so quickly, and for handling the constants yourself.

On your second point: most of the generic plumbing already exists — only the top layer is duplicated per option. `curl_params` is already a generic `CURLOPT_* => value` map, applied generically in both transport paths (`SimplePieCustom::__construct()` and `httpUtil::httpGet()`) and round-tripped through OPML as `frss:CURLOPT_*`. The gate is `sanitizeCurlParams()`, whose `$safe_params` allowlist decides what a user may set.

So `CURLOPT_IPRESOLVE` could be a single allowlist entry rather than a dedicated `ipresolve` attribute, dropping the bespoke handling in `SimplePieCustom` and `httpUtil::httpGet()`.

What `$safe_params` cannot express today is *which values* are acceptable — it allowlists keys only, with `CURLOPT_COOKIEFILE` and `CURLOPT_HTTPHEADER` special-cased inside the loop. Making it a map of `CURLOPT_* => spec` (allowed values, i18n label, widget kind) would let one entry drive the allowlist, the value validation, the form field, and the per-option `if` chain in `ImportService`.

Happy to send that as a separate PR.

Alkarex:
> Happy to send that as a separate PR.

Yes please, so we can evaluate how it looks like more concretely and whether it is positive or not

Elgeryy1:
Opened #9134 with the `safe_params` refactor into a data-driven map, scoped to `sanitizeCurlParams()` for now (see the PR description for why I left the OPML-import side out of it).

JamBalaya56562:
Thanks for working on this. If possible, could you please add before-and-after screenshots showing the new IP-version selector in both the **Add feed** and **Edit feed** forms?

Since the option is in the advanced section, please show that section expanded. This would make the new per-feed setting and its placement much easier to understand.

Elgeryy1:
Here are before/after screenshots showing the new IP version selector in the advanced section of the Add feed and Edit feed forms:

**Add feed – after (with IP resolve option, advanced section expanded)**
9012-add-feed-ipresolve

**Add feed – before (no IP resolve option)**
9012-before-add-feed-no-ipresolve

**Edit feed – after (with IP resolve option, advanced section expanded)**
9012-edit-feed-ipresolve

PR #9016, opened 2026-07-10
Warn when adding a feed that looks like a duplicate

### LINKED ISSUE #7648: [Feature] Show a warning when subscribing to an already added feed

Elgeryy1:
## Description

Adds a warning when subscribing to a feed URL that looks like a near-duplicate of one you already have (e.g. differs only by http/https or a trailing slash), with a confirmation step to add it anyway if intended. Follows the design Alkarex outlined in the previous attempt at this (#7675, closed): a pure, comparison-only URL-normalization helper that is never used to mutate the actually-stored feed URL.

The duplicate-confirmation step re-renders the full Add Feed form (category, feed kind, XPath/JSON scraping settings, proxy/auth/curl options, etc.) pre-filled with what was submitted, rather than dropping to a stripped-down confirmation view — so nothing the user configured gets silently lost if they confirm. Byte-identical URL matches are excluded from this soft warning entirely and always fall through to the existing "already subscribed" error.

Fixes #7648. Related to the earlier attempt in #7675.

## Testing

`vendor/bin/phpcs`, `vendor/bin/phpstan analyse` (whole project), and the full `phpunit` suite (665 tests): all clean/passing, including new tests for the URL-normalization helper and the duplicate-detection logic (covering the case where a submission exactly matches one existing feed while also near-matching a different one).

Manually verified end-to-end: submitting the main Add Feed form with a non-default category and an advanced option (e.g. custom user-agent) against a near-duplicate URL preserves those values through the warning-and-confirm round trip.


JamBalaya56562:
@Elgeryy1 
Could you resolve merge conflicts?

Elgeryy1:
Rebased onto the current `edge` branch and resolved the merge conflict. The updated GitHub Actions test is passing.

Elgeryy1:
Resolved with merge commit ae9366231, integrating the current `edge` changes while preserving the duplicate-feed warning changes. Follow-up commit bc3a73c45 synchronizes the generated translation status; CI is now green.

JamBalaya56562:
Thanks for working on this. If possible, could you please add before-and-after screenshots of the feed-add flow, including the new warning state shown after submitting a URL that looks similar to an existing feed?

It would be helpful for the new screenshot to include the warning and the confirmation control for continuing with the addition. This would make the new interaction and its purpose much easier to understand.

Elgeryy1:
Before (submitting a URL that only differs by case/trailing slash from an existing feed just failed with a generic "already subscribed" error — no explanation, no way to continue):

9016-before-no-duplicate-warning

After (the new check explains why it looks like a duplicate and lets you confirm and add it anyway, keeping every field you'd already filled in):

9016-duplicate-warning-full

PR #9017, opened 2026-07-10
🤖 Add feed tree item attributes hook

### LINKED ISSUE #8472: [Feature] Configuration per feed and category of an extension

There's no easy way to allow an extension to create configuration per feed or per configuration.  
Even trying to create workarounds we can find several parts which create friction.  

### Use case

I'm working on an extension that would benefit on allowing the user to create configuration per feed and/or per category.  

Allow an extension to add a section into the slider `feed_update` to manage the feed.  
Something similar to having the file `configure.phtml` to configure the extension itself.  

These are the steps I found with a bit of research, not sure if it's the best workaround.  

I created a static `feed_config.html` which needs to be loaded and sent via jsVars
```php
public function jsVars(array $vars): array {
	$vars['ntfy_feed_config_html'] = file_get_contents(__DIR__ . '/static/feed_config.html');
	return $vars;
}
```

`Minz_Extension::getFileUrl()`'s `$isStatic` can't be used since there's a restriction with the array `MIME_TYPES` on both cases.  

Once this is loaded the custom js file needs to find a way to find the slider and detect the `feed_update` element to add the custom section.  
Afterward, the html file can be injected into the `innerHTML` of an element, and manually fill and fetch the data to be saved.  

To save the data I reused the `handleConfigureAction()` callback by manually calling `/i/?c=extension&a=configure&e=${context.extensions.ntfy_ext_name}&ntfy=feed`.  

### Additional context

An extension to send summary notifications of the feed updates https://github.com/pe1uca/xExtension-ntfy  

Added new section below the "Information" section of the feed config.  
Image

Elgeryy1:
## Summary

Add an extension hook that lets extensions contribute `data-*` attributes to feed and category items in the sidebar tree.

## Changes

- Add the `FeedTreeItemAttributes` hook.
- Restrict output to `data-*` attributes.
- Add documentation and tests.

## Validation

- Targeted PHPUnit tests
- PHPCS
- PHPStan

Fixes #8472


Elgeryy1:
Closing as a duplicate -- I opened #9021 for the same issue, which also covers categories (not just feeds, per the issue title and Alkarex's comment) and uses a composable hook signature so multiple enabled extensions' attributes don't silently clobber each other. Consolidating review there. Sorry for the noise!

PR #9018, opened 2026-07-10
Fix Safari unread count repaint

### LINKED ISSUE #8249: [Bug] Unread count of feed in sidebar is slow to update
### Describe the bug

Whether I open a single feed, or a folder full of feeds, the unread count in the sidebar often takes a minute or two to update to the correct number as I read through the feed entries. 

However, if I move the mouse pointer over the list of feeds in the sidebar, each feed immediately updates to the current count as the mouse tracks over it.

Would it be possible to trigger a refresh of the sidebar information as entries are read?

This is not a new regression.

### To Reproduce

1. Open a feed
2. Observe the current unread count in the sidebar for that feed
3. Read a few entries
4. Observe that the the current unread count has not changed
5. Move the mouse cursor over the sidebar
6. Observe that the current unread count is now correct

### Expected behavior

_No response_

### FreshRSS version

1.27.1

### System information

- Database version: [e.g. SQLite, PostgreSQL 17]
- PHP version: [e.g. PHP 8.2]
- Web server type: [e.g. Apache, nginx]
- Device: [e.g. iPhone13]
- Browser: [e.g. Firefox 134]

### Additional context

_No response_

Elgeryy1:
## Summary

Force Safari to repaint the sidebar unread count after its `data-unread` value changes.

## Changes

- Add the minimal CSS transform workaround in the base and RTL themes.

## Validation

- ESLint
- Stylelint
- `git diff --check`

Safari itself was not available for direct testing.

Fixes #8249

### DIFF
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2308,6 +2308,7 @@ html.slider-active {
 	min-width: 2rem;
 	display: block;
 	content: attr(data-unread);
+	transform: translateZ(0);
 	position: absolute;
 	top: 0;
 	right: 0.75rem;
--- a/p/themes/base-theme/frss.css
+++ b/p/themes/base-theme/frss.css
@@ -2308,6 +2308,7 @@ html.slider-active {
 	min-width: 2rem;
 	display: block;
 	content: attr(data-unread);
+	/* Forces Safari to repaint this badge when `data-unread` changes; without it, Safari can leave the old count on screen until an unrelated repaint (e.g. hover). */
 	transform: translateZ(0);
 	position: absolute;
 	top: 0;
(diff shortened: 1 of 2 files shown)


Frenzie:
It's potentially worth noting that this always affects rendering (as in extra CPU / GPU time) while with a toggle via JS it'd only be at the time it's actually updated. 

In any case I'd add a brief comment explaining that this is for Safari.

Elgeryy1:
Good point on the CPU/GPU cost -- added a comment explaining the Safari repaint reasoning. I kept the CSS-only approach for now since it's the minimal fix for the reported symptom, but happy to switch to a JS-driven toggle (only paying the cost at update time) if you'd prefer that instead.

Frenzie (approved):

Elgeryy1:
Added a comment on the `transform: translateZ(0)` line explaining it's there to force Safari to repaint the unread badge. Should be ready for another look.

PR #9019, opened 2026-07-10
🤖 Ignore junk after closing RSS tag

### LINKED ISSUE #8326: [Feature] Ignore junk after closing RSS tag in XML

Cloudflare insights intermittently appends some HTML to RSS feeds unintentionally on some websites causing feed loading to fail.

Here is an example:
```
...
</rss>
 
<!--
Performance optimized by W3 Total Cache. Learn more: https://www.boldgrid.com/w3-total-cache/?utm_source=w3tc&utm_medium=footer_comment&utm_campaign=free_plugin
 
Object Caching 417/1500 objects using APC
Page Caching using Disk: Enhanced
 
Served from: ... @ 2025-12-17 11:27:25 by W3 Total Cache
--><a href="..." aria-hidden="true" rel="nofollow noopener" style="display: none !important; visibility: hidden !important"></a><script defer src="https://static.cloudflareinsights.com/beacon.min.js/..." integrity="..." data-cf-beacon='{"version":"2024.11.0","token":"...","server_timing":{"name":{"cfCacheStatus":true,"cfEdge":true,"cfExtPri":true,"cfL4":true,"cfOrigin":true,"cfSpeedBrain":true},"location_startswith":null}}' crossorigin="anonymous"></script>
 
```

### Use case

_No response_

Since the RSS data itself is fine, we could just load that and ignore any junk after the closing tag rather than inconveniencing the user with something only the site admin can fix.

_No response_

### Additional context

_No response_

Elgeryy1:
## Summary

Trim trailing junk after a valid RSS/Atom XML root so feeds wrapped by caches or proxies can still be parsed.

## Changes

- Trim trailing content after the closing XML root in `SimplePieFetch` for feed-like XML.
- Add regression coverage for junk after `</rss>` and for unchanged valid feeds.

## Validation

- PHPUnit targeted SimplePie tests
- PHPCS
- PHPStan

Fixes #8326


Elgeryy1:
Closing as a duplicate -- I opened #9020 for the same issue with broader test coverage (9 cases vs the RSS-only case here) and without mixing in unrelated files from #8472 (this branch's diff includes app/layout/aside_feed.phtml, HookType.php, etc., which don't belong to this issue). Consolidating review there. Sorry for the noise!

PR #9020, opened 2026-07-10
Ignore junk found after the closing root tag when fetching a feed

### LINKED ISSUE #8326: [Feature] Ignore junk after closing RSS tag in XML

Elgeryy1:
## Description

Some feeds (e.g. behind W3 Total Cache / Cloudflare beacon injection) append stray HTML/JS/comments after the closing root tag (`</rss>`, `</feed>`, or `</rdf:RDF>`), which breaks XML parsing even though the actual feed content is well-formed.

Adds a small helper in `FreshRSS_SimplePieFetch` that truncates the fetched body at the end of the *last* well-formed root closing tag (case-insensitive, arbitrary namespace prefix for RDF), only when there's non-whitespace content after it — a clean feed with no trailing junk is left untouched.

Fixes #8326

## Testing

`vendor/bin/phpcs`, `vendor/bin/phpstan analyse` (whole project), and the full `phpunit` suite: all clean/passing, including 9 new unit tests covering clean RSS/Atom/RDF feeds (no-op), the reported trailing-junk case, an Atom/RDF variant, and a feed whose content legitimately contains the substring `</rss>` earlier in the body (to confirm only the true last root close is used).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ See also [the FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases).
 	* Fix wrong player parent logic leading to invalid type [#8893](https://github.com/FreshRSS/FreshRSS/pull/8893), [simplepie#978](https://github.com/simplepie/simplepie/pull/978)
 	* Consistently enable `XML_OPTION_PARSE_HUGE` [#8894](https://github.com/FreshRSS/FreshRSS/pull/8894), [simplepie#977](https://github.com/simplepie/simplepie/pull/977)
 	* Fix null warning in IRI for PHP 8.5+ [#8918](https://github.com/FreshRSS/FreshRSS/pull/8918), [simplepie#979](https://github.com/simplepie/simplepie/pull/979)
+	* Ignore junk found after the closing root tag when fetching a feed [#8326](https://github.com/FreshRSS/FreshRSS/issues/8326)
 * Features
 	* New option to hide badges showing number of unread articles (*Phantom Obligation*) [#8844](https://github.com/FreshRSS/FreshRSS/pull/8844)
 	* New option to keep or not the custom sort order when navigating between categories and feeds [#8969](https://github.com/FreshRSS/FreshRSS/pull/8969)
--- a/app/Models/SimplePieFetch.php
+++ b/app/Models/SimplePieFetch.php
@@ -42,6 +42,32 @@ final class FreshRSS_SimplePieFetch extends \SimplePie\File
 		}
 
 		parent::__construct($url, $timeout, $redirects, $headers, $useragent, $force_fsockopen, $curl_options);
+
+		if ($this->success) {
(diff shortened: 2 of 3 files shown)


Alkarex:
I am not sure of this one. Supporting invalid XML might open another can of worms.
Furthermore, SimplePie is trying multiple encodings in case of error, and that behaviour would also have to be checked

Elgeryy1:
Kept trimming byte-preserving and added coverage for an ISO-8859-1 body plus a UTF-16 body that is intentionally left untouched for SimplePie to handle. The follow-up GitHub Actions run is in progress after fixing a spelling-check false positive in the test.

Elgeryy1:
Good points, let me address both:

**Invalid XML**: `removeContentAfterRootClosingTag()` is a no-op whenever no root closing tag (`</rss>`, `</feed>`, `</…:RDF>`) is found in the body at all (see the "no root closing tag at all" test case). It never tries to validate or repair XML — it only trims bytes that come *after* an already-present, well-formed closing tag. A feed that was unparseable before this change is left exactly as unparseable as before; this only helps the case where a perfectly valid feed has extra junk appended after it by a misbehaving proxy/cache.

**Encoding fallback**: this runs inside `FreshRSS_SimplePieFetch::__construct()` (which extends `\SimplePie\File`), i.e. directly on the raw fetched bytes, *before* SimplePie's own encoding detection/conversion runs on `$body`. The regex only matches contiguous ASCII byte sequences, so for encodings where the closing tag isn't representable that way (UTF-16 being the obvious case), the pattern simply won't match and the body is returned untouched — I added `test_removeContentAfterRootClosingTag_whenUtf16FeedCannotBeMatched_leavesTheBodyForSimplePieToHandle` to cover exactly that. So SimplePie's multi-encoding fallback logic downstream is unaffected either way.

PR #9022, opened 2026-07-10
Allow editing the feed title when adding a new feed

### LINKED ISSUE #2826: Change title before add
When adding a new rss feed [you cannot change](https://upload.vaa.red/jZya3#9dd3995f563a90949ad130d440b2e6ec) the title before confirming the new entry. 
Only [after adding](https://upload.vaa.red/7nKwt#5b219d48cfd5a87acf64d1d9c77b7350) it is possible to do this.

I think it's something that could be improved.

Elgeryy1:
## Description

The direct/bookmarklet "add feed" flow shows the feed title as a read-only label, while the normal subscribe flow already lets you edit it. Makes the title an editable field in the add-feed form too, matching the existing subscribe form's pattern -- if left blank, it still falls back to the feed's own detected title as before.

Fixes #2826

## Testing

`vendor/bin/phpcs` and `vendor/bin/phpstan analyse` (whole project): no errors.


JamBalaya56562:
@Elgeryy1 
Could you resolve merge conflicts?

Elgeryy1:
Rebased onto current edge and resolved the merge conflict (two unrelated CHANGELOG entries landed on the same line).

JamBalaya56562:
Thanks for working on this. If possible, could you please add before-and-after screenshots of the feed-add confirmation form shown after the feed URL has been detected?

It would be helpful to show how the feed title changes from a read-only label to an editable input field. This would make the user-facing effect of the change immediately clear.

Elgeryy1:
Before (the feed title was a read-only label — you couldn't change it before adding the feed, only after via Edit feed):

9022-before-readonly-title

After (the title is now an editable input, pre-filled with the detected name, so you can rename it right away):

9022-editable-title

PR #9025, opened 2026-07-10
Use :user-invalid instead of :invalid for form field styling

### LINKED ISSUE #3647: [BUG] New user: input is always red, when empty

Elgeryy1:
## Description

Required form fields (e.g. the admin "Add user" username field) show a red border immediately on page load, before the user has interacted with them, because every theme styles the bare `:invalid` pseudo-class unconditionally.

This was attempted before in #3724, but was closed over a browser-support concern for `:user-invalid` (only matches after interaction, not on load). That concern is resolved now -- `:user-invalid` has ~89% global support across all evergreen browsers.

Fixes #3647

## Testing

`.rtl.css` files regenerated via `npm run rtlcss`, `npm run stylelint`: no errors. Verified in a local instance: the add-user form shows no red border on load, turns red after leaving an invalid field, clears when corrected.

### DIFF
--- a/p/themes/Ansum/_forms.css
+++ b/p/themes/Ansum/_forms.css
@@ -94,6 +94,12 @@ input:focus, select:focus, textarea:focus {
 	border-color: var(--main-first);
 }
 
+input:-moz-ui-invalid, select:-moz-ui-invalid {
+	color: var(--alert-bg);
+	border-color: var(--alert-bg);
+	box-shadow: none;
+}
+
 input:user-invalid, select:user-invalid {
 	color: var(--alert-bg);
 	border-color: var(--alert-bg);
--- a/p/themes/Ansum/_forms.css
+++ b/p/themes/Ansum/_forms.css
@@ -94,12 +94,6 @@ input:focus, select:focus, textarea:focus {
 	border-color: var(--main-first);
 }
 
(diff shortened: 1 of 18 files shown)


Alkarex:
Could you please test with SeaMonkey? (Which I like to use for ensuring a minimal support for older engines)

math-GH:
Just for documentation that it is now widley useable: https://caniuse.com/?search=%3Auser-invalid

Elgeryy1:
@Alkarex Tested with SeaMonkey 2.53.23: `:user-invalid` does not parse there, while `:-moz-ui-invalid` parses and matches after `form.reportValidity()`; the legacy rule is visibly applied. Keeping the selectors as separate rules is therefore necessary for SeaMonkey, while modern browsers use the standard selector. Stylelint and GitHub Actions are passing.

Alkarex:
Ok, thanks. I have ended up removing the legacy `-moz-ui-invalid` as I would rather avoid vendor-specific code, especially when not so important.

Alkarex (approved):

PR #9026, opened 2026-07-10
Mailer: allow disabling SMTPAutoTLS via config

### LINKED ISSUE #2997: smtp mailer and SMTPAutoTLS
Hi, 

I am using a protonmail bridge as my smtp mailer. It is using a self-signed certificate and smtp mailer attempts SMTPAutoTLS regardless of 'secure' settings in config.php. 

I would like to either:
1. be able to point at the certificate and trust it for STARTTLS
2. be able to disable cert verification
3. be able to set SMTPAutoTLS => false

Thank you

Elgeryy1:
## Description

FreshRSS's mailer never sets PHPMailer's `SMTPAutoTLS`, so its default (`true`) always applies regardless of the documented `secure => ''` setting -- meaning opportunistic TLS can't actually be disabled through configuration as documented.

Adds a `smtp.auto_tls` config option (defaulting to `true`, preserving current behavior) that's threaded through to PHPMailer.

Scoped narrowly to this -- the custom root CA / Docker-image suggestion also raised in the issue is left out of scope.

Fixes #2997

## Testing

`vendor/bin/phpcs` and `vendor/bin/phpstan analyse` (whole project): no errors. Verified against a local SMTP test server with a self-signed cert that `auto_tls=false` sends unencrypted while the default preserves today's opportunistic-TLS behavior.

### DIFF
--- a/config.default.php
+++ b/config.default.php
@@ -211,6 +211,7 @@ return [
 		'username' => '',
 		'password' => '',
 		'secure' => '', // '', 'ssl' or 'tls'
+		'auto_tls' => true, // maps to PHPMailer’s `SMTPAutoTLS`; set to false to disable opportunistic STARTTLS, e.g. when using a self-signed certificate
 		'from' => 'root@localhost',
 	],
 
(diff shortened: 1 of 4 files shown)


Alkarex (approved):

Alkarex:
Thanks :-)

PR #9028, opened 2026-07-10
Migrate core codebase to Composer PSR-4 autoloading under FreshRss namespace

ColonelMoutarde:
Replace the hand-rolled classAutoloader() convention (FreshRSS_X, Minz_X global classes) with proper Composer PSR-4 autoloading under the FreshRss namespace, mirroring the existing directory layout:

- lib/Minz/          -> FreshRss\Minz\...
- app/Models/         -> FreshRss\Models\...
- app/Exceptions/      -> FreshRss\Exceptions\...
- app/Services/        -> FreshRss\Services\...
- app/Utils/           -> FreshRss\Utils\... (files renamed to StudlyCaps)
- app/Mailers/         -> FreshRss\Mailers\...
- app/Controllers/     -> FreshRss\Controllers\... (files renamed to
  StudlyCaps; Minz_Dispatcher now resolves core controllers via FQCN while
  keeping the FreshExtension_<name>_Controller convention for extensions)

Backward compatibility for third-party extensions is preserved via lib/legacy-aliases.php, which lazily aliases the classes that make up the documented extension API (Minz_Extension, FreshRSS_Entry, etc.) back to their historical global names. The old classAutoloader() fallback is kept registered alongside Composer's autoloader for anything not covered by the new namespace yet.

All inline fully-qualified class references were converted to `use` imports across app/, lib/, cli/, p/, and tests/.

Verified with php-lint, PHPStan (level 10), PHPUnit (652 tests), PHPCS, and manual smoke testing of the running app (all 17 controllers, feed actualization, and a throwaway third-party extension using only legacy class names through the real hook system).


lib/lib_rss.php
```diff
@@ -38,6 +49,15 @@ function join_path(...$path_parts): string {
 }
 
 //<Auto-loading>
+
+// Composer PSR-4 autoloading for the `FreshRss` namespace (app/, lib/Minz/), when available.
+// Not required in production yet: no runtime Composer package is mandatory, so `vendor/`
+// is not currently shipped in releases or Docker images. Falls back to the legacy
+// autoloader below when absent (e.g. a release archive without `composer install`).
+if (is_file(FRESHRSS_PATH . '/vendor/autoload.php')) {
+	require_once FRESHRSS_PATH . '/vendor/autoload.php';
```
Alkarex: I would rather not depend on `/vendor/autoload.php` for anything else than loading development libraries. The code is horrible

app/Controllers/AuthController.php
```diff
@@ -0,0 +1,292 @@
...
+use FreshRss\Minz\Translate;
+use FreshRss\Minz\Url;
+use FreshRss\Minz\User;
+use FreshRss\Models\ActionController;
+use FreshRss\Models\Auth;
+use FreshRss\Models\Context;
+use FreshRss\Models\FormAuth;
+use FreshRss\Models\UserDAO;
+use FreshRss\Models\View;
+use FreshRss\Utils\HttpUtil;
```
Alkarex, line 19: I think this is ridiculously verbose. At least something like `use FreshRss\Models as Models;` and then `Models\ActionController` in code

Alkarex:
I am generally quite negative about such change. Several thousand more lines for something that ends up being slower and depending on unreadable third-party code.
That being said, I think we could at some point gradually work on the namespace part (not the PSR autoloading), probably starting with Minz, which makes more sense as a library

app/Controllers/AuthController.php
ColonelMoutarde: PSR-4 only governs the namespace→directory mapping for autoloading; it says nothing about use statement style. The relevant standard here is PSR-12 chapter 4.1, which explicitly requires "one use keyword per declaration" and our phpcs.xml enforces the full PSR12 ruleset without excluding the imports sniff, so this per-class style is already what our own tooling mandates.                                                                                                                                                                                                                                                                                                      
Also, in practice ActionController, Auth, etc. are referenced 15 times in this file's body. Aliasing the namespace (Models\ActionController) would just move the verbosity from a one-time import block into every call site throughout the file, which is worse for readability overall, not better.

PR #9031, opened 2026-07-11
Dispatch a JS event when an entry finishes being marked read/unread

### LINKED ISSUE #8862: [Feature] Hook "mark as (un)read complete" for extensions

I'm using the [ThreePanesView extension](https://framagit.org/nicofrand/xextension-threepanesview). When marking an article as unread, the unread icon is changed to a loading spinner, but [never gets replaced](https://framagit.org/nicofrand/xextension-threepanesview/-/work_items/28) when the ajax request is done.

`main.js#send_mark_read_queue()` resets the icon, but only in the article list - not in the third pane.

### Use case

The extension should be able to register a hook that gets called when an article has been marked read or unread. Then it can restore the icons itself.

I don't know the internals, so I can't suggest a good solution.

_No response_

### Additional context

_No response_

Elgeryy1:
## Summary

Fixes #8862.

Extensions that render their own read/unread indicator outside of the main article list (e.g. a custom reading pane, like the ThreePanesView extension mentioned in the issue) have no way to know when a mark read/unread AJAX request completes, so their own UI can get stuck (e.g. a spinner that never resets).

This dispatches a `freshrss:entryStateChange` event on `document` in `send_mark_read_queue()` in `p/scripts/main.js`, once the read/unread state has actually been toggled for an entry. The event's `detail` contains `id` (entry id) and `isRead` (new state).

This follows the naming and dispatch convention already used for the other undocumented/documented `freshrss:*` events in the same file (`freshrss:globalContextLoaded`, `freshrss:openArticle`, `freshrss:load-more`), and covers every code path that toggles read state through this function: single-entry marking, batched marking, and keyboard-shortcut/auto-mark-on-scroll actions all funnel through `send_mark_read_queue()`.

Also documented the new event in `docs/en/developers/03_Backend/05_Extensions.md` and added a changelog entry.

## Test plan

- [x] `npm run eslint` passes on `p/scripts/main.js`
- [x] `npm run markdownlint` passes on the changed docs/changelog files
- [ ] Manually verified in a browser that the event fires with the correct `id`/`isRead` when marking an entry read/unread (single and batch)

### DIFF
--- a/docs/en/developers/03_Backend/05_Extensions.md
+++ b/docs/en/developers/03_Backend/05_Extensions.md
@@ -225,6 +225,7 @@ if (document.readyState && document.readyState !== 'loading' && typeof window.co
 The following events are available:
 
 * `freshrss:globalContextLoaded`: will be dispatched after load the global `context` variable, useful for referencing variables injected with the `Minz_HookType::JsVars` hook.
+* `freshrss:entryStateChange`: will be dispatched on `document` after an entry has finished being marked as read or unread (i.e. once the underlying AJAX request has completed). The `detail` property of the event contains `id` (the entry ID) and `isRead` (the new read state). Useful for extensions that display their own read/unread indicator outside of the main article list, e.g. in a custom reading pane.
 
 ### Injecting CDN content
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ See also [the FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases).
 * Features
 	* New option to hide badges showing number of unread articles (*Phantom Obligation*) [#8844](https://github.com/FreshRSS/FreshRSS/pull/8844)
 	* New option to keep or not the custom sort order when navigating between categories and feeds [#8969](https://github.com/FreshRSS/FreshRSS/pull/8969)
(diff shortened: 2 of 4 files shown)


Alkarex:
It is probably better to merge https://github.com/FreshRSS/FreshRSS/pull/9035 first?

docs/en/developers/03_Backend/05_Extensions.md
```diff
@@ -225,6 +225,7 @@ if (document.readyState && document.readyState !== 'loading' && typeof window.co
 The following events are available:
 
 * `freshrss:globalContextLoaded`: will be dispatched after load the global `context` variable, useful for referencing variables injected with the `Minz_HookType::JsVars` hook.
+* `freshrss:entryStateChange`: will be dispatched on `document` after an entry has finished being marked as read or unread (i.e. once the underlying AJAX request has completed). The `detail` property of the event contains `id` (the entry ID) and `isRead` (the new read state). Useful for extensions that display their own read/unread indicator outside of the main article list, e.g. in a custom reading pane.
```
Alkarex, line 228: Note for another PR: it would be nice to add missing events such as `freshrss:openArticle`

Alkarex (approved):

PR #9035, opened 2026-07-11
fix(js): use Event() constructor instead of deprecated initEvent()

### LINKED ISSUE #5152: JavaScript: initEvent() deprecated
https://github.com/FreshRSS/FreshRSS/blob/4f957dfc4c38c418e5b46a2dc8d34a9bdabb31db/p/scripts/main.js#L393-L394

https://github.com/FreshRSS/FreshRSS/blob/4f957dfc4c38c418e5b46a2dc8d34a9bdabb31db/p/scripts/main.js#L1745-L1746

initEvent() is marked as deprecated:
https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent

TowyTowy:
Fixes #5152

`document.createEvent('Event')` + `Event.initEvent()` are deprecated (https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent). This replaces the two remaining usages in `p/scripts/main.js` with the standard `new Event(name, { bubbles, cancelable })` constructor:

- `freshrss:openArticle`
- `freshrss:load-more`

Both `bubbles` and `cancelable` are kept `true`, exactly matching the previous `initEvent('…', true, true)`, so dispatch behaviour is unchanged (verified with a jsdom A/B: same type, bubbling to `document`, cancelable). This mirrors the pattern already used for `freshrss:slider-load` in `p/scripts/extra.js`, and follows @Frenzie's suggestion on the issue.

`p/scripts/` no longer uses jQuery, so the old concern about jQuery only catching the legacy syntax no longer applies — these events are dispatched to and listened for via native `addEventListener`.

## How to test manually

- Open an article in the normal/reader view (dispatches `freshrss:openArticle`).
- Scroll to load more articles (dispatches `freshrss:load-more` on `document.body`).
- Both continue to work with no console errors.

## Pull request checklist

- [x] clear commit messages
- [x] `eslint` clean
- [x] code manually tested

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -439,8 +439,7 @@ function mark_favorite(div) {
 	}));
 }
 
-const freshrssOpenArticleEvent = document.createEvent('Event');
-freshrssOpenArticleEvent.initEvent('freshrss:openArticle', true, true);
+const freshrssOpenArticleEvent = new Event('freshrss:openArticle', { bubbles: true, cancelable: true });
 
 function loadLazyImages(rootElement) {
 	rootElement.querySelectorAll('img[data-original], iframe[data-original], video[data-original], track[data-original]').forEach(function (el) {
@@ -2247,8 +2246,7 @@ function load_more_posts() {
 	req.send();
 }
 
-const freshrssLoadMoreEvent = document.createEvent('Event');
-freshrssLoadMoreEvent.initEvent('freshrss:load-more', true, true);
+const freshrssLoadMoreEvent = new Event('freshrss:load-more', { bubbles: true, cancelable: true });
 
 function init_load_more(box) {
 	box_load_more = box;


Alkarex:
Thanks. Could you please check:
* That is works with SeaMonkey (which I like to use as a test for older engines)
* The degree of compatibility with possible extensions, which are using the old syntax to subscribe to these events

TowyTowy:
Good questions — checked both:

**Older engines / SeaMonkey:** the `Event()` constructor has been supported in Gecko since the Firefox 11 era (2012), so every currently-maintained SeaMonkey (2.53.x is Firefox 60 ESR-based) handles it. The only engine that lacks `new Event()` is Internet Explorer (Trident), which *only* has `createEvent`/`initEvent` — but that's the reverse concern and IE isn't in FreshRSS's supported set. I don't have SeaMonkey here for a live screenshot, but there's no Gecko version in a plausible support window that would break. Worth noting FreshRSS already relies on this constructor: `extra.js` uses `new Event('freshrss:slider-load')` — if SeaMonkey couldn't handle `new Event()`, that would already be broken.

**Extension compatibility:** subscribing is done with `addEventListener('freshrss:openArticle', …)`, which is independent of how the event is constructed. The object dispatched by `new Event(name, { bubbles: true, cancelable: true })` is behaviourally identical to the old `initEvent(name, true, true)` one — same `type`, `bubbles`, and `cancelable` (I verified this in a jsdom A/B: both bubble to `document` and are cancelable). Because `bubbles` stays `true`, a jQuery `.on(...)` handler on an ancestor still catches it too. I also searched GitHub code for `freshrss:openArticle` / `freshrss:load-more`: the only hits are FreshRSS itself and forks/mirrors — no third-party extension subscribes to these — so in practice nothing external hooks them, and anything that did (via `addEventListener` or jQuery `.on`) keeps working unchanged.

If you have a specific extension in mind that dispatches or listens for these differently, point me at it and I'll check it directly.

Alkarex (approved):

PR #9036, opened 2026-07-11
Fix SQL errors breaking regular-expression search on MySQL / MariaDB

TowyTowy:
Regular-expression searches are broken for all MySQL and MariaDB users (the compressed `content_bin` path, `isCompressed() === true`): three of the four regex constructions in `FreshRSS_EntryDAO::sqlBooleanSearch()` generate invalid SQL, the query fails, and the search silently returns zero articles (the SQL error only appears in the user log).

* `intext:/regex/` appends a stray closing parenthesis after `sqlRegex(...)` → MySQL error 1248 *Every derived table must have its own alias*
* `-intext:/regex/` has the same stray parenthesis
* `-/regex/` emits `' ANT NOT '` instead of `' AND NOT '` → MySQL error 1064

`sqlRegex()` already returns a balanced expression (`REGEXP_LIKE(expr,?,'flags')` on MySQL, `expr REGEXP ?` on MariaDB), so no extra parenthesis is needed — matching the SQLite/PostgreSQL branches, which are correct. This also affects filter actions and API searches using regexes, not just the search box.

The existing MySQL/MariaDB unit tests for `intext:/regex/` had captured the buggy unbalanced SQL as their expected strings; they are corrected, and new cases are added for `-intext:/regex/` and `-/regex/` on both flavours.

Tested live on MySQL 8 (FreshRSS edge Docker + real feed, 10 articles of which 8 contain "highlights"):

| search | before | after |
|---|---|---|
| `intext:/highlights/` | 0 articles + SQL error 1248 | 8 articles |
| `-/highlights/` | 0 articles + SQL error 1064 | 2 articles |
| `-intext:/highlights/` | 0 articles + SQL error 1248 | 2 articles |
| `intext:/HIGHLIGHTS/i` vs `/HIGHLIGHTS/` | — | 8 vs 0 (flags OK) |

`tests/app/Models/SearchTest.php`: 253 tests, 513 assertions OK. Found by code audit; written with the help of AI (Claude), personally reviewed and live-tested.

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -1226,7 +1226,7 @@ class FreshRSS_EntryDAO extends Minz_ModelPdo {
 			if ($filter->getIntextRegex() !== null) {
 				if (static::isCompressed()) {	// MySQL-only
 					foreach ($filter->getIntextRegex() as $content) {
-						$sub_search .= 'AND ' . static::sqlRegex("UNCOMPRESS({$alias}content_bin)", $content, $values) . ') ';
+						$sub_search .= 'AND ' . static::sqlRegex("UNCOMPRESS({$alias}content_bin)", $content, $values) . ' ';
 					}
 				} else {
 					foreach ($filter->getIntextRegex() as $content) {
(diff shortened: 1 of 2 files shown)


Alkarex (approved):

Alkarex:
Thanks! Good catch

PR #9037, opened 2026-07-12
i18n: Improve Traditional Chinese translations

IEEE-754:
**Change(s)**
- Add missing punctuation
- Add missing spaces in the date format
- Add some new translations

**File(s) changed**
`app/i18n/zh-TW/{admin,conf,gen,install,sub}.php`
`README.md`
`README.fr.md`

### DIFF
--- a/app/i18n/zh-TW/sub.php
+++ b/app/i18n/zh-TW/sub.php
@@ -94,7 +94,7 @@ return array(
 			'html_json' => array(
 				'_' => 'HTML + XPath + JSON 點表示法 (HTML 中的 JSON)',
 				'xpath' => array(
-					'_' => 'XPath for JSON in HTML',	// TODO
+					'_' => 'XPath for JSON in HTML',	// IGNORE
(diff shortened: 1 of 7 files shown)


app/i18n/zh-TW/sub.php
```diff
@@ -94,7 +94,7 @@
 			'html_json' => array(
 				'_' => 'HTML + XPath + JSON 點表示法 (HTML 中的 JSON)',
 				'xpath' => array(
-					'_' => 'XPath for JSON in HTML',	// TODO
+					'_' => 'XPath for JSON in HTML',	// IGNORE
```
Alkarex, line 97: Is the ignore on purpose? Maybe the *for* and *in*?

IEEE-754: Yes, this was intentionally marked as `// IGNORE`.
I know the translation would be **"HTML 中適用於 JSON 的 XPath"**, 
but it does not sound natural. In fact no native speaker would say it that way.
It sounds like a machine translation, so I chose to mark it as not requiring translation.

PR #9039, opened 2026-07-12
Use key code instead of key value for hotkeys

### LINKED ISSUE #7954: [Feature] Use "key code" for the hotkeys, instead of the actual letters

Currently, when you set a hotkey for navigation (`j`/`k`/etc.) — they are processed as actual letters of the alphabet. This creates problems for users using multiple different languages on their computers. E.g. Latin letter `j` and Cyrillic letter `о` are on the same key on the keyboard, but navigation in FreshRSS will work only for the former, when the currently active layout is the Latin-based one.

### Use case

Changing the behaviour will allow users not to care about the currently selected language, and they will always be able to navigate in FreshRSS.

Rely on `event.code` for the shortcuts, instead of the `event.key` as it currently is. E.g. both Latin `j` and Cyrillic `о` will trigger an event with `code="KeyJ"`.

_No response_

### Additional context

_No response_

lcd1232:
Fixes #7954

## Problem

Keyboard shortcuts are matched against `event.key`, i.e. the character the key produces. That character depends on the active keyboard layout, so as soon as I switch to a non-Latin layout (Russian in my case) the shortcuts stop working — physical `J` produces `о`, which no longer matches the configured `j`.

## Change

Match shortcuts against `event.code` (the physical key position) instead, as suggested in the issue. `KeyJ` maps to `j` regardless of whether the layout produces `j`, `о`, or anything else, so navigation keeps working on any layout.

I pulled the normalization into a small helper `shortcutKeyFromEvent(ev)` (`p/scripts/shortcut.js`) that both `main.js` and `global_view.js` now use, so there's a single place that turns a keydown into the token we compare against. It returns the same uppercase tokens the matcher already expected (`J`, `3`, `ARROWDOWN`, `SPACE`, …), and falls back to the old `event.key` behaviour for browsers that don't populate `event.code`.

Stored config keeps the existing friendly tokens (`j`, `space`, `1`, …), so there's **no migration** and the shortcut settings page / server-side validation are untouched.

## Notes / trade-offs

- This is technically a behaviour change for people on AZERTY/QWERTZ who relied on the shortcut following the printed letter rather than the physical position (raised by @Frenzie in the issue). All the *default* shortcuts are on letter/number keys, which stay put, so in practice the common cases are unaffected.
- Matching is positional, so only keys in the existing allow-list (a–z, 0–9, function keys, named keys) are bindable — punctuation-position keys aren't, same as before.

## Testing

- Added unit tests for the helper (`tests/scripts/shortcut.test.js`) run with the built-in Node test runner (`npm run test:js`) — covers letter/digit/numpad/arrow/named keys and the `event.key` fallback, including the Cyrillic case (`KeyJ` + `key: 'о'` → `J`).
- Manually verified in the app that `j`/`k`/`r`/`f` and the view/escape shortcuts work on both a US layout and a Russian layout.


Frenzie:
> for people on AZERTY/QWERTZ

That's just a common example, but of course it also applies to Dvorak, Bépo, etc.

I'm not opposed, but I'd still prefer to add a switch and perhaps some other UI refinements. But that can be done in a follow-up. (For some reason `git fetch upstream pull/9093/head` doesn't seem to be working atm. [The docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally) still list it though. Quite strange.)

lcd1232:
@Frenzie it's because this PR id is 9039, not 9093

Frenzie:
Haha, I think I actually copy pasted the ID (in the terminal, not when typing it here) but maybe I didn't.

Frenzie:
I pushed a concept of the ability to do both.

app/Controllers/configureController.php
```diff
@@ -330,6 +334,7 @@ public function shortcutAction(): void {
 		}
 
 		FreshRSS_View::prependTitle(_t('conf.shortcut.title') . ' · ');
+		FreshRSS_View::appendScript(Minz_Url::display('/scripts/shortcut-test.js?' . @filemtime(PUBLIC_PATH . '/scripts/shortcut-test.js')));
```
Alkarex, line 337: To be removed in a final PR

Alkarex: Ah, sorry, I thought it was for debugging

Frenzie: Oops, I forgot to change the name. First it just popped up the result but then I realized that was a rather unnecessary extra step.

PR #9041, opened 2026-07-13
Update German translations

Makario1337:
Closes #

1. Not needed, just updated translations.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [ ] documentation updated

### DIFF
--- a/app/i18n/de/admin.php
+++ b/app/i18n/de/admin.php
@@ -92,12 +92,12 @@ return array(
 		'force_email_validation' => 'E-Mail-Adressprüfung erzwingen',
 		'instance-name' => 'Bezeichnung',
 		'internal-host-allowlist' => array(
-			'_' => 'Internal host allowlist',	// TODO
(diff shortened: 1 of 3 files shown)


Frenzie (approved):

math-GH:
Veto. It is not a good German translation. Was it translated with AI or another technique?

math-GH (changes requested):
veto at all.

Frenzie:
(I mainly just meant the PHP looked good, though nothing particularly jumps out to me other than that I wouldn't be inclined to say "von der … festgelegt" and yet I wouldn't be surprised if you told me that's the most idiomatic German way of saying it. :-)

Makario1337:
Why should I translate it using AI if I am a native german speaker???? I was just unsure how to translate feeds since there is no real word for it in german. So I used the word "Quelle" which means source and it is the closest I could find...

Frenzie:
I'd overlooked that. Please try to check the current translations since it's better to be consistent with them. I think that currently they're all translated as `Feed(s)`.

Makario1337:
Nope, I won't do that; you can do it yourself... It's a pity that someone who is using his spare time to commit to an open-source project gets called out using AI, even this isn't the case here... If you had behaved a bit nicer, I would have done so.

But since @math-GH is a German native, he can do it by himself :)

Alkarex:
@Makario1337 Sorry for the bad experience. With the generalisation of AI, it has become difficult to assess whether some work, especially from recent contributors, is automated or not. And while not a German speaker myself, I have the impression that there might be some variations of preferences between native speakers. Maybe due to e.g. regional differences?

PR #9043, opened 2026-07-14
Fix: infinite recursion in sanitizeHTML() when maxLength truncation doesn't converge

ColonelMoutarde:

Sanitizing can grow a truncated HTML fragment (e.g. an unclosed tag gets auto-closed), so the previous recursive shrink-and-retry could hit a fixed point and recurse forever, crashing on a stack overflow. Bound the retries and fall back to a hard truncation that always terminates.

Add unit tests for FreshRSS_SimplePieCustom::sanitizeHTML(), the XSS sanitization applied to all untrusted feed content, which had no coverage.

Pull request checklist:

- [X] clear commit messages
- [X] code manually tested
- [X] unit tests written (optional if too hard)
- [ ] documentation updated


ColonelMoutarde:
Please see this PR.

app/Models/SimplePieCustom.php
```diff
@@ -286,26 +286,36 @@ public static function sanitizeHTML(string $data, string $base = '', ?int $maxLe
...
 			//Sanitizing has made the result too long so try again shorter
-			$data = mb_strcut($result, 0, (2 * $maxLength) - strlen($result) - 2, 'UTF-8');
-			return self::sanitizeHTML($data, $base, $maxLength);
+			$data = mb_strcut($result, 0, max(0, (2 * $maxLength) - strlen($result) - 2), 'UTF-8');
+			if ($data === '') {
+				break;
+			}
 		}
-		return $result;
+		//Could not converge to the requested length: fall back to a hard truncation that is guaranteed to terminate
```
Frenzie: Nit.
```suggestion
		// Could not converge to the requested length: fall back to a hard truncation that is guaranteed to terminate.
```

Frenzie (approved):
I suppose it doesn't matter in practice since it'll be really fast, but is there an actual difference beyond two or three?

Alkarex:
I finally took the time to address some of the different cases I had in mind, so I made a little refactoring: Avoid two mb_strcut in a row, more cases for fallback, better fallback, more edge cases

I am not quite done, as there are still a few strange cases.

Alkarex:
Now I have also covered the cases of trailing incomplete tag or entity

Alkarex (approved):

PR #9047, opened 2026-07-15
Allow selecting a category in feed add URLs

### LINKED ISSUE #4160: Specify Group/Category via Bookmarklet
I scoured the documentation and couldn't find any way to supply a group/category as a parameter for a feed with the bookmarklet. 

Is there a way to add a feed via the bookmarklet and specify the category of the feed as a parameter in the URL?

Elgeryy1:
## Summary

- Honour the `category` query parameter when opening the add-feed form.
- Keep the existing `cat_id` parameter as a backwards-compatible fallback.
- Use `category` for the add-feed links in the subscription view.

This makes the URL proposed in #4160 select the intended category:
`?c=feed&a=add&category=1&url_rss=https://example.com/feed.xml`.

## Validation

- PHP syntax checks for the changed controller and templates.
- Focused request-parameter checks for explicit `category`, legacy `cat_id`, and the default category.

Fixes #4160


Alkarex:
I have simplified the approach a bit, including small fixes related to avoiding renaming `cat_id`.
Double-check welcome

Alkarex (approved):

PR #9049, opened 2026-07-15
Refactor user query translation keys

### LINKED ISSUE #6315: i18n: Query strings (refactoring)

Elgeryy1:
## What changed

- replace dynamic user-query translation keys with explicit state, order, and type lookups
- retain the existing translations and query behaviour without adding dependencies

## Why

The translation checker cannot identify keys assembled at runtime. Explicit lookups make all user-query strings visible to validation and easier to maintain.

## Validation

- `php -l app/views/configure/queries.phtml`
- `vendor/bin/phpcs app/views/configure/queries.phtml -s`
- `vendor/bin/phpstan analyse --memory-limit 512M --no-progress app/views/configure/queries.phtml`
- `vendor/bin/phpunit --bootstrap ./tests/bootstrap.php ./tests/app/Models/UserQueryTest.php --no-progress`
- verified `cli/check.translation.php --language=en --display-result` reports no unused `conf.query.*` key

Fixes #6315


app/views/configure/queries.phtml
```diff
@@ -137,6 +137,26 @@
 	</form>
 
 
+	<?php
+	$stateTranslations = [
+		0 => _t('conf.query.state_0'),
```
Frenzie: state_0 or state_all, state_read, state_unread, etc.? You could otherwise keep the array with `getState()`. The  translation files would need updating in that case but that doesn't strike me as an issue.

NB Not a rhetorical question.

Elgeryy1: Implemented this with semantic keys (`state_all`, `state_read`, `state_unread`, `state_favorite`, and `state_not_favorite`) and an explicit `getState()` lookup. The translation tables were regenerated across all locales; the targeted PHPUnit test, PHPCS, PHPStan, and translation generation pass.

app/views/configure/queries.phtml
```diff
@@ -137,6 +137,15 @@
 	</form>
 
 
+	<?php
+	$stateTranslations = [
+		7 => _t('conf.query.state_favorite'),
+		11 => _t('conf.query.state_not_favorite'),
+		13 => _t('conf.query.state_read'),
+		14 => _t('conf.query.state_unread'),
+		15 => _t('conf.query.state_all'),
```
Frenzie, line 157: Is this supposed to look like it's missing 11 states? Besides that it looks better to me.

Elgeryy1: Good catch. `getState()` can return 9 effective bitmask values: all, read, unread, favorite, not-favorite, read+favorite, unread+favorite, read+not-favorite, and unread+not-favorite. I added the four missing combination labels, mapped all 9 values explicitly, regenerated every locale, and verified PHPUnit, PHPCS, PHPStan, and translation generation.

app/views/configure/queries.phtml
```diff
@@ -137,6 +137,19 @@
...
+	$stateTranslations = [
+		5 => _t('conf.query.state_read_favorite'),
+		6 => _t('conf.query.state_unread_favorite'),
+		7 => _t('conf.query.state_favorite'),
+		9 => _t('conf.query.state_read_not_favorite'),
+		10 => _t('conf.query.state_unread_not_favorite'),
+		11 => _t('conf.query.state_not_favorite'),
+		13 => _t('conf.query.state_read'),
+		14 => _t('conf.query.state_unread'),
+		15 => _t('conf.query.state_all'),
```
Frenzie, line 157: Apparently this requires you to activate your own brain rather than the bot's since regardless how many single individual strings there are, `$query->getState()` can evidently return any number from 0 to 15 or it wouldn't have been written the way it's written.

Frenzie (approved):
Alrighty, lgtm (but didn't test)

app/views/configure/queries.phtml
Elgeryy1: Addressed in 543c8d4df: the mapping now covers every effective getState() value from 0 through 15.

PR #9050, opened 2026-07-15
Use native DOM CSS selectors when available

### LINKED ISSUE #6830: Take advantage of PHP 8.4+ native CSS selectors
https://wiki.php.net/rfc/dom_additions_84
Change our current CSS selector code (to filter content) to take advantage of PHP 8.4+ `querySelectorAll()` function.

Test for availability of `querySelectorAll()` and if available, use it instead of our code based on `PhpGt/CssXPath` (keep it as fallback as long as we support PHP 8.3-).

Elgeryy1:
## Summary

- Use PHP 8.4's native `Dom\HTMLDocument` CSS selector APIs when available.
- Keep the existing `PhpGt/CssXPath` implementation for PHP 8.1–8.3.
- Route entry extraction and content-filter selectors through the compatibility layer.

The legacy `DOMDocument` used today does not implement `querySelectorAll()`. The compatibility layer creates the modern HTML document only when the native APIs exist; otherwise it retains the current parser and CSS-to-XPath translation path.

Fixes #6830

## Validation

- Focused PHPUnit tests for native selectors, the legacy fallback, and entry filtering
- PHPCS across the repository
- PHPStan across the repository


Elgeryy1:
Resolved the merge conflict with current edge via normal merge commit ff0c54098 (no rebase/force-push). The resolution keeps the native-selector implementation and preserves the newer no-match/no-content diagnostics from edge.

PR #9053, opened 2026-07-16
Fix extension settings cancel action

### LINKED ISSUE #3662: [BUG] Manage extension: Cancel button does not close the settings panel
**Describe the bug**

**To Reproduce**

1. go to manage extensions
2. open the extension settings panel
3. click on cancel button

**Expected behavior**
click on cancel button closes the settings pannel

**Screenshots**
grafik

 - Browser: Firefox 89
 - FreshRSS version: [e.g. 1.17.1] + customCSS extension

Elgeryy1:
Closes #3662. Closes the extension settings slider when its reset/cancel control is used while preserving unsaved-change confirmation. Tests: php -l, node --check, git diff --check.

### DIFF
--- a/app/views/helpers/extension/configure.phtml
+++ b/app/views/helpers/extension/configure.phtml
@@ -8,7 +8,7 @@
 		throw new FreshRSS_Context_Exception('Extension not initialised!');
 	}
 ?>
-<div class="post">
+<div class="post extension-configure">
 	<h2>
 		<?= $this->extension->getName() ?> (<?= $this->extension->getVersion() ?>) —
 		<?= $this->extension->isEnabled() ? _t('admin.extensions.enabled') : _t('admin.extensions.disabled') ?>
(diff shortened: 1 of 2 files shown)


Alkarex:
Related: https://github.com/FreshRSS/FreshRSS/pull/8615

Alkarex:
In my quick test (Firefox), that does not seem to work. Double-check welcome

Alkarex:
If you can, I prefer that you just push and merge to the same PR without rewriting git history (so no forced pushes, no rebase) as it breaks things such as partial reviews, changes since last review, some comments, information about multiple authors, git signing, etc. and also makes git checkouts more cumbersome due to branches having diverged.

For instance here, I cannot see what has changed since my last test

Alkarex:
With a quick test (using the User CSS extension), that does not seem to work (do nothing)

Elgeryy1:
Fixed in 753711b5f. The previous listener ran before the AJAX-loaded extension form existed, so I changed it to delegated click handling on the slider. Pushed as a new commit, with no rebase or force-push.

Alkarex (approved):

PR #9056, opened 2026-07-16
Refresh sidebar unread badges immediately

### LINKED ISSUE #8249: [Bug] Unread count of feed in sidebar is slow to update

Elgeryy1:
Closes #8249. Centralizes unread badge updates and forces a layout refresh so Safari repaints pseudo-element counts immediately. Tests: node --check, eslint, git diff --check.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -113,7 +113,7 @@ function updateUnreadBadge(elem, value) {
 	// Reading the layout forces the new badge value to be painted immediately.
 	if (elem.matches('.title, .item-title')) {
 		elem.style.visibility = 'visible';
-		void elem.offsetWidth;
+		elem.getBoundingClientRect();
 		elem.style.removeProperty('visibility');
 	}
 }
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -104,16 +104,30 @@ function incLabel(p, inc, spaceAfter) {
 	return i > 0 ? ((spaceAfter ? '' : ' ') + '(' + numberFormat(i) + ')' + (spaceAfter ? ' ' : '')) : '';
 }
 
+function updateUnreadBadge(elem, value) {
+	if (!elem) {
+		return;
+	}
+	elem.setAttribute('data-unread', value);
+	// Safari can keep a pseudo-element stale after only a data attribute change.
+	// Reading the layout forces the new badge value to be painted immediately.
+	if (elem.matches('.title, .item-title')) {
+		elem.style.visibility = 'visible';
+		void elem.offsetWidth;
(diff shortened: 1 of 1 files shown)


JamBalaya56562:
Thanks for working on both approaches. I noticed that #9018 and this PR both address #8249, but use different solutions: #9018 applies the CSS `translateZ(0)` repaint workaround, while this PR centralizes the updates in JavaScript and forces a layout refresh when the count changes.

Since #9018 has already received an approval, could you please clarify how you see the relationship between the two PRs? Is #9056 intended to supersede #9018, complement it, or serve as an alternative implementation?

It would also be helpful to know which approach you recommend merging and what trade-off motivated the second implementation. This would help us understand the intended direction and avoid keeping two overlapping fixes for the same issue.

Frenzie:
> received an approval

For the code only. Someone still has to test if it actually works.

Elgeryy1:
Good question — these are two independent attempts at the same fix rather than complementary. #9018 promotes the badge onto its own compositing layer (\`transform: translateZ(0)\`), a well-known passive fix for this class of WebKit repaint bug. This PR instead forces a synchronous layout read (visibility toggle + \`getBoundingClientRect()\`) at the exact moment each badge's value changes, scoped to the \`.title\`/\`.item-title\` elements that actually carry the counter.

Both should address #8249, but #9018 is the smaller, lower-risk change (pure CSS, no JS control flow touched) and already has a review approval, so I'd recommend merging that one. Closing this PR in favour of #9018.

On testing: I don't have access to real Safari either (Windows only), so I can't personally confirm the repaint behaviour beyond the DOM/CSS reasoning above — flagging that honestly rather than claiming verification I don't have.

PR #9057, opened 2026-07-16
Use physical keys for keyboard shortcuts

### LINKED ISSUE #7954: [Feature] Use "key code" for the hotkeys, instead of the actual letters

Elgeryy1:
Closes #7954. Uses KeyboardEvent.code for letter and digit shortcuts so they survive active keyboard-layout changes. Tests: node --check, eslint, git diff --check.

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1207,7 +1207,15 @@ function init_shortcuts() {
 		}
 
 		const s = context.shortcuts;
-		let k = (ev.key.trim() || ev.code || 'Space').toUpperCase();
+		// Use the physical key for letters and digits so shortcuts keep working
+		// when the active keyboard layout changes.
+		let k = ev.key.trim() || ev.code || 'Space';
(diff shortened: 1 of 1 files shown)


Frenzie:
Duplicate of #9039.

Elgeryy1:
Closing as a duplicate of #9039.

PR #9058, opened 2026-07-16
Detect JSON feeds from .json URLs

### LINKED ISSUE #7574: [Feature] "Type of feed source" should automatically be set to "JSON Feed" based on feed URL

If a JSON feed is picked, we need to manually set type to JSON.

Clients do not seem to have the ability to do this, so they are unable to add JSON feed at all. See #7569

### Use case

- https://blog.hayman.net/feed.json

It seems like it should be easy to set JSON type based on URL file extension, or if we want to go deeper the contents of the feed file.

"Type of feed source" should automatically be set to "JSON Feed" based on feed URL

- avoid clients whilst adding JSON Feed, and manually add the feed through FreshRSS interface

### Additional context

See https://github.com/FreshRSS/FreshRSS/discussions/7569#discussioncomment-13094011

Elgeryy1:
Closes #7574. Detects .json feed URLs while adding a subscription and selects the JSON Feed source type automatically. Tests: php -l, node --check, eslint, git diff --check.

### DIFF
--- a/p/scripts/feed.js
+++ b/p/scripts/feed.js
@@ -120,7 +120,7 @@ function init_json_feed_detection(parent) {
 	const jsonKind = kind.querySelector('option[data-json-feed="1"]');
 	const details = kind.closest('details');
 	const detect = () => {
-		if (kind.value === kind.options[0]?.value && jsonKind && /\.json(?:[?#].*)?$/i.test(url.value.trim())) {
+		if (kind.value === kind.options[0]?.value && jsonKind && /(?:\b|_)json(?:\b|_)/i.test(url.value)) {
 			kind.value = jsonKind.value;
 			kind.dispatchEvent(new Event('change', { bubbles: true }));
 			if (details) {
--- a/p/scripts/feed.js
+++ b/p/scripts/feed.js
@@ -118,10 +118,14 @@ function init_json_feed_detection(parent) {
 	}
 
 	const jsonKind = kind.querySelector('option[data-json-feed="1"]');
+	const details = kind.closest('details');
 	const detect = () => {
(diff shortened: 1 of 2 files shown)


Alkarex (approved):

Alkarex:
I have added some code to automatically open the `<details>` section containing the feed type.
Double-check welcome

p/scripts/feed.js
```diff
@@ -110,6 +110,28 @@ function init_select_show(parent) {
...
+	const url = parent.querySelector('#url_rss');
+	const kind = parent.querySelector('#feed_kind');
+	if (!url || !kind) {
+		return;
+	}
+
+	const jsonKind = kind.querySelector('option[data-json-feed="1"]');
+	const details = kind.closest('details');
+	const detect = () => {
+		if (kind.value === kind.options[0]?.value && jsonKind && /\.json(?:[?#].*)?$/i.test(url.value.trim())) {
```
Frenzie: I think just any json should trigger the opening, not just .json. To clarify, I'm thinking of the not infrequent type of URL scheme that goes something like `/feed/rss(2)`, `/feed/atom` and by extension `/feed/json`.

I supposed that could be gated to something like `[.?=\/\\]json` if desired.

Alkarex: What about:

```suggestion
		if (kind.value === kind.options[0]?.value && jsonKind && /\bjson\b/i.test(url.value)) {
```

Frenzie: I'd add `_` as well in that case, but sure.

Alkarex: > I'd add `_` as well in that case, but sure.

I have not understood your suggestion

Frenzie: I mean that \b doesn't match _, so something like `(\b|_)` perhaps.

Alkarex (approved):

PR #9060, opened 2026-07-16
Refresh only feeds in the current view

### LINKED ISSUE #8025: Restrict "update feeds" to current view?
### Discussed in https://github.com/FreshRSS/FreshRSS/discussions/8023

<sup>Originally posted by **annomatik** September 27, 2025</sup>
Hi! On top of my current FreshRSS view, I have a nifty toolbar:

image

Everything relates to the current view. I can switch between seeing all or just the unread articles in the current view, mark everything read / unread in the current view ( feed / category / whatever ) and switch the viewer for the current view.

But the "update feeds" button on the right side of the toolbar is different. It does not update just the feeds from the current view, but all of them. That's over 100 in my case.

Is there a way of configuring that button to just update the feeds from the current view? That would be very nice :-)

Thanks!
</div>

Elgeryy1:
Closes #8025. Passes the current feed or category selection to the refresh endpoint and limits the batch accordingly. Tests: php -l, node --check, eslint, git diff --check.


Alkarex:
Thanks! I have made two fixes (preserve feed order, and dynamic OPML). Double-check welcome

Alkarex (approved):

PR #9061, opened 2026-07-16
Document OpenBSD httpd API configuration

### LINKED ISSUE #7493: [Feature] OpenBSDs httpd(8) server support

I can't figure out the configuration to allow the google reader api to work on OpenBSDs [httpd(8)](https://man.openbsd.org/httpd) webserver.

### Use case

OpenBSD's httpd(8) works perfectly for FreshRSS but the example config that ships with the port is incomplete.

An example config that works with the Google Reader API.

_No response_

### Additional context

I've had a tiny bit of success with location blocks like this: 
`location "*/check/compatibility" {
                fastcgi socket "/run/php-fpm.sock"
                fastcgi param PATH_INFO "/check/compatibility"
        }`
But they are a little impractical.  There are `location match` rules you can make with patterns(7) and groups but I cannot figure it out.

Elgeryy1:
Closes #7493. Adds an OpenBSD httpd example for FreshRSS with PATH_INFO preserved for the Google Reader API. Tests: markdownlint and git diff --check.

### DIFF
--- a/docs/en/admins/01_Index.md
+++ b/docs/en/admins/01_Index.md
@@ -32,8 +32,8 @@ Learn how to install, update, and backup FreshRSS, as well as how to use the com
 ### Web server configuration
 
 * [Apache/Nginx configuration files](10_ServerConfig.md)
-* [Reverse proxy with Caddy](Caddy.md)
 * [OpenBSD httpd](OpenBSD-httpd.md)
+* [Reverse proxy with Caddy](Caddy.md)
 
 ### Special server information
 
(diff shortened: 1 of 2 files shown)


Alkarex:
I do not have an OpenBSD test environment ready. Did you have a chance to test this configuration?

Furthermore, Web servers not supporting `.htaccess` are missing some other configurations, which we are providing, in particular for caching https://github.com/FreshRSS/FreshRSS/blob/edge/p/.htaccess so it could be nice to either provide a warning or (better) provide a suggestion for OpenBSD httpd compatible configuration. I do not know what default settings OpenBSD httpd is using for HTTP caching.

Alkarex (approved):

PR #9065, opened 2026-07-18
Stop the unread-count poller after the session ends

featurecreep-cron:
## Problem

When a session ends while a reading view is open, `refreshUnreads()` keeps polling `?c=javascript&a=nbUnreadsPerFeed` every 2 minutes and never stops. Each request returns 401 (auth proxy / OIDC image) or 403 (stock). Behind a reverse proxy with fail2ban watching auth failures, the repeated failures get the client IP banned.

## Fix

On a 401/403 from the poll, reload to re-authenticate instead of retrying — the same handling the other `badAjax(this.status == 403)` call sites use.

## Testing

Tested on a live instance behind `mod_auth_openidc`: with the session cookie deleted, the next poll returns one 401 and the page reloads to re-authenticate. Without the change, the same case logs a 401 every two minutes indefinitely.

## Notes

- Pausing the poll while the tab is hidden (Page Visibility) is a separate concern, not included here.


p/scripts/main.js
```diff
@@ -2107,7 +2107,9 @@ function refreshUnreads() {
 	req.onload = function (e) {
 		const json = xmlHttpRequestJson(this);
 		if (!json) {
-			return badAjax(false);
+			// A lost session (401 from an auth proxy, 403 from FreshRSS) reloads to
+			// re-authenticate instead of polling the dead session forever.
+			return badAjax(this.status == 401 || this.status == 403);
```
Frenzie: `>= 400 < 500` or something, no?

Alkarex (approved):

Alkarex:

PR #9070, opened 2026-07-21
Security: public OPML should not contain cURL parameters

Alkarex:
https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-w3xh-7jrp-3qp2

https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-wmcc-46r3-6q5f


Inverle (approved):
I would also suggest displaying a warning in the private OPML export path, when applicable, that the OPML file may contain secrets and to be careful when sharing it, but for later.

PR #9072, opened 2026-07-21
Fix auto-share shortcut on first use

### LINKED ISSUE #7299: [Bug] Sharing with shortcut requires to chose my only sharing mode
### Describe the bug

I have only one sharing mode (Wallabag).
Using the shortcut `s` to share the article no longer directly opens Wallabag but I now have to choose it.

### To Reproduce

1. Go to "One random article".
2. Use shortcut `s` to share it.
3. I now have to choose my only sharing mode.

Image

### Expected behavior

It should (as it was in 1.24.3 and before) directly open Wallabag as explained in `/i/?c=configure&a=shortcut`:

> ℹ️ If there is only one sharing mode, it is used. Otherwise, modes are accessible by their number.

### FreshRSS version

1.25.0

### System information

- Database version: Mysql
- Installation type: Docker

### Additional context

_No response_

Elgeryy1:
## Summary

- Re-read the share actions after creating the lazy-loaded share menu.
- Restore the documented one-key share behavior when exactly one action is configured.

## Why

The shortcut checked the action count before it built the menu. On first use it therefore found no action and required a second press.

Fixes #7299.

## Testing

- `npm run eslint -- p/scripts/main.js`
- `git diff --check edge...HEAD`

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -873,7 +873,7 @@ function auto_share(key) {
 	if (!share) {
 		return;
 	}
-	let shares = share.parentElement.querySelectorAll('.dropdown-menu .item [data-type]');
+	let shares;
 	if (typeof key === 'undefined') {
 		show_share_menu(share);
 		shares = share.parentElement.querySelectorAll('.dropdown-menu .item [data-type]');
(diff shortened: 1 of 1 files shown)


Alkarex (approved):

Alkarex:
I have made a light code change. Test welcome

PR #9006, opened 2026-07-09
Redirect after saving user CSS/JS so changes apply immediately

TowyTowy:
Saving custom CSS/JS rendered the response from the pre-save state (no Post/Redirect/Get), so the cache-busting URL for the stylesheet was stale and the change only appeared after a manual reload, which reads as erratic behaviour.

Redirect back to the extension config after saving (mirroring the `Minz_Request::good()` PRG pattern used elsewhere, e.g. `apiController`/`updateController`), so the next page is built after the save with a fresh cache-busting URL for the updated stylesheet/script. Applied to both User CSS and User JS extensions; version bumped in both metadata.json.

Verified with `php -l` on both changed files.

Fixes #8795

PR #9073, opened 2026-07-21
Prevent cached subscription slider content

### LINKED ISSUE #7723: [Bug] Subscription editing AJAX slider can be miscached
### Describe the bug

The AJAX slider that is loaded with `GET` can be cached, resulting in settings seemingly not being updated or applied, or if multiple submissions happen, being wiped out.

Likely a misconfiguration on my end, but feels like something the software should catch?

### To Reproduce

1. Go to subscription management
2. Click on the feed settings (gear icon)
3. Edit any field
4. Click submit
5. Open the settings again
6. Be confused why it's not updated

### Expected behavior

The new settings are loaded, or submission is not accepted when cached AJAX pane is reused. Something like a basic "this was already submitted" opaque check value would be nice

### FreshRSS version

1.26.2

### System information

- Database version: Unsure, whatever nix package has by default, sorry
- PHP version: 8.4.10
- Installation type: nix package
- Web server type: nginx
- Device: Archlinux desktop
- OS: NixOS 25.05
- Browser: Firefox Developer Edition, 139.0b6

### Additional context

Nginx setup got global `Cache-Control max-age=3600;`

Elgeryy1:
## Summary

- Add a unique query parameter to slider AJAX requests.

## Why

A cached GET response can reopen subscription settings with stale values and make a later form submission appear to overwrite changes. The timestamp makes each slider request distinct for intermediaries that cache dynamic pages.

Fixes #7723.

## Testing

- `npm run eslint -- p/scripts/extra.js`
- `git diff --check edge...HEAD`

### DIFF
--- a/p/scripts/extra.js
+++ b/p/scripts/extra.js
@@ -342,7 +342,7 @@ function open_slider_listener(ev) {
 			const req = new XMLHttpRequest();
 			slider_content.innerHTML = '';
 			slider.classList.add('sliding');
-			const ahref = a.href + '&ajax=1#slider';
+			const ahref = a.href + '&ajax=1&_=' + Date.now() + '#slider';
 			req.open('GET', ahref, true);
 			req.responseType = 'document';
 			req.onload = function (e) {


Inverle:
A better fix is probably to put `FreshRSS_UserDAO::touch()` somewhere where it's missing.
See https://github.com/FreshRSS/FreshRSS/pull/8607 for example

Elgeryy1:
Thanks for the pointer. I verified that edge already contains the underlying fix: #8607 calls FreshRSS_UserDAO::touch() when saving UserCSS/UserJS, and #9006 follows it with PRG so the next slider response is built from the saved state. The cache-busting timestamp in this draft is therefore redundant, so I am closing it.

Inverle:
Yes, but what about the feed settings slider path (like in the linked issue)? Are there any cases where it's not called after saving?

PR #9075, opened 2026-07-21
Redirect away from wrong URL paths to fix displaying of relative URLs

Inverle:
Examples of where this is needed:
* https://demo.freshrss.org/i/index.php/ (no CSS or JS)
* https://demo.freshrss.org/i/index.php/x/y/z?c=configure&a=reading (no CSS or JS)
* https://demo.freshrss.org/api/index.php/ (no PASS because JS doesn't load)
* etc.

Because of `index.php` being treated as a directory in those URLs, CSS and JS aren't loading properly due to them being displayed from a relative path via `Minz_Url::display()`:
```html
<link rel="stylesheet" href="../themes/Origine/origine.css?1779721222" />
```
https://github.com/FreshRSS/FreshRSS/blob/8c5bd58808fec14853e5f1b2582b14db3cdb563a/app/FreshRSS.php#L136

Also applies to all the other links on the page of course.
The URLs are displayed from the `PUBLIC_RELATIVE` constant via `Minz_Url::display()`, which is always expected to be `..`:
https://github.com/FreshRSS/FreshRSS/blob/8c5bd58808fec14853e5f1b2582b14db3cdb563a/constants.php#L25

The implemented fix redirects away from the bad URL paths, when `PATH_INFO` is not empty, to a manually built URL from `$_GET` parameters. For API pages, an error is shown instead.

### DIFF
--- a/p/api/query.php
+++ b/p/api/query.php
@@ -38,6 +38,18 @@ if (!FreshRSS_Context::hasSystemConf() || !FreshRSS_Context::systemConf()->api_e
 	die('Service Unavailable!');
 }
 
+// Strip path info from URL to ensure proper displaying of relative URLs
+$pathInfo = $_SERVER['PATH_INFO'] ?? $_SERVER['ORIG_PATH_INFO'] ?? '';
+if ($pathInfo !== '') {
+	$scriptName = is_string($_SERVER['SCRIPT_NAME'] ?? null) ? $_SERVER['SCRIPT_NAME'] : '';
+	$queryString = is_string($_SERVER['QUERY_STRING'] ?? null) ? $_SERVER['QUERY_STRING'] : '';
+	if ($queryString !== '') {
+		$queryString = '?' . $queryString;
+	}
+	header('Location: ' . $scriptName . $queryString);
+	exit();
+}
+
 FreshRSS_Context::initUser($user);
 if (!FreshRSS_Context::hasUserConf() || !FreshRSS_Context::userConf()->enabled) {
 	usleep(rand(100, 10000));	//Primitive mitigation of scanning for users
--- a/p/api/query.php
+++ b/p/api/query.php
@@ -38,16 +38,10 @@ if (!FreshRSS_Context::hasSystemConf() || !FreshRSS_Context::systemConf()->api_e
 	die('Service Unavailable!');
 }
 
-// Strip path info from URL to ensure proper displaying of relative URLs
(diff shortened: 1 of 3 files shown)


p/api/query.php
```diff
@@ -38,6 +38,18 @@
...
+$pathInfo = $_SERVER['PATH_INFO'] ?? $_SERVER['ORIG_PATH_INFO'] ?? '';
+if ($pathInfo !== '') {
+	$scriptName = is_string($_SERVER['SCRIPT_NAME'] ?? null) ? $_SERVER['SCRIPT_NAME'] : '';
+	$queryString = is_string($_SERVER['QUERY_STRING'] ?? null) ? $_SERVER['QUERY_STRING'] : '';
+	if ($queryString !== '') {
+		$queryString = '?' . $queryString;
+	}
+	header('Location: ' . $scriptName . $queryString);
+	exit();
+}
```
Inverle: It's done differently for API pages, since Minz isn't as well integrated there.

Alkarex:
Those URLs should probably not be allowed, as this risks opening wrong paths in multiple locations.

Instead, what about a redirection, done either:

in PHP, e.g.:

```php
if (!empty($_SERVER['PATH_INFO'])) {
	// Avoid trailing slash in URL
	Minz_Request::forward(['c' => 'index', 'a' => 'index'], redirect: true);
	exit();
}
```

or using Apache mod_alias (which is commonplace, but currently disabled in our Dockerfile):

```
RedirectMatch 301 ^/i/index\.php/(.*)$ /i/$1
```

Alkarex:
Ah sorry, I am only realising now that you are doing a redirection already. Based on the amount of code, I thought it was some kind of path fix

Alkarex:
What about either returning an HTTP error or redirecting to the root, to simplify the code and more clearly disallow such URLs?

Inverle:
In the FrontController it's just two lines of code to fix the issue, so I think it's fine there and as simple as it can be.

Maybe we should return an error only in the API pages to make it simpler there.

Alkarex:
> In the FrontController it's just two lines of code to fix the issue, so I think it's fine there and as simple as it can be.
> 
> Maybe we should return an error only in the API pages to make it simpler there.

Ok, fair enough. Fine with https://github.com/FreshRSS/FreshRSS/pull/9075/commits/83ec502c7eb535ee53ad73235def345107f3c181 ?

Inverle:
Yes

Alkarex (approved):

PR #9077, opened 2026-07-22
Log failed CSS content retrievals

### LINKED ISSUE #8875: [Feature] log errors fpr content retrievals based on CSS

user can use content retrievals based on CSS to get full post from links in RSS that point to actual website. 

I love this feature.

but sometimes the websites change their layout and this retrieval fails silently.

### Use case

_No response_

can this be logged and shown in feed (like error for feed retrieval that marks the feed red).

also in log of the freshrss in webui logs section?

_No response_

### Additional context

_No response_

Elgeryy1:
## Summary

- Log a warning when a CSS content selector matches no elements.
- Log a distinct warning when matched content becomes empty after sanitization.
- Avoid duplicate warnings for the no-match case.

## Why

CSS-based content retrieval otherwise fails silently when a publisher changes its markup.

Fixes #8875.

## Validation

- git diff --check origin/edge...HEAD`n- Local PHP/Composer tooling is unavailable on this Windows host; GitHub Actions will run the project checks.


Frenzie:
Would it be correct to assume it never reaches this part of the code if there's some kind of error (e.g., failure to connect or an HTTP error like 404)? Just double checking since otherwise it could result in a lot of extra noise.

Elgeryy1:
Correct. The CSS extraction block is entered only when the HTTP response body is non-empty, so connection failures and HTTP errors do not reach either warning. The warnings cover a successful HTML retrieval where the selector matches no elements, or where matched content becomes empty after sanitization.

Frenzie (approved):

Elgeryy1:
Yes, that's correct. A few lines above, there's:

\`\`\`php
\$response = FreshRSS_http_Util::httpGet(\$url, \$cachePath, 'html', \$feed->attributes(), \$feed->curlOptions());
\$html = \$response['body'];
if (\$html !== '') {
	...
}
\`\`\`

The whole block containing the new warnings is guarded by \`\$html !== ''\`, and \`httpGet()\` returns an empty body on connection failures / HTTP errors. So these warnings can only fire once we already have a non-empty HTML response — i.e. when the configured CSS selector matched nothing, or matched elements that produced no content after sanitizing. Network/HTTP failures never reach this code, so there shouldn't be extra noise from those.

PR #9080, opened 2026-07-22
Add feed website shortcut

### LINKED ISSUE #8841: [Feature] Add a shortcut to the site's homepage somewhere in its feed

I'd like to have a link to the site's homepage *somewhere* on a site's feed.

### Use case

A feed is *kind of* a bookmark, and it is also useful to access the homepage of a site directly.

Add a copy of this buton
Image

Here next to the site's title
Image

A regular bookmark, but it's in my browser or in another tool, so it's not in the same place nor updated when I edit my feeds.

I also found a "visit website" link in the gear dropdown menu, but it's buried just one click too deep and feels like an outsider between all the other feed management options. Regarding the UX, I don't think it makes sense to go into that option menu to find it. I believe it might be better to move it as an icon either next to the gear or the unread count, but that might create more clutter, unless there is a possibility to widen the sidebar.

Image

### Additional context

_No response_

Elgeryy1:
## What changed
- Adds a direct external-link icon beside a feed name in normal reading mode.
- Covers both the initial page and dynamically rendered entries.

## Why
The feed configuration already exposes the site homepage, but reading a feed required opening the management menu first.

## Validation
- `git diff --check`
- GitHub Actions will run the PHP checks.

Closes #8841

### DIFF
--- a/app/views/helpers/index/normal/entry_header.phtml
+++ b/app/views/helpers/index/normal/entry_header.phtml
@@ -54,6 +54,9 @@
 		<a href="<?= _url('index', 'index', 'get', 'f_' . $this->feed->id()) ?>" class="item-element" title="<?= _t('gen.action.filter') ?>: <?= $this->feed->name() ?>">
 			<?php if (FreshRSS_Context::userConf()->show_favicons && 'name' !== $topline_website): ?><img class="favicon" src="<?= $this->feed->favicon() ?>" alt="✇" loading="lazy" /><?php endif; ?><?php if ('icon' !== $topline_website): ?><span class="websiteName"><?= $this->feed->name() ?></span><?php endif; ?>
 		</a>
+		<?php if ($this->feed->website() !== ''): ?>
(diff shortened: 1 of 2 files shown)


JamBalaya56562:
Thanks for working on this. If possible, could you please add before-and-after screenshots of an article header in the normal reading view, showing the new shortcut to the feed’s website?

If practical, showing the shortcut with the feed name in its different supported positions would also be useful. This would make the placement and visual impact of the new action easier to understand.

Elgeryy1:
Before / after, cropped to the feed-name area (the header also has a pre-existing "See on original website" icon further right, unrelated to this change — same glyph, easy to confuse with the new one, so cropping tighter here for clarity):

9080-before-feedname-crop

9080-after-feedname-crop

And with **topline_website = icon** (name hidden, favicon + shortcut only) to show another supported position:

9080-article-header-icon-mode

Alkarex:
That is too much of a layout change for me, in particular when using a single line:
image

I would like to refrain adding too many options. Maybe this use-case could be addressed by improving the subscription management instead

PR #9081, opened 2026-07-22
Avoid missing unread cache index failures

### LINKED ISSUE #8492: [Bug] "There are no articles to show." when displaying only Unread or Read
### Describe the bug

Since 1.28.1, neither `state=0`, `state=1` or `state=2` are working: all three of them always displays "**There are no articles to show.**"

Only `state=3` is actually displaying both Unread and Read articles.

Before 1.28, it was working perfectly. (1.28.0 was not showing anything at all, which is why 1.28.1 has been release, I believe)

### To Reproduce

1. Go to your FreshRSS home page, like `https://example.com/freshrss/`
2. Click on `Show read` and/or `Show unread` to enable one or both of them
3. See error **There are no articles to show.** in three cases.

### Expected behavior

_No response_

### FreshRSS version

1.28.1

### System information

- Database version: SQLite
- PHP version: 8.4.16
- Installation type: Sources (since 2015, yay!)
- Web server type: Nginx
- OS: Debian 13
- Browser: Firefox 146.0.1

### Additional context

_No response_

Elgeryy1:
## What changed
- Stops forcing `entry_feed_read_index` when recalculating unread-count caches.

## Why
Older upgraded MySQL/MariaDB databases can lack that historical index. The forced index hint then turns a cache refresh into a SQL error and makes read/unread views appear empty. Letting the database optimizer select an available index keeps those installations working while current schemas still retain the composite index.

## Validation
- `git diff --check`
- GitHub Actions will run the PHP test and static-analysis suite.

Closes #8492

### DIFF
--- a/app/Models/EntryDAO.php
+++ b/app/Models/EntryDAO.php
@@ -456,13 +456,10 @@ class FreshRSS_EntryDAO extends Minz_ModelPdo {
 	 * feeds from one category or on all feeds.
 	 */
 	protected function updateCacheUnreads(?int $catId = null, ?int $feedId = null): bool {
-		// Help MySQL/MariaDB's optimizer with the query plan:
-		$useIndex = $this->pdo->dbType() === 'mysql' ? 'USE INDEX (entry_feed_read_index)' : '';
-
 		$sql = <<<SQL
 			UPDATE `_feed`
 			SET `cache_nbUnreads`=(
-				SELECT COUNT(*) AS nbUnreads FROM `_entry` e {$useIndex}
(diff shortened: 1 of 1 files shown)


Alkarex:
I think on error, we should try to add this index (by calling `autoUpdateDb()` and updating that function accordingly) and retry the query. That index is needed for performance also elsewhere

Alkarex (changes requested):
I think on error, we should try to add this index (by calling autoUpdateDb() and updating that function accordingly) and retry the query. That index is needed for performance also elsewhere

PR #9082, opened 2026-07-22
Restore remembered categories before sidebar paint

### LINKED ISSUE #8512: [Bug] "Remember open categories" animation glitch
### Describe the bug

When tapping on a feed with "Remember open categories" option enabled, only the category containing currently selected feed is initially displayed in unfolded state. All other categories and initially folded regardless of their previous state, and their folded/unfolded state is reapplied a moment later. This results in a very visible vertical jump of all feeds and categories on a larger screen.

Please see the attached video. The issue is more noticeable if you view it frame by frame.

https://github.com/user-attachments/assets/56c85de6-4774-4566-b77c-f751c2046535

### To Reproduce

1. Go to 'Reading' setting page
2. Set 'Categories to unfold' option to 'Remember open categories'
3. Save and return to the main page
4. Unfold several categories with unread feeds in them manually
5. Click on one of the feeds in an unfolded category
6. Notice the folding/unfolding reload animation

### Expected behavior

It would be great if all categories have retained their folded/unfolded state right from the start.

### FreshRSS version

1.28.1

### System information

- Database version: not sure
- PHP version: not sure
- Installation type: [LinuxServer FreshRSS docker image](https://hub.docker.com/r/linuxserver/freshrss/)
- Web server type: not sure
- Device: Desktop PC
- OS: Bazzite Linux
- Browser: Firefox 147

### Additional context

_No response_

Elgeryy1:
## What changed
- Restores remembered category expansion states synchronously after rendering the sidebar.
- Applies the saved state before the sidebar is flushed to the browser.

## Why
The existing restoration ran later from the main JavaScript initialization. On navigation it briefly rendered only the active category open, then expanded saved categories, causing a visible layout jump.

## Validation
- `git diff --check`
- GitHub Actions will run PHTML and JavaScript checks.

Closes #8512


JamBalaya56562:
Thanks for working on this. To make the behavior change easier to understand and review, would you be able to add a short before/after screen recording or GIF?

Ideally, it would show navigation or a page reload with several categories remembered as open, so the brief sidebar layout jump before the fix and the synchronous restoration after the fix can be compared.

This would be especially helpful here because the difference is mainly in the timing and may be difficult to convey with static screenshots. Thank you!

Elgeryy1:
Went to record the before/after GIF requested above and the "after" build still showed the same jump. Dug into why:

**The fix's inline `<script>` never executes — it's blocked by CSP.** Every normal-view page response sends `Content-Security-Policy: default-src 'self'` (see `indexController.php`), which blocks inline scripts without a nonce or hash. Chrome's console confirms it outright:

```
Executing inline script violates the following Content Security Policy directive 'default-src 'self''... The action has been blocked.
```

FreshRSS has no CSP-nonce plumbing anywhere in the codebase to attach to a script tag (the only `nonce` references are the unrelated crypto-login challenge), so this inline block is dead on arrival in any default installation.

To confirm it's not just imperceptible, I measured the actual DOM state (whether `.tree-folder-items` has `.active`) at each animation frame after a reload, with several categories remembered as open:

- Without this PR: closed at ~45ms → open at ~66ms
- With this PR: closed at ~45ms → open at ~66ms (identical)

Since the new script never runs, the categories are still being restored by the existing `init_column_categories()` in `main.js` (an external file, so CSP doesn't touch it) — at the same point in the load sequence as before. This PR currently adds inert code and doesn't fix #8512.

To actually fix it, the restoration needs to happen in something CSP allows to run early — e.g. a small external script loaded/executed before the rest of `main.js`, not an inline block in the phtml. I'll rework it that way rather than post a GIF that wouldn't show any real difference.

Elgeryy1:
Reworked as promised. Moved the restoration into `p/scripts/restore-categories.js`, loaded via `<script src="...">` right after the sidebar markup — same position as the inline attempt, but as an external same-origin file so `default-src 'self'` allows it to run. Removed the now-dead duplicate restoration in `main.js`'s `init_column_categories()` (and the `openCategory()` helper it alone used), since this earlier script always runs first.

Re-measured the same way as before (DOM state per animation frame after a reload, several categories remembered as open):

- Before this update: closed at ~45ms → open at ~66ms (script blocked, old late path still doing the work)
- After: already open at first paint, ~56ms — no closed state observed

Also checked the browser console: no CSP violation now, where before it explicitly logged the inline script being blocked.

`composer test` (phpunit/phpcs/phpstan) and `eslint` all pass; the only phpunit failures are the pre-existing Windows CRLF-vs-LF i18n test noise, unrelated to this change.

PR #9090, opened 2026-07-22
Complete Russian reading translations

Elgeryy1:
## Summary
- translate enclosure and unread-count sidebar settings to Russian
- use current custom-sort terminology
- provide Russian plural forms for new and unread article notifications

## Validation
- `git diff --check`


Frenzie:
If these PRs are AI translations I'd rather just skip on them since then you have to make an effort to find the original instead of just finding it right there. (And at the very least they're definitely `// DIRTY`.)

Elgeryy1:
Thanks, agreed. I am closing the translation PRs from this batch rather than treating AI output as verified translations. They should remain `// TODO` / `// DIRTY` for human contributors.

PR #9114, opened 2026-07-22
Translate Finnish username format

Elgeryy1:
## Summary
- translate the username character-format hint to Finnish

## Validation
- `git diff --check`


Alkarex:
I am not sure yet of what policy to apply for AI-translated i18n strings. Maybe marking them as  `// DIRTY`.
Also, please make a single PR for all languages as I doubt we will have the bandwidth to review and merge them independently.

Alkarex:
PS: either a single PR for all languages focussed on a part of the i18n such as "username format" or a PR per language for all the strings you would like to AI-translate

Frenzie:
> I am not sure yet of what policy to apply for AI-translated i18n strings. Maybe marking them as // DIRTY.

Cross-ref to <https://github.com/FreshRSS/FreshRSS/pull/9090#issuecomment-5046629227>.

> If these PRs are AI translations I'd rather just skip on them since then you have to make an effort to find the original instead of just finding it right there. (And at the very least they're definitely `// DIRTY`.)

My preference is not at all, as stated there. I think it gets more in the way than that it is helpful. Dirty is currently intended like "this is the old string based on a previous translation and maybe it only needs minor adjustment". Though we could consider making it something like `// DIRTY: [this is the original string]`.

PR #9127, opened 2026-07-22
Remove stale dynamic favicon links

### LINKED ISSUE #8237: [Bug] Count on Favicon does not match/stay synced with the count shown in the title
### Describe the bug

When viewing the main page of RSS articles for the feeds I'm subscribed to, I notice that over time that the count of unread articles in the favicon becomes out of sync with the count that is next to the title. Typically the number in the favicon is much higher than the count in the title.

Image

Interestingly, while looking at the underlying HTML, I see that there are two favicon entries showing, in some cases though the image that they rendered was not even the count, so I'm not sure where my browser was getting the count from.

Image

### To Reproduce

1. Go to the main article list page
2. Wait for new articles to show up
3. See the count change in the favicon and title
4. View article, see that the count changes in the title, but the favicon stays the same
5. Repeat 2 through 4

### Expected behavior

The correct count is shown in the favicon. Although, it'd also be nice if we could disable the favicon count since it is duplicative of the count that is already 

### FreshRSS version

1.27.1

### System information

- Database version: MariaDB - 11
- PHP version: 8.2
- Installation type: git source
- Web server type: Apache
- Device: MacBook Pro 
- OS: MacOS 15.7.1
- Browser: Safari 26.0.1

### Additional context

_No response_

Elgeryy1:
## Summary
- remove every previous dynamic favicon link before appending the updated unread-count icon
- prevent stale duplicate favicon links from being preferred by browsers

## Validation
- `git diff --check`
- `node --check p/scripts/main.js`

Fixes #8237.


Alkarex:
Is it on purpose that this PR is marked as draft?

p/scripts/main.js
```diff
@@ -2347,7 +2347,7 @@ function faviconNbUnread(n) {
 			temp.remove();
 		}
 		link.href = `data:image/svg+xml;base64,${btoa(svgOutput || svgBase)}`;
-		document.querySelector('#favicon').remove();
+		document.querySelectorAll('#favicon').forEach(favicon => favicon.remove());
```
Alkarex: Multiple `id="favicon"` looks like a bug (an HTML document is not supposed to have multiple identical IDs).
Probably the code handling the favicons should be fixed instead

p/scripts/main.js
```diff
@@ -2347,7 +2347,7 @@ function faviconNbUnread(n) {
...
 		document.head.appendChild(link);
```
Frenzie: Is the current code written this way on purpose to work around something or other? It seems more expedient to simply use `replaceWith()`.

Elgeryy1:
It was initially a draft while CI was pending. The checks are now green, and the PR has been marked ready for review.

p/scripts/main.js
Elgeryy1: Agreed. I have changed this to keep one DOM favicon link and replace it in place (`favicon.replaceWith(link)`) rather than append a second link. This keeps the document valid and avoids relying on cleanup of duplicate IDs.

p/scripts/main.js
Elgeryy1: I do not see a browser-specific reason for the append/remove sequence. I have replaced it with `favicon.replaceWith(link)`, which updates the existing link atomically without ever inserting a duplicate `id=favicon`.

Frenzie (approved):
lgtm

p/scripts/main.js
```diff
@@ -2351,8 +2352,7 @@ function faviconNbUnread(n) {
 			temp.remove();
 		}
 		link.href = `data:image/svg+xml;base64,${btoa(svgOutput || svgBase)}`;
-		document.querySelector('#favicon').remove();
-		document.head.appendChild(link);
+		favicon.replaceWith(link);
```
Alkarex, line 2355: Shouldn't this line be within the scope of the `if (link) {}` ?

Frenzie: But it is? Or do you mean to move it down a line or two, up one level. I don't think such separation from link.href would improve clarity but it'd otherwise behave the same.

Frenzie: Hm no, it's at exactly that level?

Alkarex: That line is just after the end of the `if (link) {}` section, meaning `link` could be null

Frenzie: Did you look at it on mobile? It does weird things with indentation sometimes. But we're inside if link here. ;-)

Alkarex: Indeed, not on mobile, but I did not look correctly

PR #9128, opened 2026-07-22
Limit Nord title hover transition to work around Chromium bug

### LINKED ISSUE #8075: [Bug] RTL Padding in Chromium-based browsers 141.0.7390.55
### Describe the bug

Chromium 141.0.7390.55 has supposedly fixed some RTL issue, but in addition caused some weird rendering in FreshRSS.

https://chromium.googlesource.com/chromium/src/+/95049732f5eb21ac4e489b46075c1fc47f03c067
https://chromium.googlesource.com/chromium/src/+log/140.0.7339.213..141.0.7390.55?pretty=fuller&n=1000

Now all my chromium-based browsers that upgraded to version 141.0.7390.55 have an issue in FreshRSS where hovering over article titles written in Arabic shifts the location of the title further to the right, and goes back to its original location upon hovering away from the title.

https://github.com/user-attachments/assets/2425837a-7256-4e99-bc81-752b52312e6a

I re-downloaded Chromium version 140.0.7339.213 to confirm, and it indeed did not have this issue. Firefox-based browsers also do not have this issue. The issue only affects titles written in RTL languages like Arabic in the latest chromium-based browsers.

### To Reproduce

1. Go to FreshRSS homepage
2. Hover over titles written in an RTL language like Arabic

### Expected behavior

_No response_

### FreshRSS version

1.27.1

### System information

- Installation type: Docker on Ubuntu 22.04
- OS: Windows 11
- Browser: Chromium/Chrome/Brave 141.0.7390.55

### Additional context

_No response_

Elgeryy1:
## Summary

- limit the Nord theme title hover transition to the background color
- avoid transitioning layout-related properties for RTL titles in Chromium

## Validation

- `npx stylelint p/themes/Nord/nord.css p/themes/Nord/nord.rtl.css`
- `git diff --check`

Fixes #8075.

### DIFF
--- a/p/themes/Nord/nord.css
+++ b/p/themes/Nord/nord.css
@@ -815,7 +815,7 @@ li.item.active {
 
 .flux:not(.current):hover .item .title {
 	background: var(--accent-bg);
-	transition: .3s;
+	transition: background-color .3s;
 }
 
 .flux .flux_header .item .title {
--- a/p/themes/Nord/nord.rtl.css
+++ b/p/themes/Nord/nord.rtl.css
@@ -815,7 +815,7 @@ li.item.active {
 
 .flux:not(.current):hover .item .title {
 	background: var(--accent-bg);
-	transition: .3s;
+	transition: background-color .3s;
 }
 
 .flux .flux_header .item .title {


Alkarex:
If possible, to ease the review, would you be able to provide a before/after screenshot for PRs affecting the style/layout/themes ?

Elgeryy1:
Sorry for the slow follow-up — thought about the best way to show this, but a static before/after screenshot doesn't really capture it well: the bug in #8075 is a *hover-triggered* layout jump (RTL/`dir="auto"` titles shifting position while the mouse is over them in Chromium 141+), not a static visual difference. The two states look identical at rest and mid-hover; the only difference is whether Chromium's style recalc glitch gets triggered by the implicit `transition: all`.

Per Frenzie's and erannave's diagnosis in #8075, animating `transition: .3s` (i.e. `all`) on `.flux:not(.current):hover .item .title` makes Chromium re-run its (buggy) transition-affected recalculation on every watched property, including whatever changed in Chromium 141 around RTL/`dir="auto"` handling — hence the jump. Scoping the transition to `background-color` only avoids putting those RTL-affected properties on Chromium's transition watch list, without changing anything visually for LTR titles.

Happiest to try a screen recording of the actual jump with an RTL feed if that would help more than the explanation above — let me know.

Inverle:
You can make a GIF or a screen recording.

Elgeryy1:
I don't have a reliable way to reproduce the specific Chromium 141.0.7390.55 RTL repaint glitch in my current environment (would need that exact browser build plus a live RTL feed to capture something faithful, rather than a recording that might not actually show the bug). The code-level reasoning above should cover the "why" for review purposes — happy to have @techie6 or anyone who could reproduce it on 141+ attach a recording if that would help close this out.

Inverle:
You can use `npx @puppeteer/browsers` to download the exact browser build, but that is probably not necessary since I would assume the bug still exists on latest Chromium.
Here's a RTL feed you can test with: https://www.almasryalyoum.com/rss/rssfeed

Elgeryy1:
Tried this properly: spun up a local instance with Nord theme + the BBC Arabic feed and hovered over RTL titles. My automated hover didn't reliably trigger the CSS `:hover` state in a way I could capture the transient jump — screenshots only show steady before/after states, and this bug is specifically about what happens *during* the transition, which a static capture can't show even when it fires correctly. So I don't have a recording to add. If either of you can reproduce it live on 141+, a quick screen recording from your end would settle it faster than more attempts from mine.

Frenzie:
Unfortunately I don't believe this PR does anything, at least not relating to #9128. It's just that in Nord the issue in question is animated with the transition, but the Chromium issue is that anything changes when hovering at all.

Frenzie:
The issue comes from a padding-right on `.flux:not(.current):hover .flux_header.has-date .item .title` that interacts badly with the RTL, but it's not immediately obvious why that padding-right is even there at all. Even in LTR the effect of it is odd.

Elgeryy1:
Sorry for how formal/repetitive those last few replies read — got carried away trying to be thorough. Bottom line from my end: I can't get a faithful repro of the transient jump with what I have here, so if either of you can grab a quick recording on 141+ that'd close this out faster than me poking at it more. Appreciate the patience!

Frenzie:
I think I'll just check if I can fix the issue in the next ten minutes or so.

Frenzie:
It's simple: #8997 should be reverted. Now to see what that's about…

Elgeryy1:
Working on getting a screen recording of the hover behavior now, will attach it here shortly.

Frenzie:
I'll submit a PR soon. The correct way to go about it is almost certainly to simplify the code, meaning to completely remove `.flux:not(.current):hover .flux_header.has-date .item .title` and to switch to `column-gap` (because it's grid!). The AI here is adding kludges upon kludges.

Elgeryy1:
https://github.com/user-attachments/assets/9d018c7f-79dc-43c2-8a05-7bd35983a577

Here's the hover jump on an RTL title

techie6:
I confirm that the RTL hover issue is still present on latest Chrome version 150.0.7871.182 using the default Origine theme on latest FreshRSS 1.29.1

Alkarex:
Please try again with the `edge` branch now that https://github.com/FreshRSS/FreshRSS/pull/9140 is merged

Elgeryy1:
Closing this as superseded by #9140, which fixed the underlying hover/RTL behavior in the shared base theme after review. This PR's Nord-only transition adjustment no longer addresses the root cause.

PR #9130, opened 2026-07-22
Persist sort selected from navigation menu

### LINKED ISSUE #8227: [Bug] Sort Order Doesn't Persist
### Describe the bug

Setting a sort order (like Publication order, as most readers do) does not persist. 

### To Reproduce

1. Sort Feed
3. Close browser
4. Go to feed

### Expected behavior

Feed is sorted as indicated

### FreshRSS version

 1.27.1 

### System information

Latest Docker, SQLite, any device

### Additional context

Maybe sorting was only done with ephemeral url params?

Elgeryy1:
## Summary

- persist the sort and order selected in the navigation menu for authenticated users
- keep ordinary sort URLs as non-persistent navigation overrides
- use the existing CSRF-protected form for the persistence request

## Validation

- `node --check p/scripts/main.js`
- `git diff --check`

PHP and Composer are not available in this Windows environment; repository CI will run PHP checks.

Fixes #8227.

### DIFF
--- a/app/Controllers/indexController.php
+++ b/app/Controllers/indexController.php
@@ -8,6 +8,22 @@ class FreshRSS_index_Controller extends FreshRSS_ActionController {
 
 	#[\Override]
 	public function firstAction(): void {
+		if (Minz_Request::isPost() && Minz_Request::paramBoolean('saveSort') && FreshRSS_Auth::hasAccess()) {
+			$sort = Minz_Request::paramString('sort', plaintext: true);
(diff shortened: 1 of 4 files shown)


Alkarex:
I am a bit worried of this one, due to risks of confusions.
Since https://github.com/FreshRSS/FreshRSS/issues/8227 was posted, there is now a server-side option to save the sorting preference. Furthermore, sorting is kept during the navigation.
So this PR is probably not needed

PR #9131, opened 2026-07-22
Fix lazy-loaded images in retrieved full content

### LINKED ISSUE #8317: Problem with missing lazy-loaded images while web scraping
Hi.

I'm facing a problem after using HTML + XPath (web scraping) to generate the RSS feed from a page without it.
The "item content" element contains lazy-loaded images which won't show in the feed. Instead, a gray rectangle is drawed.

This is an example of the code retrieved:

```kodi omega v21```

If only there was a way to replace `src=""` by the contents of `data-src=""`, that would be fixed.

If I had the skill I would try to code an extension, but sadly I'm not that savvy.
I might try though, if anyone could point me in the right direction.

Of course, any other ideas to achieve the same result are welcome.

Thank you!!

### PR DESCRIPTION
**Pull request by Elgeryy1**
Fixes #8317.

When full-content retrieval receives a lazy-loading placeholder in an image `src`, restore the usable URL from `data-src`, `data-src-template`, or `data-original` before HTML sanitization. This covers both data-URI placeholders and the reported `1x1_spacer` placeholder, while leaving regular images unchanged and accepting only HTTP(S) or relative replacements.

Validation:
- `git diff --check`
- PHP is not installed in this Windows environment; GitHub Actions will run the PHP test suite and static checks.

### CODE REVIEW

**Comment by JamBalaya56562** [comment]
Nice catch on the `data-src` fallback — that gap is real.

One part looks worth pinning down with a test, the scheme check:

```php
$isWebUrl = preg_match('/^[a-z][a-z0-9+.-]*:/i', $replacement) !== 1
	|| preg_match('/^https?:/i', $replacement) === 1;
```

I ran it against a few inputs:

| Input | Result |
|---|---|
| `javascript:alert(1)` | blocked |
| `JaVaScRiPt:alert(1)` | blocked |
| `data:image/png;base64,…` | blocked |
| `java` + TAB + `script:alert(1)` | **accepted** |
| `java` + LF + `script:alert(1)` | **accepted** |

The [URL parser removes tabs and newlines](https://url.spec.whatwg.org/#concept-basic-url-parser) before resolving, so the last two normalise back to `javascript:`. I am not claiming this is exploitable — `javascript:` in `img@src` does not execute in current browsers, and the parsed content goes through `sanitizeHTML()` afterwards — but it is exactly the kind of edge case that is cheap to pin down and easy to regress later.

The snag is that the new loop sits inside `getContentByParsing()`, which fetches over HTTP and needs a `FreshRSS_Feed`, so a unit test cannot reach it as it stands. Would you consider pulling the image-fixing part out into a small helper (taking the `DOMXPath` and the node, or simply HTML in / HTML out)? That would make the placeholder detection, the attribute precedence and the scheme check all testable. #9050 already adds `tests/app/Models/EntryTest.php`, which would be a natural home for it.

Elgeryy1:
Good catch on the tab/newline normalisation — fixed by stripping ASCII tabs/newlines/carriage-returns before the scheme check, so `java\tscript:` and `java\nscript:` are evaluated in their post-normalisation form and correctly rejected. Added test cases for both.

Also pulled the image-fixing loop out into `FreshRSS_Entry::replaceLazyLoadedImagePlaceholders(DOMXPath $xpath, DOMElement $node)` as suggested, so it's testable directly against constructed HTML rather than needing an HTTP fetch. `tests/app/Models/EntryTest.php` now covers: empty-src/data-URI/1x1-spacer placeholder detection, real-src passthrough, the data-src → data-src-template → data-original fallback order, the two bypass cases, and protocol-relative URLs.

JamBalaya56562:
@Elgeryy1 Could you resolve merge conflict?

Elgeryy1:
Done, merged `edge` in — conflict's resolved.

Elgeryy1:
The requested extraction and coverage are already included in commit 9f194cd7: 
eplaceLazyLoadedImagePlaceholders() isolates the DOM logic, and the tests cover placeholder detection, attribute precedence, safe URLs, and tab/newline-normalized javascript: schemes. The branch was also merged with current edge in cd241028b.

PR #9132, opened 2026-07-22
Filter global view feed list by state and search

### LINKED ISSUE #4564: [BUG] Global view: Favorite button has no effect to the listed feeds
**Describe the bug**
If the favorite button is active, then the feed lists are not filtered
grafik

When you open the feed, then the favorite articles are shown. 
If there are no favorite articles then an empty list is shown.

**To Reproduce**
1. Go to global view
2. active the favorite button
3. see the feed lists
4. open the feeds

**Expected behavior**
List only feeds with min. 1 favorite article

- FreshRSS version: 1.19.2, 1.20.0-dev

Elgeryy1:
Fixes #4564.

The global view now filters its category/feed list when the Favorites filter is active, using the same article state and search criteria used when a feed is opened. Empty categories are omitted rather than showing feeds that would open to no matching articles.

Validation:
- `git diff --check`
- PHP is not installed in this Windows environment; GitHub Actions will run the PHP tests and static analysis.


Alkarex:
After reflection, I have revised the logic so that in the global view, the categories and feeds shown are always filtered by the search query (if any) and the states (if consequential):

image


This makes the global view more useful and distinct in my opinion (this was otherwise not a view I was using myself, but now it will be).

Note that this behaviour would be challenging to port to the normal view for performance reasons.

Alkarex (approved):

Alkarex:
Follow-up in:
* https://github.com/FreshRSS/FreshRSS/pull/9144
* https://github.com/FreshRSS/FreshRSS/pull/9145

Elgeryy1:
Nice, glad it's useful as a base for those. Nothing needed from this PR for them to work as far as I can tell.

PR #9133, opened 2026-07-22
Preserve commas in author names

### LINKED ISSUE #7909: [Bug] Multiple "dc:creator" in feed ignored
### Describe the bug

[According to the RSS documentation](https://www.rssboard.org/rss-profile#namespace-elements-dublin-creator), a feed item can have multiple authors by adding more than one dc:creator element.

Currently, only one author is displayed for such items.

### To Reproduce

1. Subscribe to a feed with items that have multiple authors. (One current example is in [this feed](https://artikel91.eu/feed/), the item with the guid `https://artikel91.eu/feed/`
2. Look at the item in FreshRSS

**RSS source:** Screenshot of an RSS feed with multiple dc:creator elements

**Display in FreshRSS**: Screenshot of the author display of the same article in FreshRSS

### Expected behavior

If multiple authors are provided, multiple authors should be displayed. (Also, there's a space missing before the author – is this a general problem or an issue of the German localization?)

### FreshRSS version

1.27.1-dev 

### System information

- Database version: MariaDB 10.6
- PHP version: 8.3
- Installation type: git
- Web server type: nginx
- Device/OS/Browser: problem identical across browser

### Additional context

_No response_

Elgeryy1:
Fixes #7909.

FreshRSS stores multiple authors separated by semicolons. A comma is valid inside a single creator name (for example, `Family name, Given name`) and must not turn it into two authors.

Validation:
- `git diff --check`
- PHP is unavailable in this Windows environment; GitHub Actions will run the project checks.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -568,7 +568,7 @@ class FreshRSS_Entry extends Minz_Model {
 				$value = preg_split('/\s*[;]\s*/', $value, -1, PREG_SPLIT_NO_EMPTY) ?: [];
 				$value = Minz_Helper::htmlspecialchars_utf8($value);
 			} else {
-				$value = preg_split('/\s*[,]\s*/', $value, -1, PREG_SPLIT_NO_EMPTY) ?: [];
+				$value = [$value];
 			}
 		}
 		$this->authors = $value;
--- a/tests/app/Models/EntryTest.php
+++ b/tests/app/Models/EntryTest.php
@@ -1,0 +1,31 @@
+<?php
+declare(strict_types=1);
+
+final class EntryTest extends \PHPUnit\Framework\TestCase {
+	public static function testAuthorsKeepsCommaContainingNameAsSingleAuthor(): void {
+		$entry = new FreshRSS_Entry(authors: 'Smith, John');
+		self::assertSame(['Smith, John'], $entry->authors());
+	}
+
+	public static function testAuthorsKeepsCommaSeparatedListAsSingleAuthor(): void {
+		// Known trade-off: a feed genuinely listing two authors separated only by a comma
+		// is now treated as one author, since comma is no longer used as a separator.
+		$entry = new FreshRSS_Entry(authors: 'Alice, Bob');
+		self::assertSame(['Alice, Bob'], $entry->authors());
+	}
+
+	public static function testAuthorsStillSplitsOnSemicolon(): void {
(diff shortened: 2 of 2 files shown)


app/Models/Entry.php
```diff
@@ -568,7 +568,7 @@ public function _authors(array|string $value): void {
 				$value = preg_split('/\s*[;]\s*/', $value, -1, PREG_SPLIT_NO_EMPTY) ?: [];
 				$value = Minz_Helper::htmlspecialchars_utf8($value);
 			} else {
-				$value = preg_split('/\s*[,]\s*/', $value, -1, PREG_SPLIT_NO_EMPTY) ?: [];
+				$value = [$value];
```
Alkarex, line 571: I think we used to split on comma (to be double-checked more precisely), so we should first reflect a bit on the consequences. Might be ok anyway since this was a while ago

Frenzie: There's probably no right answer (give or take allowing for a custom regex).

JamBalaya56562:
Would you mind adding a test for this one? `_authors()` is a plain setter with no database or network involved, and #9050 already adds `tests/app/Models/EntryTest.php`, which could host it.

It is a one-line change, but it flips the behaviour in both directions: `"Smith, John"` now stays a single author (the fix), while a feed that genuinely lists `"Alice, Bob"` as two authors becomes one. The semicolon branch still splits. A couple of cases pinning those outcomes would make the intended contract explicit and keep it from drifting back.

Elgeryy1:
Added tests pinning both directions of the behaviour change: a comma-containing single name (\`Smith, John\`) no longer gets split, a comma-only list (\`Alice, Bob\`) is now treated as one author, and semicolon-separated lists still split while keeping any comma inside each name intact (\`Smith, John; Doe, Jane\` → two authors). Hosted in \`tests/app/Models/EntryTest.php\` as suggested.

app/Models/Entry.php
Elgeryy1: Agreed, there isn't one — any fixed separator will misparse someone's real data the other way. A configurable custom regex would handle both cases properly, but that's a bigger, separate feature (per-feed config, UI, docs) rather than something this narrow fix should grow to include. Left the trade-off explicit and pinned down by the tests added above; happy to open a follow-up issue for the configurable-regex idea if there's appetite for it.

PR #9134, opened 2026-07-22
Refactor per-feed cURL option allowlist into a data-driven map

Elgeryy1:
## Summary

Follow-up to the discussion on #9012: replaces the flat `CURLOPT_*` allowlist plus the bespoke per-option `if` chain in `FreshRSS_http_Util::sanitizeCurlParams()` with a single map of `CURLOPT_* => sanitizer`.

Previously, adding one more allowed cURL option to `$safe_params` meant touching two places if it needed any special handling (the allowlist array, and a separate `if ($k === ...)` block). Now a single map entry drives both the allowlist and the sanitizing logic for that option.

## Changes

- `sanitizeCurlParams()` now iterates a `CURLOPT_* => callable` map instead of an allowlist array + special cases.
- Behaviour is unchanged: `CURLOPT_COOKIEFILE` is still forced to an empty value, `CURLOPT_HTTPHEADER` still strips `Remote-User`/`X-WebAuth-User` override headers (and is left untouched if not an array), and every other currently-allowed option keeps passing through as before.
- Added `tests/app/Utils/httpUtilTest.php` coverage for the allowlist, the `COOKIEFILE` and `HTTPHEADER` special cases, and plain pass-through options.

## Scope note

This only covers the `sanitizeCurlParams()` side discussed on #9012. The OPML-import side (`ImportService::pushFeed()`'s per-`frss:CURLOPT_*` `if` chain, which also does type coercion and a couple of legacy quirks like the `CURLOPT_PROXYTYPE` `3` → `-1` mapping) is a separate, larger piece of work with its own OPML-compatibility constraints, so I left it out here to keep this PR focused and reviewable.

## Validation

- `phpunit` (targeted + full suite)
- `phpcs`
- `phpstan`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ See also [the FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases).
 	* Improve Spanish [#8878](https://github.com/FreshRSS/FreshRSS/pull/8878)
 	* Improve Ukrainian [#8871](https://github.com/FreshRSS/FreshRSS/pull/8871)
 * Misc.
+	* Refactor the per-feed cURL option allowlist into a single data-driven map, replacing the bespoke per-option sanitizing code [#9012](https://github.com/FreshRSS/FreshRSS/pull/9012)
 	* Update to PHPMailer 7.1.1 [#9807](https://github.com/FreshRSS/FreshRSS/pull/9807)
 	* Improve PHP code [#8906](https://github.com/FreshRSS/FreshRSS/pull/8906)
 	* Fix PHPStan 2.2.2 in CLI [#8911](https://github.com/FreshRSS/FreshRSS/pull/8911)
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -104,36 +104,45 @@ final class FreshRSS_http_Util {
 	}
 
 	/**
+	 * The per-feed cURL options that are allowed to be set, each mapped to a sanitizer that
+	 * receives the candidate value and returns the value to actually keep.
+	 * A single entry here drives both the allowlist and the value sanitization, instead of
+	 * a separate allowlist plus a bespoke per-option `if` chain.
+	 *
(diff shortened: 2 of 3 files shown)


JamBalaya56562:
@Elgeryy1 Apologies for the collision — I think we independently answered the same request from @Alkarex on #9012. I opened #9067 for it on 19 July, which is easy to miss among all the activity.

The two overlap on `sanitizeCurlParams()` and both add cases to `tests/app/Utils/httpUtilTest.php`, so they will conflict. #9067 also covers the parts you scoped out: the OPML import chain, the export side, and the `CURLOPT_PROXYTYPE` `3 → -1` remap that is currently repeated in three places. Doing the export side turned up a bug — `frss:CURLOPT_COOKIEFILE` is never actually exported, because its value is forced to an empty string and the export filter drops empty values, so the cookie engine setting is lost on an OPML round-trip.

The one real design difference is what the map holds. You use a sanitizer callable per option; I use declarative data (`name` / `type` / `values`) so that the same entry also drives the OPML round-trip and value validation. I avoided callables because there are only two special cases today and the repo instructions warn against premature abstraction — but that is a judgement call, not a certainty.

So, a genuine question rather than a rhetorical one: is there anything your version handles better? If the callable map buys something the declarative one cannot, I would rather fold it into #9067 than have the two of us maintain competing refactors. Either way it is the maintainers' call which one to take.

Elgeryy1:
Looked at #9067's diff to answer that properly rather than guess: no, the callable map here doesn't buy anything the declarative one doesn't. `curlParamSpecs()` handles both special cases through its `type` field (\`flag\` for COOKIEFILE, \`lines\` + the auth-header filter for HTTPHEADER), and additionally drives the OPML import/export round-trip and per-value validation, which this PR doesn't touch at all. It's strictly more capable, and it fixes a real bug (COOKIEFILE export) along the way.

Closing this one in favour of #9067. Thanks for digging into the overlap and laying out the comparison.

Issue #7923, opened 2025-09-06
Complete TODOs from source code


Inverle:
<details>
<summary>Generated with this script</summary>

```php
#!/usr/bin/env php
<?php
$last_commit = trim(`git log --format="%H" -n 1`);
$grep = trim(`grep -E -irn '(//|/\*|\* @|#|<!--)\s?(TODO|FIXME)' --exclude-dir i18n --exclude-dir simplepie app lib p cli Makefile *.php *.neon`);
$lines = explode("\n", $grep);
$output = "Result from $last_commit:\n---\n\n### Uncompleted (" . count($lines) . ")\n\n";

foreach ($lines as $line) {
	$split = explode(':', $line);
	if (count($split) < 3) {
		continue;
	}
	$file = $split[0];
	$line = $split[1];
	$match = addslashes(htmlspecialchars(trim(implode(':', array_slice($split, 2))), ENT_NOQUOTES, 'UTF-8'));
	$output .= "* [ ] [$file:$line](https://github.com/FreshRSS/FreshRSS/tree/$last_commit/$file#L$line): <code>$match</code>" . "\n";
}

echo $output;
```
</details>

Result from 8cb2158a390088738eaa27ed3b01d784c35d9f9d:
---

### Uncompleted (73)

* [ ] [app/views/index/reader.phtml:17](https://github.com/FreshRSS/FreshRSS/tree/8cb2158a390088738eaa27ed3b01d784c35d9f9d/app/views/index/reader.phtml#L17): <code>&lt;a href=\"&lt;?= Minz_Url::display(Minz_Request::currentRequest()) ?&gt;\"&gt;&lt;?= _t(\'gen.js.new_article\'); /* TODO: move string in JS*/ ?&gt;&lt;/a&gt;</code>
* [ ] [app/views/index/reader.phtml:55](https://github.com/FreshRSS/FreshRSS/tree/8cb2158a390088738eaa27ed3b01d784c35d9f9d/app/views/index/reader.phtml#L55): <code>&lt;a href=\"&lt;?= Minz_Url::display(Minz_Request::currentRequest()) ?&gt;\"&gt;&lt;?= _t(\'gen.js.new_article\'); /* TODO: move string in JS*/ ?&gt;&lt;/a&gt;</code>
[...]

PR #9136, opened 2026-07-22
Fix tagging an already-tagged entry on PostgreSQL

Elgeryy1:
## Summary
- Picked two small, self-contained TODOs/FIXMEs from #7923 and completed them.

## Changes
- `lib/lib_date.php`: `_noDelimit()` stripped `-` and `:` from the entire ISO 8601 string, including a trailing UTC offset. A negative offset like `2014-02-03T22:00:00-05:00` (usable in `date:`/`pubdate:`/`mdate:`/`userdate:` search filters) silently lost its sign and produced a garbage timestamp. Now a trailing offset (or `Z`) is extracted before stripping delimiters and re-appended unchanged.
- `app/Models/TagDAO*.php`: `FreshRSS_TagDAOPGSQL::sqlIgnore()` returned `''` as a placeholder (the `//TODO`), so re-tagging an already-tagged entry raised a real SQL error on PostgreSQL instead of being a silent no-op like on MySQL/SQLite. Replaced the `sqlIgnore()` keyword-injection approach with the `sqlIgnoreConflict(string $sql): string` pattern already used by `EntryDAO`/`EntryDAOSQLite`/`EntryDAOPGSQL`, and implemented it for PostgreSQL with `ON CONFLICT DO NOTHING`.

## Validation
- `phpunit` (new regression tests for both fixes, plus the full existing `Search`/`LibDate`/`Models` suites)
- `phpcs`
- `phpstan`
- `php -l` on all changed files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,6 @@ See also [the FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases).
 	* New per-feed option to show or hide enclosures (attachments) [#4999](https://github.com/FreshRSS/FreshRSS/issues/4999)
 * Bug fixing
 	* Fix lost elements while parsing search query [#8884](https://github.com/FreshRSS/FreshRSS/pull/8884)
-	* Fix search `date:`/`pubdate:`/`mdate:`/`userdate:` filters with a negative UTC offset (e.g. `-05:00`) parsing to the wrong date [#7923](https://github.com/FreshRSS/FreshRSS/issues/7923)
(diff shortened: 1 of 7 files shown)


Inverle:
Didn't you already make a PR for the first part? #9071

Elgeryy1:
Good catch — yes, #9071 already covers the negative-UTC-offset date fix, I missed that when going through the TODO list. Dropped that part here and kept only the PostgreSQL tag-insert fix, which is the other TODO from #7923 and isn't covered elsewhere.

Alkarex (approved):

PR #9138, opened 2026-07-22
Allow triggering a SQLite export from the Web UI, gated by DB size

### LINKED ISSUE #8183: Export to SQLite from Web UI

Elgeryy1:
## Summary
- Completes the remaining item of the settings-only "PR 1" plan from #8183 (the first two settings, `auto_sqlite_export.enabled` and `.retention`, were added in #8819).
- Adds a `limits.sqlite_export_max_db_size` setting (bytes, default 100 MiB) and a button on *Import / export* to generate a new SQLite export of the current user's database on demand.

## Changes
- `config.default.php`: new `limits.sqlite_export_max_db_size` (0 disables the button; the CLI stays available regardless).
- `app/Models/DatabaseDAO.php`: small pure helper `sizeWithinLimit()` to decide if the current DB size is under the configured max.
- `app/Controllers/importExportController.php`: `indexAction` now exposes whether the export button should show/be enabled; new `sqliteExportAction` triggers the export (POST only, re-checks the size limit server-side, reuses the existing `dbCopy()` SQLite export path).
- `app/views/importExport/index.phtml`: new button, disabled with an explanatory message when the database is over the limit.
- `docs/en/admins/05_Backup.md`: documents the new setting.
- i18n: new English strings added via `cli/manipulate.translation.php` (other languages get the usual `// TODO` placeholders).

## Validation
- `phpunit` (new `DatabaseDAOTest::test_sizeWithinLimit`, full suite otherwise green except the pre-existing Windows-only `tests/cli/i18n/*` EOL noise)
- `phpcs`
- `phpstan` (level 10, clean)
- `php -l` on every changed file

Manually reasoned through the flow (no live multi-DB-size setup here to click through), but the export path itself (`dbCopy()`) is the same one already used by the existing CLI/auto-export and the existing Web UI "download" action.

Fixes #8183.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ See also [the FreshRSS releases](https://github.com/FreshRSS/FreshRSS/releases).
 	* New option to hide badges showing number of unread articles (*Phantom Obligation*) [#8844](https://github.com/FreshRSS/FreshRSS/pull/8844)
 	* New option to keep or not the custom sort order when navigating between categories and feeds [#8969](https://github.com/FreshRSS/FreshRSS/pull/8969)
 	* New per-feed option to show or hide enclosures (attachments) [#4999](https://github.com/FreshRSS/FreshRSS/issues/4999)
-	* Trigger an on-demand SQLite export from the Web UI, gated by a new `limits.sqlite_export_max_db_size` setting [#8183](https://github.com/FreshRSS/FreshRSS/issues/8183)
+	* Trigger an on-demand SQLite export from the Web UI, gated by a new `limits.sqlite_export_max_db_size` setting, with an admin override to force it past that limit [#8183](https://github.com/FreshRSS/FreshRSS/issues/8183)
 * Bug fixing
 	* Fix lost elements while parsing search query [#8884](https://github.com/FreshRSS/FreshRSS/pull/8884)
 * CLI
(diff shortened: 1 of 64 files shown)


Frenzie:
> Adds a limits.sqlite_export_max_db_size setting (bytes, default 100 MiB) and a button on Import / export to generate a new SQLite export of the current user's database on demand.

Ignore this if it's too much effort, but I figure a superuser should probably be able to force it to proceed anyway?

Elgeryy1:
Good call — added a "force" checkbox that only shows up for admins once the DB is over the limit, so they can push through the size gate deliberately instead of being hard-blocked.

JamBalaya56562:
Thanks for working on this. If possible, could you please add before-and-after screenshots of the **Import / export** page showing the new SQLite export controls?

A screenshot of the normal enabled state would be helpful. If practical, it would also be useful to show the size-limit states: the force-export checkbox shown to an administrator and the disabled button shown to a non-administrator.

This would make the new workflow and its different UI states much easier to understand.

Elgeryy1:
Before (no way to trigger a new SQLite export from the Web UI — straight from the heading to the existing archive list):

9138-before-no-export-button

After — normal state (database within the size limit):

9138-export-normal-state

After — over the size limit, as admin (force-export checkbox, button stays enabled):

9138-export-admin-force-state

After — over the size limit, as a non-admin (no checkbox, button disabled):

9138-export-nonadmin-disabled-state

Elgeryy1:
Implemented the requested administrator override in commit 6e5e25d: when the DB exceeds the configured limit, only admins see the force-export checkbox; the server re-checks both the limit and admin authorization on POST. The normal user path remains disabled above the limit.

JamBalaya56562:
@Elgeryy1 
Could you resolve merge conflicts in your PRs?

PR #9143, opened 2026-07-23
Docker/OIDC: redirect to a default URL on expired auth state instead of 400-ing

featurecreep-cron:
## Problem

With OIDC on the official Docker image, interactive SSO logins intermittently return `HTTP 400` at the `/i/oidc/` callback. The failure tracks login duration. `mod_auth_openidc` sets an anti-CSRF state cookie bounded by `OIDCStateTimeout` (default 300 s). If a login through an external IdP takes longer than that (MFA, a slow provider screen, the user stepping away), the cookie is already gone when the callback returns. The image does not set `OIDCDefaultURL`, so the module has no fallback and returns an error instead of restarting the login.

Apache log:

```
oidc_response_proto_state_restore: state has expired
oidc_response_process: invalid authorization response state and
        no default SSO URL is set, sending an error...
→ GET /i/oidc/?code=…&state=… 400
```

This is [documented mod_auth_openidc behaviour](https://github.com/OpenIDC/mod_auth_openidc/blob/72c9f479c2d228477ff0a9518964f61879c83fb6/auth_openidc.conf#L723-L733) when `OIDCDefaultURL` is unset. Both `OIDCDefaultURL` and `OIDCStateTimeout` are `RSRC_CONF` (server config only), so the `./FreshRSS/p/i/.htaccess` escape hatch cannot set them and the fix has to live in the image.

Reproduce: log out, start SSO, wait more than 5 minutes on the IdP screen, complete.

## Fix

Add `OIDC_DEFAULT_URL`, wired through the existing `${OIDC_*}` env pattern, defaulting to `/i/` (the index, consistent with `OIDCRedirectURI /i/oidc/`). On an expired or unmatchable state the module now redirects to `/i/`, which is protected and starts a fresh login against the existing IdP session. `OIDCStateTimeout` stays at 300 s; the window is fine, the missing fallback was the bug.

`OIDC_STATE_TIMEOUT` is intentionally not exposed. It can be added the same way if there is a reason to tune the window.

Files:

- `Docker/entrypoint.sh`
- `Docker/FreshRSS.Apache.conf`
- `docs/en/admins/16_OpenID-Connect.md`

## Verification

Live FreshRSS Docker instance behind Authentik, same 7-minute wait, before and after:

| | Log | Result |
|---|---|---|
| Before | `no default SSO URL is set, sending an error` | `/i/oidc/… → 400` |
| After | `a default SSO URL is set, sending the user to it` | `/i/oidc/ → 302 → /i/ → 302 → /i/oidc/ → 302 → /i/ → 200` (signed in, no loop) |

The state still expires at 300 s; only the recovery changed.


docs/en/admins/16_OpenID-Connect.md
```diff
@@ -46,6 +46,7 @@ OIDC support in Docker is activated by the presence of a non-empty non-zero `OID
...
 * `OIDC_SESSION_MAX_DURATION`: Optional. Maximum duration of the application session. When not defined the default is 8 hours (3600 * 8 seconds). When set to 0, the session duration will be set equal to the expiry time of the ID token.
 * `OIDC_SESSION_TYPE`: Optional. OpenID Connect session storage type. See [mod_auth_openidc’s documentation for details](https://github.com/OpenIDC/mod_auth_openidc/blob/72c9f479c2d228477ff0a9518964f61879c83fb6/auth_openidc.conf#L587-L596).
+* `OIDC_DEFAULT_URL`: Optional. URL the user is redirected to when the OIDC module receives a callback whose anti-CSRF state has expired or cannot be matched, for example when an interactive login (external identity provider, MFA/step-up, or the user stepping away) takes longer than the state timeout. When not defined, the default is `/i/`, the FreshRSS index, so that an expired login restarts against the existing provider session instead of returning an HTTP 400 error page. Maps to mod_auth_openidc’s [`OIDCDefaultURL`](https://github.com/OpenIDC/mod_auth_openidc/blob/72c9f479c2d228477ff0a9518964f61879c83fb6/auth_openidc.conf#L723-L733).
```
Inverle: Maybe link towards a newer commit? https://github.com/OpenIDC/mod_auth_openidc/blob/4f29abe20c25fe942e85d33b612e5be7f62ceb6d/auth_openidc.conf#L841-L845 (latest)

or the version that's in the Debian container (2.4.17):
https://github.com/OpenIDC/mod_auth_openidc/blob/b2e99151bc695335089c8d4bfe5793624ac0732e/auth_openidc.conf#L786-L790

No difference between the two, but there is against the one that's linked here.

Alkarex: Done https://github.com/FreshRSS/FreshRSS/pull/9143/commits/aafbb47b5a455cff4db089837c28312cf71d84e7

Alkarex:
Thanks :-)

PR #9146, opened 2026-07-26
Security: apply encodeURIComponent to share template

Alkarex:
For instance if an article title is `<title>XSS &quot;&gt;&lt;img src=x&gt; END</title>`

Reported by @meifukun

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -822,11 +822,11 @@ function show_share_menu(el) {
 		const title = title_el.textContent;
 		const titleText = title;
 		const template = document.getElementById(templateId).innerHTML
-			.replace(/--entryId--/g, id)
-			.replace(/--link--/g, link)
-			.replace(/--titleText--/g, titleText)
-			.replace(/--websiteName--/g, websiteName)
-			.replace(/--articleAuthors--/g, articleAuthorsText);
+			.replace(/--entryId--/g, encodeURIComponent(id))
+			.replace(/--link--/g, encodeURIComponent(link))
+			.replace(/--titleText--/g, encodeURIComponent(titleText))
+			.replace(/--websiteName--/g, encodeURIComponent(websiteName))
+			.replace(/--articleAuthors--/g, encodeURIComponent(articleAuthorsText));
 
 		div.insertAdjacentHTML('beforeend', template);
 	}


Inverle:
Is there a report for this somewhere?

Alkarex:
Credits @meifukun

# Share Menu HTML Injection via Feed-Controlled Article Metadata

## Summary

An RSS feed publisher can place HTML-breaking characters in article metadata such as the title or author. When a FreshRSS user opens the share menu for that article, the client-side share-menu code substitutes the decoded metadata into an HTML template and inserts it with `insertAdjacentHTML()`, allowing attacker-controlled markup to be injected into the user's FreshRSS page.

Under the default FreshRSS CSP, inline event-handler JavaScript is blocked, so the default impact is HTML injection and UI manipulation. If the deployment relaxes CSP to allow inline scripts, the same issue becomes stored XSS. I verified this on a local self-hosted instance and did not test production services.

## Affected Component

- `app/views/helpers/index/normal/entry_share_menu.phtml`
- `p/scripts/main.js`
- `app/Models/Share.php`
- Observed version: FreshRSS `1.29.2-dev`, commit `cf298e2680ee`

## Root Cause

The share-menu template places a share URL directly into an `href` attribute:

```php
<a target="_blank" rel="noreferrer" href="<?= $share->url() ?>" data-type="<?= $share->type() ?>"><?= $share->name() ?></a>
```

The template contains placeholders such as `--titleText--`. The server-side share URL transforms are applied to those literal placeholders, not to the real article metadata that is substituted later.

On the client side, `show_share_menu()` reads decoded text from the DOM and inserts the resulting string as HTML:

```js
const title = title_el.textContent;
const template = document.getElementById(templateId).innerHTML
    .replace(/--link--/g, link)
    .replace(/--titleText--/g, titleText)
    .replace(/--websiteName--/g, websiteName)
    .replace(/--articleAuthors--/g, articleAuthorsText);

div.insertAdjacentHTML('beforeend', template);
```

If the feed title contains a double quote followed by markup, the substituted value breaks out of the `href` attribute and becomes parsed DOM.

## Proof of Concept / Verification Summary

1. I configured a sharing service in a local FreshRSS instance so the article share dropdown was rendered.
2. I served and subscribed to a local RSS feed containing this title:

```xml
<title>XSS &quot;&gt;&lt;img src=x onerror=alert(document.cookie)&gt; END</title>
```

3. After refreshing the feed, I opened the share menu for the imported article.
4. DOM inspection showed that the title broke out of the share link attribute and inserted a real `img` element with an `onerror` attribute:

```html
<a ... href="https://publish.buffer.com/compose?url=http://example.com/article1&amp;text=XSS ">
  <img src="x" onerror="alert(document.cookie)"> END - Test Feed (xss@attacker.com)" data-type="buffer">Buffer
</a>
```

5. With the default CSP, the browser blocked the inline event handler and no alert fired.
6. With CSP relaxed to allow inline script in the local test environment, the same payload executed.

## Impact

With default CSP, a malicious feed can inject attacker-controlled HTML into the trusted FreshRSS UI when the victim opens the share menu. This can break share links, alter UI content, or create phishing-style markup inside the authenticated FreshRSS page.

If a deployment relaxes CSP, strips CSP, or enables an extension/configuration that allows inline script execution, this becomes stored XSS in the victim's FreshRSS session. In that case, attacker JavaScript can perform actions as the user and read data available through the FreshRSS UI/API.

The issue requires the user to have at least one sharing service configured and to open the share menu for an attacker-controlled article.

## Suggested Fix

Avoid string substitution into HTML followed by `insertAdjacentHTML()`. Build the share menu with DOM APIs and assign URLs via `setAttribute()`, or HTML-escape substituted values before insertion. The share URL should be constructed from the real article metadata using URL encoding and then escaped for the attribute context.

Alkarex:
Additional tests welcome

Inverle:
https://github.com/FreshRSS/FreshRSS/pull/9170

PR #9150, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published image

kobihikri:
Hi, and thanks for FreshRSS.

`.github/workflows/docker-publish.yml` publishes the image, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it cannot check that it was built by this workflow, from this repository, at that tag.

FreshRSS is self-hosted precisely so someone's reading list stays on their own server, and the image is what they deploy. There is nothing between the registry and that deployment which vouches for it today.

The change is two lines on the build step:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the image. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats worth stating: `mode=max` records the full build including build arguments, so `provenance: true` is the smaller option if any have ever been sensitive; and attestations add an extra manifest to the index, which the registry supports.

No SLSA level claimed — the attestation is what BuildKit produces.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -94,3 +94,5 @@ jobs:
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
         push: ${{ (github.ref == 'refs/heads/latest' || github.ref == 'refs/heads/edge' || startsWith(github.ref, 'refs/tags/')) && github.repository_owner == 'FreshRSS' }}
+        provenance: mode=max
+        sbom: true


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

Alkarex:
Hello and thanks for the PR.
I have not reviewed it yet but if you can, please already:
* Provide an example of how this SBOM metadata looks like, both for our Debian and our Alpine images 
* Confirm that it is targeting both Docker Hub and GitHub

kobihikri:
Thanks — both answers below.

**1. What the SBOM looks like, for Debian and for Alpine.**

`sbom: true` makes BuildKit attach an in-toto **SPDX 2.3** attestation to each pushed image (one per platform), listing the packages it detected. To show the shape concretely I built each of your base images (`debian:13-slim` and `alpine:3.24`) with `--sbom=true` and read the attestation back. The two differ exactly as you'd expect, because the package managers differ:

| | base | SPDX | packages | manager |
|---|---|---|---|---|
| `Docker/Dockerfile` | `debian:13-slim` | SPDX-2.3 | ~79 | dpkg |
| `Docker/Dockerfile-Alpine` | `alpine:3.24` | SPDX-2.3 | ~17 | apk |

Debian sample (dpkg names/versions):
```
apt 3.0.3 · base-files 13.8+deb13u6 · bash 5.2.37-2+b9 · coreutils 9.7-3 · dpkg 1.22.22 · dash 0.5.12-12 …
```
Alpine sample (apk names/versions):
```
alpine-baselayout 3.7.2-r1 · apk-tools 3.0.6-r0 · busybox 1.37.0-r31 · musl 1.2.6-r2 · libssl3 3.5.7-r0 · ca-certificates-bundle …
```
Your real images add their own layers on top (PHP, extensions, FreshRSS files), so the published SBOMs will be supersets of these base lists. Anyone can read them after a push with:
```
docker buildx imagetools inspect docker.io/freshrss/freshrss:edge --format '{{ json .SBOM }}'
```

**2. Docker Hub and GitHub — yes, both.** Your workflow's `images:` block (docker-publish.yml) lists both registries:
```
docker.io/freshrss/freshrss
ghcr.io/${{ github.repository }}
```
`build-push-action` attaches the provenance and SBOM attestations to every tag it pushes, so both Docker Hub and GHCR get them, for all three platforms in your `platforms:` matrix (amd64, arm/v7, arm64).

**One heads-up on merge, unrelated to the above:** this branch's protection requires *verified (signed) commits*, and my two commits are currently unsigned, so the merge is blocked on that. I'll get them signed and force-push before you'd merge — flagging it so the red check doesn't surprise you.

Happy to adjust anything.

kobihikri:
Commits are now signed — I collapsed the branch to a single web-flow-signed commit ([`5a94788`](https://github.com/FreshRSS/FreshRSS/commit/5a9478849fe32307cb4a8b37868a4d2498139c37), "Verified"), so the *verified signatures* requirement is satisfied. The diff is unchanged: the two lines (`provenance: mode=max`, `sbom: true`) on the build-push step, +2/-0. Ready whenever you'd like to review.

kobihikri:
Small self-inflicted hiccup: while collapsing to a single signed commit I briefly reset the branch to `edge`, which momentarily gave it a zero diff and auto-closed the PR. I've reopened it — the branch is back to the one verified commit (`5a94788`), +2/-0, nothing changed in the diff. Sorry for the notification noise.

Alkarex:

PR #9152, opened 2026-07-29
Exclude hidden feeds from global view counts

JamBalaya56562:
Follow-up to #9132

## Summary

- Exclude feeds below `FreshRSS_Feed::PRIORITY_FEED` from the global view.
- Reuse the same filtered feed list for category visibility, category counts, and feed rendering.
- Keep the existing behavior for visible feeds and unread counts.

## Root cause

The global view skipped hidden feeds only while rendering feed rows. Category visibility and `STATE_ALL` totals still used every feed in the category, so hidden entries inflated category counts and categories containing only hidden feeds rendered as empty boxes.

## Before / After

| Before | After |
| --- | --- |
| `Mixed visibility` reports 8 articles while its only visible feed reports 3, and `Hidden only` appears as an empty category with 4 articles.<br><br>Before: hidden feeds inflate global view counts | `Mixed visibility` reports the same 3 articles as its visible feed, and `Hidden only` is omitted.<br><br>After: global view counts only visible feeds |

## Testing

- Verified the before/after behavior with the same SQLite data, Origine theme, English locale, and 1400×1000 viewport.
- Verified that a search matching only hidden-feed entries renders no empty category.
- Verified that the normal unread view remains at 3 for both the visible category and feed.
- `composer run-script phtml-lint`
- `composer run-script phpcs`
- `composer run-script phpstan`
- `composer run-script phpunit` (664 tests, 1231 assertions)
- `make test-all`

### DIFF
--- a/app/views/index/global.phtml
+++ b/app/views/index/global.phtml
@@ -56,8 +56,9 @@
 		FreshRSS_Context::isStateEnabled(FreshRSS_Entry::STATE_NOT_READ)));
 
 	foreach ($this->categories as $cat) {
-		$feeds = $this->feedIdsMatching === null ? $cat->feeds() :
-			array_filter($cat->feeds(), fn(FreshRSS_Feed $feed): bool => !empty($this->feedIdsMatching[$feed->id()]));
+		$feeds = array_filter($cat->feeds(), fn(FreshRSS_Feed $feed): bool =>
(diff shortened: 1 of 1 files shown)


Alkarex (approved):

Alkarex:
Good catch!

PR #9167, opened 2026-08-01
Issue #9166: Detect JSON Feed URLs when subscribing via the API

### LINKED ISSUE #9166: [Feature] Need way to subscribe to a JSON feed via the Google Reader API

The Google Reader API implementation assumes the feed URL is an RSS or Atom feed. This makes it impossible for a client to subscribe to a JSON Feed.

Issue #7574 mentions this, but the associated pull request does not address the API.

I intend to work on this and submit a pull request.

### Use case

A website might offer a JSON feed, and a user might want to subscribe to it using an app that syncs with FreshRSS.

Ideally, the API would detect that a feed is a JSON feed and set the `kind` accordingly.

Another option would be for the client to send the feed type as a separate parameter.

### Additional context

I develop [Unread](https://www.goldenhillsoftware.com/unread/) and am working on syncing with FreshRSS. I encountered this issue very quickly in testing, that I could not subscribe to JSON feeds.

jbrayton:
This pull request makes it possible for Google Reader API clients to subscribe to JSON feeds. It addresses issue #9166.

- When subscribing to a JSON feed via the Google Reader API, determine whether the feed is a JSON feed. If so, set the `kind` accordingly.

1. Determine the URL for your server.
2. Determine the GoogleLogin `Authorization` header to use with a valid account. 
3. Try subscribing to JSON feeds using the command below, after replacing `(auth header)` with an appropriate `Authorization` header and replacing `(server) with your server URL:

```
curl -i -H (auth header) \
-d "ac=subscribe&s=feed/https://www.goldenhillsoftware.com/feed.json&t=my%20feed" \
"(server)/reader/api/0/subscription/edit"
```

4. Verify in FreshRSS that the feed has articles.
5. Unsubscribe.
6. Try to subscribe again, using `quickadd`. Do the same replacement as done in Step 3:

```
curl -i -H "(auth header)" \
-d "quickadd=feed/https://www.goldenhillsoftware.com/feed.json" \
"(server)/api/greader.php/reader/api/0/subscription/quickadd"
```

Pull request checklist:

- [ X ] clear commit messages
- [ X ] code manually tested
- [ X ] documentation updated

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -146,6 +146,7 @@ People are sorted by name so please keep this order.
 * [jlefler](https://github.com/jlefler): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:jlefler)
 * [Joe Stump](https://github.com/joestump): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:joestump), [Web](http://stu.mp)
 * [Joel Garcia](https://github.com/joelchrono12): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:joelchrono12), [Web](https://joelchrono12.xyz)
+* [John Brayton](https://github.com/jbrayton): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:jbrayton), [Web](https://www.virtualsanity.com)
 * [Jonas Östanbäck](https://github.com/cez81): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:cez81)
 * [Jordi Garcia](https://github.com/jgtorcal): [contributions](https://github.com/FreshRSS/FreshRSS/pulls?q=is:pr+author:jgtorcal)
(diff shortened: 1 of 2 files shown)


Alkarex:
Thanks for the PR!

Related:
* https://github.com/FreshRSS/FreshRSS/pull/9058

Making multiple HTTP requests when adding a feed would not be acceptable (risks of getting banned), but either a heuristic based on containing `json` like in https://github.com/FreshRSS/FreshRSS/pull/9058 or some kind of HTTP caching (including the appropriate HTTP accept headers) could be viable

jbrayton:
When I test this out, I only see one initial request to my server from FreshRSS. It looks like `loadJson()` uses `FreshRSS_http_Util::httpGet`, which uses a cache.

But if I am missing something, or if you would still prefer I use heuristics around the URL string, I am happy to do that.

Thank you.

Alkarex:
The multiple requests would occur when the feed is not JSON.
For now, I would prefer a heuristic based on the URL containing `json` like in https://github.com/FreshRSS/FreshRSS/pull/9058

jbrayton:
I updated the pull request to use heuristics based on the URL string.

Thank you!

Alkarex (approved):

PR #9168, opened 2026-08-02
Sanitize all fields during JSON import

Inverle:
fix [GHSA-hx52-rwx7-wg87](https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-hx52-rwx7-wg87)

### DIFF
--- a/app/Controllers/importExportController.php
+++ b/app/Controllers/importExportController.php
@@ -262,7 +262,7 @@ class FreshRSS_importExport_Controller extends FreshRSS_ActionController {
 				continue;
 			}
 			if (filter_var($url, FILTER_VALIDATE_URL) === false) {
-				$message = 'TXT import: skipping invalid URL “' . $url . '”';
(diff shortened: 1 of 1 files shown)


Alkarex (approved):

app/Controllers/importExportController.php
```diff
@@ -482,8 +489,10 @@ private function importJson(string $article_file, bool $starred = false): bool {
 			} else {
 				$url = '';
 			}
+			$url = Minz_Helper::htmlspecialchars_utf8(FreshRSS_http_Util::checkUrl($url) ?: '');
 
 			$title = is_string($item['title'] ?? null) ? $item['title'] : $url;
+			$title = Minz_Helper::htmlspecialchars_utf8($title);
```
Inverle, line 495: I think this introduced a regression (unless it was already there, can't check right now)

A title with `&` (in freshly exported JSON) turned into `&amp;` after importing.

PR #7489, opened 2025-04-01
Use HTTP POST for logout

Alkarex:
To avoid potential CSRF risks

PR #9169, opened 2026-08-02
Refactor to use CSRF field directly instead of borrowing it from a form

Inverle:
HTML injections outside of a form with a `_csrf` field set, could instead refer to the CSRF token form with a payload such as:
```html
<button form="post-csrf" formaction="./i/?c=auth&a=logout" formmethod="post">Continue reading</button>
```
While the `post-csrf` form is useful for reducing the amount of code needed (no need to repeat the `_csrf` field in forms), I don't think it's worth the added security risk.

### DIFF
--- a/p/themes/Alternative-Dark/adark.css
+++ b/p/themes/Alternative-Dark/adark.css
@@ -323,7 +323,7 @@ a:hover > .spinner {
 }
 
 .nav-list .item > a,
-.nav-list .item > .as-link {
+.nav-list .item .as-link {
 	padding: 0 1rem;
 }
 
@@ -368,7 +368,7 @@ a:hover > .spinner {
 
 .dropdown-menu .item > a,
 .dropdown-menu .item > span,
-.dropdown-menu .item > .as-link {
+.dropdown-menu .item .as-link {
 	padding: 0 22px;
 	line-height: 2.5;
 	font-size: inherit;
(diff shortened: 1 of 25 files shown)


p/themes/Alternative-Dark/adark.css
```diff
@@ -323,7 +323,7 @@ a:hover > .spinner {
 }
 
 .nav-list .item > a,
-.nav-list .item > .as-link {
+.nav-list .item .as-link {
```
Frenzie, line 326: Is this sufficiently unique? Put another way, I personally probably wouldn't put in a `>` unless it was necessary but maybe others have different habits.

Inverle: For `.as-link` it should be. See #7489, #7526 where the class was added.

Alkarex:
We have other forms though, e.g.:

https://github.com/FreshRSS/FreshRSS/blob/6bd14a612145142035dea0689cc25c2a4effa9a9/app/layout/aside_feed.phtml#L40-L41

Inverle:
I guess we'll have to refactor those too probably. I didn't realize there were more.

Alkarex:
> I guess we'll have to refactor those too probably. I didn't realize there were more.

I have not checked in details, but removing the `id="..."` might be sufficient

Inverle:
> > I guess we'll have to refactor those too probably. I didn't realize there were more.
> 
> I have not checked in details, but removing the `id="..."` might be sufficient

Not in all cases https://github.com/FreshRSS/FreshRSS/pull/9205

PR #9170, opened 2026-08-02
Fix HTML injection via decoded website attribute in feed menu

Inverle:
To reproduce:
1. Add a feed e.g. `https://github.com/FreshRSS/FreshRSS/releases.atom`
2. Change its website URL to `https://github.com/FreshRSS/FreshRSS/releases.atom?"><marquee>test</marquee><?x`
3. Open the feed's menu from the sidebar after saving the settings

Follow-up of https://github.com/FreshRSS/FreshRSS/pull/9146 (same issue but in a different menu)

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1147,9 +1147,10 @@ function init_column_categories() {
 				const id = itemId.substr(2);
 				const feed_web = a.getAttribute('data-fweb') || '';
 				const template = document.getElementById(templateId)
-					.innerHTML.replace(/------/g, id).replace('http://example.net/', feed_web);
+					.innerHTML.replace(/------/g, id);
 				div.insertAdjacentHTML('beforeend', template);
 				dropdownMenu = div.querySelector('.dropdown-menu');
+				dropdownMenu.querySelector('li.website > a').href = feed_web;
 				dropdownMenu.style.opacity = '0%'; // Hide initially to prevent dropdown flashing
 				if (feed_web == '') {
 					const website = div.querySelector('.item.link.website');


p/scripts/main.js
```diff
@@ -1147,9 +1147,10 @@ function init_column_categories() {
 				const id = itemId.substr(2);
 				const feed_web = a.getAttribute('data-fweb') || '';
 				const template = document.getElementById(templateId)
-					.innerHTML.replace(/------/g, id).replace('http://example.net/', feed_web);
+					.innerHTML.replace(/------/g, id);
```
Frenzie, line 1150: The referenced PR writes `encodeURIComponent(id)`?

Inverle: Yes, the feed ID isn't user input though, so it's not necessary.
There are plenty of other places where you can have HTML injection with direct database write access.

And here it can only be an integer anyway:
https://github.com/FreshRSS/FreshRSS/blob/ee6dd7b73184a78555c4843adb5cdf8c758159e9/p/scripts/main.js#L1137-L1150
https://github.com/FreshRSS/FreshRSS/blob/ee6dd7b73184a78555c4843adb5cdf8c758159e9/app/layout/aside_feed.phtml#L152-L153
https://github.com/FreshRSS/FreshRSS/blob/ee6dd7b73184a78555c4843adb5cdf8c758159e9/app/Models/Feed.php#L91-L93

Frenzie: Alright, sure. It just stood out as being quite different when you wrote "same as". ;-)

Inverle: I meant the attribute contents being HTML encoded, and that when they're accessed from JS, they're not encoded anymore.

Alkarex (approved):

PR #9171, opened 2026-08-02
Fix auth CSRFs (login and register actions)

Inverle:
This is done by getting rid of the CSRF check allowlist, and instead of refreshing the CSRF token on every unauthenticated page load, only refreshing the token when needed (on login, logout, account delete, HTTP auth user change)

In addition, the actualize actions are now correctly protected as well.

The most notable part is that the user create action could potentially be exploited by another website for XSS on a given instance, by submitting the register form with known credentials, and then by the attacker logging into the same account forcibly registered by a malicious website, and setting custom User JS code.
Login action was likely not exploitable in any meaningful way, because of the JS nonce.

It's worth noting one risk of merging this PR, some vulnerabilities discovered in the past had reduced impact due to CSRF token refreshing on every page load: https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-jf4v-f8p2-8xvq#:~:text=It%27s%20important,FreshRSS%5FAuth%3A%3AremoveAccess%28%29%2E


Alkarex:
> It's worth noting one risk of merging this PR, some vulnerabilities discovered in the past had reduced impact due to CSRF token refreshing on every page load

Yes. I have started a few times the review of this PR but got sidetracked every time and worried about the drawbacks being maybe greater than the advantages.

Something else which I see positive with the current behaviour, but which I have not debugged fully, is situations when an old(ish) view is re-opened from a Web browser history (that happens to me on mobile in particular) with the view already scrolled down. I would not want articles to be automatically marked as read by accident.

Frenzie:
Yes, there's something very strange there that happens on mobile sometimes which I've never taken the time to look into, in part because it's hard to reproduce. You reopen the page from yesterday or maybe it is even longer ago, you tap to open something, it opens, and then suddenly the whole page refreshes. This doesn't lead to anything being marked as read so it just looks a bit odd.

Inverle:
> > It's worth noting one risk of merging this PR, some vulnerabilities discovered in the past had reduced impact due to CSRF token refreshing on every page load
> 
> Yes. I have started a few times the review of this PR but got sidetracked every time and worried about the drawbacks being maybe greater than the advantages.
> 
> Something else which I see positive with the current behaviour, but which I have not debugged fully, is situations when an old(ish) view is re-opened from a Web browser history (that happens to me on mobile in particular) with the view already scrolled down. I would not want articles to be automatically marked as read by accident.

I think I ran into the bug myself before, but I don't think it should make a difference here, since the CSRF token is not being refreshed in authenticated sessions anyway.

(I meant that anonymous view had the CSRF token changed on every page load)

Alkarex (approved):

PR #9172, opened 2026-08-03
docs: add Zenith Hosting deploy option

odpay:
hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we added FreshRSS to our marketplace, so this PR adds a small "Deploy with Zenith" badge to the automated install table (links to https://zenith.hosting/host/freshrss).

one-click managed FreshRSS: storage, backups, outbound email and a free subdomain included, no server or config files needed. a share of every subscription goes back to FreshRSS.

note on the diff: the table is 3 columns wide, so adding one button re-flows the last two rows. nothing was removed or re-pointed, the existing buttons just shift one cell and the last row is now full instead of having a trailing empty cell. happy to move the button to the end of the grid instead if you'd rather keep the current order.

we share a cut of revenue with the projects we host. happy to explain if you're curious: hello@zenith.hosting


Inverle:
Should be added to https://github.com/FreshRSS/freshrss.org/blob/91bec77a40de33e0d5f8bd963e5996e5d263340d/pages/cloud-providers.html#L201 as well

See https://freshrss.org/cloud-providers.html#dedicated-managed

PR #9173, opened 2026-08-04
Fix “mark as read older than…” widening the active search

TowyTowy:
Fixes the "Mark as read → Articles older than one day / one week" action marking far more than requested when a search is active.

Sibling searches of a `FreshRSS_BooleanSearch` are combined by OR (`BooleanSearch::__toString()`, and "Searches are combined by OR" in `EntryDAO::sqlBooleanSearch()`). `prepend()` added the maximum publication date as one such sibling, so it widened the selection instead of restricting it:

```
search='cats'         → (e.date <= ?) OR (…content… LIKE '%cats%')
search='cats OR dogs' → (e.date <= ?) OR (…'%cats%') OR (…'%dogs%')
```

`nav_menu.phtml` puts `search` and `maxPubDate` on the same mark-as-read URL, so with any search active and the stream sorted by publication date the action marks as read every article older than the cut-off (search ignored) **plus** every article matching the search at any date. The `idMax` path used when sorting by ID is not affected.

`enforce()` already had the correct behaviour (wrap the existing expression, then prepend) and is used that way in `indexController`. This moves that wrapping into a small private helper shared by `enforce()` and `prepend()`, so both combine with AND:

```
search='cats'         → (e.date <= ?) AND ((…'%cats%'))
search='cats OR dogs' → (e.date <= ?) AND ((…'%cats%') OR (…'%dogs%'))
search=''             → (e.date <= ?)          (unchanged)
```

`prepend()` has a single caller, `entryController::readAction()`, added together with `prepend()` itself in #8163. `enforce()` output is unchanged — the generated SQL, bound values and `__toString()` were diffed across 14 query shapes × 2 injected search types before and after, identical.

How to test manually:
1. Configuration → Reading → sort by publication date.
2. Type something in the search box that matches only some articles.
3. Mark as read → "Articles older than one day". Only articles that match the search *and* are older than the cut-off are marked read.

Checklist:
- [x] clear commit messages
- [x] code manually tested
- [x] unit tests written (`tests/app/Models/BooleanSearchTest.php`, asserts the generated SQL for empty / flat / OR searches)
- [x] documentation updated (n/a)

Written with assistance from Claude (Anthropic); fail-before/pass-after verified, plus the full suite (667 tests), phpstan and phpcs.

### DIFF
--- a/app/Models/BooleanSearch.php
+++ b/app/Models/BooleanSearch.php
@@ -432,8 +432,30 @@ class FreshRSS_BooleanSearch implements \Stringable {
 		return $this->operator;
 	}
 
-	/** @param FreshRSS_BooleanSearch|FreshRSS_Search $search */
+	/**
+	 * Wrap the existing searches in a single BooleanSearch if needed,
+	 * so that another search can be added as an additional restriction (AND)
+	 * instead of being combined by OR with the existing sibling searches.
+	 */
+	private function wrapSearches(): void {
+		if (count($this->searches) > 1 || (count($this->searches) > 0 && $this->searches[0] instanceof FreshRSS_Search)) {
+			$wrap = new FreshRSS_BooleanSearch('');
+			foreach ($this->searches as $existingSearch) {
+				$wrap->add($existingSearch);
+			}
+			if (count($wrap->searches) > 0) {
+				$this->searches = [$wrap];
+			}
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


Alkarex (approved):

Alkarex:
Thanks, good catch!
I have slightly fixed the SQL white-space.

PR #6272, opened 2024-04-06
Allow multiple authors on enclosures

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/5066

PR #9174, opened 2026-08-05
Fix only the last <media:credit> being written to the RSS output

### LINKED ISSUE #5066: [BUG] RSS standard: multiple credits not shown
**Describe the bug**
Only the first credit of enclosures is shown of the 5th example of https://www.rssboard.org/media-rss#examples (as RSS feed: https://raw.githubusercontent.com/math-GH/testfeed/main/rssboard.org/example5.xml)

grafik

That is the source:
grafik

**To Reproduce**
1. subscribe to https://raw.githubusercontent.com/math-GH/testfeed/main/rssboard.org/example5.xml
2. see that only the first credit is shown

**Expected behavior**
Both credits are shown

TowyTowy:
## Problem

`app/views/index/rss.phtml` builds the `<media:credit>` elements for an enclosure in a loop, but assigns instead of appends:

```php
$mediaCredits = '';
foreach ($credits as $credit) {
    $mediaCredits = '<media:credit>' . $credit . '</media:credit>';
}
```

So an enclosure with three credits emits only the last one. Every iteration throws away what the previous one built.

This came in with #6272 ("Allow multiple authors on enclosures", which closed #5066). Parsing and display got multi-credit support, but FreshRSS's own generated RSS still passes on a single credit, so a downstream reader subscribing to a FreshRSS feed sees the same symptom #5066 described.

## Fix

One character, `=` to `.=`.

## Tests

Added `tests/app/Views/rssViewTest.php`, which renders `index/rss.phtml` for an entry with an enclosure and asserts on the output:

- three credits produce three `<media:credit>` elements (fails before, passes after)
- a legacy string credit (entries < FreshRSS 1.24) still produces one

Ran locally against edge: `make test` (669 tests, all green), `phpcs` and `phpstan` (level 9) clean on both changed files.

## AI disclosure

Written with Claude (Claude Code). I verified the fix and the tests locally, including the failing-before state, and traced the regression back to #6272.

### DIFF
--- a/app/views/index/rss.phtml
+++ b/app/views/index/rss.phtml
@@ -83,7 +83,7 @@ foreach ($this->entries as $item) {
 					}
 					$mediaCredits = '';
 					foreach ($credits as $credit) {
-						$mediaCredits = '<media:credit>' . $credit . '</media:credit>';
+						$mediaCredits .= '<media:credit>' . $credit . '</media:credit>';
 					}
 
 					// https://www.rssboard.org/media-rss
--- a/tests/app/Views/rssViewTest.php
+++ b/tests/app/Views/rssViewTest.php
@@ -1,0 +1,49 @@
+<?php
+declare(strict_types=1);
+
+final class rssViewTest extends \PHPUnit\Framework\TestCase {
+
+	#[\Override]
+	public static function setUpBeforeClass(): void {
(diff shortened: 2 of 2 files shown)


Alkarex:
Thanks! And sorry for the delay.

PR #9182, opened 2026-08-07
Add optional feed argument to actualize_script.php

jtracey:

Adds an optional argument to the script to allow updating a particular feed, instead of all of them. This makes it easier to update specific feeds via cron and other local utilities. (My personal motivation is to add a bit of privacy to updates so they aren't as obviously linked, but this is also useful for other reasons, e.g. if you have too many feeds to feasibly update at once.)

1. apply patch
2. look up currently valid feed ID
3. call `actualize_script.php 123`

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] documentation updated

I'll update the documentation if I get the go-ahead on this being desirable, and can add a test if requested and not too difficult.

### DIFF
--- a/app/actualize_script.php
+++ b/app/actualize_script.php
@@ -9,11 +9,34 @@ ob_start();
 
 $begin_date = date_create('now');
 
+// Optional feed ID argument.
+// If none given, all feeds are updated.
+$feedId = null;
+if ($argc > 2) {
+	fwrite(STDERR, "actualize_script.php takes at most one feed ID as an argument, aborting\n");
+	die();
+}
+if ($argc == 2) {
+	$feedId = filter_var($argv[1], FILTER_VALIDATE_INT, [
+		'options' => ['min_range' => 1],
+	]);
+
+	if (!$feedId) {
+		fwrite(STDERR, 'Invalid feed ID ' . $argv[1] . "\n");
+		die();
+	}
+}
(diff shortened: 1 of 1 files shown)


Frenzie:
As far as the ID goes is that the reaching the exact same code under the hood as `?c=feed&a=actualize&id=27`?

Alkarex (changes requested):
Hum, I am not sure this makes sense for this script, which is used to refresh all users.
Instead, I suggest you add this new parameter to https://github.com/FreshRSS/FreshRSS/blob/edge/cli/actualize-user.php

Frenzie:
But then you'd have to worry about which user it belongs to. I don't think that makes sense within the confines of what OP said they wanted to do since then you'd have to collect all the users first as boilerplate in order to achieve the same end result. :-)

Alkarex:
> But then you'd have to worry about which user it belongs to. I don't think that makes sense within the confines of what OP said they wanted to do since then you'd have to collect all the users first as boilerplate in order to achieve the same end result. :-)

But feed ID only makes sense for a given user

Frenzie:
Fair enough, but then I think they might be looking for something slightly different?

jtracey:
Thanks for the suggestion, closing in favor of #9319.

PR #9186, opened 2026-08-08
Allow modifiers for open website shortcut

### LINKED ISSUE #1256: Open article in new background tab shortcut

Inverle:
Minimal changes borrowed from https://github.com/FreshRSS/FreshRSS/pull/8990

Allows usage of <kbd>Ctrl/Shift</kbd> keys with the open website shortcut, so that you can open an article either in a new background tab or a new window.

Doesn't work in Chromium browsers, the website opens normally in a new tab there instead. (Firefox only)

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -1205,12 +1205,12 @@ function init_shortcuts() {
 	});
 
 	document.addEventListener('keydown', ev => {
-		if (ev.ctrlKey || ev.metaKey || (ev.altKey && ev.shiftKey) || ev.target.closest('input, select, textarea')) {
-			return;
-		}
-
 		const s = context.shortcuts;
 		let k = (ev.key.trim() || ev.code || 'Space').toUpperCase();
+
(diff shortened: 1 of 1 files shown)


Inverle:
> Doesn't work in Chromium browsers, the website opens normally in a new tab there instead. (Firefox only)

It's possible to dispatch a click MouseEvent (instead of `window.open`) to make it work with shiftKey, but doesn't work with ctrlKey or metaKey so not worth implementing.

Frenzie (approved):
The diff is a bit hard to read, but as long as it doesn't reject opening anything at all I suppose it should be fine.

Alkarex (approved):

PR #9202, opened 2026-08-14
Improve aside keyboard navigation

Inverle:
### Feature changes

* It's now possible to jump from categories to feeds (press shift+j when focused on a category)
   * Similarly, it's possible to jump from a feed to the current category by using alt+k
* If the very first feed is focused and the user presses shift+k, there is no jump to the last unread feed in that category anymore, instead nothing happens.
   * Same if the very last is focused and the user presses shift+j (no jump to the first unread feed in that category)
* Alt+home/end (first/last category) now jumps regardless if the first/last category is read or unread
   * In addition, shift+home/end does not jump between unread feeds anymore, now it jumps between the first visible feed in the category and the last visible feed in the category 
* Alt+h (jump to next unread category) now fall backs to the same behavior as alt+j (jump to next category) if there are no more unread categories after

### Minor code changes

* Removed dead code in `prev_feed`
* `delayedClick(null)` will now throw an error

### DIFF
--- a/p/scripts/main.js
+++ b/p/scripts/main.js
@@ -612,7 +612,7 @@ function next_unread_entry(skipping) {
 	toggleContent(new_active, old_active, skipping);
 }
 
-function prev_feed(jump_to_unread) {
+function prev_feed() {
 	let found = false;
 	let adjacent = null;
 	const feeds = document.querySelectorAll('#aside_feed .feed');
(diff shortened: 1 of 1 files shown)


p/scripts/main.js
```diff
@@ -323,9 +323,7 @@ function send_mark_queue_tick(callback) {
 const delayedFunction = send_mark_queue_tick;
 
 function delayedClick(a) {
-	if (a) {
-		delayedFunction(function () { a.click(); });
-	}
+	delayedFunction(function () { a.click(); });
```
Alkarex, line 326: Are we sure `a` is always defined? Since it can come from a CSS selector, this looks a bit risky

Inverle: I'm not aware of any case where it is not. Not behaving like `querySelector('#nonexistentelement').click()` (which would throw an error) made it harder for me to realize I mistyped my selector when opening the web console.

Issue #4086, opened 2021-12-26
Import OPML file not in alphabetical order, but in the order as in the OPML file


Wombat2018:
Hello,

I am new to FreshRSS and have a question.

When I import my OPML file, the feeds are automatically sorted alphabetically. However, I would like to maintain the order as in the OPML file. Is this possible?

math-GH:
Right now all feeds within a category are ordered alphabetical. It does not matter whether they are imported via OPML or manually typed in.

math-GH:
see also this Feature Request: #1875

Wombat2018:
Would be nice to have this feature. In Firefox RSS Addons like Feedbro and Brief you can sort them manually by drag and drop.

math-GH:
When do you need a manual order? In the views, the articles are ordered chronologically (independent of the feed order)

Wombat2018:
Usually, I check feed by feed in an order that I have defined. I don't need a general overview, where the entries from all feeds are sorted chronologically.

PR #9203, opened 2026-08-14
Order categories during OPML import/export according to position

Inverle:
F.ixes https://github.com/FreshRSS/FreshRSS/issues/4086 (partially, only for categories)

I exported an OPML, and after importing it on a different instance I noticed the categories weren't in order.

This was both due to the position being ignored during the export, and the import.

### DIFF
--- a/app/Services/ImportService.php
+++ b/app/Services/ImportService.php
@@ -61,13 +61,17 @@ class FreshRSS_Import_Service {
 		// existing categories later.
 		$categories = $this->catDAO->listCategories(prePopulateFeeds: false);
 		$categories_by_names = [];
-		$highest_position = PHP_INT_MIN; // To order the positions as best as possible in case of negative positions, we choose the lowest number first
+		$highest_position = PHP_INT_MIN; // To order the numbers as best as possible in case of negative positions, we choose the lowest possible number first
 		foreach ($categories as $category) {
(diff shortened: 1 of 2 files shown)


app/Services/ImportService.php
```diff
@@ -91,7 +96,8 @@ public function importOpml(string $opml_file, ?FreshRSS_Category $forced_categor
 				$can_create_category = FreshRSS_Context::$isCli || !$limit_reached;
 
 				if ($can_create_category) {
-					$category = $this->createCategory($category_element, $dry_run);
+					// Import category in the exact order as the outline's placement in the OPML, at the end of positioned categories
+					$category = $this->createCategory($category_element, $dry_run, ++$highest_position);
 					if ($category !== null) {
```
Inverle, line 105: The `$highest_position` increment is done before it's checked if the category append succeeds, but maybe it is fine this way, since currently this makes it easier to import any failed categories as a separate OPML and correct the position numbers?
([no drag and drop is implemented yet](https://github.com/FreshRSS/FreshRSS/pull/2592#issuecomment-546367257))

Or more accurately: one would have to create the failed categories and change the numbers manually, and then drag the failed feeds from "Uncategorized" also manually, where they are unordered.

Well that could be hard in some cases..

PR #9205, opened 2026-08-14
Follow-up #9169 remove form IDs everywhere

Inverle:
Follow-up #9169


app/layout/aside_feed.phtml
```diff
@@ -37,7 +37,7 @@
 	<a href="<?= _url('index', 'about') ?>" class="about"><?= _t('index.menu.about') ?></a>
 	<?php } ?>
 
-	<form id="mark-read-aside" method="post">
+	<form class="mark-read-aside" method="post">
```
Frenzie: Is that also intended for potential side effects like selecting it in JS?

Inverle: There are no references to it in JS

Frenzie: Not in here, but that's besides the point. :-) You want important elements like forms to be easily selectable with with an ID or a name and I don't think a classname quite fills the same role.

Inverle: Well it can't be an ID, since there is a `_csrf` field inside of it (and it's exactly the reason why it's being changed)

Frenzie: Then use the former ID as a name (not exclusive with a classname), which holds the additional advantage that document.forms is the most natural method. Any console user, UserJS author or extension author will be mildly displeased otherwise.

Alkarex: So `name="xxx"` ? Sounds ok

Frenzie: Indeed.

Inverle: Done in 0359e836725622edf75f9e86286151da44b20e7d, but I left the class names too.

p/scripts/main.js
```diff
@@ -2249,7 +2249,7 @@ function load_more_posts() {
 		}
 
 		const html = this.response;
-		const streamFooter = document.getElementById('stream-footer');
+		const streamFooter = document.querySelector('.stream-footer');
```
Frenzie, line 2252: Minor aside, that means you can now also use `document.forms.stream_footer`.

Frenzie (approved):
Didn't test, but lgtm

PR #9207, opened 2026-08-15
Fix all broken links in the entire repository

Inverle:
Not only broken, but also upgrading from HTTP to HTTPS in some places, and changed some links if they were redirecting to somewhere else.

Follow-up of https://github.com/FreshRSS/FreshRSS/pull/8548 where broken links were only being fixed in the documentation.

Needed for https://github.com/FreshRSS/FreshRSS/pull/9209

PR #9209, opened 2026-08-15
Add a link checker workflow using lychee

Inverle:
Depends on #9207

https://github.com/lycheeverse/lychee-action


.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,64 @@
...
+          # Won't be available to forks in pull requests due to restrictions, so falls back to default
+          # GITHUB_TOKEN with 1,000 requests per hour instead.
+          token: ${{ secrets.LYCHEE_GITHUB_TOKEN || secrets.GITHUB_TOKEN }}
+
+          fail: true
+
+          # Config is loaded from lychee.toml
+          args: >
+            --exclude-path '(^|/)\.lycheeignore$'
+            ${{ github.event_name == 'pull_request' && '--files-from changed-files.txt' || '.' }}
```
Inverle, line 61: It looks like `--files-from` ignores the `extensions` setting:

```console
$ printf ".lycheeignore" > files.txt
$ lychee --files-from files.txt --extensions txt
[EXCLUDED] http://wiremock:8080/ (at 4:1) | This is due to your 'exclude' values
[EXCLUDED] http://freshrss/ (at 5:1) | This is due to your 'exclude' values
[EXCLUDED] http://freshrss/ (at 31:1) | This is due to your 'exclude' values
 [ERROR] https://publish/ (at 16:1) | Connection failed. Check network connectivity and firewall settings
 [ERROR] http://a/ (at 10:1) | Connection failed. Check network connectivity and firewall settings
 [ERROR] https://web/ (at 11:1) | Connection refused - server may be down or port blocked

...



```

`lycheeignore` is not included in `extensions`.

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,64 @@
+name: Link Checker
+
+on:
+  push:
+    branches: [ edge ]
+  pull_request:
+    branches: [ edge ]
+  schedule:
+    - cron: '30 12 * * *'
```
Frenzie: Every day seems a bit much.

Inverle: Should it be every 3 days or 1 week?

Frenzie: I'd start with a month.

Inverle: Done

Alkarex:
I have not checked: is it set up to distinguish between status codes:
* 302 and 307: should be kept as-is (default)
* 301 and 308: should be updated

Alkarex:
Security settings updated accordingly:

image

Inverle:
> I have not checked: is it set up to distinguish between status codes:
> 
>     * 302 and 307: should be kept as-is (default)
> 
>     * 301 and 308: should be updated

On a full scan it shows the redirect status codes: https://github.com/Inverle/FreshRSS/actions/runs/31907721553

Alkarex:
> > I have not checked: is it set up to distinguish between status codes:
> > ```
> > * 302 and 307: should be kept as-is (default)
> > 
> > * 301 and 308: should be updated
> > ```
> 
> On a full scan it shows the redirect status codes: https://github.com/Inverle/FreshRSS/actions/runs/31907721553

All right. It would be nice to filter out some of the noise by excluding 302, 307.

Inverle:
I don't see an option for that in https://lychee.cli.rs/guides/config/ or https://lychee.cli.rs/guides/cli/

Closest thing is `accept`, but that's the status code after the redirect.

lychee.toml
```diff
@@ -0,0 +1,39 @@
+# https://lychee.cli.rs/guides/config/
+
+verbose = "warn"
+no_progress = true
+cache = true
+max_cache_age = "1d"
+
+accept = "200..=204, 429"
```
Alkarex:
```suggestion
accept = "200, 302, 307, 429"
```

.lycheeignore
```diff
@@ -0,0 +1,54 @@
...
+https://archive\.ph/submit/\?url=~LINK~
+https://archive\.is/submit/\?url=~LINK~
+https://www\.reddit\.com/submit?url=~LINK~
+https://publish\.buffer\.com/compose?url=~LINK~&amp;text=~TITLE
+https://www\.journalduhacker\.net/stories/new?url=~LINK~&amp;title=~TITLE~
+https://www\.linkedin\.com/shareArticle?url=~LINK~&amp;title=~TITLE~&amp;source=FreshRSS
+https://www\.xing\.com/spi/shares/new?url=~LINK
+
+# URLs that will be published in the future
+https://github\.com/FreshRSS/FreshRSS/security/advisories/GHSA-p83h-r2jw-9whg
```
Alkarex:
```suggestion
^https://github\.com/FreshRSS/FreshRSS/security/advisories/.*
```

.lycheeignore
```diff
@@ -0,0 +1,54 @@
+# URLs used as examples
+https://codeload\.github\.com/FreshRSS/FreshRSS/zip/x\.y\.z
+^https://[^/]+\.domain/
```
Alkarex, line 3: Should there be a `^` for all lines?

lychee.toml
Alkarex: It could however be nice to be informed in case of e.g. `307 -> 404` but I am not sure whether this is supported

.lycheeignore
Inverle: It seems to work without

Alkarex: If it is indeed a regex, it would indeed work without but is less efficient because of lack of anchoring (more compute) and there might be false positives such as (not super convincing example) `http://example.net?ref=https://example.org`

Inverle: In that case we should indeed change it

lychee.toml
Alkarex: I have not tried whether that works. Tests welcome

Inverle: Though I would expect it to exclude those status codes from the summary, if added to `accept`.
But I haven't seen a way to make lychee fail on redirects.

.lycheeignore
Inverle: Done

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,64 @@
...
+          persist-credentials: false
+          fetch-depth: 0
+
+      - name: Get changed files
+        if: github.event_name == 'pull_request'
+        run: |
+          git diff --name-only --diff-filter=ACMR \
+            "${{ github.event.pull_request.base.sha }}" \
+            "${{ github.event.pull_request.head.sha }}" \
+            > changed-files.txt
```
Inverle, line 38: If there is a file renamed, which is referenced in some Markdown document, and the link is not updated, then currently this will only be caught after push.

lychee.toml
Inverle: > It could however be nice to be informed in case of e.g. `307 -> 404` but I am not sure whether this is supported

Yes, depending on whatever the final status code is, the request is either counted as successful or an error.

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,68 @@
+name: Link Checker
+
+on:
+  push:
+    branches: [ edge ]
+  pull_request:
+    branches: [ edge ]
+  schedule:
+    - cron: '30 12 1 * *'
```
Alkarex, line 9: What is the consequence of a cron job that found problems?

Frenzie: It depends at least in part on https://docs.github.com/en/subscriptions-and-notifications/how-tos/managing-github-actions-notifications

But actually that's a good point: are we mailed about a failed action we didn't initiate?

Inverle: It will show up as failed in CI, which you can configure notifications for.
In the summary of the run, you will be able to see what exactly failed.
There is also an option to extend the workflow to also create an issue, but it could spam the issue tracker too much.

.github/workflows/link-checker.yml
Alkarex: Might require `failIfEmpty: false` otherwise PRs without any link at all risk failing, aren't they?
https://github.com/lycheeverse/lychee-action/blob/e7477775783ea5526144ba13e8db5eec57747ce8/entrypoint.sh#L61-L63

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Run lychee
+        uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
+        with:
+          # PAT with read-only permissions used to avoid GitHub rate limits (allows 5,000 requests per hour)
+          # Won't be available to forks in pull requests due to restrictions, so falls back to default
+          # GITHUB_TOKEN with 1,000 requests per hour instead.
+          token: ${{ secrets.LYCHEE_GITHUB_TOKEN || secrets.GITHUB_TOKEN }}
+
+          fail: true
```
Alkarex, line 56: Isn't it too strict? I am afraid of a lot of noise

Inverle: Could be set conditionally, depending if it's a PR, cron or a push.
Yes, it could be a problem if unrelated failing links are scanned in a PR that happen to be in the set of changed files.
No option for changed lines only, unfortunately.

.github/workflows/link-checker.yml
Inverle: Possibly, was not aware of that option.

Inverle: Fixed in [`7307b3a`](https://github.com/FreshRSS/FreshRSS/pull/9209/commits/7307b3af10ec19dd801cbf313b81c262eb0ece55)

PR #9215, opened 2026-08-17
Disallow fetching of non remote uri

Inverle:
Fix https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-fgq3-88jp-7rj9

### DIFF
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -658,8 +658,9 @@ final class FreshRSS_http_Util {
 			if (in_array($c_status, [301, 302, 303, 307, 308], true)) {
 				// Handle the redirect by making another request
 				$location = \SimplePie\Misc::absolutize_url($headers['location'] ?? $url, $url);
-				if ($location === false) {
-					$location = $url;
+				if ($location === false || !\SimplePie\Misc::is_remote_uri($location)) {
+					Minz_Log::warning('Invalid redirect location: malformed URL “' . ($headers['location'] ?? $url) . '“');
+					break;
 				}
 				if (!self::compareURLOrigins($url, $location)) {
 					unset($curl_options[CURLOPT_COOKIE]);
--- a/app/Utils/httpUtil.php
+++ b/app/Utils/httpUtil.php
@@ -464,6 +464,11 @@ final class FreshRSS_http_Util {
 	 *   * `-500` `curl_init()` failure.
 	 */
 	public static function httpGet(string $url, ?string $cachePath = null, string $type = 'html', array $attributes = [], array $curl_options = []): array {
+		if (!\SimplePie\Misc::is_remote_uri($url)) {
+			Minz_Log::warning('Error fetching content: malformed URL “' . $url . '“');
+			return ['body' => '', 'effective_url' => '', 'redirect_count' => 0, 'fail' => true, 'status' => -500, 'error' => ''];
+		}
+
 		$limits = FreshRSS_Context::systemConf()->limits;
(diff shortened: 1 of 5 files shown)


lib/lib_rss.php
```diff
@@ -47,6 +47,25 @@ function actualize_mutex_file(string $tmpPath, string $dataPath): string {
...
+	$registered_wrappers = stream_get_wrappers();
+	$allowed_wrappers = ['file', 'php'];
+	foreach ($registered_wrappers as $protocol) {
+		if (!in_array($protocol, $allowed_wrappers, true)) {
+			stream_wrapper_unregister($protocol);
+		}
+	}
+}
+
+if (!defined('WITH_COMPOSER')) {
```
Inverle, line 65: For example PHPStan needs `phar://`

app/Utils/httpUtil.php
```diff
@@ -658,8 +658,9 @@ public static function httpGet(string $url, ?string $cachePath = null, string $t
 			if (in_array($c_status, [301, 302, 303, 307, 308], true)) {
 				// Handle the redirect by making another request
 				$location = \SimplePie\Misc::absolutize_url($headers['location'] ?? $url, $url);
-				if ($location === false) {
-					$location = $url;
+				if ($location === false || !\SimplePie\Misc::is_remote_uri($location)) {
```
Inverle, line 666: Should we add the same `is_remote_uri()` check at the beginning of this function?

https://github.com/FreshRSS/simplepie/blob/37ebf581e60ce90ebee7d79181de43cd8f63f936/src/Misc.php#L63-L66

lib/lib_rss.php
```diff
@@ -47,6 +47,25 @@ function actualize_mutex_file(string $tmpPath, string $dataPath): string {
...
 }
 
+/**
+ * Disable stream wrappers that the application does not need.
+ *
+ * For temporary usage of a previously unregistered wrapper, `stream_wrapper_restore()` can be used.
+ */
+function unregister_unsafe_protocols(): void {
+	$registered_wrappers = stream_get_wrappers();
+	$allowed_wrappers = ['file', 'php'];
```
Inverle, line 57: So it blocks cases such as `file_get_contents('https://example.com')` or `ftp://`, `zip://` etc.

Alkarex (approved):

app/Utils/httpUtil.php
Alkarex: Yes, that would make sense, if you cannot think of any drawback

Frenzie: In other circumstances I'd say I think it's a pity to needlessly block, say, FTP (even though the chance of it being used to host a feed is quite minuscule). But since we're always doing things with HTTP statuses I can't see how it could possibly make a difference.

Frenzie:
What happens after this PR if you submit a feed address that looks like `tweakers.net/feeds/mixed.xml`? Currently it automatically turns into `https://tweakers.net/feeds/mixed.xml`.

Inverle:
> What happens after this PR if you submit a feed address that looks like `tweakers.net/feeds/mixed.xml`? Currently it automatically turns into `https://tweakers.net/feeds/mixed.xml`.

It continues to work as before

The scheme is added by `checkUrl()` here:

https://github.com/FreshRSS/FreshRSS/blob/48ece4200f9eaa1990c791316bf0f69a8e53b019/app/Models/Feed.php#L499-L505

app/Utils/httpUtil.php
Inverle: Done

Alkarex:
PR welcome in https://github.com/FreshRSS/simplepie

Inverle:
https://github.com/FreshRSS/simplepie/pull/88

PR #9219, opened 2026-08-17
Dev updates: composer, hadolint, typos...

Alkarex:
Composer compatibility PHP 8.6.
Manual updates not supported by Dependabot

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,8 @@ test: bin/phpunit ## Run the PHP test suite
 
 bin/composer:
 	mkdir -p bin/
-	wget 'https://raw.githubusercontent.com/composer/getcomposer.org/a5abae68b349213793dca4a1afcaada0ad11143b/web/installer' -O - -q | php -- --quiet --install-dir='./bin/' --filename='composer'
+	wget 'https://raw.githubusercontent.com/composer/getcomposer.org/acf8a43b6a379f3a5ef72ebf21979f4936a608fd/web/installer' -O - | \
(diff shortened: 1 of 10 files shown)


lib/Minz/Url.php
```diff
@@ -61,7 +61,7 @@ public static function display(string|array $url = [], string $encoding = 'html'
 	}
 
 	/**
-	 * Construit l'URI d'une URL
+	 * Construct the URI of an URL
```
Frenzie: Though I'm not really sure what it means. Isn't any URL already a URI, even if more things are URIs too?
```suggestion
	 * Construct the URI of a URL
```

Makefile
```diff
@@ -75,7 +75,8 @@ test: bin/phpunit ## Run the PHP test suite
 
 bin/composer:
 	mkdir -p bin/
-	wget 'https://raw.githubusercontent.com/composer/getcomposer.org/a5abae68b349213793dca4a1afcaada0ad11143b/web/installer' -O - -q | php -- --quiet --install-dir='./bin/' --filename='composer'
+	wget 'https://raw.githubusercontent.com/composer/getcomposer.org/acf8a43b6a379f3a5ef72ebf21979f4936a608fd/web/installer' -O - | \
```
Inverle: I would do a `sha256sum -c` check here too

Inverle: I will make a PR to replace https://github.com/FreshRSS/Extensions/blob/main/.github/workflows/tests.yml#L106 with `sha256sum -c` later

Alkarex: Done

PR #9224, opened 2026-08-19
fix(favicon): fetch YouTube channel avatars

bowencool:
## Summary

- Resolve YouTube channel feed URLs to their channel pages when preparing favicons.
- Fetch the channel avatar from YouTube's `og:image` metadata before falling back to the existing favicon discovery logic.
- Preserve the existing behaviour for user-selected icons and non-YouTube feeds.

## Root cause

YouTube channel feeds expose a `channel_id` in the feed URL, but the feed URL itself does not expose the channel avatar. The existing favicon flow reduced that URL to the YouTube root favicon, so channel icons were not fetched.

## Validation

- `git diff --check`
- PHP/Composer lint and test commands were unavailable in the local environment because PHP and Composer are not installed.

image

### DIFF
--- a/app/Models/Feed.php
+++ b/app/Models/Feed.php
@@ -443,8 +443,9 @@ class FreshRSS_Feed extends Minz_Model {
 		$feedIconUrl = $this->attributeString('feedIconUrl') ?? '';
 		$websiteUrl = $this->website(fallback: false);
 		if ($websiteUrl === '' || $websiteUrl === $this->url) {
-			// Get root URL from the feed URL
-			$websiteUrl = preg_replace('%^(https?://[^/]+).*$%i', '$1/', $this->url) ?? $this->url;
+			// YouTube feed URLs expose the channel ID but no channel avatar.
+			$websiteUrl = youtubeChannelPageUrl($this->url)
+				?? (preg_replace('%^(https?://[^/]+).*$%i', '$1/', $this->url) ?? $this->url);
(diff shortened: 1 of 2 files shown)


bowencool:
`SimplePie::get_icon_url()` currently accepts an RSS `<channel><image><url>` only when it declares equal `width` and `height`. This is intentionally conservative, to avoid using a channel banner as a favicon.

That rule can omit an RSSHub channel image when the feed does not declare dimensions. In my experience, all RSSHub channel images I have seen are square, but we have not confirmed a stable RSSHub-specific marker that would safely distinguish those feeds.

A potential follow-up is to identify such a marker and accept RSSHub channel images without declared dimensions, while keeping the conservative rule for arbitrary RSS feeds.

Alkarex:
Seems to be related to:
* https://github.com/FreshRSS/Extensions/pull/513

See also:
* https://github.com/FreshRSS/Extensions/tree/main/xExtension-YouTube

In FreshRSS code, I would rather stick to only standard behaviour, avoiding site-specific code such as YouTube-specific rules, which are better placed in the [YouTube extension](https://github.com/FreshRSS/Extensions/tree/main/xExtension-YouTube).

Inverle:
The YouTube extension has this feature already

bowencool:
Ah, I’ve never used that extension, so I wasn’t aware of it.

PR #9227, opened 2026-08-21
Add drag&drop ordering for categories

### LINKED ISSUE #9204: [Feature] Add drag&drop ordering for categories
Currently categories can only be ordered by going into their settings, and setting a position number manually.
And it has to be done for all categories, since the ones that don't have a number set, are positioned after the ones that do, if you want the categories in an exact order.

See original discussion: https://github.com/FreshRSS/FreshRSS/pull/2592#issuecomment-546367257

and related in progress PR for feed drag&drop ordering: https://github.com/FreshRSS/FreshRSS/pull/8687

Sufiyan-MSA:
Closes #9204

Categories could only be ordered by opening each one and typing a
`Display position` number. Since categories without a position sort after
those with one, getting an exact order meant editing every category.

This adds drag&drop reordering of categories on the subscription
management page, reusing the drag&drop delegation already present there
for feeds.

- Category boxes on the subscription management page are now reordered by
  dragging their title bar; the drop target is highlighted while dragging.
- New `category&a=move` action stores the resulting order by assigning
  sequential `position` attributes to every category, so a single drag
  produces a fully explicit order instead of leaving some categories unset.
- Feed drag&drop is untouched: the category branch returns early in each
  handler, and the manual `Display position` field keeps working as before.

1. Create at least three categories, leaving their `Display position` empty.
2. Go to the subscription management page and drag a category by its title
   onto another category.
3. Reload the page, and check the sidebar: the new order is kept, and each
   category now has a `Display position` set in its settings.
4. Confirm dragging a feed from one category to another still works.


Inverle (approved):
Thanks for the PR. I made a few changes, let me know whether they're fine.

One thing to note: there is a small layout issue regarding the category boxes, especially in Firefox - but that can be addressed in a different PR, since the issue seems to exist partly because of some browser bug. (it's difficult to explain though)

Inverle:
Not sure what to edit the documentation with, if you have an idea, you can modify it.

https://github.com/FreshRSS/FreshRSS/blob/3d59aeb476849662ef02687bdd6a4b0de8dac244/docs/en/users/04_Subscriptions.md?plain=1#L20

Sufiyan-MSA:
Thanks for the changes — they all look good to me, and several are
straight improvements over what I had:

- Simulating the reorder on a clone and only applying it in `onload` is
  more correct than my move-then-reload approach; a failed request now
  leaves the page untouched.
- The `draggingCategory` flag correctly keeps the drag state alive past
  `dragend` until the request resolves, which my version got wrong.
- `badAjax()` is clearly the right pattern over `location.reload()`.
- 0-based positions are safe: `aside_feed.phtml` guards with
  `null === $position`, so `0` still renders, and `data-position` isn't
  read by any JS or CSS.

Two very minor things, take or leave:

- In `save_category_order`, `clone.querySelector('[data-cat-id="…"]')`
  matches both the `.box` div and its inner `<ul>` — it only resolves to
  the box because the div comes first in document order. Reusing the
  existing `catBox` constant (`.box[data-cat-id="…"]`) would make that
  explicit.
- `req.onerror` calls `cleanDragElements()` but not `badAjax()`, so a
  network-level failure reverts silently, unlike an HTTP error status.

I've updated the documentation and merged `edge` in. Since the numeric
field is gone, I removed the **Display position** bullet and folded the
behaviour into the "Subscription management" paragraph next to the
existing feed drag-and-drop sentence, keeping the one part of the old
text that still applies (never-reordered categories sort last,
alphabetically).

Unrelated follow-up for whenever you touch the layout issue:
`data-cat-position` in `subscription/index.phtml` is dead code. It calls
`attributeString('position')` on an int attribute, so it always renders
empty and the badge rule in `frss.css` never matches. It was already
broken before this PR — now that the numeric field is gone, the attribute
and its CSS rule could just be deleted.

One question: merging is blocked on "Commits must have verified
signatures". My commits are unsigned. Would a squash-merge work on your
side, or should I set up a signing key and force-push a re-signed branch?

Inverle:
I took your suggestions and implemented them. I left the `data-cat-position` attribute since it potentially has use for User JS/CSS or extensions.

> One question: merging is blocked on "Commits must have verified
signatures". My commits are unsigned. Would a squash-merge work on your
side, or should I set up a signing key and force-push a re-signed branch?

It would be nice to have signed commits, but a squash merge will do as well.

Sufiyan-MSA:
Perfect, thanks for implementing those — and good call keeping
`data-cat-position` for user JS/CSS and extensions.

Let's go with the squash merge. Re-signing would mean force-pushing a
rebase over your commits too, which doesn't seem worth it here. I'll set
up a signing key for future contributions.

Nothing further from me, happy for this to go in whenever you're ready.

Issue #1610, opened 2017-08-15
Postgresql: commitNewEntries() fail due to conflict on duplicate keys while inserting


Trim:
Hello,

I'm using FreshRSS 1.7.0 (zip release download) with PostgreSQL 9.6.4-0+deb9u1 on Debian Stretch.

It appears, that after a while, FreshRSS don't import any more articles from RSS feeds.

I've found the issue: there's an SQL Error inside the PostgreSQL logs (sorry, they are French « speaking »):

```
2017-08-15 06:36:30 CEST [23983-600] freshrss@freshrss ERREUR:  la valeur d'une clé dupliquée rompt la contrainte unique « adrien_entry_id_feed_guid_key »
2017-08-15 06:36:30 CEST [23983-601] freshrss@freshrss DÉTAIL:  La clé « (id_feed, guid)=(9, http://www.minimachines.net/?p=53460) » existe déjà.
2017-08-15 06:36:30 CEST [23983-602] freshrss@freshrss CONTEXTE :  instruction SQL « INSERT INTO "adrien_entry" (id, guid, title, author, content, link, date, "lastSeen", hash, is_read, is_favorite, id_feed, tag
s)
                        (SELECT rank + row_number() OVER(ORDER BY date) AS id, guid, title, author, content, link, date, "lastSeen", hash, is_read, is_favorite, id_feed, tags FROM "adrien_entrytmp" ORDER BY date
) »
        fonction PL/pgsql inline_code_block, ligne 6 à instruction SQL
2017-08-15 06:36:30 CEST [23983-603] freshrss@freshrss INSTRUCTION :  DO $$
        DECLARE
        maxrank bigint := (SELECT MAX(id) FROM "adrien_entrytmp");
        rank bigint := (SELECT maxrank - COUNT(*) FROM "adrien_entrytmp");
        BEGIN
                INSERT INTO "adrien_entry" (id, guid, title, author, content, link, date, "lastSeen", hash, is_read, is_favorite, id_feed, tags)
                        (SELECT rank + row_number() OVER(ORDER BY date) AS id, guid, title, author, content, link, date, "lastSeen", hash, is_read, is_favorite, id_feed, tags FROM "adrien_entrytmp" ORDER BY date
);
                DELETE FROM "adrien_entrytmp" WHERE id <= maxrank;
        END $$;
[...]

PR #9231, opened 2026-08-23
PostgreSQL: make commitNewEntries resilient to duplicate keys

aligundogar:
## Problem

On PostgreSQL, `commitNewEntries()` can fail permanently, leaving **all new articles stuck in `_entrytmp`** while the UI/API shows nothing newer than the day the failure started — even though feed fetching itself keeps working.

We hit this in production (FreshRSS 1.26.0, PostgreSQL 16, ~3200 feeds, `nb_parallel_refresh = 10`, WebSub enabled):

```
ERROR:  duplicate key value violates unique constraint "…_entry_id_feed_guid_key"
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

Within ~8 days the temporary table grew to **213,000+ articles** that never reached `_entry`:

| Table | Rows |
|---|---|
| `_entrytmp` (stuck backlog) | 211,587 → 213,247 |
| newest article visible in UI/API | frozen at the day of first failure |

## Root cause

The `DO $$ … $$` block in `commitNewEntries()` uses a plain `INSERT INTO _entry … SELECT … FROM _entrytmp WHERE NOT EXISTS (…)`, so a single duplicate `(id_feed, guid)` — or duplicate generated `id` — aborts the *entire transaction*, and the trailing `DELETE FROM _entrytmp WHERE id <= maxrank` never executes. Every subsequent refresh retries the whole ever-growing batch under the same conditions.

Two realistic sources of duplicates:

1. **Concurrent commits**: multiple parallel refresh workers and WebSub (`pshb.php`) handlers can call `commitNewEntries()` at the same time on the shared `_entrytmp`. Under `READ COMMITTED`, two transactions evaluate `NOT EXISTS` against the same snapshot; the loser then hits the unique index after the winner commits.
2. **Feeds with non-unique GUIDs** within one update batch (as documented in #1614).

MySQL and SQLite adapters already tolerate this via `INSERT IGNORE` / `INSERT OR IGNORE`; only PostgreSQL hard-fails. The related TODO ("Update to PostgreSQL 9.5+ syntax with ON CONFLICT DO NOTHING") has been in this file for years.

## Fix

- Deduplicate each batch with `SELECT DISTINCT ON (id_feed, guid) … ORDER BY date DESC` — keeps the newest version of a given GUID, matching the intent discussed in #1614 ("keep the newest article when there is no unique GUID").
- Add `ON CONFLICT DO NOTHING` so losing races and already-known GUIDs become harmless no-ops instead of aborted transactions. This also covers primary-key collisions between concurrent committers.
- ID generation semantics are unchanged: `(maxrank - nb) + row_number() OVER (ORDER BY date, id)` still produces IDs ≤ `MAX(id) OF _entrytmp`, ordered by publication date.

## Validation

- `php -l` clean.
- Executed the exact new statement against the live production database inside a transaction followed by `ROLLBACK` (dry run): moved **213,247** articles from `_entrytmp` into `_entry`, left **0** rows behind, no constraint violations, completed in seconds.

Fixes scenarios reported in #1610 and #1614.

### DIFF
--- a/app/Models/EntryDAOPGSQL.php
+++ b/app/Models/EntryDAOPGSQL.php
@@ -88,14 +88,14 @@ class FreshRSS_EntryDAOPGSQL extends FreshRSS_EntryDAOSQLite {
 			DO $$
 			DECLARE
 			maxrank bigint := (SELECT MAX(id) FROM `_entrytmp`);
-			nb bigint := (SELECT COUNT(*) FROM `_entrytmp`);
+			rank bigint := (SELECT maxrank - COUNT(*) FROM `_entrytmp`);
 			BEGIN
 				INSERT INTO `_entry`
 					(id, guid, title, author, content, link, date, `lastSeen`, hash, is_read, is_favorite, id_feed, tags, attributes)
-					SELECT (maxrank - nb) + row_number() OVER(ORDER BY etmp.date, etmp.id) AS id, guid, title, author, content,
+					(SELECT rank + row_number() OVER(ORDER BY etmp.date, etmp.id) AS id, guid, title, author, content,
 						link, date, `lastSeen`, hash, is_read, is_favorite, id_feed, tags, attributes
-					FROM `_entrytmp` etmp
-					ORDER BY etmp.date, etmp.id
+					FROM `_entrytmp` AS etmp
+					ORDER BY etmp.date, etmp.id)
 					ON CONFLICT DO NOTHING;
 				DELETE FROM `_entrytmp` WHERE id <= maxrank;
 			END $$;
--- a/app/Models/EntryDAOPGSQL.php
+++ b/app/Models/EntryDAOPGSQL.php
@@ -84,22 +84,19 @@ class FreshRSS_EntryDAOPGSQL extends FreshRSS_EntryDAOSQLite {
 
 	#[\Override]
 	public function commitNewEntries(): bool {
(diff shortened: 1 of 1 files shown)


Alkarex:
I have not looked into the code in details yet, but now that we require PostgreSQL 10+, supporting e.g. `CONFLICT DO NOTHING`, it should be possible to reduce the code a lot, to something similar to our MySQL version

aligundogar:
Thanks @Alkarex, agreed — pushed e61298195 simplifying to the same shape as the MySQL version:

- Dropped the `DISTINCT ON` pre-dedup entirely. Intra-batch duplicate GUIDs (and losing races) are now simply skipped by `ON CONFLICT DO NOTHING`, matching the `INSERT IGNORE` / `INSERT OR IGNORE` semantics of the other adapters, where the oldest row wins too.
- One deliberate difference from a literal two-statement translation: the `DO $$ … $$` block stays, because `maxrank` must be computed once up-front. Otherwise a row concurrently pushed into `_entrytmp` by another refresh worker or WebSub handler between the `INSERT` and the trailing `DELETE` would be deleted without ever having been committed into `_entry`.

Validation: `php -l` clean; executed this exact statement against a live production database inside `BEGIN … ROLLBACK` (dry run) — no errors, no constraint violations. The previous variant was additionally validated against a 213k-row stuck backlog the same way.

References #1610 #1614

Alkarex:
I have reduced a bit the diff. Double-check welcome

Alkarex (approved):

PR #9239, opened 2026-08-28
Fix spam of empty log messages

### LINKED ISSUE #9222: [Bug] Prevent spam of empty log messages
For example here:

Image

ref: https://github.com/FreshRSS/FreshRSS/pull/8400#issuecomment-4179851083, https://github.com/FreshRSS/FreshRSS/pull/8400#issuecomment-4344292895

https://github.com/FreshRSS/FreshRSS/blob/0376d402b033d14a34661b968b29d165bacc8fbc/app/Models/Entry.php#L1100-L1103

(need to investigate why the catch block is called, and why the message is empty)

Also preferably print some debug log message, whenever any `Minz_Log` function is called with an empty string. Maybe point to the source with `debug_backtrace()`, or require `non-empty-string` type, instead of `string`.

mmustafasenoglu:
Fix spam of empty log messages

When fetching full article content fails with an exception that has
an empty message, the catch block in Entry::loadCompleteContent()
logs an empty warning message, causing log spam.

Fix:
1. In Entry.php: Only log the exception message if it's not empty
2. In Minz_Log.php: Add early return for empty/whitespace messages
   to prevent any empty log entries at the framework level

Fixes #9222

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -1104,7 +1104,10 @@ class FreshRSS_Entry extends Minz_Model {
 					}
 				} catch (Exception $e) {
 					// rien à faire, on garde l’ancien contenu(requête a échoué)
-					Minz_Log::warning($e->getMessage());
+					$message = $e->getMessage();
+					if ($message !== '') {
+						Minz_Log::warning($message);
+					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


Alkarex:
Thanks! 
It would be interesting to investigate the cause, for instance by reproducing the issue.
Because if we only hide the symptoms, we risk missing the more important causes

PR #9243, opened 2026-08-29
Improve shared HTML

### LINKED ISSUE #9232: Is it possible to change the view from 'Reader' to 'Normal' in shared user query?
I like the query-share feature but in HTML mode it shares only in 'Reader' mode which is not very information-dense, 'Norma' view with openable items is much more suitable for me, is it possible to share in that mode?

Alkarex:
fix https://github.com/FreshRSS/FreshRSS/issues/9232
Allow changing between normal view and reader view in shared user queries.
Allow visitors to change view and change sorting order.
Enable some JavaScript for e.g. keyboard shortcuts.

Reduce code by sharing better between normal view, reader view, and user query HTML.

image

image


mmustafasenoglu:
#9247 has been closed. The entry rendering template refactoring from that PR can be incorporated into this Work In Progress PR if desired.

Alkarex:
@mmustafasenoglu Help welcome here, in particular for testing to avoid regressions

PR #9246, opened 2026-08-30
Strip feed URL credentials from anonymous query.php output

SamCyber001:
The public query-sharing endpoint `p/api/query.php` emits `origin.feedUrl` with the
full stored feed URL — including any `user:pass@` credentials — for `f=json` and
`f=greader`. An anonymous recipient of a "Share by RSS" link can therefore read the
credentials of the owner's private feeds.

This strips the `user:pass@` credentials from `feedUrl` on the public serialization
path (`mode === 'freshrss'`), matching what the logging and OPML export paths already do.

Note: `SimplePie\Misc::url_remove_credentials()` only removes `user:pass@` userinfo, not
secrets passed as query parameters (e.g. `?apikey=…`), which would still appear in
`feedUrl`. Since `mode === 'freshrss'` is used only by this anonymous share export (the
authenticated GReader API uses `compat`, which omits `feedUrl` entirely), if you'd like
to close that case too, `feedUrl` could instead be dropped on this path. Happy to adjust.

Reported privately in GHSA-rpmm-h4hx-5p6p.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -1273,7 +1273,7 @@ class FreshRSS_Entry extends Minz_Model {
 			if ($mode === 'compat') {
 				$item['origin']['title'] = escapeToUnicodeAlternative($feed->name(), true);
 			} elseif ($mode === 'freshrss') {
-				$item['origin']['feedUrl'] = htmlspecialchars_decode($feed->url());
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
lgtm

PR #9248, opened 2026-08-30
Gate greader/json share formats behind Share by OPML

SamCyber001:
Follow-up to #9246, per @Alkarex's suggestion on GHSA-rpmm-h4hx-5p6p.

The greader and json share formats expose origin.feedUrl (the feed's subscription URL). Sharing feed URLs is what the "Share by OPML" option explicitly means, whereas "Share by RSS" is meant to share article content. This moves greader and json in p/api/query.php from the shareRss gate to the shareOpml gate, so feed URLs are only served when the owner opted into OPML sharing.

Note: moved json alongside greader too, since both emit feedUrl (same toGReader path).

### DIFF
--- a/app/Models/UserQuery.php
+++ b/app/Models/UserQuery.php
@@ -347,7 +347,7 @@ class FreshRSS_UserQuery {
 	}
 
 	public function sharedUrlGreader(bool $xmlEscaped = true): string {
-		if ($this->shareRss && $this->token !== '') {
+		if ($this->shareOpml && $this->token !== '') {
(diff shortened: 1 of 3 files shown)


Alkarex:
Thanks! We also need to move the corresponding link in the UI from being under HTML to being under OPML

SamCyber001:
Done — moved the greader link under the "Share by OPML" group and switched
sharedUrlGreader() to gate on shareOpml so the link matches the endpoint.

Alkarex:
Perfect. Please add a line for you in https://github.com/FreshRSS/FreshRSS/blob/edge/CREDITS.md

PR #9249, opened 2026-08-30
Add remove action to feed dropdown menu

monksy:
## Summary

- add a Remove action to each feed configuration dropdown
- reuse the existing feed deletion endpoint, CSRF-protected form, confirmation message, and translation
- preserve the stats-page redirect when deleting from feed statistics

## Testing

- `php -l app/layout/aside_feed.phtml` (PHP 8.1 container)
- `git diff --check`


Alkarex:
Hello and thanks for the PR.
I am not sure I like it (by default). I would not want this menu to contain too many entries, in particular not risky ones, and it only saves one click. Is it something important for you?
However, it would be nice to expose this menu to extensions (help welcome for that).

monksy:
So right now the issue is if you're running into a bad feed, you have to go all the way to subscription manager, find the feed, click on the feed details and then find the remove button. This simplifies that flow.

Alkarex:
> So right now the issue is if you're running into a bad feed, you have to go all the way to subscription manager, find the feed, click on the feed details and then find the remove button. This simplifies that flow.

There is a much more direct method already: click on *Manage*  in the same context menu, and then *Delete*

PR #9252, opened 2026-09-01
Add the Atelier theme

mbieh:
Adds **Atelier**, the theme from the show-and-tell thread in #9217. @Alkarex, you wrote there that a PR is welcome once it feels ready, and asked to limit duplicated code by building on a base theme with overrides — that is how it is built. @nclm, thank you for the kind words in that thread.

In the discussion I proposed submitting a single palette upstream rather than all nine; this is that one, the Tailwind Neutral ramp, and it ships under the plain name `Atelier`. The other eight stay in [mbieh/Atelier](https://github.com/mbieh/Atelier) for anyone who wants a different grey. Happy to change which one lands here — the folder is generated, so swapping the ramp is a one-line change on my side.

## What it is

A calm light and dark theme derived from Mapco, inspired by [shadcn/ui](https://ui.shadcn.com/). No web fonts, no scripts, nothing loaded from an external service while a page is read.

* **No duplicated base.** `metadata.json` lists `_frss.css`, so the base theme is inherited, not copied. The folder holds only what is Atelier's own.
* **Colors are roles, not literals.** `_variables.css` maps shadcn/ui's semantic roles onto an eleven-step neutral ramp in `_palette.css`, and each role picks its step by the contrast it needs. Redefining the ramp with the CustomCSS extension moves the whole interface, which is also what makes the other eight palettes the same theme rather than eight forks of it.
* **Dark mode asks first.** Every rule behind `prefers-color-scheme: dark` is gated on `darkMode_auto`, matched inside `:where()` so the gate lends it no specificity. A reader who sets *Automatic dark mode* to No keeps a light page.
* **Icons.** Bundled Lucide SVGs in `icons/`. Most are painted as CSS masks rather than recolored with filters, so a glyph follows `currentColor` through hover, active and dark states. This is the one deliberate departure from the theme checklist, which asks for `#666666` plus filters — happy to revisit if you would rather have the convention than the behaviour.
* **Right to left.** Direction is answered inside the sheet: a declaration that reads differently in Arabic or Hebrew carries a `:dir(rtl)` rule beside it, so the mirrors are copies. Each sheet is wrapped in rtlcss's `ignore` directives, because an unguided pass would flip both halves of every pair and leave a right-to-left reader with a sheet that is wrong twice over. `npm run rtlcss` is idempotent on this folder.
* **Accessibility.** Every semantic role is held against the surfaces it is painted on: 756 pairings across the palettes, in both schemes, all clearing their WCAG bars. Visible focus states throughout, and `prefers-reduced-motion` is honoured.

## Checklist

| Item | State |
| --- | --- |
| `metadata.json`, `thumbs/original.png` | ✔ |
| `_frss.css` in `files` | ✔ |
| Icons as SVG under `icons/` | ✔ — Lucide, license header kept for attribution |
| Icons `#666666` + filters | ✘ — masks with `currentColor` instead, see above |
| CSS variables for colors | ✔ — roles over an eleven-step ramp |
| Dark mode aligned with the `darkMode` class | ✔ |
| No preprocessor | ✔ — plain CSS, partials via `@import` |
| `npm test`, `typos` | ✔ |

## Notes for review

* **The `?v=` query on the `@import`s.** `p/themes/.htaccess` caches `text/css` for a month, and FreshRSS cache-busts only the sheets named in `metadata.json`. A partial reached through `@import` would otherwise be served stale for weeks after a theme change, so the imports carry the theme's own version. It is written by the generator alongside the `version` field, so the two cannot drift. Glad to drop it if you would rather solve that centrally.
* **Browser baseline.** Nesting, `:has()`, subgrid, `color-mix()` and `:dir()` — Chrome/Edge 120+, Firefox 121+, Safari 17.2+. The six `-webkit-backdrop-filter` declarations carry a stylelint exemption each, since Safari kept that prefix until 18.
* **Maintenance.** The folder is generated from the source at [mbieh/Atelier](https://github.com/mbieh/Atelier), which lints with a verbatim copy of this repository's `.stylelintrc.json` and runs a WCAG contrast check over every palette. Fixes made here are easy for me to carry back; I am happy to maintain the theme, and interested in the variable-based theme selection you mentioned in the thread.

Marked as a draft: the palette question above is open, and I would rather settle it before this is merged than after.


Alkarex:
Any comment maybe @math-GH ?

PR #9256, opened 2026-09-01
Trim the Docker Hub description at a section boundary

### LINKED ISSUE #9211: [Bug] Docker Hub README description gets truncated

m-altaifi:
Fixes #9211.

`Docker/README.md` is currently ~27,900 characters, and Docker Hub caps the full description at [25,000 characters](https://github.com/peter-evans/dockerhub-description#content-limits). The published description is therefore cut mid-section — the tail of `## Cron job to automatically refresh feeds` just disappears, which is what makes the page [look incomplete](https://hub.docker.com/r/freshrss/freshrss#option-3-cron-as-another-instance-of-the-same-freshrss-docker-image).

This trims the description ourselves before handing it to `dockerhub-description`, so the cut lands on a section boundary instead of mid-sentence, and appends a link to the full README on GitHub.

@Alkarex suggested splitting the readme manually. I went with computing the cut instead so there is still a single source of truth and no second copy to keep in sync — and, more importantly, so this cannot silently regress the same way if the README grows again. Sections are dropped from the end until the text fits; nothing else about `Docker/README.md` changes.

### Result

Running the new step against the current `Docker/README.md`:

```
Docker/README.md      27,910 characters
README.dockerhub.md   23,541 characters
```

The generated description ends cleanly at the end of `## Alternative reverse proxy configurations`, followed by:

```markdown
---

📖 **[Read the full documentation on GitHub](https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/README.md)**
```

### Testing

Parsed the workflow with PyYAML, pulled the `run:` script back out of the parsed document, and executed it against the real `Docker/README.md` to confirm both the character count above and that the output ends on a complete section.

### DIFF
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -18,10 +18,25 @@ jobs:
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+    - name: Trim the description to Docker Hub's character limit
+      run: |
+        python3 - <<'EOF'
+        import pathlib
+        # Docker Hub truncates the full description at 25,000 characters, which cuts
+        # Docker/README.md off mid-section. Drop whole trailing sections instead, and
+        # link to the complete version. https://github.com/peter-evans/dockerhub-description#content-limits
+        LIMIT = 25000
+        MORE = '\n\n---\n\n📖 **[Read the full documentation on GitHub](https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/README.md)**\n'
+        readme = pathlib.Path('Docker/README.md').read_text(encoding='utf-8')
+        if len(readme) + len(MORE) > LIMIT:
+            readme = readme[:LIMIT - len(MORE)].rpartition('\n## ')[0]
+        pathlib.Path('README.dockerhub.md').write_text(readme + MORE, encoding='utf-8')
+        EOF
+
     - name: Update repo description
       uses: peter-evans/dockerhub-description@1b9a80c056b620d92cedb9d9b5a223409c68ddfa
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
         repository: freshrss/freshrss
-        readme-filepath: Docker/README.md
+        readme-filepath: README.dockerhub.md
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -18,18 +18,20 @@ jobs:
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-    - name: Trim the description to Docker Hub's character limit
+    - name: Trim the description to Docker Hub's byte limit
       run: |
         python3 - <<'EOF'
         import pathlib
-        # Docker Hub truncates the full description at 25,000 characters, which cuts
+        # Docker Hub truncates the full description at 25,000 bytes, which cuts
         # Docker/README.md off mid-section. Drop whole trailing sections instead, and
         # link to the complete version. https://github.com/peter-evans/dockerhub-description#content-limits
-        LIMIT = 25000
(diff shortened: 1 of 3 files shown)


.github/workflows/dockerhub-description.yml
```diff
@@ -18,10 +18,25 @@ jobs:
...
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+    - name: Trim the description to Docker Hub's character limit
+      run: |
+        python3 - <<'EOF'
+        import pathlib
+        # Docker Hub truncates the full description at 25,000 characters, which cuts
+        # Docker/README.md off mid-section. Drop whole trailing sections instead, and
+        # link to the complete version. https://github.com/peter-evans/dockerhub-description#content-limits
+        LIMIT = 25000
```
Frenzie: Mind that it says "The readme content is limited to 25,000 _bytes_".

A quick search suggests `s.encode('utf-8')[:10].decode('utf-8', 'ignore')`.
<https://stackoverflow.com/a/75608686>

m-altaifi: You're right, thanks — fixed in 56a7376. It now slices `readme.encode('utf-8')` against the byte budget and uses `decode('utf-8', 'ignore')` to drop the partial character the slice may land on, before cutting back to the last complete `## ` section.

Re-running the step against the current `Docker/README.md`:

```
source  27,967 bytes / 27,910 characters
output  24,192 bytes / 24,146 characters
```

Measuring bytes properly also keeps about 650 bytes more of the README than the character-based version did.

Frenzie: ...huh? I don't think that should be possible. A character is 1 to 4 bytes.

m-altaifi: Sorry, my numbers were wrong.

Both versions produce identical output. Measured properly this time:

```
source       27,967 bytes  27,910 chars
char budget  23,587 bytes  23,541 chars
byte budget  23,587 bytes  23,541 chars
```

What happened: I measured the byte-based run by reading back the file the script writes, on Windows, where Python's `write_text` turns every `\n` into `\r\n`. The output has 605 newlines and 23,587 + 605 = 24,192, which is the number I posted. Line-ending artifact from my machine, not anything the workflow does — the runner is Linux.

They come out the same here because the cut gets snapped back to the last complete `## ` section either way, and the whole file only carries 57 bytes of multi-byte overhead, so the boundary doesn't move. Measuring bytes is still the right thing to do since that's the actual limit, it just doesn't buy any extra content.

Frenzie (approved):
Should be fine. Consider forcing `\n`?

Alkarex:
While this approach is fine, I would still rather go for a manual action:
* Make https://github.com/FreshRSS/FreshRSS/blob/edge/Docker/README.md shorter
* Link to additional elements moved under https://github.com/FreshRSS/FreshRSS/tree/edge/docs
* Add a minimal size limit check in https://github.com/FreshRSS/FreshRSS/blob/edge/.github/workflows/tests.yml

I am not sure cutting before a random section is much better than cutting elsewhere, and to the contrary, it shows more clearly that there is something missing

Frenzie:
Yes, a simple warning for size so you can act on it makes sense.

m-altaifi:
Updated in 6a444f6 to follow the requested direction: Docker/README.md is now the published description (23,920 UTF-8 bytes), the long cron and database-migration material links to its existing docs pages, and the standard test workflow rejects a README over Docker Hub's 25,000-byte limit.

Docker/README.md
```diff
@@ -599,120 +599,12 @@ server {
...
 
-## Cron job to automatically refresh feeds
+## Further administration documentation
 
-We recommend a refresh rate of about twice per hour (see *WebSub* / *PubSubHubbub* for real-time updates).
-There are no less than 3 options. Pick a single one.
+To keep this Docker Hub guide below its description-size limit, the detailed
+operational guides live in the documentation site:
 
-### Option 1) Cron inside the FreshRSS Docker image
```
Alkarex, line 607: We need to move and keep a good part of this deleted information in the docs

PR #9264, opened 2026-09-06
Fix marking filtered label articles as read in SQLite and PostgreSQL

### LINKED ISSUE #9214: [Bug] Mark as Read from search filter doesn't
### Describe the bug

In FreshRSS 1.29.1, the "mark all as read" buttons from the search results performed in label view, advances labels without marking as read.

### To Reproduce

1. Have more than one label populated with distinct items
2. Pick the first label in the side bar to limit to only those articles
3. Take note of the article count on the label
4. Pick a semi-unique word in the first couple articles
5. Use the search from the top bar to search for that semi-unique term
6. See there is a smaller set of articles shown that match the search term
7. Click the "Mark all as read" button in the top bar
8. See the view advance to the next label
9. Return to the previous label
10. Notice the article count us unchanged from before
11. Notice the article you picked the semi-unique search term from is still unread

If your semi-unique search term results in less than one page of results, you can click the "Mark all as read" below the last result forthe same (lack of) effect.

If you have "mark as read on scroll" enabled however, scrolling thru the search results does mark them as read properly.

### Expected behavior

"Mark all as read" marls all visible articles as read.

### FreshRSS version

1.29.1

### System information

- Database version: SQLite 3.51.2
- PHP version: 8.4.21
- Installation type: Docker (alpine)
- Web server type: Apache
- Device: Google Pixel Fold
- OS: Android 17
- Browser: Water fox 1.2.6 (Firefox 153.0)

### Additional context

_No response_

XiaoSong1223:
Fixes #9214.

Marking search results as read within a label fails on SQLite and PostgreSQL: `markReadTag()` generates conditions such as `e.title`, but its UPDATE statement does not define the `e` alias. The UI advances to the next label while the matching articles remain unread. PostgreSQL inherits this implementation from `FreshRSS_EntryDAOSQLite`.

Use an empty alias for these conditions. The production fix remains one line. MySQL/MariaDB use a different query which defines `e` correctly and need no change.

`EntryDAOTest` runs the same eight regression cases against SQLite by default, or PostgreSQL/MySQL/MariaDB through `FRESHRSS_TEST_DSN`. Each case uses the real database-specific DAO and connection-local temporary tables, including compressed content on MySQL/MariaDB. Coverage includes title/content searches, individual and all labels, read/favourite state filters, the maximum article ID, unread counts, repeated marking, and marking matching read articles unread. Connection settings and commands are documented in `tests/README.md`.

Validation with PHP 8.5.10, using the same tests against the parent of the original fix and the updated code:

| Database | Before the fix | After the fix |
| --- | --- | --- |
| SQLite 3.53.4 | 7 filtered cases fail with undefined-column errors | 8/8 pass |
| PostgreSQL 16.14 | 7 filtered cases fail with missing FROM-clause entry for `e` | 8/8 pass |
| MySQL 8.4.11 | 8/8 pass | 8/8 pass |
| MariaDB 11.4.13 | 8/8 pass | 8/8 pass |

Each passing database run completes 264 assertions. SQLite remains the default CI backend; other backends were tested locally against isolated database servers.

Also verified:
- Both the top and bottom mark-as-read buttons in the local SQLite web instance, with non-matching articles and other labels unchanged.
- Full `composer test`: 703 PHPUnit tests, 1548 assertions, PHP/PHTML syntax checks, PHPCS, and PHPStan passed. Run from a temporary copy without spaces in its path because existing CLI tests do not quote shell paths.
- Markdown lint and `git diff --check` passed.

This change was prepared with AI assistance.


Alkarex:
Thanks!
Did you have a chance to test the other databases, by any chance?
This code is also used by PostgreSQL. And we should also check MySQL/MariaDB even if it is using another code, as it might have the same bug.

Alkarex (approved):

Alkarex:
Thanks!

PR #9272, opened 2026-09-08
Only allow http/https URI schemes for enclosure and thumbnail URLs

aligundogar:
Enclosure URLs (`<enclosure url="…">`, media RSS thumbnails, and the entry-level `media:thumbnail`) are stored as entry attributes and interpolated into `href`/`src` by `FreshRSS_Entry::content()`. Unlike URLs inside content HTML, these metadata values do not go through SimplePie’s URI scheme sanitization (`Sanitize::replace_urls()`), so any URI scheme can end up in user-facing links.

This PR adds `FreshRSS_http_Util::isAllowedUrlScheme()` — an allow-list of `http`/`https` (consistent with the protocols used for fetching feeds). Before extracting the scheme it strips control characters and whitespace, mirroring how browsers parse URI schemes, so obfuscated values like `java\tscript:` are not accepted either.

The check is applied:

- **at store time** in `app/Models/Feed.php`, while building the entry attributes (enclosure URL, per-enclosure thumbnails, entry-level `media:thumbnail`);
- **at render time** in `FreshRSS_Entry::content()`, so entries already stored in existing databases are covered as well, without a data migration. Since the Google Reader compatible API serves this same output, third-party clients are covered too.

Enclosure URLs using other schemes (`data:`, `ftp:`, relative paths, …) are now dropped. If relative enclosure URLs should keep working, I can switch the helper to a block-list mirroring SimplePie’s `disallowed_uri_schemes` — feedback welcome.

PHPUnit coverage added in `tests/app/Models/EntryTest.php` (scheme matrix + rendering of already-stored entries).

Security fix coordinated privately with the maintainers; opened here per maintainer request so CI can run.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -227,7 +227,7 @@ class FreshRSS_Entry extends Minz_Model {
 		$thumbnailAttribute = $this->attributeArray('thumbnail') ?? [];
 		if (!empty($thumbnailAttribute['url'])) {
 			$elink = $thumbnailAttribute['url'];
-			if (is_string($elink) && FreshRSS_http_Util::isAllowedUrlScheme($elink) &&
+			if (is_string($elink) && \SimplePie\Misc::is_remote_uri($elink) &&
 				($allowDuplicateEnclosures || !self::containsLink($content, $elink))) {
 				$content .= <<<HTML
 					<figure class="enclosure">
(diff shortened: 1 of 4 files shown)


Alkarex:
Thanks! I have made some changes. Checks welcome

Frenzie:
Don't forget to allow protocol-relative URLs, even if they're no longer as in vogue as they once were.

Alkarex:
> Don't forget to allow protocol-relative URLs, even if they're no longer as in vogue as they once were.

I have added some coverage for that in the tests

PR #9273, opened 2026-09-08
SimplePie: Harden cache deserialization calls

Inverle:
To reduce impact if a maliciously provided cache file is somehow loaded, e.g. from a manipulated backup or some kind of external arbitrary file write vulnerability.

### DIFF
--- a/lib/simplepie/simplepie/src/Cache/File.php
+++ b/lib/simplepie/simplepie/src/Cache/File.php
@@ -85,7 +85,10 @@ class File implements Base
     public function load()
     {
         if (file_exists($this->name) && is_readable($this->name)) {
-            return unserialize((string) file_get_contents($this->name));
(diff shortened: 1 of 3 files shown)


Inverle:
https://github.com/FreshRSS/simplepie/pull/90

lib/simplepie/simplepie/src/Cache/File.php
Alkarex, line 1: Please make a PR for that in https://github.com/FreshRSS/simplepie

Alkarex (approved):

lib/simplepie/simplepie/src/Cache/File.php
Inverle: It's here https://github.com/FreshRSS/simplepie/pull/90

Alkarex: Ah, excellent https://github.com/FreshRSS/simplepie/pull/90

PR #9276, opened 2026-09-09
Ignore security-relevant OPML attributes for dynamic OPML sources

aligundogar:
A dynamic OPML category re-fetches its remote OPML on every refresh cycle (`FreshRSS_Category::refreshDynamicOpml()`) and applies the imported feeds through `FreshRSS_Import_Service::importOpml()`.

That remote content is not explicitly provided by the user, so this PR adds a `trusted_source` flag to `importOpml()`:

- feed cURL parameters from `frss:CURLOPT_*` attributes are ignored for remote content, consistent with the export-side policy introduced in #9070 (public OPML exports already strip these parameters);
- `frss:opmlUrl` no longer creates further dynamic OPML categories from remote content (defense in depth; the forced-category path of `refreshDynamicOpml()` currently already skips category creation).

Imports from a local file (Settings → Import) keep the current behavior: they are an explicit user action, so feed cURL parameters and dynamic OPML categories keep working there.

cc @Alkarex @Inverle @Frenzie

### DIFF
--- a/app/Services/ImportService.php
+++ b/app/Services/ImportService.php
@@ -32,8 +32,11 @@ class FreshRSS_Import_Service {
 	 * @param string $opml_file the OPML file content.
 	 * @param FreshRSS_Category|null $forced_category force the feeds to be associated to this category.
 	 * @param bool $dry_run true to not create categories and feeds in database.
+	 * @param bool $trusted_source true when the OPML content is explicitly provided by the user (e.g. local file import);
+	 *        false by default, and for content fetched from a remote source (e.g. dynamic OPML categories), in which case
+	 *        security-relevant attributes (feed cURL parameters, nested dynamic categories) are ignored.
 	 */
-	public function importOpml(string $opml_file, ?FreshRSS_Category $forced_category = null, bool $dry_run = false): void {
+	public function importOpml(string $opml_file, ?FreshRSS_Category $forced_category = null, bool $dry_run = false, bool $trusted_source = false): void {
 		if (function_exists('set_time_limit')) {
 			@set_time_limit(300);
 		}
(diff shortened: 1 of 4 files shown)


app/Services/ImportService.php
```diff
@@ -32,8 +32,11 @@ public function lastStatus(): bool {
 	 * @param string $opml_file the OPML file content.
 	 * @param FreshRSS_Category|null $forced_category force the feeds to be associated to this category.
 	 * @param bool $dry_run true to not create categories and feeds in database.
+	 * @param bool $trusted_source true when the OPML content is explicitly provided by the user (e.g. local file import);
+	 *        false for content fetched from a remote source (e.g. dynamic OPML categories), in which case
+	 *        security-relevant attributes (feed cURL parameters, nested dynamic categories) are ignored.
 	 */
-	public function importOpml(string $opml_file, ?FreshRSS_Category $forced_category = null, bool $dry_run = false): void {
+	public function importOpml(string $opml_file, ?FreshRSS_Category $forced_category = null, bool $dry_run = false, bool $trusted_source = true): void {
```
Frenzie: Is it supposed to be true by default?

aligundogar: Good point — flipped in 2b2349d85: `trusted_source` now defaults to `false`, so OPML content is treated as untrusted unless the caller explicitly opts in.

The only opt-in call site is the local file import controller (`importExportController.php`), where the content is explicitly provided by the user; `refreshDynamicOpml()` keeps the explicit untrusted default. This way a future caller that forgets the flag fails safe.

Alkarex:
Thanks! I have made a few changes. Double-check welcome

p/api/greader.php
```diff
@@ -325,7 +325,7 @@ private static function subscriptionExport(): never {
 	private static function subscriptionImport(string $opml): never {
 		$user = Minz_User::name() ?? Minz_User::INTERNAL_USER;
 		$importService = new FreshRSS_Import_Service($user);
-		$importService->importOpml($opml);
+		$importService->importOpml($opml, trusted_source: true);
```
Alkarex, line 328: Changed here in particular. Double-check welcome

PR #9278, opened 2026-09-09
Log invalid challenge and ignore JSON error during login

Inverle:
Fixes some unexpected behaviors during a user login, including strange errors in the log related to JSON.

### DIFF
--- a/app/Models/FormAuth.php
+++ b/app/Models/FormAuth.php
@@ -11,6 +11,11 @@ class FreshRSS_FormAuth {
 			return false;
 		}
 
+		if (!preg_match('/^\$2[aby]\$(0[4-9]|10)\$[.\/0-9A-Za-z]{53}$/', $challenge)) {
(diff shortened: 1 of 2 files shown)


app/Models/FormAuth.php
```diff
@@ -11,6 +11,11 @@ public static function checkCredentials(string $username, string $hash, string $
 			return false;
 		}
 
+		if (!preg_match('/^\$2[aby]\$(0[4-9]|10)\$[.\/0-9A-Za-z]{53}$/', $challenge)) {
```
Frenzie, line 15: Where's the challenge format defined? In isolation this seems a bit mysterious.

Alkarex: We could add a link to https://en.wikipedia.org/wiki/Bcrypt in a comment

Inverle: Done

Alkarex (approved):

PR #9277, opened 2026-09-09
Limit the length and parentheses nesting depth of a search query

AdamKorcz:
Closes #

Changes proposed in this pull request:

-
-
-

How to test the feature manually:

1.
2.
3.

Pull request checklist:

- [ ] clear commit messages
- [ ] code manually tested
- [ ] unit tests written (optional if too hard)
- [ ] documentation updated

[Additional information can be found in the documentation](https://freshrss.github.io/FreshRSS/en/developers/04_Pull_requests.html).

PR #9280, opened 2026-09-09
Config + increase default values for search max length and depth

Alkarex:
Fix https://github.com/FreshRSS/FreshRSS/discussions/9279
Follow-up of https://github.com/FreshRSS/FreshRSS/pull/9277

Values can be overridden in `data/config.php`, or in `./data/config.custom.php` before the install process

### DIFF
--- a/constants.php
+++ b/constants.php
@@ -59,10 +59,6 @@ defined('MAX_LOG_SIZE') or define('MAX_LOG_SIZE', 1048576);
 // Amount of characters of text shown if feed has no title
 defined('MAX_CHARS_EMPTY_FEED_TITLE') or define('MAX_CHARS_EMPTY_FEED_TITLE', 75);
 
-// Limits for searches
-defined('MAX_SEARCH_LENGTH') or define('MAX_SEARCH_LENGTH', 16384);
-defined('MAX_SEARCH_PARENTHESES_DEPTH') or define('MAX_SEARCH_PARENTHESES_DEPTH', 32);
-
 //This directory must be writable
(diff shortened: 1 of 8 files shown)


constants.php
```diff
@@ -59,6 +59,10 @@
 // Amount of characters of text shown if feed has no title
 defined('MAX_CHARS_EMPTY_FEED_TITLE') or define('MAX_CHARS_EMPTY_FEED_TITLE', 75);
 
+// Limits for searches
+defined('MAX_SEARCH_LENGTH') or define('MAX_SEARCH_LENGTH', 16384);
+defined('MAX_SEARCH_PARENTHESES_DEPTH') or define('MAX_SEARCH_PARENTHESES_DEPTH', 32);
+
```
Inverle: Could be more convenient to edit in `data/config.php` under `limits`, or that's what I'm used to, at least.

I guess it's easier to see what options were edited under `constants.local.php`, but the locations are not consistent for all settings.

Inverle: Though this is better if we intend to change the defaults in the future, so fine as is.

Alkarex: That would indeed probably be just as fine, if not better, since searches always require the system configuration to be fully loaded anyway. I cannot do it tonight though

Alkarex: Done: https://github.com/FreshRSS/FreshRSS/pull/9280/commits/36ab12a11c721f0a0c878bbd9e09150f9bd6306b

Inverle:
Should update the PR body with the new config location

Alkarex:
> Should update the PR body with the new config location

Done

PR #9282, opened 2026-09-10
Fix infinite redirect loop due to `SCRIPT_NAME` in `PATH_INFO`

### LINKED ISSUE #9281: [Bug] ERR_TOO_MANY_REDIRECTS with 1.30.0
### Describe the bug

Clean install of 1.30.0 on my shared hosting provider (ionos), as soon as I hit 'Complete Installation' (using sqlite as db) all I can get are ERR_TOO_MANY_REDIRECTS error. Clearing cookies does not help.

1.29.1 worked.

### To Reproduce

1. Install on shared-hosting

### Expected behavior

_No response_

### FreshRSS version

1.30.0

### System information

- Database version: SQLite
- PHP version: 8.5.10
- Installation type: shared host
- Web server type: Apache
- Device:  unknown
- OS: Debian
- Browser: Chrome, Edge, Firefox

### Additional context

_No response_

Inverle:
Closes #9281 
Regression from #9075

Not sure what kind of configuration causes this issue yet, since it's on a shared hosting provider.

### DIFF
--- a/lib/Minz/Request.php
+++ b/lib/Minz/Request.php
@@ -283,6 +283,25 @@ class Minz_Request {
 	}
 
 	/**
+	 * Returns `PATH_INFO` with `SCRIPT_NAME` stripped from the beginning of it,
+	 * if it's there on some shared hosting configurations.
+	 */
+	public static function pathInfo(): string {
+		$pathInfo = $_SERVER['PATH_INFO'] ?? $_SERVER['ORIG_PATH_INFO'] ?? '';
+		if (!is_string($pathInfo)) {
+			$pathInfo = '';
+		}
+		$scriptName = $_SERVER['SCRIPT_NAME'] ?? '';
+		if (!is_string($scriptName)) {
+			$scriptName = '';
+		}
+		if ($pathInfo !== '' && $scriptName !== '' && str_starts_with($pathInfo, $scriptName)) {
+			$pathInfo = substr($pathInfo, strlen($scriptName));
+		}
+		return $pathInfo;
+	}
+
+	/**
 	 * Return true if the request is over HTTPS, false otherwise (HTTP)
 	 */
 	public static function isHttps(): bool {
--- a/lib/lib_rss.php
+++ b/lib/lib_rss.php
@@ -28,6 +28,25 @@ if (function_exists('openlog')) {
 }
 
 /**
+ * Returns `PATH_INFO` with `SCRIPT_NAME` stripped from the beginning of it,
+ * if it's there on some shared hosting configurations.
+ */
+function get_path_info(): string {
+	$pathInfo = $_SERVER['PATH_INFO'] ?? $_SERVER['ORIG_PATH_INFO'] ?? '';
+	if (!is_string($pathInfo)) {
+		$pathInfo = '';
+	}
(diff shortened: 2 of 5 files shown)


lib/lib_rss.php
```diff
@@ -27,6 +27,25 @@ function syslog(int $priority, string $message): bool {
 	}
 }
 
+/**
+ * Returns `PATH_INFO` with `SCRIPT_NAME` stripped from the beginning of it,
+ * if it's there on some shared hosting configurations.
+ */
+function get_path_info(): string {
```
Alkarex: If used in Minz, should best be moved under Minz

Alkarex:
I have moved the new function under `Minz_Request::pathInfo()`. Still good?

Inverle:
That works

lib/Minz/Request.php
```diff
@@ -282,6 +282,25 @@ public static function connectionRemoteAddress(): string {
...
+		}
+		$scriptName = $_SERVER['SCRIPT_NAME'] ?? '';
+		if (!is_string($scriptName)) {
+			$scriptName = '';
+		}
+		if ($pathInfo !== '' && $scriptName !== '' && str_starts_with($pathInfo, $scriptName)) {
+			$pathInfo = substr($pathInfo, strlen($scriptName));
+		}
+		return $pathInfo;
+	}
```
Alkarex, line 302: Might be sufficient, then:

```suggestion
	/**
	 * Returns `PATH_INFO` accounting for a bug when sometimes it may contain `SCRIPT_NAME` value instead of blank
	 */
	public static function pathInfo(): string {
		$pathInfo = $_SERVER['PATH_INFO'] ?? $_SERVER['ORIG_PATH_INFO'] ?? null;
		if (!is_string($pathInfo)) {
			return '';
		}
		$scriptName = $_SERVER['SCRIPT_NAME'] ?? null;
		if ($scriptName === $pathInfo) {
			return '';
		}
		return $pathInfo;
	}
```

https://github.com/FreshRSS/FreshRSS/issues/9281#issuecomment-5618617227

Inverle: I thought about doing that, but what if there are cases with `SCRIPT_NAME` always being at the beginning in `PATH_INFO`, instead of only when it should be blank?

Inverle: It maybe makes sense with `ORIG_PATH_INFO` though: https://www.php.net/manual/en/reserved.variables.server.php

> 'ORIG_PATH_INFO'
>    Original version of 'PATH_INFO' before processed by PHP.

Alkarex: Simplification https://github.com/FreshRSS/FreshRSS/pull/9287

Alkarex: > what if there are cases with `SCRIPT_NAME` always being at the beginning in `PATH_INFO`, instead of only when it should be blank?

I think we should address it if actually observed. Those are configuration bugs. My guess is that it happens only when `PATH_INFO` is blank

Inverle: Note that `PATH_INFO` was not being provided there, only `ORIG_PATH_INFO`.
But I guess it is fine, since we're only using the function to check for empty values at the moment.

PR #9283, opened 2026-09-10
Fix documentation of TRUSTED_PROXY syntax and document limitations.

### LINKED ISSUE #9210: [Bug] IPs without /subnet and domain names in TRUSTED_PROXY are ignored for REMOTE_USER purposes
### Describe the bug

With env `TRUSTED_PROXY=192.168.1.1`, or `TRUSTED_PROXY=gateway.localdomain`, where `gateway.localdomain` resolves via DNS to `192.168.1.1`, FreshRSS does not trust headers like `Remote-User` from `192.168.1.1` (the proxy). This behavior is inconsistent with Apache. FreshRSS only works with `192.168.1.1/32` with an explicit `/32` suffix.

### To Reproduce

1. A DNS entry is needed to repro the second part. Without it, the best way is to edit `/etc/hosts` to add `192.168.1.1 gateway.localdomain` (or the Docker equivalent)
2. Run FreshRSS with env var `TRUSTED_PROXY=192.168.1.1` or `TRUSTED_PROXY=gateway.localdomain`. This can be done with Docker env or Kubernetes
3. Set up a reverse proxy on 192.168.1.1, or whatever internal IP, in front of FreshRSS, with authentication that sends `Remote-User: foo` to FreshRSS. Authelia can do this.
4. In FreshRSS admin settings, change authentication method to HTTP headers (external authentication)
5. After authenticating with the proxy, FreshRSS shows:

```
Error 403 - Forbidden

You don’t have permission to access this page [HTTP Remote-User=foo ; Remote IP address=192.168.1.1] 
```
6. In Apache logs (stderr), observe that the real IP address is logged for each request, instead of 192.168.1.1.

Honestly, way too many steps to repro. I think it's probably possible to take some shortcuts by carefully constructing a `curl` command that sends a request with `Remote-User` header set. With enough fiddling with internal / public IPs, it's probably possible to make the request appear to come from an IP address other than 127.0.0.1, thus reproducing the issue without the trouble of setting up Authelia at all. I've also pinpointed the root cause below if you want to skip the trouble.

### Expected behavior

I expect the request to succeed and `Remote-User` to be accepted. In fact, setting `TRUSTED_PROXY=192.168.1.1/32` does exactly that but requires hard-coding the IP address, making it inconvenient if the proxy IP is not fixed. Setting `TRUSTED_PROXY=192.168.1.1` looks like it should work, but it doesn't.

My expectation came from this sentence from documentation `./docs/en/admins/09_AccessControl.md`:

> You may alternatively pass a `TRUSTED_PROXY` environment variable in a format compatible with [Apache’s `mod_remoteip` `RemoteIPInternalProxy`](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy).

Apache's linked documentation gave this example, among others: `RemoteIPInternalProxy gateway.localdomain`. And Apache itself works just fine with that -- see repro step 6 above. Apache documentation also define the syntax as `	RemoteIPInternalProxy proxy-ip|proxy-ip/subnet|hostname ...` -- see how `proxy-ip` is called out as an alternative to `proxy-ip/subnet`. But FreshRSS code only parses the variant with an explicit `/subnet` part.

### FreshRSS version

Tested on both 1.29.1 and 1.30.0 but there may be other affected versions

### System information

- PHP version: PHP 8.4.21
- Installation type: Docker
- Web server type: Apache
- Device: All client devices
- OS: Any
- Browser: All browsers

### Additional context

Internally, Apache [runs a DNS lookup to convert domains to IP addresses](https://github.com/apache/httpd/blob/dec51f1f869a8ebd7b805c6900b419f802f09680/modules/metadata/mod_remoteip.c#L251-L264) at config load time.

Such logic isn't present in FreshRSS. Instead, FreshRSS [assumes each entry is a CIDR](https://github.com/FreshRSS/FreshRSS/blob/24039afb10ae6c073682eb5fd680a155ab52e452/app/Utils/httpUtil.php#L828-L832) and attempts to parse them that way. The parsing fails silently, ignoring domain names and IPs without slashes.

It took me quite a while to figure out that FreshRSS has its own parsing logic despite linking to Apache docs and writing Apache config files for other reasons. I didn't realize FreshRSS didn't simply trust the remote IP from Apache mod_remoteip and spent a lot of time staring at Apache logs and thinking "but this is the right remote IP and Apache mod_remoteip did not log the proxy as untrusted".

If this is worth noting, I can draft a PR to add one sentence about this caveat to said docs page. Most use cases can be worked around with hard-coded CIDRs. The hard part is realizing that this limitation exists.

Also, adding a warning when FreshRSS fails to parse TRUSTED_PROXY seems in general a good idea.

I'm personally rusty in PHP and don't know how to handle the domain name properly even if we decide we want to match Apache2 behavior. Running DNS lookup *per request* sounds like a bad idea so the fix won't be as simple as modifying the linked loop. Caching the resulting IPs require a much bigger change. I bet I'm also missing a lot of context as to why FreshRSS doesn't simply trust Apache's own handling of `RemoteIPInternalProxy`. I did search the issue tracker and found nothing beyond the introduction of the `TRUSTED_PROXY` env and the fact that parsing it seems hard.

yuchenshi:
Closes #9210

- Updated documentation on `TRUSTED_PROXY` syntax.
- No code or behavior changes.

I've manually tried domain names. Then IPs without `/`. Either way, requests got rejected on my server. Then I looked at the code, as described in #9210 investigation steps, and concluded that they are WAI (working as _implemented_ , not sure if _intended_ ). I don't particularly like the existing behavior (failures with no debuggable warnings or logs) but I try to document it as-is without judgement.

Pull request checklist:

- [x] clear commit messages
- [x] code manually tested
- [ ] unit tests written (optional if too hard) -- did not write for docs
- [x] documentation updated

[Additional information can be found in the documentation](https://freshrss.github.io/FreshRSS/en/developers/04_Pull_requests.html).

### DIFF
--- a/docs/en/admins/09_AccessControl.md
+++ b/docs/en/admins/09_AccessControl.md
@@ -62,7 +62,8 @@ To enable this feature, you need to add the IP range (in CIDR notation) of your 
 To allow only one IPv4, you can use a `/32` like this: `trusted_sources => [ '192.168.1.10/32' ]`.
 Likewise to allow only one IPv6, you can use a `/128` like this: `trusted_sources => [ '::1/128' ]`.
 
-You may alternatively pass a `TRUSTED_PROXY` environment variable in a format compatible with [Apache’s `mod_remoteip` `RemoteIPInternalProxy`](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy).
+You may alternatively pass a `TRUSTED_PROXY` environment variable, which is a space-separated list of CIDR notations like this: `192.168.1.10/32 ::1/128`.
+Unlike [Apache’s `mod_remoteip` `RemoteIPInternalProxy`](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy) syntax, domain names or a single IP missing a `/32` or `/128` suffix is not supported and will silently cause requests to be rejected as unauthorized.
 
 > ☠️ WARNING: FreshRSS will trust any IP configured in the `trusted_sources` option, if your proxy isn’t properly secured, an attacker could simply attach this header and get admin access.
 
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -352,7 +352,7 @@ services:
 
       # Optional parameter, remove for automatic settings, set to 0 to disable,
       # or (if you use a proxy) to a space-separated list of trusted IP ranges
-      # compatible with https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy
+      # in CIDR notation. Write `/32` or `/128` explicitly to allowlist a single IPv4 or v6 address.
       # This impacts which IP address is logged (X-Forwarded-For or REMOTE_ADDR).
       # This also impacts external authentication methods;
       # see https://freshrss.github.io/FreshRSS/en/admins/09_AccessControl.html


docs/en/admins/09_AccessControl.md
```diff
@@ -62,7 +62,8 @@ To enable this feature, you need to add the IP range (in CIDR notation) of your
 To allow only one IPv4, you can use a `/32` like this: `trusted_sources => [ '192.168.1.10/32' ]`.
 Likewise to allow only one IPv6, you can use a `/128` like this: `trusted_sources => [ '::1/128' ]`.
 
-You may alternatively pass a `TRUSTED_PROXY` environment variable in a format compatible with [Apache’s `mod_remoteip` `RemoteIPInternalProxy`](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy).
+You may alternatively pass a `TRUSTED_PROXY` environment variable, which is a space-separated list of CIDR notations like this: `192.168.1.10/32 ::1/128`.
+Unlike [Apache’s `mod_remoteip` `RemoteIPInternalProxy`](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy) syntax, domain names or a single IP missing a `/32` or `/128` suffix is not supported and will silently cause requests to be rejected as unauthorized.
```
Alkarex, line 66: It could be worth checking if we could easily support the missing part of the syntax

PR #9289, opened 2026-09-10
fix: TRUSTED_PROXY accepts hostnames like mod_remoteip (#9210)

jabrailkhalil:
`#9301` now handles bare IP addresses by treating them as single-host CIDR ranges. This PR is rebased on that change and adds the remaining `mod_remoteip`-compatible behavior: trusted proxy hostnames.

Hostname entries are resolved to validated IPv4 and IPv6 addresses and cached for the current PHP execution. `checkTrustedIP()` uses the same path for IP, CIDR, and hostname entries; IP/CIDR matching delegates directly to the accepted `checkCIDR()` implementation.

The focused PHP 8.4 run passes all 37 `httpUtilTest` cases, including hostname resolution through `localhost`, unresolved hostnames, bare IPs, CIDR ranges, IPv4, IPv6, and invalid input.

Refs #9210


app/Utils/httpUtil.php
```diff
@@ -809,11 +809,78 @@ private static function checkCIDR(string $ip, string $range): bool {
...
+		return false;
+	}
+
 	/**
 	 * Check if the client (e.g. last proxy) is allowed to send unsafe headers.
 	 * This uses the `TRUSTED_PROXY` environment variable or the `trusted_sources` configuration option to get an array of the authorized ranges,
 	 * The connection IP is obtained from the `CONN_REMOTE_ADDR`
 	 * (if available, to be robust even when using Apache mod_remoteip) or `REMOTE_ADDR` environment variables.
+	 * The entries are interpreted like Apache's `RemoteIPInternalProxy`:
+	 * an IP address, an IP/CIDR range, or a hostname (resolved once per process).
```
yuchenshi: How true is the "resolved once per process" part given the typical Apache setup?

yuchenshi: In other words, how much does the caching help the overall performance? I haven't looked deep enough for this but someone else may know better

jabrailkhalil: You're right: my original 'once per process' wording was too broad for Apache/PHP-FPM. The current head (4f184655) now explicitly says one PHP execution: normally one HTTP request, or one CLI invocation, with no cross-request cache.

The benefit is limited to avoiding repeated hostname/DNS resolution calls for the same hostname if the helper is called multiple times within that execution. It does not save lookups between HTTP requests, and I have not measured an overall performance improvement, so I would not claim a significant one.

Alkarex:
Now that https://github.com/FreshRSS/FreshRSS/pull/9301 is merged, could you please check whether the current PR might be adjusted?

jabrailkhalil:
Adjusted after #9301 in `68408bf6`: the branch is rebased on the accepted bare-IP implementation, duplicate mask handling and tests were removed, and this PR now adds only hostname resolution plus its public-path tests. PHP 8.4 syntax checks and all 37 focused PHPUnit assertions pass.

jabrailkhalil:
The signed-commit blocker is now addressed: both commits on the current head 4ee03186d12cfa44bd4a5618097cc780398cb6a9 are GitHub-verified. I recreated the existing implementation and tests through the GitHub editor, then verified that the entire Git tree is exactly identical to the previous head 3d4ce5de84b944bba9b93685a83e7fe17e8db9f8 (tree e4cc2677f53fd74f795b0b15c610a37585d1fe63).

No source code, assertions, or behavior changed. PHP syntax checks pass for both files; the full test suite was not rerun for this signature-only history replacement. The merge panel no longer reports the unsigned-commit requirement. The required tests and maintainer review are still pending.

PR #9300, opened 2026-09-13
Fit the Docker Hub description in the 25000 character limit

### LINKED ISSUE #9211: [Bug] Docker Hub README description gets truncated

jamalkamaladdin:
Docker/README.md is 28050 bytes. Docker Hub cuts the description at 25000, mid-sentence inside "Option 3) Cron as another instance of the same FreshRSS Docker image".

The workflow now cuts the file at the last heading that fits, appends a link to the full text on GitHub, and passes that copy to the description action. A file under the limit is passed through unchanged.

On the current file the step produces 24915 bytes, cut at "### Option 3)", with code fences balanced.

Closes #9211

### DIFF
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -18,10 +18,30 @@ jobs:
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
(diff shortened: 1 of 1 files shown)


Alkarex:
Hello and thanks for the PR.
We have a couple of duplicates:
https://github.com/FreshRSS/FreshRSS/pull/9256#issuecomment-5505296678

PR #9311, opened 2026-09-16
Warn during session regenerate fail

Alkarex:
Might provide more logs for e.g. https://github.com/FreshRSS/FreshRSS/discussions/9305


lib/Minz/Session.php
```diff
@@ -209,26 +209,30 @@ public static function deleteLegacyCookie(string $name): void {
 
 	/**
 	 * Regenerate a session id.
+	 *
+	 * @return bool True if the session was successfully regenerated, false otherwise (e.g. unwritable session storage)
 	 */
-	public static function regenerateID(string $name): void {
+	public static function regenerateID(string $name): bool {
```
glyn, line 215: It would be more helpful to return a string recording why the session could not be regenerated, or an empty string in the successful case. Then the caller could log the string.

Alkarex: I finally went for an exception instead

PR #9333, opened 2026-09-19
Rotate session ID on all authenticated transitions (CWE-384)

gigioneggiando:

The session ID is regenerated on explicit form login, reauth, logout and password change, but **not** when access is (re)established from a remember-me cookie, via HTTP authentication, or on self-registration auto-login. A fixed/attacker-known session ID can therefore become authenticated (session fixation).

This rotates the session ID at those transitions too, mirroring the existing form-login idiom:
- `FreshRSS_Auth::init()` — when `accessControl()` first grants access (covers remember-me restore and HTTP auth; skipped for `auth_type = none`, which has no authentication boundary);
- self-registration auto-login in `userController`.

It also enables `session.use_strict_mode` before `session_start()` as defence in depth, so an uninitialised (attacker-supplied) session ID is not adopted.


Inverle:
What about the CSRF token, is it rotated sufficiently or not? (after the https://github.com/FreshRSS/FreshRSS/pull/9171 changes)

app/Models/Auth.php
```diff
@@ -34,9 +34,22 @@ public static function init(): bool {
 		if (self::$login_ok && self::giveAccess()) {
 			return self::$login_ok;
 		}
-		if (self::accessControl() && self::giveAccess()) {
-			FreshRSS_UserDAO::touch();
-			return self::$login_ok;
+		if (self::accessControl()) {
+			// Rotate the PHP session ID on the unauthenticated->authenticated
+			// transition (remember-me cookie restore, HTTP auth, HTTP auto-register)
+			// to prevent session fixation, mirroring the explicit form-login path.
```
Inverle: > mirroring the explicit form-login path.

It's not an important detail here, since that's what should be being done anyway.

Inverle:
I'm having the `FreshRSS` session cookie being sent twice during the remember-me auto login with this PR, which should be avoided.

Inverle (changes requested):

gigioneggiando:
Addressed the requested changes in 5baf803a4. During session regeneration, an already queued FreshRSS session cookie is now replaced while preserving other Set-Cookie headers (including FreshRSS_login), so remember-me auto-login emits exactly one FreshRSS cookie. I reproduced the flow locally over HTTP and confirmed one FreshRSS plus one FreshRSS_login header. Regarding CSRF: the token is rotated on each newly covered authenticated transition. Remember-me and HTTP auth clear csrf in accessControl(), and self-registration clears it when installing the authenticated session parameters; the next csrfToken() call generates a fresh value, consistent with #9171. I also removed the unnecessary comparison wording from the inline comment. Focused SessionTest, full PHPCS, and full PHPStan pass.

PR #9334, opened 2026-09-19
Make the login challenge nonce one-time (CWE-294)

gigioneggiando:

The per-login challenge nonce (`nonceAction` → session `nonce`) was never consumed or cleared, so a captured challenge stayed valid for reuse and survived logout (capture-replay).

The client (`p/scripts/extra.js`) already requests a fresh nonce for every login attempt, so making it one-time doesn't affect legitimate retries. This:
- consumes (unsets) the nonce as soon as it is read for verification — in form login, reauth, password change and account self-deletion;
- clears it in `FreshRSS_Auth::removeAccess`, so logout invalidates any outstanding challenge.

Per-form binding + short expiry would be a reasonable further hardening; this PR keeps the change minimal and focused on the one-time/replay issue.


jamalkamaladdin (commented):
One measured note below.

app/Models/Auth.php
```diff
@@ -168,6 +168,7 @@ public static function removeAccess(): void {
 			'lastReauth' => false,
 			// 'csrf' => false, // Must be refreshed separately
 			'REMOTE_USER' => false,
+			'nonce' => false,	// Invalidate any outstanding login challenge (e.g. on logout)
```
jamalkamaladdin: `FreshRSS_Auth::init()` runs on every request through `FreshRSS.php::initAuth()`. For a request with no session and no remember me cookie it reaches `removeAccess()`, which unsets this key before `formLoginAction` reads it at line 121.

Measured on this head sha with a SQLite install, `auth_type=form`, nonce fetched through `c=javascript&a=nonce`, then a login POST:

with this line, HTTP 403 and `Invalid session during login for user=admin, nonce=`, password never reached

with this line removed, `Password mismatch for user=admin, nonce=31c2d8ec...`

```suggestion
```

gigioneggiando:
Good catch, thank you — you're right: `removeAccess()` runs on the pre-login request via `FreshRSS_Auth::init()` and wiped the nonce before `formLoginAction` read it. Removed that line in bb202d7. The one-time guarantee is preserved by consuming the nonce when it's read for verification (login / reauth / password change / self-deletion), so logout no longer needs a separate clear.

PR #9335, opened 2026-09-19
Require POST+CSRF for self-update mutations; guard post-config step (CWE-352)

gigioneggiando:

The self-update `check`/`apply` mutations were triggered by GET links, and GET is never anti-CSRF validated (the token is only checked on POST). The privileged post-configuration step (`apply&post_conf=1` → `do_post_update()` / `PostUpdate` hook / removal of the staged update) was additionally reauth-exempt and reachable by a cross-site top-level GET navigation.

Changes:
- `checkAction` and the initial `apply` branch now require POST, so the global anti-CSRF check applies. The update page's **Check** / **Apply** controls become POST forms carrying `_csrf` (loading-button behaviour preserved).
- The `post_conf` finalisation is bound to a **one-time session marker** set by the successful apply right before its internal redirect, so it can only run as that continuation — not via a direct cross-site GET, and independently of the reauth setting.
- Reauth is now applied consistently (the `apply`+`post_conf` exemption is removed); the legitimate internal redirect still passes on the reauth performed moments earlier by the apply request.

Note on the apply flow: with apply now POST-only, `save_info_update()` runs on the apply POST (it was previously gated on POST vs. the GET trigger). For updates that don't need an info step this is a no-op; for those that do, the info form (`ask_info_update()`) already POSTs with the token and still drives `need_info_update()`/`apply_update()` as before. Verified against phpcs + phpstan (level 10).


Inverle:
Related: [GHSA-4f8m-6r45-85qq](https://github.com/FreshRSS/FreshRSS/security/advisories/GHSA-4f8m-6r45-85qq)

PR #9336, opened 2026-09-19
Reject token access (RSS/OPML export, feed refresh) for disabled accounts (CWE-613)

gigioneggiando:

`Minz_Request::tokenIsOk()` validated the per-user token against the user configuration but did not check that the account is **enabled**. A disabled account's token therefore still authorised token-based endpoints — RSS/OPML export (`c=index&a=rss|opml`) and feed refresh (`c=feed&a=actualize`).

Session-based access already enforces `enabled` (in `FreshRSS_Auth::giveAccess()`); this applies the same check in token validation. Re-enabling the account restores token access as before, so the existing token stays valid (no rotation needed).


Alkarex (approved):

Alkarex:
Thanks!

gigioneggiando:
Thanks — added in #9338!

